Issue #2421, opened 2022-04-05
NFT bridge between L1 and Optimism


sam-goldman-zz:
The purpose of this issue is to solicit feedback on an initial proposal for an ERC721 bridge between L1 and Optimism. The bridge allows owners of L1 NFTs to make deposits into L2 and withdrawals back to L1 in the same manner that ERC20s can be transferred using the [Standard Token Bridge](https://community.optimism.io/docs/developers/bridge/standard-bridge/#).

The relevant code can be found in [this PR](https://github.com/ethereum-optimism/optimism/pull/2577) or in the standalone repo [here](https://github.com/sam-goldman/optimism-nft-bridge). All contracts have been tested locally. For more information regarding the bridge, check out [these slides](https://docs.google.com/presentation/d/e/2PACX-1vQDgyIjiWeXjkR3Yq-DDZSnyGqvxQ5efrY5WrwgRg7YjumPegQzYMaGwOHtYK_R5SGw6EGZT6_CGju1/pub?start=false&loop=false&delayms=3000&slide=id.g120f23c86d9_0_0) (credit goes to Mark!)

The design of the ERC721 Bridge is almost exactly the same as the Standard Token Bridge, with one main difference: token URIs are transferred off-chain using a permissioned oracle. The reason for this design choice is mostly due to the cost of transferring a token URI from L1 to L2 on-chain, which can be expensive since it is a large string. Also, it is common for NFT projects to change token URIs for various reasons, which means users would need to pay gas fees every time they want to to update their L2 token URI. We view this as a temporary solution until it is possible to read L1 state from L2, which would allow us to permissionlessly view tokenURIs from L2 without using an oracle. To be clear, only token URIs would be transferred off-chain; token IDs would be sent on-chain.

We'd love to hear the community's feedback on the design of the bridge, as well as the following open questions:
- What is the best way to handle the transfer of token URIs from L1 to L2?
- Should there be a token list to maintain a source of truth for high-profile NFT projects, similar to [Optimism's ERC20 Token List](https://github.com/ethereum-optimism/ethereum-optimism.github.io/blob/master/optimism.tokenlist.json)?
  - Who would maintain this token list?

annieke:
the optimism team (@maurelian @bdresser me) and quixotic (@sam-goldman @markdawson) met about this yesterday, and we have some open questions for reviewers!

1. How should we deal with TokenURIs?
[...]

PR #2577, opened 2022-05-18
feat: nft bridge betwen l1 and optimism

sam-goldman:
**Description**
An ERC721 bridge between Mainnet and Optimism. The issue describing the bridge can be found [here](https://github.com/ethereum-optimism/optimism/issues/2421).

After the PR has been reviewed, the address of the L2ERC721Bridge needs to be updated from its placeholder address (0xA77...) [here](https://github.com/sam-goldman/optimism/blob/eth-nft-bridge/packages/contracts/test/contracts/L2/messaging/L2StandardERC721Factory.spec.ts#L14) and [here](https://github.com/sam-goldman/optimism/blob/eth-nft-bridge/packages/contracts/contracts/L2/messaging/L2StandardERC721Factory.sol#L16). Once this is done, the PR will be ready to merge.


sam-goldman:
@tynes It's all ready for review!

packages/contracts/contracts/L2/messaging/L2StandardERC721Factory.sol
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @title L2StandardERC721Factory
+ * @dev Factory contract for creating standard L2 ERC721 representations of L1 ERC721s
+ * compatible with and working on the NFT bridge.
+ */
+contract L2StandardERC721Factory {
+    event StandardL2ERC721Created(address indexed _l1Token, address indexed _l2Token);
+
+    // Address of the L2 ERC721 Bridge. It currently exists only on Optimistic Kovan.
+    address public constant L2_ERC721_BRIDGE = 0xA779A0cA89556A9dffD47527F0aad1c2e0d66e46;
```
tynes, line 16: This should be turned into a constructor argument

packages/contracts/contracts/L2/messaging/L2StandardERC721Factory.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    // Maps an L2 ERC721 token address to a boolean that returns true if the token was created
+    // with the L2StandardERC721Factory.
+    mapping(address => bool) public isStandardERC721;
```
tynes, line 20: What is this mapping for exactly? Because developers can deploy custom L2 ERC721s that implement the correct interface and not be included in this mapping. Is is supposed to signify a "token list"?

packages/contracts/contracts/L2/messaging/L2StandardERC721Factory.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    // Maps an L1 ERC721 token address to an L2 Standard ERC721 token address. This mapping can
+    // only be updated once per L1 ERC721 token.
+    mapping(address => address) public standardERC721Mapping;
```
tynes, line 24: Same comment as above - I feel like an offchain tokenlist is better for this because using this factory is permissionless so its not the best indicator of legitimacy

packages/contracts/contracts/standards/IL2StandardERC721.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import { IERC721 } from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+
+interface IL2StandardERC721 is IERC721 {
+    function l1Token() external returns (address);
+
+    function mint(address _to, uint256 _amount) external;
```
tynes, line 9: The `_amount` args in this file should be `_tokenId`

packages/contracts/contracts/standards/L2StandardERC721.sol
```diff
@@ -0,0 +1,80 @@
...
+        l1Token = _l1Token;
+        l2Bridge = _l2Bridge;
+
+        // Creates a base URI in the format specified by EIP-681:
+        // https://eips.ethereum.org/EIPS/eip-681
+        baseTokenURI = string(
+            abi.encodePacked(
+                "ethereum:0x",
+                Lib_Strings.addressToString(_l1Token),
+                "@1/tokenURI?uint256="
```
tynes, line 36: `block.chainid` should be used instead of hardcoding `1` here so that this can work with testnets and L3s

.changeset/big-hotels-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@eth-optimism/contracts': minor
```
tynes, line 2: This should be a patch release as we do minor releases with each regenesis and a major release with bedrock

packages/contracts/contracts/L2/messaging/L2StandardERC721Factory.sol
sam-goldman: The purpose is just to give frontends an easy way to determine if an L2 ERC721 is a Standard ERC721. In other words, it gives frontends an easy way to tell users, "this L2 ERC721 contract doesn't implement any custom (i.e. potentially dangerous) functionality". The mapping isn't supposed to signify a token list, since that can include custom ERC721s too.

packages/contracts/contracts/L2/messaging/L2StandardERC721Factory.sol
sam-goldman: The purpose of this mapping isn't actually to maintain an on-chain token list. The mapping is there to ensure that there's one, and only one, L2 Standard ERC721 contract for each L1 ERC721 contract (which is enforced [here](https://github.com/ethereum-optimism/optimism/pull/2577/files/e307559010d67a4cc8449ebdf081a682cca5a5ab#diff-d1c461b922bbc3893e80ce8fb6371f7955bd47914f604debaf5ac1633b6a9011R40)). Since these standard contracts implement the exact same functionality, there's no need to have more than one for each L1 contract. Also, this mapping makes it easy for frontends to find out which L2 Standard ERC721 corresponds to an L1 ERC721, which makes it easy to know which L2 contract to bridge to.

maurelian:
Replaced by https://github.com/ethereum-optimism/optimism/pull/2662

guoshijiang (commented):
sdk test erc721 bridge code also need write

PR #2580, opened 2022-05-18
feat: bedrock predeploys

tynes:
**Description**
This PR adds the rest of the predeploys to the bedrock contracts package. The commits are split up to include each contract + their tests. There are no tests for weth9, the contract is a copy/paste from what is currently running in production with tons of value locked.

- `OVM_ETH` was extended to `revert` for `burn` and `mint`
- `SequencerFeeVault` was updated to call the new `L2StandardBridge` - this could be updated to use the periphery dripper contract in theory, not sure if we want to do that since what we have has been working fine.
- `DeployerWhitelist` has the `owner` storage slot set to `address(0)` which turns it off, leaving its bytecode in there for historical reasons, there is a mapping for addresses that were whitelisted before arbitrary contract deployment was turned on
- `forge-std` dependency was updated to the latest version which includes the abi for new cheatcodes, see https://github.com/foundry-rs/foundry/issues/1568 - be sure to `foundryup` to run this locally!


trianglesphere (approved):
lgtm

mslipper (approved):

PR #2583, opened 2022-05-18
feat: L1Block uses average L1 basefee

tynes:
**Description**
This PR adds L1 basefee smoothing. The final numbers are pending but I think its ok to merge this for now and then tweak the numbers later pending internal devnet simulations


contracts-bedrock/contracts/L2/L1Block.sol
```diff
@@ -66,8 +66,17 @@ contract L1Block {
 
         number = _number;
         timestamp = _timestamp;
-        basefee = _basefee;
```
protolambda, line 69: I think it is very confusing to track only our own averaged basefee, and not the actual basefee, while still calling it basefee.

Can we instead create a separate `averageBasefee` field, and track both?

protolambda: Related: the PR description is outdated, it still describes the averaging based on history instead of this new approach.

tynes:
I created some figures based on the formula used in this PR to see how changes in the basefee would impact the average value. You can see that the more skewed the ratio is, the less volatile the changes are

---

`average * (uint256(7) / uint256(10)) + basefee * (uint256(3) / uint256(10))`


`average * (uint256(8) / uint256(10)) + basefee * (uint256(2) / uint256(10))`


`average * (uint256(5) / uint256(10)) + basefee * (uint256(5) / uint256(10))`

contracts-bedrock/contracts/L2/L1Block.sol
smartcontracts: Strongly agree with having `basefee` be real base fee and having a separate averaged fee. I would also ask that the type of average be explicitly named, e.g.,`basefeeRollingAverage`.

trianglesphere:
Leaving comment to update gas usage of the L1 Info Deposit (see https://github.com/ethereum-optimism/optimism/pull/2582). The other option (annoying) is to not meter the L1 Info transaction's gas usage at all (but that does have some nice benefits).

tynes:
@K-Ho - Any progress on the numbers that we can use for this? We could do something that is "good enough" for devnet/testnet and tweak later on if necessary

K-Ho:
> @K-Ho - Any progress on the numbers that we can use for this? We could do something that is "good enough" for devnet/testnet and tweak later on if necessary

Currently working out the best formula to use w Michael. For now let’s just leave the averaging as a TODO and just use the vanilla basefee

packages/contracts-bedrock/contracts/L2/L1Block.sol
```diff
@@ -67,7 +84,17 @@ contract L1Block {
...
         hash = _hash;
         sequenceNumber = _sequenceNumber;
     }
+
+    /**
+     * @notice Compute the running average L1 basefee
+     */
+
+    function _averageBasefee(uint256 _basefee) internal returns (uint256) {
+        return averageBasefee * (avgWeight / uint256(10)) +
```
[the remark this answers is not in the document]
tynes: Ah good catch

tynes:
> Formula's wrong

I'm gonna write the formula as a library and test directly

trianglesphere:
> I'm gonna write the formula as a library and test directly

That's one good solution. The reason that this wasn't caught in the tests is that you compared the formula to itself rather than hardcoded values which would have made it obvious that the formula was not producing numbers we expect.

tynes:
> > I'm gonna write the formula as a library and test directly
> 
> That's one good solution. The reason that this wasn't caught in the tests is that you compared the formula to itself rather than hardcoded values which would have made it obvious that the formula was not producing numbers we expect.

I've turned the logic into a library but its not yet properly tested. I think that I am using too many decimals (1e18), the basefee could technically grow large enough to lose precision. Maybe we should use less decimals (1e3) ?
Also do you think the best way to test this would be to differentially test it against an implementation in another language to ensure that there aren't precision errors?

packages/contracts-bedrock/contracts/test/L1Block.t.sol
```diff
@@ -38,6 +44,11 @@ contract L1BlockTest is CommonTest {
         assertEq(lb.basefee(), 3);
     }
 
+    function test_averageBasefee() external {
+        uint256 expect = Lib_FeeSmoothing.rollingAverage(0, 3);
```
tynes, line 48: This is not sufficient as a test, we need unit tests that are realistic

packages/contracts-bedrock/contracts/test/GasPriceOracle.t.sol
```diff
@@ -81,7 +82,8 @@ contract GasPriceOracle_Test is CommonTest {
 
     function test_l1BaseFee() external {
         uint256 l1BaseFee = gasOracle.l1BaseFee();
-        assertEq(l1BaseFee, 100);
+        uint256 expect = Lib_FeeSmoothing.rollingAverage(0, 100);
```
tynes, line 85: The value here is `30` which I believe is realistic if 70% weighted towards prev value and 30% weighted towards next value. We need unit tests for the `rollingAverage` function, not 100% sure of the best way to go about it

trianglesphere:
> > > I'm gonna write the formula as a library and test directly
> > 
> > 
> > That's one good solution. The reason that this wasn't caught in the tests is that you compared the formula to itself rather than hardcoded values which would have made it obvious that the formula was not producing numbers we expect.
> 
> I've turned the logic into a library but its not yet properly tested. I think that I am using too many decimals (1e18), the basefee could technically grow large enough to lose precision. Maybe we should use less decimals (1e3) ? Also do you think the best way to test this would be to differentially test it against an implementation in another language to ensure that there aren't precision errors?

I don't actually care that much about precision errors, just that the output is close enough.

In terms of decimals, during. the Yuga labs spike, the L1 base fee got up to 8 terra-wei (8e12). A uint256 can hold up to 1e77. So using 18 decimals is totally fine & won't overflow (it makes the effective max base fee 1e59 which is stupidly large).

PR #2591, opened 2022-05-18
Bedrock: use Engine API authentication

protolambda:
This PR:
- adds a `--l2.jwt-secret` flag to point to a file with a secret key for the JWT client authentication. This file should match what the execution engine uses with `--authrpc.jwt-secret`
- uses the JWT rpc code from @trianglesphere  to authenticate the L2 RPC
- removes the `--l2.eth` option, we already have `--l2`. The `eth` namespace is served under the same engine RPC endpoint, and we should just use the first entry in the `--l2` list whenever we need to use the L2 data in the RPC server etc. instead of a driver instance.
- Refactors the node config system a little, so we can pass prepared RPC connections. This enables us to skip authentication, or use an in-process RPC connection to a geth node. In a later PR we can further cleanup the go e2e testing to use more in-process rpc connections instead of websocket connections.
- Updates the go e2e tests
- Updates the devnet

Note: devnet needs to be fixed (see #2584 ) and dependencies need to be updated (see #2589 ) for this PR to really work in CI as expected.


op-e2e/setup.go
```diff
@@ -324,14 +330,28 @@ func (cfg SystemConfig) start() (*System, error) {
...
 
+	// Configure connections to L1 and L2 for rollup nodes.
+	// TODO: refactor testing to use in-process rpc connections instead of websockets.
+	for name, rollupCfg := range cfg.Nodes {
+		rollupCfg.L1 = &rollupNode.L1EndpointConfig{
+			L1NodeAddr: l1Node.WSEndpoint(),
+			L1TrustRPC: false,
+		}
+		rollupCfg.L2s = &rollupNode.L2EndpointsConfig{
+			L2EngineAddrs:      []string{sys.nodes[name].WSEndpoint()},
```
trianglesphere, line 341: Is this the line that needs to be changed to the `AuthEndpoint` for this to work?

trianglesphere, line 341: Is this the line that needs to be changed to the `AuthEndpoint` for this to work?

protolambda:
@trianglesphere I just need to rebase it, now that the geth-update PR is merged. And maybe fix a minor issue or two to use the auth endpoints properly, with the new auth-endpoint getters on the geth node.

op-e2e/setup.go
protolambda: yes, updating it. Upstream geth still allows the engine API to be used through the non-authenticated endpoint, but that will change in the future.

protolambda:
This PR is ready, but depends on #2665 to be merged. So I changed the base to the branch of that PR to isolate the diff for review here. Once that PR is merged the base should become `develop` again, and we can then merge this.

protolambda:
Rebased on `develop`, ready for review.

trianglesphere (approved):

trianglesphere:
Bedrock solidity tests are failing. Assuming unrelated.

maurelian:
> Bedrock solidity tests are failing. Assuming unrelated.

The error is `src/generateProofs.ts(2,29): error TS2307: Cannot find module '@eth-optimism/core-utils' or its corresponding type declarations.`

I'm at a loss as to why that would be happening. It works on my machine™️. 
I hit 'rerun' now 🤞

mslipper (approved):

PR #2594, opened 2022-05-20
feat: op bindings predeploy addresses

tynes:
**Description**
Create a common go package that can be used across the monorepo that holds each predeploy address. This allows us to import this package into any go project that wants to interact with a predeploy contract


op-bindings/predeploys/addresses.go
```diff
@@ -0,0 +1,17 @@
+package predeploys
+
+import "github.com/ethereum/go-ethereum/common"
+
+var (
```
tuxcanfly, line 5: Since these are globals, it would be good to have them as string `const`s imo

tuxcanfly: Callers can instantiate `common.Address`

mslipper (approved):

op-bindings/predeploys/addresses.go
trianglesphere: TBH, any chance to do both (i.e. have const string and address versions). We can also use address literals here - slightly more work, but still const.

trianglesphere: Something like this `var WithdrawalContractAddress2 = common.Address{0x42, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 16}`
Unfortunately you can't make it const. As such, const strings + a set of `common.HexToAddresses` (to avoid the conversion is probably what makes sense)

tynes:
This PR now uses `const` strings that are exported

trianglesphere (approved):

PR #2596, opened 2022-05-22
contracts: optimize `L1Block.setL1BlockValues`

tynes:
**Description**

The amount of gas required to update the
L1 block values can be reduced by ~5000 gas by tightly
packing the `uint64`s into a single storage slot.
This is important because there will be a single
transaction at the beginning of each block that will
be updating these values. ~100 gas is saved by using
yul instead of straight solidity. I don't feel like the
yul is particularly difficult to read in this context,
and saving 100 gas per block will add up to a lot over
the history of the chain. This logic is covered by foundry
fuzzing.


mslipper:
Recusing myself from review because I don't have enough Solidity knowledge to be helpful.

tynes:
I have a preference for this implementation because it is much more clear that the underlying storage slots are being packed

Inphi (commented):
kinda annoying that solc cannot perform such a basic optimization.
Looks good. But I wanna make sure that we have sufficient testing such that things doesn't break if the storage layout changes.

tynes:
`solc` can pack variables into the same storage slot if they are next to each other. I can remove the assembly if highly preferred but it is covered with fuzzing. I prefer this because the packing is very obvious and tests will fail on storage slot changes

packages/contracts-bedrock/contracts/L2/L1Block.sol
```diff
@@ -64,10 +64,16 @@ contract L1Block {
...
         }
 
-        number = _number;
-        timestamp = _timestamp;
-        basefee = _basefee;
-        hash = _hash;
-        sequenceNumber = _sequenceNumber;
+        bytes32 slot;
+        assembly {
+            // _number, _timestamp and _sequenceNumber are all uint64 and will
```
trianglesphere, line 69: If solidity does slot packing on layout & reads, can it do it on writes as well?

tynes: Yup it can, it's slightly more expensive and hides the fact that they are in the same storage slot which I think is the benefit of doing it in yul

trianglesphere (approved):
lgtm. I did look into how solidity does slot packing and I'm happy with the how the yul is storing into the slot.

mslipper (approved):

packages/contracts-bedrock/contracts/L2/L1Block.sol
```diff
@@ -64,10 +64,16 @@ contract L1Block {
...
+            // be tightly packed together in the first storage slot
+            slot := or(slot, _number)
+            slot := or(slot, shl(64, _timestamp))
+            slot := or(slot, shl(128, _sequenceNumber))
+            sstore(0, slot)
+            sstore(basefee.slot, _basefee)
```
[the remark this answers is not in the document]
tynes: By consensus rules the basefee is a uint256, see https://github.com/ethereum/go-ethereum/blob/138f0d74943a054b5c8bb6943e9041404defdf6d/core/types/block.go#L140

PR #2597, opened 2022-05-23
Replica doc fix

qbzzt:
**Description**

Clearer instructions and somewhat easier configuration

### DIFF
--- a/infra/op-replica/docker-compose/README.md
+++ b/infra/op-replica/docker-compose/README.md
@@ -18,7 +18,9 @@ The `data-transport-layer` should run with 1 CPU and 256Mb of memory.
 
 The `l2geth` process should run with 1 or 2 CPUs and between 4 and 8Gb of memory.
 
-With this configuration a synchronization from block 0 to current height is expect to take about 8 hours.
+With this configuration a synchronization from block 0 to current height is expect to take about 8 hours (mid 2022, obviously it will take longer as time passes).
+
+For disk we recomment at least 1 TB, solid state.
 
 ## Software Packages
 
--- a/infra/op-replica/docker-compose/README.md
+++ b/infra/op-replica/docker-compose/README.md
@@ -1,16 +1,14 @@
 # Running a Network Node
 
-This project lets you set up a local replica of the Optimistic Ethereum chain (either the main one or the Kovan testnet). [New
-transactions are submitted either to the sequencer outside of Ethereum or to the Canonical Transaction Chain on
-L1](https://research.paradigm.xyz/optimism#data-availability-batches). To submit transactions via a replica, set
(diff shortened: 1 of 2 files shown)


infra/op-replica/docker-compose/README.md
```diff
@@ -20,74 +18,86 @@ The `data-transport-layer` should run with 1 CPU and 256Mb of memory.
...
-| GCMODE                   | Whether to run l2geth as an `archive` or `full` node     | archive
-| L2GETH_IMAGE_TAG         | L2geth version                                           | 0.5.8 (see below)
-| DTL_IMAGE_TAG            | Data transport layer version                             | latest (see below)
-| HC_IMAGE_TAG             | Health check version                                     | latest (see below)
+| SHARED_ENV_PATH          | Path to a directory containing env files                 | [a directory under .../op-replica](https://github.com/ethereum-optimism/optimism/tree/develop/infra/op-replica/envs)
+| GCMODE                   | [Whether to run l2geth as an `archive` or `full` node](https://www.quicknode.com/guides/infrastructure/ethereum-full-node-vs-archive-node)     | archive
+| L2GETH_IMAGE_TAG         | L2geth version                                           | [Go here](https://hub.docker.com/r/ethereumoptimism/l2geth/tags) and find the latest version. At writing this is 0.5.19 (1).
```
optimisticben, line 70: I really don't like the idea of setting a manual tag version in the README. It updates too often and this change is missed.

We should have a way to publish the current versions supported and let user's choose from the list.

This README should point to that list.

infra/op-replica/docker-compose/README.md
```diff
@@ -20,74 +18,86 @@ The `data-transport-layer` should run with 1 CPU and 256Mb of memory.
...
+| DTL_IMAGE_TAG            | Data transport layer version                             | [Go here](https://hub.docker.com/r/ethereumoptimism/data-transport-layer/tags) and find the latest version. At writing this is 0.5.30 (1).
```
optimisticben, line 71: Same with these tags

infra/op-replica/docker-compose/replica.yml
```diff
@@ -15,7 +15,7 @@ services:
       - ${SHARED_ENV_PATH}/data-transport-layer.env
       - .env
     volumes:
-      - dtl:/db
+      - /var/replica/dtl:/db
```
optimisticben, line 18: I'm fine adding README on how to preserve the data from the replica.
But, this change will clobber whatever the user has on their locale system at the `/var/replica` path.

If you feel strongly that we should default to host paths instead of docker volumes, the paths should be set in the `.env` file for the user to review before it makes changes to their system.

optimisticben (changes requested):

PR #2598, opened 2022-05-23
feat(contracts): add `isOutputFinalized` function

tynes:
**Description**

An `external` function that can be used for apps to
check to see if a particular output is finalized.


contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -148,6 +148,19 @@ contract OptimismPortal {
         emit TransactionDeposited(from, _to, msg.value, _value, _gasLimit, _isCreation, _data);
     }
 
+    /**
+     * @notice Determine if an L2 Output is finalized
+     * @param _l2Timestamp The timestamp of the L2 block
+     */
+
+    function isOutputFinalized(uint256 _l2Timestamp) external view returns (bool) {
```
smartcontracts, line 156: I think the OutputOracle would be a better location for this function. The OutputOracle should have the finalization window baked in because outputs cannot be deleted beyond the finalization window.

smartcontracts: Might require some other minor refactoring

tynes: I agree that it makes more sense for the `OutputOracle` to have this function, but it doesn't have access to the finalization window. I initially wanted to add it there. I do think having this `view` function will be nice for frontends or off chain applications that want to know if an output has been finalized

smartcontracts: Hm. Actually this is a good question. Where the heck should this value live.

tynes: When we refactor things to work off blocknumbers instead of timestamps, it could be a good time to introduce this change. I think it makes sense for the portal to hold the finalization window as its required for withdrawals and we don't want to have to call the output oracle to get it. The only other solution that i can think of is combining the output oracle and the portal but then thats a beefy contract and we will have to deal with its storage

smartcontracts: Word. I'll accept it for now.

smartcontracts: Let's use this function internally in `finalizeWithdrawalTransaction`

tynes: If we use it internally, then we cannot return a bool, we would need to return `(bool, OutputProposal)` because the root is required downstream of the "is finalized" check, otherwise we add extra sloads

tynes: This function is going to need to be refactored for https://github.com/ethereum-optimism/optimism/pull/2707 https://github.com/ethereum-optimism/optimism/pull/2703

We should expose this functionality for end users though, its very useful for offchain applications

maurelian: Those were dupes so 2707 is the only one that matters. Also, I'm pretty sure that  that the only change necessary is to modify the argument name:

```suggestion
    function isOutputFinalized(uint256 _l2BlockNumber) external view returns (bool) {
```

smartcontracts: I see the problem. Feel free to ignore my comment. Not worth the extra SLOADs.

contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -148,6 +148,19 @@ contract OptimismPortal {
...
+        L2OutputOracle.OutputProposal memory proposal = L2_ORACLE.getL2Output(_l2Timestamp);
+        if (proposal.outputRoot == bytes32(uint256(0))) {
+            return false;
+        }
+        return block.timestamp > proposal.timestamp + FINALIZATION_PERIOD;
```
smartcontracts, line 161: Can alternatively combine the above IF statement into this to make a single unified assertion. I'm ok either way.

smartcontracts (approved):
Approving with optional convos left unresolved, feel free to address or simply mark resolved

contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: +1, but also fine with it as is.

maurelian (approved):

maurelian:
I rebased this to fix the conflicts and make it compatible with the recent oracle changes in #2707.
Since I'm one of the approvers, I think we should get another approval from @tynes before merging, so I applied the mergify-ignore label.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -141,6 +141,21 @@ contract OptimismPortal is ResourceMetering {
...
 
+    /**
+     * @notice Determine if an L2 Output is finalized.
+     *
+     * @param _l2BlockNumber The number of the L2 block.
+     */
+
+    function isOutputFinalized(uint256 _l2BlockNumber) external view returns (bool) {
+        L2OutputOracle.OutputProposal memory proposal = L2_ORACLE.getL2Output(_l2BlockNumber);
+        // todo: Should this not return true if the previous check pointed block _is_ finalized?
```
maurelian, line 152: Oops, I mean "the following checkpointed block"

maurelian:
I wrote a little poc for looking up blocknumbers that aren't written to the oracle: 
[`develop...maurelian:feat/is-output-finalized`#diff-db27494cd2](https://github.com/ethereum-optimism/optimism/compare/develop...maurelian:feat/is-output-finalized#diff-db27494cd2a9e5b3dfecf8c9035186f2685e183db7ce4d3b81bfc838773914d6R144-R165)

tynes:
@maurelian This is ready for review again

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -141,6 +141,37 @@ contract OptimismPortal is ResourceMetering {
...
+     *
+     * @param _l2BlockNumber The number of the L2 block.
+     */
+
+    function isOutputFinalized(uint256 _l2BlockNumber) external view returns (bool) {
+        L2OutputOracle.OutputProposal memory proposal = L2_ORACLE.getL2Output(_l2BlockNumber);
+
+        if (proposal.outputRoot == bytes32(uint256(0))) {
+            uint256 interval = L2_ORACLE.SUBMISSION_INTERVAL();
+            uint256 staringBlockNumber = L2_ORACLE.STARTING_BLOCK_NUMBER();
```
maurelian, line 155: typo
```suggestion
            uint256 startingBlockNumber = L2_ORACLE.STARTING_BLOCK_NUMBER();
```

maurelian: I know this was probably my typo originally. :/

maurelian:
Fix typo and lint and should be good.

PR #2616, opened 2022-05-25
fix: update ethers deps

smartcontracts:
**Description**
Updates all ethers dependencies in response to the BN.js bug reported
for versions below 5.8.6. Also removes various extra dependencies to
reduce the total number of deps and simplify the process of updating in
the future.


contracts-bedrock/helpers/utils.ts
```diff
@@ -1,19 +1,7 @@
...
-  hexConcat,
-  stripZeros,
-  zeroPad,
-  arrayify,
-  BytesLike,
-} from '@ethersproject/bytes'
-import { keccak256 } from '@ethersproject/keccak256'
-import { Zero } from '@ethersproject/constants'
-import { ContractReceipt, Event } from '@ethersproject/contracts'
+import { BigNumber, BigNumberish, BytesLike, ContractReceipt, ethers, Event } from 'ethers'
```
smartcontracts, line 1: Using these from ethers directly reduces the number of dependencies, which makes it easier to update dependencies if we ever need to do this again in the future.

tynes (approved):

tynes:
```
Unused devDependencies
* @ethersproject/bytes
```

packages/contracts/package.json
```diff
@@ -98,10 +98,10 @@
     "eslint-plugin-react": "^7.24.0",
     "eslint-plugin-unicorn": "^32.0.1",
     "ethereum-waffle": "^3.3.0",
-    "ethers": "^5.5.4",
```
tynes, line 101: `ethers` is a dev dep?

smartcontracts: dev dev + peer dep

tynes (approved):

mslipper (approved):

PR #2619, opened 2022-05-26
contracts-bedrock: migrate to `packages`

tynes:
**Description**
This PR migrates the `contracts-bedrock` to the `packages` directory. It removes the dependency on `@eth-optimism/contracts` by copy/pasting the code from that package into this package. Longer term, we will be replacing the `contracts` package with `bedrocks-contracts` with a major release, the `bedrock-contracts` package is temporary.

`contracts-bedrock` uses the foundry toolchain for compilation and I have custom code to write the hardhat style artifacts to disk. I plan on upstreaming that code so that foundry compilation will create hardhat style artifacts on disk

The submodules have been removed, and the packages that do not have npm modules are installed from a github repo at a specific commit hash. I'm using a commit hash as releases on github aren't up to date


tynes:
CI is facing a problem

```
Caused by:
  ERROR: You're using an RSA key with SHA-1, which is no longer allowed. Please use a newer client or a different key type.
  Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

Will investigate exactly the reason why, a potential fix is using git to overwrite such that downloads happen over https instead of ssh

.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
   yarn-monorepo:
     docker:
-      - image: ethereumoptimism/js-builder:latest
+      - image: ethereumoptimism/js-builder:nightly
```
tynes, line 66: This might be able to be `latest`

.circleci/config.yml
```diff
@@ -97,7 +98,8 @@ jobs:
             - integration-tests/node_modules
       - run:
           name: Build monorepo
-          command: yarn build
+          command: |
```
tynes, line 101: unneeded

.prettierrc.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
         bracketSpacing: true,
         // These options are specific to the Solidity Plugin
         explicitTypes: 'always',
-        compiler: '0.8.9',
```
tynes, line 20: `^0.8.9`

package.json
```diff
@@ -41,25 +44,39 @@
     "release": "yarn build && yarn changeset publish"
   },
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "^4.26.0",
+    "@babel/eslint-parser": "^7.18.2",
+    "@types/chai": "^4.2.18",
+    "@types/chai-as-promised": "^7.1.4",
+    "@types/mocha": "^8.2.2",
+    "@types/node": "^12.0.0",
+    "@typescript-eslint/eslint-plugin": "^5.26.0",
```
tynes, line 52: yarn depcheck

depcheck: depcheck --config ../../.depcheckrc

Can we extend the depcheck with multiple configs?

PR #2624, opened 2022-05-26
fix(hc): reset metrics on ok connection

smartcontracts:
**Description**
Has replica healthcheck reset connection failure metrics whenever the
connection is successful. This is more useful on the UI standpoint since
we're generally more interested in consecutive failures, not total
failures.

### DIFF
--- a/.changeset/smooth-bugs-sneeze.md
+++ b/.changeset/smooth-bugs-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/replica-healthcheck': patch
+---
+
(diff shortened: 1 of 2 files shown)


optimisticben:
What a user typically cares about is how many events occurred in a time period.

You accomplish this by observing the rate of change on the metric series instead of resetting it.

Something like
```
rate(healthcheck_reference_connection_failures[5m])
```

When the condition causing the errors is resolved, the rate will return to zero.

Inphi:
^^
this can be accomplished on the metrics backend

PR #2628, opened 2022-05-27
proxyd: Handle unexpected JSON-RPC responses

Inphi:
This fixes a bug where the infura backend would be labeled offline because it
returns an unexpected JSON-RPC response. Unexpected, but well-formed,
JSON-RPC response are handled specially. Such errors are surfaced up to
the backend proxier so failover still occurs.

### DIFF
--- a/proxyd/integration_tests/failover_test.go
+++ b/proxyd/integration_tests/failover_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/alicebob/miniredis"
 	"github.com/ethereum-optimism/optimism/proxyd"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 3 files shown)


proxyd/integration_tests/failover_test.go
```diff
@@ -240,3 +242,52 @@ func TestBatchWithPartialFailover(t *testing.T) {
...
+	require.NoError(t, err)
+	defer shutdown()
+
+	res, statusCode, err := client.SendBatchRPC(
+		NewRPCReq("1", "eth_chainId", nil),
+		NewRPCReq("2", "eth_chainId", nil),
+	)
+	require.NoError(t, err)
+	require.Equal(t, 200, statusCode)
+	RequireEqualJSON(t, []byte(asArray(goodResponse, goodResponse)), res)
```
tynes, line 284: What exactly is this line testing? Should it be checking `res` against `goodResponse` instead of `goodResponse` against itself?

Inphi: That's indeed what it's doing, isn't it? There are two `goodResponse`  JSON-RPC objects returned. It's that batch response that's checked against `res`.

tynes (approved):

PR #2629, opened 2022-05-28
fix(fd): start at beginning of challenge period

smartcontracts:
**Description**
Updates the fault-detector to use a smarter starting height by default.
Specifically, the fault-detector will now start at the first batch that
has not yet crossed the challenge period, since the general trust
assumption is that invalid batches will not pass the challenge period.
Significantly reduces verification time, especially since the fault
detector is stateless and does not keep track of the highest verified
batch.


packages/fault-detector/src/service.ts
```diff
@@ -41,7 +42,7 @@ export class FaultDetector extends BaseServiceV2<Options, Metrics, State> {
         },
         startBatchIndex: {
           validator: validators.num,
-          default: 0,
+          default: -1,
```
smartcontracts, line 45: `BaseServiceV2` doesn't cleanly handle "undefined" defaults but -1 is good enough here.

packages/fault-detector/src/service.ts
```diff
@@ -67,19 +68,56 @@ export class FaultDetector extends BaseServiceV2<Options, Metrics, State> {
...
+      let end = (await this.state.scc.getTotalBatches()).toNumber()
+      while (start !== end) {
+        const mid = Math.ceil((start + end) / 2)
+        const filter = this.state.scc.filters.StateBatchAppended(mid)
+        const events = await this.state.scc.queryFilter(filter)
+        const block = await this.options.l1RpcProvider.getBlock(
+          events[0].blockNumber
+        )
+
+        if (block.timestamp > latestBlock.timestamp - fpw) {
```
smartcontracts, line 102: Will converge on index 0 if the chain is less than `fpw` seconds old.

tynes:
How exactly are you testing this?

smartcontracts:
> How exactly are you testing this?

Added tests

packages/fault-detector/package.json
```diff
@@ -32,13 +33,22 @@
...
+    "chai-as-promised": "^7.1.1",
     "dateformat": "^4.5.1",
+    "ethereum-waffle": "^3.4.4",
     "ethers": "^5.6.8",
+    "hardhat": "^2.9.6",
+    "lodash": "^4.17.21",
     "ts-node": "^10.7.0"
   },
   "dependencies": {
+    "@defi-wonderland/smock": "^2.0.7",
```
tynes, line 49: Should this be a devdep? Not that big of a deal

tynes (approved):

mslipper (approved):

PR #2635, opened 2022-05-31
proxyd: Filter out archive requests

Inphi:
**Description**
Change targets proxyd-3.8.4

### DIFF
--- a/go/proxyd/rpc.go
+++ b/go/proxyd/rpc.go
@@ -1,10 +1,13 @@
 package proxyd
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"io/ioutil"
 	"strings"
+
+	"github.com/ethereum/go-ethereum/log"
 )
 
 type RPCReq struct {
@@ -108,7 +111,7 @@ func ParseRPCRes(r io.Reader) (*RPCRes, error) {
 	return res, nil
 }
 
-func ValidateRPCReq(req *RPCReq) error {
+func ValidateRPCReq(req *RPCReq, getBlockNum GetLatestBlockNumFn) error {
 	if req.JSONRPC != JSONRPCVersion {
 		return ErrInvalidRequest("invalid JSON-RPC version")
 	}
@@ -121,6 +124,15 @@ func ValidateRPCReq(req *RPCReq) error {
 		return ErrInvalidRequest("invalid ID")
 	}
 
+	archive, err := isArchiveRequest(req, getBlockNum)
+	if err != nil {
+		// ignore errors and pass em up the request handler chain
+		log.Warn("failed to decode request for archive", "err", err)
+		return nil
+	}
+	if archive {
+		return ErrInvalidRequest("unsupported archive request")
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


go/proxyd/rpc.go
```diff
@@ -152,3 +164,52 @@ func IsBatch(raw []byte) bool {
...
+	return blockNum+uint64(BLOCK_NUM_ARCHIVE_HEIGHT) < latestBlockNum, nil
+}
+
+// quick hack: if the request looks like a request containing a quantity|tag at the end, then return the tag
+func extractTerminalBlockTag(req *RPCReq) (string, error) {
+	var input []json.RawMessage
+	if err := json.Unmarshal(req.Params, &input); err != nil {
+		return "", err
+	}
+	if len(input) < 2 {
```
tynes, line 209: are there any requests that take only a block tag?

Inphi: also note that that function is only [called for eth_call, eth_getBalance and eth_getTransactionCount rpcs ](https://github.com/ethereum-optimism/optimism/pull/2635/files#diff-326039202cfd8c96893c534b0a1be54828dbaf252544893173db3d3d85579448R177)

go/proxyd/rpc.go
```diff
@@ -152,3 +164,52 @@ func IsBatch(raw []byte) bool {
 	}
 	return false
 }
+
+func isArchiveRequest(req *RPCReq, getBlockNum GetLatestBlockNumFn) (bool, error) {
+	const BLOCK_NUM_ARCHIVE_HEIGHT = 256
```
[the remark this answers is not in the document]
tynes: `geth` does hold on to archival state for the last 256 blocks, trailing the tip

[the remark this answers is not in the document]
Inphi: l2geth?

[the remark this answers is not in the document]
tynes: Its fine to reduce it to be safe

tynes:
The code generally looks good to me, the tests are failing which is not ideal

go/proxyd/rpc.go
tynes: all of those take more than 2 elements so we good

go/proxyd/rpc.go
```diff
@@ -152,3 +164,52 @@ func IsBatch(raw []byte) bool {
...
+	// NOTE: we're reading from an in-memory getBlockNum LVC, which doesn't block
+	if latestBlockNum, err = getBlockNum(context.Background()); err != nil {
+		return false, err
+	}
+
+	blockNum, err := decodeBlockInput(tag)
+	if err != nil {
+		return false, err
+	}
+	return blockNum+uint64(BLOCK_NUM_ARCHIVE_HEIGHT) < latestBlockNum, nil
```
tynes, line 200: Should this be `>`?

Inphi: hmm no. We return true (i.e. it's an archive request) if `blockNum` is too old/ too far behind.

PR #2655, opened 2022-06-02
state-surgery: Add state surgery

mslipper:
See the README for an overview.


tynes (approved):

state-surgery/util.go
```diff
@@ -0,0 +1,22 @@
+package state_surgery
+
+import (
+	"fmt"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+func wrapErr(err error, msg string, ctx ...any) error {
+	return fmt.Errorf("%s: %v", fmt.Sprintf(msg, ctx...), err)
```
trianglesphere, line 9:
```suggestion
	return fmt.Errorf("%s: %w", fmt.Sprintf(msg, ctx...), err)
```

nit: Using `%w` enables checking nested errors. Not that important here, but something that I think we should be using more across our go code.

state-surgery/db.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// MustOpenDB opens a Geth database, or panics. Note that
+// the database must be opened with a freezer in order to
+// properly read historical data.
+func MustOpenDB(dataDir string) ethdb.Database {
+	dir := filepath.Join(dataDir, "geth", "chaindata")
+	db, err := rawdb.NewLevelDBDatabaseWithFreezer(
+		dir,
+		0,
```
trianglesphere, line 17: Does increasing the cache here improve performance?

state-surgery/addresses.go
```diff
@@ -0,0 +1,120 @@
...
+func IterateDBAddresses(inDB ethdb.Database, cb AddressCB) error {
+	iter := inDB.NewIterator(AddressPreimagePrefix, nil)
+	for iter.Next() {
+		if iter.Error() != nil {
+			return iter.Error()
+		}
+
+		addr := common.BytesToAddress(bytes.TrimPrefix(iter.Key(), AddressPreimagePrefix))
+		cbErr := cb(addr)
+		if cbErr == ErrStopIteration {
```
trianglesphere, line 42: nit: use errors.Is

state-surgery/addresses.go
```diff
@@ -0,0 +1,120 @@
...
+// calling the callback with the address.
+func IterateAddrList(r io.Reader, cb AddressCB) error {
+	scan := bufio.NewScanner(r)
+	for scan.Scan() {
+		addrStr := scan.Text()
+		if !common.IsHexAddress(addrStr) {
+			return fmt.Errorf("invalid address %s", addrStr)
+		}
+		err := cb(common.HexToAddress(addrStr))
+		if err == ErrStopIteration {
```
trianglesphere, line 62: no else clause here if `cb` errs

state-surgery/cli.go
```diff
@@ -0,0 +1,417 @@
...
+	})
+}
+
+// Migrate performs the actual state migration. It does quite a lot:
+//
+//   1. It uses address lists, allowance lists, Mint events, and address preimages in
+//      the input state database to create a comprehensive list of storage slots in the
+//      OVM ETH contract.
+//   2. It iterates over the slots in OVM ETH, and compares then against the list in (1).
+//      If the list doesn't match, or the total supply of OVM ETH doesn't match the su of
```
trianglesphere, line 55:
```suggestion
//      If the list doesn't match, or the total supply of OVM ETH doesn't match the sum of
```
correct typo fix?

state-surgery/cli.go
```diff
@@ -0,0 +1,417 @@
...
+	if err != nil {
+		return wrapErr(err, "error opening output state DB")
+	}
+
+	// Iterate over the Genesis allocation accounts. These will override
+	// any accounts found in the state.
+	log.Info("importing allocated accounts")
+	logAllocProgress := ProgressLogger(1000, "allocated accounts")
+	for addr, account := range genesis.Alloc {
+		outStateDB.AddBalance(addr, account.Balance)
```
trianglesphere, line 233: Add balance or set balance here? - edit, these are the same b/c balance should always be zero pre-bedrock?

state-surgery/cli.go
```diff
@@ -0,0 +1,417 @@
...
+	rawdb.WriteReceipts(outDB, block.Hash(), block.NumberU64(), nil)
+	rawdb.WriteCanonicalHash(outDB, block.Hash(), block.NumberU64())
+	rawdb.WriteHeadBlockHash(outDB, block.Hash())
+	rawdb.WriteHeadFastBlockHash(outDB, block.Hash())
+	rawdb.WriteHeadHeaderHash(outDB, block.Hash())
+	rawdb.WriteChainConfig(outDB, block.Hash(), genesis.Config)
+	return nil
+
+	return nil
+}
```
trianglesphere, line 387:
```suggestion
}
```

trianglesphere (commented):
Generally looks good. It's missing some integrations with outer level makefiles / CI & testing, but I'm also not sure what level of rigor / long term support we plan for this package.

Also, if 2hrs is too long, running a CPU profile w/ pprof can often identify simple things that make a big difference.

state-surgery/db.go
mslipper: I haven't benchmarked this. It may in theory, so I've added CLI flags to tune this.

state-surgery/cli.go
mslipper: Correct, they are identical since balance should be zero pre-Bedrock. There's a check for that above. Swapped it to `SetBalance` anyway since it's clearer.

PR #2335, opened 2022-03-17
l2geth: verifier sync in parallel with dtl

tonykogias:
Have verifier node sync in parallel with dtl to save time.

**Metadata**
- Fixes #2329

PR #2656, opened 2022-06-02
l2geth: verifier sync in parallel with dtl

### LINKED ISSUE #2329: Have L2Geth sync in parallel with the DTL
First, some background. Right now, the Optimism node software has two main components: the "data transport layer" (DTL for short) and L2Geth. A while back, we added some code that stops L2Geth from syncing if the DTL hasn't synced all the way to the tip yet. Here's where that code sits:

https://github.com/ethereum-optimism/optimism/blob/17b11bee0ad164283ee7ccb92981249ec27216aa/l2geth/rollup/sync_service.go#L175-L188

This makes sync quite slow because the DTL has to sync to the tip first and then L2Geth has to start syncing. We want to try to have L2Geth sync in parallel with the DTL to speed things up.

Let's talk about how to make this happen. I *think* it should be possible to simply remove the code that I linked above. I'm not 100% sure about that, but it seems reasonable. Hopefully that'll be the entire PR. Once that's done, try running the integration tests to make sure everything is working as expected.

Once all that's ready to go, we'll spin up a mainnet node using the old version of the code and another node with the new version of the code. We'll make sure both nodes are syncing correctly and then we'll also record the time savings. Hopefully it's pretty significant!

tonykogias:
Have verifier node sync in parallel with dtl to save time.

**Metadata**
- Fixes #2329 
- Reopened #2335 here after rebase


smartcontracts (commented):
I have successfully synced an L1-syncing node with this change. We should try to sync an L2-syncing node as well.

smartcontracts:
So this definitely works for L1 sync. Maybe we can just enable this feature for L1 sync for now? L2 sync is generally much faster, so this is less necessary.

tynes:
> So this definitely works for L1 sync. Maybe we can just enable this feature for L1 sync for now? L2 sync is generally much faster, so this is less necessary.

I think this should also work when syncing from L2, we would just need to confirm that it is the case it does work by trying it out

smartcontracts:
I'd really like to get this merged, but I haven't had the opportunity to test on L2. In the interest of speed, let's just have the IF statement be:

```
if !cfg.Verifier || cfg.Backend == BackendL2 {
 ...
}
```

smartcontracts (approved):
Can confirm that this properly syncs an L1-syncing node

smartcontracts:
cc @tynes do you mind dropping one quick review here

PR #2658, opened 2022-06-02
specs: Fix inconsistencies within documentation

trianglesphere:
**Description**
Three small commits + one wontfix to fix the four inconsistencies that ToB found in our documentation.

**Metadata**
- Fixes ENG-2152

### DIFF
--- a/specs/deposits.md
+++ b/specs/deposits.md
@@ -256,8 +256,7 @@ The deposit contract handles two special cases:
 [address-aliasing]: #address-aliasing
 
 If the caller is a contract, the address will be transformed by adding
-`0x1111000000000000000000000000000000001111` to it. This is the reverse of the
-transformation described in the [withdrawals spec](./withdrawals.md#address-aliasing). This prevents attacks in which a
+`0x1111000000000000000000000000000000001111` to it. This prevents attacks in which a
 contract on L1 has the same address as a contract on L2 but doesn't have the same code. We can safely ignore this
 for EOAs because they're guaranteed to have the same "code" (i.e. no code at all). This also makes
 it possible for users to interact with contracts on L2 even when the Sequencer is down.
--- a/specs/deposits.md
+++ b/specs/deposits.md
@@ -84,7 +84,7 @@ The `sourceHash` of a deposit transaction is computed based on the origin:
   Where the `l1BlockHash`, and `l1LogIndex` all refer to the inclusion of the deposit log event on L1.
   `l1LogIndex` is the index of the deposit event log in the combined list of log events of the block.
 - L1 attributes deposited:
-  `keccak256(bytes32(uint256(1)), keccak256(l1BlockHash), bytes32(uint256(seqNumber))`.
+  `keccak256(bytes32(uint256(1)), keccak256(l1BlockHash), bytes32(uint256(seqNumber)))`.
   Where `l1BlockHash` refers to the L1 block hash of which the info attributes are deposited.
   And `seqNumber = l2BlockNum - l2EpochStartBlockNum`,
   where `l2BlockNum` is the L2 block number of the inclusion of the deposit tx in L2,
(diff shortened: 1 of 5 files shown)


specs/deposits.md
```diff
@@ -257,7 +257,7 @@ The deposit contract handles two special cases:
 
 If the caller is a contract, the address will be transformed by adding
 `0x1111000000000000000000000000000000001111` to it. This is the reverse of the
-transformation described in the [deposits spec](./withdrawals.md#address-aliasing). This prevents attacks in which a
+transformation described in the [withdrawals spec](./withdrawals.md#address-aliasing). This prevents attacks in which a
```
smartcontracts, line 260: Note that we do not do this transformation for withdrawals anymore. Another spec change to watch out for.

smartcontracts, line 260: Note that we do not do this transformation for withdrawals anymore. Another spec change to watch out for.

maurelian: Why do we not do it for withdrawals anymore?

specs/rollup-node.md
```diff
@@ -93,13 +93,13 @@ Note that the sequencing windows overlap.
 The rollup reads the following data from the [sequencing window][g-sequencing-window]:
 
 - Of the *first* block in the window only:
+  - L1 log entries emitted for [user deposits][g-deposits], augmented with a [sourceHash](./deposits.md#).
+- Of each block in the window:
```
maurelian, line 97: The report says: 
> Some portions of documentation state that the L1 attributes deposited transaction is included only in the first block of a sequencing window, while other portions of the documentation state that it is included in every L2 block. (The latter is the correct behavior.)

This change doesn't quite seem to reflect what they describe. It refers to reading data from L1, not posting to L2. Is what they describe as the actual behavior somehow implied by this change?

trianglesphere: gah, you're right. This fix was wrong. The sequencing window is not the same as the L2 epoch (though one sequencing window does correspond to a sequencing window.

specs/deposits.md
trianglesphere: Can we add a rationale section to the address aliasing (and I guess explain why we only need it one way)?

maurelian: Also, just noting that the issue says

> Links to the “deposits spec” throughout the documentation should read
“withdrawals spec” instead.

But I double checked and only see one instance of this error.

maurelian (changes requested):
The changes look good to me, though for the sake of the auditors I'd like if we can explain the reason for the one wontfix.

specs/rollup-node.md
trianglesphere: This fix request first links to the line "Of the first block in the window only". The window in question is the sequencing window, the set of L1 blocks from which an L2 epoch is derived. It is correct that the L1 Attributes are read from the first block of the L1 Sequencing window. The second link points to the fact that there must be a L1 Attributes Transaction for every L2 block. Given the first point is about L1 blocks, and the second is about L2 blocks, there is no inconsistency.

maurelian (approved):

specs/deposits.md
```diff
@@ -84,7 +84,7 @@ The `sourceHash` of a deposit transaction is computed based on the origin:
   Where the `l1BlockHash`, and `l1LogIndex` all refer to the inclusion of the deposit log event on L1.
   `l1LogIndex` is the index of the deposit event log in the combined list of log events of the block.
 - L1 attributes deposited:
-  `keccak256(bytes32(uint256(1)), keccak256(bytes32(uint256(l1BlockHash)), bytes32(uint256(seqNumber))`.
+  `keccak256(bytes32(uint256(1)), keccak256(l1BlockHash), bytes32(uint256(seqNumber))`.
```
protolambda, line 87:
```suggestion
  `keccak256(bytes32(uint256(1)), keccak256(l1BlockHash), bytes32(uint256(seqNumber)))`.
```

Missing `)`

protolambda (commented):
Looks good, minor typo-like issue to fix still, see review.

mslipper (approved):

PR #2661, opened 2022-06-02
core-utils: update geth genesis type

tynes:
Can specify the private key in the json file which
is nice for testing

**Description**
Updates core-utils to have updated geth types that make building a genesis block more easy


mslipper (approved):

mslipper:
LGTM, but shouldn't it be a major release if we're doing a backwards-incompatible API change?

PR #2657, opened 2022-06-02
bedrock: update geth, now includes rpc auth utils

protolambda:
This updates the bedrock go modules to use the latest `optimism-prototype` branch commit of our bedrock geth fork.
This geth version is rebased on the latest upstream master branch of geth, and includes cherry-picked commits of an open upstream PR that adds RPC auth utils/tests to unblock the engine-auth PR (#2591)

PR #2665, opened 2022-06-02
Bedrock: update geth, fix bug

protolambda:
Update `reference-optimistic-geth`. Same as #2657, but this time:
- Properly update the `go.work` and `go.work.sum` (tests were running with old geth previously, and thus missed a bug)
- Fix bug: geth prepares an "empty" block by default, and then kicks off the background process to build a better block. This empty block must still include the deposits / other force-txs, only when sequencing with the tx pool it makes sense to run the async block production. For the default "empty" block though, I was returning `block` (the parent) instead of `empty` (the prepared default block). This caused it to repeat the genesis over and over again, without any txs. See https://github.com/ethereum-optimism/reference-optimistic-geth/blob/dd2e24b3359f7948ec7eabe903d2af94269c8f78/eth/catalyst/api.go#L220

In `reference-optimistic-geth` we are also charging gas for deposits now, and the L1 info deposit didn't seem to have enough, it ran out of gas. I increased it to 150_000 gas for now, and then we can fix the unexpected high gas usage in a separate PR (the spec defined 75_000, but clearly that is not enough).


tynes:
Maybe we should have a script in the root of the repo that can easily ensure that the versions are matching or up to date?

tynes (approved):

mslipper (approved):

PR #2669, opened 2022-06-03
contracts-bedrock: move core-utils to deps

tynes:
It was previously in dev dependencies which could
cause problems for production builds

**Description**
Moves core-utils to deps from devdeps, should resolve build issues


protolambda (approved):

smartcontracts (approved):

smartcontracts:
Force merged to override issue in CI

PR #2670, opened 2022-06-03
package: contracts-governance

tynes:
**Description**
Adds the governance contracts to the monorepo. This will let me import `GovernanceToken` into the devnet genesis config as a predeploy


mslipper (approved):

.circleci/config.yml
```diff
@@ -178,6 +178,24 @@ jobs:
...
+  contracts-governance-tests:
+    docker:
+      - image: ethereumoptimism/js-builder:latest
+    resource_class: xlarge
+    steps:
+      - restore_cache:
+          keys:
+            - v2-cache-yarn-build-{{ .Revision }}
+      - checkout
+      - run:
```
smartcontracts, line 190: Should we also add slither?

packages/contracts-governance/.nvmrc
```diff
@@ -0,0 +1 @@
+v14.18.1
```
smartcontracts, line 1: Included at root level

smartcontracts (approved):
Code itself could use some cleanup but not really here to address that. Let's get it merged then do cleanup PRs as we go.

packages/contracts-governance/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "files": [
+    "dist/**/*.js",
+    "dist/**/*.d.ts",
+    "dist/types/*.ts",
+    "artifacts/**/*.json",
+    "deployments/**/*.json"
+  ],
+  "scripts": {
+    "build": "npx hardhat compile",
+    "test": "npx hardhat test",
```
maurelian, line 17: Any reason to use `npx` here but `yarn solhint` below?

tynes: this was copy pasted from the other repo, can clean up

.circleci/config.yml
tynes: slither isn't configured for this package and I don't feel like setting it up, not much dev will happen here and we could consider consolidating this package into others in the future

PR #2673, opened 2022-06-03
ci: build contracts-bedrock build

tynes:
**Description**
Should fix the ci build for `contracts-bedrock`


mslipper (approved):

smartcontracts (changes requested):
#2672 fixes instead

PR #2674, opened 2022-06-03
feat(ctp): use solmate instead of oz

smartcontracts:
**Description**
Replaces OZ contracts with solmate, now that v7 is MIT licensed.


tynes (approved):

Inphi (approved):
Looks fine I guess. But why replace OZ?

PR #2675, opened 2022-06-03
Frozen lockfile

roninjin10:
CI should fail if lockfile is wrong


smartcontracts (approved):

tynes:
The integration tests are now failing CI due to foundry not being able to find its compiler. It should auto download it, does `--frozen-lockfile` prevent changes to the filesystem? I don't think this really makes sense

tynes (approved):

PR #2575, opened 2022-05-17
feat(contracts): deposit gas metering via 1559

smartcontracts:
**Description**
Implements deposit gas metering via a 1559-like mechanism.

TODO:
- [x] Fix bug preventing tests from running
- [x] Add tests

**Metadata**

- Fixes ENG-2187

PR #2679, opened 2022-06-03
specs: Add deposit gas metering spec 

trianglesphere:
**Description**
Add specs for the new deposit fee model.

Implementation is done in the following PRs:
- Tracking guaranteed gas on L2: https://github.com/ethereum-optimism/reference-optimistic-geth/pull/21
- Versioning deposits: https://github.com/ethereum-optimism/reference-optimistic-geth/pull/19
- Solidity Fee Market: https://github.com/ethereum-optimism/optimism/pull/2575
- Adjust L1 Info Tx Gas: https://github.com/ethereum-optimism/optimism/pull/2582

**Metadata**
- Fixes ENG-2109


specs/deposits.md
```diff
@@ -138,6 +143,11 @@ follows:
 - `context.gas` set to `gasLimit`
 - `context.value` set to `sendValue`
 
+No gas is bought on L2 and no refund is provided. The gas used for the deposit is subtracted from
+the gas pool on L2. Gas usage exactly matches other transaction types (including intrinsic gas).
```
trianglesphere, line 147: Comment from @maurelian "is 'gas pool' a commonly used term? I think 'block gas limit' is more intuitive."

trianglesphere: (answer from Joshua): It's from geth. Maybe a little used outside of geth.

maurelian: OK, sgtm

specs/deposits.md
```diff
@@ -163,7 +173,7 @@ This transaction MUST have the following values:
    contract][predeploy]).
 3. `mint` is `0`
 4. `value` is `0`
-5. `gasLimit` is set to the maximum available.
+5. `gasLimit` is set to 75,000.
```
tynes, line 176: Is this enforced by consensus or is it a policy? This should be made clear and if i understand correctly it is a policy

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,138 @@
...
+    now_basefee_wide = prev_basefee + base_fee_per_gas_delta
+
+    now_basefee = clamp(now_basefee_wide, min=MINIMUM_BASEFEE, max=UINT_64_MAX_VALUE)
+    now_bought_gas =  requested_gas
+
+# If we skipped multiple blocks between the previous block and now update the basefee again.
+# This is not exactly the same as iterating the above function, but quite close for reasonable
+# gas target values. It is also constant time wrt the number of missed blocks which is important
+# for keeping gas usage stable.
+# TODO: Update this to a fixed point exponentiation for constant time results. This loop is currently
```
tynes, line 102: This todo can be addressed now

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Rationale for burning L1 Gas
+
+If we collect ETH directly we need to add the payable selector. Some projects are not able to do
+this. The alternative is to burn L1 gas. Unfortunately this is quite wastefull. As such, we provide
+two options to buy L2 gas:
+
+1. Burn L1 Gas 2. Send ETH to the Optimism Portal
+
+The payable version (Option 2) will likely have discout applied to it (or conversly, #1 has a
```
tynes, line 135: I believe we dropped support for option 2?

trianglesphere: For the initial release yes. I do like calling out that we intend to add it though.

specs/deposits.md
trianglesphere: It's a parameter in the rollup node. It causes a chain split if it's not set the same for every node (at least right now).

trianglesphere:
> @trianglesphere it'll just keep putting those reviewers back. I can remove that in Mergify.

I didn't realize it was mergify that kept adding them. If you use GH code owners it doesn't re-add them if you manually remove them (also code owners lints for people to make sure they are in the project).

specs/deposits.md
tynes: Ah that makes sense, so its a consensus param

specs/glossary.md
```diff
@@ -139,11 +140,20 @@ Different transaction types can contain different payloads, and be handled diffe
...
 > TODO: define and link to those concepts
 
+## Priority Gas Auction
+
+Transactions are in ethereum are ordered by the price that the transaction pays to the miner. Priority Gas Auctions
+(PGAs) occur when multiple parties are competing to be the first transaction in a block. Each party continuously
+updates the gas price of their transaction. PGAs occur when there is value in submitting a transaction before other
+parties (like being the first deposit or submitting a deposit before there is not more guaranteed gas remaining).
+PGAs tend to have negative externalities on the network due to a large amount of transactions being submitted in a
+very short amount of time.
```
tynes, line 155: I don't think it's a great explanation of the negative externalities but I don't think it's really our responsibility to explain them as part of this doc. Perhaps flashbots has a blog post someplace but no need to update this

tynes (approved):

specs/deposits.md
norswap: Was going to comment about this as well.
Could be ".. from the gas pool on L2 (i.e. it reduces the amount of gas that can be consumed by direct L2 transactions)."

specs/glossary.md
```diff
@@ -139,11 +140,20 @@ Different transaction types can contain different payloads, and be handled diffe
...
 
-L2 also has a fork choice rule, although the rules vary depending on wether we want the sequencer-confirmed head, the
+L2 also has a fork choice rule, although the rules vary depending on whether we want the sequencer-confirmed head, the
 on-chain-confirmed head, or the on-chain-finalized head.
 
 > TODO: define and link to those concepts
 
+## Priority Gas Auction
+
+Transactions are in ethereum are ordered by the price that the transaction pays to the miner. Priority Gas Auctions
```
norswap, line 150: s/are in Ethereum/in Ethereum

specs/glossary.md
norswap: Agreed, that an external link could serve as alternative. Though I remember there was sometimes pushback for including external links.

Assuming this is kept, then:
> are competing to be the first transaction in a block. 

Not necessarily, could be for inclusion in the block at all, for instance.

> PGAs tend to have negative externalities on the network due to a large amount of transactions being submitted in a
very short amount of time.

I don't think that's accurate. In fact it's rather the reverse: a lot of transaction being submitted (e.g. for an NFT mint) create a PGA to be included in the block.

The negative consequence is more about normal people being unable to transact for an extended period of time.
What EIP-1559 achieves is temporarily extended throughput by raising the cost (which PGA participants are willing to bear anyway) so that the period of congestion due to the PGA subsides faster.

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+(`guaranteed gas * L2 deposit basefee`). The contract then accepts that amount of ETH (in a future
+upgrade) or (only method right now), burns an amount of L1 gas that corresponds to the L2 cost
+(`L2 cost / L1 Basefee`).
+
+## Gas Stipend
+
+To offset the gas spent on the deposit event, we credit the amount of gas spent on the metering step
+times the current basefee to the cost of the L2 gas. The amount of gas is selected to represent the
+cost to the user. If the ETH price of the gas (gas times current L1 baseefee) is greater than the
+requested guaranteed gas times the L2 gas price, no L1 gas is burnt.
```
norswap, line 38: s/spent on the metering step times the current basefee/spent on the metering step, multiplied by the current basefee, to the cost of the L2 gas/

It's not quite clear here what the "metering step" is, and in fact, I'm not sure it's what we want to say at all (the whole processing of the deposit is what we want to offset, right?).

Suggestion:

> To offset the gas spent on the deposit event, we credit `gas spent * L1 basefee` ETH to the cost of the L2 gas, where `gas spent` is the amount of L1 gas spent processing the deposit. If the ETH value of this credit is greater than the ETH value of the requested guaranteed gas (`requested guaranteed gas * L2 gas price`), no L1 gas is burnt.

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Limiting Guaranteed Gas
+
+The total amount of guaranteed gas that can be bought in a single L1 block must be limited to
+prevent a denial of service attack against L2 as well as allow the total amount of guaranteed gas
+to be below the L2 block gas limit.
```
norswap, line 44: s/allow [...] to be below/ensure [...] stays below/

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+
+We set a guaranteed gas limit of 8,000,000 gas per L1 block and a target of 2,000,000 gas per L1
+block. These numbers enabled occasional large transactions while staying within our target and
+maximum gas usage on l2.
```
norswap, line 48: s/l2/L2

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+
+## 1559 Fee Market
+
+To reduce [Priority Gas Auctions](./glossary.md#priority-gas-auction) and accurately price gas, we
+implement a 1559 style fee market on L1 with the following pseudocode. We also use this opporunity
+to place a hard limit on the amount of guaranteed gas that is provided.
```
norswap, line 54: s/opporunity/opportunity/

Generaly like "EIP-1559-style ..." better than "1559 style" (pretty sure the dash is required, and EIP-1559 is a bit more specific).

I think we can drop the PGA thing, which also solves the glossary issue.

I think we could combine this section and the previous one, as the previous one gives most of the rationale and then add a paragraph that explains that because we limit the gas, we must implement a fee market.

e.g.:
> Because the amount of guaranteed L2 gas that can be purchased in a single block is now limited, we implement an EIP-1559-style fee market to reduce congestion on deposits. By setting the limit at a multiple of the target, we enable deposits to temporarily use more L2 gas at a greater cost.

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+# bought in a block. Calling code must handle the gas burn and validity checks on
+# the ability of the account to afford this gas.
+BASE_FEE_MAX_CHANGE_DENOMINATOR = 8
+ELASTICITY_MULTIPLIER = 4
+MAX_RESOURCE_LIMIT = 8,000,000
+TARGET_RESOURCE_LIMIT = MAX_RESOURCE_LIMIT / ELASTICITY_MULTIPLIER
+MINIMUM_BASEFEE=10000
+
+# prev_basefee is a u128, prev_bought_gas and prev_num are u64s
+prev_basefee, prev_bought_gas, prev_num = load_and_unpack_storage()
```
norswap, line 67: Isn't this changing to be their own slots?

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+    else:
+        return u128(v)
+
+# If this is a new block, update the basefee and reset the total gas
+# If not, just update the total gas
+if prev_num == now_num:
+    now_basefee = prev_basefee
+    now_bought_gas = prev_bought_gas + requested_gas
+elif prev_num != now_num :
+    # New formula
```
norswap, line 87: Do we need this "New formula" here?

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+    base_fee_per_gas_delta = prev_basefee * gas_used_delta / TARGET_RESOURCE_LIMIT / BASE_FEE_MAX_CHANGE_DENOMINATOR
+    now_basefee_wide = prev_basefee + base_fee_per_gas_delta
+
+    now_basefee = clamp(now_basefee_wide, min=MINIMUM_BASEFEE, max=UINT_64_MAX_VALUE)
+    now_bought_gas =  requested_gas
+
+# If we skipped multiple blocks between the previous block and now update the basefee again.
+# This is not exactly the same as iterating the above function, but quite close for reasonable
+# gas target values. It is also constant time wrt the number of missed blocks which is important
+# for keeping gas usage stable.
```
norswap, line 101: Why is it not exactly the same? Looks exactly the same to me.

If last update was at block 1 and we are at block 4, we update based on block 1's old value above (update "logically" at "slot 2" , and here we account for empty block 2 and 3 (logically at slot 3 and 4). For empty blocks `gas_used_delta == -TARGET_RESOURCE_LIMIT`, which yield a 1/8 reduction, so 7/8 multiplier.

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+### Exponent Based Fee Reduction
+
+When there are stretches where no deposits are executed on L1, the basefee should be decaying, but
+is not. If there is the case that the basefee spiked, this mechanism is needed to enable a more
+accurate decay. It uses exponentiation to run in constant (relative to the number of missed blocks)
+gas.
+
+The current loop based approach will always converge in no more than 12 steps, however it is possible
+to go to a loopless form (with slightly more error), but adpoting a fixed point exponentiation
+algorithm.
```
norswap, line 123: I would just get rid of this whole section and let the algorithm + comments there do the talking — I think it's clear enough.

Also seems outdated since the above uses the exponential decay, and this talks about "the current loop-based approach".

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+## Rationale for burning L1 Gas
+
+If we collect ETH directly we need to add the payable selector. Some projects are not able to do
+this. The alternative is to burn L1 gas. Unfortunately this is quite wastefull. As such, we provide
+two options to buy L2 gas:
+
+1. Burn L1 Gas
+2. Send ETH to the Optimism Portal (Not yet supported)
+
+The payable version (Option 2) will likely have discout applied to it (or conversly, #1 has a
```
norswap, line 134: s/discout/discount/

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+
+The current loop based approach will always converge in no more than 12 steps, however it is possible
+to go to a loopless form (with slightly more error), but adpoting a fixed point exponentiation
+algorithm.
+
+## Rationale for burning L1 Gas
+
+If we collect ETH directly we need to add the payable selector. Some projects are not able to do
+this. The alternative is to burn L1 gas. Unfortunately this is quite wastefull. As such, we provide
+two options to buy L2 gas:
```
norswap, line 129: Replace by:

> If we collect ETH directly to pay for L2 gas, every (indirect) caller of the deposit function will need to be marked with the `payable` selector. This won't be possible for many existing projects. Unfortunately this is quite wasteful. As such, we provide two options to buy L2 gas:

norswap (changes requested):
It's in pretty good shape!

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,137 @@
...
+it is sometimes paid for with a gas burn and there may not be any ETH left to refund.
+
+The **guaranteed gas** is composed of a gas stipend, and of any guaranteed gas the user would like
+to purchase (on L1) on top of that.
+
+Guaranteed gas on L2 is bought in the following manner. An L2 gas price is calculated via a 1559
+style gas market. The total amount of ETH required to buy that gas is then calculated
+(`guaranteed gas * L2 deposit basefee`). The contract then accepts that amount of ETH (in a future
+upgrade) or (only method right now), burns an amount of L1 gas that corresponds to the L2 cost
+(`L2 cost / L1 Basefee`).
```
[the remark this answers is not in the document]
trianglesphere: Fixed all three. Would appreciate a double check on the last point.

specs/glossary.md
trianglesphere: They talk about it a bit, but don't have a good reference that I could easily find.

specs/guaranteed-gas-market.md
trianglesphere: We place them into a struct which solidity packs into a single slot.

trianglesphere: https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/contracts/L1/ResourceMetering.sol

specs/guaranteed-gas-market.md
trianglesphere: nope

specs/guaranteed-gas-market.md
trianglesphere: it's logically the same, but we are working with essentially fixed point numbers & integer division so we have some error when using these modified formulas.

specs/guaranteed-gas-market.md
trianglesphere: Will do. The main loop got updated in the implementation when solmate got released with a new license.

specs/guaranteed-gas-market.md
trianglesphere: Combined.

specs/guaranteed-gas-market.md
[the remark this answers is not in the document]
norswap: extra pedantic: still missing a dash after EIP-1559 :p

specs/guaranteed-gas-market.md
norswap: Given that the solidity doesn't have `load_and_unpack_storage()`, just refers to the struct fields, could we just do something like:
`prev_basefee, prev_bought_gas, prev_num = <values from previous update>` ?

Or just have top-level `prev_basefee, prev_bought_gas, prev_num` vars with proper initial value, then change the algorithm to be contained with an `update` function that takes some parameters (`now_num`, `requested_gas` ... maybe I'm missing something else). At the end of the function assign `prev_basefee, prev_bought_gas, prev_num` to their proper values.

My objections here is that `load_and_unpack_storage()` is more like an implementation detail that doesn't have a *semantic* meaning in the algorithm. It doesn't really matter how we store the values to compute the amount of gas, this is a platonic algorithm which can then be impemented on chain or even somewhere else for simulation purpose.

specs/guaranteed-gas-market.md
norswap: I think I would include that in the comment then, something like what you just said "logically identical, but because we are using point numbers & integer division so we have some error when using these modified formulas."

norswap (approved):
Made some followup comments, fix to you liking, and after that gtg!

PR #2683, opened 2022-06-04
fix(ctp): re-deploy RetroReceiver with new owner

smartcontracts:
**Description**
Re-deploys the RetroReceiver with a new default owner called the
Dedicated Deterministic Deployer address. When deploying authenticated
smart contracts deterministically to the same address on many chains,
it's necessary to have a known static owner that then transfers
ownership to the final owner after deployment (which can be different on
every chain). This PR re-deploys the original contract with that new
default owner, the DDD.


tynes (changes requested):
The `deployments` directory is in the npm package, this should get a changeset

smartcontracts:
> The `deployments` directory is in the npm package, this should get a changeset

Done

tynes (approved):

tynes (approved):

PR #2684, opened 2022-06-05
devnet: Fix devnet

mslipper:


Inphi (approved):
looks good. Though removing images that match a glob can be surprising to users that build custom images (myself included) with a matching glob. It would be better if we removed specific image names. But this is acceptable too.

tynes (approved):

PR #2687, opened 2022-06-05
feat: introduce the drippie-mon service 

smartcontracts:
**Description**
Introduces drippie-mon, a basic service for monitoring Drippie
deployments. drippie-mon will increment a metric whenever a drip is
currently executable. Clients of drippie-mon can watch for this metric
to see if a drip has been executable for a while but has not yet been
executed.


packages/drippie-mon/src/service.ts
```diff
@@ -0,0 +1,117 @@
...
+  protected async init(): Promise<void> {
+    this.state.drippie = new ethers.Contract(
+      this.options.drippieAddress,
+      DrippieArtifact.abi,
+      this.options.rpc
+    )
+  }
+
+  protected async main(): Promise<void> {
+    const dripCreatedEvents = await this.state.drippie.queryFilter(
```
tynes, line 59: Does this default to the range [start, end] or just tip? Will fail on backends like alchemy if range is too large

smartcontracts: Actually should work fine on Alchemy, which is why I did it this way. Alchemy is [10k blocks OR 2k events] I believe. Since we don't expect to have many drips, should work fine to search the entire chain because Alchemy caches these results.

packages/contracts-periphery/contracts/universal/drippie/Drippie.sol
```diff
@@ -201,6 +213,29 @@ contract Drippie is AssetReceiver {
...
+     * assumption being made here is that setting the drip to ACTIVE is an affirmative signal that
+     * the drip should be executable according to the drip parameters, drip check, and drip
+     * interval. Note that drip parameters are read entirely from the state and are not supplied as
+     * user input, so there should not be any way for a non-authorized user to influence the
+     * behavior of the drip.
+     *
+     * @param _name Name of the drip to trigger.
+     */
+    function drip(string memory _name) external {
+        DripState storage state = drips[_name];
```
tynes, line 231: Do you need to `sload` a bunch here before the `require` statement?

smartcontracts: No only SLOADs when things actually get used

.github/workflows/release.yml
```diff
@@ -372,6 +373,33 @@ jobs:
           push: true
           tags: ethereumoptimism/fault-detector:${{ needs.release.outputs.fault-detector }},ethereumoptimism/fault-detector:latest
 
+  drippie-mon:
+    name: Publish Fault Detector Version ${{ needs.release.outputs.drippie-mon }}
```
smartcontracts, line 377:
```suggestion
    name: Publish Drippie Mon Version ${{ needs.release.outputs.drippie-mon }}
```

packages/drippie-mon/src/service.ts
```diff
@@ -0,0 +1,117 @@
...
+          validator: validators.provider,
+          desc: 'Provider for network where Drippie is deployed',
+        },
+        drippieAddress: {
+          validator: validators.str,
+          desc: 'Address of Drippie contract',
+        },
+      },
+      metricsSpec: {
+        executableDripCount: {
```
mslipper, line 41: Swap this to a gauge metric - set it to zero if the drip is not executable, set to 1 if it is.

packages/drippie-mon/src/service.ts
```diff
@@ -0,0 +1,117 @@
...
+          type: Counter,
+          desc: 'Number of times a drip has spotted being executable',
+          labels: ['name'],
+        },
```
mslipper, line 45: Add a metric that tracks the number of successful drips.

mslipper (changes requested):
See inline comments

mslipper (approved):

tynes (approved):

PR #2688, opened 2022-06-05
feat(cmn): BaseServiceV2 exposes internal options

smartcontracts:
**Description**
BaseServiceV2 will expose internal options (loop interval, metrics port,
etc) as environment variables or command line options.

### DIFF
--- a/.changeset/ten-owls-return.md
+++ b/.changeset/ten-owls-return.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/common-ts': patch
+---
+
+Expose service internal options as environment or cli options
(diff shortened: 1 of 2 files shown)


Inphi (commented):
can we remove the `loopIntervalMs` and the others from the constructor? They're no longer used now that `options` provides them.

smartcontracts:
> can we remove the `loopIntervalMs` and the others from the constructor? They're no longer used now that `options` provides them.

I want to do this later, for now I still want apps to be able to configure defaults for themselves via the constructor

Inphi:
> > can we remove the `loopIntervalMs` and the others from the constructor? They're no longer used now that `options` provides them.
> 
> I want to do this later, for now I still want apps to be able to configure defaults for themselves via the constructor

gotcha. I also incorrectly thought they were no longer used. This is fine.

Inphi (approved):

TreTuna (approved):

roninjin10 (approved):

PR #2690, opened 2022-06-06
teleportr: Improve metrics

mslipper:
- Update metrics collection to happen on an interval in addition to on each tick of the driver loop. This lets us continue collecting on-chain metrics if the driver loop halts for some reason.
- Adds an additional metric to track the last on-chian deposit ID. This will let add Teleportr halting alerts more easily.

### DIFF
--- a/teleportr/drivers/disburser/driver.go
+++ b/teleportr/drivers/disburser/driver.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"math/big"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ethereum-optimism/optimism/bss-core/metrics"
@@ -48,6 +49,7 @@ type Config struct {
 	DisburserAddr        common.Address
 	ChainID              *big.Int
 	PrivKey              *ecdsa.PrivateKey
+	ChainMetricsEnable   bool
 }
 
 type Driver struct {
@@ -58,10 +60,12 @@ type Driver struct {
 	walletAddr           common.Address
 	metrics              *Metrics
 
-	currentDepositIDs []uint64
+	currentDepositIDs   []uint64
+	chainMetricsEnabled bool
+	metricsMu           sync.Mutex
 }
 
-func NewDriver(cfg Config) (*Driver, error) {
+func NewDriver(cfg Config, parentCtx context.Context) (*Driver, error) {
 	if cfg.NumConfirmations == 0 {
 		panic("NumConfirmations cannot be zero")
 	}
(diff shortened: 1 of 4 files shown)


teleportr/drivers/disburser/driver.go
```diff
@@ -140,8 +150,8 @@ func (d *Driver) ClearPendingTx(
 func (d *Driver) GetBatchBlockRange(
 	ctx context.Context) (*big.Int, *big.Int, error) {
 
-	// Update balance metrics on each iteration.
-	d.updateBalanceMetrics(ctx)
+	// Update metrics on each iteration.
+	d.collectChainMetrics(ctx)
```
Inphi, line 154: Now that there's a bg gorotuine also doing this, would it be better to only send a signal to `collectChainMetricsBackground` for metrics to be immediately collected?

teleportr/drivers/disburser/driver.go
```diff
@@ -720,19 +730,61 @@ func (d *Driver) deletePendingTx(startID, endID uint64) error {
...
+	if err != nil {
+		log.Error("Error getting deposit contract total disbursements")
+		nextDisbursementID = big.NewInt(0)
+	}
+
+	d.metricsMu.Lock()
 	d.metrics.DisburserBalance.Set(float64(disburserBal.Uint64()))
 	d.metrics.DepositContractBalance.Set(float64(depositBal.Uint64()))
+	d.metrics.ContractNextDepositID.Set(float64(nextDisbursementID.Uint64()))
+	d.metrics.ContractNextDisbursementID.Set(float64(nextDepositID.Uint64()))
```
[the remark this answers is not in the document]
mslipper: Fixed

teleportr/drivers/disburser/driver.go
mslipper: I feel like this is a bit cleaner. The goroutine just wraps this method. Would like to avoid having to do signal lifecycle management.

teleportr/drivers/disburser/driver.go
```diff
@@ -720,19 +730,61 @@ func (d *Driver) deletePendingTx(startID, endID uint64) error {
...
+	nextDepositID, err := d.depositContract.TotalDeposits(&bind.CallOpts{
+		Context: subCtx,
+	})
+	if err != nil {
+		log.Error("Error getting deposit contract total deposits")
+		nextDepositID = big.NewInt(0)
+	}
+
+	nextDisbursementID, err := d.disburserContract.TotalDisbursements(
+		&bind.CallOpts{},
```
tuxcanfly, line 777: Does this need `Context` as well?

mslipper: Ah yes, it does. Fixed.

PR #2699, opened 2022-06-06
op-node: Remove multi L2 Engine Option

trianglesphere:
**Description**
This stops the rollup node from driving multiple L2 engines at once.
This makes node lifecycle management easier.
It also has associated command line and testing changes.

**Additional context**
I tested the devnet locally and I don't believe that I need to make any changes there.
The flag names are actually the same, so as long as no one was passing in multiple
addresses or jwtsecrets, nothing changes.

**Metadata**
- Fixes ENG-2243


mslipper (approved):
I think this is much simpler. Thanks.

op-node/node/config.go
```diff
@@ -9,8 +9,8 @@ import (
 )
 
 type Config struct {
-	L1  L1EndpointSetup
-	L2s L2EndpointsSetup
+	L1 L1EndpointSetup
+	L2 L2EndpointsSetup
```
protolambda, line 13: This can now be renamed to L2EndpointSetup.

op-node/node/node.go
```diff
@@ -129,48 +126,28 @@ func (n *OpNode) initL1(ctx context.Context, cfg *Config) error {
...
-	engLog := n.log.New("engine", tag)
-
-	client, err := l2.NewSource(cl, &cfg.Rollup.Genesis, engLog)
+func (n *OpNode) initL2(ctx context.Context, cfg *Config, snapshotLog log.Logger) error {
+	rpcClient, err := cfg.L2.Setup(ctx, n.log)
+	if err != nil {
+		return fmt.Errorf("failed to setup L2 execution-engine RPC client: %w", err)
+	}
+	n.l2Node = rpcClient
+	// TODO: New tag on the log?. Note: must be key-value pair.
```
protolambda, line 135: Passing the logger as-is is fine

op-node/node/config.go
trianglesphere: I found one other plural version that I removed as well.

PR #2705, opened 2022-06-07
Bedrock - temporary batch-builder reorg fix - don't omit empty batches

protolambda:
Previously the node API, used by the batcher, wouldn't submit batches if they are empty.

However, if the sequencer lags or has other hiccups, it may build one or more extra L2 blocks in a sequencing window with the margin it has, before adopting the next L1 origin. If that is not submitted on-chain, then the driver does not see that the L1 origin is delayed, and builds a different chain. The sequencer runs this same derivation process, and would even reorg itself.

As temporary fix we can just include those empty batches, and the `epoch` of those will be synced between sequencer and verifier, even if the block is empty. This should just avoid the reorgs caused by different `epoch` values.

While we're still working on the new batch-submitter/derivation code this temporary fix should fix the reorg issue (tm).


trianglesphere (approved):

trianglesphere (approved):
lgtm

tynes:
This needs a rebase and CI should work

tynes:
It seems like where ever hh downloads `solc` from isn't available:

```
#27 47.18 lerna ERR! yarn run build exited 1 in '@eth-optimism/contracts'
#27 47.18 lerna ERR! yarn run build stdout:
#27 47.18 $ yarn build:contracts && yarn autogen:artifacts && yarn build:typescript
#27 47.18 $ hardhat compile --show-stack-traces
#27 47.18 Downloading compiler 0.5.17
#27 47.18 
#27 47.18 info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
#27 47.18 info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
#27 47.18 lerna ERR! yarn run build stderr:
#27 47.18 Error HH501: Couldn't download compiler version 0.5.17. Please check your connection.
#27 47.18 HardhatError: HH501: Couldn't download compiler version 0.5.17. Please check your connection.
```

tynes (approved):

PR #2707, opened 2022-06-07
bedrock: use block number for key in output oracle

maurelian:
**Description**
Replaces L2 timestamps with block numbers as the key in `mapping(uint => OutputProposal)`. 

**Additional context**

This also enables us to remove the `computeL2BlockNumber()` function.

I will handle RBAC implementation in a subsequent PR. 

TODO:
- [x] Get go tests working
- [x] Update natspec in solidity to the new style

**Metadata**

- Fixes ENG-2207


packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -11,6 +11,19 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
+
+    /// @notice OutputProposal represents a commitment to the L2 state.
+    /// The timestamp is the L1 timestamp that the output root is posted.
+    /// This timestamp is used to verify that the finalization period
+    /// has passed since the output root was submitted.
+    struct OutputProposal {
+        bytes32 outputRoot;
+        uint256 timestamp;
+    }
+
```
maurelian, line 26: Just relocated this from below.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -110,18 +101,20 @@ contract L2OutputOracle is Ownable {
...
      * This function may only be called by the Sequencer.
      * @param _l2Output The L2 output of the checkpoint block.
-     * @param _l2timestamp The L2 block timestamp that resulted in _l2Output.
+     * @param _l2BlockNumber The L2 block number that resulted in _l2Output.
      * @param _l1Blockhash A block hash which must be included in the current chain.
-     * @param _l1Blocknumber The block number with the specified block hash.
+     * @param _l1BlockNumber The block number with the specified block hash.
      */
-    function appendL2Output(
+    function appendL2Output( // todo: rename to proposeL2Output?
```
maurelian, line 108: Now is the time to rename anything we don't like in this contract.

trianglesphere:
@maurelian contract tests are failing as well

op-e2e/system_test.go
```diff
@@ -171,21 +172,21 @@ func TestL2OutputSubmitter(t *testing.T) {
...
-		l2ooTimestamp, err := l2OutputOracle.LatestBlockTimestamp(&bind.CallOpts{})
+		l2ooBlockNumber, err := l2OutputOracle.LatestBlockNumber(&bind.CallOpts{})
 		require.Nil(t, err)
 
 		// Wait for the L2 output oracle to have been changed from the initial
 		// timestamp set in the contract constructor.
-		if l2ooTimestamp.Cmp(initialSroTimestamp) > 0 {
+		if l2ooBlockNumber.Cmp(initialOutputBlockNumber) > 0 {
 			// Retrieve the l2 output committed at this updated timestamp.
-			committedL2Output, err := l2OutputOracle.GetL2Output(&bind.CallOpts{}, l2ooTimestamp)
```
maurelian, line 181: @trianglesphere should this check that the return value is non-zero?

trianglesphere: Probably. There's also the option to revert inside the contract.
Fixing this proper likely means adding a second wrapper around the contract

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -134,71 +149,58 @@ contract L2OutputOracle is Ownable {
...
     }
 
     /**
      * @notice Deletes the most recent output.
+     *
      * @param _proposal Represents the output proposal to delete
      */
-    function deleteL2Output(OutputProposal memory _proposal) external onlyOwner {
-        OutputProposal memory outputToDelete = l2Outputs[latestBlockTimestamp];
+    function deleteL2Output(OutputProposal memory _proposal) external onlySequencer {
```
maurelian, line 168: We still have an issue here to deal with, which is partially [discussed here](https://github.com/ethereum-optimism/optimism/pull/2598#discussion_r892978171). 

But in summary I see these problems: 
1. It should not be possible to delete outputs back to the full finalization window, as it would allow the sequencer to post a faulty proposal which is immediately finalized. However the FINALIZATION_WINDOW value
2. Even if we could easily add `require(block.timestamp > proposal.timestamp + FINALIZATION_WINDOW)`, 
  2.1. it would still allow the sequencer to delete all outputs in a block, then create a malicious withdrawal in the next.
  2.2. every time a proposal is deleted and replaced, the finalization time gets pushed back to the current `block.timestamp` + the `FINALIZATION_WINDOW`. 
  
The consequences of 2.2 are similar to that of a sequencer which goes offline and is late in posting new outputs, except that it bakes in a method for intentionally delaying withdrawals.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -6,96 +6,104 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
-        // solhint-disable-next-line not-rely-on-time
-        l2Outputs[_startingBlockTimestamp] = OutputProposal(_genesisL2Output, block.timestamp);
+        l2Outputs[_startingBlockNumber] = OutputProposal(_genesisL2Output, block.timestamp);
         HISTORICAL_TOTAL_BLOCKS = _historicalTotalBlocks;
-        // solhint-disable-next-line not-rely-on-time
-        latestBlockTimestamp = _startingBlockTimestamp;
-        // solhint-disable-next-line not-rely-on-time
-        STARTING_BLOCK_TIMESTAMP = _startingBlockTimestamp;
+        latestBlockNumber = _startingBlockNumber;
+        // TODO: I think we can just remove this now. It's no longer needed in computeL2BlockNumber.
```
maurelian, line 105: I'm working on this in a new PR: https://github.com/ethereum-optimism/optimism/pull/2735

op-e2e/system_test.go
```diff
@@ -206,6 +207,7 @@ func TestL2OutputSubmitter(t *testing.T) {
 
 		select {
 		case <-timeoutCh:
+			// fails here
```
trianglesphere, line 210:
```suggestion
```

op-e2e/system_test.go
```diff
@@ -173,21 +174,21 @@ func TestL2OutputSubmitter(t *testing.T) {
...
 
 			// Compute the committed L2 output's L2 block number.
-			l2ooBlockNumber, err := l2OutputOracle.ComputeL2BlockNumber(
-				&bind.CallOpts{}, l2ooTimestamp,
-			)
-			require.Nil(t, err)
+			// l2ooBlockNumber, err := l2OutputOracle.ComputeL2BlockNumber(
+			// 	&bind.CallOpts{}, l2ooBlockNumber,
+			// )
+			// require.Nil(t, err)
```
trianglesphere, line 191:
```suggestion
```

op-e2e/system_test.go
maurelian: >  a second wrapper around the contract

I assume you mean a second wrapper in Go?
I don't love adding a revert here, as it will make it more complicated to handle in any contract that calls this function, and generally we'd prefer to move complexity off-chain.

op-e2e/setup.go
```diff
@@ -380,7 +381,7 @@ func (cfg SystemConfig) start() (*System, error) {
 	sys.cfg.RollupConfig.Genesis = sys.RolupGenesis
 	sys.cfg.RollupConfig.BatchSenderAddress = batchSubmitterAddr
 	sys.cfg.RollupConfig.P2PSequencerAddress = p2pSignerAddr
-	sys.cfg.L2OOCfg.L2StartTime = new(big.Int).SetUint64(l2GenesisTime)
+	sys.cfg.L2OOCfg.L2StartingBlock = new(big.Int)
```
maurelian, line 384: @trianglesphere I think the following would be more accurate, and I'd like to add a check to the constructor to enforce `_historicalTotalBlocks > _startingBlockNumber`. 

```suggestion
	sys.cfg.L2OOCfg.L2StartingBlock = new(big.Int).SetInt64(int64(l2GenesisID.Number))
```

op-e2e/system_test.go
trianglesphere: yea, a wrapper around the go bindings (which is a wrapper around the eth & solidity API).
Let's manually do this check for now & build golang wrapper later.

op-e2e/setup.go
trianglesphere: Ok, let me update this to make a little more sense with the golang types, but we can go ahead with this.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -98,6 +98,10 @@ contract L2OutputOracle is Ownable {
         uint256 _startingBlockNumber,
         address sequencer
     ) {
+        require(
+            _startingBlockNumber > _historicalTotalBlocks,
```
trianglesphere, line 102: `0 > 0 => false`. This does bring up some interesting questions around the L2 genesis - is it a block that actually exists or not and what relation is there between created blocks & that block

maurelian: Right, my sense is that our testing is currently assuming that both of these numbers are zero.

maurelian: I've removed this commit for now to get this PR through.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -106,23 +114,26 @@ contract L2OutputOracle is Ownable {
...
+     * @param _l1BlockNumber The block number with the specified block hash.
      */
     function appendL2Output(
         bytes32 _l2Output,
-        uint256 _l2timestamp,
+        uint256 _l2BlockNumber,
         bytes32 _l1Blockhash,
-        uint256 _l1Blocknumber
+        uint256 _l1BlockNumber
     ) external payable onlyOwner {
```
protolambda, line 131: timestamp and block number were swapped in the argument order. The specs still define this contract interface in the old order. Can you update the specs definition? (`specs/proposals.md`)

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -6,96 +6,104 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
-    uint256 public latestBlockTimestamp;
+    /**
+     * @notice The number of the most recent L2 block recorded in this contract.
+     */
+    uint256 public latestBlockNumber;
 
-    /// @notice A mapping from L2 timestamps to the output root for the block with that timestamp.
+    /**
+     * @notice A mapping from L2 block numbers to the respective output root.
+     */
```
protolambda, line 77: Worth to add a warning that these proposals cannot be trusted until they are finalized (i.e. l1 timestamp of proposal must be old enough)

protolambda (changes requested):
Generally looks good, but the spec should be updated (the solidity interface for the proposer is part of the spec).

And when fault-proofs with cannon are ready, we'll need to change the interface/storage to link proposals together like a chain. We only want to fault-proof one proposal, and then remove all proposals that build on top of that with a simple proof. cc @norswap 
That change can be made in a later PR however, and I don't believe we have a spec for it yet, but it will change the current contract.

op-e2e/system_test.go
tynes: its kind of like the case in erc721 when `ownerOf(uint256)` reverts when the nft hasn't been minted

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -6,96 +6,104 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
  * @title L2OutputOracle
  * @notice The L2 state is committed to in this contract
- * The payable keyword is used on appendL2Output to save gas on the msg.value check.
- * This contract should be deployed behind an upgradable proxy
+ *         The payable keyword is used on appendL2Output to save gas on the msg.value check.
+ *         This contract should be deployed behind an upgradable proxy
  */
 // slither-disable-next-line locked-ether
 contract L2OutputOracle is Ownable {
+    /*********
```
tynes, line 14: I don't believe the contracts have the section headers elsewhere in the codebase

specs/proposals.md
```diff
@@ -91,28 +91,28 @@ where:
...
- * block. The timestamp must be equal to the current value returned by
- * `nextTimestamp()` in order to be accepted.
+ * @notice Accepts an L2 outputRoot and the timestamp of the corresponding L2 block. The
+ * timestamp must be equal to the current value returned by `nextTimestamp()` in order to be
+ * accepted.
  * This function may only be called by the Sequencer.
  * @param _l2Output The L2 output of the checkpoint block.
  * @param _l2timestamp The L2 block timestamp that resulted in _l2Output.
  * @param _l1Blockhash A block hash which must be included in the current chain.
  * @param _l1Blocknumber The block number with the specified block hash.
```
protolambda, line 108: Can you update the order of the doc comment params to match the function?

specs/proposals.md
```diff
@@ -91,28 +91,28 @@ where:
...
-    uint256 _l2timestamp,
-    bytes32 _l1Blockhash,
-    uint256 _l1Blocknumber
-)
+  function appendL2Output(
+      bytes32 _l2Output,
+      uint256 _l2BlockNumber,
+      bytes32 _l1Blockhash,
+      uint256 _l1BlockNumber
+  )
```
protolambda, line 115: duplicate argument name, 2nd should be `_l2timestamp` I think (and might need to fix capitalization for consistency with other arg names).

protolambda (changes requested):
some nice spec fixes, but there's a ~~duplicate arg + out of order~~ outdated doc comment.

specs/proposals.md
protolambda: Oh wait, I can't read, it's `l1BlockNumber`

specs/proposals.md
```diff
@@ -91,28 +91,28 @@ where:
...
 /**
- * Accepts an L2 output checkpoint and the timestamp of the corresponding L2
- * block. The timestamp must be equal to the current value returned by
- * `nextTimestamp()` in order to be accepted.
+ * @notice Accepts an L2 outputRoot and the timestamp of the corresponding L2 block. The
+ * timestamp must be equal to the current value returned by `nextTimestamp()` in order to be
+ * accepted.
  * This function may only be called by the Sequencer.
  * @param _l2Output The L2 output of the checkpoint block.
  * @param _l2timestamp The L2 block timestamp that resulted in _l2Output.
```
protolambda, line 106: doc comment needs to refer to `_l2BlockNumber` now

protolambda, line 106:
```suggestion
 * @param _l2BlockNumber The L2 block number that resulted in _l2Output.
```

protolambda (approved):
Looks good to me after fixing doc comment

specs/proposals.md
maurelian: Amended previous commit and fixed.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -134,71 +127,58 @@ contract L2OutputOracle is Ownable {
...
-            "Can only delete the most recent output."
+            "OutputOracle: The output root to delete does not match the latest output proposal."
+        );
+        require(
+            _proposal.timestamp == outputToDelete.timestamp,
+            "OutputOracle: The timestamp to delete does not match the latest output proposal."
         );
-        require(_proposal.timestamp == outputToDelete.timestamp, "");
 
         emit l2OutputDeleted(
```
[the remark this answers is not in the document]
maurelian: Fixed

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -20,19 +20,12 @@ contract OptimismPortal_Test is CommonTest {
...
-            1800,
-            2,
-            keccak256(abi.encode(0)),
-            100,
-            1,
-            address(666)
-        );
+    function setUp() public override {
+        L2OutputOracle_Initializer.setUp();
+        // oracle = new L2OutputOracle(1800, keccak256(abi.encode(0)), 100, 1, address(666));
```
tynes, line 28: dead code?

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
maurelian: I think it's fine as si.

op-e2e/system_test.go
maurelian: OK, I reconsidered and actually think that this is a good idea and will add in a future PR.
I've created a separate issue to track it.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -6,123 +6,116 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
+     */
+    struct OutputProposal {
+        bytes32 outputRoot;
+        uint256 timestamp;
+    }
 
-    /// @notice Emitted when an output is appended.
-    event l2OutputAppended(
+    /**
+     * @notice Emitted when an output is appended.
```
smartcontracts, line 26: Event parameters should be documented here, also we're moving away from underscores for names of event params but I can fix that in a different PR.

maurelian: I'll just fix it here.

maurelian: Hmmm, scratch that. I tried it and got a non-obvious go test failure.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -6,123 +6,116 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
...
      */
     function appendL2Output(
         bytes32 _l2Output,
-        uint256 _l2timestamp,
+        uint256 _l2BlockNumber,
         bytes32 _l1Blockhash,
-        uint256 _l1Blocknumber
+        uint256 _l1BlockNumber
     ) external payable onlyOwner {
-        require(_l2timestamp < block.timestamp, "Cannot append L2 output in future");
```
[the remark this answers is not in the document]
maurelian: Added.

smartcontracts (approved):

PR #2711, opened 2022-06-08
proxyd: Fix concurrent WS write panic

mslipper:
Fixes a panic in the websocket proxyd logic. Normally, the `clientPump` and `backendPump` methods in `WSProxier` send data in one direction. However, when the client sends an invalid RPC, the `clientPump` will send a response _directly to the client_ in order to avoid unnecessary roundtrips to the backend. This could be interleaved with concurrent writes to the client's WS in `backendPump`, and would cause a panic in the WS library.

To test this, this PR includes a dedicated integration test that reliably triggers the issue. In addition, this PR adds additional testing for WS functionality.

### DIFF
--- a/proxyd/integration_tests/ws_test.go
+++ b/proxyd/integration_tests/ws_test.go
@@ -1,0 +1,281 @@
+package integration_tests
+
+import (
+	"os"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/ethereum-optimism/optimism/proxyd"
+	"github.com/gorilla/websocket"
+	"github.com/stretchr/testify/require"
+)
+
+// TestConcurrentWSPanic tests for a panic in the websocket proxy
+// that occurred when messages were sent from the upstream to the
+// client right after the client sent an invalid request.
+func TestConcurrentWSPanic(t *testing.T) {
+	var backendToProxyConn *websocket.Conn
+	var setOnce sync.Once
+
+	readyCh := make(chan struct{}, 1)
+	quitC := make(chan struct{})
+
+	// Pull out the backend -> proxyd conn so that we can spam it directly.
+	// Use a sync.Once to make sure we only do that once, for the first
+	// connection.
+	backend := NewMockWSBackend(func(conn *websocket.Conn) {
+		setOnce.Do(func() {
+			backendToProxyConn = conn
+			readyCh <- struct{}{}
+		})
+	}, nil, nil)
+	defer backend.Close()
+
+	require.NoError(t, os.Setenv("GOOD_BACKEND_RPC_URL", backend.URL()))
+
+	config := ReadConfig("ws")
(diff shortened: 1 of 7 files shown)


proxyd/integration_tests/ws_test.go
```diff
@@ -0,0 +1,253 @@
...
+			select {
+			case <-quitC:
+				return
+			default:
+				_ = client.WriteMessage(websocket.TextMessage, []byte("{\"id\": 1, \"method\": \"eth_foo\", \"params\": [\"newHeads\"]}"))
+			}
+		}
+	}()
+
+	// 1 second is enough to trigger the panic without locks in place
```
Inphi, line 70: nit: It's not super clear to others what locks we're referring to.
reword this to - 1 second is enough to trigger a panic due to "concurrent write to websocket connection"

proxyd/integration_tests/ws_test.go
```diff
@@ -0,0 +1,253 @@
...
+			"{}",
+			"{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32700,\"message\":\"parse error\"},\"id\":null}",
+			"{\"jsonrpc\": \"2.0\", \"method\": true}",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			timeout := time.NewTicker(30 * time.Second)
+			doneCh := make(chan struct{}, 1)
+			backendHdlr.msgCB = func(conn *websocket.Conn, msgType int, data []byte) {
```
Inphi, line 153: The msgCB is read in a separate goroutine in `MockWSBackend.ServeHTTP`. This races with the write here and would make the test flaky.

Ensure that access to all `backendHandler` callbacks are exclusive.

Inphi (commented):
the fix looks good. i left a couple comments on unit testing.

proxyd/integration_tests/ws_test.go
mslipper: Strange that the race detector didn't pick this up. Will fix.

Inphi: it's probably because the `client.WriteMessage` call on L160 and the [ReadMessage](https://github.com/ethereum-optimism/optimism/pull/2711/files/62b4e9f82af4ad5f93b679fa6341a8ebd52db25c#diff-a8386cda07790cbe1234c20956a09b1c026d20cc1b77c7f771d0b647851186a9R295) in `MockWSBackend` act as memory barriers (internally in gorilla) so the cb access is inadvertently consistent. Though I wouldn't rely on this behavior especially if we upgrade the websocket module.

mslipper: Ah, that makes sense. These are hidden behind an `atomic.Value` now, so should be thread safe.

Inphi (approved):

PR #2718, opened 2022-06-08
feat(cmn): add kill function to BaseServiceV2

smartcontracts:
**Description**
Adds a kill function to BaseServiceV2 so you can shut things down
quickly if necessary.

### DIFF
--- a/.changeset/five-pens-press.md
+++ b/.changeset/five-pens-press.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/common-ts': patch
+---
+
+Add kill function to BaseServiceV2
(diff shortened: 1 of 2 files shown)


tynes (approved):

roninjin10 (commented):
This pr might not be necessary I'm putting up a pr that does what I need for test.  This won't work for my test because process.exit(0) would exit from the entire test

mslipper:
Couldn't you `return` from the loop rather than killing the process?

roninjin10:
@mslipper the problem was the service would have to finish "waiting" when looping to be able to end.   In https://github.com/ethereum-optimism/optimism/pull/2731 we make it so we can cancel the waiting without needing to wait for the "waiting" to finish before being able to return.

PR #2725, opened 2022-06-08
feat(cmn): have BaseService expose a full server

smartcontracts:
**Description**
Updates BaseServiceV2 to expose a full server instead of simply exposing
the server for metrics. Services can now add new custom routes to this
server. Mainly useful because we're already running the metrics server,
so we might as well allow people to add more things to it.

Requested by @roninjin10


mslipper (approved):

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -307,30 +318,67 @@ export abstract class BaseServiceV2<
...
-        this.metricsServer = app.listen(
-          this.metricsServerPort,
-          this.metricsServerHostname,
-          () => {
-            resolve(null)
-          }
-        )
+      // Registery user routes.
+      if (this.routes) {
+        this.routes(app)
```
roninjin10, line 369: The better way to do this is to accept a 

routes: express.Router 

Instead of a function that takes app.   Then we with that router we call

app.use('api' /** For some services in future we might end up having different versions of api */, router)

This way we aren't exposing the entire app to anybody using this.  They should only be able to define routes.

Alternatively, this.routes can be a function that passes in an express.Router() instead of an entire app.

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -307,30 +318,67 @@ export abstract class BaseServiceV2<
...
+      }
+
+      // Wait for server to come up.
+      await new Promise((resolve) => {
```
roninjin10, line 373: To be able to use supertest to write unit tests like we do herer https://github.com/ethereum-optimism/gateway/blob/develop/packages/backend/src/index.spec.ts#L20

We need app.listen to not get called by default.   My quick hack was to just wrap it in a if(nodenv !== 'test') but to do this cleanly we need to seperate this function into two methods

1. something called something like `private readonly initServer` returns the server
2. Something called something like `private readonly startServer`
3. Then expose a utility seperate from this server that wraps supertest 

import request from 'supertest'
const testBaseService = (service: BaseServiceV2): request.SuperTest<request.Test> => {
    // as any because we are using private method to do this test util
    // alternatively if you prefer you can make it public and throw an error if NODE_ENV is not test.
    request((service as any).initServer())
}

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -307,30 +318,67 @@ export abstract class BaseServiceV2<
...
+          includeStatusCode: true,
+        })
+      )
 
-        app.get('/metrics', async (_, res) => {
-          res.status(200).send(await this.metricsRegistry.metrics())
+      // Health status.
+      app.get('/healthz', async (req, res) => {
+        return res.json({
+          ok: this.healthy,
```
roninjin10, line 363: We should add a script to package.json called `yarn generate-sha` or something.   That script should run git rev-parse HEAD or whatever that command is to get the current git sha and generate a file called something like gitsha.txt.   We should then fs.readFile() to get the git sha and return it here so we not only know the service is up but can easily verify what version is running.

roninjin10 (changes requested):

packages/common-ts/src/base-service/base-service-v2.ts
roninjin10: In general it's bad practice to call app.listen by default.   You usually want to initialize the app and start the app in two different steps.

roninjin10 (approved):

packages/common-ts/src/base-service/base-service-v2.ts
smartcontracts: Adding in a separate PR

PR #2726, opened 2022-06-08
fix(fd): off by one in metric

smartcontracts:
**Description**
Fixes a simple off-by-one error in the highestCheckedBatchIndex metric.


Inphi (approved):
LGTM. Though the commit title makes it out like it's the dtl being updated.

mslipper (approved):

tynes (approved):

PR #2728, opened 2022-06-08
contracts bedrock: implement simple proxy + admin

tynes:
**Description**

Implement a simple proxy contract based on the chugsplash
proxy code. It is a transparent proxy and can be used
on both L1 and L2.


packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,155 @@
...
+ * @title Proxy
+ * @dev Basic transparent proxy contract
+ */
+contract Proxy {
+    /*************
+     * Constants *
+     *************/
+
+    // bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1)
+    bytes32 internal constant IMPLEMENTATION_KEY =
```
smartcontracts, line 14: Let's use the style we agreed on for the other contracts (seaport style)

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,155 @@
...
+     */
+    function setImplementation(address _implementation) external onlyOwner {
+        _setImplementation(_implementation);
+    }
+
+    /**
+     * Changes the owner of the proxy contract. Only callable by the owner.
+     * @param _owner New owner of the proxy contract.
+     */
+    // slither-disable-next-line external-function
```
maurelian, line 59: This disable comment is unnecessary, it enforces [this](https://github.com/crytic/slither/wiki/Detector-Documentation#description-108):
> public functions that are never called by the contract should be declared external to save gas.

maurelian: similar for the other occurences of it.

tynes:
One thing that this contract doesn't have is "upgrade and call" functionality

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,155 @@
...
+     * eth_call and setting the "from" address to address(0).
+     * @return Implementation address.
+     */
+    // slither-disable-next-line external-function
+    function getImplementation() external view returns (address) {
+        return _getImplementation();
+    }
+
+    /**********************
+     * Internal Functions *
```
maurelian, line 85: `_setImplementation()` and `_doProxyCall()` are both called only once, I think it'd be much nicer to just inline them into the external functions.

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    /**********************
+     * External Functions *
+     **********************/
+
+    /**
+     * @param _implementation The address of the implementation contract
+     */
+    function setImplementation(address _implementation) external onlyOwner {
```
maurelian, line 51: I think everything other than the fallback should have `proxyCallIfNotOwner()` on it 
https://github.com/ethereum-optimism/optimism/blob/255eea5751516f20e8bbdc981a145dda6f71acc6/packages/contracts/contracts/chugsplash/L1ChugSplashProxy.sol#L92-L99

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Queries the owner of the proxy contract.
+     * @return owner address.
+     */
+    function _getOwner() internal view returns (address owner) {
+        assembly {
+            owner := sload(OWNER_KEY)
+        }
+    }
```
maurelian, line 126: I think we've generally agreed to avoid named return arguments. In this case though it's much nice than this, so I'd be OK to let this in if @smartcontracts is.

```solidity
    function _getOwner() internal view returns (address) {
		address owner;
        assembly {
            owner := sload(OWNER_KEY)
        }
        return owner;
    }
```

maurelian:
> One thing that this contract doesn't have is "upgrade and call" functionality so that a single tx can upgrade + call initialize

Good callout. That also raises the fact we probably want to use a single [`ProxyAdmin`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/a5445b0afb8b350417b6e6ab3160554967bc151f/contracts/proxy/transparent/ProxyAdmin.sol#L13) contract to control all of these proxies, for the [reasons explained here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/a5445b0afb8b350417b6e6ab3160554967bc151f/contracts/proxy/transparent/TransparentUpgradeableProxy.sol#L22-L27). Fortunately that `ProxyAdmin` by OZ is pretty simple.

Specific to our situation, we'll probably want a single multisig to be the `owner` in the implementations behind all these contracts, therefore we'll need to use the `ProxyAdmin` if we want to use the multisig to control the proxies.

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,155 @@
...
+
+            // Success == 0 means a revert. We'll revert too and pass the data up.
+            if iszero(success) {
+                revert(0x0, returndatasize())
+            }
+
+            // Otherwise we'll just return and pass the data up.
+            return(0x0, returndatasize())
+        }
+    }
```
maurelian, line 154: This function is basically [identical](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/a5445b0afb8b350417b6e6ab3160554967bc151f/contracts/proxy/Proxy.sol#L23-L44) to OZ' base proxy, so LGTM

packages/contracts-bedrock/contracts/universal/Proxy.sol
smartcontracts: Although I agree it's nicer here, let's use named returns for consistency. I'd like to be able to install a "no named returns" linter rule in the future.

maurelian: You mean:
> let's _not_ use named returns for consistency
👍

maurelian:
Since the goal is a simple version of the TransparentUpgradeableProxy; I'd like to see [this](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/proxy/transparent/.test.js) test suite repurposed and applied here.

packages/contracts-bedrock/contracts/universal/Proxy.sol
maurelian: I changed my mind, at least about `_doProxy`, it's nice to have it at the bottom of the file so it's obviously identical to chugsplash. Alternatively we could dedicate the code into a common parent contract.

tynes:
Do we want to prevent the owner from accessing the fallback function like oz does?
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/afb20119b33072da041c97ea717d3ce4417b5e01/contracts/proxy/transparent/TransparentUpgradeableProxy.sol#L122

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * @title Proxy
+ * @notice Proxy is a transparent proxy that passes through the call
+ *         if the caller is the owner or if the caller is `address(0)`,
+ *         meaning that the call originated from an offchain simulation.
+ */
+contract Proxy {
+    /**
+     * @notice An event that is emitted each time the implementation is changed.
+     *         This event is part of the EIP 1967 spec.
```
smartcontracts, line 13: Parameters should also be documented here

smartcontracts (commented):
Generally looks good, a few minor comments to address

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,203 @@
...
+        assembly {
+            owner := sload(OWNER_KEY)
+        }
+        return owner;
+    }
+
+    /**
+     * @dev Performs the proxy call via a delegatecall.
+     */
+    function _doProxyCall() internal {
```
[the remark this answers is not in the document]
tynes: Yes this is intentional, the call will revert anyways

maurelian:
> Do we want to prevent the owner from accessing the fallback function like oz does? [OpenZeppelin/openzeppelin-contracts@`afb2011`/contracts/proxy/transparent/TransparentUpgradeableProxy.sol#L122](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/afb20119b33072da041c97ea717d3ce4417b5e01/contracts/proxy/transparent/TransparentUpgradeableProxy.sol#L122)

@tynes Seems like an SLOAD it would be nice to avoid.

If we use the [ProxyAdmin](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.6.0/contracts/proxy/transparent/ProxyAdmin.sol) as owner, then all the function selectors are hardcoded and there's no risk that it will trigger the fallback.

packages/contracts-bedrock/contracts/universal/Proxy.sol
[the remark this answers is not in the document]
smartcontracts: Hmmm are you sure it will revert? We should have a test for this if we don't already have that.

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,203 @@
...
+     */
+    bytes32 internal constant IMPLEMENTATION_KEY =
+        0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
+
+    /**
+     * @notice The storage slot that holds the address of the owner.
+     *         bytes32(uint256(keccak256('eip1967.proxy.admin')) - 1)
+     */
+    bytes32 internal constant OWNER_KEY =
+        0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103;
```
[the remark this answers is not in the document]
tynes: There is test coverage of this btw

packages/contracts-bedrock/contracts/universal/Proxy.sol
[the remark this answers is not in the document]
tynes: Ok so it does not revert if the implementation is set to `address(0)`. Is the point of this check meant to prevent it to not call a "burn address"? Its possible to call accounts without code and non zero burn addresses

tynes:
> > One thing that this contract doesn't have is "upgrade and call" functionality so that a single tx can upgrade + call initialize
> 
> Good callout. That also raises the fact we probably want to use a single [`ProxyAdmin`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/a5445b0afb8b350417b6e6ab3160554967bc151f/contracts/proxy/transparent/ProxyAdmin.sol#L13) contract to control all of these proxies, for the [reasons explained here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/a5445b0afb8b350417b6e6ab3160554967bc151f/contracts/proxy/transparent/TransparentUpgradeableProxy.sol#L22-L27). Fortunately that `ProxyAdmin` by OZ is pretty simple.
> 
> Specific to our situation, we'll probably want a single multisig to be the `owner` in the implementations behind all these contracts, therefore we'll need to use the `ProxyAdmin` if we want to use the multisig to control the proxies.

I'm going to update the code to use the same interface as the oz proxy over the chugsplash proxy

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -0,0 +1,97 @@
...
+
+import { Proxy } from "./Proxy.sol";
+import { Owned } from "@rari-capital/solmate/src/auth/Owned.sol";
+
+/**
+ * @title ProxyAdmin
+ * @dev This is an auxiliary contract meant to be assigned as the admin of a
+        Proxy, based on the OpenZeppelin implementation.
+ */
+contract ProxyAdmin is Owned {
```
smartcontracts, line 12: Can we have an "isUpgrading" function here so we can finally resolve the ChugSplashProxy issues

tynes:
I've addressed all review comments on this PR

.circleci/config.yml
```diff
@@ -406,7 +406,9 @@ jobs:
           working_directory: packages/contracts-bedrock
       - run:
           name: gas snapshot
-          command: forge snapshot && git diff --exit-code
+          command: |
+            forge --version
```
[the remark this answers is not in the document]
tynes: This is for debugging purposes, in case the version of `forge` causes the snapshot to be different

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
maurelian: Does this contract work with chugsplash though? I thought the interface of the proxy changed.

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -0,0 +1,200 @@
...
+     * @param proxy The Proxy to return the implementation of.
+     * @return The address of the implementation
+     */
+    function getProxyImplementation(Proxy proxy) external view returns (address) {
+        ProxyType proxyType = getProxyType(address(proxy));
+
+        // We need to manually run the static call since the getter cannot be flagged as view
+        if (proxyType == ProxyType.OpenZeppelin) {
+            // bytes4(keccak256("implementation()")) == 0x5c60da1b
+            (bool success, bytes memory returndata) = address(proxy).staticcall(hex"5c60da1b");
```
smartcontracts, line 89: Might be better to just wrap the proxy in the interface to avoid needing to manually encode stuff

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -0,0 +1,200 @@
...
+     * @param implementation The address of the implementation
+     */
+    function upgrade(Proxy proxy, address implementation) public onlyOwner {
+        ProxyType proxyType = getProxyType(address(proxy));
+
+        if (proxyType == ProxyType.OpenZeppelin) {
+            proxy.upgradeTo(implementation);
+        } else if (proxyType == ProxyType.Chugsplash) {
+            bytes memory code = address(proxy).code;
+            L1ChugSplashProxy(payable(proxy)).setStorage(
```
smartcontracts, line 162: Lmfao

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * @dev Upgrades `proxy` to `implementation`.
+     *      This contract must be the admin of `proxy`.
+     *
+     * @param proxy          The address of the proxy
+     * @param implementation The address of the implementation
+     */
+    function upgrade(Proxy proxy, address implementation) public onlyOwner {
```
smartcontracts, line 155: I believe the correct interface is `upgradeTo`, and `upgradeToAndCall` below

smartcontracts: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/429b2242d8e7dedec5e69e7e2065d33e0fc76dc4/contracts/proxy/transparent/TransparentUpgradeableProxy.sol#L106

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -0,0 +1,200 @@
...
+        ProxyType proxyType = getProxyType(address(proxy));
+
+        if (proxyType == ProxyType.OpenZeppelin) {
+            proxy.upgradeTo(implementation);
+        } else if (proxyType == ProxyType.Chugsplash) {
+            bytes memory code = address(proxy).code;
+            L1ChugSplashProxy(payable(proxy)).setStorage(
+                0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc,
+                Bytes32AddressLib.fillLast12Bytes(implementation)
```
smartcontracts, line 164: Presumably can avoid the extra import by just doing `bytes32(bytes20(implementation))`

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -0,0 +1,200 @@
...
+            );
+        } else if (proxyType == ProxyType.ResolvedDelegate) {
+            string memory name = getProxyName(address(proxy));
+            Lib_AddressManager(address(proxy)).setAddress(name, implementation);
```
smartcontracts, line 168: `Lib_AddressManager(address(proxy))` is wrong here, proxy is not the AM

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
smartcontracts: Sorry I was looking at the wrong contract, nvm

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
smartcontracts: We're modifying this to be compatible

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
smartcontracts: + Same comment where used elsewhere

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
tynes: Ah good call, need to add unit tests for each of these

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -0,0 +1,266 @@
...
+     * @param proxy           The proxy to call
+     * @param implementation  The implementation to upgrade the proxy to
+     * @param data            The calldata to pass to the implementation
+     */
+    function upgradeAndCall(
+        Proxy proxy,
+        address implementation,
+        bytes memory data
+    ) external payable onlyOwner {
+        upgrade(proxy, implementation);
```
tynes, line 251: If OZ, do upgrade and call, otherwise do the low level call + pass value

tynes: in ci, make sure there is no function selector clash with L1std bridge and chugsplash proxy

packages/contracts-bedrock/contracts/test/Proxy.t.sol
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+
+    function test_upgradeToAndCall() external {
+        {
+            // There should be nothing in the current proxy storage
+            uint256 result = SimpleStorage(address(proxy)).get(1);
+            assertEq(result, 0);
+        }
+
```
tynes, line 156: TODO: check event

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -0,0 +1,266 @@
...
+     */
+    function setAddressManager(address _address) external onlyOwner {
+        addressManager = Lib_AddressManager(_address);
+    }
+
+    /**
+     * @notice Set an address in the address manager. This is required
+     *         because only the owner of the AddressManager can set the
+     *         addresses in it.
+     * @custom:legacy
```
tynes, line 122: put custom above notice

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
tynes: This will not work because of the possibility of the delegatecall that may modify state

packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * @notice A modifier that reverts if not called by the owner
+     *         or by `address(0)` to allow `eth_call` to interact
+     *         with the proxy without needing to use low level storage
+     *         inspection. It is assumed that nobody controls the private
+     *         key for `address(0)`.
+     */
+    modifier proxyCallIfNotOwner() {
```
maurelian, line 67: proxyCallIfNotAdmin feels like a better name here

smartcontracts (approved):

packages/contracts-bedrock/contracts/legacy/L1ChugSplashProxy.sol
```diff
@@ -0,0 +1,290 @@
...
+
+    /***************
+     * Constructor *
+     ***************/
+
+    /**
+     * @param _owner Address of the initial contract owner.
+     */
+    constructor(address _owner) {
+        _setOwner(_owner);
```
maurelian, line 47: similar to my previous comment, I think we should just say `_admin` here and avoid the term `owner` in general as the implementation contracts will typically have an owner.

tynes: I would prefer to have 0 diff with the previous contract

maurelian: Oh good point. I was actually meant this comment in the context of the new Proxy. 


https://github.com/ethereum-optimism/optimism/blob/22b8cf19aa5715177aa553a6fe3878a4c2252c5c/packages/contracts-bedrock/contracts/universal/Proxy.sol#L51

maurelian: That'd also be more consistent with the usage: https://github.com/ethereum-optimism/optimism/blob/22b8cf19aa5715177aa553a6fe3878a4c2252c5c/packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol#L204

maurelian (approved):
Approving to unblock Kovan deployment. 
Will continue reviewing and request any changes in a future PR.

PR #2731, opened 2022-06-09
feat(BaseServiceV2):End immediately when looping

roninjin10:
The server does not end when we call stop because async loops do not end.  This makes it so the async loop stops right away.

### DIFF
--- a/packages/common-ts/src/base-service/base-service-v2.ts
+++ b/packages/common-ts/src/base-service/base-service-v2.ts
@@ -4,7 +4,6 @@ import Config from 'bcfg'
 import * as dotenv from 'dotenv'
 import { Command, Option } from 'commander'
 import { ValidatorSpec, Spec, cleanEnv } from 'envalid'
-import { sleep } from '@eth-optimism/core-utils'
 import snakeCase from 'lodash/snakeCase'
 import express, { Router } from 'express'
 import prometheus, { Registry } from 'prom-client'
@@ -62,6 +61,12 @@ export abstract class BaseServiceV2<
   TMetrics extends MetricsV2,
   TServiceState
 > {
+  /**
+   * Stops the main loop from running
+   */
+  private stopLoop: () => void = () => {
+    this.running = false
+  }
   /**
    * Whether or not the service will loop.
    */
@@ -446,7 +451,8 @@ export abstract class BaseServiceV2<
     if (this.loop) {
       this.logger.info('starting main loop')
       this.running = true
-      while (this.running) {
+
+      const doLoop = async () => {
         try {
           await this.main()
         } catch (err) {
@@ -460,9 +466,14 @@ export abstract class BaseServiceV2<
 
         // Sleep between loops if we're still running (service not stopped).
         if (this.running) {
-          await sleep(this.loopIntervalMs)
+          const timeout = setTimeout(doLoop, this.loopIntervalMs)
+          this.stopLoop = () => {
+            this.running = false
+            clearTimeout(timeout)
+          }
         }
       }
+      doLoop()
     } else {
       this.logger.info('running main function')
       await this.main()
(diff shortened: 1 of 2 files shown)


packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -51,6 +50,27 @@ export abstract class BaseServiceV2<
...
+   * If clearTimeout(this.pollingTimeout) is called the timeout will stop
+   */
+  private pollingTimeout: NodeJS.Timeout
+
+  /**
+   * The promise representing this.main
+   */
+  private mainPromise: ReturnType<typeof this.main>
+
+  private stopMainLoop = async () => {
```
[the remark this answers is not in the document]
roninjin10: done

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -406,29 +411,28 @@ export abstract class BaseServiceV2<
 
         // Sleep between loops if we're still running (service not stopped).
         if (this.running) {
```
tynes, line 413: Hmm what is the intended behavior here? Unclear where `this.running` is set

roninjin10: It gets set immediately on 398 before the service starts running and when server is told to shut down it gets set to false.   The reason we need this.running in addition to ending the polling promise is for one specific corner case of the server is told to stop while this.main on like 402 is in the middle of running.   It only exists to make the the polling still stops if that race condition happens.

roninjin10: Now that I think about it, we could get rid of this.running and use this.loop as source of truth here. f Set this.loop to false when looping is supposed to end.[

roninjin10: Currently this.loop is set in the user options and never changes.

roninjin10: I got rid of this.running in favor of using this.loop in this commit https://github.com/ethereum-optimism/optimism/pull/2731/commits/f59c504d2e850963ee5cf9edf266e5085b203c4f

smartcontracts: I prefer having a separate `running` boolean. `this.loop` is supposed to represent whether or not the service is supposed to loop, `this.running` represents whether the service is currently running.

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -438,29 +443,28 @@ export abstract class BaseServiceV2<
...
-
-    this.done = true
   }
 
   /**
    * Tries to gracefully stop the service. Service will continue running until the current loop
    * iteration is finished and will then stop looping.
    */
   public async stop(): Promise<void> {
+    this.logger.info('Stopping main loop...')
```
smartcontracts, line 461: Last minor nit but we're using lower case log strings

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -460,29 +465,28 @@ export abstract class BaseServiceV2<
...
-    // Wait until the main loop has finished.
-    this.logger.info('stopping service, waiting for main loop to finish')
-    while (!this.done) {
-      await sleep(1000)
-    }
+    clearTimeout(this.pollingTimeout)
+    this.logger.info('waiting for main to complete')
+    // if main is in the middle of running wait for it to complete
+    await this.mainPromise
+    this.logger.info('main loop stoped.')
```
smartcontracts, line 489:
```suggestion
    this.logger.info('main loop stopped')
```

small typo

smartcontracts (commented):
Small typo, otherwise lgtm. Let's fix that then can approve and merge.

tynes (approved):

PR #2735, opened 2022-06-09
bedrock: remove starting block number from output oracle

maurelian:
This is a single commit building on #2707 which removes the `STARTING_BLOCK_NUMBER` value from the output oracle, as it is now redundant with `HISTORICAL_TOTAL_BLOCKS`. 

However the e2e go tests are now failing on it and I'm confused as to why that might be.


maurelian:
I no longer think this change is beneficial.

PR #2736, opened 2022-06-09
feat(cmn): have BSV2 export metadata metrics

smartcontracts:
**Description**
Updates BaseServiceV2 to require a version string. Also now exports
parameters in a synthetic metric but ignores any parameters explicitly
marked secret.


roninjin10 (approved):

packages/drippie-mon/src/service.ts
```diff
@@ -76,15 +73,6 @@ export class DrippieMonService extends BaseServiceV2<
       DrippieArtifact.abi,
       this.options.rpc
     )
-
-    this.metrics.metadata.set(
-      {
-        // eslint-disable-next-line @typescript-eslint/no-var-requires
-        version: require('../package.json').version,
```
roninjin10, line 83: nice, definitely way more readable to do it this way

PR #2737, opened 2022-06-09
contracts-governance: fix the mint manager

tynes:
**Description**
Fixes the `MintManager` contract which allowed for 20% inflation per year instead of 2% inflation per year. We will need to upgrade the `MintManager` and deploy this new contract. Also follow the checks, effects, interactions pattern inside of `mint`. Reentrency isn't possible assuming that this contract is initialized with the correct `GovernanceToken` but no need to keep that assumption. Also update to use seaport style comments


smartcontracts (approved):

maurelian (approved):
LGTM

PR #2748, opened 2022-06-10
contracts: Fix L-01 use _from instead of msg.sender

maurelian:
This PR fixes the Low and Informational severity issues issues in the OpenZeppelin audit (which is not yet public). Each commit in this PR clearly references the finding that it addresses.

In order to review, I suggest comparing between the audit report (see the "Open Zeppelin Audit Resolution" doc in Notion), and the commit which mentions it.


smartcontracts:
Do we need to pull this into the current version of the codebase?

smartcontracts:
I propose we make a simultaneous PR that commits these changes into develop

maurelian:
> I propose we make a simultaneous PR that commits these changes into develop

Good call, I'll do that. I'm having some trouble getting the patch to apply, atm though, and don't think it should block this so we can pass back to OZ>

maurelian:
Merging this. 
Internal review can be done on #2763

PR #2749, opened 2022-06-10
OZ Audit fix medium severity issues (on Bedrock)

maurelian:
This PR fixes the Low and Informational severity issues issues in the OpenZeppelin audit (which is not yet public). Each commit in this PR clearly references the finding that it addresses.

In order to review, I suggest comparing between the audit report (see the "Open Zeppelin Audit Resolution" doc in Notion), and the commit which mentions it.

This PR is being applied to the branch in #2748, which builds on the state of the develop branch as it was immediately after merging the bedrock branch.


maurelian:
There is no changeset here, because contracts-bedrock was not a package at this point in history.

maurelian:
Merging this. 
Internal review can be done on #2765

PR #2750, opened 2022-06-10
feat(common):Normalize promBundle stats

roninjin10:
For our metrics to not suck we need to be able to pass in a bypass and normalize into promBundle

### DIFF
--- a/packages/common-ts/src/base-service/base-service-v2.ts
+++ b/packages/common-ts/src/base-service/base-service-v2.ts
@@ -49,6 +49,23 @@ export type MetricsSpec<TMetrics extends MetricsV2> = {
   }
 }
 
+export interface BaseServiceV2Params<TOptions, TMetrics extends MetricsV2> {
+  name: string
+  version: string
+  optionsSpec: OptionsSpec<TOptions>
+  metricsSpec: MetricsSpec<TMetrics>
+  options?: Partial<TOptions>
+  loop?: boolean
+  loopIntervalMs?: number
+  port?: number
+  hostname?: string
+  /**
+   * Optionally override promBundleConfig defaults
+   * https://www.npmjs.com/package/express-prom-bundle
+   */
+  promBundleConfig?: promBundle.Opts
+}
+
 export type ExpressRouter = Router
 
 /**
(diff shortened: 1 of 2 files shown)


smartcontracts:
Hmmmm before we do this, I want to take a look at what the metrics would look like right now. My hope is we can normalize this automatically instead of having the user pass it in.

tynes (approved):

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -407,8 +435,6 @@ export abstract class BaseServiceV2<
 
       // Registery user routes.
       if (this.routes) {
-        const router = express.Router()
-        this.routes(router)
         app.use('/api', router)
```
smartcontracts, line 438: Can this whole block be moved up?

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -389,12 +393,36 @@ export abstract class BaseServiceV2<
...
           includeStatusCode: true,
+          bypass: (req) => {
+            // Note we are bypassing all routes that aren't at /api/* by only including the router.stack
+            for (const r of router.stack) {
+              if (req.path.includes(r.path)) {
+                return false
+              }
+            }
+            return true
+          },
```
smartcontracts, line 417: Does this mean we're not recording metrics for paths that don't exist?

roninjin10: Yes this was a specific request from @ben-chain because people be pinging our server for routes that don't exist all the time.

packages/common-ts/src/base-service/base-service-v2.ts
roninjin10: The other block where I define router be move down. though.  Should I do that?

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -141,17 +153,9 @@ export abstract class BaseServiceV2<
...
-    optionsSpec: OptionsSpec<TOptions>
-    metricsSpec: MetricsSpec<TMetrics>
-    options?: Partial<TOptions>
-    loop?: boolean
-    loopIntervalMs?: number
-    port?: number
-    hostname?: string
-  }) {
+  constructor(
+    private readonly params: BaseServiceV2Params<TOptions, TMetrics>
```
smartcontracts, line 157: I think it's ok to get rid of `private readonly` since we're parsing the params how we want them

mslipper (approved):

roninjin10:
Holding off on clicking merge for now[ ](https://optimism-workspace.slack.com/archives/C03J2VAAVDG/p1655508834480539)

mslipper:
Closing for now since this is fixed elsewhere.

PR #2753, opened 2022-06-10
feat(common-ts): Give access to server in tests

roninjin10:
This allows us to get the server without having to (this.server as any).server it.


packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -116,6 +116,12 @@ export abstract class BaseServiceV2<
    * App server.
    */
   protected server: Server
```
smartcontracts, line 118: I guess we could just make this public

roninjin10: It shouldn't be public because the router that is registered on `routes` is supposed to be a self contained server.  The user of base service shouldn't be able to touch the actual server object and if they do, they are likely doing something wrong.

tynes: I don't like the practice of having functions that are specific for tests. Can `server` be immutable and then we have a getter for it?

roninjin10:
@tynes yea I get that sentiment.   There is no way to make the server immutable.  What we can do though is provide a completely different function.

import {testServer} from 'eth-optimism/common-ts'

testServer(server)

That basically just wraps

import reqest from 'super-test'

And does the (app as any).server internally.

roninjin10:
@smartcontracts @tynes 

I added this as a seperate library so that it doesn't install supertest mocha jest whatever else ends up being added here as main dependencies with common-ts.  That way this can be a dev dependency.

I held off on the method to control the polling interval in tests for now.  Will add in future pr.

roninjin10:
we will finish this at some point but for now this isn't a big enough priority to push forward right away

PR #2755, opened 2022-06-10
feat: introduce hardhat-deploy-config 

smartcontracts:
**Description**
Creates a new package hardhat-deploy-config. We're using the same
configuration system for all of our contracts package, so might as well
turn it into a hardhat plugin to avoid duplicating code.


packages/hardhat-deploy-config/.env.example
```diff
@@ -0,0 +1,5 @@
+# URL for an L1 RPC provider, used to query L2 output proposals
+FAULT_DETECTOR__L1_RPC_PROVIDER=
+
```
tynes, line 3: This is a dead file

packages/hardhat-deploy-config/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "name": "@eth-optimism/hardhat-deploy-config",
+  "version": "0.1.0",
+  "description": "[Optimism] Hardhat deploy configuration plugin",
+  "main": "dist/src/index.js",
+  "types": "dist/src/index.d.ts",
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "start": "ts-node ./src/service.ts",
```
tynes, line 11: `start` is dead code?

smartcontracts:
@tynes to get this over the finish line

tynes:
The last major problem that I see here is that we lose the ability to have network specific config. There are a few possibilities on how to fix this, going to think about it a bit

maurelian (approved):
All looks quite reasonable to me.

PR #2764, opened 2022-06-13
feat(ctb): update L2 contract style

smartcontracts:
**Description**
Updates L2 contract style to match L1 contract style.


packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
```diff
@@ -1,43 +1,47 @@
...
+ *         transaction fee will be.
  */
 contract GasPriceOracle is Ownable {
-    /*************
-     * Variables *
-     *************/
-
-    // backwards compatibility
+    /**
+     * @custom:legacy
```
maurelian, line 20: Does anything parse these currently?

maurelian (approved):

packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
smartcontracts: Not yet, but we'll set that up

PR #2767, opened 2022-06-13
feat(ci): download and cache solidity compilers

smartcontracts:
**Description**
Downloads and caches the Solidity compilers to avoid CI errors where
circle seems to be getting rate limited. Checks compilers against known
hashes to guarantee that we're not getting any sort of error during this
caching step.


mslipper (approved):

trianglesphere (approved):

tynes (approved):
Missing the 0.5 solc for weth but lgtm, would recommend breaking into a bash function

PR #2771, opened 2022-06-13
(WIP) derive: Implement BatchQueue with full window

trianglesphere:
This implements a simple algorithm for the batch queue. It waits
until it has a full sequence window and then runs the historical
batch derivation process over that data.

The WIP part is that it needs more data that it does not yet have.

**Description**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -1,13 +1,22 @@
 package derive
 
 import (
+	"context"
+	"errors"
 	"fmt"
+	"time"
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


op-node/rollup/derive/batch_queue.go
```diff
@@ -45,16 +51,64 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) error {
...
+			PrevRandao:            eth.Bytes32(l1Info.MixDigest()),
+			SuggestedFeeRecipient: bq.config.FeeRecipientAddress,
+			Transactions:          txns,
+			// we are verifying, not sequencing, we've got all transactions and do not pull from the tx-pool
+			// (that would make the block derivation non-deterministic)
+			NoTxPool: true,
+		}
+		attributes = append(attributes, attrs) // TOOD: direct assignment here
+	}
+
```
protolambda, line 110: Don't forget we'll need something like
```
bq.inputs = bq.inputs[1:]
```
to continue to the next origin, after we return all batches.

PR #2772, opened 2022-06-14
make: Misc Makefile Improvements

norswap:
- add `make nuke` which removes ALL build output/caches/etc (everything that is `.gitignore`d)
    - also has `clean` and `devnet-clean` as dependencies
- add `-v` flag to `xargs` in `make devnet-clean` so that it doesn't error when there are no volumes or images to delete
- if `nvm` is installed, emit `nvm use` to switch to the `.nvmrc` version before compiling js
- add `yarn install` to `build-ts`: it doesn't add much overhead, and avoids confusion when dummies like me forget to run it beforehand (there is really nothing in the subsequent error message that tells you that this is what is going on)


Inphi (approved):
nvmind. This looks good.

mslipper (approved):

Issue #2776, opened 2022-06-14
Fresh installation and build hits a Hardhat error


snario:
To reproduce:

```bash
gh repo clone ethereum-optimism/optimism
cd optimism
yarn
yarn build
```

Error log:
```
lerna ERR! yarn run build exited 1 in '@eth-optimism/contracts-bedrock'
lerna ERR! yarn run build stdout:
$ hardhat compile && tsc && hardhat typechain

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run build stderr:
Solidity 0.8.10 is not fully supported yet. You can still use Hardhat, but some features, like stack traces, might not work correctly.

Learn more at https://hardhat.org/reference/solidity-support

An unexpected error occurred:
TypeError: Cannot read property 'code' of null
    at analyzeModuleNotFoundError (/Users/liam/code/optimism/optimism/node_modules/hardhat/src/internal/core/config/config-loading.ts:161:13)
    at Environment.run (/Users/liam/code/optimism/optimism/node_modules/hardhat/src/internal/core/runtime-environment.ts:136:33)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async main (/Users/liam/code/optimism/optimism/node_modules/hardhat/src/internal/cli/cli.ts:218:5)
error Command failed with exit code 1.
lerna ERR! yarn run build exited 1 in '@eth-optimism/contracts-bedrock'
lerna WARN complete Waiting for 2 child processes to exit. CTRL-C to exit immediately.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

PR #2777, opened 2022-06-14
Check if Foundry is installed before building

snario:
**Description**
Checks if Foundry is installed before running `build` in `contracts-bedrock`.

**Metadata**
Fixes https://github.com/ethereum-optimism/optimism/issues/2776


packages/contracts-bedrock/package.json
```diff
@@ -14,6 +14,7 @@
   ],
   "scripts": {
     "build:forge": "forge build",
+    "prebuild": "./forge-install-check",
```
tynes, line 17: Can you put this in the scripts directory?

tynes:
This prevents users on windows from building unless they are in a wsl environment. I don't believe anybody on the team uses windows besides @norswap, I'm not sure how many are out there. I'm not opposed if everybody thinks this is fine

I originally wanted to use something like [this](https://github.com/foundry-rs/hardhat/blob/ba96df24332eec3b38bf6997ef74f8391c0bcccf/packages/easy-foundryup/src/binary.ts#L41) but never got around to it

snario:
@tynes replaced Shell script with TypeScript script using `command-exists` which is what the script you linked uses.

mslipper (approved):

tynes (approved):

PR #2778, opened 2022-06-14
bedrock: fix OZ-H-O2 Token address ordering

maurelian:
This change addresses the H-O2 finding in the OZ audit extension.

This PR is on develop. I've merged the same fixes to the bedrock branch in #2779.


packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -226,8 +226,8 @@ abstract contract StandardBridge {
             // Something went wrong during the bridging process, return to sender.
             // Can happen if a bridge UI specifies the wrong L2 token.
             _initiateBridgeERC20Unchecked(
-                _remoteToken,
                 _localToken,
+                _remoteToken,
```
maurelian, line 230: This ordering matches the actual order of arguments in the definition of `_initiateBridgeERC20Unchecked()`:

https://github.com/ethereum-optimism/optimism/blob/01a515db6e8a8b262b6c166c3e6e2a91076107e0/packages/contracts-bedrock/contracts/universal/StandardBridge.sol#L342-L350

There is no need to reverse the ordering here, since they are reversed in the construction of the cross-domain message (see the comment added below).

tynes:
It appears as if this change doesn't cause any tests to fail - it would be ideal to add a unit test that would fail without the change and then passes with the change

maurelian:
@tynes Added that test.

tynes (approved):

PR #2780, opened 2022-06-14
feat(ctp): add deploy configuration for NFT bridge

smartcontracts:
**Description**
Introduces deploy configuration + deploy scripts for the NFT bridge.


packages/hardhat-deploy-config/src/types.ts
```diff
@@ -0,0 +1,10 @@
+export type DeployConfigSpec<
+  TDeployConfig extends {
+    [key: string]: any
+  }
+> = {
+  [K in keyof TDeployConfig]: {
+    type: 'address' | 'number' | 'string' | 'boolean'
+    default?: any
```
tynes, line 8: `default` should be a string or number?

PR #2781, opened 2022-06-14
wip: genesis builder

tynes:
**Description**
WIP script for building the L2 genesis file


tynes:
This PR places a proxy at every possible predeploy addr with the admin set to a configurable value and the implementations of existing predeploys set at in the `0xc0d3..` namespace

packages/contracts-bedrock/tasks/genesis.ts
```diff
@@ -0,0 +1,268 @@
...
+
+      const storageLayout = await getStorageLayout(hre, name)
+      const slots = computeStorageSlots(storageLayout, variables[name])
+
+      for (const slot of slots) {
+        alloc[allocAddr].storage[slot.key] = slot.val
+      }
+    }
+
+    const genesis: Genesis | any = {
```
tynes, line 236: Modern `geth` seems to like different types than the geth we are using for the current system, this is why the type is extended to `any`. Will require a breaking change for the type, so a major version of core-utils or packages that use this type need to be pinned but that isn't useful for outside devs

packages/contracts-bedrock/deploy-config/devnetL1.ts
```diff
@@ -6,8 +6,31 @@ const config = {
   genesisOutput: ethers.constants.HashZero,
   historicalBlocks: 0,
   startingBlockTimestamp:
-    parseInt(process.env.L2OO_STARTING_BLOCK_TIMESTAMP, 10) || Date.now(),
+    parseInt(process.env.L2OO_STARTING_BLOCK_TIMESTAMP, 16) || Date.now(),
```
mslipper, line 9: Isn't it easier to specify as a decimal rather than in hex?

tynes: The existing devnet specifies it as hex, happy to make a larger change if requested:
- https://github.com/ethereum-optimism/optimism/blob/9aa8049c199758b1a3656fe285f7cfc889db0be7/ops-bedrock/devnet-up.sh#L63
- https://github.com/ethereum-optimism/optimism/blob/9aa8049c199758b1a3656fe285f7cfc889db0be7/ops-bedrock/devnet-up.sh#L82

packages/contracts-bedrock/deploy-config/devnetL1.ts
```diff
@@ -6,8 +6,31 @@ const config = {
...
+
+  l1BlockInitialNumber: 0,
+  l1BlockInitialTimestamp: 0,
+  l1BlockInitialBasefee: 0,
+  l1BlockInitialHash: ethers.constants.HashZero,
+  l1BlockInitialSequenceNumber: 0,
+
+  proxyAdmin: '0x829BD824B016326A401d083B33D092293333A830',
+  genesisBlockExtradata: ethers.utils.hexConcat([
+    '0x',
```
tynes, line 26: This line is dead code, the blank `0x`

packages/contracts-bedrock/tasks/genesis.ts
```diff
@@ -0,0 +1,268 @@
...
+        symbol: 'OP',
+        _owner: hre.deployConfig.proxyAdmin,
+      },
+    }
+
+    assertEvenLength(implementationSlot)
+    assertEvenLength(adminSlot)
+    assertEvenLength(hre.deployConfig.proxyAdmin)
+
+    const predeployAddrs = new Set(Object.values(predeploys))
```
tynes, line 129: This set should be initialized as empty and all address comparisons should be done on checksummed addresses

packages/contracts-bedrock/hardhat.config.ts
```diff
@@ -80,6 +81,20 @@ const config: HardhatUserConfig = {
...
+      {
+        artifacts: '../contracts/artifacts',
+      },
+      {
+        artifacts: '../contracts-governance/artifacts',
+      },
+    ],
+    deployments: {
+      kovan: ['../contracts/deployments/kovan'],
+      goerli: ['../contracts/deployments/goerli'],
```
tynes, line 95: This might not be necessary

packages/contracts-bedrock/deploy-config/devnetL1.ts
tynes: Perhaps a fix to this would be to have two tasks:
- `genesis-l1`
- `genesis-l2`

Both of them are ran as part of `devnet-up.sh`

mslipper: Ah got it, then let's just leave as-is. Thanks for clarifying.

trianglesphere:
This PR needs to figure out how to integrate the new fee config: https://github.com/ethereum-optimism/optimism/pull/2804

packages/contracts-bedrock/foundry.toml
```diff
@@ -13,4 +13,6 @@ remappings = [
   'ds-test/=node_modules/ds-test/src'
 ]
 extra_output = ['devdoc', 'userdoc', 'metadata', 'storageLayout']
-bytecode_hash = "none"
+bytecode_hash = 'none'
+
+# build_info = true
```
tynes, line 18: Waiting on a new release of `foundry-rs/hardhat`

trianglesphere (approved):
lgtm.
Do you know when the new version of foundry will be coming out / when we need to remove that comment?

mslipper (approved):

mslipper:
@mergifyio refresh

PR #2782, opened 2022-06-14
ci: Remove buggy assignments rule

mslipper:
Removes the Mergify rule that requests reviewers. We need to redo who gets reviewed when. In the meantime, this prevents spammy review requests from getting through.


Inphi (approved):
😌

trianglesphere (approved):

tynes (approved):

PR #2784, opened 2022-06-15
ci: Fix broken js-builder build

mslipper:
The curl commands weren't outputting files and the checksum verification was breaking other commands.


smartcontracts (approved):

Inphi (approved):
Looks good.
But i'm curious as to how it worked before if the curl command download anything?

PR #2786, opened 2022-06-15
fix: revert es target back to 2017

smartcontracts:
**Description**
We accidentally bumped the es target to 2022 when it should've remained
at 2017.


Sednaoui:
thanks!

PR #2787, opened 2022-06-15
bedrock: Proxy Admin improvements

maurelian:
**Description**
Some changes to improve the Proxy Admin. See inline comments.


packages/contracts-bedrock/contracts/test/ProxyAdmin.t.sol
```diff
@@ -90,7 +90,7 @@ contract ProxyAdmin_Test is Test {
     function test_proxyType() external {
         assertEq(
             uint256(admin.proxyType(address(proxy))),
-            uint256(ProxyAdmin.ProxyType.OpenZeppelin)
+            uint256(ProxyAdmin.ProxyType.ERC1967)
```
maurelian, line 93: I prefer this naming to avoid confusion in the future, as we're not using the exact OpenZeppelin implementation.

packages/contracts-bedrock/contracts/test/ProxyAdmin.t.sol
```diff
@@ -164,19 +164,26 @@ contract ProxyAdmin_Test is Test {
...
+        // The proxy is no longer the admin and can
         // no longer call the proxy interface except for
-        // the ResolvedDelegate type which anybody can call
-        // the admin interface
-        if (proxyType != ProxyAdmin.ProxyType.ResolvedDelegate) {
-            vm.expectRevert();
+        // the ResolvedDelegate type on which anybody can
+        // call the admin interface.
+        if (proxyType == ProxyAdmin.ProxyType.ERC1967) {
+            vm.expectRevert("Proxy: implementation not initialized");
```
maurelian, line 172: A change below in the ProxyAdmin uses a high level syntax to make a static call, which also bubbles up the error message, so can't just use a generic `expectRevert()` here. 

(I was surprised that `expectRevert()` fails in the case there there is an error message).

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -169,25 +174,15 @@ contract ProxyAdmin is Owned {
...
+        if (proxyType == ProxyType.ERC1967) {
+            return IStatic_ERC1967Proxy(address(proxy)).admin();
         } else if (proxyType == ProxyType.Chugsplash) {
-            target = address(proxy);
-            data = abi.encodeWithSelector(L1ChugSplashProxy.getOwner.selector);
+            return IStatic_L1ChugSplashProxy(address(proxy)).getOwner();
         } else if (proxyType == ProxyType.ResolvedDelegate) {
-            target = address(addressManager);
-            data = abi.encodeWithSignature("owner()");
+            return Lib_AddressManager(address(addressManager)).owner();
```
smartcontracts, line 182: Same comment as above, unnecessary cast

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -6,25 +6,40 @@ import { Owned } from "@rari-capital/solmate/src/auth/Owned.sol";
...
+     * @custom:value ERC1967          Represents the ERC1967 transparent proxy
      *                                interface, this is the standard.
      * @custom:value Chugsplash       Represents the Chugsplash proxy interface,
      *                                this is legacy.
      * @custom:value ResolvedDelegate Represents the ResolvedDelegate proxy
      *                                interface, this is legacy.
      */
     enum ProxyType {
-        OpenZeppelin,
+        Unknown,
```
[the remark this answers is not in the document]
tynes: Agreed with this, we should remove the unknown type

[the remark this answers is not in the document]
maurelian: I realized that was the intent, but still it seems like pretty subtle behavior to me, and not that difficult to call `setProxyType`. 
Not having a default type is also more future proof IMO, especially in case we (god forbid), adopt yet another proxy type in the future.

tynes:
Need to regenerate the gas snapshot

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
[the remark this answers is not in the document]
maurelian: OK.

tynes (approved):

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -228,6 +224,8 @@ contract ProxyAdmin is Owned {
         } else if (proxyType == ProxyType.ResolvedDelegate) {
             string memory name = implementationName[address(proxy)];
             Lib_AddressManager(addressManager).setAddress(name, implementation);
```
smartcontracts, line 226: Final location for the address manager thing

tynes:
This could also use a changeset

smartcontracts (approved):

PR #2795, opened 2022-06-15
fix: switch between empty/forked chain without the need to recreate t…

vasapower:
…he entire image

**Description**
Currently `l1_chain` Docker image doesn't allow to fork a preexistent chain because the `dotenv ` repo is not loaded nor initialized. Adding the latter to the `package.json` and using it in the `hardhat.config.js` file solves the issue.

**Additional context**
The current image works well if you need to run a fresh L1 chain. However, the `l1_chain.env` configuration file is not read correctly by the container, so it's not possible to use a forked mainnet.

**Metadata**
No issues links

### DIFF
--- a/.changeset/old-carpets-worry.md
+++ b/.changeset/old-carpets-worry.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/hardhat-node': minor
+---
+
+Correct configuration file behavior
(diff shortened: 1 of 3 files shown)


tynes:
Can you add a changeset for the hardhat package?

vasapower:
> Can you add a changeset for the hardhat package?

Done

tynes (approved):

PR #2797, opened 2022-06-16
op-bindings: Provide common.Address types of predeploys

trianglesphere:
**Description**
We provide both a string literal which contains the address as const
a hex string and as the type common.Address.

We typically use addresses in the common.Address type so this avoids
users having to do the conversion manually.


mslipper (approved):

protolambda (approved):
LGTM

PR #2801, opened 2022-06-16
proxyd: Use canned response for eth_accounts

mslipper:
We never want to expose Geth's accounts to the public internet, so proxyd will now return `[]` for `eth_accounts` RPC calls without hitting the backend.


optimisticben (approved):
lmtm

Inphi (approved):
Looks good.

Super nit incoming:
However, this makes proxyd even more specific to our infrastructure. A couple folks have expressed interest into using proxyd for generic JSON-RPC routing. I dunno if anyone else deploys proxyd for other networks, but this would be a surprising change in behavior.

optimisticben:
> Super nit incoming: However, this makes proxyd even more specific to our infrastructure. A couple folks have expressed interest into using proxyd for generic JSON-RPC routing. I dunno if anyone else deploys proxyd for other networks, but this would be a surprising change in behavior.

That's a good point. This should be behind a feature flag tbh.

Issue #2798, opened 2022-06-16
Crash: replica-healthcheck


breezytm:
**Describe the bug**
The ``replica-healthcheck`` container is constantly crashing. Both ``data-transport-layer`` and ``l2geth-replica`` seem to be running. The only deviation from the documentation is I changed the ports in``replica-shared.yml`` from 8545, and 8546 to 8555 and 8556. Both ports are currently being used by Ethereum mainnet on the same host. A little help would be nice 🙏

```replica-healthcheck_1   | > @eth-optimism/replica-healthcheck@1.1.2 start
replica-healthcheck_1   | > ts-node ./src/service
replica-healthcheck_1   | 
replica-healthcheck_1   | /opt/optimism/packages/common-ts/src/base-service/base-service-v2.ts:303
replica-healthcheck_1   |         throw new Error(`missing required option: ${optionName}`)
replica-healthcheck_1   |               ^
replica-healthcheck_1   | Error: missing required option: referenceRpcProvider
replica-healthcheck_1   |     at new BaseServiceV2 (/opt/optimism/packages/common-ts/src/base-service/base-service-v2.ts:303:15)
replica-healthcheck_1   |     at new HealthcheckService (/opt/optimism/packages/replica-healthcheck/src/service.ts:34:5)
replica-healthcheck_1   |     at Object.<anonymous> (/opt/optimism/packages/replica-healthcheck/src/service.ts:208:19)
replica-healthcheck_1   |     at Module._compile (node:internal/modules/cjs/loader:1105:14)
replica-healthcheck_1   |     at Module.m._compile (/opt/optimism/node_modules/ts-node/src/index.ts:1310:23)
replica-healthcheck_1   |     at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
replica-healthcheck_1   |     at Object.require.extensions.<computed> [as .ts] (/opt/optimism/node_modules/ts-node/src/index.ts:1313:12)
replica-healthcheck_1   |     at Module.load (node:internal/modules/cjs/loader:981:32)
replica-healthcheck_1   |     at Function.Module._load (node:internal/modules/cjs/loader:822:12)
replica-healthcheck_1   |     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
replica-healthcheck_1   | npm ERR! Lifecycle script `start` failed with error: 
replica-healthcheck_1   | npm ERR! Error: command failed 
[...]

PR #2803, opened 2022-06-16
op-replica env should include new values

optimisticben:
**Description**
The v1+ version of the replica-healthcheck requires new env values.

Closes https://github.com/ethereum-optimism/optimism/issues/2798


infra/op-replica/docker-compose/default-mainnet.env
```diff
@@ -14,6 +13,8 @@ L2GETH_HTTP_PORT=9991
 L2GETH_WS_PORT=9992
 DTL_PORT=7878
 GETH_INIT_SCRIPT=check-for-chaindata-berlin.sh
+HEALTHCHECK__REFERENCE_RPC_PROVIDER=https://mainnet.optimism.io
```
smartcontracts, line 16: We might want to ask people to bring their own L2 RPC endpoints rather than pointing them at mainnet.optimism.io. More and more people are running nodes, so this will start to get taxing on the public endpoint.

mslipper (approved):

PR #2804, opened 2022-06-16
ops: Enabled Optimism fee config

trianglesphere:
**Description**
Enable L2 Fees to be properly collected and paid out in the devnet.

The baseFeeRecipient is as follows:
- HD Path: `m/44'/60'/0'/0/10`
- Address: `0xBcd4042DE499D14e55001CcbB24a551F3b954096`

The l1FeeRecipient is as follows:
- HD Path: `m/44'/60'/0'/0/11`
- Address: `0x71bE63f3384f5fb98995898A86B02Fb2426c5788`

### DIFF
--- a/ops-bedrock/genesis-l2.json
+++ b/ops-bedrock/genesis-l2.json
@@ -13,7 +13,12 @@
     "berlinBlock": 0,
     "londonBlock": 0,
     "mergeForkBlock": 0,
-    "terminalTotalDifficulty": 0
+    "terminalTotalDifficulty": 0,
+    "optimism": {
+      "enabled": true,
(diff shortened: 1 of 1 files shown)


mslipper (approved):

trianglesphere:
Damn, this is gonna conflict with mark's PR which generates this from scratch every time

tuxcanfly:
Should updating values in `GasPriceOracle` be a part of devnet script? That will fix the issue with `l1FeeRecipient`

tynes:
If we stayed.setbalance it doesn't execute evm so the l1cost recipient should receive the eth?

PR #2805, opened 2022-06-16
fix(cmm): BSV2 update metrics to more permissive type

roninjin10:
The index type makes any type extending it require the same index type (in more recent typescript versions).   Using Record makes the type work more as intended.   

e.g.
https://github.com/ethereum-optimism/gateway/pull/418/files#diff-ef065f0877f6777b77bdbeadb0a7d176efd075a0f33f46efbef92a8f320873daR41


packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -34,9 +34,7 @@ export type OptionsSpec<TOptions extends Options> = {
   }
 }
 
-export type MetricsV2 = {
-  [key: string]: Metric
-}
+export type MetricsV2 = Record<any, Metric>
```
roninjin10, line 37: We could be nits here with 
export type MetricsV2<TMetricKeys extends string> = Record<TMetricKeys, Metrics>
But that adds little to no value while bubbling up generic typescript types making base service more complicated.

mslipper (approved):

smartcontracts:
Ah I see. I've been "solving" this by making metrics a `type` instead of an `interface`

tynes (approved):

PR #2808, opened 2022-06-17
Fix typo

Vsevo1od:
**Description**
safeTransferFrom checks if the *token* contract has code, not a *from* address. So this looks like a typo.

**Additional context**
-

**Metadata**
-

### DIFF
--- a/packages/contracts/contracts/L1/messaging/L1StandardBridge.sol
+++ b/packages/contracts/contracts/L1/messaging/L1StandardBridge.sol
@@ -182,7 +182,7 @@ contract L1StandardBridge is IL1StandardBridge, CrossDomainEnabled {
     ) internal {
         // When a deposit is initiated on L1, the L1 Bridge transfers the funds to itself for future
         // withdrawals. safeTransferFrom also checks if the contract has code, so this will fail if
-        // _from is an EOA or address(0).
+        // _l1Token is an EOA or address(0).
         // slither-disable-next-line reentrancy-events, reentrancy-benign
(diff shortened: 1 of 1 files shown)


tynes (approved):

maurelian:
Thanks for this @Vsevo1od! Though I'm closing as the fix is not quite right, and I have a replacement in #2831. 

For ref, see the impl of `safeTransferFrom` [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d4fb3a89f9d0a39c7ee6f2601d33ffbf30085322/contracts/token/ERC20/utils/SafeERC20.sol#L88).

Vsevo1od:
@maurelian thank you!
I think what the comment may meant is that inside `_callOptionalReturn` (which is called by `safeTransferFrom`) is a [`.functionCall`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d4fb3a89f9d0a39c7ee6f2601d33ffbf30085322/contracts/token/ERC20/utils/SafeERC20.sol#L93)


Which checks if the called address [has code](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d4fb3a89f9d0a39c7ee6f2601d33ffbf30085322/contracts/utils/Address.sol#L135)
```solidity
        require(isContract(target), "Address: call to non-contract");
```


And the called address is `_l1token` in this case.

But your fix is fine with me. 
I just got a little confused while reading the code and decided to share :)

maurelian:
Thanks, and definitely appreciate the help! :)

PR #2825, opened 2022-06-20
bedrock: Make portal upgradeable

maurelian:
**Description**

Adds upgradability to the `OptimismPortal` contract, and makes its parent contract `ResourceMetering` amenable to upgradeability.

**Additional context**

1. Used OZ's `Initializable` code and pattern.
2. `ResourceMetering` doesn't inherit from `Initializable` itself, but the contract has been made `abstract`, and the constructor was converted to an `internal` function called `__ResourceMetering_init`.  This makes it easier to use in 3rd party contracts which are not  upgradeable.
3. Because the Portal's constructor calls `initialize()`. The contract's implementation still works without being behind a proxy. This is nice because it means we don't need to add a proxy to unit tests, but integration tests should be updated to deploy behind a proxy.


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -83,12 +89,28 @@ contract OptimismPortal is ResourceMetering {
...
+     * @notice The constructor sets immutable values in the implementation.
+     *         This means that these values can only changed by an upgrade. But the efficiency gains
+     *         are worthwhile.
+     *         Also ensures that the implementation is initialized upon deployment.
      */
     constructor(L2OutputOracle _l2Oracle, uint256 _finalizationPeriodSeconds) {
+        // Set these immutable values into the bytcode of the implementation.
         L2_ORACLE = _l2Oracle;
         FINALIZATION_PERIOD_SECONDS = _finalizationPeriodSeconds;
+        initialize();
```
tynes, line 106: OZ recommends this pattern in the constructor. It would break any existing testing infra, we would need to write the code that sets the proxy in both the solidity tests + the go tests.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/83277ff916ac4f58fec072b8f28a252c1245c2f1/contracts/proxy/utils/Initializable.sol#L52

tynes: I'm not super opinionated, curious what you think

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -83,12 +89,28 @@ contract OptimismPortal is ResourceMetering {
...
+     */
+    uint256[48] private __gap;
+
+    /**
+     * @notice The constructor sets immutable values in the implementation.
+     *         This means that these values can only changed by an upgrade. But the efficiency gains
+     *         are worthwhile.
+     *         Also ensures that the implementation is initialized upon deployment.
      */
     constructor(L2OutputOracle _l2Oracle, uint256 _finalizationPeriodSeconds) {
```
tynes, line 102: These constructor args need to be set via `initialize`

maurelian: They are immutable, so they have to be in the constructor. I verified there are [no gotchas](https://forum.openzeppelin.com/t/un-safe-usage-of-immutable-in-uups/28783/2?u=maurelian) with this approach, and these values should not be expected to change so often that they can't be modified via an upgrade.

https://github.com/ethereum-optimism/optimism/blob/m/upgradeable-portal/packages/contracts-bedrock/contracts/L1/OptimismPortal.sol#L62-L67

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -151,20 +142,28 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
         vm.expectCall(target, hex"1111");
 
         // set the value of op.l2Sender() to be the L2 Cross Domain Messenger.
-        uint256 senderSlotIndex = 1;
-        vm.store(address(op), bytes32(senderSlotIndex), bytes32(abi.encode(sender)));
+        uint256 senderSlotIndex = 50;
```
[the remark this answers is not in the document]
maurelian: Is this the syntax you had in mind? It doesn't look as though solidity supports this for variables in other contracts.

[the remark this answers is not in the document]
tynes: Ah yes this was the syntax that I was thinking of, unfortunately it does not work so I think the way that you are doing it now is fine

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: If they *must* be deployed behind proxies, then we should test them behind proxies in the unit tests. All of the proxy stuff could be done in the common tests. We will also need to update the hh deploy scripts to deploy proxies

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -206,8 +205,24 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
         L1Messenger.xDomainMessageSender();
 
         address sender = Lib_PredeployAddresses.L2_CROSS_DOMAIN_MESSENGER;
-        uint256 senderSlotIndex = 1;
-        vm.store(address(op), bytes32(senderSlotIndex), bytes32(abi.encode(sender)));
+
+        uint256 senderSlotIndex = 50;
```
tynes, line 209: I wish there was an easy way to prevent this from being brittle, sure its possible using `ffi` and grabbing the value from the storage layout snapshot but that is pretty slow. foundry has a `fs` library on the way, that could be used in the future

maurelian: I'm not worried, these values shouldn't change too much after this.

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -151,20 +142,28 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
...
+        uint256 senderOffset = 16;
+
+        // Retrieve the current value for the full slot, zero out the sender section, then
+        // write over that section with the address of the L2 Messenger.
+        bytes32 slotValue = vm.load(address(op), bytes32(senderSlotIndex));
+        bytes32 slotValueWithSender;
+        assembly {
+            slotValueWithSender := or(
+                and(0xffff, slotValue),
+                shl(16,sender)
```
tynes, line 157: nit: no space between args to `shl`

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -206,8 +205,24 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
...
```
tynes, line 221: same nit

tynes:
Could also use an update to the hh deploy script to call `updateToAndCall` where `initialize()` is called

tynes:
Perhaps a rebase will trigger CI

maurelian:
> Could also use an update to the hh deploy script to call `updateToAndCall` where `initialize()` is called

Already had that locally, just pushed :)

packages/contracts-bedrock/contracts/test/L2OutputOracle.t.sol
```diff
@@ -384,4 +385,31 @@ contract L2OutputOracleUpgradeable_Test is L2OutputOracle_Initializer {
...
+            abi.encodeWithSelector(
+                L2OutputOracle.initialize.selector,
+                genesisL2Output,
+                startingBlockNumber,
+                sequencer,
+                owner
+            )
+        );
+        assertEq(proxy.implementation(), address(newOracleImpl));
+
```
tynes, line 412: nit: extra newlines

tynes:
The tests are failing with this error:

```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
```

maurelian:
@tynes: https://github.com/foundry-rs/foundry/issues/2106

maurelian:
I can run subsets of the tests successfully, so we might need to hack around it by running `forge test --match-contract <name>_Test` a bunch of times?

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -271,14 +263,13 @@ contract OptimismPortal_Test is L2OutputOracle_Initializer {
     }
 }
 
-contract OptimismPortalUpgradeable_Test is CommonTest {
-    OptimismPortal opImpl;
+contract OptimismPortalUpgradeable_Test is Portal_Initializer {
     Proxy internal proxy;
-    L2OutputOracle oracle = L2OutputOracle(address(0));
+    // L2OutputOracle oracle = L2OutputOracle(address(0));
```
maurelian, line 268: I'll fix this.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -15,7 +15,7 @@ import { Initializable } from "@openzeppelin/contracts/proxy/utils/Initializable
  *         and L2. Messages sent directly to the OptimismPortal have no form of replayability.
  *         Users are encouraged to use the L1CrossDomainMessenger for a higher-level interface.
  */
-contract OptimismPortal is ResourceMetering, Initializable {
+contract OptimismPortal is Initializable, ResourceMetering {
```
maurelian, line 18: I realized that the Portal's `_initializing`/`_initialized` values were not in the first slot as they are in our other upgradable contracts. I thought that was worth preserving, and it meant I didn't have to do that bit shifting/masking in the messenger tests anymore.

tynes (approved):

packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -11,7 +11,7 @@ import { Burn } from "../libraries/Burn.sol";
  * @notice ResourceMetering implements an EIP-1559 style resource metering system where pricing
  *         updates automatically based on current demand.
  */
-contract ResourceMetering {
+abstract contract ResourceMetering {
```
Nicca42, line 14: Why is this marked abstract? It doesn't have any un-implemented functions, not sure if it needs this. Non-blocking though.

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -269,3 +262,43 @@ contract OptimismPortal_Test is L2OutputOracle_Initializer {
         op.finalizeWithdrawalTransaction(0, alice, alice, 0, 0, hex"", 0, outputRootProof, hex"");
     }
 }
```
Nicca42, line 264: Can we add a test for accessing the reserved `__gap` in an upgrade?

Nicca42 (changes requested):

packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
maurelian: Yeah, good call out... 
I think I made it `abstract` because of long-ish chain of things that are no longer entirely true.
So, I'll add a constructor and remove `abstract`.

maurelian: Did this [here](https://github.com/ethereum-optimism/optimism/pull/2825/commits/84868c79dff3a3385f509b8bf192991fd1947cf2).

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
maurelian: TBH I wasn't totally sure what you were looking for, but I think this gets at it: 

[`0965fc4` (#2825)](https://github.com/ethereum-optimism/optimism/pull/2825/commits/0965fc43075da77659783dc7db7e3962e65ad92b#diff-5e974f8d325ddec55609e9fac6947a369a9376db659fb40eeaf426537b54b3fe)

tynes (approved):

tynes:
This is almost ready to be merged, looks like it just has a merge conflict now

tynes (approved):

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
tynes: This does look like its testing what was requested

PR #2826, opened 2022-06-20
Make OuputOracle upgradeable

maurelian:
**Description**

_This PR is being opened as a draft, mostly because it builds on #2810._

It applies to the Oracle the same upgradability approach as #2825.


packages/contracts-bedrock/.storage-layout
```diff
@@ -67,28 +67,44 @@
...
 | Name              | Type                                                     | Slot | Offset | Bytes |
 +======================================================================================================+
-| _owner            | address                                                  | 0    | 0      | 20    |
+| _initialized      | uint8                                                    | 0    | 0      | 1     |
 |-------------------+----------------------------------------------------------+------+--------+-------|
-| sequencer         | address                                                  | 1    | 0      | 20    |
+| _initializing     | bool                                                     | 0    | 1      | 1     |
 |-------------------+----------------------------------------------------------+------+--------+-------|
-| latestBlockNumber | uint256                                                  | 2    | 0      | 32    |
+| __gap             | uint256[50]                                              | 1    | 0      | 1600  |
```
tynes, line 74: Why is this gas useful here? In case the ownable implementation upgrades?

maurelian: Yeah. It's built into [OwnableUpgradeable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/f5819bc47bdf0c049f9b1c2eed6b940cbff7ada2/contracts/access/OwnableUpgradeable.sol#L94)

tynes (approved):

tynes:
Its expected that the bedrock tests fail as the storage layout changes

tynes:
This can be done in a follow up PR, but this will need to be updated to use `upgradeToAndCall` with a call to initialize: https://github.com/ethereum-optimism/optimism/blob/700dcbb01eeeac93b4608d92e3f5d4366aedf039/packages/contracts-bedrock/deploy/000-L2OutputOracle.deploy.ts#L49

mslipper (approved):

PR #2832, opened 2022-06-22
fix(fd): properly handle connection failures

smartcontracts:
**Description**
Updates the Fault Detector to properly handle connection failures when
trying to query the L2 node. Also increments a new metric whenever this
happens so we can keep track of these failures if necessary.


tynes:
Looks like there is a type issue that is preventing it from building

packages/fault-detector/src/service.ts
```diff
@@ -151,15 +164,26 @@ export class FaultDetector extends BaseServiceV2<Options, Metrics, State> {
...
     for (let i = 0; i < batchSize; i += 1000) {
-      const provider = this.options
-        .l2RpcProvider as ethers.providers.JsonRpcProvider
-      blocks = blocks.concat(
-        await provider.send('eth_getBlockRange', [
+      let newBlocks: any[]
+      try {
+        newBlocks = await (
+          this.options.l2RpcProvider as ethers.providers.JsonRpcProvider
+        ).send('eth_getBlockRange', [
```
tynes, line 171: I recommend we attempt to migrate away from `eth_getBlockRange` so that the services can work against infura/alchemy since that RPC is not provided, this can be achieved with the batch json rpc provider

smartcontracts: Alchemy definitely supports this RPC, not sure about infura.

packages/fault-detector/src/service.ts
```diff
@@ -20,10 +20,12 @@ type Metrics = {
   highestCheckedBatchIndex: Gauge
   highestKnownBatchIndex: Gauge
   isCurrentlyMismatched: Gauge
-  inUnexpectedErrorState: Gauge
```
smartcontracts, line 23: Metric was not being set, can safely delete it

mslipper (approved):

tynes (approved):

PR #2839, opened 2022-06-22
contracts-bedrock: fix some compiler warnings

tynes:
**Description**
This PR fixes some of the compiler warnings emitted when compiling `contracts-bedrock`. There are more compiler warnings, but this gets some of the higher priority/low hanging fruit


mslipper (approved):

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
```diff
@@ -147,13 +147,13 @@ contract ProxyAdmin is Owned {
      * @return The address of the implementation.
      */
     function getProxyImplementation(Proxy proxy) external view returns (address) {
-        ProxyType proxyType = proxyType[address(proxy)];
+        ProxyType _proxyType = proxyType[address(proxy)];
 
-        if (proxyType == ProxyType.ERC1967) {
+        if (_proxyType == ProxyType.ERC1967) {
```
maurelian, line 152: this is a compiler warning?

maurelian (approved):

packages/contracts-bedrock/contracts/universal/ProxyAdmin.sol
tynes: Its shadowing another variable so renaming it with the `_` prefix fixes that

PR #2840, opened 2022-06-23
op-node: Add metrics

mslipper:
Adds a metrics server config and tracks metrics for the RPC server and RPC client calls. This PR is just a start - there are many more things we can track. Anything that might be affected by the batch derivation work was deliberately excluded.


op-node/metrics/metrics.go
```diff
@@ -0,0 +1,150 @@
...
+const (
+	Namespace = "op_node"
+
+	RPCServerSubsystem = "rpc_server"
+	RPCClientSubsystem = "rpc_client"
+
+	BatchMethod = "<batch>"
+)
+
+var (
```
protolambda, line 25: Should these metrics really be globals? We technically can have multiple node instances in the same process. And we do that in the go e2e testing for example. So these metrics would get data from entirely different nodes.
A metrics struct that we can pass to the rpc server and other node main components to avoid such collision would be nice.

mslipper: Done, updated to pass around a `metrics` object.

op-node/metrics/metrics.go
```diff
@@ -0,0 +1,177 @@
...
+// convert the passed-in error into something metrics friendly.
+// Nil errors get converted into <nil>, RPC errors are converted
+// into rpc_<error code>, HTTP errors are converted into
+// http_<status code>, and everything else is converted into
+// <unknown>.
+func (m *Metrics) RecordRPCClientResponse(method string, err error) {
+	var errStr string
+	if err == nil {
+		errStr = "<nil>"
+	} else if rpcErr, ok := err.(rpc.Error); ok {
```
trianglesphere, line 152: can use errors.Is and errors.As here

mslipper: Done

op-node/l2/source.go
```diff
@@ -6,29 +6,30 @@ import (
...
-	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
 type Source struct {
-	rpc     *rpc.Client       // raw RPC client. Used for the consensus namespace
-	client  *ethclient.Client // go-ethereum's wrapper around the rpc client for the eth namespace
+	rpc     client.RPC    // raw RPC client. Used for the consensus namespace
+	client  client.Client // go-ethereum's wrapper around the rpc client for the eth namespace
```
protolambda, line 24: Nice work on supporting metrics for the raw RPC calls too.

It's in my backlog to not use the `client.Client` bindings here at all; the default bindings have drawbacks (we don't need to decode txs, we're encoding them again after receiving them...) and similarly we're decoding/requesting other things that we don't actually read.

op-node/l2/source.go
```diff
@@ -218,16 +219,16 @@ func blockToBlockRef(block *types.Block, genesis *rollup.Genesis) (eth.L2BlockRe
 }
 
 type ReadOnlySource struct {
-	rpc     *rpc.Client       // raw RPC client. Used for methods that do not already have bindings
-	client  *ethclient.Client // go-ethereum's wrapper around the rpc client for the eth namespace
+	rpc     client.RPC    // raw RPC client. Used for methods that do not already have bindings
+	client  client.Client // go-ethereum's wrapper around the rpc client for the eth namespace
 	genesis *rollup.Genesis
 	log     log.Logger
```
protolambda, line 225: Planning to potentially unify this with the other L2 source bindings, or at least reduce duplication. Thank you for integrating metrics though.

op-node/l2/source.go
protolambda: And we can keep an eye on the metrics - I suspect we could really use caching well here (like we already have in the L1 source).

op-node/node/config.go
```diff
@@ -34,17 +38,38 @@ type RPCConfig struct {
...
+	ListenAddr string
+	ListenPort int
+}
+
+func (m MetricsConfig) Check() error {
+	if !m.Enabled {
+		return nil
+	}
+
+	if m.ListenPort < 0 || m.ListenPort > math.MaxUint16 {
```
protolambda, line 52: Could maybe type it as `uint16` directly, if the CLI library supports `uint16`. This works too.

protolambda (approved):
Looks good to me. Some nitpicks, but we can merge as-is.

The op-batcher and op-proposer already import op-node, so we can reuse the metrics tooling in those that way, but maybe we want to consider having a more general go module of go service building blocks that we want to reuse. E.g. the log configuration and CLI flags in the rollup-node are also useful to share (and that way cli flags are also more consistent).

op-node/node/config.go
mslipper: It doesn't, unfortunately :(. Just `uint` and `uint64`.

mslipper:
> maybe we want to consider having a more general go module of go service building blocks that we want to reuse

I think we should definitely do this. Something similar to what Kelvin did with `BaseServiceV2` would be ideal.

op-node/p2p/rpc_server.go
```diff
@@ -52,18 +54,22 @@ type Node interface {
...
+func NewP2PAPIBackend(node Node, log log.Logger, m *metrics.Metrics) *APIBackend {
 	return &APIBackend{
 		node: node,
 		log:  log,
+		m:    m,
 	}
 }
 
 func (s *APIBackend) Self(ctx context.Context) (*PeerInfo, error) {
+	recordDur := s.m.RecordRPCServerRequest("opp2p_self")
```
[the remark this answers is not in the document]
mslipper: We can modify the base RPC server, but I'd like to do that outside this PR.

trianglesphere (approved):

tynes (approved):

PR #2849, opened 2022-06-23
Correct l33t sp34k in toCodeAddr

maurelian:
A small change to ensure the code address prefix is repeated correctly. Not a big deal, but could prevent errors of manually rewriting an address.


trianglesphere (approved):
no more coed

PR #2853, opened 2022-06-27
ci: Use conditional builds

mslipper:
- Updates CircleCI jobs to exit early if the git diff doesn't require them to run. This will dramatically reduce our CircleCI usage and speed up the pipeline for projects like Bedrock that don't touch legacy code. The entire pipeline will still run on `develop` and `master`. The conditional logic is aware of package dependencies, so changing an upstream package will correctly trigger a build on downstream dependencies.
- Creates a new `ci-builder` image to replace `js-builder` and `go-builder`. The `ci-builder` image contains everything we need to run builds on CircleCI.
- Fixes a bunch of misspellings that should have failed the linter, but didn't.

You can see the conditional builds in action on this PR. The JS tests in the monorepo exit early but the integration tests still run since `bss-core` and the batch submitter changed. Here's a screenshot:

Screen Shot 2022-06-26 at 7 03 13 PM

### DIFF
--- a/ops/docker/ci-builder/check-changed.sh
+++ b/ops/docker/ci-builder/check-changed.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# Usage: check-changed.sh <diff-pattern>.
+#
+# This script compares the files changed in the <diff-pattern> to the git diff,
+# and writes TRUE or FALSE to stdout if the diff matches/does not match. It is
+# used by CircleCI jobs to determine if they need to run.
+
+set -e
+
+echoerr() { echo "$@" 1>&2; }
+
+# Check if this is a CircleCI PR.
+if [[ -n $CIRCLE_PULL_REQUEST ]]; then
+	PACKAGE=$1
+	# Craft the URL to the GitHub API. The access token is optional for the monorepo since it's an open-source repo.
+	GITHUB_API_URL=$(echo "https://api.github.com/repos/${CIRCLE_PULL_REQUEST}?access_token=$GITHUB_ACCESS_TOKEN" | sed "s/\/pull\//\/pulls\//")
+	# Grab the PR's base ref using the GitHub API.
+	REF=$(curl -s "$GITHUB_API_URL" | jq -r ".base.ref")
+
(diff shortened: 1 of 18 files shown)


ops/docker/ci-builder/check-changed.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -e
+
+echoerr() { echo "$@" 1>&2; }
+
+# Check if this is a CircleCI PR.
+if [[ -n $CIRCLE_PULL_REQUEST ]]; then
+	PACKAGE=$1
+	# Craft the URL to the GitHub API. The access token is optional for the monorepo since it's an open-source repo.
+	GITHUB_API_URL=$(echo "https://api.github.com/repos/${CIRCLE_PULL_REQUEST:19}?access_token=$GITHUB_ACCESS_TOKEN" | sed "s/\/pull\//\/pulls\//")
```
tynes, line 17: Why is the default value here set to `19`? It should always be set if we make it to this point

roninjin10 (approved):

ops/docker/ci-builder/check-changed.sh
mslipper: Fixed

ops/docker/ci-builder/check-changed.sh
```diff
@@ -0,0 +1,32 @@
...
+ 	echoerr "Curr Ref:     $(git rev-parse --short HEAD)"
+
+ 	# Compare HEAD to the PR's base ref, stripping out the change percentages that come with git diff --dirstat.
+ 	# Pass in the diff pattern to grep, and echo TRUE if there's a match. False otherwise.
+ 	(git diff --dirstat=files,0 "$REF...HEAD" | sed 's/^[ 0-9.]\+% //g' | grep -q -E "$PACKAGE" && echo "TRUE") || echo "FALSE"
+else
+	# Non-PR builds always require a rebuild.
+	echoerr "Not a PR build, requiring a total rebuild."
+	echo "TRUE"
+fi
```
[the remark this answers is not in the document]
mslipper: Fixed

PR #2855, opened 2022-06-27
spec: derivation spec rework

norswap:
- add overview
- greatly expand & correct glossary
- add section on batch submission, and consolidate encoding specs
- described every stage in the pipeline + its general function
- edited & added a lot of explanation regarding resetting the pipeline

July 8th: everything is integrated, but a few things still need work
- I didn't incorporate feedback from Kelvin and from our call with Proto & Josh (all of which are present as review comments) yet
- the payload derivation section is a klutz that I need to fix
- I didn't re-read through the later section, past what we reviewed w/ Proto & Josh (i.e. overview + wire format)

Still, this should be good enough for reviews.


specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+### Frame Format
+
+A [channel frame][g-channel-frame] is encoded as:
 
 ```text
-transaction_data = version_byte | rollup_payload
+frame = channel_id ++ frame_number ++ frame_data_length ++ frame_data ++ is_last
+
+channel_id        = random ++ timestamp
+random            = bytes32
```
smartcontracts, line 220: Why `random`? Why not use an incrementing channel ID?

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+[batcher-spec]: batching.md
+
+### Channel Format
+
+A channel is encoded as `channel_encoding`, defined as:
+
+```text
+rlp_batches = []
+for batch in batches:
+    rlp_batches.append(rlp_encode(batch))
```
smartcontracts, line 260: Why RLP?

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+Unknown versions make the batch invalid (it must be ignored by the rollup node), as do malformed contents.
+
+------------------------------------------------------------------------------------------------------------------------
+
+# Architecture
+
+[architecture]: #architecture
+
+The above describes the general process of L2 chain derivation, and specifies how batches are encoded within [batcher
+transactions][g-batcher-transactions].
```
norswap, line 319: dead link, drop the "s"

specs/derivation.md
```diff
@@ -367,7 +808,7 @@ A sequencing window is derived into a variable number of L2 blocks, defined by a
     `batch.timestamp` refers to the L2 block timestamp encoded in the batch.
   - `next_l1_timestamp` is the timestamp of the next L1 block.
 
-The L2 chain is extended to `new_head_l2_timestamp` with blocks at a fixed block time (`l2_block_time`).
+The L2 chain is extended to `new_head_l2_timestamp` with blocks at a fixed [block time][g-block-time] (`l2_block_time`).
 This means that every `l2_block_time` that has no batch is interpreted as one with no sequenced transactions.
```
norswap, line 812: link to block time glossary entry

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
-- There may be different actors that submit the data, the system cannot rely on a single EOA nonce value.
-- The fraud proof processes submitted data, which has to be bounded
-- The submissions confirmed on L1 may be confirmed out of order
-- The L1 may [reorganize][g-reorg], requiring the submissions to reset to a previous point
-- The L2 grows, and does so in smaller intervals than L1
-- The L2 may have bursts that cannot be fully confirmed in a single piece of submitted data
+This means that even if the sequencer applies a state transition incorrectly, the transactions in the batch will stil be
+considered part of the canonical L2 chain. Batches are still subject to validity checks (i.e. they have to be encoded
+correctly), and so are individual transactions within the batch (e.g. signatures have to be valid). Invalid batches and
+invalid individual transactions within an otherwise valid batch are discarded by correct nodes.
```
norswap, line 129: TODO: check whether the whole gets dropped if we have a bad transaction
I reckon we should only drop the invalid transaction.

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
-The types of data are illustrated in the diagram below.
+[batcher-spec]: batching.md
+
+> **TODO** rewrite the batch submission specification
+>
+> Here are some things that should be included there:
+>
+> - There may be different concurrent data submissions to L1
+> - There may be different actors that submit the data, the system cannot rely on a single EOA nonce value.
+> - The batcher requests batches from the rollup-node then builds channels and frames.
```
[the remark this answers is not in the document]
norswap: in the future we might be able to get the safe head information from the execution engine directly
not possible it right now, but there is an upstream geth PR open

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+frame = channel_id ++ frame_number ++ frame_data_length ++ frame_data ++ is_last
+
+channel_id        = random ++ timestamp
+random            = bytes32
+timestamp         = uvarint
+frame_number      = uvarint
+frame_data_length = uvarint
+frame_data        = bytes
+is_last           = bool
+```
```
norswap, line 226: maybe replace uvarint by fixed-size integers
@trianglesphere will do it

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+- `frame_number` identifies the index of the frame within the channel (in bytes)
+- `frame_data_length` is the length of `frame_data` in bytes
+- `frame_data` is a sequence of bytes belonging to the channel, at index `max(frame_number-1, 0)`
+- `is_last` is a single byte with a value of 1 if the frame is the last in the channel, 0 if there are frames in the
+  channel. Any other value makes the frame invalid (it must be ignored by the rollup node).
+
+> **TODO** why is frame_number not 0 based?
+
+> **TODO** channel timestamps must be validated to not be in the future - is that currently done? + specify
+> Also: is this even safe? i.e. could L1 block producers manipulate the timestamp?
```
norswap, line 248: - it is done — document this constraint above
- L1 has a soft (client) constraint that timestamp cannot be higher than the wallclock time by a certain margin

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+- `uvarint` is a variable-length encoding of a 64-bit unsigned integer into between 1 and 9 bytes, [as specified in
+  SQLite 4][sqlite-uvarint].
+- `channel_id` uniquely identifies a channel as the concatenation of a random value and a timestamp
+    - `random` is a random value such that two channels with different batches should have a different random value
+    - `timestamp` is the time at which the channel was created (UNIX time in seconds)
+    - The ID includes both the hash and the timestamp, in order to prevent a malicious sequencer from reusing the random
+      value after the channel has [timed out][g-channel-timeout] (refer to the [batcher specification][batcher-spec] to
+      learn more about channel timeouts). This will also allow us substitute `random` by a hash commitment to the
+      batches, should we want to do so in the future.
+- `frame_number` identifies the index of the frame within the channel (in bytes)
```
norswap, line 239: incorrect, it's an index not an offset
similarly `max(frame_number-1, 0)` is incorrect — should talk about it being after he previous frames

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+- `frame_data_length` is the length of `frame_data` in bytes
+- `frame_data` is a sequence of bytes belonging to the channel, at index `max(frame_number-1, 0)`
+- `is_last` is a single byte with a value of 1 if the frame is the last in the channel, 0 if there are frames in the
+  channel. Any other value makes the frame invalid (it must be ignored by the rollup node).
+
+> **TODO** why is frame_number not 0 based?
```
norswap, line 245: it *is* 0-based

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+### Channel Format
+
+A channel is encoded as `channel_encoding`, defined as:
+
+```text
+rlp_batches = []
+for batch in batches:
+    rlp_batches.append(rlp_encode(batch))
+channel_encoding = compress(rlp_batches)
+```
```
norswap, line 262: The above is not correct: `batch` has a version byte which is not rlp-encoded, only the payload is.

specs/derivation.md
norswap: - variable-length size encoding
- my bad for not making the spec good, but `batch` is actually a structure that can itself be RLP-encoded (I just said they are bytes are, that map to something RLP-encoded, but in practice you would just use one RLP-encoding)

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+
+where:
+
+- `batches` is the input, a sequence of batches byte-encoded as per the next section ("Batch Encoding")
+- `rlp_encode` is a function that encodes a batch according to the [RLP format] (where `batch` is treated as a string of
+  bytes) — concretely this adds a (variable-length) length prefix to the batch
+- `rlp_batches` is the concatenation of the RLP-encoded batches
+- `compress` is a function performing compression, using the ZLIB algorithm with no dictionary
+- `channel_encoding` is the compressed version of `rlp_batches`
```
norswap, line 271: Fix this so that we only have one instance of RLP-encoding

norswap: code comment that defines the encoding, but is outdated: https://github.com/ethereum-optimism/optimism/blob/297af0837bc49ec2abdef5b59b695995d09e6ba4/op-node/rollup/derive/batch.go#L22

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+
+> **TODO** we probably need to specify the compression better (e.g. which ZLIB version, which flags)
```
norswap, line 273: look this up

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+
+### Batch Format
+
+Recall that a batch contains a list of transactions to be included in a specific L2 block.
+
+A batch is encoded as `batch_version ++ content`, where `content` depends on the version:
+
+| `batch_version` | `content`                                         |
+| --------------- | ------------------------------------------------- |
+| 0               | `rlp_encode([epoch, timestamp, transaction_list)` |
```
norswap, line 298: missing a closing `]`

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+## L2 Chain Derivation Pipeline
+
+[pipeline]: #l2-chain-derivation-pipeline
+
+Our architecture decomposes the derivation process into a pipeline made up of the following stages:
+
+1. L1 Traversal
+2. L1 Retrieval
+3. Channel Buffering
+4. Batch Decoding
```
norswap, line 337: there is a stage missing here

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+
+> **NOTE**
+>
+> The currently implementation uses slightly different names for each stage, respectively:
+>
+> 1. L1 Traversal → `L1Traversal`
+> 2. L1 Retrieval → `L1Retrieval`
+> 3. Channel Buffering → `ChannelBank`
+> 4. Batch Decoding → `ChannelInReader`
+> 5. Payload Attributes Derivation → `BatchQueue`
```
norswap, line 411: split into batch queue and attributes queue

norswap:
Everything is integrated, but a few things still need work:
- I didn't incorporate feedback from Kelvin and from our call with Proto & Josh (all of which are present as review comments) yet
- the payload derivation section is a klutz that I need to fix
- I didn't re-read through the later section, past what we reviewed w/ Proto & Josh (i.e. overview + wire format)

Still, this should be good enough for reviews.

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+[pipeline]: #l2-chain-derivation-pipeline
+
+Our architecture decomposes the derivation process into a pipeline made up of the following stages:
+
+1. L1 Traversal
+2. L1 Retrieval
+3. Channel Bank
+4. Batch Decoding (called `ChannelInReader` in the code)
+5. Payload Attributes Derivation (called `BatchQueue` in the code)
+6. Engine Queue
```
norswap, line 350: now there batch queue then payload attributes derivation

norswap: - batch decoding: decompresses the batches, and send the unordered to the next stage
- batch queue: re-orders the batches and fills in missing batches
     - if contiguous batches available at the start, flush to next stage (eager batch derivation, no need to wait for a full sequencing window)

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+
+> **TODO** can we change code names for these two things? maybe as part of a refactor
+
+The data flows flows from the start (outer) of the pipeline towards the end (inner). Each stage is able to push data to
+the next step.
```
norswap, line 355: s/step/stage

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+
+However, data is *processed* is reverse order. Meaning that if there is any data to be processed in the last stage, it
```
norswap, line 357: in reverse order

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+will be processed first. Processing proceeds in "steps" that can be taken at each stage. We try to take as many steps as
+possible in the last (most inner) stage before taking steps any step in its outer stage, etc.
```
norswap, line 359: s/step/steps

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+
+This ensures that we use the data we already have before pulling more data and minimizes the latency of data traversing
+the derivation pipeline.
+
+Each stage can maintain its own inner state as necessary. In particular, each stage maintains a L1 block reference
+(number + hash) to the latest L1 block such that all data originating from previous blocks has been processed, and the
+data from that block is or has been processed.
```
norswap, line 366: maybe bold?

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+L1 blocks as they get created, though we can also read old blocks while syncing, or in case of an L1 [re-org][g-reorg].
+
+### L1 Retrieval
+
+In the *L1 Retrieval* stage, we read the block we get from the outer stage (L1 traversal), and extract data for it. In
+particular we extract a byte string that corresponds to the concatenation of the data in all the [batcher
+transaction][g-batcher-transaction] belonging to the block. This byte stream encodes a stream of [channel
+frames][g-channel-frame] (see the [Batch Submission Wire Format][wire-format] section for more info).
+
+This frames are parsed, then grouped them per [channel][g-channel] into a structure we call the *channel bank*.
```
norswap, line 382: remove "them"

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+
+Some frames are ignored:
+
+- Frames where `frame.frame_number <= progress`, where `progress` is the number of the latest frame that was read.
+    - i.e. in case of duplicate frame, the first frame read from L1 is considered canonical.
```
norswap, line 387: remove "latest frame" as "highest number"

norswap: change `progress` to a different name, because `progress` is used for L1 reference in code

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+- Frames past the end of the channel (i.e. past the first frame marked with `frame.is_last == 1`) are ignored.
```
norswap, line 388: "past end of the channel" → in terms of number, not in terms of order in the stream

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+stage, and decompressing batches from these frames.
+
+In principle, we should be able to read any channel that has any number of sequential frames at the "front" of the
+channel (i.e. right after any frames that have been read from the bank already) and decompress batchers from them. (Note
+that if we did this, we'd need to keep partially decompressed batches around.)
+
+However, our current implementation doesn't support streaming decompression, so currently we have to wait until either:
+
+- We have received all frames in the channel (i.e. we received the last frame in the channel (`is_last == 1`) and every
+  frame before it)
```
norswap, line 404: "before it" == lower numbers

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+- The channel has timed out (in which we case we read all contiguous sequential frames from the start of the channel).
+    - A channel is considered to be *timed out* if `currentL1Block > channeld_id.timestamp + CHANNEL_TIMEOUT`.
```
norswap, line 406: s/`currentL1Block`/`currentL1Block.timestamp`

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+        - where `currentL1Block` is the L1 block maintained by this stage, which is the latest L1 block whose frames
+          have been added to the channel bank.
+
+> **TODO** There is currently `MAX_CHANNEL_BANK_SIZE`, a notion about the maximum amount of channels we can keep track
+> of.
+> - Is this a semantic detail (i.e. if the batcher opens too many frames, valid channels can be dropped?)
+> - If so, I feel **very strongly** about changing this. This ties us very much to the current implementation.
+> - And it doesn't feel necessary given the channel timeout - if DOS is an issue we can reduce the channel timeout.
```
norswap, line 414: - Note: the size is a BYTE size.
     - And it doesn't matter if we have many channels or few channels with many frames.
- We need this to avoid DOS from malicious batchers.
- Drop full channels at a time.
- Open question: drop old channels or drop new channels.
     - Only relevant in a malicious batcher case, malicious batcher can do damage in both cases.

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+
+### Batch Decoding
+
+In the *Batch Decoding* stage, we simply decode [batches][g-sequencer-batch] from the frames we received from the last
+stage.
```
norswap, line 419: decoding = decompression + parsing of the decompressed data

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+
```
norswap, line 420: TODO: insert batch queue stage, see comment above for what it does

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+### Payload Attributes Derivation
+
+In the *Payload Attributes Derivation* stage, we convert the decoded batches into instances of the
+[`PayloadAttributes`][g-payload-attr] structure. Such a structure encodes the transactions that need to figure into a
+block, as well as other block inputs (timestamp, fee recipient, etc). Payload attributes derivation is detailed in the
+section [Deriving Payload Attributes section][deriving-payload-attr] below.
+
+If batches are missing for a given [time slot][g-time-slot], this stage generates empty batches to fill gaps.
```
norswap, line 428: this is done in the previous section

specs/derivation.md
norswap: Consider renaming "batch queue" as "batch buffer".

specs/derivation.md
```diff
@@ -0,0 +1,820 @@
...
+In particular, the following fields of the payload attributes are checked for equality with the block:
+
+- `parent_hash`
+- `timestamp`
+- `randao`
+- `transactions_list` (first length, then equality of each of the encoded transactions)
+
+TODO link payload attributes spec / definition
+
+If consolidation fails, the unsafe L2 head is reset to the safe L2 head.
```
norswap, line 458: and then execution proceeds to create a new safe L2 head on top of the old safe L2 head

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+  the sequencing window size (note that this means that epochs are overlapping). In particular we need:
+    - The [batcher transactions][g-batcher-transactions] included in the sequencing window. These allow us to
+      reconstruct [sequencer batches][g-sequencer-batch] containing the transactions to include in L2 blocks (each batch
+      maps to a single L2 block).
+    - The [deposits][g-deposits] made in L1 block `E` (in the form of events emitted by the [deposit
+      contract][g-deposit-contract]).
+    - The L1 block attributes from L1 block `E` (to derive the [L1 attributes deposited transaction][g-l1-attr-deposit]).
+- The state of the L2 chain after the last L2 block of epoch `E - 1`, or — if epoch `E - 1` does not exist — the
+  [genesis state][g-l2-genesis] (cf. TODO) of the L2 chain.
+    - An epoch `E` does not exist if `L2CI <= E`, where `L2CI` is the [L2 chain inception][g-l2-chain-inception].
```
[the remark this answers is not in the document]
norswap: It should!

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+[Architecture section][architecture] for more information on how we implement this in practice.
+
+Each epoch may contain a variable number of L2 blocks, at the discretion of [the sequencer][g-sequencer], but subject to
+the following constraints for each block:
+
+- `min_l2_timestamp <= block.timestamp < max_l2_timestamp`, where
+    - all these values are denominated in seconds
+    - `min_l2_timestamp = prev_l2_timestamp + l2_block_time`
+        - `prev_l2_timestamp` is the timestamp of the previous L2 block
+        - `l2_block_time` is a configurable parameter of the time between L2 blocks (on Optimism, 2s)
```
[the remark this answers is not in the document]
norswap: Agree it would be nice to have, but probably not going to happen right now.

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+    - `max_l2_timestamp = max(l1_timestamp + max_sequencer_drift, min_l2_timestamp + l2_block_time)`
```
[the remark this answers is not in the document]
norswap: No, rather that the number of blocks within an L2 epoch can float. At minimum it must comprises one L2 block, while the max is capped by the sequencer drift.

I'm going to clarify the preceding paragraph to say that l2 blocks occur every 2s.

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+## Eager Block Derivation
 
-The L1 data can be thought of as a stream, and L2 rollup nodes are the ***readers***, transforming their state with the inputs.
+In practice, it is often not necesary to wait for a full sequencing window of L1 blocks in order to start deriving the
+L2 blocks in an epoch. Indeed, as long as we are able to reconstruct sequential batches, we can start deriving the
+corresponding L2 blocks. We call this *eager block derivation*.
 
-The L2 batcher nodes are the ***writers*** to the stream: they take L2 changes and submit the necessary data to L1.
-Others can then reproduce the same L2 changes.
+However, in the very worst case, we can only reconstruct the batch for the first L2 block in the epoch by reading the
```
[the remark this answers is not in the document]
norswap: Correct, adding this in.

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+last L1 block of the sequencing window. This means we also can't derive any further L2 block in the epoch until then, as
```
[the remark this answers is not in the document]
norswap: Correct, adding this.

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+## Sequencing & Batch Submission Overview
 
-## Rollup Stream format
+The [sequencer][g-sequencer] accepts L2 transactions from users. It is responsible for building blocks out of these. For
+each such block, it also creates a corresponding [sequencer batch][g-sequencer-batch]. It is also responsible for
+submitting each batch to a [data availability provider][g-avail-provider] (e.g. Ethereum calldata), which it does via
+its [batcher][g-batcher] component.
 
-The stream is ***multiplexed*** and ***buffered***, to handle serve the required properties:
+The difference between an L2 block and a batch is subtle but important: the block includes an L2 state root, whereas the
```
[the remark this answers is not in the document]
norswap: True, included with the following explanation of why this matters:
> This matters in some edge case where a L1 reorg would occur and a batch would be reposted to the L1 chain but not
the preceding batch, whereas the predecessor of an L2 block cannot possibly change.

specs/derivation.md
norswap: Only invalid transactions get dropped.

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
-- The L1 may [reorganize][g-reorg], requiring the submissions to reset to a previous point
-- The L2 grows, and does so in smaller intervals than L1
-- The L2 may have bursts that cannot be fully confirmed in a single piece of submitted data
+This means that even if the sequencer applies a state transition incorrectly, the transactions in the batch will stil be
+considered part of the canonical L2 chain. Batches are still subject to validity checks (i.e. they have to be encoded
+correctly), and so are individual transactions within the batch (e.g. signatures have to be valid). Invalid batches and
+invalid individual transactions within an otherwise valid batch are discarded by correct nodes.
 
-The multiplexed and buffered stream consists of:
+If the sequencer applies a state transition incorrectly, he will post an incorrect [output root][g-l2-output], which
```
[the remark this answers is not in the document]
norswap: Sorts of clarify what happens if a batch with an invalid transaction is posted, where the sequencer interprets the transaction as valid, but validators do not. Pretty harmless extra detail.

Will change text to "if the sequence applies [...] and posts a state root, this state root will be incorrect [...].

specs/derivation.md
```diff
@@ -27,181 +43,606 @@ TODO
...
+The [batcher][g-batcher] submits [batcher transactions][g-batcher-transaction] to a [data availability
+provider][g-avail-provider]. These transactions contain one or multiple [channel frames][g-channel-frame], which are
+chunks of data belonging to a [channel][g-channel].
+
+A [channel][g-channel] is a sequence of [sequencer batches][g-sequencer-batch] (for sequential blocks) compressed
+together. The reason to group multiple batches together is simply to obtain a better compression rate, hence reducing
+data availability costs.
+
+Channels might be too large to fit in a single [batcher transaction][g-batcher-transaction], hence we need to split it
+into chunks known as [channel frames][g-channel-frame]. A single batcher transaction can also carry multiple frames
```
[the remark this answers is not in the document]
norswap: The way I see it, allowing this gives us maximum flexibility, and ease of implementation (since we do not need to check the constraint of one frame per channel).

I can think of cases where this could be desirable too, e.g. if we limit frame size but data availability space is pretty big, or just let the batcher schedule things according to its own logic (say he hasn't seen new things in a while, he does a frame, but then suddenly new things come in, and can just put those in another frame).

specs/derivation.md
norswap: Good question. I think we want to avoid relying on the number in the implem so that we're free to swap out the random value by a hash commitment in the future. @protolambda ?

specs/derivation.md
norswap: I've document the drop requirement as:

> Channels whose timestamp are higher than that of the L1 block they first appear in must be ignored. Note that L1 nodes have a soft constraint to ignore blocks whose timestamps that are ahead of the wallclock time by a certain margin. (A soft constraint is not a consensus rule — nodes will accept such blocks in the canonical chain but will not attempt to build directly on them.)

Is this correct and consistent with the current implementation? Dropping just the first frame should have the same result (as it causes the channel to be unreadable).

It does make a difference to the size of the channel bank however (!!) so must be specific here.

@protolambda @trianglesphere

specs/derivation.md
norswap: Okay so actually, `batch` is NOT RLP encoded, only its content, but it has a non-RLP encoded version byte.

I'm thinking this is meh and we should fix, it. Resolving this comment however, consolidating these concerns in another comment below.

norswap: Okay so actually, only the content of the batch is encoded, the encoding can be dropped here.

specs/derivation.md
protolambda: A nonce-like number would be stateful and thus much harder to recover (with current design there's no need to read L1 to make a valid batch-submission)

specs/derivation.md
protolambda: Not implemented yet, but there's a Linear issue open for it. Not high-priority.

specs/rollup-node.md
```diff
@@ -8,349 +8,45 @@
...
-      - [Building individual Payload Attributes](#building-individual-payload-attributes)
-  - [From Payload Attributes to L2 Block](#from-payload-attributes-to-l2-block)
-    - [Inductive Derivation Step](#inductive-derivation-step)
-    - [Engine API Error Handling](#engine-api-error-handling)
-    - [Finalization Guarantees](#finalization-guarantees)
-  - [Whole L2 Chain Derivation](#whole-l2-chain-derivation)
-  - [L2 Output RPC method](#l2-output-rpc-method)
-    - [Output Method API](#output-method-api)
-- [Handling L1 Re-Orgs](#handling-l1-re-orgs)
+TODO
```
protolambda, line 22: TODO: Update ToC before merging

protolambda (approved):

trianglesphere (approved):

protolambda:
@mergifyio update

protolambda:
@mergifyio refresh

PR #2860, opened 2022-06-27
Full ToC in specs README

maurelian:
**Description**

All I'm trying to do here is ensure that the ToC is complete. I'm sure there are other ways that I could improve on the presentation, but I'd prefer not to expand the scope of this PR.


tynes:
This isn't caught by the markdown linter?

tynes (approved):

maurelian:
It's not generated by doctoc like other ToCs. I suppose maybe that's possible, but have only seen it used for a single page ToC that lists the internal headings.

PR #2863, opened 2022-06-27
op-node: Add Epoch Hash to batch

trianglesphere:
**Description**
This commits a batch to a specific L1 origin block by hash rather
than just by number. This help in the case of L1 reorgs by stopping
batches from being applied in weird ways.

**Additional context**
This does not add a block number to the batch. The plan for eager batch derivation
is to continue to use timestamps, but slightly modify derivation rules (similar to how
they would be modified even if it was switched over to using block numbers).

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/op-node/rollup/derive/batches.go
+++ b/op-node/rollup/derive/batches.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/ethereum-optimism/optimism/op-node/eth"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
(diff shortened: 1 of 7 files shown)


op-node/rollup/derive/batches.go
```diff
@@ -34,12 +35,15 @@ func FilterBatches(log log.Logger, config *rollup.Config, epoch rollup.Epoch, mi
 	return
 }
 
-func ValidBatch(batch *BatchData, config *rollup.Config, epoch rollup.Epoch, minL2Time uint64, maxL2Time uint64) error {
+func ValidBatch(batch *BatchData, config *rollup.Config, epoch rollup.Epoch, epochHash common.Hash, minL2Time uint64, maxL2Time uint64) error {
```
protolambda, line 38: Maybe we should just pass `epoch eth.BlockID`?

op-node/rollup/derive/batches.go
```diff
@@ -5,16 +5,17 @@ import (
...
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
 )
 
 var DifferentEpoch = errors.New("batch is of different epoch")
 
-func FilterBatches(log log.Logger, config *rollup.Config, epoch rollup.Epoch, minL2Time uint64, maxL2Time uint64, batches []*BatchData) (out []*BatchData) {
+func FilterBatches(log log.Logger, config *rollup.Config, epoch rollup.Epoch, epochHash common.Hash, minL2Time uint64, maxL2Time uint64, batches []*BatchData) (out []*BatchData) {
```
protolambda, line 15: Same here

op-node/rollup/derive/batches.go
```diff
@@ -62,7 +66,7 @@ func ValidBatch(batch *BatchData, config *rollup.Config, epoch rollup.Epoch, min
 }
 
 // FillMissingBatches turns a collection of batches to the input batches for a series of blocks
-func FillMissingBatches(batches []*BatchData, epoch, blockTime, minL2Time, nextL1Time uint64) []*BatchData {
+func FillMissingBatches(batches []*BatchData, epochNumber uint64, epochHash common.Hash, blockTime, minL2Time, nextL1Time uint64) []*BatchData {
```
protolambda, line 69: Same here

op-node/rollup/derive/batches.go
```diff
@@ -34,12 +35,15 @@ func FilterBatches(log log.Logger, config *rollup.Config, epoch rollup.Epoch, mi
...
 }
 
-func ValidBatch(batch *BatchData, config *rollup.Config, epoch rollup.Epoch, minL2Time uint64, maxL2Time uint64) error {
+func ValidBatch(batch *BatchData, config *rollup.Config, epoch rollup.Epoch, epochHash common.Hash, minL2Time uint64, maxL2Time uint64) error {
 	if batch.Epoch != epoch {
 		// Batch was tagged for past or future epoch,
 		// i.e. it was included too late or depends on the given L1 block to be processed first.
 		return DifferentEpoch
 	}
+	if batch.OriginHash != epochHash {
```
protolambda, line 44: Can merge this with above check, and return a wrapped error around `DifferentEpoch` with the two different id values (batch vs epoch).

protolambda (commented):
Generally looks good, maybe consider using `eth.BlockID` (and we'd get rid of the `rollup.Epoch` type then I guess).

trianglesphere:
switched to `eth.BlockID` in calling code

op-node/rollup/derive/batches.go
trianglesphere: check merged, but no differentiation in the error case

op-node/rollup/derive/channel_out.go
```diff
@@ -164,7 +164,7 @@ func blockToBatch(block *types.Block, w io.Writer) error {
 	}
 
 	batch := &BatchData{BatchV1{
-		Epoch:        rollup.Epoch(l1Info.Number),
+		EpochNum:     rollup.Epoch(l1Info.Number),
```
protolambda, line 167: think you missed the epoch hash here, I'll add it

PR #2865, opened 2022-06-27
batcher: Handle multiple frames per channel

trianglesphere:
**Description**
The batcher is still very simple generating a new channel full of
L2 blocks since the last channel that it created, but it is just
a tad smarter now in that if will handle the case of multiple frames
per channel.

This is the bare minimum functionality to handle happy case batching
on a real network. The only other thing that it can't handle is
reorgs, but it can now handle very larger L2 transactions and blocks.
**Additional context**

**Metadata**
- Fixes #[Link to Issue]


op-batcher/batch_submitter.go
```diff
@@ -253,58 +254,65 @@ mainLoop:
...
-			}
+				// Wait until one of our submitted transactions confirms. If no
+				// receipt is received it's likely our gas price was too low.
+				// TODO: does the tx manager nicely replace the tx?
+				//  (submit a new one, that's within the channel timeout, but higher fee than previously submitted tx? Or use a cheap cancel tx?)
+				ctx, cancel = context.WithTimeout(l.ctx, time.Second*time.Duration(l.cfg.ChannelTimeout))
+				receipt, err := l.txMgr.Send(ctx, updateGasPrice, l.cfg.L1Client.SendTransaction)
+				cancel()
+				if err != nil {
+					l.log.Error("unable to publish tx", "err", err)
```
protolambda, line 303: Once one tx goes missing, there's a missing frame, and it's not useful anymore to try and submit the next frames. It should continue to the `mainLoop` here.

op-batcher/batch_submitter.go
```diff
@@ -253,58 +254,65 @@ mainLoop:
...
-			// Actually send the transaction
-			//
-			walletAddr := crypto.PubkeyToAddress(l.cfg.PrivKey.PublicKey)
+				// Create the transaction
+				ctx, cancel = context.WithTimeout(l.ctx, time.Second*10)
+				tx, err := l.CraftTx(ctx, data.Bytes(), nonce)
+				cancel()
+				if err != nil {
+					l.log.Error("unable to craft tx", "err", err)
+					continue
```
protolambda, line 286: Same here, once there's a gap we'd need to reconstruct the frame that is going missing or continue with the `mainLoop`

op-batcher/batch_submitter.go
```diff
@@ -253,58 +254,65 @@ mainLoop:
...
 
-			// Poll for new L1 Block and make a decision about what to do with the data
+				// Query for the submitter's current nonce.
+				walletAddr := crypto.PubkeyToAddress(l.cfg.PrivKey.PublicKey)
+				ctx, cancel = context.WithTimeout(l.ctx, time.Second*10)
+				nonce, err := l.cfg.L1Client.NonceAt(ctx, walletAddr, nil)
+				cancel()
+				if err != nil {
+					l.log.Error("unable to get current nonce", "err", err)
+					continue
```
protolambda, line 277: Same here. Maybe get the nonce before constructing the frame, so it can be a regular `continue` and we can retry getting the nonce.

op-batcher/batch_submitter.go
trianglesphere: switch to `continue mainLoop`

op-batcher/batch_submitter.go
trianglesphere: switch to `continue mainLoop`

PR #2714, opened 2022-06-08
bedrock batch derivation v2

protolambda:
Work in progress, see Batch derivation v2 doc.

PR #2871, opened 2022-06-28
Bedrock: new batch-derivation changes

protolambda:
This PR tracks all batch-derivation changes, starting a squashed version of the earlier dev branch `batch-deriv-v2` (#2714).

Op-node and op-e2e tests are passing, except an unrelated e2e fees test which is being looked at separately.

We'll split out some of the `eth`/`l2`/`testutils` changes into separate PRs to reduce the size of this one. And then likely review in smaller PRs that isolate changes more.


op-batcher/db/history_db.go
```diff
@@ -4,67 +4,42 @@ import (
 	"encoding/json"
```
trianglesphere, line 4: delete the history files? IIRC they aren't used anymore

trianglesphere (commented):
I left a couple comments on packages other than `op-node`
I wrote most of `op-batcher` so I don't want to do most of the review
`op-e2e` lgtm
`op-proposer` lgtm
`ops-bedrock` has a slight inconsistency

Will help with splitting out refactor code from `op-node` & then start reviewing it

op-node/rollup/derive/channel_emitter_test.go
```diff
@@ -0,0 +1,114 @@
+package derive
+
+// import (
+// 	"context"
```
trianglesphere, line 4: dead file, should remove

op-node/rollup/derive/channel_out.go
```diff
@@ -0,0 +1,178 @@
...
+	return blockToBatch(block, co.compress)
+}
+
+func makeUVarint(x uint64) []byte {
+	var tmp [binary.MaxVarintLen64]byte
+	n := binary.PutUvarint(tmp[:], x)
+	return tmp[:n]
+}
+
+func (co *ChannelOut) ReadyBytes() int {
```
trianglesphere, line 90:
```suggestion
// ReadyBytes returns the number of bytes that the channel out can immediately output into a frame.
// Use `Flush` or `Close` to move data from the compression buffer into the ready buffer if more bytes
// are needed. Add blocks may add to the ready buffer, but it is not guaranteed due to the compression stage.
func (co *ChannelOut) ReadyBytes() int {
```

op-node/rollup/derive/channel_out.go
```diff
@@ -0,0 +1,178 @@
...
+	return co.buf.Len()
+}
+
+func (co *ChannelOut) Flush() error {
```
trianglesphere, line 94:
```suggestion
// Flush flushes the internal compression stage to the ready buffer. It enables pulling a larger & more
// complete frame. It reduces the compression efficiency.
func (co *ChannelOut) Flush() error {
```

op-batcher/db/history_db.go
protolambda: done

op-node/rollup/driver/state.go
```diff
@@ -408,87 +291,68 @@ func (s *state) loop() {
...
-	l1WindowBufJSON, _ := json.Marshal(s.l1WindowBuf)
-
-	s.snapshotLog.Info("Rollup State Snapshot",
-		"event", event,
-		"l1Head", string(l1HeadJSON),
-		"l2Head", string(l2HeadJSON),
-		"l2SafeHead", string(l2SafeHeadJSON),
-		"l2FinalizedHead", string(l2FinalizedHeadJSON),
-		"l1WindowBuf", string(l1WindowBufJSON))
+	// l1HeadJSON, _ := json.Marshal(s.l1Head)
```
trianglesphere, line 345: I commented this out & it needs to be uncommented

op-node/rollup/driver/state.go
```diff
@@ -408,87 +291,68 @@ func (s *state) loop() {
...
-			// behind the L1 head. We can therefore attempt to shift the safe head forward by at
-			// least one L1 block. If the node is holding on to unsafe blocks, this may trigger a
-			// reorg on L2 in the case that safe (published) data conflicts with local unsafe
-			// block data.
-			if s.l1Head.Number-s.l2SafeHead.L1Origin.Number >= s.Config.SeqWindowSize {
-				s.log.Trace("Requesting next step", "l1Head", s.l1Head, "l2Head", s.l2Head, "l1Origin", s.l2SafeHead.L1Origin)
-				reqStep()
-			}
-
+			s.handleNewL1Block(newL1Head)
```
trianglesphere, line 310: How does this handle reorgs?

protolambda: We don't handle reorgs unless we actually hit a L1 block on traversal. I.e. we currently don't immediately handle reorgs to shorter L1 chains. But the L2 won't grow one step further either without detecting the reorg (i.e. as soon as it can get L1 input N+1 it will see it doesn't build on N and then the pipeline will reset).

trianglesphere: ok, I've figured out how reorgs are being handled & it makes sense to me

specs/rollup-node-p2p.md
```diff
@@ -276,7 +276,7 @@ An [extended-validator] checks the incoming messages as follows, in order of ope
 
 - `[REJECT]` if the compression is not valid
 - `[REJECT]` if the block encoding is not valid
-- `[REJECT]` if the `payload.timestamp` is older than 20 seconds in the past
+- `[REJECT]` if the `payload.timestamp` is older than 60 seconds in the past
```
trianglesphere, line 279: did the value change in the code in this PR?

protolambda: If L1 block time is 15 seconds, and there's some response from the sequencer (e.g. a reorg or some other stalling), it's likely to produce L2 blocks with a delay. So I bumped it up to be more flexible.

PR #2873, opened 2022-06-28
l2geth: Fix NPE in API tracer

mslipper:
The `debug_standardTraceBlockToFile` RPC was panicing here:

```go
// Line 572 in eth/api_tracer.go
if config != nil && config.Overrides != nil {
```

`config` is an instance of `StdTraceConfig`, which embeds a pointer to `vm.LogConfig`. This pointer is `nil` when the user doesn't pass in any overrides. Since `vm.LogConfig` exposes the `Overrides` parameter, this caused a panic.

`debug_standardTraceBlockToFile` is required for users like Dune, who need access to transaction traces that are too complex to serve over RPC.

### DIFF
--- a/.changeset/stupid-ties-reply.md
+++ b/.changeset/stupid-ties-reply.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/l2geth': patch
+---
+
+fix NPE in debug_standardTraceBlockToFile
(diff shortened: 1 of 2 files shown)


tynes:
This looks generally good to me, haven't tested locally but looks simple enough

tynes (approved):

smartcontracts (approved):
Same fix was made upstream, so I'm happy with it: https://github.com/ethereum/go-ethereum/blob/9a5c1000c73c39e036973f4dfa828acc707e6d7c/eth/tracers/api.go#L678

PR #2878, opened 2022-06-28
ops-bedrock: Increase timeout on wait_up

maurelian:
Increases the timeout on waiting for a URL to respond in `devnet-up`. 

Also a couple minor changes related to running `devnet-up`.

### DIFF
--- a/packages/contracts-bedrock/.gitignore
+++ b/packages/contracts-bedrock/.gitignore
@@ -6,3 +6,4 @@ coverage.out
 .deps
 deployments
 broadcast
+genesis.json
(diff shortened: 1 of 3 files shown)


maurelian:
> This PR changes implementation code, but doesn't include a changeset. Did you forget to add one?

This PR doesn't change any packages that `yarn changeset` provides for me. I think mergify is too sensitive here.

packages/contracts-bedrock/.gitignore
```diff
@@ -6,3 +6,4 @@ coverage.out
 .deps
 deployments
```
tynes, line 7: i think we do want to commit in the deployments directory besides `deployments/devnetL1` but fine for now

tynes (approved):

PR #2880, opened 2022-06-28
fix(ctb): bug computing L2 timestamp

smartcontracts:
**Description**
Fixes a bug where the wrong constant was being used to compute L2
timestamps.


tynes:
The fact that changing this value and no tests break shows that we do not have sufficient coverage over this function

smartcontracts:
Agree

maurelian:
> The fact that changing this value and no tests break shows that we do not have sufficient coverage over this function

They did break, this PR just didn't run them.

maurelian:
> Why didn't they run?

Maybe because it was in draft at the time? Regardless, I was only able to see the two GHA checks having completed at the time of my comment.

This will fail tho:
```
git checkout 7898717
forge test
```

norswap (approved):

tynes:
Generally looks good to me, going to rebase + test on the local devnet before approval

tynes (changes requested):
Will approve once I observe devnet is working with this change

tynes (approved):

tynes:
Pushed a commit that fixes the devnet on top of this as well as rebased it

PR #2882, opened 2022-06-28
[WIP] Eager Batch Derivation

trianglesphere:
**Description**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


op-node/rollup/derive/batch_queue.go
```diff
@@ -4,205 +4,303 @@ import (
...
-	log       log.Logger
-	inputs    []BatchesWithOrigin
-	resetting bool // true if we are resetting the batch queue
-	config    *rollup.Config
-	dl        L1ReceiptsFetcher
-	next      BatchQueueOutput
-	progress  Progress
+	log                     log.Logger
+	originOpen              bool // true if the last origin expects more batches
+	highestL1InclusionBlock eth.L1BlockRef
```
protolambda, line 30: these two vars seem like duplicates with the `progress` field.

op-node/rollup/derive/payload_queue.go
```diff
@@ -0,0 +1,144 @@
...
+		return err
+	}
+	pq.next.AddSafeAttributes(attr)
+	return nil
+}
+
+func (pq *AttributesQueue) ResetStep(ctx context.Context, l1Fetcher L1Fetcher) error {
+	pq.batches = pq.batches[:0]
+	// TODO: progress thing here
+	pq.progress = pq.next.Progress()
```
protolambda, line 74: this TODO looks fine. It's already correctly adopting the progress of the inner stage upon reset.

op-node/rollup/derive/batch_queue.go
```diff
@@ -4,205 +4,301 @@ import (
...
 
-	minL2Time := uint64(lastL2Timestamp) + bq.config.BlockTime
-	maxL2Time := l1Origin.Time + bq.config.MaxSequencerDrift
-	if minL2Time+bq.config.BlockTime > maxL2Time {
-		maxL2Time = minL2Time + bq.config.BlockTime
+	// TODO (VERY IMPORTANT): Filter this batch out if the origin of the batch is too far past the epoch
+	// TODO (ALSO VERY IMPT): Get this equality check correct
+	// TODO: Also do this check when ingesting batches.
+	if uint64(batch.Batch.EpochNum) >= batch.L1InclusionBlock.Number+cfg.SeqWindowSize {
+		return false
```
protolambda, line 160: The batch epoch num shouldn't be equal or higher than the inclusion block to begin with, since it's referencing that block by hash.
Do you mean:
```
epochNum := uint64(batch.Batch.EpochNum)
// note: overflow safe, since epochNum has been validated above to match current or next epoch. So it does not have an absurdly high value.
if epochNum+cfg.SeqWindowSize < batch.L1InclusionBlock.Number {
  return false
}
```

op-node/rollup/derive/engine_consolidate.go
```diff
@@ -25,6 +27,18 @@ func AttributesMatchBlock(attrs *eth.PayloadAttributes, parentHash common.Hash,
...
+				err := atx.UnmarshalBinary(expect)
+				fmt.Println(err)
+				btx.UnmarshalBinary(otx)
+				a, _ := L1InfoDepositTxData(atx.Data())
+				b, _ := L1InfoDepositTxData(btx.Data())
+				fmt.Println("unsafe")
+				spew.Dump(a)
+				fmt.Println("safe")
+				spew.Dump(b)
+			}
```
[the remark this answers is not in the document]
trianglesphere: was, debugged. Will pull these changes out

op-node/rollup/derive/batch_queue.go
trianglesphere: yea, figured this out. I like that comparison

op-node/rollup/derive/batch_queue_test.go
```diff
@@ -0,0 +1,223 @@
...
+		BlockTime:         2,
+		MaxSequencerDrift: 600,
+		SeqWindowSize:     2,
+	}
+	bq := NewBatchQueue(log, cfg, next)
+
+	l1 := L1Chain([]uint64{10, 15, 20})
+
+	// Open
+	bq.OpenOrigin(l1[0])
```
trianglesphere, line 142: API has changed so these tests need to be updated

protolambda: Will help update tests tomorrow morning :+1:

op-node/rollup/derive/payload_queue.go
```diff
@@ -0,0 +1,144 @@
...
+		pq.log.Info("advancing epoch in the payload queue", "l2SafeHead", l2SafeHead, "l2SafeHead_origin", l2SafeHead.L1Origin, "batch_timestamp", batch.batch.Timestamp, "batch_epoch", batch.batch.Epoch())
+		seqNumber = 0
+	}
+
+	fetchCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
+	defer cancel()
+	l1Info, _, receipts, err := pq.dl.Fetch(fetchCtx, batch.batch.EpochHash)
+	if err != nil {
+		pq.log.Error("failed to fetch L1 block info", "l1Origin", batch.batch.Epoch(), "err", err)
+		return nil, err
```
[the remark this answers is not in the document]
trianglesphere: why return `nil, nil`? though I guess this is out of date

op-node/testlog/testlog.go
```diff
@@ -140,9 +140,10 @@ func (l *logger) flush() {
 	l.t.Helper()
 	// 2 frame skip for flush() + public logger fn
 	decorationLen := estimateInfoLen(2)
-	padding := 20
-	if decorationLen <= 25 {
-		padding = 25 - decorationLen
+	padding := 0
+	targetWidth := 30
+	if decorationLen <= targetWidth {
+		padding = targetWidth - decorationLen
```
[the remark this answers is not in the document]
trianglesphere: yea, would be nice. This also fixes the issue where padding was adding an extra 20 chars if the line was too long

protolambda:
in the test: "adding batch while attributes queue is closed"

The panic in the attributes queue is wrong, the batch queue may produce empty batches when the origin data is closed & sequence window runs out. Maybe we replace it with a condition that checks if the batch origin is older than a sequence window compared to the local progress, to sanity check the filling is correct.

op-node/rollup/derive/batch_queue.go
```diff
@@ -4,205 +4,291 @@ import (
...
+	if batch.Batch.EpochNum != rollup.Epoch(prevEpoch) && batch.Batch.EpochNum != rollup.Epoch(prevEpoch+1) {
+		return false
+	}
+	// TODO: Also check EpochHash (hard b/c maybe extension)
 
-	minL2Time := uint64(lastL2Timestamp) + bq.config.BlockTime
-	maxL2Time := l1Origin.Time + bq.config.MaxSequencerDrift
-	if minL2Time+bq.config.BlockTime > maxL2Time {
-		maxL2Time = minL2Time + bq.config.BlockTime
+	if uint64(batch.Batch.EpochNum)+cfg.SeqWindowSize < batch.L1InclusionBlock.Number {
```
protolambda, line 153: nit: maybe worth a comment that there's no overflow risk due to above check.

op-node/rollup/derive/batch_queue.go
```diff
@@ -4,205 +4,291 @@ import (
...
-		l1SafeHead, err := l1Fetcher.L1BlockRefByHash(ctx, l2SafeHead.L1Origin.Hash)
+		// Timestamp bounds
+		minL2Time := l2SafeHead.Time + bq.config.BlockTime
+		maxL2Time := l1OriginTime + bq.config.MaxSequencerDrift
+		if minL2Time+bq.config.BlockTime > maxL2Time {
+			maxL2Time = minL2Time + bq.config.BlockTime
+		}
+
+		// Note: Don't check epoch change here, check it in `validExtension`
+		epoch, err := bq.dl.L1BlockRefByNumber(context.TODO(), uint64(batch.Batch.EpochNum))
```
protolambda, line 265: `ctx` should be passed in, and the function should return `*BatchWithL1InclusionBlock, error` to handle the error. Same for the `deriveBatches` func which calls this func. And the error should be handled as non-critical in the `Step` func, i.e. a failed RPC call should not result in an EOF or error, just a nil + subsequent retry.

op-node/rollup/derive/batch_queue_test.go
```diff
@@ -0,0 +1,323 @@
...
+func (f *fakeBatchQueueOutput) SafeL2Head() eth.L2BlockRef {
+	return f.safeL2Head
+}
+
+func (f *fakeBatchQueueOutput) Progress() Progress {
+	return f.progress
+}
+
+func b(timestamp uint64, epoch eth.L1BlockRef) *BatchData {
+	rng := rand.New(rand.NewSource(1234))
```
protolambda, line 46: If we're using the same rng seed for each batch, then they'll all look the same.

op-node/testlog/testlog.go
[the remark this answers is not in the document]
protolambda: Still like me to open this PR? I think we can drop these changes from this PR.

protolambda (changes requested):
Getting really close, nice work, but there are some things to resolve before we merge.

op-node/testlog/testlog.go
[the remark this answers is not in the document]
trianglesphere: yea, I'll drop these changes. At the very least pulling this patch in would be nice, but auto width would be nice as well.

op-node/rollup/derive/batch_queue_test.go
trianglesphere: length is different. Could do fixed length & rng seed based on timestamp

op-node/rollup/derive/batch_queue.go
trianglesphere: actually, I'm a little confused, is the attack that the `EpochNum` could be very large when it's submitted?

protolambda: Yes, I think the epoch num is filtered in several other places, but if it affects other batches it might still be an issue. Here we've a check close by that it's one of two non-overflowing values, so it's fine.

op-node/rollup/derive/progress.go
```diff
@@ -23,7 +23,7 @@ func (pr *Progress) Update(outer Progress) (changed bool, err error) {
 	}
 	if pr.Closed {
 		if outer.Closed {
-			if pr.Origin != outer.Origin {
+			if pr.Origin.ID() != outer.Origin.ID() {
```
protolambda, line 26: Fine to me to do the comparison like this, but is there even a functional difference, other than a possible (not sure) performance gain?

trianglesphere: I think I was running into issues with the tests by not having a full eth.L1BlockRef instead of the IDs

trianglesphere: TBH all you need is the hash for equality checks

protolambda (approved):
Changes look good to me. And thanks for `go mod tidy`-ing all, had to be done at some point, I hope other PRs don't get bloated with line count changes the same way.

mslipper (approved):

PR #2894, opened 2022-06-29
ops: use nightly foundry

tynes:
**Description**
Uses `foundry:nightly` for the build

### DIFF
--- a/ops/docker/Dockerfile.packages
+++ b/ops/docker/Dockerfile.packages
@@ -2,7 +2,7 @@
 # be used to build any of the follow-on services
 #
 # ### BASE: Install deps
-FROM ghcr.io/foundry-rs/foundry:latest as foundry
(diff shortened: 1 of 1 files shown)


mslipper (changes requested):
I'm concerned about using a nightly build in something that produces production code. Can we pick a version of the Foundry image, and retag it on our Docker Hub repo so that we can control changes?

maurelian (approved):

tynes:
> I'm concerned about using a nightly build in something that produces production code. Can we pick a version of the Foundry image, and retag it on our Docker Hub repo so that we can control changes?

Thats a good call, there is one change that we need that hasn't been merged into foundry yet, we can pin once that change gets in. The ethers-rs dep needs to be updated in foundry and then we need to make sure it fixes the problem we were seeing with the hh oom

tynes:
Will update this PR once the version of foundry is live that has the features we need

PR #2897, opened 2022-06-29
Bedrock: fix new batcher, use sync status api, start from safe head

protolambda:
Previously the batcher would continue batch submitting from the L2 head.
However, if the sequencer runs  while the batch submitter is offline, the L2 chain continues, and the batch-submitter will be skipping a part. This causes a reorg, after which the batch submitter can't recover without restart (since it didn't detect reorgs).

This PR fixes the above, to make restarting of sequencer and batcher work without so much disruption of the L2 chain.

### DIFF
--- a/op-batcher/sequencer/driver.go
+++ b/op-batcher/sequencer/driver.go
@@ -5,6 +5,8 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/ethereum-optimism/optimism/op-proposer/rollupclient"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/log"
@@ -19,6 +21,8 @@ type Config struct {
 
 	// API to hit for batch data
 	L2Client *ethclient.Client
+
+	RollupNode *rollupclient.RollupClient
 
 	// Limit the size of txs
 	MinL1TxSize uint64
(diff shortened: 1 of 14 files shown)


op-node/rollup/driver/state.go
```diff
@@ -336,12 +348,35 @@ func (s *state) eventLoop() {
 				s.l2Head = unsafe
 				reqStep() // continue with the next step if we can
 			}
+		case respCh := <-s.syncStatusReq:
+			respCh <- SyncStatus{
+				CurrentL1:   s.derivation.Progress().Origin,
```
trianglesphere, line 353: this is the tip of what we may consider while working on, realistically going to be the l1Head. Mainly a comment on the fact that "current L1" is not well defined (but then fine to define as is).

op-batcher/sequencer/driver.go
```diff
@@ -5,6 +5,8 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/ethereum-optimism/optimism/op-proposer/rollupclient"
```
trianglesphere, line 8: more weird goland import stuff.

trianglesphere (approved):
these changes look fine. Nice work marking the restarts way more efficient

op-node/rollup/driver/state.go
```diff
@@ -13,6 +13,14 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
+type SyncStatus struct {
```
tynes, line 16: Having a godoc here would be nice

protolambda:
Rebased onto `develop`, added `SyncStatus` go doc, and fixed import style.

PR #2898, opened 2022-06-29
contracts-bedrock: fix slither

tynes:
**Description**
Fixes slither to work in CI, this is a temp fix until we can configure the outdir for `build-info` using foundry, see https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/3883/workflows/d23c46f9-501a-4853-9305-d12ff1a12d0b/jobs/52916/steps

### DIFF
--- a/packages/contracts-bedrock/scripts/slither.sh
+++ b/packages/contracts-bedrock/scripts/slither.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+if [ ! -d forge-artifacts/build-info ]; then
+    npx hardhat compile
+fi
+
+cp -rf forge-artifacts/build-info artifacts/build-info
+slither .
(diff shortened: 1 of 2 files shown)


packages/contracts-bedrock/scripts/slither.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ ! -d forge-artifacts/build-info ]; then
+    npx hardhat compile
```
maurelian, line 4: I tested and this does seem to work, but for my own info, how does `npx hardhat compile` generate files in `forge-artifacts/build-info`?

maurelian (approved):

packages/contracts-bedrock/scripts/slither.sh
tynes: `foundry` outputs the build info directory in the artifacts directory || `build-info` and isn't configurable currently, i'd like to add in a way to configure the outdir and then we could just configure it to go in the same directory that hh outputs them

PR #2900, opened 2022-06-30
chore: Add yarn-deduplicate script to deduplicate along with github action

roninjin10:
This will help keep our node moduels smaller and avoid some issues yarn has from time to time. 

https://optimism-workspace.slack.com/archives/CQZ0U9FAN/p1656513572407569


roninjin10:
This technically changes any package that depends on the packages I changed but I didn't include a changeset.   not sure how to handle it since I'm not entirely sure which packages changed.

tynes:
Unsure if the actual deps used will be changed that go into the builds, we could add changesets for each typescript package just in case

smartcontracts:
Pretty sure this will change deps, so let's do a patch release of every TS package

roninjin10:
Ok I’ll bump packages.   I think I can add yarn deduplicate as a post install task too so nobody needs to remember to do it.

tynes (approved):

tynes:
@roninjin10 I pushed a commit that adds changesets, ready to merge with another approval

mslipper (approved):

smartcontracts (approved):

PR #2902, opened 2022-06-30
make: Add devnet-logs command

maurelian:
A handy dandy make command which enables easy access to the logs of a running devnet.


maurelian:
🙈

PR #2904, opened 2022-06-30
rewind functionality for channel-out

protolambda:
Builds on top of #2897 (although no special dependencies, just makes more sense to test this way).

This implements a `Rewind(frame)` function, so we can revert `ChannelOut` to a previous frame.

Later on this may be useful to improve the batcher recovery, but we also need it for a more urgent feature: we need to estimate the size of a frame, and we can't do so without creating the frame, which affects the compression state.

So the idea is that we can output a frame, check if it's large enough to be cost-efficient (or just too old that we'd have to submit it), and rewind the channel if it's not.

### DIFF
--- a/op-batcher/batch_submitter.go
+++ b/op-batcher/batch_submitter.go
@@ -12,6 +12,9 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/ethereum-optimism/optimism/op-proposer/rollupclient"
+	"github.com/ethereum/go-ethereum/rpc"
+
 	"github.com/ethereum-optimism/optimism/op-batcher/sequencer"
 	"github.com/ethereum-optimism/optimism/op-node/eth"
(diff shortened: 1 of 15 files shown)


trianglesphere:
So we need to characterize it, but you can look at `co.ReadyBytes` to determine how many compressed bytes are ready without messing up the compression ratio.

In general I think this is pretty complex without a clear need at this point & we should focus on stabilizing batch derivation & do this after.

protolambda:
@trianglesphere yea agreed, not merge ready yet. But I do believe we need a consistent way to re-attempt a batch creation, or else we end up submitting many tiny transactions. And for it to be consistent I believe we need to flush data the same way, or else the internal compression state may be wrong.

protolambda:
Closing this in favor of something minimal with `ReadyBytes` like @trianglesphere suggests.

PR #2914, opened 2022-07-01
devnet: Get devnet running

mslipper:
- Deploys to Goerli
- Adds a Hardhat script to generate testnet sequencer keys from a mnemonic
- Minor updates to the batch submitter to get better output


tynes (approved):

trianglesphere (commented):
batcher changes lgtm
Do you intend to check in the deployments? Also, it looks like the deployments have been renamed/changed from the contract itself to the proxy (but maybe including the contract, IDK  the deployment system that well)

tynes:
The `.chainId` file should be added back, some additional hh commands choke without it

maurelian (approved):

PR #2915, opened 2022-07-01
bedrock: check L1 origin changes when sequencing new blocks

protolambda:
@mslipper hit a bug where the sequencer is producing a block with bad data. To ensure we catch it when it happens again, so we can investigate the issue, we should defensively check the L1 origin changes.

### DIFF
--- a/op-node/rollup/driver/step.go
+++ b/op-node/rollup/driver/step.go
@@ -37,9 +37,19 @@ func (d *outputImpl) createNewBlock(ctx context.Context, l2Head eth.L2BlockRef, 
 	// case we need to fetch all transaction receipts from the L1 origin block so we can scan for
 	// user deposits.
 	if l2Head.L1Origin.Number != l1Origin.Number {
+		if l2Head.L1Origin.Hash != l1Origin.ParentHash {
+			d.log.Error("SEQUENCING BUG: cannot create new block on top of l2Head with new origin", "head", l2Head,
+				"head_origin", l2Head.L1Origin, "head_seq_nr", l2Head.SequenceNumber, "new_origin", l1Origin, "new_origin_parent", l1Origin.ParentID())
+			return l2Head, nil, fmt.Errorf("cannot create new block with L1 origin %s (parent %s) on top of L1 origin %s", l1Origin, l1Origin.ParentID(), l2Head.L1Origin)
+		}
 		l1Info, _, receipts, err = d.dl.Fetch(fetchCtx, l1Origin.Hash)
 		seqNumber = 0 // reset sequence number at the start of the epoch
 	} else {
+		if l2Head.L1Origin.Hash != l1Origin.Hash {
+			d.log.Error("SEQUENCING BUG: cannot create new block on top of l2Head with different origin at same height",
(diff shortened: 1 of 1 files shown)


op-node/rollup/driver/step.go
```diff
@@ -37,9 +37,19 @@ func (d *outputImpl) createNewBlock(ctx context.Context, l2Head eth.L2BlockRef,
...
+				"head_origin", l2Head.L1Origin, "head_seq_nr", l2Head.SequenceNumber, "new_origin", l1Origin, "new_origin_parent", l1Origin.ParentID())
+			return l2Head, nil, fmt.Errorf("cannot create new block with L1 origin %s (parent %d) on top of L1 origin %s", l1Origin, l1Origin.ParentID(), l2Head.L1Origin)
+		}
 		l1Info, _, receipts, err = d.dl.Fetch(fetchCtx, l1Origin.Hash)
 		seqNumber = 0 // reset sequence number at the start of the epoch
 	} else {
+		if l2Head.L1Origin.Hash != l1Origin.Hash {
+			d.log.Error("SEQUENCING BUG: cannot create new block on top of l2Head with different origin at same height",
+				"head", l2Head, "head_origin", l2Head.L1Origin, "head_seq_nr", l2Head.SequenceNumber, "new_origin", l1Origin, "new_origin_parent", l1Origin.ParentID())
+			return l2Head, nil, fmt.Errorf("cannot create new block with L1 origin %s (parent %d) on top of L1 origin %s", l1Origin, l1Origin.ParentID(), l2Head.L1Origin)
```
trianglesphere, line 51: How is %d working with `l1Origing.ParentID()`?

op-node/rollup/driver/step.go
```diff
@@ -37,9 +37,19 @@ func (d *outputImpl) createNewBlock(ctx context.Context, l2Head eth.L2BlockRef,
 	// case we need to fetch all transaction receipts from the L1 origin block so we can scan for
 	// user deposits.
 	if l2Head.L1Origin.Number != l1Origin.Number {
+		if l1Origin.ParentHash != l2Head.L1Origin.Hash {
```
trianglesphere, line 40: comparison is opposite above comparison

op-node/rollup/driver/step.go
protolambda: nice catch

op-node/rollup/driver/step.go
protolambda: I'm not sure if I understand your comment. If the l1 origin changes by number, and the parent of the new origin does not match the previous origin, then it is some inconsistent L1 chain reference and should thus error. That should be correct right?

protolambda: Oh I see, you mean the order of the operands, not the equality check itself.

trianglesphere: first comparison is `l2Head != l1Origin`, second is `l1Origin != l2Head`. The order of the arguments is different in each.

op-node/rollup/driver/step.go
trianglesphere: Also, I'm surprised a lint didn't catch this, I feel like they can look into format strings (or maybe that's the LSP).

op-node/rollup/driver/step.go
protolambda: Fixed (and fixed the fix, now all three comparisons match)

trianglesphere (approved):

PR #2917, opened 2022-07-01
feat(ctb): use ERC721Enumerable for standard token

smartcontracts:
**Description**
Replaces ERC721 with ERC721Enumerable for the standard version of the
OptimismMintableERC721 token. Enumerable is generally less useful on L1
because it consumes more gas but it makes indexing significantly easier
and gas is cheap on L2. Also removes reinitializer functions because we
don't want to be initializing contracts every time we re-deploy.

### DIFF
--- a/packages/contracts-periphery/contracts/L1/messaging/L1ERC721Bridge.sol
+++ b/packages/contracts-periphery/contracts/L1/messaging/L1ERC721Bridge.sol
@@ -21,7 +21,7 @@ contract L1ERC721Bridge is CrossDomainEnabled, OwnableUpgradeable {
     /**
      * @notice Contract version number.
      */
-    uint8 public constant VERSION = 1;
+    uint256 public constant VERSION = 2;
 
     /**
      * @notice Emitted when an ERC721 bridge to the other network is initiated.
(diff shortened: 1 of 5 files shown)


packages/contracts-periphery/contracts/L2/messaging/L2ERC721Bridge.sol
```diff
@@ -19,11 +17,11 @@ import { IOptimismMintableERC721 } from "../../universal/op-erc721/IOptimismMint
  *         acts as a minter for new tokens when it hears about deposits into the L1 ERC721 bridge.
  *         This contract also acts as a burner for tokens being withdrawn.
  */
-contract L2ERC721Bridge is CrossDomainEnabled, OwnableUpgradeable {
```
smartcontracts, line 22: We weren't using Ownable for anything, so we shouldn't have it

tynes (approved):

packages/contracts-periphery/contracts/L1/messaging/L1ERC721Bridge.sol
```diff
@@ -21,7 +21,7 @@ contract L1ERC721Bridge is CrossDomainEnabled, OwnableUpgradeable {
     /**
      * @notice Contract version number.
      */
-    uint8 public constant VERSION = 1;
+    uint256 public constant VERSION = 2;
```
maurelian, line 24: Breaks interface, that ok/desirable?

smartcontracts: Yes that's fine

smartcontracts: Communicated with Quixotic about it.

maurelian (approved):

PR #2923, opened 2022-07-02
style(ctb): remove Lib_ and OVM_ prefixes

smartcontracts:
**Description**
Removes Lib_ and OVM_ prefixes from all contracts. We're no longer using
prefixes for new contracts, I see no good reason to include prefixes for
all of the older contracts that are being included and overhauled as
part of this upgrade. Keeps everything consistent while also removing
some legacy cruft.


tynes:
This is missing changes to the hh tasks

packages/contracts-bedrock/contracts/L2/ETH.sol
```diff
@@ -11,18 +11,18 @@ import { OptimismMintableERC20 } from "../universal/OptimismMintableERC20.sol";
  * @custom:legacy
  * @custom:proxied
  * @custom:predeploy 0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000
- * @title OVM_ETH
- * @notice OVM_ETH is a legacy contract that held ETH balances before the Bedrock upgrade. All
- *         ETH balances held within this contract were migrated to the state trie as part of the
+ * @title ETH
```
maurelian, line 14: Having a contract named `ETH` is confusing.
I think we should keep the `OVM` prefix, or else change it to something like `DEAD_ETH` or just `DEAD`.

packages/contracts-bedrock/contracts/test/OptimismMintableTokenFactory.t.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity 0.8.10;
 
 import { Bridge_Initializer } from "./CommonTest.t.sol";
-import { LibRLP } from "./Lib_RLP.t.sol";
+import { LibRLP } from "./RLP.t.sol";
```
maurelian, line 5: Probably should do this then:

```suggestion
import { RLP } from "./RLP.t.sol";
```

smartcontracts: I believe LibRLP is an external library

packages/contracts-bedrock/contracts/L2/ETH.sol
smartcontracts: Having a contract that held ETH balances *was* confusing, lol. Can we meet in the middle and keep it called `ETH` but move it into the `legacy` folder? I'm trying to use `legacy` as the place to stuff things that people can ignore.

packages/contracts-bedrock/contracts/test/OptimismMintableTokenFactory.t.sol
smartcontracts: Was taken from here I think https://github.com/Rari-Capital/solmate/issues/207

Needs to be attributed though, idk why it doesn't have any attribution attached

maurelian: Agree about attribution. The prefix is not a big deal to me tho, do what you want here. :)

packages/contracts-bedrock/contracts/L2/ETH.sol
maurelian: I'm not sure `legacy` is exactly 'stuff you can ignore', considering it has the proxies which run the bridge/messenger.

I just really don't like `ETH`, and if people can/should ignore it, why not give it a name that makes its status as deprecated more obvious?

Maybe @tynes can break the tie?

packages/contracts-bedrock/contracts/test/OptimismMintableTokenFactory.t.sol
smartcontracts: Attribution: https://github.com/ethereum-optimism/optimism/pull/2927

packages/contracts-bedrock/contracts/L2/ETH.sol
smartcontracts: `LegacyERC20ETH`?

maurelian: works for me

smartcontracts: kk will change

tynes:
Generally looks good to me, waiting on CI to finish. I kind of liked the `Lib` prefix on things like `BytesUtils` but am ok either way

maurelian:
> Generally looks good to me, waiting on CI to finish. I kind of liked the `Lib` prefix on things like `BytesUtils` but am ok either way

Me too.

maurelian (approved):

tynes:
This is going to need a rebase

norswap (approved):

PR #2924, opened 2022-07-02
style(ctb): move external contracts to vendor dir

smartcontracts:
**Description**
Moves the two contracts we use that were written by external parties
into their own folder named vendor. By separating these contracts into
their own folder, we can be very explicit that these contracts are not
meant to be tested or modified.


mslipper (approved):

tynes:
No problems with this, we should eventually get slither configured now that its running to ignore the vendor directory. We could also try to figure out how to ignore it with coverage

tynes (approved):

PR #2925, opened 2022-07-02
op-node: unify verifier and sequencer payload attributes preparation

protolambda:
This started as a weekend idea - I want to test/secure the sequencer block preparation, but also help test eager batch derivation. There's significant overlap in functionality, so let's de-duplicate the code.

This PR:
- Factors out the payload attributes construction (excl. adding batch transactions) into a function that can be used by both verifier and sequencer
- Tests the edge cases of this attributes creation
- Adds a general test for the attributes queue, which is now more minimal.

Fixes ENG-2393


op-node/rollup/derive/attributes_queue.go
```diff
@@ -58,82 +48,43 @@ func (aq *AttributesQueue) Step(ctx context.Context, outer Progress) error {
...
-		Transactions:          txns,
-		// we are verifying, not sequencing, we've got all transactions and do not pull from the tx-pool
-		// (that would make the block derivation non-deterministic)
-		NoTxPool: true,
-	}
-	aq.log.Info("generated attributes in payload queue", "tx_count", len(txns), "timestamp", batch.Timestamp)
+	// we are verifying, not sequencing, we've got all transactions and do not pull from the tx-pool
+	// (that would make the block derivation non-deterministic)
+	attrs.NoTxPool = true
+	attrs.Transactions = append(attrs.Transactions, batch.Transactions...)
```
trianglesphere, line 71: Manually modifying the attributes to add the batch transactions is a weird interface.

op-node/rollup/derive/batch_queue.go
```diff
@@ -105,6 +105,7 @@ func (bq *BatchQueue) ResetStep(ctx context.Context, l1Fetcher L1Fetcher) error
 	bq.log.Info("found reset origin for batch queue", "origin", l1BlockStart)
 	bq.l1Blocks = append(bq.l1Blocks, l1BlockStart)
 	bq.progress.Origin = l1BlockStart
+	bq.progress.Closed = false
```
trianglesphere, line 108: Probably important (though I still don't really understand the progress API), but it's not related to the other changes in this PR.

op-node/rollup/derive/attributes_queue.go
```diff
@@ -43,9 +36,6 @@ func NewAttributesQueue(log log.Logger, cfg *rollup.Config, l1Fetcher L1Receipts
 }
 
 func (aq *AttributesQueue) AddBatch(batch *BatchData) {
-	if aq.progress.Closed {
```
trianglesphere, line 46: Why do some stages panic when adding stuff while closed & others don't?

trianglesphere (commented):
Generally makes sense

op-node/rollup/derive/attributes_queue.go
protolambda: Because that's what the stages expect. This stage can be called with empty batches that are created based on the close. Whereas other stages only ingest data immediately; never ingesting new data when the input is already exhausted.

op-node/rollup/derive/attributes_queue.go
protolambda: I'm thinking of it as preparing a template, building the default empty block attributes. It's then up to the user to modify the template. Do you have a suggestion for a better interface, without duplicating code?

trianglesphere: after reading it a bit more, I think I'm fine with this pattern as long as it's documented in the `Prepare` function

trianglesphere (approved):
lgtm. Really nice code deduplication.

protolambda:
Rebased onto `develop`, dropped the stale go mod tidy fixes (CI was not checking mod files at the time when this PR was opened, but they are good on current `develop`).

op-node/rollup/derive/attributes.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type L1ReceiptsFetcher interface {
+	InfoByHash(ctx context.Context, hash common.Hash) (eth.L1Info, error)
```
tynes, line 15: I think a devdoc here would be nice describing when to use these methods

op-node/rollup/derive/deposits.go
```diff
@@ -21,26 +22,30 @@ func UserDeposits(receipts []*types.Receipt, depositContractAddr common.Address)
 			if log.Address == depositContractAddr && len(log.Topics) > 0 && log.Topics[0] == DepositEventABIHash {
 				dep, err := UnmarshalDepositLogEvent(log)
 				if err != nil {
-					errs = append(errs, fmt.Errorf("malformatted L1 deposit log in receipt %d, log %d: %w", i, j, err))
+					result = multierror.Append(result, fmt.Errorf("malformatted L1 deposit log in receipt %d, log %d: %w", i, j, err))
```
tynes, line 25: Do we need multierror here when we can use native go error wrapping?
`err = fmt.Errorf("%v, %v", result, err)`? Unsure if it just works if `result` is `nil` (the first error)

trianglesphere: I mean you could chain errors natively. I kinda like multi-err b/c this was previously an array of errors.

protolambda:
Go-bedrock tests failing at the same point as in other PRs, I think `develop` is flaky due to the op-proposer bug and/or reorgs.

protolambda:
Rebased onto latest `develop`.

tynes (approved):

PR #2927, opened 2022-07-05
fix(ctb): attribute LibRLP

smartcontracts:
**Description**
Adds attribution to LibRLP.


maurelian (approved):

tynes:
> Are we sure that this is used anymore? It was pulled in originally because the `OptimismMintableERC20Factory` didn't return the address of the contract that was deployed and I now believe that it does

Its unused in `test/OptimismMintableERC20` but useful in the factory test because we want to know the addr beforehand so that we can assert on the event. We could get rid of the lib and use `ffi` + `cast` but I think its fine to keep it

tynes (approved):

PR #2930, opened 2022-07-05
chore: Add todo please

roninjin10:
In case the teams working on the monorepo want this too.  This matches what we did in gateway.   Encourage devs to make linear tickets for their TODOs.   

- Includes an example of what a TODO would look like in this pr
- Made it warn instead of error so I didn't have to make a linear ticket for the ~382 existing TODO comments in the repo.


tynes:
I see how this is useful but it makes the codebase less forkable and also ties our code to a centralized service

roninjin10:
@tynes makes sense to me

PR #2936, opened 2022-07-06
feat: Deduplicate node modules in post install

roninjin10:
Post install script to deduplicate node modules.   In previous pr [0fc0a05](https://github.com/ethereum-optimism/optimism/commit/0fc0a055c37b749fabb1cb12db78b178c97cc071) we were checking for this in CI.  This is unsafe to do so though with many devs possibly touching node modules at the same time.  If it's running in postInstall we have little to worry about as everytime we add a node module deduplication will happen.


roninjin10:
fyi @mslipper I think it skipped running checks on typescript packages.  Technically anytime yarn.lock changes all typescript checks should be triggered

roninjin10:
Slapped a do-not-merge on this since typescript checks never actually ran

roninjin10:
low priority fwiw no hurry here

PR #2938, opened 2022-07-06
contracts-bedrock: add tests for rlp-reader

tynes:
**Description**

Ports over the JS tests from the `contracts` package.
The next step is to get some fuzz tests going.


smartcontracts (approved):

maurelian (approved):

tynes:
It seems like there is a cache out there with an old docker image in it

```
#30 113.4 error: Found argument '--build-info' which wasn't expected, or isn't valid in this context
#30 113.4 
#30 113.4 	If you tried to supply `--build-info` as a value rather than a flag, use `-- --build-info`
#30 113.4 
#30 113.4 USAGE:
#30 113.4     forge build [OPTIONS]
```

PR #2942, opened 2022-07-06
feat(ctb): clean up initialization logic

smartcontracts:
**Description**
Cleans and standardizes the initialization logic across all contracts.
Importantly, removes reinitialization.


packages/contracts-bedrock/.storage-layout
```diff
@@ -49,15 +47,15 @@
 +---------------+-------------------------------------------------+------+--------+-------+----------------------------------------------------+
 | Name          | Type                                            | Slot | Offset | Bytes | Contract                                           |
 +==============================================================================================================================================+
-| messenger     | contract CrossDomainMessenger                   | 0    | 0      | 20    | contracts/L1/L1StandardBridge.sol:L1StandardBridge |
+| _initialized  | uint8                                           | 0    | 0      | 1     | contracts/L1/L1StandardBridge.sol:L1StandardBridge |
```
smartcontracts, line 50: Storage layout changed, but it's OK. We'll need to review all of the storage layouts before we go to mainnet. I spoke with @tynes and we agreed that it would be better to make storage layout changes now for the sake of readability/speed for now and make tweaks to match the original storage layout at the very end of this process.

packages/contracts-bedrock/contracts/L1/L1CrossDomainMessenger.sol
```diff
@@ -13,34 +13,26 @@ import { CrossDomainMessenger } from "../universal/CrossDomainMessenger.sol";
  *         interface instead of interacting with lower-level contracts directly.
  */
 contract L1CrossDomainMessenger is CrossDomainMessenger {
-    /**
-     * @notice Contract version number.
-     */
-    uint8 public constant VERSION = 1;
```
smartcontracts, line 19: Removing `VERSION` in favor of using a base contract like the `Semver` we have in `contracts-periphery` instead.

tynes: we could import that contract here?

packages/contracts-bedrock/contracts/universal/OptimismMintableTokenFactory.sol
```diff
@@ -41,15 +41,12 @@ contract OptimismMintableTokenFactory {
     /**
      * @notice Address of the StandardBridge on this chain.
      */
-    address public bridge;
```
tynes, line 44: This contract will technically be upgradable as its a predeploy

tynes (approved):

tynes:
I think we need a follow up PR that adds the upgradability scheme to the predeploys as well

smartcontracts:
> I think we need a follow up PR that adds the upgradability scheme to the predeploys as well

Will do

packages/contracts-bedrock/contracts/universal/OptimismMintableTokenFactory.sol
smartcontracts: Yeah I'll put it behind Initializable (along with every other predeploy) in an upcoming PR... we'll need to figure out how that impacts storage layout and if there's any easy way to fix that :-/

maurelian: FWIW, the only predeploy that has a constructor is `GasPriceOracle`, so I don't think the other ones will need `Initializable`?

Though it would be good to add `uint256 constant public version` to them.

smartcontracts: Note my comment here regarding versioning: https://github.com/ethereum-optimism/optimism/pull/2942#discussion_r915135867

packages/contracts-bedrock/contracts/L1/L1CrossDomainMessenger.sol
```diff
@@ -13,34 +13,29 @@ import { CrossDomainMessenger } from "../universal/CrossDomainMessenger.sol";
...
 
     /**
-     * @notice Intializes mutable variables.
-     *
-     * @param _portal Address of the OptimismPortal to send and receive messages through.
+     * @notice Initializer.
      */
-    function initialize(OptimismPortal _portal) public reinitializer(VERSION) {
-        portal = _portal;
+    function initialize() public initializer {
```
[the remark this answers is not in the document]
maurelian: Oh, right, that's for the `__X_init` functions.

packages/contracts-bedrock/contracts/universal/OptimismMintableERC20Factory.sol
```diff
@@ -41,15 +41,12 @@ contract OptimismMintableERC20Factory {
...
+    address public immutable bridge;
 
     /**
-     * @notice Initializer.
-     *
      * @param _bridge Address of the StandardBridge on this chain.
      */
-    function initialize(address _bridge) public {
-        require(bridge == address(0), "OptimismMintableERC20Factory: already initialized");
+    constructor(address _bridge) {
```
maurelian, line 49: This is no longer upgradeable, or it doesn't need an initializer because we don't need to change values when changing the impl (ie. the conversation we had the other day).

tynes: If we want to upgrade this, then we can just deploy a new implementation and then call `upgradeTo` on the implementation since the constructor has no state access

tynes (approved):

maurelian (approved):

tynes (changes requested):
We need to update the deploy scripts:

```
An unexpected error occurred:

Error: ERROR processing /home/circleci/project/packages/contracts-bedrock/deploy/002-L1CrossDomainMessenger.ts:
Error: no matching function (argument="signature", value="initialize(address)", code=INVALID_ARGUMENT, version=abi/5.6.3)
```

tynes:
Looks like it needs another rebase for the gas snapshot

tynes:
Looks like a flake in the `bedrock-go-tests`

packages/contracts-bedrock/deploy/002-L1CrossDomainMessenger.ts
```diff
@@ -37,9 +37,7 @@ const deployFn: DeployFunction = async (hre) => {
 
   const upgradeTx = await Proxy.upgradeToAndCall(
     messenger.address,
-    L1CrossDomainMessenger.interface.encodeFunctionData('initialize(address)', [
-      portal.address,
-    ])
+    L1CrossDomainMessenger.interface.encodeFunctionData('initialize')
```
tynes, line 40: Does this need to be `initialize()` or does it not matter?

tynes:
I'm going to investigate as to why this happened:

```
An unexpected error occurred:

TypeError: Cannot read properties of undefined (reading 'storage')
    at computeStorageSlots (/home/circleci/project/node_modules/@defi-wonderland/smock/src/utils/storage.ts:85:38)
    at SimpleTaskDefinition.action (/home/circleci/project/packages/contracts-bedrock/tasks/genesis-l2.ts:250:21)
```

It is in the generation of the L2 `genesis.json`

packages/contracts-bedrock/deploy/002-L1CrossDomainMessenger.ts
smartcontracts: Doesn't matter

tynes (approved):

PR #2944, opened 2022-07-06
contracts-bedrock: fix typechain

tynes:
**Description**

Update to the latest version of typechain and no longer
use the hh plugin. Directly call typechain as the hardhat
plugin is unreliable when using foundry.


smartcontracts (approved):

roninjin10:
curious why the hardhat plugin is unreliable?

PR #2946, opened 2022-07-06
docker: add ethereumoptimism/foundry docker image

tynes:
**Description**

Based on the following commit:
https://github.com/foundry-rs/foundry/commit/b7b1ec471bdd38221773e1a569dc4f20297bd7db

Allows us to control the version of foundry being used in our toolchain. This dockerfile is exactly the same as upstream's but pins to a specific commit hash.

### DIFF
--- a/.changeset/big-scissors-remain.md
+++ b/.changeset/big-scissors-remain.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/foundry': minor
+---
+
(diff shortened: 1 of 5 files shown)


trianglesphere (approved):
Does this require this commit to land on master before the docker image is built?
I'm just not that familiar with how often change get pulled into master from develop.

smartcontracts (approved):

PR #2947, opened 2022-07-06
contracts-bedrock: speed up compilation time

tynes:
**Description**

This commit greatly speeds up the compilation
time for `hardhat compile` by updating to the
latest version of the plugin.

This is known to work with the following forge version:
```
forge 0.2.0 (4ad4bb6 2022-07-06T17:55:43.107018225Z)
```

To use this version:
```
$ foundryup -C 4ad4bb660ad0b429e76e980fd6c1d50aee7c112b
```

### DIFF
--- a/packages/contracts-bedrock/hardhat.config.ts
+++ b/packages/contracts-bedrock/hardhat.config.ts
@@ -47,9 +47,6 @@ const config: HardhatUserConfig = {
   },
   foundry: {
     buildInfo: true,
-    // call runSuper in the hardhat task so that
-    // other things execute
-    runSuper: true,
   },
   paths: {
     deploy: './deploy',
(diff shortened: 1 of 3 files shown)


trianglesphere (approved):

packages/contracts-bedrock/hardhat.config.ts
```diff
@@ -47,9 +47,6 @@ const config: HardhatUserConfig = {
   },
   foundry: {
     buildInfo: true,
-    // call runSuper in the hardhat task so that
-    // other things execute
-    runSuper: true,
```
smartcontracts, line 52: What's the effect of removing this?

tynes: It speeds up compilation speed by not calling `runSuper` in the hh task. Since we no longer are using hh-typechain, we don't need to call `runSuper`. Its about 50% faster with this removed

PR #2950, opened 2022-07-07
fix(ctb): rename token factory to erc20 factory

smartcontracts:
**Description**
With the intent of mirroring the naming of the ERC721 bridge, I'm
renaming the OptimismMintableTokenFactory to
OptimismMintableERC20Factory.


tynes:
I think we need updates in the hh tasks

smartcontracts:
> I think we need updates in the hh tasks

Done

tynes (approved):

maurelian (approved):

PR #2951, opened 2022-07-07
fix(ctb): solc warnings in Proxy.sol

smartcontracts:
**Description**
Cleans up a few solc warnings in Proxy.sol.


packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -57,6 +57,12 @@ contract Proxy {
         _doProxyCall();
     }
 
+    // slither-disable-next-line locked-ether
```
tynes, line 60: A payable fallback fn isn't the same as a receive?

tynes (approved):

maurelian (approved):

PR #2953, opened 2022-07-07
style(ctb): clean up legacy comments and errs

smartcontracts:
**Description**
Cleans up comments and errors for legacy contracts. It's fine to update
the bytecode for these legacy contracts since we're really only using
them for their interfaces.


packages/contracts-bedrock/contracts/legacy/AddressResolver.sol
```diff
@@ -1,40 +0,0 @@
-// SPDX-License-Identifier: MIT
-pragma solidity ^0.8.9;
-
-/* Library Imports */
-import { AddressManager } from "./AddressManager.sol";
-
-/**
- * @title AddressResolver
- */
-abstract contract AddressResolver {
```
maurelian, line 10: When did this cease to be in use?

maurelian (approved):

packages/contracts-bedrock/contracts/legacy/AddressResolver.sol
smartcontracts: It was never in use for contracts-bedrock, not sure why it was copied in the first place

tynes (changes requested):
This needs the bindings to be regenerated

tynes:
Just pushed new bindings

tynes (approved):

tynes (approved):

PR #2954, opened 2022-07-07
style(ctb): Remove underscores from event args

maurelian:
**Description**
All events args in the bedrock package should now be without the prefixed `_`


packages/contracts-bedrock/.gas-snapshot
```diff
@@ -1,242 +1,242 @@
 GasBenchMark_L1CrossDomainMessenger:test_L1MessengerSendMessage_benchmark_0() (gas: 160669)
-GasBenchMark_L1CrossDomainMessenger:test_L1MessengerSendMessage_benchmark_1() (gas: 77078)
+GasBenchMark_L1CrossDomainMessenger:test_L1MessengerSendMessage_benchmark_1() (gas: 77100)
```
maurelian, line 2: Why would this change the gas costs? I have no idea, and I'm trying not to get nerd sniped by it, but if anyone knows offhand I'd love to hear it.

tynes: The bytecode didn't change according to `l2outputoracle.go`, only the abi changed. Unclear as to why the gas costs changed, there must be another variable someplace. Perhaps something with a different amount of 0 bytes?

maurelian: Does setUp code influence the gas cost of a test? I didn't think so.

Even the L1Block costs changed, that contract has no deps and was untouched.
I'm gonna assume its foundry. Hopefully CI fails on this.

maurelian: oh good: https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/4222/workflows/e1e3b5c0-c640-4633-a777-16092ab0762e/jobs/60445/parallel-runs/0/steps/0-107

smartcontracts (approved):
Perfect, thanks. Worth noting that changing event arg names *technically* changes the ABI but in practice has no impact because the arg names are purely used a client-side utility. So this is safe to do.

tynes:
This needs a new gas snapshot to be taken

tynes (approved):

PR #2957, opened 2022-07-07
feat(ctb): use external BytesLib

smartcontracts:
**Description**
Replaces our BytesUtils with the BytesLib upon it was originally based.
Only significant function within BytesUtils was slice, which was
copy/pasted from the external BytesLib that we're now importing. Pulls
out the nibble-related functions into a separate library.


packages/contracts-bedrock/contracts/libraries/Nibbles.sol
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * @notice Converts a byte array into a nibble array by splitting each byte into two nibbles.
+     *         Resulting nibble array will be exactly twice as long as the input byte array.
+     *
+     * @param _bytes Input byte array to convert.
+     *
+     * @return Resulting nibble array.
+     */
+    function toNibbles(bytes memory _bytes) internal pure returns (bytes memory) {
+        bytes memory nibbles = new bytes(_bytes.length * 2);
```
tynes, line 18: I'd like to follow up with a PR that caches the length and increments in an unchecked block

tynes:
The go bindings need to be regenerated

tynes:
Getting an error in the e2e tests:

```
  system_test.go:896: 
        	Error Trace:	/root/project/op-e2e/system_test.go:896
        	Error:      	Expected nil, but got: &rpc.jsonError{Code:3, Message:"execution reverted: slice_outOfBounds", Data:"0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000011736c6963655f6f75744f66426f756e6473000000000000000000000000000000"}
        	Test:       	TestWithdrawals
```

.circleci/config.yml
```diff
@@ -143,7 +143,7 @@ jobs:
           working_directory: packages/contracts-bedrock
       - run:
           name: check go bindings
-          command: make && git diff --exit-code
+          command: make -j 8 && git diff --exit-code
```
smartcontracts, line 146: 8x parallel, speeds up bindings by ~4x on my machine.

tynes: This check is pretty slow, recent foundry commit https://github.com/foundry-rs/foundry/commit/cca65b1fc6c7ab50b5c84c99c9d3537cf00dbbbb should speed it up.

Open to having this as a separate PR, altho we should make `8` be dynamic based on the # of cores

smartcontracts: Yeah I had to revert this because -j 8 didn't work on CI

PR #2958, opened 2022-07-07
ci: add devnet to ci

tynes:
**Description**

Prevent bugs that break the devnet by
doing a deposit in CI on the devnet


.circleci/config.yml
```diff
@@ -384,6 +384,51 @@ jobs:
...
+      - checkout
+      - run:
+          name: Check if we should run
+          command: |
+            CHANGED=$(bash ./ops/docker/ci-builder/check-changed.sh "(contracts-bedrock|op-bindings|op-batcher|op-node|op-proposer|ops-bedrock)/")
+            if [[ "$CHANGED" = "FALSE" ]]; then
+              circleci step halt
+            fi
+      - run:
+          name: foundryup
```
trianglesphere, line 403: Use a builder that has foundry? This seems to be the only reference to foundry.

tynes: I wasn't sure if we can run docker in docker so i am installing foundry this way

trianglesphere: docker in docker would be a bit nice to have the same foundry version, but this seems fine then

protolambda (approved):

.circleci/config.yml
mslipper: I've run docker in docker on Circle before, but it's a bit complicated. I'm also good with merging as-is, we can refactor later.

mslipper:
@mergifyio refresh

PR #2960, opened 2022-07-08
feat(bedrock): Add version to TransactionDeposited event

maurelian:
**Description**
Adds a version to the `TransactionDeposited` event simply by renaming the event. This approach does not require an extra data, it simply modifies the hash that is already logged as the first topic.


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -159,9 +159,9 @@ contract OptimismPortal is Initializable, ResourceMetering {
             from = AddressAliasHelper.applyL1ToL2Alias(msg.sender);
         }
 
-        // Emit a TransactionDeposited event so that the rollup node can derive a deposit
+        // Emit a TransactionDepositedV0 event so that the rollup node can derive a deposit
         // transaction for this deposit.
-        emit TransactionDeposited(from, _to, msg.value, _value, _gasLimit, _isCreation, _data);
+        emit TransactionDepositedV0(from, _to, msg.value, _value, _gasLimit, _isCreation, _data);
```
tynes, line 164: I thought the idea was to do something like this:

```
TransactionDeposited(uint256 version, address indexed to, address indexed from, bytes transaction)
```

cc @protolambda @trianglesphere

maurelian: Hmmm, yeah. TBH I forgot there was an existing spec. I like this approach because it just changes topic0 and doesn't cost extra gas, but seeing as there's been [discussion already](https://www.notion.so/optimismpbc/Versioning-Deposits-c759d50a9bf94ca3b77a9c0387d5c3e6), I'll just do that.

PR #2961, opened 2022-07-08
fix(ctb): remove DefaultValues

smartcontracts:
**Description**
Unnecessary library, only being used in a single place now and it can
just be an internal constant. I see no strong reason for this library to
continue to exist.


maurelian (approved):

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -153,7 +159,7 @@ abstract contract CrossDomainMessenger is
      */
     function xDomainMessageSender() external view returns (address) {
         require(
-            xDomainMsgSender != DefaultValues.DEFAULT_XDOMAIN_SENDER,
+            xDomainMsgSender != DEFAULT_XDOMAIN_SENDER,
             "xDomainMessageSender is not set"
```
maurelian, line 163: linter error is here

tynes:
Since `xDomainMessageSender` reverts, this is a good change. It doesn't need to be in a lib for the caller to do a comparison against

tynes (changes requested):
needs a linter fix

smartcontracts:
> needs a linter fix

Resolved

tynes (approved):

PR #2962, opened 2022-07-08
contracts-bedrock: add semver to L2 contracts

tynes:
**Description**
Adds `contracts-periphery` as a dep and then adds semver to each of the predeploys. They are all `0.0.1`, perhaps we make them all `1.0.0` on release of bedrock


packages/contracts-bedrock/package.json
```diff
@@ -50,6 +50,7 @@
   },
   "devDependencies": {
     "@eth-optimism/hardhat-deploy-config": "^0.2.0",
+    "@eth-optimism/contracts-periphery": "^0.1.5",
```
smartcontracts, line 53: I would prefer that the dependency chain go from `contracts-bedrock` to `contracts-periphery`. I'm fine with moving `Semver` into `contracts-bedrock` to achieve this. I just don't want a circular dependency and I can imagine building peripheral contracts that interop with some of our core contracts.

smartcontracts (approved):
Nice, I like it. +1 on using `@custom:semver`.

maurelian (approved):

PR #2964, opened 2022-07-08
fix(ctb): clean up BytesUtils

smartcontracts:
**Description**
Cleans up BytesUtils and properly attributes the slice code (although
it's licensed as Unlicense so I don't think strictly necessary, but
attribution is always nice and I think we just missed it the first time
around).


packages/contracts-bedrock/contracts/libraries/trie/MerkleTrie.sol
```diff
@@ -160,7 +160,7 @@ library MerkleTrie {
             } else {
                 // Nodes smaller than 31 bytes aren't hashed.
                 require(
-                    BytesUtils.toBytes32(currentNode.encoded) == currentNodeID,
+                    bytes32(currentNode.encoded) == currentNodeID,
```
smartcontracts, line 163: This was suggested by Spearbit. Has an identical effect, and is available since 0.8.5.

packages/contracts-bedrock/contracts/libraries/BytesUtils.sol
```diff
@@ -3,19 +3,27 @@ pragma solidity ^0.8.9;
...
+     *
+     * @return Slice of the input byte array.
+     */
     function slice(
         bytes memory _bytes,
         uint256 _start,
         uint256 _length
     ) internal pure returns (bytes memory) {
         require(_length + 31 >= _length, "slice_overflow");
-        require(_start + _length >= _start, "slice_overflow");
```
smartcontracts, line 18: Unnecessary check as of 0.8 since overflow is checked by default. Note that this is also removed in the upstream code.

packages/contracts-bedrock/contracts/libraries/BytesUtils.sol
```diff
@@ -3,19 +3,27 @@ pragma solidity ^0.8.9;
...
  * @title BytesUtils
+ * @notice BytesUtils is a library for manipulating byte arrays.
  */
 library BytesUtils {
-    /**********************
-     * Internal Functions *
-     **********************/
-
+    /**
+     * @custom:attribution https://github.com/GNSPS/solidity-bytes-utils/blob/6458fb2780a3092bc756e737f246be1de6d3d362/contracts/BytesLib.sol#L228-L295
```
smartcontracts, line 10: I think it would be nice to include `@custom:attribution` tags for any code that we're using via external sources. Then we can easily link to attribution in the generated docs.

packages/contracts-bedrock/contracts/libraries/BytesUtils.sol
tynes: Following the same logic, the line above could be removed as well?

smartcontracts: Actually sorry this is incorrect, since we're using this in assembly it's actually necessary to do the checks manually. Needs to be wrapped in `unchecked` though or you'll get incorrect errors.

smartcontracts: Not sure exactly why that check was removed upstream but in practice you'd have to have `_start` or `_length` be close to max u256, which would use an extraordinary amount of gas. I doubt you could even get remotely close to that.

smartcontracts: Anyway I added the check back in just to be safe

tynes:
This needs a rebase + regeneration of the bindings

maurelian (approved):

tynes (approved):

PR #2965, opened 2022-07-08
feat(bedrock): Versioned TransactionDeposited event

maurelian:
**Description**

Adds a version to the `TransactionDeposited` event with an extra argument. 

And update Go code (Josh/Proto).

**Metadata**
- Fixes ENG-2293


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -21,26 +21,25 @@ contract OptimismPortal is Initializable, ResourceMetering {
...
      */
     event TransactionDeposited(
         address indexed from,
         address indexed to,
-        uint256 mint,
-        uint256 value,
-        uint64 gasLimit,
-        bool isCreation,
-        bytes data
+        uint256 indexed version,
```
maurelian, line 41: Making this indexed saves about 100 gas. I'm not attached to it tho.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -159,9 +159,11 @@ contract OptimismPortal is Initializable, ResourceMetering {
             from = AddressAliasHelper.applyL1ToL2Alias(msg.sender);
         }
 
+        bytes memory opaqueData = abi.encode(msg.value, _value, _gasLimit, _isCreation, _data);
```
maurelian, line 162: `opaqueData` was the term used in the spec, but IMO `depositData` would be more reader friendly.

maurelian:
superceded by #2976

tynes (approved):

op-node/rollup/derive/deposit_log.go
```diff
@@ -12,34 +12,55 @@ import (
...
+	from := common.BytesToAddress(ev.Topics[1][12:])
+	// indexed 1
+	to := common.BytesToAddress(ev.Topics[2][12:])
+	// indexed 2
+	version := ev.Topics[3]
+	// unindexed data
+	var opaqueContentOffset uint256.Int
+	opaqueContentOffset.SetBytes(ev.Data[0:32])
+	if !opaqueContentOffset.IsUint64() || opaqueContentOffset.Uint64() != 32 {
+		return nil, fmt.Errorf("invalid opaqueData slice header offset: %s", &opaqueContentOffset)
```
tuxcanfly, line 55: nit: can use `%d` to format `uint256.Int`?

op-node/rollup/derive/deposit_log.go
```diff
@@ -12,34 +12,55 @@ import (
...
+	// indexed 0
+	from := common.BytesToAddress(ev.Topics[1][12:])
+	// indexed 1
+	to := common.BytesToAddress(ev.Topics[2][12:])
+	// indexed 2
+	version := ev.Topics[3]
+	// unindexed data
+	var opaqueContentOffset uint256.Int
+	opaqueContentOffset.SetBytes(ev.Data[0:32])
+	if !opaqueContentOffset.IsUint64() || opaqueContentOffset.Uint64() != 32 {
```
tuxcanfly, line 54: Can use a comment to explain the fixed offset `32`. Is the offset from the start of `ev.Data`, and if so shouldn't it be `64` considering we're reading `ev.Data[64+...` down below?

op-node/rollup/derive/deposit_log.go
tuxcanfly: yep seems better:

```
x = %s [%!s(uint64=0) %!s(uint64=0) %!s(uint64=0) %!s(uint64=0)]
x = %d [0 0 0 0]
```

op-node/rollup/derive/deposit_log.go
```diff
@@ -12,34 +12,55 @@ import (
...
+		return nil, fmt.Errorf("invalid opaqueData slice header offset: %s", &opaqueContentOffset)
+	}
+	var opaqueContentLength uint256.Int
+	opaqueContentLength.SetBytes(ev.Data[32:64])
+	// make sure the length is an uint64, it's not larger than the remaining data, and the log is using minimal padding (i.e. can't add 32 bytes without exceeding data)
+	if !opaqueContentLength.IsUint64() || opaqueContentLength.Uint64() > uint64(len(ev.Data)-64) || opaqueContentLength.Uint64()+32 <= uint64(len(ev.Data)-64) {
+		return nil, fmt.Errorf("invalid opaqueData slice header length: %s", &opaqueContentLength)
```
tuxcanfly, line 61: nit: `%d`

tuxcanfly (commented):
OK, needs rebase to fix merge conflicts.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tuxcanfly: is it the same with opaqueData? Sounds more useful to that indexed instead.

op-node/rollup/derive/deposit_log.go
tuxcanfly: Considering the var is called `opaqueContentOffset`, it seems that the offset must indicate start of the content.

tuxcanfly: I see that the offset `0x20` is document below. Going to give the benefit of doubt and assume it means content including the length.

tuxcanfly (approved):
OK.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: Indexed arguments are taken from the stack, so if `opaqueData` were indexed then it would be hashed to fit into 256 bits.

PR #2968, opened 2022-07-08
contracts-bedrock: optimization of initialize

tynes:
**Description**

Will slightly reduce the cost of deployment. Still
is very simple to read. Note that the `.gas-snapshot`
is not changed due to the initialization of the
cross domain messengers happening inside of the `setUp`
function.

### DIFF
--- a/.changeset/honest-bags-press.md
+++ b/.changeset/honest-bags-press.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/contracts-bedrock': patch
+---
+
+Optimize the initialize function of the xdomain messenger slightly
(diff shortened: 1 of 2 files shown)


maurelian (approved):
Based on the description I was afraid I was going to see assembly in the constructor. I think this change is arise just a best practice in solidity. Deserves a linter rule...

smartcontracts:
I feel like this isn't really necessary in the context of the constructor. Generally I would prefer legibility over optimization unless the gas savings will be significant over a long period of time. Since this is only running once, I would rather have the more legible original version of the code.

tynes:
Thats understandable but i don't think its that illegible

PR #2972, opened 2022-07-11
style(ctb): clean up enums

smartcontracts:
**Description**
Enums should have CapitalCase names and UPPER_SNAKE_CASE values.
Standardizes around that. Also removes an unused enum in MerkleTrie.sol.


tynes (approved):

tynes:
Looks like tests need to be updated too

```
  --> contracts/test/ProxyAdmin.t.sol:55:49:
```

PR #2974, opened 2022-07-11
style(ctb): refactor Hashing and Encoding libs

smartcontracts:
**Description**
Cleans up CrossDomainUtils, CrossDomainHashing, and WithdrawalVerifier
into two new libraries, Hashing and Encoding. Will be followed up by a
second PR to clean up function names and comments.


packages/contracts-bedrock/contracts/libraries/Encoding.sol
```diff
@@ -0,0 +1,153 @@
...
+    {
+        assembly {
+            nonce := or(shl(240, _version), _nonce)
+        }
+    }
+
+    /**
+     * @notice Gets the version out of the nonce
+     */
+    function getVersionFromNonce(uint256 _nonce) internal pure returns (uint16 version) {
```
tynes, line 148: No named return args

packages/contracts-bedrock/contracts/libraries/Encoding.sol
```diff
@@ -0,0 +1,153 @@
...
+                _sender,
+                _message,
+                _messageNonce
+            );
+    }
+
+    /**
+     * @notice RLP encode a deposit transaction
+     * This only works for user deposits, not system deposits
+     * TODO: better name + rearrange the input param ordering?
```
maurelian, line 39: I think this is a good thing to do.
Would be nice to do it here rather than introducing a new TODO.

packages/contracts-bedrock/contracts/libraries/Encoding.sol
maurelian: +1

packages/contracts-bedrock/contracts/libraries/Encoding.sol
maurelian: gotcha

smartcontracts:
As noted in the commit message, I will be following this PR up with a second PR to clean up function names, comments, and any other TODOs. I just wanted to minimize the size of this PR by simply copy/pasting the old content.

maurelian (approved):

tynes:
Looks like some linting errors but generally lgtm

tynes (approved):

tynes:
@mergifyio refresh

PR #2976, opened 2022-07-11
Node: Update fuzz test to use new deposit type

trianglesphere:
**Description**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


op-node/rollup/derive/deposit_log_test.go
```diff
@@ -11,6 +11,7 @@ import (
 )
 
 func TestUnmarshalLogEvent(t *testing.T) {
+	t.Skip("not working because deposit_log_create not working properly")
```
maurelian, line 14: When unskipped, this test fails with: 
`/Users/maurelian/Projects/O/mono/op-node/rollup/derive/deposit_log_test.go:29: expected 4 event topics (event identity, indexed from, indexed to, indexed version), got 3`

maurelian: To fix, `MarshalDepositLogEvent` will need to return the version on the `Topics` field.

https://github.com/ethereum-optimism/optimism/blob/19490d318/op-node/rollup/derive/deposit_log.go#L119-L156

maurelian:
Summary from speaking with @trianglesphere about what is needed here: 

1. In `MarshalDepositLogEvent`:
	1. return the version on the Topics array ([here](https://github.com/ethereum-optimism/optimism/blob/19490d318/op-node/rollup/derive/deposit_log.go#L156))
	2. generate the opaque data bytes & then encode that again
2. In `UnmarshalDepositLogEvent` add length and offset checks fix the decoding of `ev.Data`

PR #2979, opened 2022-07-11
getL2Output reverts if not found

maurelian:
**Description**

Calls to `oracle.getL2Output` revert if no output is found for the given block number. 

**Pros:**

Reverting on blocks which are not found forces 3rd parties to consider and handle this failure mode, instead of possibly missing this 'edge case'. This feels more semantically correct, and is similar to how many of ERC721's functions handle non-existent token IDs (ex. [getApproved](https://docs.openzeppelin.com/contracts/2.x/api/token/erc721#ERC721-getApproved-uint256-)).

**Cons:**

This change forces _us_ to introduce some pretty ugly try/catch code in `isOutputFinalized()`.  

**Metadata**
- Fixes #ENG-2279


maurelian:
@smartcontracts: I discussed this change with @tynes. Would appreciate your input before I spend time sending the go tests.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -162,32 +162,40 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
+     * @notice Determine if an L2 Output is finalized for a given block number.
      *
      * @param _l2BlockNumber The number of the L2 block.
      */
-
     function isOutputFinalized(uint256 _l2BlockNumber) external view returns (bool) {
-        L2OutputOracle.OutputProposal memory proposal = L2_ORACLE.getL2Output(_l2BlockNumber);
-
-        if (proposal.outputRoot == bytes32(uint256(0))) {
+        L2OutputOracle.OutputProposal memory proposal;
```
smartcontracts, line 170: I don't think this try/catch stuff is really necessary. I would prefer that we keep this simple by doing the following:
1. Compute the block number that corresponds to the output that would've been published for a given L2 block number
2. Get the L2 output for that computed block number, it's fine if the error gets thrown if that output does not exist.
3. If it does exist, do the simple timestamp check

maurelian: Yeah, I was thinking about a similar alternative approach yesterday evening. Will do.

tynes (approved):

maurelian:
@Mergify refresh

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -162,32 +162,39 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
-    function isOutputFinalized(uint256 _l2BlockNumber) external view returns (bool) {
-        L2OutputOracle.OutputProposal memory proposal = L2_ORACLE.getL2Output(_l2BlockNumber);
+        uint256 interval = L2_ORACLE.SUBMISSION_INTERVAL();
+        uint256 startingBlockNumber = L2_ORACLE.STARTING_BLOCK_NUMBER();
+        if (startingBlockNumber > _l2BlockNumber) {
+            return false;
+        }
+
+        // Find the distance between _l2BlockNumber, and the checkpoint block before it.
+        uint256 offset = (_l2BlockNumber - startingBlockNumber) % interval;
```
smartcontracts, line 179: I think this math makes more sense in `getL2Output`, since I would also want to use it on the client-side and this saves me doing some math.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -162,32 +162,39 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
-
-            // Find the distance between the _l2BlockNumber, and the checkpoint block before it.
-            uint256 offset = (_l2BlockNumber - startingBlockNumber) % interval;
-            // Look up the checkpoint block after it.
-            proposal = L2_ORACLE.getL2Output(_l2BlockNumber + (interval - offset));
-            // False if that block is not yet proposed.
-            if (proposal.outputRoot == bytes32(uint256(0))) {
-                return false;
-            }
+        try L2_ORACLE.getL2Output(lookupBlockNumber) returns (
```
smartcontracts, line 187: I don't think the try/catch is necessary, I think it's fine if this reverts instead of returning false.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -228,17 +235,15 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
             "OptimismPortal: you cannot send messages to the portal contract"
         );
 
-        // Get the output root.
+        // Get the output root. This will fail if there is no
+        // output root for the given block number.
         L2OutputOracle.OutputProposal memory proposal = L2_ORACLE.getL2Output(_l2BlockNumber);
```
smartcontracts, line 240: Argument for doing the math in L2OutputOracle is because otherwise this will fail if the user doesn't provide the exact correct block number

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -235,15 +226,16 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
             "OptimismPortal: you cannot send messages to the portal contract"
         );
 
-        // Get the output root.
+        // Get the output root. This will fail if there is no
+        // output root for the given block number.
```
maurelian, line 230: Note that this function will now succeed for multiple values of `_l2BlockNumber`.

maurelian:
@Mergify refresh

maurelian:
@mergify refresh

PR #2980, opened 2022-07-11
contracts-bedrock: differential fuzzing

tynes:
**Description**
Begin more thorough testing of withdrawals by starting with some differential fuzzing of versioned encodings and versioned hashes. The config was updated with `ffi = true` so that the `vm.ffi` cheatcode can be used.

The version of `forge-std` was also updated so that the `vm.toString` cheatcodes could be used.

The encodings/hashing lib was implemented in `core-utils`


tynes:
These tests fail on nodejs v16 but pass on nodejs 14. There must have been a change with std? Investigating more. The ci builder uses 16

packages/contracts-bedrock/scripts/differential-testing.ts
```diff
@@ -0,0 +1,135 @@
...
+      gasLimit,
+      data
+    )
+    const output = utils.defaultAbiCoder.encode(['bytes32'], [hash])
+    process.stdout.write(output)
+    break
+  }
+  case 'hashDepositTransaction': {
+    // The solidity transaction hash computation currently only works with
+    // user deposits. System deposit transaction hashing is not supported.
```
maurelian, line 70: Why is that?

tynes: System deposits compute hash differently based on a different way to compute the source hash

packages/contracts-bedrock/package.json
```diff
@@ -19,7 +19,7 @@
     "build": "hardhat compile && yarn build:ts && yarn typechain",
     "build:ts": "tsc -p tsconfig.json",
     "deploy": "hardhat deploy",
-    "test": "forge test",
+    "test": "yarn build:ts && forge test",
```
maurelian, line 22: This does slow things down a bit. Is it reasonable to expect ppl to run `yarn build` first?

packages/contracts-bedrock/scripts/differential-testing.ts
maurelian: OK, I see that in the specs. Missed when that got added.

maurelian: Is it possible we'll need it in solidity anytime soon?

packages/contracts-bedrock/package.json
tynes: This is mostly for ci, running locally you just need to run `forge test` after building ts once. We can lower the fuzz runs in the local profile to speed it up

packages/contracts-bedrock/scripts/differential-testing.ts
tynes: > Is it possible we'll need it in solidity anytime soon?

I think we could add it in the future if needed, right now it doesn't solve an end user problem which is why i'm omitting it

tynes:
I rebased this PR and got the differential fuzzing of withdrawals covered. It uses ethereumjs packages to create pmt proofs. CI is configured to run 512 fuzz runs

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -314,6 +313,81 @@ contract OptimismPortal_Test is Portal_Initializer {
...
+    ) external {
+        uint256 _nonce = messagePasser.nonce();
+
+        (
+            bytes32 stateRoot,
+            bytes32 storageRoot,
+            bytes32 outputRoot,
+            bytes32 withdrawalHash,
+            bytes memory withdrawalProof
+        ) = ffi.finalizeWithdrawalTransaction(
```
maurelian, line 332: IMO this name should be changed to something like `ffi.getFinalizeWithdrawalTransactionInputs` (or less cumbersome-ly `ffi.getFinalizationInputs`). Most of the other `case`s in `differential-testing.ts` are a direct reimplementation of the solidity functions they are named after.

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -314,6 +313,81 @@ contract OptimismPortal_Test is Portal_Initializer {
         vm.expectRevert("L2OutputOracle: No output found for that block number.");
         assertEq(op.isBlockFinalized(checkpoint + 1), false);
     }
+
+    function test_finalizeWithdrawalTransaction_differential(
```
maurelian, line 317: Is correct to call this `_differential`? It looks to me like it's doing differential testing on the `outputRoot` and `withdrawalHash`, but not on a different implementation of `finalizeWithdrawalTransaction`.

tynes: The proofs and output roots are determined by the fuzzer so I would say yes that its a differential test of finalize withdrawal tx

smartcontracts (approved):

PR #2981, opened 2022-07-11
ci: move `op-bindings` check to own job

tynes:
**Description**

This should speed up the CI so that the check for the
`op-bindings` runs in parallel with the `contracts-bedrock`
tests. This step is often very slow compared to others and
I often find myself waiting on it.


.circleci/config.yml
```diff
@@ -143,6 +143,24 @@ jobs:
           name: storage snapshot
           command: yarn storage-snapshot && git diff --exit-code .storage-layout
           working_directory: packages/contracts-bedrock
+
```
smartcontracts, line 146: Don't we also need to remove the old step from `contracts-bedrock-tests`?

maurelian (approved):

.circleci/config.yml
smartcontracts: Oh, lawl. Weird diff. Ok lgtm.

smartcontracts (approved):

PR #2983, opened 2022-07-11
style(ctb): clean up hashing and encoding libs

smartcontracts:
**Description**
Cleans up Hashing and Encoding libraries. Updates natspec and cleans up
function names.


packages/contracts-bedrock/contracts/test/Encoding.t.sol
```diff
@@ -6,8 +6,27 @@ import { Encoding } from "../libraries/Encoding.sol";
 
 contract Encoding_Test is CommonTest {
     function test_nonceVersioning(uint240 _nonce, uint16 _version) external {
-        uint256 nonce = Encoding.addVersionToNonce(uint256(_nonce), _version);
-        uint16 version = Encoding.getVersionFromNonce(nonce);
+        uint256 nonce = Encoding.encodeVersionedNonce(uint256(_nonce), _version);
```
tynes, line 9: I don't love that `encodeVersionedNonce` takes a `uint256` because it can only safely work up to `uint240` - I'd rather be explicit with that with its interface

smartcontracts: That's a really good point.

packages/contracts-bedrock/contracts/libraries/Encoding.sol
```diff
@@ -130,24 +138,33 @@ library Encoding {
...
+     * @param _version Version number to encode into the message nonce.
+     *
+     * @return Message nonce with version encoded into the first two bytes.
      */
-    function addVersionToNonce(uint256 _nonce, uint16 _version)
-        internal
-        pure
-        returns (uint256 nonce)
-    {
+    function encodeVersionedNonce(uint256 _nonce, uint16 _version) internal pure returns (uint256) {
```
tynes, line 148: We could do this as well:

```
    function toUint240(uint256 value) internal pure returns (uint240) {
        require(value <= type(uint240).max, "SafeCast: value doesn't fit in 240 bits");
        return uint240(value);
    }
```

packages/contracts-bedrock/contracts/libraries/Encoding.sol
```diff
@@ -130,24 +138,36 @@ library Encoding {
...
+     *
+     * @return Nonce without encoded version.
+     * @return Version of the message.
      */
-    function getVersionFromNonce(uint256 _nonce) internal pure returns (uint16 version) {
+    function decodeVersionedNonce(uint256 _nonce) internal pure returns (uint240, uint16) {
+        uint240 nonce;
+        uint16 version;
         assembly {
+            nonce := and(_nonce, 0x0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)
```
tynes, line 168: I don't see any test coverage of the nonce value that is returned here?

packages/contracts-bedrock/contracts/libraries/Encoding.sol
```diff
@@ -10,106 +10,114 @@ import { RLPWriter } from "./rlp/RLPWriter.sol";
...
+     * @notice RLP encodes the L2 transaction that would be generated when a given deposit is sent
+     *         to the L2 system. Useful for searching for a deposit in the L2 system.
+     *
+     * @param _from        Address of the sender of the deposit.
+     * @param _to          Address of the receiver of the deposit.
+     * @param _value       ETH value to send to the receiver.
+     * @param _mint        ETH value to mint on L2.
+     * @param _gasLimit    Gas limit to use for the transaction.
+     * @param _isCreation  Whether or not the transaction is a contract creation.
+     * @param _data        Data to send with the transaction.
```
maurelian, line 22: Not directly related to this PR, but these values get passed around together all the time. WDYT on putting them into a struct and treating them as a proper type?

smartcontracts: Strongly agree it should be a type

packages/contracts-bedrock/contracts/libraries/Encoding.sol
smartcontracts: Good point, I can add

tynes (approved):

PR #2986, opened 2022-07-11
ops-bedrock: move docker files into their module dirs

protolambda:
In preparation of the go module versioning work: move the dockerfiles to the dir of their module, and use a temporary `go.work` for the docker build so the modules build with the local monorepo contents.

We still have the `replace github.com/ethereum-optimism/optimism/op-something v0.0.0 => ../op-something` directives in the mod files, but the `use` directives in the new `docker.go.work` files make those unnecessary during the docker build now.
So we can remove them in the next PR that introduces versioning for them.

With a separate toy go workspace project I tested that:
- the go.work file thing in docker works: it's not building with the module version referenced in the `require` of the  `go.mod`, but rather with the local copy referenced by the `docker.go.work`.
- we don't need to copy the go-ethereum replace statement into the `docker.go.work`, the `go.work` is adding/overriding, but not remove directives of the `go.mod`.

### DIFF
--- a/op-batcher/Dockerfile
+++ b/op-batcher/Dockerfile
@@ -2,6 +2,8 @@ FROM golang:1.18.0-alpine3.15 as builder
 
 RUN apk add --no-cache make gcc musl-dev linux-headers git jq bash
 
+# build op-batcher with local monorepo go modules
+COPY ./op-batcher/docker.go.work /app/go.work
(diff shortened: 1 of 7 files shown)


tynes (approved):

tuxcanfly (approved):
OK.

PR #2990, opened 2022-07-12
Naming improvements in Oracle

maurelian:
**Description**

The main changes are: 

1. Rename `sequencer` to `proposer`
3. Use `propose` instead of `append`
2. Use `outputRoot` instead of `l2Output` when referring to the `bytes32` hash of the Output


packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -103,10 +103,10 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
     mapping(uint256 => OutputProposal) internal l2Outputs;
 
     /**
-     * @notice Reverts if called by any account other than the sequencer.
```
tynes, line 106: Perhaps we want to rename the `l2Outputs` mapping as well?

tynes:
What do you think about also renaming `getL2Output` to be more in style?

tynes:
I like the idea here

tynes:
Looks like a new gas snapshot is necessary

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
maurelian: I did play around with that, but didn't find anything I liked more. I considered: 
- `outputProposals`: but they are only proposals until they are finalized then they are outputs
- `outputs`: I do find the `l2` prefix a bit clunky and repetitive in this contract. However `output` is a vague term, and `l2` makes it more readable especially for external devs.

maurelian:
> What do you think about also renaming `getL2Output` to be more in style?

What name do you have in mind? 
FWIW I did try `getOutput` as well as `proposeOutput`, `deleteOutput` etc. See my [comment here](https://github.com/ethereum-optimism/optimism/pull/2990#discussion_r919060561) about removing `l2`.

smartcontracts (approved):
Nice, good changes.

tynes:
@mergifyio refresh

maurelian:
> @mergifyio refresh

What does that do @tynes ?

protolambda (approved):
LGTM. You may need to update the bindings or gas-snapshot later, depending on the merge order with the other PRs in review currently.

tynes (approved):

PR #2991, opened 2022-07-12
contracts-bedrock: better logging in deposit task

tynes:
**Description**

This task is ran in CI and it started failing recently
for unknown reasons. This increases the logging so
that it is easier to understand what is going on
in the script. It detects the case where the deposit
goes through but is undetected by client side hash
computation. It also periodically logs the block number
and hash of the latest block so it can be known if there
has been a reorg.


tynes:
I think that https://github.com/ethereum-optimism/optimism/commit/d55b80a0db6ce1e76a864f4ff40cd922eef7eb38 broke the devnet, cc @trianglesphere @protolambda

packages/contracts-bedrock/tasks/deposits.ts
```diff
@@ -77,26 +81,39 @@ task('deposit', 'Deposits funds onto L2.')
...
+      const postL2Balance = await l2Provider.getBalance(to)
+      if (postL2Balance.gt(preL2Balance)) {
+        console.log(
+          `Unexpected balance increase without detecting deposit transaction`
+        )
+      }
+
+      if (i % 20 === 0) {
+        const block = await l2Provider.getBlock('latest')
+        console.log(`latest block [${block.number}]:${block.hash}`)
```
maurelian, line 113: Maybe print a timestamp here to provide some context within the [wall of logs this outputs](https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/4423/workflows/2f7ab782-08f1-4a7b-a826-2b7ed011acad/jobs/65188/parallel-runs/0/steps/0-106)

trianglesphere:
> I think that [d55b80a](https://github.com/ethereum-optimism/optimism/commit/d55b80a0db6ce1e76a864f4ff40cd922eef7eb38) broke the devnet, cc @trianglesphere @protolambda

It's fully possible. The way the script is run is that the L1 Origin time is set ahead of the L2 starting time (b/c the L1 origin time is set from when the contracts are deployed). Arguably that commit is correct with how we intended the system to run (& how it gets to in a steady state).

packages/contracts-bedrock/tasks/deposits.ts
smartcontracts: Yeah either that or reduce the period between logs

PR #2997, opened 2022-07-12
feat(ctb): include Semver version function

smartcontracts:
**Description**
Updates the Semver contract to include a getter function to retrieve the
full version string.


packages/contracts-bedrock/contracts/universal/Semver.sol
```diff
@@ -35,4 +37,22 @@ contract Semver {
         MINOR_VERSION = _minor;
         PATCH_VERSION = _patch;
     }
+
+    /**
+     * @notice Returns the full semver contract version.
+     *
+     * @return Semver contract version as a string.
+     */
+    function version() public view returns (string memory) {
```
tynes, line 46: Could we get rid of the other getters?

smartcontracts: I guess, any strong reason why?

tynes: Less pollution of the jump table, each extra fn will add gas overhead. The other getters are only useful if something onchain is going to act based on it imo. If they are just human readable, then we can convert to uint8s which should save on codesize + only have 1 getter function so that we don't have to add 4 fns to each contract that inherits this

maurelian: Weakly agree with @tynes here. Approving now anyways.

packages/contracts-bedrock/contracts/universal/Semver.sol
```diff
@@ -9,17 +11,17 @@ contract Semver {
...
      * @notice Contract version number (major).
      */
-    uint256 public immutable MAJOR_VERSION;
+    uint256 private immutable MAJOR_VERSION;
 
     /**
      * @notice Contract version number (minor).
      */
-    uint256 public immutable MINOR_VERSION;
+    uint256 private immutable MINOR_VERSION;
```
tynes, line 19: Will making these uint8s make the bytecode smaller?

packages/contracts-bedrock/contracts/universal/Semver.sol
tynes: slight nit: isn't there `strings.concat`?

packages/contracts-bedrock/contracts/universal/Semver.sol
smartcontracts: Probably, but not significantly.

maurelian (approved):

tynes (approved):

tynes:
@mergifyio refresh

PR #2998, opened 2022-07-12
fix(ctb): minor L1 semver/init cleanup

smartcontracts:
**Description**
Minor L1 contract semver/initialization cleanup after spotting an
inconsistency with ResourceMetering.


packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -62,19 +63,11 @@ contract ResourceMetering {
      */
     uint256[49] private __gap;
```
tynes, line 64: If this is initializable, we don't need the gap?

tynes (approved):

maurelian (approved):

PR #3005, opened 2022-07-13
style(ctb): fix revert messages globally

smartcontracts:
**Description**
Updates revert messages across all contracts to use the standardized
format "ContractName: lowercase message".


packages/contracts-bedrock/contracts/universal/OptimismMintableERC20Factory.sol
```diff
@@ -88,11 +88,6 @@ contract OptimismMintableERC20Factory {
             "OptimismMintableERC20Factory: must provide remote token address"
         );
 
-        require(
```
tynes, line 91: Good catch here

tynes:
Looks like a linting error

maurelian (approved):

trianglesphere:
I'm getting very similar issues when running locally. I don't think it's flaky test for this

PR #3007, opened 2022-07-13
ci: updates version of bash in ci-builder

tynes:
Also deletes leftover copies of solc in the
root of the image. They are copied to the places
where they are expected.

**Description**
Updates bash in the `ci-builder` image, this seems to fix problems with `vm.ffi` in https://github.com/ethereum-optimism/optimism/pull/2980. That PR works on my machine locally and I rebuilt this docker image locally and ran the tests in it and they all passed. In the existing `ci-builder`, the differential tests fail for some reason.

Will need to trigger a release of `ci-builder`

### DIFF
--- a/.changeset/slow-peaches-sit.md
+++ b/.changeset/slow-peaches-sit.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/ci-builder': patch
+---
+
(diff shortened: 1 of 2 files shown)


tuxcanfly (approved):
OK.

PR #3008, opened 2022-07-13
op-node: update engine queue to reset full seq window

protolambda:
Bug found by @trianglesphere

This fix does not re-introduce the old sync find-heads algorithm, since that one would fail and restart on the slightest RPC error, which is bad when it traverses the L2 chain with many requests.

This is a draft, I need to update some batch queue tests, and add tests to the engine queue for this.

Fixes ENG-2471


protolambda:
L2 output submitter go test is failing again. Not sure how related it exactly is. But definitely need to fix the op-proposer thing to enable proposals to get accepted.

protolambda:
Rebased on develop

tynes:
Is this PR still relevant?

protolambda:
@tynes not really, there are some improvements we can make to the current code to handle RPC errors in smaller increments, but I'm fine with the way it is. This can be closed.

PR #3012, opened 2022-07-13
chore(ctb): Remove some todos in messenger and portal

maurelian:
**Description**

1. Removed some todos, I'll provide the rationale in my self-review. 
2. Made error messages conform to style in other contracts.


packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -261,26 +264,29 @@ abstract contract CrossDomainMessenger is
 
         if (_isSystemMessageSender()) {
             // Should never happen.
-            require(msg.value == _value, "Mismatched message value.");
+            require(msg.value == _value, "CrossDomainMessenger: Mismatched message value.");
         } else {
-            // TODO(tynes): could require that msg.value == 0 here
```
maurelian, line 266: Hmmm, I should add a test for this actually.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -261,26 +264,29 @@ abstract contract CrossDomainMessenger is
...
-            require(receivedMessages[versionedHash], "Message cannot be replayed.");
+            require(msg.value == 0, "CrossDomainMessenger: Cannot relay messages with value");
+            require(
+                receivedMessages[versionedHash],
+                "CrossDomainMessenger: Message cannot be replayed."
+            );
         }
 
-        // TODO: Should blocking happen on sending or receiving side?
-        // TODO: Should this just return with an event instead of reverting?
```
maurelian, line 272: Either would be secure, but I think reverting on an invalid action is much more standard.

packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol
```diff
@@ -16,7 +16,6 @@ import { OptimismMintableERC20 } from "../universal/OptimismMintableERC20.sol";
  *         Examples of some token types that may not be properly supported by this contract include,
  *         but are not limited to: tokens with transfer fees, rebasing tokens, and
  *         tokens with blocklists.
- *         TODO: ensure that this has 1:1 backwards compatibility
```
maurelian, line 19: I assume that this just refers to function sigs and storage layouts, which we're already tracking.

tynes (approved):

tynes:
Gas snapshot out of date

PR #3019, opened 2022-07-13
fix (op-proposer): use safe head in block range

tuxcanfly:
op-proposer: use safe head in `BlockRange` response

WIP: adding tests.

Fixes: ENG-2348


protolambda:
@Mergifyio refresh

tuxcanfly:
Closing in favor of #3057

PR #3020, opened 2022-07-13
fix(ctb): clean up compiler warnings

smartcontracts:
**Description**
Cleans up various compiler warnings throughout the package.


packages/contracts-bedrock/contracts/legacy/LegacyERC20ETH.sol
```diff
@@ -23,87 +23,62 @@ contract LegacyERC20ETH is OptimismMintableERC20 {
     {}
 
     /**
+     * @custom:blocked
      * @notice Mints some amount of ETH.
-     *
-     * @param _to     Address of the recipient.
-     * @param _amount Amount of ETH to mint.
      */
-    function mint(address _to, uint256 _amount) public virtual override {
```
tynes, line 31: slight preference for doing something like:

```
function mint( address _to) {
  _to;
}
```

to preserve the abi json

tynes (approved):

packages/contracts-bedrock/contracts/legacy/LegacyERC20ETH.sol
smartcontracts: I think old ABI is mostly unused

tynes:
@mergifyio refresh

PR #3027, opened 2022-07-13
fix(github): fix changesets workflow: run yarn, not uses yarn

protolambda:
I believe this should be `run:` instead of `uses`, since yarn is a command and not an action


tynes (approved):

tynes:
Good catch

smartcontracts (approved):

PR #3032, opened 2022-07-13
tmp: sanity check that latest go module changes are visible to op-e2e

protolambda:
Quick sanity check with CI. Do not merge.


protolambda:
CI panics on the new inserted panic as expected, latest local module changes are visible between modules thanks to go-workspace as expected, even without version updates. Closing now.

PR #3038, opened 2022-07-14
CTB: Pass arguments as calldata in bridge and messengers when possible

maurelian:
This saves about 200 gas when applied to `baseGas`. 

The other changes in the xDM and bridge just feel like a best practice.


packages/contracts/contracts/L1/messaging/L1CrossDomainMessenger.sol
```diff
@@ -136,7 +136,7 @@ contract L1CrossDomainMessenger is
     // slither-disable-next-line external-function
     function sendMessage(
         address _target,
-        bytes memory _message,
```
tynes, line 139: Looks like this is in the wrong repo

maurelian: you mean dir, but yes you're right thx.

packages/contracts-bedrock/contracts/L1/L1StandardBridge.sol
```diff
@@ -264,7 +264,7 @@ contract L1StandardBridge is StandardBridge, Semver {
         address _from,
         address _to,
         uint32 _minGasLimit,
-        bytes memory _extraData
+        bytes calldata _extraData
```
maurelian, line 267: This is how it's being done in the L2 bridge already:

https://github.com/ethereum-optimism/optimism/blob/3f07d1bc2a2a7b0339d839545de57f57702a582d/packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol#L179-L185

smartcontracts (approved):

maurelian:
@mergify refresh

PR #3041, opened 2022-07-14
test(ctb): bridge return to sender on fail

smartcontracts:
**Description**
Adds a test that bridge contracts return to sender on failed deposits or
withdrawals.


packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -391,4 +391,48 @@ contract L1StandardBridge_Test is Bridge_Initializer {
...
+        vm.store(address(L1Bridge), bytes32(slot), bytes32(uint256(100)));
+        assertEq(L1Bridge.deposits(address(BadL1Token), address(L2Token)), 100);
+
+        vm.expectEmit(true, true, true, true);
+
+        emit ERC20BridgeInitiated(
+            address(BadL1Token),
+            address(L2Token),
+            alice,
+            alice,
```
maurelian, line 415: This misses the possibility (which we've gotten wrong previously), that the tokens are returned to the wrong sender on the other domain. I'd suggest this change and the one below:
```suggestion
            bob,
            alice,
```

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -391,4 +391,48 @@ contract L1StandardBridge_Test is Bridge_Initializer {
...
+            address(L1Bridge.messenger()),
+            abi.encodeWithSelector(CrossDomainMessenger.xDomainMessageSender.selector),
+            abi.encode(address(L1Bridge.otherBridge()))
+        );
+        vm.prank(address(L1Bridge.messenger()));
+        L1Bridge.finalizeBridgeERC20(
+            address(BadL1Token),
+            address(L2Token),
+            alice,
+            alice,
```
maurelian, line 430: I think this should be 
```suggestion
            alice,
            bob,
```

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
smartcontracts: Done

maurelian: jc that's fast

maurelian (approved):

tynes (approved):

PR #3042, opened 2022-07-14
op-node: conf depth filter should not filter out genesis

protolambda:
If we are requesting block 0 by number it must not be filtered out because of the confirmation depth filter, the genesis (even if below rollup genesis) always needs to be available as starting point.


trianglesphere (approved):

protolambda:
Added a testcase

tynes (approved):

PR #3044, opened 2022-07-14
Op-Node: Fix divergence due to bad reset logic

trianglesphere:
**Description**
Fixes a divergence bug that occurred whenever the derivation pipeline was reset.
The `progress` of each stage was being set back in history correctly, but the safe head was not
properly updated. This caused a large step back in the L1 origin of L2 blocks & because the pipeline reset
occurred at any time, the nodes would reset at different times & diverge. 

**Additional context**
The simpler version of this commit (without the batch queue reset fix) was tested for a day on goerli without a divergence (up to block 70k for all three replicas, up to block 120k for a pair of nodes).

I'm currently working on testing the change to the batch queue reset step.

**Metadata**
- Fixes ENG-2313


trianglesphere:
Ok, looked into this, on a reset this goes back to the end of a epoch & accidentally produces one more block than it should b/c of a weird interaction of sections 2b of the batch queue & the `FillBatches` code

trianglesphere:
Ok, I put the latest up as a bedrock goerli replica to test it. Currently seems like it's working.

protolambda (approved):

norswap (approved):

PR #3050, opened 2022-07-14
Fix issue related to resolving genesis l2 head

AwaitFuture:
comparison between l1 and l2 time wrong

**Description**
The comparison of the l2 block time and l1 block time is flipped; engine queue has no tests to detect this issue.

### DIFF
--- a/op-node/rollup/derive/engine_queue.go
+++ b/op-node/rollup/derive/engine_queue.go
@@ -297,7 +297,7 @@ func (eq *EngineQueue) ResetStep(ctx context.Context, l1Fetcher L1Fetcher) error
 				Origin: canonicalRef,
 				Closed: false,
 			}
-			if eq.safeHead.Time < canonicalRef.Time {
+			if eq.safeHead.Time > canonicalRef.Time {
 				return fmt.Errorf("cannot reset block derivation to start at L2 block %s with time %d older than its L1 origin %s with time %d, time invariant is broken",
 					eq.safeHead, eq.safeHead.Time, canonicalRef, canonicalRef.Time)
 			}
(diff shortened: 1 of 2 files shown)


op-node/rollup/derive/engine_queue.go
```diff
@@ -297,7 +297,7 @@ func (eq *EngineQueue) ResetStep(ctx context.Context, l1Fetcher L1Fetcher) error
 				Origin: canonicalRef,
 				Closed: false,
 			}
-			if eq.safeHead.Time < canonicalRef.Time {
+			if eq.safeHead.Time > canonicalRef.Time {
```
protolambda, line 300: In Bedrock the L2 time is always equal or higher than the L1 time. Throwing an error if it's higher does not look correct to me.

AwaitFuture: My understanding is that the Safehead is behind the cannon L1 time, since a Safehead needs to be minted to the L1 to be considered Safe. So it appears to me that a SafeHead must necessarily be behind whatever the current L1 time is, since the only way it can be made is for the L1 to step forward.

AwaitFuture: If the comparison is already right then I think the error message is wrong.

`cannot reset block derivation to start at L2 block %s with time %d **older than** its L1 origin %s with time %d, time invariant is broken`

this message is tripped if safehead time is _less than_ the L1 time.

protolambda: canonicalRef is not the L1 head, but the origin that the safe-head is referring to. See above line `canonicalRef, err := l1Fetcher.L1BlockRefByNumber(ctx, eq.safeHead.L1Origin.Number);`

PR #3051, opened 2022-07-15
fix: get events with a smaller block range

### LINKED ISSUE #2653: L1 DTL sync can fail when pointing at nodes that don't cache events
**Describe the bug**
When `DATA_TRANSPORT_LAYER__SYNC_FROM_L1=true`, the dtl can stall and fail to sync new transactions. 
We see the following error in logs:
```
Error: bad response (status=502, headers={"content-length":"0","connection":"close","date":"Sun, 29 May 2022 17:34:38 GMT","via":"kong/2.8.1.0-enterprise-edition"}, body=null, requestBody="{\"method\":\"eth_getLogs\",\"params\":[{\"fromBlock\":\"0xcf7732\",\"toBlock\":\"0xde7af2\",\"address\":\"0xde1fcfb0851916ca5101820a69b13a4e276bd81f\",\"topics\":[\"0x9416a153a346f93d95f94b064ae3f148b6460473c6e82b3f9fc2521b873fcd6c\",\"0x02b616af23339f1e031e76333e2d5b1c3067beb78578c961911872cc2127ef8b\"]}],\"id\":650,\"jsonrpc\":\"2.0\"}", requestMethod="POST", url="https://venasaur.com/", code=SERVER_ERROR, version=web/5.5.1)
```
The RPC request seems to be originating from this [line](https://github.com/ethereum-optimism/optimism/blob/8478377523c6f1480458567af7fe4c52250c7968/packages/data-transport-layer/src/services/l1-ingestion/service.ts#L444). The DTL queries eth_getLogs to retrieve events emitted by the `AddressManager` contract. However, the DTL creates this query with a block range, where the starting block is always the configured `DATA_TRANSPORT_LAYER__L1_START_HEIGHT`. This causes issues because as the JSON-RPC response of such a request would be ginormous, as dtl syncing is close to the tip. In the above log, the block range spans 1 million blocks. This most likely explains the HTTP 5xx error returned by the backend. This problem is compounded as the DTL repeatedly tries to retrieve logs from a large block range during each loop of the l1-ingestion service.

**Desired Outcome**
The DTL shouldn't resync logs starting from `DATA_TRANSPORT_LAYER__L1_START_HEIGHT`. It should remember previous events from prior queries and adjust the start_block `eth_getLog` parameter on successive syncs.

ericlee42:
**Description**

dtl uses a large block range when fetching events, which can cause requests to time out and not synchronize properly

**Additional context**

**Metadata**
- Fix #2653

### DIFF
--- a/packages/data-transport-layer/src/services/l1-ingestion/service.ts
+++ b/packages/data-transport-layer/src/services/l1-ingestion/service.ts
@@ -4,7 +4,7 @@ import { BaseService, Metrics } from '@eth-optimism/common-ts'
 import { TypedEvent } from '@eth-optimism/contracts/dist/types/common'
 import { BaseProvider, StaticJsonRpcProvider } from '@ethersproject/providers'
 import { LevelUp } from 'levelup'
-import { constants } from 'ethers'
+import { constants, Event } from 'ethers'
 import { Gauge, Counter } from 'prom-client'
 
 /* Imports: Internal */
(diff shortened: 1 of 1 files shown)


smartcontracts:
Hey @ericlee42! Since this touches protocol-critical code and much of the team is in Paris for EthCC, it may take a while to review this PR. Thanks for your patience here, and thanks for your contributions to the Optimism codebase!

ericlee42:
ping

PR #3053, opened 2022-07-15
feat(ctb): introduce Types.sol

smartcontracts:
**Description**
Introduces a new library, Types.sol, meant to hold various different
structs that get used all over the place.


packages/contracts-bedrock/contracts/libraries/Types.sol
```diff
@@ -0,0 +1,57 @@
...
+        uint256 mint;
+        uint64 gasLimit;
+        bool isCreation;
+        bytes data;
+        bytes32 l1BlockHash;
+        uint256 logIndex;
+    }
+
+    /**
+     * @notice Struct representing a withdrawal transaction.
```
tynes, line 47: The nonce here is the nonce in the `L2ToL1MessagePasser` correct? Perhaps good to note so that it can be differentiated from a legacy tx. If solidity had type aliases then we could require that only a specific type of nonce is used here

packages/contracts-bedrock/contracts/test/Encoding.t.sol
```diff
@@ -15,15 +16,17 @@ contract Encoding_Test is CommonTest {
 
     function test_encodeDepositTransaction() external {
         bytes memory raw = Encoding.encodeDepositTransaction(
```
tynes, line 18: We need to delete this test because the encoding changed and now we have coverage via differential testing

packages/contracts-bedrock/contracts/test/Hashing.t.sol
```diff
@@ -22,15 +23,17 @@ contract Hashing_Test is CommonTest {
...
-            0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266,
-            0xB79f76EF2c5F0286176833E7B2eEe103b1CC3244,
-            0xde0b6b3a7640000,
-            0xe043da617250000,
-            0x2dc6c0,
-            false,
-            hex"",
-            0xd25df7858efc1778118fb133ac561b138845361626dfb976699c5287ed0f4959,
-            0x1
+            Types.UserDepositTransaction(
```
tynes, line 26: We need to delete this test, we have coverage via differential testing + the ci devnet deposit test covers this

tynes:
Generally looks good to me, can approve after merge conflicts addressed

packages/contracts-bedrock/contracts/test/Hashing.t.sol
tynes: Note that the hashing scheme changed so this test should fail if you rebase

packages/contracts-bedrock/contracts/libraries/Types.sol
maurelian: I think this would achieve what you're [looking for](https://docs.soliditylang.org/en/latest/types.html?highlight=custom%20types#user-defined-value-types)?

packages/contracts-bedrock/contracts/test/Encoding.t.sol
tynes: This test will break once this is rebased, i'm ok with re-adding it in the future

tynes (approved):

packages/contracts-bedrock/contracts/libraries/Types.sol
tynes: > I think this would achieve what you're [looking for](https://docs.soliditylang.org/en/latest/types.html?highlight=custom%20types#user-defined-value-types)?

Yes it could - I do think introducing type safety for certain interfaces is a good idea

maurelian (approved):

PR #3054, opened 2022-07-15
op-node: fix p2p static peers flag env var

protolambda:
Found by @trianglesphere and @optimisticben :clap:


optimisticben (approved):
LGTM

PR #3056, opened 2022-07-15
op-e2e: deploy output oracle with correct config (bedrock)

protolambda:
This PR:
- deploys output oracle with correct config
- de-duplicates block time and genesis setup info.

Previously the contract was being deployed with a block time of 2 instead of 1, and thus failing to confirm the latest outputs, and increasingly lagging behind the L2 chain.


trianglesphere (approved):
tysm for tracking this down

PR #3057, opened 2022-07-15
fix (op-proposer): use safe head in block range

tuxcanfly:
**Description:**

Currently the proposer does not use a safe head for creating output proposal which causes a failure in proposing with: `L2OutputOracle: cannot propose L2 output in the future`, as the output oracle does not know the unsafe head.

To fix this, we limit the block range the output proposer works with to the safe head instead.

Edit: does not include a test atm.

Fixes: ENG-2348

### DIFF
--- a/op-proposer/drivers/l2output/driver.go
+++ b/op-proposer/drivers/l2output/driver.go
@@ -109,7 +109,12 @@ func (d *Driver) GetBlockRange(
 		d.l.Error(name+" unable to get next block number", "err", err)
 		return nil, nil, err
 	}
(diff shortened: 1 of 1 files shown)


protolambda (approved):

tuxcanfly:
Blocked on testing this. `op-e2e` doesn't expose much of the proposer except `Start` / `Stop`. Since the logs show the same error with `eth_estimateGas` I thought it would be an alternative way to test this. But it seems that the error is only logged, the geth client doesn't return any error.

tuxcanfly:
Removing the test as it's not working as expected. Looking to merge this fix now and add the test later.

trianglesphere (approved):
Lgtm

PR #3059, opened 2022-07-16
contracts-bedrock: bump to use solidity 0.8.15

tynes:
**Description**
Bumps the bedrock contracts to use the latest version of `solc`. Some tests have slight gas increases and some have slight gas decreases.


packages/contracts-bedrock/contracts/L1/L1CrossDomainMessenger.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.9;
+pragma solidity ^0.8.15;
```
maurelian, line 2: I think we should [pin the version](https://swcregistry.io/docs/SWC-103):
```suggestion
pragma solidity 0.8.15;
```

This removes any doubt about which version is being used to compile at any given time.

tynes: Will that create problems for people that import? Would need interfaces for things we expect to be public interfaces

maurelian: Hmmm... honestly I haven't considered that. Is that really necessary even just for interfaces? If that's the case then we need to go with the caret

maurelian (approved):

packages/contracts-bedrock/contracts/L1/L1CrossDomainMessenger.sol
tynes: Using the caret will only solve people importing the code if they are using a newer solidity version. I'm down to pin to the exact version and then we can automate creation of interfaces that have a lower version. People that want to interact with our on chain APIs (messengers, bridge, portal) can use the interfaces which will be pegged to a lower version.

tynes:
Libraries/interfaces are `^0.8.0` and everything else is `0.8.15`

PR #3060, opened 2022-07-16
contracts-bedrock: add size snapshot

tynes:
**Description**
To track the sizes of the contracts over time


tynes:
I'd like to modify this such that it doesn't require the artifact to be committed into the repo but instead writes the artifact to the CircleCI artifact store

PR #3065, opened 2022-07-19
fix(sdk): throw a better error for bad chain ID

smartcontracts:
**Description**
Updates the SDK to throw a better error when the user provides a bad
chain ID. Would've helped an end-user more easily debug an issue where
the chain ID parameter was missing.


smartcontracts:
Need to fix a few bugs...

tynes (approved):

maurelian (approved):

PR #3066, opened 2022-07-19
feat(ctb): introduce extra data event to XDM

smartcontracts:
**Description**
Introduces a new SentMessageExtraData event to the CrossDomainMessenger
base contract. We need this event so that we can properly find all
relevant message data in the SDK. SDK cannot function without this
event. We also cannot add to the old SentMessage event without breaking
the ABI of our contract, so we're stuck adding an extra data event.


packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -44,6 +44,15 @@ abstract contract CrossDomainMessenger is
...
     );
 
+    /**
+     * @notice Additional event data to emit, required as of Bedrock. Cannot be merged with the
+     *         SentMessage event without breaking the ABI of this contract, this is good enough.
+     *
+     * @param sender Address of the sender of the message.
+     * @param value  ETH value sent along with the message to the recipient.
+     */
+    event SentMessageExtraData(address indexed sender, uint256 value);
```
maurelian, line 54: When I read the description, I assumed that this event would have something to do with the `extraData` argument in the Standard Bridge, or at least include some arbitrary data. Maybe a better name would be `SentMessageValue`.

smartcontracts: Maybe `SentMessageAugmentation1`? We may end up having to add more of these in the future if we end up changing stuff...

smartcontracts: I think for the sake of not making significant changes to this contract in the final hours of audit prep, I'd prefer `SentMessageExtension1`. My hope is that we won't need to add more data later. If we realize we need to add more data in a future release, we can consider looking at a more future-proof model.

PR #3067, opened 2022-07-19
feat(ctb): export ABI when building

smartcontracts:
**Description**
Updates the TypeScript build to also generate ABI objects that can be
imported on the client-side. Specifically necessary so that these ABIs
can be imported on the frontend where fs/glob is not available.


tynes (approved):
Looks good to me, we could optimize the deps a bit in the future but for now this is good

mslipper (approved):

roninjin10:
This is very ugly

PR #3075, opened 2022-07-21
Fix Optimism Config in L2 Genesis builder

trianglesphere:
**Description**
This places the Optimism chain config in the correct place as
well as deleting the spurious Clique config in the L2 genesis.

**Metadata**
- Fixes the inability of op-nodes to direct a reorg on the L2 execution engine.


packages/core-utils/src/external/geth/geth-types.ts
```diff
@@ -38,6 +38,7 @@ export interface ChainConfig {
   grayGlacierBlock?: number
   mergeNetsplitBlock?: number
   terminalTotalDifficulty?: number
+  // TODO: Make this an alternation (one xor other) of clique of optimism config
```
trianglesphere, line 41:
```suggestion
```

tynes (approved):

protolambda (approved):

PR #3085, opened 2022-07-22
ops: Use starting L1 Block for timestamp everywhere

trianglesphere:
**Description**
This transitions the starting timestamp to a new flow. The L2 rollup
is anchored on a L1 block. The L2 genesis block & rollup config use
the timestamp of the L1 start block as the their time. Properly
threading this through the HH tasks is a little tricky but possible.
This is because we have two flows: creating a L1 network & placing
the rollup on that and creating a rollup on an existing L1 network
(like goerli). There is still a L1 starting time for the first flow.

This also fixes a circular dependcy that previously existed. The
starting timestamp was provided and served as the starting timestamp
for the L1 genesis & the "L2 Starting Time" in the L2 Output Oracle.
The actual L2 genesis & rollup start time were based on when the
Optimism Portal contract was deployed (after the L2 Output Oracle
contract must have been deployed). The rollup is resilient to being
started before contracts are fully deployed, so using a specific
L1 block as the start is the cleanest solution I have seen.

**Additional context**
This is part of work to improve automation of devnet deployments.

I have confirmed that this works with the local devnet. It needs testing
with the goerli devnet.

**Metadata**
- Fixes ENG-2473


mslipper (approved):

protolambda:
There's a linting issue and I'd like an ACK from @tynes to change the bedrock hardhat deploy code, but LGTM otherwise. And it looks like I can use this in the Hive setup code.

optimisticben (approved):
lgtm

mslipper:
Fixed lint. Added `do-not-merge` until @tynes acks. Mark - if everything looks good to you just remove the label and this will auto-merge.

packages/contracts-bedrock/deploy/000-L2OutputOracle.deploy.ts
```diff
@@ -10,14 +10,8 @@ const deployFn: DeployFunction = async (hre) => {
...
-  if (
-    typeof deployConfig.startingTimestamp !== 'number' ||
-    isNaN(deployConfig.startingTimestamp)
-  ) {
-    throw new Error(
-      'Cannot deploy L2OutputOracle without specifying a valid startingTimestamp.'
-    )
-  }
+  const l1 = hre.ethers.getDefaultProvider()
+  const l1StartingBlock = await l1.getBlock(deployConfig.l1StartingBlockTag)
```
tynes, line 14: Its possible that the `l1StartingBlock` is `null`, probably want to handle that case

packages/contracts-bedrock/deploy-config/goerli.ts
```diff
@@ -1,15 +1,15 @@
 import { ethers } from 'ethers'
 
 const sequencerAddress = '0x6c23a0dcdfc44b7a57bed148de598895e398d984'
-const startingTimestamp = 1658777882
+const l1StartingBlockTag = ''
```
tynes, line 4: Why is this empty? Should we find the tag that corresponds with the previous `startingTimestamp`?

tynes: `cast find-block <timestamp>`

tynes (approved):
A couple nits that are easy to fix

packages/contracts-bedrock/deploy-config/goerli.ts
trianglesphere, line 4:
```suggestion
const l1StartingBlockTag = '0xafce66a0a2446856112e4069b275ad32b1f4a607888f9c4c59eddf9be81f8670'
```

packages/contracts-bedrock/deploy/000-L2OutputOracle.deploy.ts
trianglesphere: addressed at all usages

packages/contracts-bedrock/deploy-config/goerli.ts
trianglesphere: fixed (had to use block hash from k8s repo)

packages/contracts-bedrock/tasks/rollup-config.ts
```diff
@@ -23,13 +23,16 @@ task('rollup-config', 'create a genesis config')
     const l2Genesis = await l2.getBlock('earliest')
 
     const portal = await hre.deployments.get('OptimismPortalProxy')
-    const l1StartingBlock = await l1.getBlock(portal.receipt.blockHash)
+    if (deployConfig.l1StartingBlockTag === undefined) {
+      throw new Error('must provide L1 Starting Block Tag')
+    }
+    const l1StartingBlock = await l1.getBlock(deployConfig.l1StartingBlockTag)
```
tynes, line 29: I think what we are looking for is something like this:

```typescript
if (l1StartingBlock === null) {
  throw new Error(`Cannot fetch block tag ${deployConfig.l1StartingBlockTag}`)
}

```

tynes: This same check should be added in each place where the `l1.getBlock` is called

PR #3086, opened 2022-07-24
feat(sdk): add Bedrock support to SDK

smartcontracts:
**Description**
Updates the SDK to add support for Bedrock. Includes a minor fix to the
build pipeline for the contracts-bedrock package. Removes some utilities
that are now exported as part of core-utils.


packages/sdk/src/adapters/eth-bridge.ts
```diff
@@ -178,7 +178,10 @@ export class ETHBridgeAdapter extends StandardBridgeAdapter {
           amount,
           0, // L1 gas not required.
           '0x', // No data.
-          opts?.overrides || {}
+          {
+            ...omit(opts?.overrides || {}, 'value'),
```
roninjin10, line 182: nit: the omit here is unnecessary it's being overwritten on line 183

packages/sdk/src/adapters/eth-bridge.ts
```diff
@@ -187,7 +190,10 @@ export class ETHBridgeAdapter extends StandardBridgeAdapter {
...
```
roninjin10, line 194: nit: omit unnecessary

roninjin10 (approved):
lgtm couple of nits but non blocking

mslipper (approved):

mslipper:
This looks good to me. @tynes if you want another set of eyes please have them review, then remove `mergify-ignore`.

packages/sdk/src/adapters/eth-bridge.ts
tynes: Going to not change this as to be extra sure no behavior is changed in the old sdk version

tynes: This will need to be refactored once its only bedrock support anyways

PR #3088, opened 2022-07-25
chore: Fix foundry config to use profiles key

maurelian:
Addresses this warning: 

```
Unknown section [default] found in foundry.toml. This notation for profiles has been deprecated and may result in the profile not being registered in future versions. Please use [profile.default] instead or run `forge config --fix`.
```


tynes:
We can update the version of foundry in the `ci-builder` docker image which should fix the broken `yarn-monorepo` build

maurelian:
> We can update the version of foundry in the `ci-builder` docker image which should fix the broken `yarn-monorepo` build

Just to whatever is latest?

tynes:
Just released a new `ci-builder` image, potentially will unblock, otherwise will need to release a foundry image then another `ci-builder`

tynes (approved):

tynes:
We should open a new PR that fixes the bindings build

protolambda (approved):

PR #3102, opened 2022-07-26
chore: Commit formatting changes from vscode startup

maurelian:
**Description**

VSCode creates this diff on startup. 
One is a formatting change to `.vscode/settings.json`.
The other adds a second version of the `BurntSushi/toml` package to `go.work.sum`.


trianglesphere (approved):
I don't use vscode so I don't know much about it's settings, but the go.work.sum lgtm

tynes (approved):

PR #3103, opened 2022-07-26
Bedrock goerli 1948678e

optimisticben:
**Description**
Bedrock goerli 1948678e


tynes:
Does this PR contain the live contracts?

optimisticben:
> Does this PR contain the live contracts?

Yeah, but I wasn't going to commit until we're happy with the stability.

tynes:
Going to close this PR as I believe it is no longer relevant after we have iterated on devnets

PR #3104, opened 2022-07-27
contracts-bedrock: fix `OptimismMintableERC20Created`

tynes:
**Description**

The names of the event args were backwards from what they
were supposed to be. I noticed this while writing integration
tests for the sdk.

The event is used like this:

```solidity
emit OptimismMintableERC20Created(_remoteToken, address(localToken), msg.sender);
```

The event was updated to look like this:

```solidity
event OptimismMintableERC20Created(
    address indexed remoteToken,
    address indexed localToken,
    address deployer
);
```

The `remoteToken` is first and the `localToken` is second.


maurelian:
Can we move this [here](https://github.com/ethereum-optimism/optimism-audit-fixes) so that it can get reviewed along with audit fixes?

maurelian:
I moved your commit to [here](https://github.com/ethereum-optimism/optimism-audit-fixes/pull/2), so I think we can close this for now.

packages/contracts-bedrock/contracts/universal/OptimismMintableERC20Factory.sol
```diff
@@ -32,13 +32,13 @@ contract OptimismMintableERC20Factory {
...
      *
-     * @param localToken  Address of the created token on the local chain.
      * @param remoteToken Address of the corresponding token on the remote chain.
+     * @param localToken  Address of the created token on the local chain.
      * @param deployer    Address of the account that deployed the token.
      */
     event OptimismMintableERC20Created(
-        address indexed localToken,
         address indexed remoteToken,
+        address indexed localToken,
```
maurelian, line 41: Is it better to change the ordering here in the definition, or where it's emitted?

https://github.com/ethereum-optimism/optimism/blob/761ab09335b0a1496d05a6e9d4ea9c41432198e1/packages/contracts-bedrock/contracts/universal/OptimismMintableERC20Factory.sol#L99

Could this possible cause issues for existing tools Quixotic has written?

maurelian:
@tynes I think this got stale because it was moved into the fixes repo temporarily. I'm reopening it now, and will push a rebase shortly.

packages/contracts-bedrock/contracts/universal/OptimismMintableERC20Factory.sol
maurelian: Coming back to this, I see that the ordering in the `StandardBridge` matches the current definition rather than this change, and I think that we should standardize on `local` then `remote` ordering. cc @sam-goldman as we should make sure the [NFT events](https://github.com/ethereum-optimism/optimism/blob/2b273136f7e3541b246f35e38d009d1336862844/packages/contracts-periphery/contracts/universal/op-erc721/OptimismMintableERC721Factory.sol#L21) match that.

https://github.com/ethereum-optimism/optimism/blob/9bb9fb2835551fbf4771b4489be2bdf3a271de0b/packages/contracts-bedrock/contracts/universal/StandardBridge.sol#L111-L137

sam-goldman: Fixed this [here](https://github.com/ethereum-optimism/optimism/pull/3491)

smartcontracts (approved):

maurelian:
Need to do a bit of debugging to figure out why this broke the deposit-erc20 integration test

tynes:
See this comment for fixing the tests: https://github.com/ethereum-optimism/optimism/blob/4a5e1832686f1375be3fb1b20768a67625455d68/packages/sdk/tasks/deposit-erc20.ts#L80

mslipper (approved):

PR #3110, opened 2022-07-27
proxyd: Parameterize full RPC request logging

mslipper:
Adds two new config params:

- `enable_request_log`, which enables/disables the full RPC request log
- `max_request_body_log_len`, which configures the maximum number of characters that will be logged for each RPC request. Values over 2000 are ignored.

This should reduce our GCP spend.


Inphi (approved):

optimisticben (approved):
lgtm

PR #3111, opened 2022-07-27
test(ctb): change gasLimit arg to uint256

maurelian:
Failing test case decodes to:
```
address _sender =  0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84;
address _target = 0x028188030556FaF8383A2BA80d95E1415C6Ed6F6;
uint64 _value = 6060582089520090887;
uint8 _gasLimit = 164;
bytes memory _data = hex"945a53f5ffffffffffffffff0000000000000000000000000000000000000000";
```

Caveats:
1. I'm really now sure how any of the tests passed with a maximum input of 255 for `_gasLimit`
2. Even with this fix, if I locally do 512 runs, I still get a failure case with the following values: 

```
address _sender =  0x0000000000000000000000000000000000000000;
address _target = 0x0000000000000000000000000000000000000000;
uint64 _value = 0;
uint256 _gasLimit = 8937393460516492772;
bytes memory _data = hex"0101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101";
```

Conclusion so far: it's got something to do with gasLimit.


packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -526,8 +526,8 @@ contract OptimismPortal_FinalizeWithdrawal_Test is Portal_Initializer {
     function test_finalizeWithdrawalTransaction_differential(
         address _sender,
         address _target,
-        uint64 _value,
-        uint8 _gasLimit,
```
maurelian, line 530: The values in the actual transaction type are both uint256, but I get a failure when I set `_gasLimit` to `uint256`. I assume because the `_gasLimit` arg is just too high for the vm?
https://github.com/ethereum-optimism/optimism/blob/ad7fe27b540c173ae7f0ef95e069bebfd007f5b4/packages/contracts-bedrock/contracts/libraries/Types.sol#L53-L54

tynes: It has to do with the gaslimit check here: https://github.com/ethereum-optimism/optimism/blob/915f3b28791c543961cc34bd16a3732b52637f9e/packages/contracts-bedrock/contracts/L1/OptimismPortal.sol#L191

tynes: Now that I think about it, I think the solution is to control the amount of gas during this call:

```solidity
op.finalizeWithdrawalTransaction{ value: _tx.value, gas: _tx.gasLimit + FINALIZE_GAS_BUFFER + n }(...)
```
Something like that?

maurelian: That might work... but I assumed that forge would always provide enough gas if it could, so I thought we might be bumping up against some upper bound in what the VM could accept for gasLimit.

tynes: I think controlling the gas limit passed to the call is more realistic because right now we have no idea what the `gaslimit()` value is returning, I believe it should be considered best practice to pass in a gaslimit to a `CALL` if it uses `gasleft()`

maurelian: Is [this](https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/5105/workflows/df6628a9-bd29-458a-b09e-13d42aa0d652) what you mean? I keep using larger `n` and it's still failing for me.

mslipper:
Denet should be fixed after a rebase.

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
maurelian: still looking at this, its weird.

maurelian: OK, so after a bunch of testing, I'd like to keep it as is without specifying the gas amount to pass to the call. The reason being that: 
- This amount works: `_tx.gasLimit + 20_000 + 1_000_000`
- This amount fails: `_tx.gasLimit + 20_000 + 500_000`

In the second case, the reason for the failure just seems to be an inaccurate estimation of how much gas actually needs to be provided to the transaction (I verified that there is no difference in actual gas usage between the two cases). 

For this reason, not specifying the gas amount feels much less brittle to me.

tynes: Ok i'm cool with the implementation that you have. It feels like a todo still to figure out the exact `FINALIZE_GAS_BUFFER`

tynes (approved):

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
maurelian: Agree. It's tracked [here](https://linear.app/optimism/issue/ENG-2521/evaluate-and-document-the-purpose-the-relay-gas-buffer-check-in).

mslipper (approved):

PR #3113, opened 2022-07-28
ci: devnet foundry version

tynes:
**Description**

Pin foundry to `64fe4ac` using the command:

```bash
$ foundryup -C 64fe4ac
```

This will require foundry to be compiled from scratch instead
of downloading a github release. We can add caching later as
well as a better scheme for controlling the versioning.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -434,6 +434,7 @@ jobs:
       - run:
           name: foundryup
           command: |
+            curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh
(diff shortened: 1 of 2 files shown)


tynes:
Added a quick readme to `ops-bedrock` so that the ci for `devnet` would trigger

mslipper (approved):

tynes:
I don't love this because it causes the devnet task to take 20 mins

tynes:
Since we fixed the build issue in another PR, going to close this until we have a faster impl that pins foundry

PR #3114, opened 2022-07-28
contracts-bedrock: fix build

tynes:
**Description**

Standardize on the upgradable initializable since the contracts
are technically upgradable it is more clear. There are no real
implementation differences between the upgradable and standard
initializable implementations. `OwnableUpgradable` is `initializable`
imported from the upgradable package, and the `L2OutputOracle`
inherits from `OwnableUpgradable`. This means that the only
way to standardize on a single implementation of `Initializable`
is to use the upgradable version.

This also bumps the version of the openzeppelin contracts
dependency because they refactored the initializable implementation
and made it easier to understand.

https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/releases/tag/v4.7.0
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3450


maurelian (approved):

maurelian:
Bindings need a rebuild

mslipper (approved):

PR #3115, opened 2022-07-28
pkg derive: wrap errors by severity

tuxcanfly:
**Description**
This PR wraps errs in the derive package to include severity, which can be used by callers to optionally retry and backoff with timeout or reset the pipeline or halt.

**Additional context**
This addresses stability and some tech debt in the derive package. The goal is to have a single place where errors can be handled based on severity and retried with backoff, if possible.

**Metadata**
- Fixes #ENG-2446


tuxcanfly:
Merged with #3130

PR #3116, opened 2022-07-28
op-bindings: speed up build

tynes:
**Description**
Speeds up the `op-bindings` build as foundry doesn't need to recompile the entire set of contracts if you give it the fully qualified name instead of just the name. I noticed significant speed ups with this.

Building in parallel seems to fail due to contention on downloading the compiler: https://github.com/ethereum-optimism/optimism/pull/3040


tynes:
This halves the amount of time it takes to run the build op bindings ci step

mslipper (approved):

PR #3119, opened 2022-07-28
contracts-bedrock: in place handling of immutables

tynes:
**Description**

This updates the hardhat task `genesis-l2` to handle
setting the immutables in the deployed bytecode. This
is necessary because the predeploys "live" in the L2
state and are not deployed. Any immutables must be manually
placed into the bytecode by looking at the compiler output
and then finding the offsets and slicing in the values.

A high level API is exposed for setting arbitrary immutables
with the `replaceImmutables` function. This was inspired by
smock's `computeStorageSlots`.

Longer term, this code will be rewritten in go and be able to
operate on either a LevelDB database (upgrade an existing system)
or a `genesis.json` (good for starting a new network). This is
a requirement because the state surgery is large enough such that
it must bind directly to LevelDB and be in a language that is
relatively fast.

### DIFF
--- a/packages/contracts-bedrock/tasks/check-l2-config.ts
+++ b/packages/contracts-bedrock/tasks/check-l2-config.ts
@@ -22,6 +22,7 @@ task('check-l2-config', 'Validate L2 config')
     )
 
     const bridge = await OptimismMintableERC20Factory.bridge()
+    console.log(`OptimismMintableERC20Factory.bridge() -> ${bridge}`)
     if (bridge !== predeploys.L2StandardBridge) {
       throw new Error(
         `L2StandardBridge not set correctly. Got ${bridge}, expected ${predeploys.L2StandardBridge}`
--- a/packages/contracts-bedrock/tasks/check-l2-config.ts
+++ b/packages/contracts-bedrock/tasks/check-l2-config.ts
@@ -1,0 +1,30 @@
+import { task, types } from 'hardhat/config'
+import { providers } from 'ethers'
+import '@nomiclabs/hardhat-ethers'
+
+import { predeploys, getContractInterface } from '../src'
+
+task('check-l2-config', 'Validate L2 config')
+  .addParam(
+    'l2ProviderUrl',
+    'L2 provider URL.',
+    'http://localhost:9545',
+    types.string
+  )
+  .setAction(async (args, hre) => {
+    const { l2ProviderUrl } = args
+    const l2Provider = new providers.JsonRpcProvider(l2ProviderUrl)
+
+    const OptimismMintableERC20Factory = new hre.ethers.Contract(
+      predeploys.OptimismMintableERC20Factory,
+      getContractInterface('OptimismMintableERC20Factory'),
+      l2Provider
+    )
+
+    const bridge = await OptimismMintableERC20Factory.bridge()
+    if (bridge !== predeploys.L2StandardBridge) {
+      throw new Error(
+        `L2StandardBridge not set correctly. Got ${bridge}, expected ${predeploys.L2StandardBridge}`
+      )
+    }
+  })
--- a/packages/contracts-bedrock/tasks/genesis-l2.ts
+++ b/packages/contracts-bedrock/tasks/genesis-l2.ts
@@ -5,11 +5,18 @@ import assert from 'assert'
 import { OptimismGenesis, State } from '@eth-optimism/core-utils'
 import 'hardhat-deploy'
 import '@eth-optimism/hardhat-deploy-config'
-import { ethers } from 'ethers'
+import { ethers, utils, BigNumber } from 'ethers'
 import { task } from 'hardhat/config'
 import { HardhatRuntimeEnvironment } from 'hardhat/types'
+import {
+  CompilerOutputSource,
+  CompilerOutputContract,
+  BuildInfo,
+} from 'hardhat/types/artifacts'
 
 import { predeploys } from '../src'
+
+const { hexZeroPad, hexConcat, hexDataSlice, getAddress } = utils
 
 const prefix = '0x420000000000000000000000000000000000'
 const implementationSlot =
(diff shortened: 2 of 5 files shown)


packages/contracts-bedrock/tasks/genesis-l2.ts
```diff
@@ -251,6 +369,42 @@ task('genesis-l2', 'create a genesis config')
...
+        MAJOR_VERSION: toBytes32(0),
+        MINOR_VERSION: toBytes32(0),
+        PATCH_VERSION: toBytes32(1),
+      },
+      L2ToL1MessagePasser: {
+        MAJOR_VERSION: toBytes32(0),
+        MINOR_VERSION: toBytes32(0),
+        PATCH_VERSION: toBytes32(1),
+      },
+      SequencerFeeVault: {
```
maurelian, line 401: Should this PR also make [this value](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/contracts/L2/SequencerFeeVault.sol#L24) and set it's value?

tynes: I didn't want to do so because it was a contract implementation change and the audit is going on

packages/contracts-bedrock/tasks/genesis-l2.ts
```diff
@@ -46,6 +58,120 @@ const getStorageLayout = async (
...
+  const contracts = buildInfo.output.contracts
+
+  const compilerOutputContracts: CompilerOutputContract[] = []
+  for (const [contractName, contract] of Object.entries(contracts)) {
+    if (path.basename(contractName, '.sol') === name) {
+      compilerOutputContracts.push(contract[name])
+    }
+  }
+  if (compilerOutputContracts.length !== 1) {
+    console.log(`Unexpected number of contracts for ${name}`)
```
maurelian, line 73: This is kind of a confusing error message if the number is 0.

tynes: You are right, this should throw if none are found

tynes: Pushed a fix for this

packages/contracts-bedrock/tasks/genesis-l2.ts
```diff
@@ -46,6 +58,125 @@ const getStorageLayout = async (
...
+    buildInfo
+  )
+
+  // Get the immutable references. They look like this:
+  // { ast-id: [ {start, length} ] }
+  const immutableReferences =
+    outputContract.evm.deployedBytecode.immutableReferences
+
+  const names = {}
+  // Recursively fine all of the immutables by traversing the solc output ast
```
maurelian, line 116: nit: find

packages/contracts-bedrock/tasks/genesis-l2.ts
```diff
@@ -46,6 +58,125 @@ const getStorageLayout = async (
...
+    }
+  }
+
+  findNames(outputSource.ast)
+
+  let deployedBytecode = artifact.deployedBytecode
+  const presize = deployedBytecode.length
+
+  // For each of the immutables, put the value into the bytecode
+  for (const [key, value] of Object.entries(immutables)) {
```
maurelian, line 150: Could we first check that the value is `0`?

packages/contracts-bedrock/tasks/genesis-l2.ts
```diff
@@ -262,17 +429,28 @@ task('genesis-l2', 'create a genesis config')
       const allocAddr = name === 'LegacyERC20ETH' ? addr : toCodeAddr(addr)
```
maurelian, line 429: Because this is the only one not behind a proxy?

packages/contracts-bedrock/tasks/genesis-l2.ts
maurelian: Actually, I don't think this would require a change to the solidity?

packages/contracts-bedrock/tasks/check-l2-config.ts
```diff
@@ -0,0 +1,31 @@
...
+    )
+
+    const bridge = await OptimismMintableERC20Factory.bridge()
+    console.log(`OptimismMintableERC20Factory.bridge() -> ${bridge}`)
+    if (bridge !== predeploys.L2StandardBridge) {
+      throw new Error(
+        `L2StandardBridge not set correctly. Got ${bridge}, expected ${predeploys.L2StandardBridge}`
+      )
+    }
+  })
```
maurelian, line 31: Should this also check the semvers? Find to add in a future PR.

maurelian (approved):
Some comments worth addressing, but LGTM.

packages/contracts-bedrock/tasks/check-l2-config.ts
tynes: Was thinking of doing that in the future, thinking of how to be more robust with it

packages/contracts-bedrock/tasks/genesis-l2.ts
tynes: Yes that is why

packages/contracts-bedrock/tasks/genesis-l2.ts
tynes: The value can be anything - its the value of the immutable

maurelian: But what is being overwritten?

tynes: Oh i see what you mean, check that the immutable was set to 0 in the bytecode itself

packages/contracts-bedrock/tasks/check-l2-config.ts
maurelian: Yeah, it seems like this task will naturally expand, and I'm OK with it not being exhaustive now.

packages/contracts-bedrock/tasks/genesis-l2.ts
maurelian: Yeah, since the initcode never runs, that should always be the case right?

maurelian:
Thanks for adding this check:  https://github.com/ethereum-optimism/optimism/pull/3119/commits/1124f30caa7f19cd6f975afcc8481e80116f7be5#diff-863a227d497b7afe88109084bf749cc3ee7b8f9674c4d47fb101ee264933b434R168-R176

PR #3122, opened 2022-07-28
makefile: `clean-node-modules`

tynes:
**Description**

Adds a helper to the makefile for deleting all of the js
deps. This is useful for when checking out a new branch
and ensuring that the deps are up to date. I find
myself doing this often manually, so adding a make command
makes sense.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,3 +93,7 @@ semgrep:
 	$(eval DEV_REF := $(shell git rev-parse develop))
 	SEMGREP_REPO_NAME=ethereum-optimism/optimism semgrep ci --baseline-commit=$(DEV_REF)
 .PHONY: semgrep
+
+clean-node-modules:
+	rm -rf node_modules
+	rm -rf packages/**/node_modules


maurelian (approved):

Makefile
```diff
@@ -93,3 +93,7 @@ semgrep:
 	$(eval DEV_REF := $(shell git rev-parse develop))
 	SEMGREP_REPO_NAME=ethereum-optimism/optimism semgrep ci --baseline-commit=$(DEV_REF)
 .PHONY: semgrep
+
+clean-node-modules:
+	rm -rf node_modules
```
trianglesphere, line 98: add `yarn clean`? `yarn clean` does more than just node modules

trianglesphere (approved):

Makefile
tynes: Slight preference for not but it does make sense to do a full clean and rebuild after this

tynes: Would approve a follow up PR

PR #3124, opened 2022-07-28
core-utils,contracts-bedrock,ci: OpNodeProvider + check

tynes:
**Description**

Implements a simple `OpNodeProvider` that can be used
to query the op-node. It is used as part of a ci check
against the devnet. It should add some more health related
things, now it just prints of rpc responses. This will
help to debug things.


packages/contracts-bedrock/tasks/check-op-node.ts
```diff
@@ -0,0 +1,20 @@
+import { task, types } from 'hardhat/config'
+import { OpNodeProvider } from '@eth-optimism/core-utils'
+
+// TODO(tynes): add in config validation
+task('check-op-node', 'Deposits funds onto L2.')
```
trianglesphere, line 5: old task name?

tynes: Ah the description, good catch

tynes:
I don't think L2 ever finalizes based on the output of the CI here:

```
 "finalizedL2": {
    "hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "number": {
      "type": "BigNumber",
      "hex": "0x00"
    },
    "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "timestamp": {
      "type": "BigNumber",
      "hex": "0x00"
    },
    "l1Origin": {
      "hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "number": {
        "type": "BigNumber",
        "hex": "0x00"
      }
    },
    "sequencerNumber": {
      "type": "BigNumber",
      "hex": "0x00"
    }
  }

```

trianglesphere:
> I don't think L2 ever finalizes based on the output of the CI here:

That is correct. We don't set an L2 finalization block.

trianglesphere (approved):

mslipper (approved):

PR #3125, opened 2022-07-28
op-node: Isolated channel frame serialization

trianglesphere:
**Description**

This adds a new `Frame` struct to represent a channel frame as well as serialization &
deserialization methods on that struct. The serialization is based on a reader/writer API
rather than a bytes buffer API. For the deserialization this is a more natural fit given that
multiple frames can be present in a single buffer. On the writing side some of the accounting
around where to place bytes is a bit easier.

This does not change the frame data format.


op-node/rollup/derive/channel_bank.go
```diff
@@ -81,85 +81,73 @@ func (ib *ChannelBank) IngestData(data []byte) error {
 	if data[0] != DerivationVersion0 {
 		return fmt.Errorf("unrecognized derivation version: %d", data)
 	}
+	var buf bytes.Buffer
+	buf.Write(data[1:])
```
protolambda, line 85: Maybe just avoid copying the complete input with something like `input := bytes.NewReader(data[1:])`? This provides the same byte-reading methods as the bytes buffer, without copy.

op-node/rollup/derive/channel_frame.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return fmt.Errorf("error reading Id: %w", err)
+	}
+	f.Id.Time, err = binary.ReadUvarint(r)
+	if err != nil {
+		return fmt.Errorf("error reading Id time: %w", err)
+	}
+	// stop reading and ignore remaining data if we encounter a zeroed ID
+	if f.Id == (ChannelID{}) {
+		return nil
```
protolambda, line 100:
```suggestion
		return io.EOF
```
We can stop reading the input reader completely at this point, instead of allowing the channel bank to read a next frame after this.

op-node/rollup/derive/channel_frame.go
```diff
@@ -0,0 +1,134 @@
...
+// channel_id        = random ++ timestamp
+// random            = bytes32
+// timestamp         = uvarint
+// frame_number      = uvarint
+// frame_data_length = uvarint
+// frame_data        = bytes
+// is_last           = bool
+
+type Frame struct {
+	Id          ChannelID
```
protolambda, line 26: nit: `ID` instead of `Id` in Go, like the `ChannelID` type.

protolambda (commented):
Nice, there are a few minor issues to fix and then it looks good to merge.

op-node/rollup/derive/channel_frame.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("error reading frame number: %w", err)
+	}
+
+	frameLength, err := binary.ReadUvarint(r)
+	if err != nil {
+		return fmt.Errorf("error reading frame length: %w", err)
+	}
+
+	var dataBuf bytes.Buffer
+	n, err := io.CopyN(&dataBuf, r, int64(frameLength))
```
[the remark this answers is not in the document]
trianglesphere: I've got a new check (cap it to `int32` max size - 2GB) b/c that's the max size of an array on some platforms in go.

op-node/rollup/derive/channel_frame.go
```diff
@@ -0,0 +1,130 @@
...
+	frameLength, err := binary.ReadUvarint(r)
+	if err != nil {
+		return fmt.Errorf("error reading frame length: %w", err)
+	}
+
+	// Cap frame length to int32 because a golang int is either 32 or 64 bits
+	// and it cannot allocate an array larger than math.MaxInt
+	if frameLength > math.MaxInt32 {
+		return fmt.Errorf("frameLength is too larger")
+	}
```
protolambda, line 112: I think we should be much stricter about the frame size.

It's trivially cheap to create many invalid frames, with absurd frame lengths, creating a potential DoS vector if we put the limit too high (although only the batch-submitter can exploit it).

```suggestion
	if frameLength > MaxFrameLen {
		return fmt.Errorf("frame length is too large: %d", frameLength)
	}
```

And would need to define the limit something like:
```go
// Frames cannot be larger than 1 MB.
// Data transactions that carry frames are generally not larger than 128 KB due to L1 network conditions, but we leave space to grow larger anyway (gas limit allows for more data).
const MaxFrameLen = 1000_000
```

protolambda (commented):
LGTM, but would lower the frame length limit to avoid potential of DoS by batcher.

op-node/rollup/derive/channel_frame.go
trianglesphere: fixed

protolambda (approved):

mslipper (approved):

PR #3126, opened 2022-07-28
op-node: Fixed Integers in channel frame header

trianglesphere:
**Description**
This switches the channel frame header over to using fixed integers. This avoids a circular
dependency when filling the serialized frame up to a maximum size. The circular dependency
is caused by the frame data length being a `uvarint` which can vary in size when serialized.

**Metadata**
- Fixes ENG-2371
- Fixes ENG-2537

### DIFF
--- a/op-node/rollup/derive/channel_bank_test.go
+++ b/op-node/rollup/derive/channel_bank_test.go
@@ -1,6 +1,8 @@
 package derive
 
 import (
+	"bytes"
+	"fmt"
 	"math/rand"
 	"strconv"
 	"strings"
@@ -106,32 +108,28 @@ func (tf testFrame) Content() []byte {
 	return []byte(strings.TrimSuffix(parts[3], "!"))
 }
 
-func (tf testFrame) Encode() []byte {
-	chID := tf.ChannelID()
-	var out []byte
-	out = append(out, chID.Data[:]...)
-	out = append(out, makeUVarint(chID.Time)...)
-	out = append(out, makeUVarint(tf.FrameNumber())...)
-	content := tf.Content()
-	out = append(out, makeUVarint(uint64(len(content)))...)
-	out = append(out, content...)
-	if tf.IsLast() {
-		out = append(out, 1)
-	} else {
-		out = append(out, 0)
+func (tf testFrame) ToFrame() Frame {
+	return Frame{
+		ID:          tf.ChannelID(),
+		FrameNumber: uint16(tf.FrameNumber()),
+		Data:        tf.Content(),
+		IsLast:      tf.IsLast(),
 	}
-	return out
 }
 
 func (bt *bankTestSetup) ingestData(data []byte) {
 	require.NoError(bt.t, bt.cb.IngestData(data))
 }
 func (bt *bankTestSetup) ingestFrames(frames ...testFrame) {
-	data := []byte{DerivationVersion0}
+	data := new(bytes.Buffer)
+	data.WriteByte(DerivationVersion0)
 	for _, fr := range frames {
-		data = append(data, fr.Encode()...)
+		f := fr.ToFrame()
+		if err := (&f).MarshalBinary(data); err != nil {
+			panic(fmt.Errorf("error in making frame during test: %w", err))
+		}
 	}
-	bt.ingestData(data)
+	bt.ingestData(data.Bytes())
 }
 func (bt *bankTestSetup) repeatStep(max int, outer int, outerClosed bool, err error) {
 	require.Equal(bt.t, err, RepeatStep(bt.t, bt.cb.Step, Progress{Origin: bt.origins[outer], Closed: outerClosed}, max))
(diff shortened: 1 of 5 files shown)


op-node/rollup/derive/channel_frame.go
```diff
@@ -91,21 +77,20 @@ func (f *Frame) UnmarshalBinary(r ByteReader) error {
 	if err != nil {
 		return fmt.Errorf("error reading ID: %w", err)
 	}
-	f.ID.Time, err = binary.ReadUvarint(r)
-	if err != nil {
-		return fmt.Errorf("error reading ID.Time: %w", err)
+	if err := binary.Read(r, binary.BigEndian, &f.ID.Time); err != nil {
```
mslipper, line 80: This is cool, didn't know about `binary.Read`. I've always used `binary.BigEndian.ReadUint64`, etc.

specs/derivation.md
```diff
@@ -265,15 +265,23 @@ frame = channel_id ++ frame_number ++ frame_data_length ++ frame_data ++ is_last
...
+frame_number      = uint16
+frame_data_length = uint32
 frame_data        = bytes
 is_last           = bool
 ```
 
 > **TODO** replace `uvarint` by fixed size integers
 
+timestamp: 64 bits
+frame_number: 16 bits?
```
mslipper, line 278: [nit] you can delete the **TODO** above.

mslipper (approved):

op-node/rollup/derive/channel_frame.go
trianglesphere: I can't actually find a `binary.BigEndian.ReadUint64` which works on readers, a lot of the binary encoding API works on buffers

specs/derivation.md
```diff
@@ -270,22 +270,16 @@ frame_number      = uint16
...
-timestamp: 64 bits
-frame_number: 16 bits?
-frame_data_length: 32 bits?
-
-Fixed overhead: 32 + 8 + 2 + 4 + 1 = 47 bytes. Can get down to 16 + 6 = 22 bytes (128 bit UID, 16 bit #, 32 bit len)
+Where `uint64`, `uint32` and `uint16` are all big-endian unsigned integers.
+```
 
-Why fixed length: don't have circular dependency b/c have target data length
+All data in a frame is fixed-size, except the `frame_data`. The fixed overhead is `32 + 8 + 2 + 4 + 1 = 47 bytes`.
```
trianglesphere, line 277: Out of scope for this PR, but I think we can cut this down to ~22 bytes

op-node/rollup/derive/channel_frame.go
```diff
@@ -12,70 +12,56 @@ import (
 // but we leave space to grow larger anyway (gas limit allows for more data).
 const MaxFrameLen = 1_000_000
```
trianglesphere, line 13:
```suggestion
const MaxFrameLen = 10_000_000
```

specs/derivation.md
```diff
@@ -290,7 +292,7 @@ where:
       margin. (A soft constraint is not a consensus rule — nodes will accept such blocks in the canonical chain but will
       not attempt to build directly on them.)
 - `frame_number` identifies the index of the frame within the channel
-- `frame_data_length` is the length of `frame_data` in bytes
+- `frame_data_length` is the length of `frame_data` in bytes. It is capped to 1,000,000 bytes.
```
tynes, line 295: I feel like this should be a power of two? Why this number?

trianglesphere: This was proto's suggestion. I'd probably go a tad bigger, but I don't have a strong opinion on what it should be. We probably need to look into cannon's limits here

protolambda: Any limit is fine for now, we're not hitting it for now. Doesn't have to be a power of 2. Feel free to change it to a power of 2 though (eth2 specs do it everywhere, except the 1MB network gossip limit, haha).

op-node/rollup/derive/channel_bank_test.go
```diff
@@ -106,32 +108,28 @@ func (tf testFrame) Content() []byte {
...
 	require.NoError(bt.t, bt.cb.IngestData(data))
 }
 func (bt *bankTestSetup) ingestFrames(frames ...testFrame) {
-	data := []byte{DerivationVersion0}
+	data := new(bytes.Buffer)
+	data.WriteByte(DerivationVersion0)
 	for _, fr := range frames {
-		data = append(data, fr.Encode()...)
+		f := fr.ToFrame()
+		if err := (&f).MarshalBinary(data); err != nil {
```
protolambda, line 128:
```suggestion
		if err := f.MarshalBinary(data); err != nil {
```
We can call methods with pointer receivers on symbols without doing this thing without negative effects I think.

protolambda (commented):
looks good, but not approving just yet to avoid auto-merge, one nit style fix first.

protolambda (approved):

PR #3130, opened 2022-07-29
pkg driver: use derive error types to optionally retry with backoff

tuxcanfly:
**Description**
This PR updates the driver package to make use of the error severity returned by derive package and backoff with retry, if the error is temporary.

**Additional context**
The earlier PR which introduced on `derive` pkg changes has been merged to make it easier to review. Similarly the epoch hash validation PR has been merged as it depends on custom error types.

**Metadata**
- Fixes #ENG-2525


op-node/rollup/derive/error.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	return fmt.Sprintf("Unknown ErrorCode (%d)", int(e))
+}
+
+// Error provides a single type for errors that can happen during derivation.
+type Error struct {
+	ErrorCode   ErrorCode // Describes the kind of error
+	Description string    // Human readable description of the issue
+	Err         error     // Underlying error
+}
```
protolambda, line 82: I think it makes more sense to create `ErrCritical, ErrTemporary, `ErrReorg` types that wrap an error and rely on the `errors.Is` type checking, while preserving the `Unwrap` functionality to rely the *underlying* error, instead of representing the underlying error as string + returning only 3 non-informative errors on Unwrap. Intuition here is that with wrapping we can preserve the original error information, instead of obfuscating the error.

protolambda (changes requested):

op-node/rollup/derive/error.go
```diff
@@ -0,0 +1,101 @@
...
+	// ErrCritical is returned when there is a reorg and the pipeline MUST be
+	// reset
+	ErrCritical = errors.New("critical error")
+	// ErrHalt is returned when there is a system misconfiguration and the node
+	// needs to be halted.
+	ErrHalt = errors.New("misconfiguration error")
+)
+
+// ErrorCode identifies a kind of error.
+type ErrorCode int
```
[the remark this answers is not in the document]
trianglesphere: ditto - unless we make programatic decisions based on this, we shouldn't be using errcodes. The more idiomatic was is to use errors.Is/errors.As & wrapped sentinel errors if we need this.

op-node/rollup/derive/error.go
trianglesphere: Along this vein, I think something like the following makes sense
```
func NewCritical(err error) error
func NewTemporaryError(err error) error
```
Each of these could be checked against with `errors.Is` and `errors.As`. Each would also support the unwrap interface which would enable using other sentinel errors.

op-node/rollup/driver/conf_depth.go
```diff
@@ -31,7 +33,20 @@ func (c *confDepth) L1BlockRefByNumber(ctx context.Context, num uint64) (eth.L1B
...
+			} else {
+				bOff := backoff.Exponential()
+				err = backoff.Do(10, bOff, func() error {
+					var err error
+					nextOrigin, err = c.L1Fetcher.L1BlockRefByNumber(ctx, num)
+					return err
+				})
+			}
+		}
+		return nextOrigin, err
```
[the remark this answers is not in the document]
tuxcanfly: Ok, restored.

op-node/rollup/driver/step.go
```diff
@@ -54,7 +68,20 @@ func (d *outputImpl) createNewBlock(ctx context.Context, l2Head eth.L2BlockRef,
...
+			return l2Head, nil, err
+		} else {
+			bOff := backoff.Exponential()
+			err = backoff.Do(10, bOff, func() error {
+				var err error
+				ref, err = derive.PayloadToBlockRef(payload, &d.Config.Genesis)
+				return err
+			})
+		}
+	}
```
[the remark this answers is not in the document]
tuxcanfly: Restored.

op-node/rollup/derive/error.go
```diff
@@ -0,0 +1,101 @@
...
+var (
+	// ErrTemporary is returned when the error is temporary in nature and
+	// recovery is possible, so the caller can reattempt with a backoff
+	ErrTemporary = errors.New("temporary error")
+	// ErrCritical is returned when there is a reorg and the pipeline MUST be
+	// reset
+	ErrCritical = errors.New("critical error")
+	// ErrHalt is returned when there is a system misconfiguration and the node
+	// needs to be halted.
+	ErrHalt = errors.New("misconfiguration error")
```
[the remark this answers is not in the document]
tuxcanfly: Renamed

op-node/rollup/driver/state.go
```diff
@@ -100,9 +103,22 @@ func NewState(driverCfg *Config, log log.Logger, snapshotLog log.Logger, config
 // The loop will have been started iff err is not nil.
 func (s *state) Start(ctx context.Context) error {
 	l1Head, err := s.l1.L1HeadBlockRef(ctx)
+	if err != nil {
+		if !errors.Is(err, derive.ErrTemporary) {
+			return err
+		} else {
+			bOff := backoff.Exponential()
```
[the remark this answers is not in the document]
tuxcanfly: This was removed according to the below suggestion to use backoff mainly in `Step` function.

op-node/rollup/derive/error.go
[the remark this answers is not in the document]
tuxcanfly: Removed error codes and switched to custom error type with `Is/As` support as discussed below.

op-node/rollup/derive/error.go
tuxcanfly: Implemented custom error type according to the suggestions. The sentinel can be checked with `errors.Is`. The underlying error is now unwrapped properly.

op-node/rollup/driver/state.go
```diff
@@ -100,9 +103,22 @@ func NewState(driverCfg *Config, log log.Logger, snapshotLog log.Logger, config
 // The loop will have been started iff err is not nil.
 func (s *state) Start(ctx context.Context) error {
```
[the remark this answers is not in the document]
tuxcanfly: @protolambda handling the `Step` error in the event loop by addition additional `if/else` clauses based on `errors.Is` like so:
```

			} else if err != nil && errors.Is(err, derive.ErrReset) {
				// If the pipeline corrupts, e.g. due to a reorg, simply reset it
				s.log.Warn("Derivation pipeline is reset", "err", err)
				s.derivation.Reset()
				s.metrics.RecordPipelineReset()
				}
				continue
			} else {

```

for `ErrReset` as we want to reset derivation pipeline.

Likewise for `ErrTemporary`, added backoff with exponential retry:

```
			} else if err != nil && errors.Is(err, derive.ErrTemporary) {
				s.log.Warn("Derivation process temporary error", "err", err)
				// If there's a temporary error, retry with backoff
				err := backoff.Do(10, bOff, func() error {
					reqStep()
					return nil
				})
				if err != nil {
					s.log.Warn("Derivation retry with backoff failed permanently", "err", err)
				}
				continue

```

I'm calling `reqStep` in the `backoff` func as want to keep retrying the derivation step in case of temporary error. As such the backoff is now a part of the event loop. Please let me know if this is how you thought about integrating the backoff and error type.

op-node/rollup/driver/state.go
```diff
@@ -358,11 +364,22 @@ func (s *state) eventLoop() {
...
+			} else if err != nil && errors.Is(err, derive.ErrTemporary) {
+				s.log.Warn("Derivation process temporary error", "err", err)
+				// If there's a temporary error, retry with backoff
+				err := backoff.Do(10, bOff, func() error {
+					reqStep()
+					return nil
+				})
+				if err != nil {
+					s.log.Warn("Derivation retry with backoff failed permanently", "err", err)
+				}
```
protolambda, line 381: I don't think this works. If the backoff function does not return an error, the backoff time will never increase. I'd like it to request the next step, and update some local state in this function above the loop to keep track of the consecutive errors + do a backoff accordingly (i.e. don't do steps by nilling the stepReqCh temporarily)

tuxcanfly: Got it, sorry overlooked that we're returning `nil` unconditionally.

op-node/rollup/driver/state.go
```diff
@@ -358,11 +364,22 @@ func (s *state) eventLoop() {
...
+				continue
```
protolambda, line 382: - We should add a `else if err != nil { ... } ` case, to catch when there's an error that is not marked as temporary/reset/critical, just to be sure we log it.

- We should add a case to catch the critical type of error. We can log the error, and return from the function. Maybe we can propagate the error to the thing that created/started the driver, so the node can shut down. Leaving the node idle upon a critical error is good enough for now, we can implement handling of that case in a different PR.

tuxcanfly: Added these conditions.

op-node/rollup/driver/state.go
tuxcanfly: Updated to handle errs in a buffered channel and added case in the event loop select to dequeue and backoff by setting `stepReqCh` to nil. We're _always_ returning the err in the backoff func now, so it will set `stepReqCh` to nil until _all_ the errs are dequeued and so we are backing off from more derivation steps exponentially.

op-node/rollup/driver/state.go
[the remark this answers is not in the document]
tuxcanfly: Resolving as this was discussed below.

op-node/rollup/derive/attributes.go
```diff
@@ -33,25 +33,42 @@ func PreparePayloadAttributes(ctx context.Context, cfg *rollup.Config, dl L1Rece
...
+					epoch, info.ParentHash(), l2Parent.L1Origin),
+			)
 		}
 		deposits, err := DeriveDeposits(receipts, cfg.DepositContractAddress)
 		if err != nil {
-			return nil, true, fmt.Errorf("failed to derive some deposits: %w", err)
+			return nil, NewTemporaryError(
+				err,
+				"failed to derive some deposits",
+			)
```
protolambda, line 53:
```suggestion
			return nil, NewResetError(
				err,
				"failed to derive some deposits",
			)
```

op-node/rollup/derive/attributes.go
```diff
@@ -60,7 +77,10 @@ func PreparePayloadAttributes(ctx context.Context, cfg *rollup.Config, dl L1Rece
 
 	l1InfoTx, err := L1InfoDepositBytes(seqNumber, l1Info)
 	if err != nil {
-		return nil, true, fmt.Errorf("failed to create l1InfoTx: %w", err)
+		return nil, NewTemporaryError(
+			err,
+			"failed to create l1InfoTx",
+		)
```
protolambda, line 83:
```suggestion
		return nil, NewResetError(
			err,
			"failed to create l1InfoTx",
		)
```

protolambda (approved):
Just two errors to fix before merging. These would previously force a reorg and they still should, i.e. not `NewTemporaryError` but `NewResetError` in these cases. LGTM otherwise.

op-node/rollup/derive/attributes.go
tuxcanfly: Done.

tuxcanfly:
Sorry, missed that. Fixed.

op-node/rollup/derive/error.go
```diff
@@ -0,0 +1,94 @@
...
+func NewResetError(err error, desc string) error {
+	return NewError(
+		err,
+		desc,
+		LevelReset,
+	)
+}
+
+// NewCrititalError returns a critical error.
+func NewCrititalError(err error, desc string) error {
```
protolambda, line 82:
```suggestion
// NewCriticalError returns a critical error.
func NewCriticalError(err error, desc string) error {
```

op-node/rollup/derive/error.go
```diff
@@ -0,0 +1,94 @@
...
+		desc,
+		LevelCritical,
+	)
+}
+
+// Sentinel errors, use these to get the severity of errors by calling
+// errors.Is(err, ErrTemporary) for example.
+var ErrTemporary = NewTemporaryError(nil, "temporary error")
+var ErrReset = NewResetError(nil, "pipeline reset error")
+var ErrCritical = NewCrititalError(nil, "critical error")
```
protolambda, line 94:
```suggestion
var ErrCritical = NewCriticalError(nil, "critical error")
```

protolambda (commented):
typo

op-node/rollup/derive/error.go
tuxcanfly: Fixed.

op-node/rollup/derive/error.go
tuxcanfly: Fixed.

protolambda:
Triggered CI to rerun. It's past midnight UTC, new nightly foundry build with fix may be up, don't need to wait for the foundry change in #3182 if so.

PR #3134, opened 2022-07-29
pkg derive: match epoch hash with l1 block hash

tuxcanfly:
**Description**
This PR updates the batch queue to also match the epoch hash with the l1 block hash to make sure we are not re-orged out during the batch building.

**Additional context**
Additionaly, we want to handle temporary errors and skip validation if there is an RPC or connection error.

**Metadata**
- Fixes #ENG-2364


tuxcanfly:
Since this depends on #3130 and is a small change, rebased and pushed to #3130 directly.

PR #3135, opened 2022-07-29
contracts-bedrock: L2ToL1MessagePasser event extension

tynes:
**Description**

Emit an additional event during `initiateWithdrawal` that includes
the withdrawal hash so that it is easy to observe changes to the
`sentMessages` mapping.

### DIFF
--- a/packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
+++ b/packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
@@ -50,6 +50,14 @@ contract L2ToL1MessagePasser is Semver {
     );
 
     /**
+     * @notice Emitted any time a withdrawal is initiated. An extension to
+     *         WithdrawalInitiated so that the interface is maintained.
+     *
+     * @param hash The hash of the withdrawal
+     */
+    event WithdrawalInitiatedExtension1(bytes32 indexed hash);
+
+    /**
      * @notice Emitted when the balance of this contract is burned.
      *
      * @param amount Amount of ETh that was burned.
@@ -106,6 +114,8 @@ contract L2ToL1MessagePasser is Semver {
         sentMessages[withdrawalHash] = true;
 
         emit WithdrawalInitiated(nonce, msg.sender, _target, msg.value, _gasLimit, _data);
+        emit WithdrawalInitiatedExtension1(withdrawalHash);
+
         unchecked {
             ++nonce;
         }
--- a/.changeset/hip-shrimps-cheat.md
+++ b/.changeset/hip-shrimps-cheat.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/contracts-bedrock': patch
+---
+
+Emit an extra event when withdrawals are initiated to make chainops easier


maurelian:
I guess you need this here rather than the fixes repo because of the SDK?

maurelian (commented):
Should add a test for this new event.

op-node/withdrawals/utils.go
```diff
@@ -255,14 +263,54 @@ func ParseWithdrawalInitiated(receipt *types.Receipt) (*bindings.L2ToL1MessagePa
...
 		return nil, err
 	}
-	if len(receipt.Logs) != 1 {
-		return nil, errors.New("invalid length of logs")
+	abi, err := bindings.L2ToL1MessagePasserMetaData.GetAbi()
+	if err != nil {
+		return nil, err
 	}
-	ev, err := contract.ParseWithdrawalInitiated(*receipt.Logs[0])
+
```
tynes, line 270: This will allow the withdrawal to come from a contract that emits events and still be parsed, altho it won't work on multiple withdrawals in the same tx. It will always parse the first one. I think this is fine for now

op-node/withdrawals/utils.go
```diff
@@ -172,8 +173,15 @@ func FinalizeWithdrawalParameters(ctx context.Context, l2client ProofClient, txH
 	if err != nil {
```
protolambda, line 173: Could add a log Index argument to this method to point to the withdrawal log. (and can make that index relative to other withdrawals, or to all log events).

packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
```diff
@@ -49,6 +49,14 @@ contract L2ToL1MessagePasser is Semver {
         bytes data
     );
 
+    /**
+     * @notice Emitted any time a withdrawal is initiated. An extension to
+     *         WithdrawalInitiated so that the interface is maintained.
+     *
+     * @param hash The hash of the withdrawal
+     */
+    event WithdrawalInitiatedExtension1(bytes32 indexed hash);
```
protolambda, line 58: Can you please add this to the `specs/withdrawals.md` doc?

protolambda (changes requested):
please keep the specs in sync

op-node/withdrawals/utils.go
```diff
@@ -255,14 +263,54 @@ func ParseWithdrawalInitiated(receipt *types.Receipt) (*bindings.L2ToL1MessagePa
...
-	if len(receipt.Logs) != 1 {
-		return nil, errors.New("invalid length of logs")
+	abi, err := bindings.L2ToL1MessagePasserMetaData.GetAbi()
+	if err != nil {
+		return nil, err
 	}
-	ev, err := contract.ParseWithdrawalInitiated(*receipt.Logs[0])
+
+	for _, log := range receipt.Logs {
+		topic := log.Topics[0]
```
trianglesphere, line 272: can also look at the topic directly, but this does work.

trianglesphere (approved):
go code's good. Adding to the specs would be nice as well.

op-node/withdrawals/utils.go
tynes: I could update this function to return a `[]FinalizedWithdrawalParameters` but didn't because we don't really need that right now, will do it if you think I should

op-node/withdrawals/utils.go
tynes: updated

op-node/withdrawals/utils.go
protolambda: it's probably fine, not too difficult to do later once we actually need it

protolambda (approved):

tynes:
Not sure why the e2e tests hang when running in the cloud, they work fine locally

PR #3137, opened 2022-07-30
contracts-bedrock: fix deployment

tynes:
**Description**

It was being deployed with the implementation address
of the `L2OutputOracle` instead of the proxy. This
was preventing output commitments to actually be posted.


protolambda (approved):
Nice catch

PR #3139, opened 2022-07-30
ops-bedrock: simple config fixes

tynes:
**Description**

Deletes an old config option and also changes the poll
interval on the `op-proposer` to `1s` so that it submits
outputs faster. I found that the proposer will lag a lot,
especially if the commitment interval is small.

We need to carefully consider the implications of the commitment
interval and understand if it can be updated dynamically or not.


protolambda:
Is the lag not because of the proposals being only for safe data, i.e. the op-node conf depth + batch submission time delay the L2 block from being proposed on L1?

tynes:
> Is the lag not because of the proposals being only for safe data, i.e. the op-node conf depth + batch submission time delay the L2 block from being proposed on L1?

To reframe - we need to make sure that the output commitment interval is large enough to not fall behind 2s blocktime, given some number of L1 confs for each output commitment tx sent to L1. I'm not sure what this value is, and I think setting it too large could make creating fault proof witnesses longer/larger

tynes:
But i suppose parallel tx submission solves this problem, the failure case that i'm worried about is the proposer has a liveliness issue and can't catch up

protolambda (approved):

PR #3140, opened 2022-07-30
op-node/cmd: update op-node command description [bedrock]

protolambda:
The op-node command description was still mentioning the old deposit-only prototype so I updated it.


tynes (approved):

protolambda:
rebased

mslipper (approved):

PR #3141, opened 2022-07-30
contracts-bedrock: add forge contract verification support

tynes:
**Description**

Add etherscan contract verification for the hh/foundry
compiler toolchain. This will eventually be upstreamed,
but is implemented here to ensure that it works thoroughly
first.

To run the task:

```
$ export ETHERSCAN_API_KEY=$(cat api-key.txt)
$ npx hardhat forge-verify --network goerli
```

In the future, other contract verification backends
will be added to forge and this task will be updated
to support them.

Example contract that was verified: https://goerli.etherscan.io/address/0x1234662682c85fa6fb375416d14db965eba222ba#code

An individual contract can be targeted instead of all of them
with the `--contract` flag. If hardhat deploy has configured
external deployments, forge will not be able to verify them.
It may be possible to implement that in the future.


packages/contracts-bedrock/tasks/forge-verify.ts
```diff
@@ -0,0 +1,128 @@
...
+    'Name of the contract to verify',
+    '',
+    types.string
+  )
+  .addOptionalParam(
+    'etherscanApiKey',
+    'Etherscan API key',
+    process.env.ETHERSCAN_API_KEY,
+    types.string
+  )
```
protolambda, line 77: not all `ForgeVerifyArgs` are defined here, is that correct?

tynes: Yes, because they are pulled from the deployment artifacts. Makes it easier from a devex perspective

mslipper (approved):

PR #3143, opened 2022-07-31
contracts-bedrock: fix slither

tynes:
**Description**

`slither` crashes on new solidity features, one is used
in the forge tests. Not exactly sure what it is, but
this will prevent that problem by moving the tests
to a temp dir when running slither and then moving them
back.


tynes:
> This didn't seem to fix the problem but it did work locally, need to double check why

Pushed a fix that works

maurelian (approved):

PR #3148, opened 2022-08-01
contracts-bedrock: Fix provider usage in L2OutputOracle deployment script

mslipper:
`getDefaultProvider()` returns a fallback mainnet provider, which doesn't work.

Builds on https://github.com/ethereum-optimism/optimism/pull/3147.

### DIFF
--- a/packages/contracts-bedrock/deploy/000-L2OutputOracle.deploy.ts
+++ b/packages/contracts-bedrock/deploy/000-L2OutputOracle.deploy.ts
@@ -10,7 +10,7 @@ const deployFn: DeployFunction = async (hre) => {
   const { deployer } = await hre.getNamedAccounts()
   const { deployConfig } = hre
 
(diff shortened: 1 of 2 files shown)


packages/contracts-bedrock/deploy/000-L2OutputOracle.deploy.ts
```diff
@@ -10,7 +10,7 @@ const deployFn: DeployFunction = async (hre) => {
   const { deployer } = await hre.getNamedAccounts()
   const { deployConfig } = hre
 
-  const l1 = hre.ethers.getDefaultProvider()
```
tynes, line 13: I thought this would return a provider that is associated with the `--network` config? I guess not

mslipper: Unfortunately not :(

tynes (approved):

tynes (approved):

PR #3150, opened 2022-08-01
op-node: Add p2p utils

mslipper:
Adds utilities for handling P2P keys. Note that to prevent having to specify irrelevant CLI flags, I created a new `start` command that actually stats the node.

### DIFF
--- a/op-node/cmd/main.go
+++ b/op-node/cmd/main.go
@@ -9,6 +9,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/ethereum-optimism/optimism/op-node/cmd/p2p"
+
 	"github.com/ethereum-optimism/optimism/op-node/metrics"
 
 	opnode "github.com/ethereum-optimism/optimism/op-node"
@@ -53,13 +55,22 @@ func main() {
 	)
 
 	app := cli.NewApp()
-	app.Flags = flags.Flags
 	app.Version = VersionWithMeta
 	app.Name = "opnode"
 	app.Usage = "Optimism Rollup Node"
 	app.Description = "The deposit only rollup node drives the L2 execution engine based on L1 deposits."
+	app.Commands = []cli.Command{
+		{
+			Name:        "p2p",
+			Subcommands: p2p.Subcommands,
+		},
+		{
+			Name:   "start",
+			Action: RollupNodeMain,
+			Flags:  flags.Flags,
+		},
+	}
 
-	app.Action = RollupNodeMain
 	err := app.Run(os.Args)
 	if err != nil {
 		log.Crit("Application failed", "message", err)
--- a/op-node/cmd/main.go
+++ b/op-node/cmd/main.go
@@ -56,6 +56,7 @@ func main() {
 
 	app := cli.NewApp()
 	app.Version = VersionWithMeta
+	app.Flags = flags.GlobalFlags
 	app.Name = "opnode"
 	app.Usage = "Optimism Rollup Node"
 	app.Description = "The deposit only rollup node drives the L2 execution engine based on L1 deposits."
--- a/op-node/cmd/p2p/cmd.go
+++ b/op-node/cmd/p2p/cmd.go
@@ -97,7 +97,9 @@ var Subcommands = cli.Commands{
 		Usage: "Generates a private key",
 		Action: func(ctx *cli.Context) error {
 			buf := make([]byte, 32)
-			_, _ = rand.Read(buf)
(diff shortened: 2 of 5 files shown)


op-node/cmd/main.go
```diff
@@ -53,13 +55,22 @@ func main() {
...
 	app.Name = "opnode"
 	app.Usage = "Optimism Rollup Node"
 	app.Description = "The deposit only rollup node drives the L2 execution engine based on L1 deposits."
+	app.Commands = []cli.Command{
+		{
+			Name:        "p2p",
+			Subcommands: p2p.Subcommands,
+		},
+		{
+			Name:   "start",
```
trianglesphere, line 68: k8s sister PR?

op-node/cmd/p2p/cmd.go
```diff
@@ -0,0 +1,105 @@
...
+	data, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+
+	rawStr := strings.TrimSpace(string(data))
+	rawStr = strings.TrimPrefix(rawStr, "0x")
+	b, err := hex.DecodeString(rawStr)
+	if err != nil {
+		return nil, errors.New("p2p key is not formatted in hex chars")
```
trianglesphere, line 65:
```suggestion
		return nil, fmt.Errorf("p2p key is not formatted in hex chars: %w", err)
```

op-node/cmd/p2p/cmd.go
```diff
@@ -0,0 +1,105 @@
...
+			fmt.Println(key)
+			return nil
+		},
+	},
+	{
+		Name:  "genkey",
+		Usage: "Generates a private key",
+		Action: func(ctx *cli.Context) error {
+			buf := make([]byte, 32)
+			_, _ = rand.Read(buf)
```
trianglesphere, line 100:
```suggestion
			if _, err := rand.Read(buf); err != nil {
						return fmt.Errorf("failed to get entropy: %w", err)
			}
```
note: probably need to gofmt after suggestion

op-node/cmd/main.go
mslipper: Once this is merged, yes.

trianglesphere (approved):

op-node/cmd/p2p/cmd.go
```diff
@@ -0,0 +1,106 @@
...
+	rawStr := strings.TrimSpace(string(data))
+	rawStr = strings.TrimPrefix(rawStr, "0x")
+	b, err := hex.DecodeString(rawStr)
+	if err != nil {
+		return nil, fmt.Errorf("p2p key is not formatted in hex chars: %w", err)
+	}
+	return b, nil
+}
+
+var Subcommands = cli.Commands{
```
protolambda, line 69: Technically multi-addresses are being used in the p2p stack, like: `/ip/1.2.3.4/tcp/1234/p2p/peeridgibberishhere` but string variables in shell scripts can probably do that with the `peeridgibberishhere` output from one of the commands here. Might add a command for this later if it's too error prone.

op-node/flags/flags.go
```diff
@@ -162,5 +159,12 @@ var optionalFlags = append([]cli.Flag{
...
 
-// Flags contains the list of configuration options available to the binary.
+// GlobalFlags contains the list of configuration options available to every command.
+var GlobalFlags = []cli.Flag{
+	LogLevelFlag,
+	LogFormatFlag,
+	LogColorFlag,
+}
+
+// Flags contains the list of configuration options available to the start command.
```
protolambda, line 169: We don't use a logger in the p2p commands. Maybe the other commands later, so it's fine to change, but still surprised.

op-node/cmd/main.go
protolambda: Does start really have to be a sub-command? I would like to run the `RollupNodeMain` as the main command by default. Same thing geth does.

ops-bedrock/docker-compose.yml
```diff
@@ -45,6 +45,7 @@ services:
       dockerfile: ./op-node/Dockerfile
     command: >
       op-node
+      start
```
protolambda, line 48: See comment about sub-command vs main command.

mslipper: Geth doesn't use required flags (the `Required` field in `cli.Command`), so subcommands can run without specifying the flag for the main command.

protolambda (approved):

PR #3151, opened 2022-08-01
op-node: Properly set timestamp in PreparePayloadAttributes

trianglesphere:
**Description**
This now uses a passed in timestamp for PreparePayloadAttributes.
This is important because when generating a payload attributes from
a batch it should be using the timestamp of the batch, not the
L2 parent + block time. When we are creating payloads based on the
L2 parent time, that is easy to set there instead of pulling it into
this function.

This enabled/exacerbated a divergence bug where the timestamp of the
batch was not the timestamp that was being set inside
PreparePayloadAttributes.

**Metadata**
- Fixes#ENG-2531


protolambda (approved):
nice catch

tynes (approved):

mslipper:
@mergifyio refresh

PR #3155, opened 2022-08-01
op-proposer, op-batcher: Support private keys in addition to mnemonics

mslipper:


op-proposer/flags/flags.go
```diff
@@ -80,6 +80,12 @@ var (
 		Required: true,
 		EnvVar:   prefixEnvVar("L2_OUTPUT_HD_PATH"),
 	}
+	PrivateKey = cli.StringFlag{
+		Name:     "private-key",
+		Usage:    "The private key to use with the l2output wallet. Must notbe used with mnemonic.",
```
trianglesphere, line 85:
```suggestion
		Usage:    "The private key to use with the l2output wallet. Must not be used with mnemonic.",
```

trianglesphere (approved):

op-batcher/batch_submitter.go
```diff
@@ -142,19 +144,31 @@ func NewBatchSubmitter(cfg Config, l log.Logger) (*BatchSubmitter, error) {
...
-	if err != nil {
-		return nil, err
-	}
+	var sequencerPrivKey *ecdsa.PrivateKey
+	var addr common.Address
 
-	sequencerPrivKey, err := wallet.PrivateKey(acc)
-	if err != nil {
-		return nil, err
+	if cfg.PrivateKey == "" {
```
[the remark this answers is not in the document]
mslipper: ah yeah nice catch, fixed

tynes:
I think @protolambda wanted to move to using encrypted keystores
https://github.com/ethereum/go-ethereum/blob/a0b88ce869d6be243c6b1900d69b9401f2d2c092/cmd/ethkey/generate.go#L109
https://github.com/ethereum/go-ethereum/blob/a0b88ce869d6be243c6b1900d69b9401f2d2c092/accounts/keystore/key.go#L66

I think this is fine because it won't require any changes in tooling or the way we handle secrets

op-proposer/l2_output_submitter.go
```diff
@@ -121,20 +126,33 @@ func NewL2OutputSubmitter(
...
 
 	ctx := context.Background()
+	var l2OutputPrivKey *ecdsa.PrivateKey
+	var err error
 
-	// Parse l2output wallet private key and L2OO contract address.
-	wallet, err := hdwallet.NewFromMnemonic(cfg.Mnemonic)
-	if err != nil {
-		return nil, err
+	if cfg.PrivateKey != "" && cfg.Mnemonic != "" {
```
tynes, line 132: Implementation in the batch-submitter also checks that one of them is enabled
https://github.com/ethereum-optimism/optimism/blob/0e7184b0f09eee193608e78cd0feb1431a2b145a/batch-submitter/config.go#L259

mslipper: Do you want me to change the implementation here to match the one in the BSS?

tynes: Not super opinionated but could reduce problems at deploy time

protolambda:
Maybe ideally we define one go module that can support the loading of private keys, in various ways:
- mnemonic + path
- private key
- encrypted keystore + password

That way we don't duplicate security-critical code, make it easy to audit, and easy to use.

Also ideally with discouraged use of mnemonic and private key. And warnings if you put the mnemonic, private key, or password in CLI instead of env var. (Since shell history / log streaming may log the CLI command and leak it)

I've been thinking we should have an `op-service` go module, for exactly this type of thing. As well as logging configuration, metrics, and pprof.

I'm fine with merging this PR as-is, but it would be nice if we can refactor it all into a common `op-service` module later.

op-proposer/l2_output_submitter.go
mslipper: Will leave as-is for now.

mslipper:
I agree with the `op-service` module. We have a lot of duplication among our various Go projects and it would be great to clean it up. We'll also need a separate "Signer" function to support things like Cloud HSMs in prod.

I'll merge this as-is, but will ticket the `op-service` work so we can get started on it.

PR #3156, opened 2022-08-02
op-node: Derivation metrics

mslipper:
Adds support for the following batch derivation metrics:

- 0/1 gauge for if the derivation is idle
- Count of total pipeline resets
- Count of total unsafe payloads received
- Count of total derivation errors
- Gauge tracking the various L1/L2 safe/unsafe heads

### DIFF
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/metrics"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
 	"github.com/ethereum/go-ethereum/log"
 )
@@ -67,6 +68,7 @@ type state struct {
 	output           outputInterface
 	network          Network // may be nil, network for is optional
 
+	metrics     *metrics.Metrics
 	log         log.Logger
 	snapshotLog log.Logger
 	done        chan struct{}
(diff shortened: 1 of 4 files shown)


op-node/rollup/driver/state.go
```diff
@@ -361,6 +371,9 @@ func (s *state) eventLoop() {
 				s.l2Finalized = finalized
 				s.l2SafeHead = safe
 				s.l2Head = unsafe
+				s.metrics.SetHead("l2_finalized", finalized.Number)
+				s.metrics.SetHead("l2_safe", safe.Number)
+				s.metrics.SetHead("l2_unsafe", unsafe.Number)
```
protolambda, line 376: Can you move this up, next to the log statement that only fires on changes? This code path is hit a lot repeatedly with the same values, and I'm worried the metrics collection will slow things down.

op-node/rollup/driver/state.go
```diff
@@ -330,14 +334,17 @@ func (s *state) eventLoop() {
...
 
 		case newL1Head := <-s.l1Heads:
 			s.log.Info("new l1 Head")
 			s.snapshot("New L1 Head")
 			s.handleNewL1Block(newL1Head)
+			s.metrics.SetHead("l1", newL1Head.Number)
 			reqStep() // a new L1 head may mean we have the data to not get an EOF again.
 		case <-stepReqCh:
+			s.metrics.SetDerivationIdle(false)
 			s.idleDerivation = false
```
protolambda, line 348: would be nice to refactor this into a method that changes the idle state, and only updates metrics if the new state is not the same as the old state.

protolambda (commented):
Nice work! Some nits on application of metrics in driver event loop, then good to merge.

op-node/rollup/driver/state.go
mslipper: Prom metrics are super efficient. Under the hood this is just an atomic integer set, so I think adding the logic to conditionally set the metric on change will have a negligible performance impact.

protolambda (approved):

PR #3157, opened 2022-08-02
op-node: Use unmetered L1 Attributes Transaction

trianglesphere:
**Description**
This enables the IsSystemTransaction flag in the L1 Attributes
deposit & updates to the latest version of geth.


protolambda (changes requested):
Can you update the specs?

deposits doc needs the system tx format bool: https://github.com/ethereum-optimism/optimism/blob/develop/specs/deposits.md

and derivation doc needs to set it to true for block info, false for user deposits: https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md

specs/deposits.md
```diff
@@ -246,6 +250,7 @@ feed contract][deposit-feed-contract] on L1.
 6. `isCreation` is set to `true` if the transaction is a contract creation, `false` otherwise.
 7. `data` is unchanged from the emitted value. Depending on the value of `isCreation` it is handled
    as either calldata or contract initialization code.
+8. `isSystemTransaction` is set by the rollup node for certain transactions that have unmetered execution.
```
protolambda, line 253: Can you make it more explicit it's `false` for user-deposited transactions?

trianglesphere, line 253:
```suggestion
8. `isSystemTransaction` is set by the rollup node for certain transactions that have unmetered execution.
  It is `false` for user deposited transactions
```

protolambda (approved):
need to rerun CI, op-e2e seems to have failed to start / output anything.

mslipper (approved):

mslipper (changes requested):
CI is flagging a real bug, so requesting change until we fix it.

protolambda:
Ah, I think this breaks the deposit hash calculation in TS (used in devnet test), because of the new deposit systemTx field. 

The op-e2e timing out happens sometimes in CI though, I think that's unrelated.

protolambda:
The devnet `ethereumoptimism/reference-optimistic-geth:latest` image may still need to be updated. We switched from `optimism-prototype` to `optimism` branch as default in the reference-optimistic-geth repo. Both to keep the audit branch nice and stable, as well as to get rid of the `-prototype` suffix: it's not a prototype anymore, we're about to enter alpha public testnet.

protolambda:
Cancelled CI, `devnet` was running for 2h, we should add a timeout on that job. I think we just need to update the L2 reference geth docker image to point to the new branch that this PR introduces.

protolambda:
Geth repo nows builds docker images for the new default `optimism` branch, including the updates required for this PR. I don't think the devnet is using a specific docker tag, but just uses the latest, so other CI might temporarily break until we merge this in.

tynes:
We need to merge this because its breaking the devnet for other PRs

protolambda:
I think either the CI is running an older docker image, or my JS change is not good, I'm seeing this in CI logs:
```
Sending from 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
Depositing 1 ETH to 0xB79f76EF2c5F0286176833E7B2eEe103b1CC3244
0xB79f76EF2c5F0286176833E7B2eEe103b1CC3244 has 0.0 ETH on L2
Got TX hash 0x4b6f7e960d69276ef9c9a431e8acf27b8ced8a01a6b83bf8b61ec3f518a9e025. Waiting...
Included in block 0xc00acaa964fcdfa65490c20b2c92c6baeadcf28efdb710fd041e27403a26e102
Deposit has log index 1
Waiting for L2 TX hash 0x67251c1f210dd63a82eb58aaf2f0317f66bc72aec1a9fddd8b35834ab6b61743
Unexpected balance increase without detecting deposit transaction
latest block 211:0x7c59d94d829089088090d3acb3d64d4c4b839798eb2e441266ad84715766d50a
Unexpected balance increase without detecting deposit transaction
latest block 236:0x965047b4588d06a40b7c52f7d1682bc36c727ee02fdb3dd8a1b20de0fafe93d4
```
Which I think is because of the depsit-tx hash computation not matching between Go and JS. cc @tynes

packages/core-utils/src/optimism/deposit-transaction.ts
```diff
@@ -10,6 +10,10 @@ import {
 
 const { hexDataSlice, stripZeros, hexConcat, keccak256, zeroPad } = utils
```
tynes, line 12: I pulled the other helpers from ethers internal functions, there arent any for formatting bools so this looks good to me
https://github.com/ethers-io/ethers.js/blob/56fabe987bb8c1e4891fdf1e5d3fe8a4c0471751/packages/transactions/src.ts/index.ts#L118

tynes:
There are some solidity contracts that needs to be updated:
- https://github.com/ethereum-optimism/optimism/blob/bd65ee51a5ab592c53ef793e900e4add2924a8bc/packages/contracts-bedrock/contracts/libraries/Encoding.sol#L34
- https://github.com/ethereum-optimism/optimism/blob/bd65ee51a5ab592c53ef793e900e4add2924a8bc/packages/contracts-bedrock/contracts/libraries/Types.sol#L34

As well as the differential testing:
- https://github.com/ethereum-optimism/optimism/blob/bd65ee51a5ab592c53ef793e900e4add2924a8bc/packages/contracts-bedrock/scripts/differential-testing.ts#L87
- https://github.com/ethereum-optimism/optimism/blob/bd65ee51a5ab592c53ef793e900e4add2924a8bc/packages/contracts-bedrock/contracts/test/Hashing.t.sol#L126

tynes:
> I think either the CI is running an older docker image, or my JS change is not good, I'm seeing this in CI logs:
> 
> ```
> Sending from 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
> Depositing 1 ETH to 0xB79f76EF2c5F0286176833E7B2eEe103b1CC3244
> 0xB79f76EF2c5F0286176833E7B2eEe103b1CC3244 has 0.0 ETH on L2
> Got TX hash 0x4b6f7e960d69276ef9c9a431e8acf27b8ced8a01a6b83bf8b61ec3f518a9e025. Waiting...
> Included in block 0xc00acaa964fcdfa65490c20b2c92c6baeadcf28efdb710fd041e27403a26e102
> Deposit has log index 1
> Waiting for L2 TX hash 0x67251c1f210dd63a82eb58aaf2f0317f66bc72aec1a9fddd8b35834ab6b61743
> Unexpected balance increase without detecting deposit transaction
> latest block 211:0x7c59d94d829089088090d3acb3d64d4c4b839798eb2e441266ad84715766d50a
> Unexpected balance increase without detecting deposit transaction
> latest block 236:0x965047b4588d06a40b7c52f7d1682bc36c727ee02fdb3dd8a1b20de0fafe93d4
> ```
> 
> Which I think is because of the depsit-tx hash computation not matching between Go and JS. cc @tynes

This means the JS isn't computing the hash of the deposit tx correctly

protolambda:
Just checked the CI logs. Pulled down the exact docker image that is being used, and grepped the symbols in the geth binary to make sure it includes the latest updates, and it does. So it's some code in our testing that can't find the deposit correctly. Likely JS/solidity updates I missed.

protolambda:
Devnet stuck due to other PR, cancelled CI job because it didn't time out automatically (1h...)

protolambda:
Fixed a bug in deposit tx decoding (it was still skipping an extra byte, from when deposits had a version byte) + fixed encoding/decoding of the new boolean.

mslipper (approved):

PR #3158, opened 2022-08-02
Add `src` to files so sourcemaps work

roninjin10:
Source maps did not work before because the source code was not included in npm modules when downloading from npm 
image


.changeset/cyan-ladybugs-obey.md
```diff
@@ -0,0 +1,6 @@
+---
+'@eth-optimism/common-ts': minor
+'@eth-optimism/sdk': minor
```
tynes, line 3: Can you add changesets for the other packages too?

roninjin10:
Closing because I'm not 100% sure this will actually fix the problem and want to verify that first

PR #3159, opened 2022-08-02
fix: wsteth is a dai bridge not a standard bridge

roninjin10:
In previous pr we added wsteth as a standard bridge but it turns out it's a dai bridge actually


TreTuna (approved):

tynes:
This is due to the interface? How can i verify that this is correct?

nickbalestra (approved):

roninjin10:
Oops failed to include a changeset here

@tynes yea I originally assumed it was a standard bridge but they reached out to us telling us we made a mistake and it's a DAI bridge.  I should have recorded video but chaning it to DAI got rid of the error they were seeing.
image

PR #3160, opened 2022-08-03
fix: eth deposits through bridge

tynes:
**Description**

Withdrawals don't seem to update balances correctly
for some reason, still need to debug why that is.

### DIFF
--- a/packages/sdk/tasks/deposit.ts
+++ b/packages/sdk/tasks/deposit.ts
@@ -15,8 +15,9 @@ import {
 
 // TODO(tynes): this task could be modularized in the future
 // so that it can deposit an arbitrary token. Right now it
+// deposits and withdraws ETH through the bridge,
 // deploys a WETH9 contract, mints some WETH9 and then
-// deposits that into L2 through the StandardBridge
+// deposits that into L2 through the StandardBridge.
 task('deposit', 'Deposits WETH9 onto L2.')
   .addParam(
     'l2ProviderUrl',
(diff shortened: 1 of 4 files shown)


packages/sdk/tasks/deposit.ts
```diff
@@ -108,6 +119,60 @@ task('deposit', 'Deposits WETH9 onto L2.')
       bedrock: true,
     })
 
+    const opBalanceBefore = await signer.provider.getBalance(
+      OptimismPortal.address
+    )
+
+    // Deposit ETH
+    console.log('Depositing ETH through StandardBridge')
+    const ethDeposit = await messenger.depositETH(utils.parseEther('1'))
```
mslipper, line 128: Doesn't this just wait for the transaction on L1 to confirm? Don't we have to call `messenger.waitForMessageReceipt(ethDeposit)`?

packages/sdk/tasks/deposit.ts
```diff
@@ -108,6 +119,60 @@ task('deposit', 'Deposits WETH9 onto L2.')
...
+    if (messageReceipt.status !== 1) {
+      throw new Error('deposit failed')
+    }
+    console.log('Deposit complete')
+
+    const opBalanceAfter = await signer.provider.getBalance(
+      OptimismPortal.address
+    )
+
+    if (!opBalanceBefore.add(utils.parseEther('1')).eq(opBalanceAfter)) {
```
mslipper, line 139: This checks the portal balance, but not the actual balance on L2. Intentional?

mslipper (changes requested):

tynes:
We can merge this and I can follow up with a PR that fixes the bug in the `OptimismPortal`

packages/sdk/tasks/deposit.ts
tynes: This is intentional, the call in the `OptimismPortal` sometimes doesn't actually transfer the ETH

packages/sdk/tasks/deposit.ts
tynes: I didn't think that would wait for the L2 conf

mslipper (approved):

PR #3161, opened 2022-08-03
actor-tests: Bedrock actor tests

mslipper:
- Moves actors into their own package
- Minor updates to logging, etc.
- Adds Cloudbuild config so that we can get Docker images on every commit to Develop
- Updates Actors to support Bedrock

Currently supported actors:

- Deposits
- Contract deployments
- Simple sends


.circleci/config.yml
```diff
@@ -581,6 +581,12 @@ workflows:
       - op-bindings-build:
           requires:
             - yarn-monorepo
+      - js-lint-test:
+          name: actor-tests-tests
+          package_name: actor-tests
+          dependencies: "(core-utils|sdk)"
```
tynes, line 587: Should this also run on changes to `op-node`? Then we could see if there are performance changes. Perhaps even batcher/proposer

cloudbuild.yaml
```diff
@@ -32,5 +32,14 @@ steps:
       - --cache=true
       - --cache-ttl=48h
     waitFor: ['-']
```
tynes, line 34: Do we want to also publish to dockerhub and add to changesets? It feels like changesets is really convenient for js

.circleci/config.yml
mslipper: The goal of this job is to test the testing framework rather than run the actor tests against the devnet. Right now it just runs the linter. We can add automated perf tests in the future, but that will require a different job and some updates to how actors work to collect additional metrics.

cloudbuild.yaml
mslipper: I plan on deploying this to internal infrastructure only. The infra would pull a specific commit or `develop` so we don't strictly need changesets for this. We can always add it later.

protolambda:
Planning to review this in ~2h from now. Was not able to yesterday due to focus on landing the geth changes. Sorry for the delay. I'm still excited to have this feature.

packages/actor-tests/bedrock/account-sends.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  let wallets: Wallet[]
+
+  setupActor(async () => {
+    provider = new providers.JsonRpcProvider(process.env.L2_RPC)
+    wallets = devWalletsL2()
+  })
+
+  run(async (b, ctx: Context, logger) => {
+    const sender = wallets[Math.floor(Math.random() * wallets.length)]
```
tynes, line 22: With enough concurrency, this could hit nonce issues. Is there a way to ensure that `wallets.length === concurrency` and then deterministically select a wallet?

packages/actor-tests/bedrock/account-sends.ts
```diff
@@ -0,0 +1,34 @@
...
+    const recipient = Wallet.createRandom().connect(provider)
```
tynes, line 23: Do we want to send the funds to one of the wallets so we cycle the ETH instead of losing it forever?

packages/actor-tests/bedrock/deposit-eth.ts
```diff
@@ -0,0 +1,88 @@
...
+  let messenger: CrossChainMessenger
+
+  let contracts: any
+
+  setupActor(async () => {
+    l1Provider = new providers.JsonRpcProvider(process.env.L1_RPC)
+    l2Provider = new providers.JsonRpcProvider(process.env.L2_RPC)
+    wallet = new Wallet(process.env.PRIVATE_KEY)
+    contracts = JSON.parse(
+      fs.readFileSync(process.env.CONTRACTS_JSON_PATH).toString('utf-8')
```
tynes, line 34: You can `require` a json file and its automatically turned into an object, could reduce some boilerplate

packages/actor-tests/bedrock/deposit-eth.ts
```diff
@@ -0,0 +1,88 @@
...
+import { predeploys } from '@eth-optimism/contracts-bedrock'
+import { sleep } from '@eth-optimism/core-utils'
+
+import { actor, setupActor, run } from '../lib/convenience'
+
+interface Context {
+  wallet: Wallet
+}
+
+actor('Dev account sender', () => {
```
tynes, line 18: I think that this is good for now, we could technically incorporate `hardhat-deploy` and use a network to pull in deployment artifacts if we really wanted to, although that adds a lot of overhead so just using a path to a json file is more simple

packages/actor-tests/bedrock/deposit-eth.ts
```diff
@@ -0,0 +1,88 @@
...
+    const recipient = Wallet.createRandom().connect(l2Provider)
+    logger.log(`Depositing funds to ${recipient.address}.`)
+    const depositTx = await messenger.depositETH(utils.parseEther('0.0001'), {
+      recipient: recipient.address,
+    })
+    logger.log(`Awaiting receipt for deposit tx ${depositTx.hash}.`)
+    await depositTx.wait()
+    // Temporary until this is supported in the SDK.
+    for (let i = 0; i < 60; i++) {
+      const recipBal = await recipient.getBalance()
```
tynes, line 79: A less brittle way to do this would be this: https://github.com/ethereum-optimism/optimism/blob/09658f16b6e1e02dc9f589821b82746fa947fa50/packages/contracts-bedrock/tasks/deposits.ts#L92

Poll on the transaction hash that is computed client side

tynes: https://github.com/ethereum-optimism/optimism/pull/3160 fixes the hanging, still need to debug why the value isn't being fully passed through

tynes:
Generally looks good to me, some small nits

packages/actor-tests/bedrock/account-sends.ts
mslipper: This test uses the dev accounts in state, which have a gigantic amount of eth. The test itself only sends 66 wei, so  don't think this will be an issue. I'll make sure eth is recoverable once we use non-dev accounts.

packages/actor-tests/bedrock/deposit-eth.ts
mslipper: Yeah, I want to avoid incorporating Hardhat to keep thing simple.

packages/actor-tests/bedrock/deposit-eth.ts
mslipper: Did you see my comments on #3160?

I'll update to do the client side poll.

packages/actor-tests/bedrock/account-sends.ts
mslipper: This is tough - in previous iterations of this script it would fund intermediate wallets to make sure that nonces don't collide. I can have it do that again here.

protolambda (approved):
Maybe port over previous `nft.test.ts` and `uniswap.test.ts` actors, or create a ticker for extending actor tests? LGTM otherwise

packages/actor-tests/bedrock/account-sends.ts
mslipper: Fixed to use recoverable account.

packages/actor-tests/bedrock/deposit-eth.ts
mslipper: Hmm client side poll is more difficult since the Ethers types don't line up. The deposit TX utility in core utils expects an `Event` but because the deposit is initiated by the SDK i just get back `Logs`, which isn't compatible with `Event` even after inflating them using the Ethers `Interface` class.

Any objections to me updating this once the SDK is merged?

mslipper:
> Maybe port over previous nft.test.ts and uniswap.test.ts actors, or create a ticker for extending actor tests? LGTM otherwise

Done, ticketed.

tynes (approved):

PR #3163, opened 2022-08-03
contracts-bedrock: poc OptimismSystemConfig

tynes:
**Description**
PoC for the L1 optimism system config contract. The `OptimismPortal` would need a reference to it so that we know what the canonical system config contract is. It would be deployed behind a proxy so that it could be upgraded easily. This needs comments but is a poc to show how simple this contract could be.


maurelian:
Can we close this for now?

PR #3165, opened 2022-08-03
proxyd: Unwrap single RPC batches

mslipper:


proxyd/backend.go
```diff
@@ -349,7 +349,17 @@ func (b *Backend) setOffline() {
 }
 
 func (b *Backend) doForward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool) ([]*RPCRes, error) {
-	body := mustMarshalJSON(rpcReqs)
+	isSingleElementBatch := len(rpcReqs) == 0
```
Inphi, line 352: shouldn't this be `len(rpcReqs) == 1`?

mslipper: You are absolutely right. Nice catch, this was sloppy on my part.

mslipper: Fixed.

Inphi (approved):

tynes (approved):

mslipper:
@Inphi I re-requested review since I had to modify an existing test to handle this case.

Inphi (approved):
LGTM!

mslipper:
Thanks!

PR #3180, opened 2022-08-05
op-node: fix invalid payload insertion error handling

protolambda:
This PR relies on https://github.com/ethereum-optimism/reference-optimistic-geth/pull/40 to return `INVALID` when the payload attributes cannot be processed.

When payload attributes cannot be processed due to a true payload-error (i.e. rpc errors, syncing state, etc. don't count) then we first re-attempt with only the deposits, and then return a critical error if we cannot even process a block with just the deposits.

I've chosen this approach, instead of looking for alternative batches, to keep the batch derivation simple (i.e. allow multiple batches to be queued up with the eager-batch-derivation logic, instead of having to check them 1 by 1 with the engine).

When the typed error handling lands we can handle the deposit processing error as critical, and stop the syncing entirely. Safety over liveness, deposits always have to work.

During goerli devnet testing there was a batch with invalid nonce usage, and so we have to drop the user transaction(s) from that batch to continue rollup processing, or we get stuck. This is attributable to the batch-submitter, who should not have included the invalid transaction. So we drop the batch transactions as a whole, instead of finding a single invalid transaction, to keep things simple.

### DIFF
--- a/op-node/rollup/derive/engine_queue.go
+++ b/op-node/rollup/derive/engine_queue.go
@@ -3,6 +3,8 @@ package derive
 import (
 	"context"
 	"fmt"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
 	"io"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mslipper (approved):
LGTM, just fix the lint error.

PR #3182, opened 2022-08-05
ci: used fixed foundry in ci

tynes:
**Description**
Installs git in the devnet ci to fix https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/5500/workflows/510e0d29-2a9d-4b74-9cae-515b697e8892/jobs/91276

Pushed a commit in `contracts-bedrock` simply fixing a typo in the readme so that the devnet ci step will be triggered

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -439,7 +439,7 @@ jobs:
       - run:
           name: foundryup
           command: |
-            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -- -y
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
             curl -L https://foundry.paradigm.xyz | bash
             source $HOME/.bashrc
             foundryup --pr 2634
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -439,9 +439,11 @@ jobs:
       - run:
           name: foundryup
           command: |
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+            export CARGO_NET_GIT_FETCH_WITH_CLI=true
             curl -L https://foundry.paradigm.xyz | bash
             source $HOME/.bashrc
-            foundryup
```
tynes, line 444: If we could cache this build, it would solve our problems for now

.circleci/config.yml
```diff
@@ -439,9 +439,11 @@ jobs:
...
+            foundryup --pr 2634
```
protolambda, line 446: They merged the PR, maybe we can stick to the old approach. Although freezing foundry to a stable version may still be useful.

protolambda (approved):

tynes:
Closing this PR because it was merged

PR #3184, opened 2022-08-05
op-node: event backoff (extends #3130)

protolambda:
Suggestion on top of the work of #3130: This implements the derivation step backoff without blocking the other event handling.

### DIFF
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -293,8 +293,15 @@ func (s *state) eventLoop() {
 		}
 	}
 
-	// reqStep requests a derivation step to be taken. Won't deadlock if the channel is full.
-	reqStep := func() {
+	// channel, nil by default (not firing), but used to schedule re-attempts with delay
+	var delayedStepReq <-chan time.Time
+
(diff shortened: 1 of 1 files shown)


op-node/rollup/driver/state.go
```diff
@@ -331,6 +349,7 @@ func (s *state) eventLoop() {
 			if err != nil {
 				s.log.Error("Error creating new L2 block", "err", err)
 				s.metrics.DerivationErrorsTotal.Inc()
+				break // if we fail, we wait for the next block creation trigger.
```
protolambda, line 352: Noticed we were ignoring this error, and quickly rebuilding the block upon failure. Sequencing is important to not delay by as much as an exponential backoff, but on an error I think it's fair to wait till the next block-building trigger to re-attempt (it will catch up if there are no errors again). So I'm adding this `break` while we're working on the event hub anyway.

tuxcanfly: OK, I see. Yeah I saw that there are multiple triggers that request next step, triggering block-building.

PR #3187, opened 2022-08-08
feat(contracts-bedrock,sdk): deploy config typing, hivenet

protolambda:
This PR:
- Fixes the deploy config type spec and defaults to cover all variables again. Previously some deploy configs were just adding things without documenting/typing them.
- Updates deploy config spec to handle L2 output oracle more nicely:
  - The l2 timestamp can be configured, but if negative, it will use the l1 starting block tag (like it did before), assuming the l2 chain config and rollup config scripts use this for generating a starting l2 timestamp too.
- Fix deploy config correctly configure clique signer. Previously it was putting the extra-data into the L2 genesis, and hardcoding the other thing in L1 genesis. That breaks the max l2 extra-data length, as well as being plain misleading.
- Add rollup-config, l1 genesis block, and l2 genesis block config to deploy config spec. With a note that these should only be used by the respective config generating scripts. These have defaults, and don't have to be defined by deploy-configs which are made for existing networks (as opposed to those that create full new L1/L2 networks)
- Update L1Block info predeploy contract contents in L2 genesis to use contents of starting L1 block, to not duplicate configuration more than necessary.
- Fix ops-bedrock: the L1 config generation was not specifying the hardhat network correctly.
- Support "hivenet" deploy config:
  - This deploy config loads its contents from an adjacent json file `deploy-config/hivenet_data.json`, which Hive can insert before continuing to use hardhat for configs/deployments.
  - Added a reference to hivenet in the sdk. Hoping to get sdk testing working in Hive.

Note: couldn't test hivenet in Hive yet, it's generating the chain configs very well, but it gets stuck on contract artifacts, since I think forge doesn't work in the docker image due to GLIB issues. I asked @mslipper to update `ethereumoptimism/ci-builder:latest` from Debian 10 to Debian 11, that should fix it.


mslipper:
FYI this will impact the regenesis deployer tool. I can update it.

packages/contracts-bedrock/deploy-config/hivenet.ts
```diff
@@ -0,0 +1,6 @@
+import hivenetData from "./hivenet_data.json"
```
tynes, line 1: You should be able to rename `hivenet_data.json` to `hivenet.json` and it should just work, see https://github.com/ethereum-optimism/optimism/blob/ac576a7e3caa552bf19968a18f30d74a485a904e/packages/hardhat-deploy-config/src/plugin.ts#L36

packages/contracts-bedrock/hardhat.config.ts
```diff
@@ -35,6 +35,11 @@ const config: HardhatUserConfig = {
         'ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80',
       ],
     },
+    hivenet: {
+      chainId: Number(process.env.CHAIN_ID),
```
tynes, line 39: Does hive use a particular chainid or is it nice to be able to configure it? I don't love that this file uses so many env vars, I always find myself commenting out lines because env vars are not defined

packages/contracts-bedrock/tasks/genesis-l1.ts
```diff
@@ -83,20 +77,26 @@ task('genesis-l1', 'create a genesis config')
...
+        ethers.utils.hexZeroPad('0x', 65),
+      ]),
+      gasLimit: deployConfig.l1GenesisBlockGasLimit,
+      difficulty: deployConfig.l1GenesisBlockDifficulty,
+      mixHash: deployConfig.l1GenesisBlockMixHash,
+      coinbase: deployConfig.l1GenesisBlockCoinbase,
       alloc,
+      number: deployConfig.l1GenesisBlockNumber,
+      gasUsed: deployConfig.l1GenesisBlockGasUsed,
+      parentHash: deployConfig.l1GenesisBlockParentHash,
```
tynes, line 98: Parent hash should always always be `ethers.constants.HashZero` right?

packages/contracts-bedrock/tasks/genesis-l1.ts
```diff
@@ -83,20 +77,26 @@ task('genesis-l1', 'create a genesis config')
...
+        ethers.constants.HashZero,
+        deployConfig.cliqueSignerAddress,
+        ethers.utils.hexZeroPad('0x', 65),
+      ]),
+      gasLimit: deployConfig.l1GenesisBlockGasLimit,
+      difficulty: deployConfig.l1GenesisBlockDifficulty,
+      mixHash: deployConfig.l1GenesisBlockMixHash,
+      coinbase: deployConfig.l1GenesisBlockCoinbase,
       alloc,
+      number: deployConfig.l1GenesisBlockNumber,
```
tynes, line 96: Number and gas used can be configurable for the genesis block? Do values other than 0 make sense here?

packages/contracts-bedrock/tasks/genesis-l2.ts
```diff
@@ -166,14 +166,14 @@ const replaceImmutables = async (
...
       // Ensure that the value being sliced out is 0
-      const val = hexDataSlice(
-        deployedBytecode,
-        offset.start,
-        offset.start + offset.length
-      )
-      if (!BigNumber.from(val).eq(0)) {
-        throw new Error(`Unexpected value in immutable bytecode ${val}`)
-      }
+      // const val = hexDataSlice(
```
tynes, line 169: Is this causing a bug? If so, something is off

packages/contracts-bedrock/tasks/genesis-l1.ts
protolambda: Not if we start doing migration tests with previous chain history. But generally, yes.

packages/contracts-bedrock/tasks/genesis-l2.ts
protolambda: Yes, forgot, this doesn't work in Hive for some reason. The contents are non-zero.

tynes: I will need to investigate this, it would lead me to believe there is a bug in the way immutables are handled then. Weird that it works for non hive deployments..

packages/contracts-bedrock/tasks/genesis-l1.ts
tynes: I see your above comment. I didn't think geth would just work if we created a genesis block with a non 0 number

packages/contracts-bedrock/tasks/genesis-l2.ts
tynes: It makes me think the artifacts are out of date

tynes:
Could you provide instructions on how you have been running hive with this? Or add something to CI so that it runs hive

protolambda:
@tynes hope this helps. Happy to pair program on a call to debug hardhat in Hive.
```
# get a dev install of hive
git clone git@github.com:ethereum-optimism/hive.git
cd hive
# important: checkout the refactored work
git checkout test-setup-refactor
go build -o hive .

# make sure docker is running
sudo systemctl start docker

# run the hive coordinator, load all the clients that may be involved.
# run it in dev mode, so we don't specify any simulator
# note: the _hivenet suffix forces the op-contracts client to use `hivenet` as branch argument in docker image build, which translates into pulling that monorepo branch.
./hive --client=go-ethereum,op-geth,op-proposer,op-batcher,op-node,op-contracts_hivenet --docker.output -loglevel=5 -dev


# in second terminal:
# instead of making hive load and run the simulator with the --sim option as a docker container, we run it as go binary, so we can debug it with breakpoints etc.
# go into simulator go-module. Note: the go workspace in hive repository root includes the `optimism` dir in hive, which provides all the testnet building blocks common between different optimism test simulators
cd simulators/optimism/testnet
# start the simulator
go run .

# in `workspace/logs` it will collect logs of each of the containers
# I'd recommend to set a breakpoint (or add a time.Sleep()) before the simulator exits on a failed script execution, see `optimism/hardhat.go` `RunScript` function.
# This way you can `docker exec -it <containername> /bin/bash` to go into the running op-contracts container to experiment with hardhat commands and inspect the contracts artifacts build etc.

File: packages/contracts-bedrock/hardhat.config.ts [hunk 6]
```diff
@@ -60,27 +65,275 @@ const config: HardhatUserConfig = {
...
+      type: 'string', // uint64
+      default: '0x0',
+    },
+    l2GenesisBlockParentHash: {
+      type: 'string', // bytes32
+      default: ethers.constants.HashZero,
+    },
+    l2GenesisBlockBaseFeePerGas: {
+      type: 'string', // uint256
+      default: ethers.BigNumber.from(1000_000_000).toHexString(), // 1 gwei
```
tynes, line 294: https://github.com/ethers-io/ethers.js/blob/ce8f1e4015c0f27bf178238770b1325136e3351a/packages/units/src.ts/index.ts#L64

`ethers.utils.formatUnits(1, 'gwei')`

ops-bedrock/devnet-up.sh
```diff
@@ -69,7 +69,7 @@ if [ ! -f ./.devnet/genesis-l1.json ]; then
   echo "Regenerating L1 genesis."
   (
     cd $CONTRACTS_BEDROCK
-    L1_GENESIS_TIMESTAMP=$L1_GENESIS_TIMESTAMP npx hardhat genesis-l1 \
+    L1_GENESIS_TIMESTAMP=$L1_GENESIS_TIMESTAMP npx hardhat --network $NETWORK genesis-l1 \
```
tynes, line 72: I think we should get rid of the env var here and add a flag to the task itself so that we can migrate away from stateful config files that parse env vars

packages/contracts-bedrock/deploy-config/devnetL1.ts
```diff
@@ -2,58 +2,46 @@ import { ethers } from 'ethers'
 
 const { env } = process
 
-const l1GenesisTimestamp =
+const l1GenesisBlockTimestamp =
```
tynes, line 5: We should be able to get rid of this and then parse the value from a flag passed to the hardhat task, we could use the `Math.floor(Date.now() / 1000)` as the default value in the task.

Something like:
if flag passed, use that value, otherwise use the value from the config

packages/contracts-bedrock/hardhat.config.ts
protolambda: The L1 RPC env var definitely changes, so I'm inclined to keep the other env vars, as the env is being configured anyway.

packages/contracts-bedrock/tasks/genesis-l1.ts
protolambda: It has a default value in some place, so you don't have to set it in the actual deploy config, I'm leaving this as-is so we've control over genesis data if we need to.

packages/contracts-bedrock/tasks/genesis-l2.ts
protolambda: Fixed, problem with deploy/config order, thanks for your help yesterday :+1:

packages/contracts-bedrock/hardhat.config.ts
protolambda: geth and some others are flexible with changing between hex and decimal values for uint64 strings, but I don't know if every client is. Sticking to hex formatting here.

packages/contracts-bedrock/deploy-config/devnetL1.ts
protolambda: done

ops-bedrock/devnet-up.sh
protolambda: done

packages/contracts-bedrock/tasks/genesis-l2.ts
protolambda: Unresolving this. It's happening in devnet too.

protolambda: Will rebuild and try again, this is weird

protolambda: Still happens in devnet with a rebuild.
Getting:
```
An unexpected error occurred:

Error: Unexpected value in immutable bytecode, contract L2ToL1MessagePasser ast 22099 key MAJOR_VERSION path contracts/L2/L2ToL1MessagePasser.sol >  id 2517 >inherit contracts/universal/Semver.sol id 22158 >  id 22157 >  id 22099: 0x617a5ac1cc7eba2d6f37570a0135afa950d8bb77cdd35f0d0b4e85a16f906000
    at replaceImmutables (/home/proto/projects/optimism/packages/contracts-bedrock/tasks/genesis-l2.ts:181:15)
    at async SimpleTaskDefinition.action (/home/proto/projects/optimism/packages/contracts-bedrock/tasks/genesis-l2.ts:467:11)
    at async Environment._runTaskDefinition (/home/proto/projects/optimism/node_modules/hardhat/src/internal/core/runtime-environment.ts:219:14)
    at async Environment.run (/home/proto/projects/optimism/node_modules/hardhat/src/internal/core/runtime-environment.ts:131:14)
    at async main (/home/proto/projects/optimism/node_modules/hardhat/src/internal/cli/cli.ts:218:5)
make: *** [Makefile:59: devnet-up] Error 1

```
(added ast info to the error log)
So it looks like the MAJOR_VERSION immutable var is not zero in the bytecode.
And I'm surprised it has this large random looking string, and not a version number like the name indicates.

protolambda: Going to clear all caches and rebuild, maybe some bytecode is not correct

protolambda: deleting caches and rebuilding worked, oh well

packages/contracts-bedrock/hardhat.config.ts
tynes: I meant to populate `BigNumber.from` to ensure that the math is correct

packages/contracts-bedrock/deploy-config/devnetL1.ts
```diff
@@ -1,59 +1,37 @@
...
-    : Math.floor(Date.now() / 1000)
-
 const config = {
-  submissionInterval: 20,
-  genesisOutput: ethers.constants.HashZero,
-  historicalBlocks: 0,
+  // general
   l1StartingBlockTag: 'earliest',
-  startingBlockNumber: 0,
+  l1ChainID: 901,
```
tynes, line 4: These should be 900 and 901, they somehow changed

ops-bedrock/devnet-up.sh
```diff
@@ -84,6 +84,13 @@ fi
   wait_up $L1_URL
 )
 
+echo "Waiting for L1 RPC to be responsive."
```
tynes, line 87: This is what `wait_up` does right?

ops-bedrock/devnet-up.sh
tynes, line 72: Passing through the env var here can be removed and we can use the flag instead

ops-bedrock/devnet-up.sh
protolambda: missed that, thanks

ops-bedrock/devnet-up.sh
protolambda: Don't even need to use the flag. It'll generate a timestamp as part of the task. And if the l1 genesis already exists, we don't run the genesis l1 task anyway, so no need to replicate the timestamp.

tynes (approved):

packages/contracts-bedrock/deploy-config/hivenet.json
```diff
@@ -0,0 +1 @@
+{}
```
tynes, line 1: Is this meant to be empty?

protolambda: Yes, in Hive we swap in a full config json. I thought it was awkward to reference the network in hardhat config / sdk / etc. and not add a placeholder file. So I added this empty thing.

mslipper (approved):

mslipper:
I've verified that the bedrock go tests work. CircleCI is having issues right now, so I don't want Circle's issues to block this.

PR #3188, opened 2022-08-08
op-node: Bedrock l1 cache metrics

protolambda:
This PR:
- Wraps a LRU cache with metrics
- Defines cache metrics as a group of metrics
- Updates the L1 source cache to use the cache with metrics
- Updates the L1 source cache with increased defaults: this should help respond much faster to L1 reorgs, replaying the old data from the cache instead of fetching everything again. On the testnet we see that a L1 reorg that is past the confirmation depth takes up to 3 minutes to resolve. This aims to fix that, and provide metrics so we can watch it.

Note: I moved the cache wrapper into `op-node/source/caching` package.  I'm thinking we can move `op-node/{l1,l2}` into `op-node/source/{l1,l2}`. And then later update `l2` with the same caching too (see ENG-2280)

Fix ENG-2555

### DIFF
--- a/op-node/l1/source.go
+++ b/op-node/l1/source.go
@@ -4,14 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/ethereum-optimism/optimism/op-node/sources/caching"
 
 	"github.com/ethereum-optimism/optimism/op-node/client"
-	"github.com/ethereum/go-ethereum"
-
-	"github.com/ethereum-optimism/optimism/op-node/rollup"
-
 	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum-optimism/optimism/op-node/sources/caching"
+	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
--- a/op-node/l1/source.go
+++ b/op-node/l1/source.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/ethereum-optimism/optimism/op-node/sources/caching"
 
 	"github.com/ethereum-optimism/optimism/op-node/client"
 	"github.com/ethereum/go-ethereum"
@@ -16,7 +17,6 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rpc"
-	lru "github.com/hashicorp/golang-lru"
 )
 
 type SourceConfig struct {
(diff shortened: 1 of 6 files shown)


op-node/metrics/metrics.go
```diff
@@ -37,6 +37,9 @@ type Metrics struct {
 	RPCClientRequestDurationSeconds *prometheus.HistogramVec
 	RPCClientResponsesTotal         *prometheus.CounterVec
 
+	L1SourceCache *CacheMetrics
+	// TODO: L2SourceCache *CacheMetrics
```
tynes, line 41: Is this a dead comment?

op-node/sources/caching/cache.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// LRUCache wraps hashicorp *lru.Cache and tracks cache metrics
+type LRUCache struct {
+	m     Metrics
+	label string
+	inner *lru.Cache
+}
+
+func (c *LRUCache) Get(key interface{}) (value interface{}, ok bool) {
```
tynes, line 17: Curious what you think about using `any` vs `interface{}` here

protolambda: I just copied the interface that the underlying cache uses. I'm fine with `any` too.

op-node/metrics/metrics.go
protolambda: Not dead, it's a hint for the implementer of ENG-2555 (adding caching to L2 source bindings)

tynes (approved):

op-node/l1/source.go
```diff
@@ -70,16 +68,15 @@ func (c *SourceConfig) Check() error {
...
-		ReceiptsCacheSize: 20,
-
-		// Optimal if at least a few times the size of a sequencing window.
-		// When smaller than a window, requests would be repeated every window shift.
-		// Additional cache-size for handling reorgs, and thus more unique blocks, also helps.
-		TransactionsCacheSize: int(config.SeqWindowSize * 4),
-		HeadersCacheSize:      int(config.SeqWindowSize * 4),
+		ReceiptsCacheSize:     span * 400,
+		TransactionsCacheSize: span * 400,
+		HeadersCacheSize:      span,
```
tuxcanfly, line 79: Since we're allocating 1 header and 400 tx per block, I guess the previous header cache size was very generous? As it was same as tx cache size.

tuxcanfly (commented):
OK.

op-node/sources/caching/cache.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// LRUCache wraps hashicorp *lru.Cache and tracks cache metrics
+type LRUCache struct {
+	m     Metrics
+	label string
+	inner *lru.Cache
+}
+
+func (c *LRUCache) Get(key any) (value any, ok bool) {
```
trianglesphere, line 17: I know the underlying cache is using `any`, but would we able to switch to generics here & enforce some amount of type safety? I we don't mix types in our caches

tuxcanfly (approved):
OK.

op-node/l1/source.go
protolambda: You can read the `span` as number of blocks to retain in the cache, with 400 txs per block to cache (mainnet L1 includes between about 0-300 txs per block). The span includes a full sequencing window, up to a sanity limit, and with some margin, to buffer enough to make processing a reorg fast.

op-node/sources/caching/cache.go
protolambda: Tried, it requires casting the inner LRU value type back from an interface to the `V` type param, which Go doesn't support. Switching to another LRU lib that supports generics isn't a great option either: the most popular libs I could find on github wrap the LRU contents more naively with sync methods, or don't expose the evictions / cache size. The library we use now is more mature w.r.t. these features. So I'm going to stick with that. We can switch LRU to generics in a different PR.

PR #3193, opened 2022-08-08
ci: Use faucet in actor tests

mslipper:
To improve the concurrency of the Bedrock actor tests, I hooked up an ETH faucet. Now, the actors can use the faucet to fund themselves before each run.

### DIFF
--- a/packages/actor-tests/bedrock/deposit-eth.ts
+++ b/packages/actor-tests/bedrock/deposit-eth.ts
@@ -1,4 +1,4 @@
-import { utils, Wallet, providers, constants } from 'ethers'
+import { utils, Wallet, constants } from 'ethers'
 import {
   CrossChainMessenger,
   ETHBridgeAdapter,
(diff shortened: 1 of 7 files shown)


packages/actor-tests/lib/faucet.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class Faucet {
+  private url: string
+
+  private provider: providers.Provider
+
+  constructor(url: string, provider: providers.Provider) {
+    this.url = url
+    this.provider = provider
+  }
```
roninjin10, line 12: Don't change this but in case you didn't know, you can write this with way less boilerplate

```typescript
constructor(
  private url: string,
  private provider: providers.Provider,
) {}
```

tynes (approved):

tynes:
I haven't seen these run, but they look good to me

packages/actor-tests/bedrock/deposit-eth.ts
```diff
@@ -1,71 +1,74 @@
-import { utils, Wallet, providers, constants } from 'ethers'
+import { utils, Wallet, constants } from 'ethers'
 import {
   CrossChainMessenger,
   ETHBridgeAdapter,
   StandardBridgeAdapter,
 } from '@eth-optimism/sdk'
-import { predeploys } from '@eth-optimism/contracts-bedrock'
+import { predeploys } from '@eth-optimism/contracts-bedrock/dist'
```
protolambda, line 7: Does this not work without `dist`?

protolambda (commented):
LGTM, but wondering about one imports `dist` change, is that correct?

packages/actor-tests/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "resolveJsonModule": true
   },
   "include": [
-    "./lib",
+    "./lib/**/*.ts",
```
roninjin10, line 7: This is wierd why is this necessary?

mslipper: I was seeing compilation warnings.

roninjin10: Odd.   I got no guesses on how this change would change the behavior of typescript

packages/actor-tests/bedrock/deposit-eth.ts
mslipper: Nope, it didn't work without dist :(.

mslipper: Maybe @tynes knows how to fix it? It was trying to find an index file that didn't exist.

tynes: Try updating to the latest release? https://github.com/ethereum-optimism/optimism/blob/6f66c71f4163dc28f142ef87285631f4b64d2f05/packages/contracts-bedrock/package.json#L3

Your `package.json` refers to an older release. Other packages don't need to import from dist like this when working with this package?

tynes:
If you rebase then it will get rid of the yarn.lock diff

mslipper:
Will rebase + upgrade the package.

packages/actor-tests/bedrock/deposit-eth.ts
mslipper: Fixed!

protolambda (approved):

PR #3194, opened 2022-08-09
op-node: update to latest libp2p, update deps, Go 1.19 linting [bedrock]

protolambda:
This fully updates the op-node dependencies, including an update from `libp2p` `v0.18.1` to `v0.21.0`, and `libp2p-core` `v0.15.0` to `v0.19.1`.

In this range of libp2p updates a lot of packages were moved into their monorepo, and the libp2p crypto package uses a version of btcec now that wraps the decred crypto lib (much faster secp256k1 in Go). This required some updates in the way we configure libp2p and load/handle p2p keys.

These updates help move away from indirect dependencies that were broken in Go 1.19, enabling the newest Go ci linting to work again (I was unable to run old lint tooling on Go 1.19 due to the dependency error) + Go 1.19 linting is fixed (they have some doc format rules now).

Fix ENG-2556


tynes:
For some reason, the devnet is flaky - I was not experiencing this before but now it sometimes hangs

protolambda:
rebased, hope devnet CI is less flaky now

op-node/p2p/discovery.go
```diff
@@ -11,6 +11,8 @@ import (
 	"net"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec/v2"
+	decredSecp "github.com/decred/dcrd/dcrec/secp256k1/v4"
```
tynes, line 15: Why decred, btcd and geth ec?

protolambda: Summary of the secp256k1 Go crypto situation:
- libp2p updated from btcd/btcec to btcd/btcec v2
- btcd/btcec v1 wrapped Go ecdsa
- btcd/btcec v2 wraps decred
- decred is faster, and thus btcd updated
- btcd earlier also went from monolithic releases to per-module go releases, and had versioning mistakes in v1. These were fixed with beta versions / new releases eventually, but end-users started looking to update to v2.
- geth uses a C library for secp256k1, and uses Go ecdsa curve types to integrate that into Go (but use their own instance of the curve definitions, same values, but not same memory address, making shallow equality checks not work...)
- decred has functions to convert to Go ecdsa types
- After type casting the general libp2p pubkey/privkey interface to a concrete type, it can be casted to the underlying types (btcd, decred) to use their functionality to convert to Go ecdsa, and then you can swap the curve instance to point to the one by geth for the shallow equality checks in the Geth discovery v5 package to work.

tynes (approved):

mslipper (approved):

PR #3195, opened 2022-08-09
contracts-bedrock: use `MemorySafeCall`

tynes:
**Description**

Replace `ExcessivelySafeCall` with a local `SafeCall`
that does not copy any calldata back. We do not care about
returndata being returned so this prevents returndata bombs.


tynes:
Hoping that https://github.com/ethereum-optimism/optimism/pull/3199 fixes the issues we are seeing in CI when it hangs sometimes

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -284,7 +285,7 @@ abstract contract StandardBridge is Initializable {
         require(_to != address(this), "StandardBridge: cannot send to self");
 
         emit ETHBridgeFinalized(_from, _to, _amount, _extraData);
-        (bool success, ) = _to.call{ value: _amount }(new bytes(0));
+        bool success = SafeCall.call(_to, gasleft(), _amount, hex"");
```
maurelian, line 288: Is there a good reason we didn't have ESC here before?

tynes: No good reason, I feel like it should be here

mslipper:
Is this still necessary? I think we've since fixed the SDK, so we don't need to use SafeCall instead anymore.

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
```diff
@@ -0,0 +1,55 @@
...
+        // don't call the console
+        vm.assume(
+            to != address(0x000000000000000000636F6e736F6c652e6c6f67)
+        );
+        // don't call the create2 deployer
+        vm.assume(
+            to != address(0x4e59b44847b379578588920cA78FbF26c0B4956C)
+        );
+        // don't send funds to self
+        vm.assume(from != to);
```
maurelian, line 31: Does something actually break if without these?

maurelian:
> Is this still necessary? I think we've since fixed the SDK, so we don't need to use SafeCall instead anymore.

@mslipper I talked to @tynes and he convinced me it's still beneficial.
@tynes: Can we rename it to `MemorySafeCall`? The `Safe` prefix has been overused and lacks specificity.

Otherwise I'm good with the change.

tynes:
I've rebased this and then updated the name to `MemorySafeCall`

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
tynes: Yes, things revert in unexpected ways when you call the special addresses and the balance assertions fail if you send funds to yourself

smartcontracts (approved):

tynes:
@smartcontracts Just rebased and fixed the merge conflict

maurelian (approved):

packages/contracts-bedrock/package.json
```diff
@@ -55,6 +53,7 @@
     "@foundry-rs/easy-foundryup": "^0.1.3",
     "@nomiclabs/hardhat-ethers": "^2.0.0",
     "@nomiclabs/hardhat-waffle": "^2.0.0",
+    "@rari-capital/solmate": "https://github.com/rari-capital/solmate.git#8f9b23f8838670afda0fd8983f2c41e8037ae6bc",
```
tynes, line 56: Moves solmate to devdeps bc its at a commit hash right now

op-bindings/bindings/weth9.go
```diff
@@ -31,7 +31,7 @@ var (
...
 }
```
tynes, line 35: The old weth9 has:
- bed565cbea90624f5b56e2f56c4d5fc05493e40220bba9bd15ace6aa4cabb51c
The new weth9 has:
- 2282f823392ab6b42c17b31235784e88e108cd8d10e77cd732aa3e1e7a88a7d

This looks like a bytecode hash

The diff is right at the very end

PR #3196, opened 2022-08-09
Fix Readme with proper links and added some great stylings

Developer-piyush:
**Description**
I have fixed and added proper links to readme with Table of contents, Also added some symmetry styling to the whole content.

**Metadata**
- To see live results of fixed: head over to forked repo:
- https://github.com/Developer-piyush/optimism/blob/develop/README.md


Developer-piyush:
:)

PR #3198, opened 2022-08-09
deps: fix yarn.lock

tynes:
**Description**

It got out of date, this updates it. The changes
are very minimal.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2911,11 +2911,6 @@
   dependencies:
     squirrelly "^8.0.8"
 
-"@rari-capital/solmate@git+https://github.com/rari-capital/solmate.git#8f9b23f8838670afda0fd8983f2c41e8037ae6bc":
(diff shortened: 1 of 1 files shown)


maurelian:
Doesn't this check fail if the lockfile is out of date? 

https://github.com/ethereum-optimism/optimism/blob/c003bbf6615425e409617794f2223418683d5c97/.circleci/config.yml#L32

tynes:
> Does this command not fail when the lockfile is out of date?
> 
> https://github.com/ethereum-optimism/optimism/blob/c003bbf6615425e409617794f2223418683d5c97/.circleci/config.yml#L32

It looks like this is a yarn v2 feature but not yarn v1

tynes:
I tried updating to yarn v3 but it doesn't support dependencies that do not have a package.json - the solidity packages that install via github don't have them

PR #3200, opened 2022-08-09
ops: update ci-builder

tynes:
**Description**

geth to v1.10.21
go to 1.19
golangci-lint to 1.48.0


tynes:
Will need to trigger a release to see this run in action

protolambda:
Not sure if it'll catch the CI failures, since this doesn't touch the actual Go code some CI checks might not be running. So implicitly this depends on #3194 I think. (latest linter without those changes doesn't work for me locally)

mslipper (approved):

PR #3202, opened 2022-08-10
op-batcher op-proposer op-service: Introduce op-service

mslipper:
op-service is a utilities library that automates much of the boilerplate setup we've been duplicating across different daemons. This PR adds the following to op-batcher:

- Common setup utilities for logging, metrics, pprof, and RPC.
- For each of the above functions, a set of CLI flags and configs that can be used to parse + validate CLI configuration for that module.
- A base RPC server that implementers can extend.
- HTTP metrics utilities.

Fix ENG-2309
Fix ENG-2365
Fix ENG-2366
Fix ENG-2367


protolambda:
Added `replace` statements to the go mod files to find the `op-service` module while it's not canonical yet. That way go mod tidy can pass and CI can run.

protolambda (approved):
This is amazing, a much needed cleanup of logging/metrics/rpc setup.

In a different PR I think I'll move the `testlog` package to `op-service` so it doesn't have to live in `op-node` anymore.

After it's merged we can get rid of the `replace` statements, and after this + refactor hivenet PR + Go 1.19 PRs it's time for new release tags of the bedrock go modules, so we can do that too.

op-batcher/flags/flags.go
```diff
@@ -101,39 +105,11 @@ var (
 		Required: true,
 		EnvVar:   prefixEnvVar("SEQUENCER_BATCH_INBOX_ADDRESS"),
 	}
+)
 
-	/* Optional Flags */
+func init() {
```
tynes, line 111: Is this meant to be an empty `init` function?

op-batcher/go.mod
```diff
@@ -2,12 +2,15 @@ module github.com/ethereum-optimism/optimism/op-batcher
...
+replace github.com/ethereum-optimism/optimism/op-service v0.0.0 => ../op-service
+
 require (
 	github.com/ethereum-optimism/optimism/op-node v0.3.0
 	github.com/ethereum-optimism/optimism/op-proposer v0.3.0
+	github.com/ethereum-optimism/optimism/op-service v0.0.0
 	github.com/ethereum/go-ethereum v1.10.21
 	github.com/miguelmota/go-ethereum-hdwallet v0.1.1
-	github.com/urfave/cli v1.22.5
+	github.com/urfave/cli v1.22.9
```
tynes, line 13: Perhaps out of scope of this PR, but upstream has migrated to `urfave/cli` v2
https://github.com/ethereum/go-ethereum/blob/c0cc6f63629678ad7a8d9a33dc17bcf26122bde0/go.mod#L58

op-batcher/flags/flags.go
mslipper: Removed.

op-batcher/go.mod
mslipper: This'll require a much larger refactor - can take care of that separately.

op-service/httputil/wrapped_response_writer.go
```diff
@@ -0,0 +1,38 @@
+package httputil
+
+import "net/http"
+
```
tynes, line 4: I really liked this article on hn yesterday https://preslav.me/2022/08/09/i-dont-like-golang-default-http-handlers/

I like the idea of returning errors in http handlers like echo vs just returning early if there is an error

This seems like it would be a headache in the amount of changes to make, but they do give an example on how to implement this without using echo.

Thoughts?

mslipper: I like what that article suggests, though this particularly piece of code is unrelated to that. We'll still need a `ResponseWriter`, and the wrapper tracks response size and status code so we can log them/add them to metrics.

mslipper: If we want to convert all of our HTTP handlers to this pattern we should bring it up in OH and agree on it together.

op-batcher/flags/flags.go
```diff
@@ -1,13 +1,17 @@
...
+	opmetrics "github.com/ethereum-optimism/optimism/op-service/metrics"
+	oppprof "github.com/ethereum-optimism/optimism/op-service/pprof"
+	oprpc "github.com/ethereum-optimism/optimism/op-service/rpc"
 	"github.com/urfave/cli"
 )
 
-const envVarPrefix = "BATCH_SUBMITTER_"
+const envVarPrefix = "BATCH_SUBMITTER"
 
 func prefixEnvVar(name string) string {
```
tynes, line 13: This can be replaced with the helper in the `op-service`

op-proposer/flags/flags.go
```diff
@@ -1,13 +1,17 @@
...
+	opmetrics "github.com/ethereum-optimism/optimism/op-service/metrics"
+	oppprof "github.com/ethereum-optimism/optimism/op-service/pprof"
+	oprpc "github.com/ethereum-optimism/optimism/op-service/rpc"
 	"github.com/urfave/cli"
 )
 
-const envVarPrefix = "OUTPUT_SUBMITTER_"
+const envVarPrefix = "OUTPUT_SUBMITTER"
 
 func prefixEnvVar(name string) string {
```
tynes, line 13: This can be replaced with the helper from `op-service`

tynes (approved):

PR #3205, opened 2022-08-10
bedrock: v0.4.0

protolambda:
updates:
- use latest reference-optimistic-geth
- new releases for: op-service, op-bindings, op-node, op-proposer, op-batcher, op-e2e

Marking this as `do-not-merge` so mergify does not touch the commits that were tagged to create the incremental releases (each v0.4.0 module only has v0.4.0 dependencies, starting with the modules that have no monorepo dependencies). Just need to avoid squashing and merge these using a regular merge-commit.


tynes:
I feel like the docker builds should still work, but just pointing it out so that we can be sure that doesn't break. We should also publish docker images based on the tag and do github releases based on the tag, but that doesn't need to happen now

PR #3206, opened 2022-08-11
state-surgery: hardhat utils

tynes:
**Description**

Add hardhat utils for reading hardhat artifacts
from disk. This can read hardhat artifacts, buildinfo
and hardhat deploy artifacts. This is necessary for
implementing the state surgery in go, because these
files must be consumed as part of the state surgery process.


mslipper (approved):
LGTM - I don't have a strong opinion about the big build-info file. Defer to you/Proto on that.

protolambda (approved):

PR #3207, opened 2022-08-11
op-bindings: add rest of predeploys

tynes:
**Description**

Add the rest of the predeploy contracts to `op-bindings`
so that they can be used as part of the state surgery process.


op-bindings/Makefile
```diff
@@ -1,42 +1,65 @@
 SHELL := /bin/bash
 
+pkg := bindings
```
maurelian, line 3: what is this for?

tynes: Its defining a variable that is used as an arg to each of the scripts called

maurelian: is it a default that can be overridden?

tynes: Its a hardcoded variable like `const foo = 'bar'` in JS, i believe that for vars referenced in `$(..)`, they can be set at runtime via cli flag but we are hardcoding the value because its generally static for us

protolambda (approved):

maurelian (approved):

PR #3209, opened 2022-08-11
contracts-bedrock: latest goerli deployment

tynes:
**Description**
Latest goerli deployment


optimisticben:
We're going to have multiple deployments simultaneously, yeah?

Should we have the contracts for each deployment, or just the most recent, or some canonical deployment?

tynes:
> We're going to have multiple deployments simultaneously, yeah?
> 
> Should we have the contracts for each deployment, or just the most recent, or some canonical deployment?

Hmm good question, its going to be a pain to keep adding these, if we have multiple at the same time then the "right" way to do it would be to update the hh config to add a new network and then commit each set of artifacts. That seems like a pain and would reduce velocity so I lean towards just keeping the one that people are using the most the best that we can up to date in here but not the end of the world if it falls behind.

If there are contract changes, then it would def be nice to update here so that we can reverify on etherscan

optimisticben:
Can hh config take a url? Then we can have a way to reference network configurations that aren't in the git repo.

maurelian:
> Can hh config take a url?

It's a JS file so should be possible

mslipper:
👍 on making this take a URL. We will have lots of Goerli deployments, and committing them will become increasingly time consuming.

maurelian (approved):

mslipper:
Do we still need this PR? We're going to have lots of Goerli networks, and I don't think that we should commit these networks to the repo until we have a stable public devnet.

PR #3216, opened 2022-08-13
feat: storage slot setting + memory db

tynes:
**Description**
Implements basic storage slot setting. The storage slot setting still needs to support mappings and slot offsets and more thorough testing, then it will be ready to be used for the new state surgery pipeline.

Also implements a `MemoryStateDB` that acts on a `core.Genesis` so that we can build a fresh genesis from scratch to replace the hardhat task. This will allow us to use the same code used to generate the L2 `genesis.json` on an actual LevelDB backed `StateDB` instance.

Closes ENG-2586


state-surgery/state/state.go
```diff
@@ -0,0 +1,196 @@
...
+type StorageValues map[string]any
+
+// EncodedStorage represents the storage key and value serialized
+// to be placed in Ethereum state.
+type EncodedStorage struct {
+	Key   common.Hash
+	Value common.Hash
+}
+
+// EncodedStorage will encode a storage layout
```
tynes, line 27: Going to follow up with a PR that adds handling of offsets + mappings, those are the only required things to be able to fully run on our L2 contracts based on the output of this script:

```bash
#!/bin/bash

for file in contracts/L2/*; do
    fqn="$file:$(basename $file | cut -d '.' -f1)"
    forge inspect $fqn storageLayout
done
```

We need to set a mapping in the `L2CrossDomainMessenger` and one of the contracts has a variable with an offset

state-surgery/state/state_test.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+}
+
+func TestComputeStorageSlots(t *testing.T) {
+	values := state.StorageValues{}
+	values["time"] = 12
+	values["addr"] = "0xEA674fdDe714fd979de3EdF0F56AA9716B898ec8"
+	values["boolean"] = true
+
+	slots, err := state.ComputeStorageSlots(&layout, values)
```
mslipper, line 32: What is this test asserting? It looks like it only asserts that the encoding was successful, and doesn't check the values of the storage slots themselves.

state-surgery/state/memory_db.go
```diff
@@ -0,0 +1,305 @@
...
+		Nonce:   0,
+	}
+}
+
+func (db *MemoryStateDB) SubBalance(addr common.Address, amount *big.Int) {
+	db.rw.Lock()
+	defer db.rw.Unlock()
+
+	account, ok := db.genesis.Alloc[addr]
+	if !ok {
```
mslipper, line 70: Is it always the case that `AddBalance` and `SubBalance` are always called after `CreateAccount`? In the current implementation, calling these methods before `CreateAccount` will lead to a silent failure to set the appropriate balance.

mslipper (changes requested):

state-surgery/state/memory_db.go
tynes: This matches the current behavior of the statedb in geth, we could panic in the case of it being called without the account existing to be more strict

state-surgery/state/state_test.go
tynes: I can add assertions for this, was planning on thorough testing in a follow up PR because this is critical code. I have a few ideas on ways to test this thoroughly, it still needs a bit of functionality to be ready for our usage

state-surgery/state/memory_db.go
mslipper: I think we should panic. It's a deviation from Geth but silent failures always scare me.

mslipper: I think we should panic. It's a deviation from Geth but silent failures always scare me.

state-surgery/state/state_test.go
mslipper: I'm fine adding testing in a separate PR if the tests are substantial and will need dedicated review.

tynes:
@mslipper Will follow up with an additional PR that adds the thorough testing + final required functionality

mslipper (approved):

trianglesphere:
The go code that I've read seems fine. I just don't have a great idea of the solidity ABI / slot setting / packing. If there's a good reference for that I could do a cross check & I'd be happy with that portion.

The in memory DB looks good to me.

state-surgery/state/state_test.go
```diff
@@ -174,3 +174,232 @@ func testContractStateValuesAreEmpty(t *testing.T, contract *testdata.Testdata)
...
+			offset:  1,
+			expect:  common.Hash{30: 0x01},
+		},
+		{
+			boolean: false,
+			offset:  1,
+			expect:  common.Hash{},
+		},
+		{
+			boolean: "true",
```
trianglesphere, line 354: duplicate to above

tynes: The difference is the type of the argument, it handles strings and bools

tynes:
> The go code that I've read seems fine. I just don't have a great idea of the solidity ABI / slot setting / packing. If there's a good reference for that I could do a cross check & I'd be happy with that portion.
> 
> The in memory DB looks good to me.

Not super in depth but there is this: https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html
I've seen a doc going around that shows the values in depth, not sure where it is though. Should keep that link handy to be honest

fwiw there is test coverage that setting the storage slots of a deployed contract works

trianglesphere (approved):

PR #3217, opened 2022-08-13
op-node: remove parallel batch fetching / deadlock, introduce iterative batch fetching

protolambda:
Previously receipts were being fetched in parallel RPC batches, but there was a deadlock risk due to a race-condition between accepting batch results and scheduling new requests, causing the node to lock up.

This PR gets rid of the old parallel batch fetching, in favor of a new design that:
- Lets the caller parallelize by simply calling the Fetch method in multiple go routines, and avoiding that go routine complexity in the fetcher itself.
- Lets the caller stop/pause/backoff on any fetching error, using multi-error to expose the errors of the individual RPC calls.
- Propagates RPC errors to the caller nicely, instead of the previous logging and poor retry attempts
- Eventually completes and exposes the Results

The `l1.ReceiptsFetcher` wraps the `IterativeBatchCall` to fetch receipts this way, in the future we can reuse the batching logic for other RPC interactions.

The `eth.ReceiptsFetcher` is an interface which can be used by the derivation pipeline, which does not directly import any rpc / concurrency related code. And `eth.FetchedReceipts` makes it super easy to provide a list of already available receipts, for testing and a happy path for 0 receipts to fetch.

The current implementation still calls the `Fetch` method repeatedly until complete, but in a follow-up PR I think we can refactor `PreparePayloadAttributes` to also progress more iteratively, since it does many RPC calls it's worth it to recover from any one of them, instead of starting over on any error.

This PR also fixes the L1 receipts cache: we were not adding the receipts to the cache. Now we cache the `eth.ReceiptsFetcher` job of fetching the receipts, which is safe to concurrently fetch for, and stores the results.

And in #3188 I made the mistake of scaling the receipts/transactions cache sizes by a multiple of the blocks cache, even though receipts/transactions are cached as lists per block (@tuxcanfly sorry I misunderstood your comment in that PR, I think you were right about the cache size issue).

Fix ENG-2505

### DIFF
--- a/op-node/l1/batching.go
+++ b/op-node/l1/batching.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"sync"
 	"sync/atomic"
 
 	"github.com/ethereum/go-ethereum/rpc"
@@ -11,25 +12,70 @@ import (
 )
 
 // IterativeBatchCall is an util to create a job to fetch many RPC requests in batches,
-// and enable the caller to parallelize easily, handle and re-try error,
+// and enable the caller to parallelize easily and safely, handle and re-try errors,
 // and pick a batch size all by simply calling Fetch again and again until it returns io.EOF.
-type IterativeBatchCall struct {
-	completed uint32
-	getBatch  batchCallContextFn
-	requests  []rpc.BatchElem
-	scheduled chan rpc.BatchElem
+type IterativeBatchCall[K any, V any, O any] struct {
+	completed uint32       // tracks how far to completing all requests we are
+	resetLock sync.RWMutex // ensures we do not concurrently read (incl. fetch) / reset
+
+	requestsKeys []K
+	batchSize    int
+
+	makeRequest func(K) (V, rpc.BatchElem)
+	makeResults func([]K, []V) (O, error)
+	getBatch    batchCallContextFn
+
+	requestsValues []V
+	scheduled      chan rpc.BatchElem
+
+	results *O
 }
 
-func NewIterativeBatchCall(requests []rpc.BatchElem, getBatch batchCallContextFn) *IterativeBatchCall {
-	scheduled := make(chan rpc.BatchElem, len(requests))
-	for _, r := range requests {
+// NewIterativeBatchCall constructs a batch call, fetching the values with the given keys,
+// and transforms them into a verified final result.
+func NewIterativeBatchCall[K any, V any, O any](
+	requestsKeys []K,
+	makeRequest func(K) (V, rpc.BatchElem),
+	makeResults func([]K, []V) (O, error),
+	getBatch batchCallContextFn,
+	batchSize int) *IterativeBatchCall[K, V, O] {
+
+	if len(requestsKeys) < batchSize {
+		batchSize = len(requestsKeys)
+	}
+	if batchSize < 1 {
+		batchSize = 1
+	}
+
+	out := &IterativeBatchCall[K, V, O]{
+		completed:    0,
+		getBatch:     getBatch,
+		requestsKeys: requestsKeys,
+		batchSize:    batchSize,
+		makeRequest:  makeRequest,
+		makeResults:  makeResults,
+	}
+	out.Reset()
+	return out
+}
+
+// Reset will clear the batch call, to start fetching all contents from scratch.
+func (ibc *IterativeBatchCall[K, V, O]) Reset() {
+	ibc.resetLock.Lock()
+	defer ibc.resetLock.Unlock()
+
+	scheduled := make(chan rpc.BatchElem, len(ibc.requestsKeys))
+	requestsValues := make([]V, len(ibc.requestsKeys))
+	for i, k := range ibc.requestsKeys {
+		v, r := ibc.makeRequest(k)
+		requestsValues[i] = v
 		scheduled <- r
 	}
-	return &IterativeBatchCall{
-		completed: 0,
-		getBatch:  getBatch,
-		requests:  requests,
-		scheduled: scheduled,
+
+	ibc.requestsValues = requestsValues
+	ibc.scheduled = scheduled
+	if len(ibc.requestsKeys) == 0 {
+		close(ibc.scheduled)
 	}
 }
 
@@ -37,9 +83,12 @@ func NewIterativeBatchCall(requests []rpc.BatchElem, getBatch batchCallContextFn
 // This method is safe to call concurrently: it will parallelize the fetching work.
 // If no work is available, but the fetching is not done yet,
 // then Fetch will block until the next thing can be fetched, or until the context expires.
-func (ibc *IterativeBatchCall) Fetch(ctx context.Context, maxBatchSize uint) error {
+func (ibc *IterativeBatchCall[K, V, O]) Fetch(ctx context.Context) error {
+	ibc.resetLock.RLock()
+	defer ibc.resetLock.RUnlock()
+
 	// collect a batch from the requests channel
-	batch := make([]rpc.BatchElem, 0, maxBatchSize)
+	batch := make([]rpc.BatchElem, 0, ibc.batchSize)
 	// wait for first element
 	select {
 	case reqElem, ok := <-ibc.scheduled:
@@ -53,7 +102,7 @@ func (ibc *IterativeBatchCall) Fetch(ctx context.Context, maxBatchSize uint) err
 
 	// collect more elements, if there are any.
 	for {
-		if uint(len(batch)) >= maxBatchSize {
+		if len(batch) >= ibc.batchSize {
 			break
 		}
 		select {
@@ -89,7 +138,7 @@ func (ibc *IterativeBatchCall) Fetch(ctx context.Context, maxBatchSize uint) err
 			continue
 		} else {
 			atomic.AddUint32(&ibc.completed, 1)
-			if atomic.LoadUint32(&ibc.completed) >= uint32(len(ibc.requests)) {
+			if atomic.LoadUint32(&ibc.completed) >= uint32(len(ibc.requestsKeys)) {
 				close(ibc.scheduled)
 				return io.EOF
 			}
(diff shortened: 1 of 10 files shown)


op-node/l1/batching.go
```diff
@@ -2,134 +2,103 @@ package l1
...
 		case <-ctx.Done():
 			return ctx.Err()
+		default:
+			break
 		}
+		break
 	}
+
+	if err := ibc.getBatch(ctx, batch); err != nil {
+		for _, r := range batch {
```
tynes, line 75: I see that we reschedule the entire batch here, would it be too complex to use multierror to only reschedule the requests that need to be rescheduled? Maybe I'm making some assumptions that aren't correct in this context

op-node/l1/batching.go
```diff
@@ -2,134 +2,103 @@ package l1
...
+				close(ibc.scheduled)
+				return io.EOF
+			}
+		}
+	}
+	return result
+}
+
+func (ibc *IterativeBatchCall) Complete() bool {
+	return ibc.completed >= uint32(len(ibc.requests))
```
tynes, line 99: Would it be a bug if `completed > ibc.requests`?

tynes (commented):
I like this utility and think it could be used throughout our codebases when we want to make a bunch of JSON RPC requests, basically batching them together and just getting them back behind a nice and reliable API is a good idea.

op-node/l1/receipts.go
```diff
@@ -78,9 +83,45 @@ func fetchReceipts(ctx context.Context, block eth.BlockID, receiptHash common.Ha
...
+	}
+	if batchSize < 1 {
+		batchSize = 1
+	}
+
+	receipts := make([]*types.Receipt, len(txs))
+	receiptRequests := make([]rpc.BatchElem, len(txs))
+	for i := 0; i < len(txs); i++ {
+		receipts[i] = new(types.Receipt)
+		receiptRequests[i] = rpc.BatchElem{
```
tynes, line 112: I think it could be nice to have a helper function here that creates the `rpc.BatchElem`, something like `NewBatchGetTransactionReceipt(common.Hash)`. We could then have a library of these constructor functions and very cleanly build whatever sort of batch calls that we want

op-node/l1/batching.go
protolambda: Technically yes, but it doesn't hurt to do `>=` instead of `==`

op-node/l1/receipts.go
protolambda: Agreed, the RPC iterative batch fetching is a nice util, also for L2 RPC. Would like to minimize changes in this PR though, we can move it to a common package + polish for more general usage in a later PR.

op-node/l1/batching.go
protolambda: This is a global error, it only happens if the get_batch RPC as a whole fails, so all items are always all failing, and have to be put back in the scheduling channel. See below for the per-item error handling, where only the failing items are put back into the scheduling channel.

op-node/l1/batching.go
```diff
@@ -2,134 +2,103 @@ package l1
...
+	var result error
+	for _, elem := range batch {
+		if elem.Error != nil {
+			result = multierror.Append(result, elem.Error)
+			elem.Error = nil // reset, we'll try this element again
+			ibc.scheduled <- elem
+			continue
+		} else {
+			atomic.AddUint32(&ibc.completed, 1)
+			if ibc.completed >= uint32(len(ibc.requests)) {
```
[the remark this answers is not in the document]
protolambda: Not sure 100% sure about the Go memory handling. Assuming it's aligned to the top of the struct, and only 32 bits, it might never partially mutate the uint32 while reading. But I'll add the atomic Loading here and in `Complete` just to be sure, nice catch.

op-node/l1/batching.go
tynes: Perhaps a warning log could help give insights into this happening?

trianglesphere:
I need to fully review this, but my sense is that I prefer having a concurrent (& rate/concurrency limited) fetcher without exposing an iterator based / caller managed concurrency API.

protolambda:
Rebasing this onto `develop` for devnet fix (other PR building on top of this is stuck on devnet CI).

op-node/l1/receipts.go
```diff
@@ -78,9 +83,45 @@ func fetchReceipts(ctx context.Context, block eth.BlockID, receiptHash common.Ha
...
-		return nil, fmt.Errorf("failed to fetch list of receipts: expected receipt root %s but computed %s from retrieved receipts", receiptHash, computed)
+	computed := types.DeriveSha(types.Receipts(rf.receipts), hasher)
+	if rf.receiptHash != computed {
+		return nil, fmt.Errorf("failed to fetch list of receipts: expected receipt root %s but computed %s from retrieved receipts", rf.receiptHash, computed)
 	}
-	return receipts, nil
+	return rf.receipts, nil
+}
+
+// NewReceiptsFetcher creates a receipt fetcher that can iteratively fetch the receipts matching the given t
```
trianglesphere, line 93: `given t` - it seems like the rest of the comment got chopped off

op-node/l1/receipts.go
```diff
@@ -13,44 +14,48 @@ import (
...
+// An io.EOF error will be returned once the fetching is done.
+func (rf *ReceiptsFetcher) Fetch(ctx context.Context) error {
+	return rf.iterBatchCall.Fetch(ctx, rf.batchSize)
+}
 
+func (rf *ReceiptsFetcher) Complete() bool {
+	return rf.iterBatchCall.Complete()
+}
+
+func (rf *ReceiptsFetcher) Result() (types.Receipts, error) {
```
trianglesphere, line 40: The verification of receipts should be a pure function

op-node/rollup/derive/attributes.go
```diff
@@ -40,6 +41,17 @@ func PreparePayloadAttributes(ctx context.Context, cfg *rollup.Config, dl L1Rece
 				fmt.Errorf("cannot create new block with L1 origin %s (parent %s) on top of L1 origin %s",
 					epoch, info.ParentHash(), l2Parent.L1Origin))
 		}
+		for {
+			if err := receiptsFetcher.Fetch(ctx); err == io.EOF {
```
trianglesphere, line 45: So we don't actually use this concurrently

trianglesphere (changes requested):
Ok, I gave this a closer look. It's not obviously incorrect, but a lot of l1 source code is quite complex.

My preferred solution is to simplify the concurrency inside the fetcher rather than move to an iterative based model. We often check the receipt hash where it doesn't make sense to expose an iterative API as we have to wait for all data to be fetched before returning.

I'm also worried about the overhead of managing error handling and backoffs inside the main loop. If we really want to yield control back we can schedule a job & poll for completion. I think exposing the iterative based API removes a nice API boundary that we had and instead tangles code together in a way that makes upgrading & reasoning about the code much harder.

Looking through the surrounding code, I also forgot that we send concurrent batch requests rather than individual requests. Did we benchmark that sending multiple batch requests was the way to go? Right now it's a bit better that we iteratively request batches rather than single items.

As another aside, receipt / transaction fetching is actually something that could easily be always going on in the background - that pipelines the fetching with accessing it when we actually use it. That pulls fetching data out of the critical path.

protolambda:
Applied a minimal fix for the cache issue, but not quite happy with the code either. I do like the new channel pattern, exposing errors better, and not forcing a certain retry/parallel schedule when the derivation pipeline only really needs to backoff on errors properly. But the locking isn't nice, and it's still a bit tricky code. Will make a new version next week after working on some other more important issues.

protolambda:
`makeReceiptsFn` is now a pure function, and the receipts-fetcher implementation is removed in favor of generic usage of the underlying batch fetcher. This will work for now, and we can wrap it with a fetcher that runs the Fetch calls in parallel by default later.

protolambda:
Rebased onto develop for `devnet` CI.

op-node/l1/batching.go
```diff
@@ -2,134 +2,181 @@ package l1
...
+func (ibc *IterativeBatchCall[K, V, O]) Complete() bool {
+	ibc.resetLock.RLock()
+	defer ibc.resetLock.RUnlock()
+	return atomic.LoadUint32(&ibc.completed) >= uint32(len(ibc.requestsKeys))
+}
+
+// Result returns the fetched values, checked and transformed to the final output type, if available.
+// If the check fails, the IterativeBatchCall will Reset itself, to be ready for a re-attempt in fetching new data.
+func (ibc *IterativeBatchCall[K, V, O]) Result() (O, error) {
+	ibc.resetLock.RLock()
```
tynes, line 160: Any reason to not defer runlock here?

protolambda: Because the later `Lock` during the later `Reset` and results-caching would conflict. Alternatively we can just wrap the function as a whole with `Lock` / `defer Unlock`, less optimal, but cleaner

op-node/l1/batching.go
```diff
@@ -2,134 +2,181 @@ package l1
...
-	requests = requests[:len(requests):len(requests)]
+	results *O
+}
 
-	expectedBatches := (len(requests) + maxPerBatch - 1) / maxPerBatch
+// NewIterativeBatchCall constructs a batch call, fetching the values with the given keys,
+// and transforms them into a verified final result.
+func NewIterativeBatchCall[K any, V any, O any](
+	requestsKeys []K,
+	makeRequest func(K) (V, rpc.BatchElem),
```
tynes, line 38: `makeRequest` is fine for now but would have to be refactored if we wanted to use it with something that took more than 1 argument in the rpc params

protolambda: `K` can be any type, e.g. a bundle of request arguments, and the `makeRequest` function can then output a batch request with multiple RPC arguments by unpacking it in some request specific way.

op-node/l1/receipts.go
```diff
@@ -13,74 +12,82 @@ import (
...
-			if log.Index != logIndex {
-				return nil, fmt.Errorf("log %d (%d of tx %d) has unexpected log index %d", logIndex, j, i, log.Index)
+		if len(txHashes) == 0 {
+			if receiptHash != types.EmptyRootHash {
+				return nil, fmt.Errorf("no transactions, but got non-empty receipt trie root: %s", receiptHash)
 			}
-			if log.TxIndex != uint(i) {
-				return nil, fmt.Errorf("log %d has unexpected tx index %d", log.Index, log.TxIndex)
+		}
+		// We don't trust the RPC to provide consistent cached receipt info that we use for critical rollup derivation work.
```
tynes, line 25: This logic looks correct after looking at it closely, I dont think it would hurt to modularize it into a function so that it could be unit tested against mainnet data

op-node/l1/batching.go
tynes: Ah i wasn't sure if it would need to be `[]K`

tynes (approved):
Looks good to me, generally easier to read than the last time that I looked at it

PR #3218, opened 2022-08-13
feat: `op-hardhat-chainops` package

tynes:
**Description**
Create a package that can be used for hardhat task chainops. See the readme for an example on how
it can be used.
Also migrate the sdk hh tasks to using the new chainops package and add some additional logging to
make it clear what is going on during those tasks.


tynes:
Its failing with:

```
#30 3.004 error Couldn't find package "@eth-optimism/op-hardhat-chainops@^0.0.1" required by "@eth-optimism/sdk@1.5.0" on the "npm" registry.
#30 3.004 info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
#30 ERROR: executor failed running [/bin/sh -c yarn install --frozen-lockfile && yarn cache clean]: exit code: 1
------
 > [base 24/27] RUN yarn install --frozen-lockfile && yarn cache clean:
------
executor failed running [/bin/sh -c yarn install --frozen-lockfile && yarn cache clean]: exit code: 1
ERROR: Service 'deployer' failed to build : Build failed
```

tynes:
This is going stale, I'd like to get some sort of functionality from this into the sdk where we can go from an hre -> an object of all connected contracts. This would reduce the boilerplate so much in random scripts that I write

maurelian (approved):
Usage in the SDK looks so much nicer with this. 👏

mslipper:
Is this still something we want to get reviewed?

tynes:
After talking to @smartcontracts , an alternative approach to having this functionality was suggested. I like the idea of being able to go from `hre` to an object of connected contracts if we are planning on using hardhat tasks in the future. This will eliminate a lot of boilerplate in any script that needs to attach contracts to a deployment.

I believe the proposed change was to add a function to the SDK that would be something like `sdk.fromHre(hre: HardhatRuntimeEnvironment) -> OptimismContracts` where `OptimismContracts` has a `contracts` object where each key is the name of the contract and the value is a connected `ethers.Contract`

PR #3221, opened 2022-08-14
op-node/specs: refactor batch queue, add parent_hash to batches, extend testing

protolambda:
This adds `parent_hash` to the batch definition, to ensure we never skip a batch. Skipping a batch can cause transactions with invalid nonce values to be inserted into the execution engine, causing reorgs or complete halts.

The `parent_hash` matches the L2 block hash of the block the batch will be building on top of, i.e. the L2 safe head.
To enable this, we have to change the batch-queue processing to only output 1 batch at the time. Without this, the previous L2 block hash would not be available, and so we could not check the hash otherwise.

In a previous iteration of this I considered hashing just the batch inputs, and chaining those hashes. However, we can't include those hashes into the L2 chain elegantly, which is required to find a sync continuation point without loading all historical batches. I.e. we would need to add additional data to the L1 block info transaction, and it would almost be self-referencing, as it hashes together the other transactions, which are included later.
Using the native block hashes thus is preferred to not add additional data to the info deposit, and avoid the near cyclic dependency issue.

Not outputting multiple batches at a time really means that we are not filling multiple batches at a time when there is a gap to fill. The eager batch derivation already outputted a single batch otherwise.

Previously the validity conditions on batches were spread between the batch-filtering, batch-filling, extension-checking and batch queue main function. I changed this so we can output 1 batch at a time, check all validity conditions, and isolate the validity conditions in one function that now has 100% test coverage.

Part of this unwinding of the different checks also meant we can now remove the L1 RPC from the function: all necessary L1 data is there, and is easy to access (first L1 block always matches L2 safe head L1 origin, second L1 block is optional).

And validating the batches this way means there is no difference between eager and regular batch-derivation: we always consider the validity of the first batch, since we output one at a time.

The spec is updated to reflect this same methodology, so we don't need eager batch queue <> regular batch queue distinction anymore.

I hope the PR is not too large to review, this was not an easy change, at least it should reduce total lines of regular code, while adding more tests + ensuring 100% coverage of batch validation.

Fix ENG-2593

### DIFF
--- a/op-node/rollup/derive/attributes_queue.go
+++ b/op-node/rollup/derive/attributes_queue.go
@@ -2,6 +2,7 @@ package derive
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"time"
 
@@ -63,9 +64,14 @@ func (aq *AttributesQueue) Step(ctx context.Context, outer Progress) error {
 	}
 	batch := aq.batches[0]
 
+	safeL2Head := aq.next.SafeL2Head()
+	// sanity check parent hash
+	if batch.ParentHash != safeL2Head.Hash {
+		return NewCriticalError(fmt.Errorf("valid batch has bad parent hash %s, expected %s", batch.ParentHash, safeL2Head.Hash))
+	}
 	fetchCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
 	defer cancel()
-	attrs, err := PreparePayloadAttributes(fetchCtx, aq.config, aq.dl, aq.next.SafeL2Head(), batch.Timestamp, batch.Epoch())
+	attrs, err := PreparePayloadAttributes(fetchCtx, aq.config, aq.dl, safeL2Head, batch.Timestamp, batch.Epoch())
 	if err != nil {
 		return err
 	}
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -2,6 +2,7 @@ package derive
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 
@@ -97,22 +98,13 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
 	if len(bq.l1Blocks) == 0 {
 		panic(fmt.Errorf("cannot add batch with timestamp %d, no origin was prepared", batch.Timestamp))
 	}
-	log := bq.log.New(
-		"current_l1", bq.progress.Origin,
-		"batch_timestamp", batch.Timestamp,
-		"batch_epoch", batch.Epoch(),
-		"txs", len(batch.Transactions),
-	)
-	log.Trace("queuing batch")
-
 	data := BatchWithL1InclusionBlock{
 		L1InclusionBlock: bq.progress.Origin,
 		Batch:            batch,
 	}
 	validity := CheckBatch(bq.config, bq.log, bq.l1Blocks, bq.next.SafeL2Head(), &data)
 	if validity == BatchDrop {
-		log.Warn("ingested invalid batch from L1, dropping it instead of buffering it")
-		return
+		return // if we do drop the batch, CheckBatch will log the drop reason with WARN level.
 	}
 	bq.batches[batch.Timestamp] = append(bq.batches[batch.Timestamp], &data)
 }
@@ -123,7 +115,7 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
 // If no batch can be derived yet, then (nil, io.EOF) is returned.
 func (bq *BatchQueue) deriveNextBatch(ctx context.Context) (*BatchData, error) {
 	if len(bq.l1Blocks) == 0 {
-		panic("cannot derive next batch, no origin was prepared")
+		return nil, NewCriticalError(errors.New("cannot derive next batch, no origin was prepared"))
 	}
 	epoch := bq.l1Blocks[0]
 	l2SafeHead := bq.next.SafeL2Head()
@@ -169,7 +161,7 @@ batchLoop:
 			bq.batches[nextTimestamp] = remaining
 			return nil, io.EOF
 		default:
-			panic(fmt.Errorf("unknown batch validity type: %d", validity))
+			return nil, NewCriticalError(fmt.Errorf("unknown batch validity type: %d", validity))
 		}
 	}
 	// clean up if we remove the final batch for this timestamp
@@ -183,10 +175,6 @@ batchLoop:
 		// advance epoch if necessary
 		if nextBatch.Batch.EpochNum == rollup.Epoch(epoch.Number)+1 {
 			bq.l1Blocks = bq.l1Blocks[1:]
-		}
-		// sanity check
-		if nextBatch.Batch.EpochNum > rollup.Epoch(epoch.Number)+1 {
-			return nil, NewCriticalError(fmt.Errorf("batch is advancing more than 1 epoch, from %s to %s", epoch, nextBatch.Batch.Epoch()))
 		}
 		return nextBatch.Batch, nil
 	}
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -41,8 +41,8 @@ type BatchQueue struct {
 
 	l1Blocks []eth.L1BlockRef
 
-	// batches in order of when we've first seen them
-	batches []*BatchWithL1InclusionBlock
+	// batches in order of when we've first seen them, grouped by L2 timestamp
+	batches map[uint64][]*BatchWithL1InclusionBlock
 }
 
 // NewBatchQueue creates a BatchQueue, which should be Reset(origin) before use.
@@ -83,7 +83,7 @@ func (bq *BatchQueue) ResetStep(ctx context.Context, l1Fetcher L1Fetcher) error 
 	// Copy over the Origin from the next stage
 	// It is set in the engine queue (two stages away) such that the L2 Safe Head origin is the progress
 	bq.progress = bq.next.Progress()
-	bq.batches = bq.batches[:0]
+	bq.batches = make(map[uint64][]*BatchWithL1InclusionBlock)
 	// Include the new origin as an origin to build on
 	bq.l1Blocks = bq.l1Blocks[:0]
 	bq.l1Blocks = append(bq.l1Blocks, bq.progress.Origin)
@@ -114,10 +114,12 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
 		log.Warn("ingested invalid batch from L1, dropping it instead of buffering it")
 		return
 	}
-	bq.batches = append(bq.batches, &data)
+	bq.batches[batch.Timestamp] = append(bq.batches[batch.Timestamp], &data)
 }
 
-// deriveNextBatch derives the next batch.
+// deriveNextBatch derives the next batch to apply on top of the current L2 safe head,
+// following the validity rules imposed on consecutive batches,
+// based on currently available buffered batch and L1 origin information.
 // If no batch can be derived yet, then (nil, io.EOF) is returned.
 func (bq *BatchQueue) deriveNextBatch(ctx context.Context) (*BatchData, error) {
 	if len(bq.l1Blocks) == 0 {
@@ -138,14 +140,14 @@ func (bq *BatchQueue) deriveNextBatch(ctx context.Context) (*BatchData, error) {
 
 	// Go over all batches, in order of inclusion, and find the first batch we can accept.
 	// We filter in-place by only remembering the batches that may be processed in the future, or those we are undecided on.
-	remaining := bq.batches[:0]
+	var remaining []*BatchWithL1InclusionBlock
+	candidates := bq.batches[nextTimestamp]
 batchLoop:
-	for i, batch := range bq.batches {
+	for i, batch := range candidates {
 		validity := CheckBatch(bq.config, bq.log.New("batch_index", i), bq.l1Blocks, l2SafeHead, batch)
 		switch validity {
 		case BatchFuture:
-			remaining = append(remaining, batch)
-			continue
+			return nil, NewCriticalError(fmt.Errorf("found batch with timestamp %d marked as future batch, but expected timestamp %d", batch.Batch.Timestamp, nextTimestamp))
 		case BatchDrop:
 			bq.log.Warn("dropping batch",
 				"batch_timestamp", batch.Batch.Timestamp,
@@ -158,63 +160,26 @@ batchLoop:
 			continue
 		case BatchAccept:
 			nextBatch = batch
-			// remove the current batch since we are processing it now, and retain every batch we didn't get to yet.
-			remaining = append(remaining, bq.batches[i+1:]...)
+			// don't keep the current batch in the remaining items since we are processing it now,
+			// but retain every batch we didn't get to yet.
+			remaining = append(remaining, candidates[i+1:]...)
 			break batchLoop
 		case BatchUndecided:
 			remaining = append(remaining, batch)
-			bq.batches = remaining
+			bq.batches[nextTimestamp] = remaining
 			return nil, io.EOF
 		default:
 			panic(fmt.Errorf("unknown batch validity type: %d", validity))
 		}
 	}
-	bq.batches = remaining
-
-	if nextBatch == nil {
-		// If the current epoch is too old compared to the L1 block we are at,
-		// i.e. if the sequence window expired, we create empty batches
-		expiryEpoch := epoch.Number + bq.config.SeqWindowSize
-		forceNextEpoch :=
-			(expiryEpoch == bq.progress.Origin.Number && bq.progress.Closed) ||
-				expiryEpoch < bq.progress.Origin.Number
-
-		if !forceNextEpoch {
-			// sequence window did not expire yet, still room to receive batches for the current epoch,
-			// no need to force-create empty batch(es) towards the next epoch yet.
-			return nil, io.EOF
-		}
-		if len(bq.l1Blocks) < 2 {
-			// need next L1 block to proceed towards
-			return nil, io.EOF
-		}
-
-		nextEpoch := bq.l1Blocks[1]
-		// Fill with empty L2 blocks of the same epoch until we meet the time of the next L1 origin,
-		// to preserve that L2 time >= L1 time
-		if nextTimestamp < nextEpoch.Time {
-			return &BatchData{
-				BatchV1{
-					ParentHash:   l2SafeHead.Hash,
-					EpochNum:     rollup.Epoch(epoch.Number),
-					EpochHash:    epoch.Hash,
-					Timestamp:    nextTimestamp,
-					Transactions: nil,
-				},
-			}, nil
-		}
-		// As we move the safe head origin forward, we also drop the old L1 block reference
-		bq.l1Blocks = bq.l1Blocks[1:]
-		return &BatchData{
-			BatchV1{
-				ParentHash:   l2SafeHead.Hash,
-				EpochNum:     rollup.Epoch(nextEpoch.Number),
-				EpochHash:    nextEpoch.Hash,
-				Timestamp:    nextTimestamp,
-				Transactions: nil,
-			},
-		}, nil
+	// clean up if we remove the final batch for this timestamp
+	if len(remaining) == 0 {
+		delete(bq.batches, nextTimestamp)
 	} else {
+		bq.batches[nextTimestamp] = remaining
+	}
+
+	if nextBatch != nil {
 		// advance epoch if necessary
 		if nextBatch.Batch.EpochNum == rollup.Epoch(epoch.Number)+1 {
 			bq.l1Blocks = bq.l1Blocks[1:]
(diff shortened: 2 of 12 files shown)


protolambda:
Rebased onto `develop` for the devnet CI fix.

norswap:
> The parent_hash matches the L2 block hash of the block the batch will be building on top of, i.e. the L2 safe head.
To enable this, we have to change the batch-queue processing to only output 1 batch at the time. Without this, the previous L2 block hash would not be available, and so we could not check the hash otherwise.

This has a pretty big consequence which is that if the sequencer does something bad on the black transition of batch `X`, then all the subsequent batches are rendered useless and cannot be replayed anymore. That's a pretty change and I feel like it should warrant some discussions of pros & cons.

> In a previous iteration of this I considered hashing just the batch inputs, and chaining those hashes. However, we can't include those hashes into the L2 chain elegantly, which is required to find a sync continuation point without loading all historical batches. I.e. we would need to add additional data to the L1 block info transaction, and it would almost be self-referencing, as it hashes together the other transactions, which are included later.

I'm not entirely sure why we can't include the hash of the previous batch in the L1 infos of the next block?

specs/derivation.md
```diff
@@ -490,27 +491,63 @@ Note that the presence of any gaps in the batches derived from L1 means that thi
...
+
+- `parent_hash = safe_l2_head.hash`
+- `timestamp = next_timestamp`
+- `transactions` is empty, i.e. no sequencer transactions. Deposited transactions may be added in the next stage.
+- If `next_timestamp < next_epoch.time`: the current L1 origin is repeated, to preserve the L2 time invariant.
+  - `epoch_num = epoch.number`
+  - `epoch_hash = epoch.hash`
+- Otherwise,
+  - `epoch_num = next_epoch.number`
+  - `epoch_hash = next_epoch.hash`
```
norswap, line 551: I'm wondering if these spec changes are really necessary.

...

I no joke spent > 30 minutes writing about pros & cons of current & proposed spec, but truth be told, I'm frustrated with both.

I think what I'd love is for this to be rewritten more precisely in terms of:
- input: a stream of batches
- inner state: an ordered buffer + l2 safe head
- output: a stream of ordered batches (one for every slot)

and the transition function being what appens when  either (a) you process a batch from the input stream, or (b) the "L1 head" (i.e. latest fully processed L1 block) changes.

(btw I think ideally the whole spec should be structured like this and the implem should try to follow that format too)

I think for this format, the old spec will form a stronger base, because basically when a batch comes in you want to apply the timestamp + duplication filtering criteria from the old spec. Then you want to see if after adding the batch you have some contiguous batches at the start of your buffer, and if so push that to the next stage.

At that point you *haven't* applied the `parent_hash` or epoch number checks (though you probably could do the epoch number check by verifying that the epoch number does not go down as the timestamp go up?? maybe). These would be perfect to push to the next stage (e.g. only derive payload attributes when we have an up to date L2 safe head to build upon).

norswap (commented):
Didn't review the code really, mostly spec + general principles.

I think we need to talk about the implications of the block hash (see my comment separate from this review).

Specs-wise, I have some suggestions about how to better structure this bit of the spec. Maybe with impact to the code (new stage?)

Regarding the code, this is a lot of changes. I hear that it's quite hard to split, but it would be incredibly easy for something to slip through the cracks.

Wonder if we could have worked to temporary stages that are fully functional but a bit "frankenstein" in that they introduce code that will be removed in the very next PR? Not sure, but worth considering.

I think in this specific case, it would have been good to implement the change entirely without a refactor, then consider refactors separately.

And we could have split refactors in multiple PRs where each PR moves only a single check for one place to another.

Imho right now our options are basically: hope our testing catches any issues, or do the above. Don't feel very comfortable reviewing the changes and seeing "yup, I'm confident that before/after is the same behaviour modulo the new bit we're adding".

op-node/rollup/derive/batch_queue_test.go
```diff
@@ -244,76 +249,56 @@ func TestBatchQueueMissing(t *testing.T) {
...
+	// We start with an open L1 origin as progress in the first step
+	progress := bq.progress
 	require.Equal(t, bq.progress.Closed, false)
 
-	// Add batches
-	// NB: The batch at 18 is skipped to skip over the ability to
-	// do eager batch processing for that batch. This test checks
-	// that batch timestamp 12 & 14 is created & 16 is used.
-	batches := []*BatchData{b(16, l1[0]), b(20, l1[1])}
+	// The batches at 18 and 20 are skipped to stop 22 from being eagerly processed.
```
trianglesphere, line 259: really nice comments here.

op-node/rollup/derive/batches.go
```diff
@@ -1,106 +1,121 @@
...
-		return fmt.Errorf("batch was meant for alternative L1 chain")
+	if l1Blocks[0].Hash != l2SafeHead.L1Origin.Hash {
+		log.Warn("safe L2 head L1 origin does not match batch first l1 block",
+			"safe_l2", l2SafeHead, "safe_origin", l2SafeHead.L1Origin, "l1_block", l1Blocks[0])
+		return BatchUndecided
 	}
-	if (batch.Timestamp-config.Genesis.L2Time)%config.BlockTime != 0 {
-		return fmt.Errorf("bad timestamp %d, not a multiple of the block time", batch.Timestamp)
+
+	if batch.Batch.Timestamp < nextTimestamp {
```
trianglesphere, line 53: this plus `batch.Batch.Timestamp > nextTimestamp` is a weird way of saying
```
if batch.Batch.Timestamp != nextTimestamp {
    return ...
}
```

trianglesphere (changes requested):
I went through and reviewed everything. I didn't realize how much you had refactored in there. Splitting this up into a refactor + adding parent hash as two steps (and maybe even smaller refactor steps depending on how it is) would have made this easier to review.

There's a couple spots that need some more comments.

My main concern is the repeated linear scan through `bq.batches` + how that array is modified. It does reduce the chance of leaking compared to the old design, but now requires (potentially) much more work.

specs/derivation.md
```diff
@@ -490,27 +491,63 @@ Note that the presence of any gaps in the batches derived from L1 means that thi
...
+If no batch can be `accept`-ed, and the stage has completed buffering of all batches that can fully be read from the L1
+block at height `epoch.number + sequence_window_size`, and the `next_epoch` is available,
+then an empty batch can be derived with the following properties:
+
+- `parent_hash = safe_l2_head.hash`
+- `timestamp = next_timestamp`
+- `transactions` is empty, i.e. no sequencer transactions. Deposited transactions may be added in the next stage.
+- If `next_timestamp < next_epoch.time`: the current L1 origin is repeated, to preserve the L2 time invariant.
+  - `epoch_num = epoch.number`
+  - `epoch_hash = epoch.hash`
```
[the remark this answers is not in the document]
protolambda: That's exactly what we did previously: we can't ever break the time invariant or introduce a gap in L2 blocks, and must thus produce empty blocks with the old epoch, until we can meet the L1 time. And moving ahead to the next epoch is the right thing to do if we don't break the invariant, when we're forced to produce an empty block to move forward the deposit processing as the sequence window is expired.

specs/derivation.md
```diff
@@ -490,27 +491,63 @@ Note that the presence of any gaps in the batches derived from L1 means that thi
...
+    i.e. a batch that changes the L1 origin cannot be processed until we have the L1 origin data.
+  - If known, then define `batch_origin` as `next_epoch`
+- `batch.epoch_num > epoch.number+1` -> `drop`: i.e. the L1 origin cannot change by more than one L1 block per L2 block.
+- `batch.epoch_hash != batch_origin.hash` -> `drop`: i.e. a batch must reference a canonical L1 origin,
+  to prevent batches from being replayed onto unexpected L1 chains.
+- `batch.timestamp > batch_origin.time + max_sequencer_drift` -> `drop`: i.e. a batch that does not adopt the next L1
+  within time will be dropped, in favor of an empty batch that can advance the L1 origin.
+- `batch.transactions`: `drop` if the `batch.transactions` list contains a transaction
+  that is invalid or derived by other means exclusively:
+  - any transaction that is empty (zero length byte string)
```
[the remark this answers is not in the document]
protolambda: Ignoring an empty tx bytestring is required since we at least need 1 byte to check the tx type.

op-node/rollup/derive/batch_queue.go
```diff
@@ -122,196 +97,132 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
...
+	// We may not have sufficient information to proceed filtering, and then we stop.
+	// There may be none: in that case we force-create an empty batch
+	nextTimestamp := l2SafeHead.Time + bq.config.BlockTime
+	var nextBatch *BatchWithL1InclusionBlock
+
+	// Go over all batches, in order of inclusion, and find the first batch we can accept.
+	// We filter in-place by only remembering the batches that may be processed in the future, or those we are undecided on.
+	remaining := bq.batches[:0]
+batchLoop:
+	for i, batch := range bq.batches {
```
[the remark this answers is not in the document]
protolambda: Changing this to be grouped by timestamp :+1:

op-node/rollup/derive/batch_queue.go
```diff
@@ -122,196 +97,132 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
...
+
+	// Find the first-seen batch that matches all validity conditions.
+	// We may not have sufficient information to proceed filtering, and then we stop.
+	// There may be none: in that case we force-create an empty batch
+	nextTimestamp := l2SafeHead.Time + bq.config.BlockTime
+	var nextBatch *BatchWithL1InclusionBlock
+
+	// Go over all batches, in order of inclusion, and find the first batch we can accept.
+	// We filter in-place by only remembering the batches that may be processed in the future, or those we are undecided on.
+	remaining := bq.batches[:0]
```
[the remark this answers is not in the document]
protolambda: Changing this anyway as part of the group timestamp thing, so it won't be a problem. Although it's somewhat tricky, I do think in-place filtering like this is an undervalued Go pattern.

op-node/rollup/derive/batch_queue.go
```diff
@@ -122,196 +97,132 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
...
-	if !ok {
-		return nil, io.EOF
-	}
+	bq.batches = remaining
 
-	// Find the first batch saved for this timestamp.
-	// Note that we expect the number of batches for the same timestamp to be small (frequently just 1 ).
-	for _, batch := range batches {
-		l1OriginTime := bq.l1Blocks[0].Time
+	if nextBatch == nil {
```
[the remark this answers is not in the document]
protolambda: Will un-nest the `nextBatch==nil` case so it's more readable :+1:

op-node/rollup/derive/batches.go
protolambda: True, although the results are different: too old = drop, too new = buffer, exact equal = process.

specs/derivation.md
```diff
@@ -490,27 +491,63 @@ Note that the presence of any gaps in the batches derived from L1 means that thi
...
-
-> **TODO** specify `max_sequencer_drift`
+A batch can have 4 different forms of validity:
+
+- `drop`: the batch is invalid, and will always be in the future, unless we reorg. It can be removed from the buffer.
+- `accept`: the batch is valid and should be processed.
+- `undecided`: we are lacking L1 information until we can proceed batch filtering.
+- `future`: the batch may be valid, but cannot be processed yet and should be checked again later.
+
+The batches are processed in order of the inclusion on L1: if multiple batches can be `accept`-ed the first is applied.
```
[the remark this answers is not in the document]
protolambda: Clarified the usage of the rules is in the given order :+1:

protolambda:
- Rebased onto `develop`
- Implemented the grouping of batches by timestamp, to handle the performance/DoS case, otherwise still in inclusion order, clarified in specs that future batches can be deferred in processing since these are marked as `future` anyway.
- Clarified a bunch of comments/code based on review, changed structure a tiny bit to make it more readable

Ready for final review/merge.

op-node/rollup/derive/attributes_queue.go
```diff
@@ -63,9 +64,14 @@ func (aq *AttributesQueue) Step(ctx context.Context, outer Progress) error {
 	}
 	batch := aq.batches[0]
 
+	safeL2Head := aq.next.SafeL2Head()
+	// sanity check parent hash
+	if batch.ParentHash != safeL2Head.Hash {
+		return NewCriticalError(fmt.Errorf("valid batch has bad parent hash %s, expected %s", batch.ParentHash, safeL2Head.Hash))
```
tynes, line 70: What exactly does a `CriticalError` indicate? Could a malicious sequencer use this line to dos verifiers?

op-node/rollup/derive/batch_queue.go
```diff
@@ -122,196 +97,140 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
...
+				"txs", len(batch.Batch.Transactions),
+				"l2_safe_head", l2SafeHead.ID(),
+				"l2_safe_head_time", l2SafeHead.Time,
+			)
+			continue
+		case BatchAccept:
+			nextBatch = batch
+			// don't keep the current batch in the remaining items since we are processing it now,
+			// but retain every batch we didn't get to yet.
+			remaining = append(remaining, candidates[i+1:]...)
```
tynes, line 165: Can this go out of bounds on the last element?

protolambda: The last element is index `len(candidates)-1`, adding `1` makes it `len(candidates)` and `candidates[len(candidates):]` is a valid slice of 0 length.

op-node/rollup/derive/attributes_queue.go
protolambda: It indicates that there's some programming error. This should never happen: the batch should never be "future" if we're looking up the batch candidates for the exact timestamp we need.

op-node/rollup/derive/batch_queue.go
```diff
@@ -122,196 +97,140 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
...
-	} else {
-		bq.log.Trace("Trying to eagerly find batch")
-		next, err := bq.tryPopNextBatch(ctx, l2SafeHead)
-		if err != nil {
-			return nil, err
-		} else {
-			bq.log.Info("found eager batch", "batch", next.Batch)
-			return []*BatchData{next.Batch}, nil
+		// sanity check
+		if nextBatch.Batch.EpochNum > rollup.Epoch(epoch.Number)+1 {
```
tynes, line 188: This check could be more strict

```
if nextBatch.Batch.EpochNum != rollup.Epoch(epoch.Number)
  && nextBatch.Batch.EpochNum != rollup.Epoch(epoch.Number)+1
```

This would silently fail for going backwards (which I assume isn't allowed). Could also just add another `if` clause that explicitly checks the going backwards case

op-node/rollup/derive/batch_queue.go
```diff
@@ -122,196 +97,140 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
...
-	l1BlockRef, err := bq.dl.L1BlockRefByNumber(ctx, batch.Batch.Epoch().Number)
-	cancel()
-	if err != nil {
-		return false, err
-	}
-
-	if l1BlockRef.Hash != batch.Batch.EpochHash {
-		bq.log.Debug("Batch epoch hash does not match expected L1 block hash", "batch_epoch", batch.Batch.Epoch(), "expected", l1BlockRef.ID())
-		return false, nil
+	validity := CheckBatch(bq.config, bq.log, bq.l1Blocks, bq.next.SafeL2Head(), &data)
```
tynes, line 112: Passing a logger into `CheckBatch` feels kinda weird to me but I get it, why use `bq.log` here instead of the defined `log` above?

op-node/rollup/derive/batch_queue.go
protolambda: `CheckBatch` already checks these edge-cases. What do you think about removing the sanity check here? It was mostly a product of testing/debugging

op-node/rollup/derive/batch_queue.go
```diff
@@ -122,196 +97,140 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
...
-// the sequencing window.
-func (bq *BatchQueue) deriveBatches(ctx context.Context, l2SafeHead eth.L2BlockRef) ([]*BatchData, error) {
+// deriveNextBatch derives the next batch to apply on top of the current L2 safe head,
+// following the validity rules imposed on consecutive batches,
+// based on currently available buffered batch and L1 origin information.
+// If no batch can be derived yet, then (nil, io.EOF) is returned.
+func (bq *BatchQueue) deriveNextBatch(ctx context.Context) (*BatchData, error) {
 	if len(bq.l1Blocks) == 0 {
-		return nil, io.EOF
+		panic("cannot derive next batch, no origin was prepared")
```
tynes, line 126: Are you sure a `panic` in the hot code path is a good idea? Just worried about a bug triggering this since there is a bunch of slicing of `bq.l1Blocks`. Could it return a scary error instead?

op-node/rollup/derive/batch_queue.go
protolambda: `CheckBatch` already adds the same logging details, so that the details are there for every caller.

op-node/rollup/derive/batch_queue.go
```diff
@@ -122,196 +97,140 @@ func (bq *BatchQueue) AddBatch(batch *BatchData) {
...
+			// don't keep the current batch in the remaining items since we are processing it now,
+			// but retain every batch we didn't get to yet.
+			remaining = append(remaining, candidates[i+1:]...)
+			break batchLoop
+		case BatchUndecided:
+			remaining = append(remaining, batch)
+			bq.batches[nextTimestamp] = remaining
+			return nil, io.EOF
+		default:
+			panic(fmt.Errorf("unknown batch validity type: %d", validity))
```
tynes, line 172: Exhaustive pattern matching is a nice feature to have in a language, the compiler would force this to be unreachable

op-node/rollup/derive/batch_queue.go
protolambda: Will change it to a scary `NewCriticalError`

op-node/rollup/derive/batch_queue.go
protolambda: Rust vs Go :smile:

op-node/rollup/derive/batches.go
```diff
@@ -1,106 +1,124 @@
...
+func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l2SafeHead eth.L2BlockRef, batch *BatchWithL1InclusionBlock) BatchValidity {
+	nextTimestamp := l2SafeHead.Time + cfg.BlockTime
+	if batch.Batch.Timestamp > nextTimestamp {
+		// no log, very common case, this happens when batches are included early or out of order.
+		return BatchFuture
 	}
-	return
-}
+	// add details to the log
+	log = log.New(
```
tynes, line 38: Is this additive in its properties or does it overwrite them? In one case, the caller passes in a logger that already has details, in another case it does not. I think the caller should be responsible for adding details to the logger before passing in

protolambda: It applies the latest properties, without mutating the logger of the caller. It copies copies the context values to a new slice within `New`, and concatenates them all with the old ones. Up to the log handler technically to pick the last context values if there are duplicates.

op-node/rollup/derive/batches.go
```diff
@@ -1,106 +1,124 @@
...
+		return BatchDrop
 	}
-	for i, txBytes := range batch.Transactions {
-		if len(txBytes) == 0 {
-			return fmt.Errorf("transaction data must not be empty, but tx %d is empty", i)
-		}
-		if txBytes[0] == types.DepositTxType {
-			return fmt.Errorf("sequencers may not embed any deposits into batch data, but tx %d has one", i)
+
+	epoch := l1Blocks[0]
```
tynes, line 73: Perhaps move this declaration to the top of the function after checking the length and use it instead of referencing `l1Blocks[0]`

op-node/rollup/derive/batches.go
```diff
@@ -1,106 +1,124 @@
...
+
+	// Check the L1 origin of the batch
+	batchOrigin := epoch
```
tynes, line 76: Why this extra intermediate variable `batchOrigin`? I don't see it changed

op-node/rollup/derive/batch_queue.go
protolambda: I'll try clean up the logging a bit

op-node/rollup/derive/batch_queue.go
tynes: I'm cool with removing this sanity check if its already checked elsewhere

specs/derivation.md
```diff
@@ -392,14 +392,15 @@ Recall that a batch contains a list of transactions to be included in a specific
...
 A batch is encoded as `batch_version ++ content`, where `content` depends on the version:
 
-| `batch_version` | `content`                                                             |
-| --------------- | --------------------------------------------------------------------- |
-| 0               | `rlp_encode([epoch_number, epoch_hash, timestamp, transaction_list])` |
+| `batch_version` | `content`                                                                          |
+| --------------- |------------------------------------------------------------------------------------|
+| 0               | `rlp_encode([parent_hash, epoch_number, epoch_hash, timestamp, transaction_list])` |
 
 where:
```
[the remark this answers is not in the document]
protolambda: Yep, same as transaction-typing, will clarify the specs :+1:

op-node/rollup/derive/batch_queue.go
protolambda: Removed the logging there, in favor of making the `CheckBatch` logging, which covers all cases now, and has warn-level if the batch needs to be dropped.

op-node/rollup/derive/batches.go
protolambda: See updates, de-duplicated this log context data.

op-node/rollup/derive/batch_queue.go
tynes: This one looks to be still not resolved

protolambda: Ah, good catch

protolambda: Confused it with a different panic that needed to be fixed in the same way

mslipper (approved):

norswap (approved):
Approving to move things along, flagging two things:

- We need to have a conversation about batch replayability (will create linear ticket).
- Still have my concerns with the state-machine-ish spec structures, but this can be easily addressed when doing l2 chain derivation refactoring work, so perfectly fine for now.
     - This will conflict with my last derivation spec PR, what I'll do there is keep both version side by side (old version just need an extra constraint for the parent hash), and we'll consolidate as part of refactors.

PR #3224, opened 2022-08-15
contracts-bedrock: deploy additional contracts

tynes:
**Description**

Deploy the `AddressManager` and the `ProxyAdmin`.
These contracts are required as part of the system on
L1. The indexer requires the `AddressManager` to be
deployed. Having the `ProxyAdmin` deployed will make
the devnet setup more similar to how the system will
be deployed in production.

Note that these deployments are only good for fresh
deployments. We will need to update the deploy scripts
to run against existing networks in the future.


packages/contracts-bedrock/deploy/005-AddressManager.ts
```diff
@@ -0,0 +1,22 @@
+import { DeployFunction } from 'hardhat-deploy/dist/types'
+import 'hardhat-deploy'
+import '@eth-optimism/hardhat-deploy-config'
+
+// TODO(tynes): This needs to be deployed for fresh networks
+// but not for upgrading existing networks
```
maurelian, line 6: What does this TODO mean in practice?

packages/contracts-bedrock/deploy/006-ProxyAdmin.ts
```diff
@@ -0,0 +1,55 @@
...
+    'AddressManager',
+    addressManager.address
+  )
+
+  const postConfig = [
+    await AddressManager.transferOwnership(admin.address),
+    await ProxyAdmin.setAddressManager(addressManager.address),
+  ]
+  await Promise.all(postConfig.map((tx) => tx.wait()))
+}
```
maurelian, line 51: It looks like this script, as well as the pre-existing ones for the messenger and bridge assume that all the contracts will be using Proxy.sol?
Do we plan to update this later to include the delegate and chugsplash proxies? 
If not, then do we even need the AddressManager here?

packages/contracts-bedrock/deploy/005-AddressManager.ts
tynes: This means that we need to configure this to run in "upgrade mode" and "devnet mode", so that it attaches to the existing address manager instead of deploying it in the code

packages/contracts-bedrock/deploy/006-ProxyAdmin.ts
tynes: For the devnet, everything currently uses `Proxy.sol` - for the "upgrade mode" we will need to take into account that not all proxies are `Proxy.sol`

tynes: We need the `AddressManager` for the indexer to be able to work against the devnet

maurelian (approved):

mslipper (approved):

PR #3225, opened 2022-08-15
indexer: clean up duplicated code

tynes:
**Description**

Remove `ParseL1Address` and `ParseL2Address` and
instead use `ParseAddress`. Addresses on L1 and L2
are the same and we can use as much upstream code
as possible instead of code from `l2geth`.


indexer/crypto.go
```diff
@@ -3,24 +3,14 @@ package indexer
 import (
 	"fmt"
 
-	l2common "github.com/ethereum-optimism/optimism/l2geth/common"
 	"github.com/ethereum/go-ethereum/common"
 )
 
-// ParseL1Address parses a L1 ETH address from a hex string. This method will
+// ParseAddress parses a L1 ETH address from a hex string. This method will
```
tuxcanfly, line 9:
```suggestion
// ParseAddress parses a ETH address from a hex string. This method will
```

tuxcanfly (approved):
OK.

mslipper (approved):

PR #3226, opened 2022-08-15
indexer: delete dead file

tynes:
**Description**

Deletes an unused file from the indexer. The debug
file is created by hardhat and will point to the build info
file and this isn't needed by the indexer.


tuxcanfly (approved):
OK.

mslipper (approved):

PR #3229, opened 2022-08-15
op-node: Cleanup channel definition

trianglesphere:
**Description**
This moves some files around and then expands upon the Channel struct.
It is better built for adding frames + then reading data when it is
ready.

**Metadata**

- Fixes ENG-2607


op-node/rollup/derive/channel.go
```diff
@@ -0,0 +1,153 @@
...
+		if !ok {
+			return false
+		}
+	}
+	return true
+}
+
+// Read full channel content (it may be incomplete if the channel is not Closed)
+// This should only be called after `IsReady` returns true.
+func (ch *Channel) Read() (out []byte) {
```
trianglesphere, line 115: TODO: make this return a reader

op-node/rollup/derive/channel.go
```diff
@@ -0,0 +1,153 @@
...
+		out = append(out, data...)
+	}
+	return out
+}
+
+// ReadBatch returns a decoded rollup batch, or an error:
+//   - io.EOF when the channel is empty
+//   - any other error (e.g. invalid compression or batch data)
+// The L1 Inclusion block is the highest L1 inclusion block inside the channel.
+func (ch *Channel) ReadBatch() (BatchWithL1InclusionBlock, error) {
```
trianglesphere, line 131: todo: make a function that takes a reader & returns a function that can be repeatedly called to pull out batches from the reader.

tynes:
devnet failed due to a compilation error, hope that its not something that broke upstream again

op-node/rollup/derive/channel_in_reader.go
```diff
@@ -53,80 +39,50 @@ func (cr *ChannelInReader) Progress() Progress {
...
 	}
 
-	var batch BatchData
-	if err := cr.ReadBatch(&batch); err == io.EOF {
+	if cr.nextBatchFn == nil {
+		return io.EOF
+	}
+
+	// TODO: can batch be non nil while err == io.EOF
+	// This depends on the behavior of rlp.Stream
```
protolambda, line 70: Good comment, checking this now

trianglesphere: one of the todo's that I have is to switch over to using a custom sentinel error `EOD` which has the semantics that if `EOD` is returned, no data is returned.

protolambda: `io.EOF` is not returned early. It does return when it attempts to read incomplete data, which can be the kind of rlp entry upon reading the first byte, or anything of the value content later. But if the underlying reader returns an early EOF after returning sufficient data, then it's zeroed, based on `readFull` decs (only reader, except for the byte-by-byte reader (which would transform an `io.EOF` upon reading a single byte to a `io.ErrUnexpectedEOF`)
```
			// Readers are allowed to give EOF even though the read succeeded.
			// In such cases, we discard the EOF, like io.ReadFull() does.
```

protolambda: Edit: on the top-level of the rlp decoding stack it does return `EOF` rather than `ErrUnexpectedEOF` when reading of the first element fails. So it will hit the `EOF` case correctly here.

protolambda (approved):

mslipper (approved):

PR #3231, opened 2022-08-16
L2 RPC refactor (depends on #3217) [bedrock]

protolambda:
**Description**

This PR:
- Moves the `l1` and `l2` sources into the same package, since they have so much in common.
- Renames `Source` (prev `l1` package) to `L1Client`
- Renames `Source` (prev `l2` package) to `L2Client`
- Factors out common functions from l1 `Source` into `EthClient`: only adding the `L1BlockRefBy{Label,Number,Hash}` methods.
- Refactors `L2Client` to wrap `EthClient`: reuse all of `EthClient`, only adding the `L2BlockRefBy{Label,Number,Hash}` methods.
- Moves execution payload fetching to `EthClient`
- `EthClient` sanity checks the block contents are compatible with `ExecutionPayload` if `MustBePostMerge` is set to true.
- Moves engine-API calls from `Source` (prev `l2` package) to `EngineClient`, which wraps `L2Client`. 
- Removes `ReadonlySource` (l2): this is what `L2Client` is, but cleaner, since it can now be reused by `EngineClient`
- Remove unused `BlockToBatch` function (Channel in derive package already does this type of conversion)
- Remove unused `FetchAllTransactions` function (with eager derivation we only read one block at a time anyway)
- Rename `L1BlockInfo` to `BlockInfo` to reuse it between L1/L2 to present header data. Remove `SequenceNumber` field, which shouldn't have been part of the struct (test was using it to add test data, but updated the test to do it better).
- Move `ComputeL2OutputRoot` to `rollup` package: this shouldn't have been in the RPC package.
- Move `AccountResult` to `eth` package: `testutils` and eth client both need it, this avoids an import cycle.
- Update `testutils` package to structure the mock RPC clients the same way: `MockEthClient`, `MockL1Client`, `MockL2Client`, `MockEngineClient`
- Adds `Unsafe`/`Safe`/`Finalized` globals to `eth` package for use in `-ByLabel` functions, which replace the previous `Head-` functions: unblocks finalized/safe RPC usage for finality + sync fixes.
- Minor test fix: update e2e L2 genesis to set the pow-nonce value to zero to be more compatible with execution-payload format.
- Remove duplicate l2 RPC client mock implementation that was used in output-root testing in favor of testutils `MockL2Client`.
- Update tests to use updated mock client types.
- Update makefile to point to `eth` package for fuzzing execution payload (`l2` package doesn't exist anymore, and doesn't contain the execution payload fuzzing for a while already, makefile was outdated)

Quite some code-moving to get this cleanup done, apologies for the large PR. Aside from improved types/caching no functionality was changed.

This PR builds on top of #3217 since that also includes RPC changes, to avoid conflicts (3217 should be merged first, since it's an important fix)

**Additional context**

This is tech 2+ month old tech debt that needed a cleanup, and now unblocks usage of the Finalized/Safe/Unsafe RPC labels across L1 and L2.

**Metadata**

Fix ENG-2280


protolambda:
Planning to split this PR into smaller pieces where possible. Working on other issues first.

PR #3233, opened 2022-08-16
op-node: fix execution payload fuzzing target [bedrock]

protolambda:
The execution payload fuzzing happens in the `eth` package, not the `l2` package.


trianglesphere (approved):
lgtm

mslipper (approved):

PR #3238, opened 2022-08-16
op-node: Add ResetDerivationPipeline RPC

trianglesphere:
**Description**
This is an RPC call this is useful for testing the robustness of
the derivation pipeline to resets.

**Additional context**
I think regularly calling this on the sequencer (maybe every 5-10 minutes)
would quickly find divergence bugs or give us the confidence that there are
not more divergence bugs. Another testing approach is to turn off the batch
submitter but to apply load to the network.


mslipper (approved):

tynes:
This RPC method cannot be exposed to the public, i feel like it should live in an `admin_*` namespace

tynes:
If we add this now then it'll be tech debt needed to clean up later. We don't want to have to force anybody running a node with public rpc access to have to run proxyd in front. The admin namespace should be another rpc server that binds to localhost, similar to how geth does it

trianglesphere:
Do we expect that the `optimism` namespace is consumable for public use?
This is probably more in the admin namesapace, but I also don't expect to provide much stability around this RPC

trianglesphere:
@tynes this is in a new namespace + has a flag to optionally enable it

op-node/node/server.go
```diff
@@ -49,6 +49,16 @@ func newRPCServer(ctx context.Context, rpcCfg *RPCConfig, rollupCfg *rollup.Conf
 	return r, nil
 }
 
+func (s *rpcServer) EnableAdminAPI(api *adminAPI) {
+	s.apis = append(s.apis, rpc.API{
+		Namespace:     "admin",
+		Version:       "",
+		Service:       api,
+		Public:        true, // TODO: this field is deprecated. Do we even need this anymore?
```
tynes, line 57: I think if we update geth, this field wont exist anymore

tynes (approved):

PR #3241, opened 2022-08-17
fix(dtl): sync bug when addresses change

smartcontracts:
**Description**
Fixes a bug in the DTL that would break sync when addresses are changed
in the AddressManager. Slightly reduces complexity at the same time by
removing an additional request.

Note that this does NOT impact mainnet because we have never changed addresses.


tynes:
We should cut a canary release here an ensure it syncs all the networks

Inphi (approved):
Tested the canary release on goerli with a l2geth verifier. The verifier is now able to sync to tip.

tynes (approved):

PR #3245, opened 2022-08-18
fix(sdk): withdrawal bug for some custom bridges

smartcontracts:
**Description**
Fixes a bug in the SDK that would cause withdrawals to revert for
certain custom bridges on Kovan.


roninjin10 (approved):

tynes:
The state surgery test is a flake

nickbalestra (approved):

PR #3252, opened 2022-08-18
contracts-bedrock: make a library call internal

tynes:
**Description**

So we don't have to handle library linking
in the state surgery


packages/contracts-bedrock/contracts/test/RLP.t.sol
```diff
@@ -10,7 +10,7 @@ import { Bytes32AddressLib } from "@rari-capital/solmate/src/utils/Bytes32Addres
 library LibRLP {
     using Bytes32AddressLib for bytes32;
 
-    function computeAddress(address deployer, uint256 nonce) public pure returns (address) {
+    function computeAddress(address deployer, uint256 nonce) internal pure returns (address) {
```
maurelian, line 13: This is in a test file. Why is it an issue for state surgery?

maurelian (approved):

packages/contracts-bedrock/contracts/test/RLP.t.sol
tynes: Its an issue because artifacts are created for it

PR #3253, opened 2022-08-18
state-surgery: handle fs errors

tynes:
**Description**

Handle errors when walking the filesystem so that they do
not silently fail.


trianglesphere (approved):
lgtm. QQ - is this code getting listed with the same checks as other go code? The missing err handling should have been flagged in a lint

tynes:
> lgtm. QQ - is this code getting listed with the same checks as other go code? The missing err handling should have been flagged in a lint

It is not, we should def add this

PR #3266, opened 2022-08-19
contracts-bedrock: L2CrossDomainMessenger legibility

tynes:
**Description**

The `initialize()` function in the L2CrossDomainMessenger
sets some blocked system addresses. It was using `address(this)`
but in reality the contract is a predeploy at a specific address.
It should not be deployed, the `initialize` function will never
run. For the unit tests, the contracts are placed at their correct
addresses. This makes the testing environment as close as possible
to what is experienced on a live network.

This PR hardcodes in the L2 cross domain messenger constant
from the predeploys library instead of using `address(this)`.
In practice this value is set using the chain ops tooling,
so this makes it more clear as to what the value should be.


maurelian (approved):
Could you please also mirror this in the fixes repo?

PR #3273, opened 2022-08-19
ci: Add hive to pipeline

mslipper:


.circleci/config.yml
```diff
@@ -593,6 +596,50 @@ jobs:
           name: "Go mod tidy"
           command: make mod-tidy && git diff --exit-code
 
+  build-hive:
+    docker:
+      - image: cimg/go:1.18
```
trianglesphere, line 601: maybe separate, but switch to go1.19?

trianglesphere:
an intermediate build step failed

.circleci/config.yml
mslipper: Suce, I can do that. Hive builds on ancient versions of Go so it shouldn't be a big issue.

mslipper:
Yeah, seeing I/O errors pulling Docker containers. Looking into it.

tynes (approved):

.circleci/config.yml
```diff
@@ -593,6 +596,40 @@ jobs:
...
+      - go/mod-download
+      - go/save-cache
+      - run: {command: "go build ."}
+      - run:
+          command: |
+            ./hive \
+            -sim=<<parameters.sim>> \
+            -sim.loglevel=5 \
+            -docker.pull=true \
+            -client=go-ethereum,op-geth_optimism-history,op-proposer_<<parameters.version>>,op-batcher_<<parameters.version>>,op-node_<<parameters.version>>,op-contracts_<<parameters.version>> |& tee /tmp/hive.log
```
protolambda, line 623: Is there some way to confirm it pulls the correct images for the hive client containers, i.e. if the images that we pushed as dependency are available soon enough on the docker repo to be used in the CI hive build?

protolambda (approved):
Question about cycle time between image build and hive usage, but LGTM otherwise.

.circleci/config.yml
mslipper: Hive will fail if the passed-in image doesn't exist on the repo.

mslipper: I confirmed it's pulling the correct containers by cross referencing the params with the logs.

PR #3274, opened 2022-08-19
op-node: Invert L1 Traversal stage to be pull instead of push

trianglesphere:
**Description**
Proof of concept about switching the L1 Traversal stage to be pull based.
The plan is to work my way up the derivation pipeline such that every stage
becomes pull based. Then I can remove the `Progress` API.

**Additional Context**
The end to end tests work but the unit tests don't b/c I didn't update the mocks.

### DIFF
--- a/op-node/rollup/derive/l1_retrieval.go
+++ b/op-node/rollup/derive/l1_retrieval.go
@@ -34,6 +34,12 @@ type L1Retrieval struct {
 	log     log.Logger
 	dataSrc DataAvailabilitySource
 	next    L1SourceOutput
+	prev    *L1Traversal
+
+	// We maintain a `done` flag separate from `progress.Closed`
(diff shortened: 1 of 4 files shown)


op-node/rollup/derive/l1_retrieval.go
```diff
@@ -101,6 +114,8 @@ func (l1r *L1Retrieval) Step(ctx context.Context, outer Progress) error {
 func (l1r *L1Retrieval) ResetStep(ctx context.Context, l1Fetcher L1Fetcher) error {
 	l1r.progress = l1r.next.Progress()
 	l1r.datas = nil
+	l1r.done = false
 	l1r.data = nil
+	l1r.log.Info("Reset of L1Retrieval done", "origin", l1r.progress.Origin)
```
tynes, line 119: This feels like it should not be logged on info level

PR #3276, opened 2022-08-20
indexer: rm AddressManager, StateCommitmentChain

tuxcanfly:
**Description**
This PR removes types not required for indexing bedrock anymore - `AddressManager` `StateCommitmentChain`

**Metadata**
- Fixes ENG-2664


indexer/config.go
```diff
@@ -121,18 +118,17 @@ type Config struct {
...
-		DBPort:                  ctx.GlobalUint64(flags.DBPortFlag.Name),
-		DBUser:                  ctx.GlobalString(flags.DBUserFlag.Name),
-		DBPassword:              ctx.GlobalString(flags.DBPasswordFlag.Name),
-		DBName:                  ctx.GlobalString(flags.DBNameFlag.Name),
+		BuildEnv:           ctx.GlobalString(flags.BuildEnvFlag.Name),
+		EthNetworkName:     ctx.GlobalString(flags.EthNetworkNameFlag.Name),
+		ChainID:            ctx.GlobalInt64(flags.ChainIDFlag.Name),
+		L1EthRpc:           ctx.GlobalString(flags.L1EthRPCFlag.Name),
+		L2EthRpc:           ctx.GlobalString(flags.L2EthRPCFlag.Name),
+		L2GenesisBlockHash: ctx.GlobalString(flags.L2GenesisBlockHashFlag.Name),
```
tynes, line 126: What do you think about removing the L2 genesis block hash and instead use an L1 chain id and L2 chain id flag for validation that its talking to the correct chains?

tynes:
This just means that no new release of the indexer can come out until bedrock ships, which i think is fine

tynes:
Looks like there is a linting error

mslipper (approved):
LGTM assuming you fix the linting issue.

indexer/config.go
mslipper: +1 on this. Can be handled separately but it's easier to pass in a chain ID than a genesis hash.

tuxcanfly: OK, think it was being used to partially index the chain (flag should be called `L2StartBlockHashFlag`), but I don't see it being much useful outside testing. Will remove.

tuxcanfly: Will handle this in a new PR.

PR #3278, opened 2022-08-21
chore: Release Bedrock v0.5.0

mslipper:
This PR upgrades Bedrock dependencies v0.5.0. Hive needs this in order to perform withdrawals properly since v0.4.0 does not include the change to index withdrawals by block number.

It also adds a Makefile command to generate future Bedrock version tags and push them.

Note the GOPRIVATE change in `mod-tidy` - I needed to add this to circumvent the Go version proxy since it's slow to index new versions.


tynes:
Would super appreciate it if op chain ops could get into this too

mslipper:
Sure - I'll merge the chain ops rename once test pass then add that to this.

protolambda:
Technically we can version the modules independently with different versions. Right now there's still enough churn in the code to make releases across all modules worth it, but in the future we really don't need e.g. a op-proposer/v0.8.0 if there are no changes since v0.7.0: it would just be noisy.

So lgtm, but will probably revisit this once the release cycle becomes more stable.

mslipper:
@tynes I'll make a separate PR to add op-chain-ops since the tests are failing.

PR #3279, opened 2022-08-21
ci: Add hive l1ops suite

mslipper:
This suite tests deposits and withdrawals.


protolambda (approved):
Need to fix the job name, LGTM otherwise.

PR #3283, opened 2022-08-22
fix: proxy admin deploy script

tynes:
**Description**
Reduce the amount of config by removing the concept of proxy admin
address because we should be using the address of the deployed
proxy admin contract.


op-chain-ops/genesis/config.go
```diff
@@ -157,7 +156,8 @@ func NewStorageConfig(hh *hardhat.Hardhat, config *DeployConfig, chain ethereum.
 	storage["GovernanceToken"] = state.StorageValues{
 		"_name":   "Optimism",
 		"_symbol": "OP",
-		"_owner":  config.ProxyAdmin,
+		// TODO: this should be set to the MintManager
```
tynes, line 159: Will follow up with a PR that addresses this

maurelian (approved):

PR #3284, opened 2022-08-22
contracts-bedrock: parallel deployment fix

tynes:
**Description**

Uses local nonce management for sending configuration
transactions. This will make the parallel deployments
more reliable.


mslipper (approved):

optimisticben (approved):
I was able to deploy with commit fcde9dd60d29c5f3aecae98720a914eed31eb0a3

PR #3285, opened 2022-08-22
contracts-bedrock: remove fund dev accounts

tynes:
**Description**
Always fund the dev accounts since these scripts will only be used for devnet. Begin to refactor the `op-chain-ops` so that it can be used with a live db instead of just the memory db and clean up the `genesis` package. Still need some more refactoring for `op-chain-ops` to work on a live db, but this should be ok for now since its more explicit


packages/contracts-bedrock/hardhat.config.ts
```diff
@@ -320,10 +320,6 @@ const config: HardhatUserConfig = {
       type: 'number',
       default: 6,
     },
-    fundDevAccounts: {
-      type: 'boolean',
-      default: false,
-    },
```
protolambda, line 326: Can you update the testnet setup tool to remove this flag too? `bedrock-regentool/regentool/config.py` uses this variable.

protolambda (approved):

packages/contracts-bedrock/hardhat.config.ts
tynes: Its not merged yet. https://github.com/ethereum-optimism/optimism/pull/3243

protolambda: I mean the script in the k8s repository

mslipper (approved):

PR #3287, opened 2022-08-22
feat: generate genesis with op-node

tynes:
**Description**

Adds the `genesis` subcommand to the `op-node` so that it can be used to create genesis files. To create a L2 developer genesis file that is suitable for devnets, use the following command:

```bash
$ op-node genesis devnet-l2
```

This name was chosen such that it is obvious that it is creating a genesis block for a devnet instead of trying to upgrade an existing database. The code to upgrade an existing database will be implemented in the future.

This PR also includes a fix to `op-chain-ops` because it was skipping the correct implementation of the `OptimismMintableERC20Factory`, which is required for ERC20 deposits to work.

The hardhat task responsible for checking that a L2 node's predeploys were configured properly was updated to catch this sort of error and then moved to the front of the `devnet` CI job so that it can fail before a mysterious error message comes from the sdk trying to do a deposit.

### DIFF
--- a/op-node/cmd/genesis/cmd.go
+++ b/op-node/cmd/genesis/cmd.go
@@ -1,0 +1,104 @@
+package genesis
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/urfave/cli"
+
+	"github.com/ethereum-optimism/optimism/op-chain-ops/genesis"
+	"github.com/ethereum-optimism/optimism/op-chain-ops/hardhat"
+	"github.com/ethereum/go-ethereum/log"
+
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+var Subcommands = cli.Commands{
+	{
+		Name:  "devnet-l2",
+		Usage: "Initialized a new devnet genesis file",
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "artifacts",
+				Usage: "Comma delimeted list of hardhat artifact directories",
+			},
+			cli.StringFlag{
+				Name:  "network",
+				Usage: "Name of hardhat deploy network",
+			},
+			cli.StringFlag{
+				Name:  "deployments",
+				Usage: "Comma delimated list of hardhat deploy artifact directories",
+			},
+			cli.StringFlag{
+				Name:  "deploy-config",
+				Usage: "Path to hardhat deploy config directory",
+			},
+			cli.StringFlag{
+				Name:  "rpc-url",
+				Usage: "L1 RPC URL",
+			},
+			cli.StringFlag{
+				Name:  "outfile",
+				Usage: "Path to file to write output to",
+			},
+		},
+		Action: func(ctx *cli.Context) error {
+			// Turn off logging for this command unless it is a critical
+			// error so that the output can be piped to jq
+			log.Root().SetHandler(
+				log.LvlFilterHandler(
+					log.LvlCrit,
+					log.StreamHandler(os.Stdout, log.TerminalFormat(true)),
+				),
+			)
+
+			artifact := ctx.String("artifacts")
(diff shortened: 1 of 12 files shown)


ops-bedrock/devnet-up.sh
```diff
@@ -88,12 +91,17 @@ else
...
-      npx hardhat --network $NETWORK genesis-l2
-      mv genesis.json ../../.devnet/genesis-l2.json
+      cd $OP_NODE
+      go run cmd/main.go genesis devnet-l2 \
+          --artifacts $CONTRACTS_BEDROCK/artifacts,$CONTRACTS_GOVERNANCE/artifacts \
+          --network $NETWORK \
+          --deployments $CONTRACTS_BEDROCK/deployments \
+          --deploy-config $CONTRACTS_BEDROCK/deploy-config \
+          --rpc-url http://localhost:8545 \
+          --outfile $DEVNET/genesis-l2.json
```
tuxcanfly, line 104: Same as above, `genesis-l2.json` was 2 dirs up.

ops-bedrock/devnet-up.sh
```diff
@@ -109,20 +117,20 @@ fi
...
 
 # Start putting together the rollup config.
-if [ ! -f ./.devnet/rollup.json ]; then
+if [ ! -f $DEVNET/rollup.json ]; then
     (
       echo "Building rollup config..."
       cd $CONTRACTS_BEDROCK
       npx hardhat --network $NETWORK rollup-config
-      mv rollup.json ../../.devnet/rollup.json
+      mv rollup.json $DEVNET/rollup.json
```
tuxcanfly, line 125: Same as above.

op-node/cmd/genesis/cmd.go
```diff
@@ -0,0 +1,104 @@
...
+			if err != nil {
+				return err
+			}
+
+			gen, err := genesis.BuildOptimismDeveloperGenesis(hh, config, client)
+			if err != nil {
+				return err
+			}
+
+			file, _ := json.MarshalIndent(gen, "", " ")
```
tuxcanfly, line 93: better to return err

op-node/cmd/genesis/cmd.go
```diff
@@ -0,0 +1,104 @@
...
+
+			outfile := ctx.String("outfile")
+			if outfile == "" {
+				fmt.Println(string(file))
+			} else {
+				_ = ioutil.WriteFile(outfile, file, 0644)
```
tuxcanfly, line 99: can just call `os.WriteFile`: https://pkg.go.dev/io/ioutil#WriteFile

also return err

tuxcanfly (approved):
Ok.

mslipper (changes requested):
Agree with Javed that we need to handle write errors here. LGTM otherwise.

ops-bedrock/devnet-up.sh
```diff
@@ -59,12 +62,12 @@ mkdir -p ./.devnet
...
 # Regenerate the L1 genesis file if necessary. The existence of the genesis
 # file is used to determine if we need to recreate the devnet's state folder.
-if [ ! -f ./.devnet/genesis-l1.json ]; then
+if [ ! -f $DEVNET/genesis-l1.json ]; then
   echo "Regenerating L1 genesis."
   (
     cd $CONTRACTS_BEDROCK
     npx hardhat --network $NETWORK genesis-l1 --outfile genesis-l1.json
-    mv genesis-l1.json ../../.devnet/genesis-l1.json
+    mv genesis-l1.json $DEVNET/genesis-l1.json
```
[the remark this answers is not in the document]
tuxcanfly: Nevermind, the path is absolute now, so this should be good.

tynes:
thanks for the feedback, good calls, will fix

tynes:
> Agree with Javed that we need to handle write errors here. LGTM otherwise.

I've addressed the error handling and fixed the merge conflict

mslipper (approved):

PR #3288, opened 2022-08-22
op-node: split l1 source into eth and l1 client

protolambda:
In preparation of refactoring the L2 RPC, this splits the L1 RPC  source functionality into a more general eth client, and a L1 client that wraps it.

### Changes

- Move L1 source into `sources` package
- Split L1 source into eth client (to be shared with L2 client code) and L1 client (adds the L1 specific client methods)
- Move `AccountResult` to `eth` package so it can be shared/used everywhere
- Move `ComputeL2OutputRoot` to `rollup` package so it can be used without importing a RPC bindings package
- Change `L1Info` to more general `BlockInfo`, and add coinbase header field
- Remove `FetchAllTransactions` from L1 source bindings (unused for a long time now)
- Cleanup Eth client testing a bit (used to be L1 client test)
- Mirror client bindings structure in mock test utils
- Remove unused `BlockToBatch`

#### Old RPC structure.

rpc_old

#### New RPC structure

rpc_new


op-node/sources/eth_client.go
```diff
@@ -58,39 +58,16 @@ func (c *SourceConfig) Check() error {
...
-		TrustRPC: trustRPC,
-	}
-}
-
-type batchCallContextFn func(ctx context.Context, b []rpc.BatchElem) error
-
-// Source to retrieve L1 data from with optimized batch requests, cached results,
-// and flag to not trust the RPC.
-type Source struct {
+// EthClient retrieves ethereum data from with optimized batch requests, cached results, and flag to not trust the RPC.
```
tuxcanfly, line 61:
```suggestion
// EthClient retrieves ethereum data with optimized batch requests, cached results, and flag to not trust the RPC.
```

op-node/eth/label.go
```diff
@@ -0,0 +1,18 @@
...
+	// Unsafe is:
+	// - L1: absolute head of the chain
+	// - L2: absolute head of the chain, not confirmed on L1
+	Unsafe = "latest"
+	// Safe is:
+	// - L1: Justified checkpoint
+	// - L2: Derived chain tip from L1 data
+	Safe = "safe"
+	// Finalized is:
+	// - L1: Finalized checkpoint
```
tuxcanfly, line 15: Would be nice to have an explanation of justified vs finalized checkpoint.

op-node/rollup/derive/l1_block_info_test.go
```diff
@@ -12,39 +12,45 @@ import (
...
-		})},
-		{"all zero", func(rng *rand.Rand) *testutils.MockL1Info {
-			return &testutils.MockL1Info{InfoBaseFee: new(big.Int)}
+		}), randomSeqNr},
+		{"zero seq", testutils.MakeBlockInfo(nil), func(rng *rand.Rand) uint64 {
+			return 0
 		}},
+		{"all zero", func(rng *rand.Rand) *testutils.MockBlockInfo {
+			return &testutils.MockBlockInfo{InfoBaseFee: new(big.Int)}
+		}, randomSeqNr},
```
tuxcanfly, line 46: the test name reads "all zero", should be it passing `func(rng ...) { return 0 }` instead of `randomSeqNr`?

op-node/sources/l1_client.go
```diff
@@ -0,0 +1,95 @@
...
+			MaxConcurrentRequests: 10,
+			TrustRPC:              trustRPC,
+		},
+		L1BlockRefsCacheSize: span,
+	}
+}
+
+// L1Client to retrieve L1 data from with optimized batch requests, cached results,
+// and flag to not trust the RPC.
+type L1Client struct {
```
tuxcanfly, line 43: struct comment here could be better.

tuxcanfly (approved):
OK.

op-node/rollup/derive/l1_block_info_test.go
protolambda: nice catch

op-node/eth/label.go
```diff
@@ -0,0 +1,19 @@
+package eth
+
+type BlockLabel string
+
+const (
```
tynes, line 5: Perhaps this type in upstream is useful?
https://github.com/ethereum/go-ethereum/blob/6d711f0c001ccb536c5ead8bd5d07828819e7d61/rpc/types.go#L61

op-node/rollup/derive/pipeline.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 type L1Fetcher interface {
-	L1HeadBlockRef(ctx context.Context) (eth.L1BlockRef, error)
+	L1BlockRefByLabel(ctx context.Context, label eth.BlockLabel) (eth.L1BlockRef, error)
```
tynes, line 14: Perhaps the rpc type from upstream is useful here

op-node/eth/label.go
protolambda: I don't particularly like using the negative integers to present the special values, and if we ever end up adding our own it'll be less flexible. That said, maybe in a future PR we can consolidate the `ByLabel/ByNumber/ByHash/ByRpcNumber` variants of essentially the same thing all into one function which takes some interface that allows the argument to be encoded for the request, which can then be implemented by any of these types.

op-node/eth/account_proof.go
```diff
@@ -82,40 +64,3 @@ func (res *AccountResult) Verify(stateRoot common.Hash) error {
 	}
 	return err
 }
-
-// BlockToBatch converts a L2 block to batch-data.
-// Invalid L2 blocks may return an error.
```
tynes, line 87: Is this no longer used? Looking for where it was moved to

tynes (approved):

op-node/eth/account_proof.go
protolambda: Forgot to list this in the PR description, sorry. That's indeed unused code. The channel encoding code in the `derive` package has its own implementation of this.

op-node/rollup/derive/pipeline.go
protolambda: Will open a ticket to consolidate `ByLabel`/`ByNumber`/`ByHash`/`ByRPCNumber`. There are 4 query param types I'm aware of, we can just generalize it, and support upstream param types as part of that.

op-node/eth/label.go
protolambda: Opened ENG-2673 to address this in some later PR.

PR #3291, opened 2022-08-23
op-node: l2 rpc refactor

protolambda:
**Description**

- Refactors previous `l2.Source` into `sources.L2Client`, now sharing a lot of commons with `sources.L1Client` through `sources.EthClient`
- Remove `l2.ReadOnlySource`: this was duplicate legacy code from the first op-proposer iteration. We have `sources.L2Client` now with caching + deduplicated code with L1 client.
- Add `sources.EngineClient`: wraps the `sources.L2Client` with the 3 engine API methods we use. (note: upstream execution APIs state that the `eth` namespace is part of the engine API endpoint in addition to the `engine` namespace, so this works).
- Refactor `sources/types.go` to:
  - Be more strict about blocks retrieved from post-merge RPC sources. Execution payloads are a subset of the valid block responses we can get.
    - As part of the above: fix the genesis setup in e2e to set the PoW nonce value to 0
  - Define a json definition for the header type. Could improve this more, but the intention is to remove `big.Int` out of the decoding/encoding picture, make the type flat instead of a pointer structure, and have more control to implement stricter checks on the contents.
- Port over and optimize `Payload` retrieval methods to EthClient. (Hopefully we can use this instead of block retrieval for L1 at some point. Especially if we can read raw transactions with some future RPC improvements, instead of the slow transaction encoding/decoding roundtrips we are forced to have now).
- Remove duplciate/outdated `mockL2Client`: `testutils` has everything any test may need for RPC mocking.

See previous L1/eth RPC refactor PR for more context: #3288

**Metadata**

Fix ENG-2280


op-node/sources/engine_client.go
```diff
@@ -0,0 +1,110 @@
...
+	err := s.client.CallContext(fcCtx, &result, "engine_forkchoiceUpdatedV1", fc, attributes)
+	if err == nil {
+		e.Trace("Shared forkchoice-updated signal")
+		if attributes != nil {
+			e.Trace("Received payload id", "payloadId", result.PayloadID)
+		}
+		return &result, nil
+	} else {
+		e = e.New("err", err)
+		if rpcErr, ok := err.(rpc.Error); ok {
```
mslipper, line 61: Could pull this out into a helper.

op-node/sources/engine_client.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *EngineClient) ForkchoiceUpdate(ctx context.Context, fc *eth.ForkchoiceState, attributes *eth.PayloadAttributes) (*eth.ForkchoiceUpdatedResult, error) {
+	e := s.log.New("state", fc, "attr", attributes)
+	e.Trace("Sharing forkchoice-updated signal")
+	fcCtx, cancel := context.WithTimeout(ctx, time.Second*5)
+	defer cancel()
+	var result eth.ForkchoiceUpdatedResult
+	err := s.client.CallContext(fcCtx, &result, "engine_forkchoiceUpdatedV1", fc, attributes)
+	if err == nil {
+		e.Trace("Shared forkchoice-updated signal")
+		if attributes != nil {
```
tynes, line 55: Why does this check attributes here but then reference result? It looks like this if not nil check is to prevent referencing a nil pointer

op-node/sources/engine_client.go
```diff
@@ -0,0 +1,110 @@
...
+	var result eth.ExecutionPayload
+	err := s.client.CallContext(ctx, &result, "engine_getPayloadV1", payloadId)
+	if err != nil {
+		e = e.New("payload_id", payloadId, "err", err)
+		if rpcErr, ok := err.(rpc.Error); ok {
+			code := eth.ErrorCode(rpcErr.ErrorCode())
+			if code != eth.UnavailablePayload {
+				e.Warn("unexpected error code in get-payload response", "code", code)
+			} else {
+				e.Warn("unavailable payload in get-payload request")
```
tynes, line 101: Should this logline also include the code?

op-node/sources/eth_client.go
```diff
@@ -182,6 +216,21 @@ func (s *EthClient) InfoAndTxsByLabel(ctx context.Context, label eth.BlockLabel)
 	return s.blockCall(ctx, "eth_getBlockByNumber", string(label))
 }
```
tynes, line 218: Longer term these should definitely be consolidated to work with a type similar to upstream's `rpc.BlockNumberOrHash`

op-node/sources/eth_client.go
```diff
@@ -224,7 +273,7 @@ func (s *EthClient) BlockIDRange(ctx context.Context, begin eth.BlockID, max uin
 			if result == nil {
 				break // no more headers from here
 			}
-			info, err := result.Info(s.trustRPC)
+			info, err := result.Info(s.trustRPC, s.mustBePostMerge)
```
tynes, line 276: I think its slightly ugly to pass in 2 args that are static properties to `hdr.Info`, I feel like having a method on the `EthClient` that you pass a `hdr` to that is optionally used before inserting an item into the cache or returning remote data to the caller would feel cleaner

op-node/sources/l2_client.go
```diff
@@ -0,0 +1,124 @@
...
+
+	Genesis rollup.Genesis
+}
+
+func L2ClientDefaultConfig(config *rollup.Config, trustRPC bool) *L2ClientConfig {
+	// Cache 3/2 worth of sequencing window of payloads, block references, receipts and txs
+	span := int(config.SeqWindowSize) * 3 / 2
+	// Estimate number of L2 blocks in this span of L1 blocks
+	// (there's always one L2 block per L1 block, L1 is thus the minimum, even if block time is very high)
+	if config.BlockTime < 12 && config.BlockTime > 0 {
```
tynes, line 29: I don't think its a bad idea to add some loglines here, geth does this when sanitizing config. A bug resulting from silent modifications of config can be really annoying

tynes (commented):
Generally looks good to me, left some comments

op-node/sources/engine_client.go
```diff
@@ -0,0 +1,110 @@
...
+	e.Trace("sending payload for execution")
+
+	execCtx, cancel := context.WithTimeout(ctx, time.Second*5)
+	defer cancel()
+	var result eth.PayloadStatusV1
+	err := s.client.CallContext(execCtx, &result, "engine_newPayloadV1", payload)
+	e.Trace("Received payload execution result", "status", result.Status, "latestValidHash", result.LatestValidHash, "message", result.ValidationError)
+	if err != nil {
+		e.Error("Payload execution failed", "err", err)
+		return nil, fmt.Errorf("failed to execute payload: %v", err)
```
tuxcanfly, line 83: Is this worth wrapping?

op-node/sources/types.go
```diff
@@ -83,71 +87,175 @@ func (info *HeaderInfo) ReceiptHash() common.Hash {
...
+	if hdr.BaseFee == nil {
+		return fmt.Errorf("post-merge block header requires EIP-1559 basefee field, but got %s", hdr.BaseFee)
+	}
+	if len(hdr.Extra) > 32 {
+		return fmt.Errorf("post-merge block header requires 32 or less bytes of extra data, but got %d", len(hdr.Extra))
+	}
+	if hdr.UncleHash != types.EmptyUncleHash {
+		return fmt.Errorf("post-merge block header requires uncle hash to be of empty uncle list, but got %s", hdr.UncleHash)
+	}
+	return nil
```
tuxcanfly, line 134: Should we additionally check `MixDigest != 0x000` here?

op-node/sources/types.go
```diff
@@ -83,71 +87,175 @@ func (info *HeaderInfo) ReceiptHash() common.Hash {
...
+	TxHash      common.Hash      `json:"transactionsRoot"`
+	ReceiptHash common.Hash      `json:"receiptsRoot"`
+	Bloom       eth.Bytes256     `json:"logsBloom"`
+	Difficulty  hexutil.Big      `json:"difficulty"`
+	Number      hexutil.Uint64   `json:"number"`
+	GasLimit    hexutil.Uint64   `json:"gasLimit"`
+	GasUsed     hexutil.Uint64   `json:"gasUsed"`
+	Time        hexutil.Uint64   `json:"timestamp"`
+	Extra       hexutil.Bytes    `json:"extraData"`
+	MixDigest   common.Hash      `json:"mixHash"`
```
tuxcanfly, line 104: Got confused between `MixDigest` and `mixHash`. Why not use a consistent name?

op-node/sources/types.go
```diff
@@ -83,71 +87,175 @@ func (info *HeaderInfo) ReceiptHash() common.Hash {
...
+
+	// untrusted info included by RPC, may have to be checked
 	Hash common.Hash `json:"hash"`
 }
 
-type rpcHeader struct {
-	cache  rpcHeaderCacheInfo
-	header types.Header
+// checkPostMerge checks that the block header meets all criteria to be a valid ExecutionPayloadHeader,
+// see EIP-3675 (block header changes) and EIP-4399 (mixHash usage for prev-randao)
```
tuxcanfly, line 115: These EIPs also mention `ommers` and `ommersHash`. Is that something we need to consider?

op-node/sources/types.go
tuxcanfly: Guessing `0x00` is perfectly valid random hash, but need a way to check if it's actively set I guess.

op-node/sources/engine_client.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	return &EngineClient{
+		L2Client: l2Client,
+	}, nil
+}
+
+// ForkchoiceUpdate updates the forkchoice on the execution client. If attributes is not nil, the engine client will also begin building a block
+// based on attributes after the new head block and return the payload ID.
+// May return an error in ForkChoiceResult, but the error is marshalled into the error return
```
tuxcanfly, line 45: The statement about `ForkChoiceResult` error could be a bit clearer. Where is the error being marshalled?

mslipper (approved):
My feedback is non blocking. OK to merge as-is.

tuxcanfly (approved):
OK.

op-node/node/node.go
```diff
@@ -139,33 +137,31 @@ func (n *OpNode) initL2(ctx context.Context, cfg *Config, snapshotLog log.Logger
...
-	n.l2Client = client.NewInstrumentedClient(rpcClient, n.metrics)
-	source, err := l2.NewSource(n.l2Node, n.l2Client, &cfg.Rollup.Genesis, n.log)
+
+	n.l2Source, err = sources.NewEngineClient(
+		client.NewInstrumentedRPC(rpcClient, n.metrics), n.log, n.metrics.L2SourceCache,
+		sources.EngineClientDefaultConfig(&cfg.Rollup),
+	)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create Engine client: %v", err)
```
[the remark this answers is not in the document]
tynes: Like `%w` right?

op-node/sources/l2_client.go
```diff
@@ -0,0 +1,124 @@
...
+		EthClientConfig: EthClientConfig{
+			// receipts and transactions are cached per block
+			ReceiptsCacheSize:     span,
+			TransactionsCacheSize: span,
+			HeadersCacheSize:      span,
+			PayloadsCacheSize:     span,
+			MaxRequestsPerBatch:   20, // TODO: tune batch param
+			MaxConcurrentRequests: 10,
+			TrustRPC:              trustRPC,
+			MustBePostMerge:       true,
```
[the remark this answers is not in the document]
tynes: I think that `MustBePostMerge` should be configurable, its only going to cause a headache for projects that want to use our codebase on altchains

op-node/node/node.go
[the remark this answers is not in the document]
tuxcanfly: yep

op-node/sources/l2_client.go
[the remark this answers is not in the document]
tynes: Ideally any sort of local validation should be configurable with defaults to true so that altcoins can explicitly opt out of security to run on their chain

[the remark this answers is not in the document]
protolambda: It's always `true` for L2, since the L2 is the post-merge chain we have with op-geth. For L1 it defaults to`false`, and although we could set it to `true` post-merge on ethereum, we don't strictly have to, so I'm going to leave the default as-is and revisit it later.

op-node/sources/engine_client.go
protolambda: `attributes` is an optional argument, not all forkchoice updates initiate block building. I'll add a comment

op-node/sources/engine_client.go
protolambda: Unfortunately there is an edge case in the other similar method, and there's no other code-duplication of this check, so we end up only adding more code when factoring it into a helper. This code is moving as part of this refactor, but didn't change, so I prefer to keep it like this for now.

op-node/sources/engine_client.go
protolambda: Didn't make any engine client implementation changes, just moved the code into the new engine client type, but this is indeed useful to add to the log, so I'll add it :+1:

op-node/sources/eth_client.go
protolambda: Agreed, `ENG-2673` covers this

op-node/sources/eth_client.go
protolambda: Agreed, although I'm planning to remove this `BlockIDRange` method since it's unused, so I think I'll keep this as is and open a separate PR to remove the method.

op-node/sources/l2_client.go
protolambda: It's not modifying any config value, it's calculating a new one. Right now this default-config-generating function is a pure function without logging/error edge cases, And no changes to the input config, just a new output. So I think it's fine.

op-node/sources/types.go
```diff
@@ -83,71 +87,175 @@ func (info *HeaderInfo) ReceiptHash() common.Hash {
...
+		if err := block.verify(); err != nil {
+			return nil, err
 		}
 	}
-	return info, block.extra.Transactions, nil
+	var baseFee uint256.Int
+	baseFee.SetFromBig((*big.Int)(block.BaseFee))
+
+	// Unfortunately eth_getBlockByNumber either returns full transactions, or only tx-hashes.
+	// There is no option for encodes transactions.
```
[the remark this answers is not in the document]
protolambda: For L2 we can, and it's definitely an optimization that's worth looking at, but for L1 we can't change it. However, we use this primarily to just check if the derived block attributes data matches the L2 chain, so maybe there's a way to refactor it to only check the tx-root from an `ExecutionPayloadHeader` instead of full payload. That's worth a separate PR if we're refactoring derivation related code though, I'm ok with the way this is now, my previous L2 payload fetching code was worse :upside_down_face:

op-node/sources/types.go
protolambda: It'll still fit in the execution payload format, and the execution payload attributes are checked, so I don't think we've to be so strict with that attribute as with the others which are not fully compatible with the new `ExecutionPayload` type post-merge.

op-node/sources/types.go
protolambda: `mixHash` is used in the JSON RPC + specs/EIPs, but `MixDigest` is used in the geth header/block types. And it may change to `PrevRandao` once the json field is finally renamed. Going to leave it consistent with geth for now.

op-node/sources/types.go
protolambda: That's the same as the uncle hash, and is already checked. The EIPs follow yellow paper naming, but "sha3Uncles" is used in the RPC and geth.

op-node/sources/engine_client.go
protolambda: Good catch, it's legacy from when we still tried to present the non-success status values with a regular error. But we don't anymore. I'll update the description to clarify the exact error usage.

protolambda: Will also update the `NewPayload` doc :+1:

tynes (approved):

PR #3292, opened 2022-08-23
op-bindings: add OptimismMintableERC20Factory

tynes:
**Description**

Was forgotten because it lives in the `universal`
directory and not the `L2` directory


protolambda (approved):

tuxcanfly (approved):
OK.

PR #3293, opened 2022-08-23
op-bindings: add more bindings

tynes:
**Description**
Adds more bindings that will be useful for the indexer and the L2 genesis creation


protolambda (approved):

tuxcanfly (approved):
OK.

PR #3298, opened 2022-08-23
op-node: L1 genesis generation

tynes:
**Description**
Generate L1 genesis files using the `op-node` and `op-chain-ops` tooling. This replaces the hardhat task in the devnet setup that was used previously.


op-chain-ops/genesis/setters.go
```diff
@@ -147,3 +147,14 @@ func MigrateDepositHashes(hh *hardhat.Hardhat, db vm.StateDB) error {
 	})
 	return nil
 }
+
+// SetPrecompileBalances will set a single wei at each precompile address.
+// This is an optimization to make calling them cheaper. This should only
+// be used for devnets.
+func SetPrecompileBalances(db vm.StateDB) {
+	for i := 0; i <= 9; i++ {
```
protolambda, line 155: Can we do this for the first 256 addresses? This is what Geth does during genesis setup: https://github.com/ethereum/go-ethereum/blob/4c114af5024943a7471aeccdf5e536584c1e21b4/cmd/puppeth/wizard_genesis.go#L124

Rinkeby was also set up like this iirc.

Also, I believe that due to a bug in the spurious dragon hardfork, prefunding the precompiles is useful in some weird edge cases with different client / EVM implementations, see https://github.com/ethereum/EIPs/issues/716

op-node/cmd/genesis/cmd.go
```diff
@@ -89,6 +89,54 @@ var Subcommands = cli.Commands{
...
+				if err := os.WriteFile(outfile, file, 0644); err != nil {
+					return err
+				}
+			}
+			return nil
+		},
+	},
+	{
+		Name:  "devnet-l1",
+		Usage: "Initialized a new L2 devnet genesis file",
```
tuxcanfly, line 105:
```suggestion
		Usage: "Initialized a new L1 devnet genesis file",
```

protolambda (approved):
LGTM. One nit about prefunding precompiles. And you may need to wait with this PR until @optimisticben ACKs to get the right timing, since it may break testnet regenesis tooling, and I'm not sure if we're currently only updating nodes with old configs, or doing a full regenesis.

tuxcanfly (approved):
OK.

op-chain-ops/genesis/setters.go
```diff
@@ -147,3 +147,14 @@ func MigrateDepositHashes(hh *hardhat.Hardhat, db vm.StateDB) error {
...
+		addr := common.BytesToAddress([]byte{byte(i)})
+		db.CreateAccount(addr)
+		db.AddBalance(addr, big.NewInt(1))
```
tuxcanfly, line 158: perf nit: move big.Int out of the loop

tuxcanfly (approved):
OK.

op-chain-ops/genesis/setters.go
tynes: good catch

op-chain-ops/genesis/setters.go
tynes: I've updated it to set the balance for the first 256 addresses

op-chain-ops/genesis/setters.go
```diff
@@ -152,9 +152,9 @@ func MigrateDepositHashes(hh *hardhat.Hardhat, db vm.StateDB) error {
 // This is an optimization to make calling them cheaper. This should only
 // be used for devnets.
 func SetPrecompileBalances(db vm.StateDB) {
-	for i := 0; i <= 9; i++ {
+	for i := 0; i <= 256; i++ {
```
protolambda, line 155: First 256 would be `<= 255` or `< 256` since 0 counts. Now we visit 0 twice.

mslipper: Updated.

tynes: Whoops, good catch

mslipper (approved):
Fixed Proto's comments. LGTM.

PR #3299, opened 2022-08-23
op-node: update readme with genesis docs

tynes:
**Description**

Add documentation on how to use the `op-node` to generate
L2 devnet genesis blocks. This will be expanded in the future
to include upgrading pre-bedrock databases.


tuxcanfly (approved):
OK.

protolambda (approved):

PR #3286, opened 2022-08-22
indexer: upgrade bridge contracts to bedrock

tuxcanfly:
**Description**
This PR upgrades the contract bindings being used to index L1 / L2 chains to the bedrock contracts.

**Additional Context**
Also removes dead artifacts related to l2geth and state commitment chain.

**Metadata**
- Fixes ENG-2665

PR #3300, opened 2022-08-23
bindings: add L1StandardBridge bindings

tuxcanfly:
**Description**
This PR adds L1StandardBridge bindings to `op-bindings` package.

**Additional context**
Will be used in #3286


tynes (approved):

tynes:
I think we are getting close to a point where we are going to need to optimize the CI task, I've tried before using `make -j $(nproc)` but it failed due to the lock on the compiler downloader breaking things, using that command locally worked for me, then my workaround of downloading the compiler beforehand didn't work so idk what the deal was exactly, but if we could run these build steps in parallel or use an alternative approach to building the go bindings that is faster it would be more ideal.

`foundry` is open to having building go bindings natively built in, since they support building rust bindings, we would just need to implement that which is something that I tried to do once but never got around to finishing

protolambda (approved):

PR #3301, opened 2022-08-23
contracts: add inspect hh task

tynes:
**Description**

This hardhat task mirrors functionality of `forge inspect`
where you can pass a contract name and a compiler output field
and it will print that field for you. This will be useful for
building sets of `storageLayout`s and `methodIdentifiers` to
compare against the bedrock contracts.

Example usage:

```
$ npx hardhat inspect AddressDictator storageLayout
```

Where `AddressDictator` can be any contract in the package
and `storageLayout` can be one of the following:

- abi
- bytecode
- deployedBytecode
- storageLayout
- methodIdentifiers

This is one of the more useful features of foundry so porting it
to hardhat will make storageLayout inspections much easier.


tynes:
I don't love the linter settings on this PR

smartcontracts:
Lol, this is useful enough that it made me consider if we should have  a tasks package for "generally useful hardhat tasks"

maurelian (approved):

PR #3302, opened 2022-08-23
predeploys: add L2 `ProxyAdmin`

tynes:
**Description**

We need an instance of the `ProxyAdmin` on L2 to manage each
of the other predeploys. This change ranges across the entire
codebase and the spec is updated with the new predeploy.


tynes:
The markdown linting is failing due to a quicknode webpage being down

op-chain-ops/genesis/config.go
```diff
@@ -158,5 +158,9 @@ func NewStorageConfig(hh *hardhat.Hardhat, config *DeployConfig, chain ethereum.
 		// TODO: this should be set to the MintManager
 		"_owner": common.Address{},
 	}
+	storage["ProxyAdmin"] = state.StorageValues{
+		// TODO: this needs to be configurable
```
smartcontracts, line 162: Do we want to set this config up now? Or is it fine if that's done in a different PR?

op-chain-ops/genesis/setters.go
```diff
@@ -29,23 +29,22 @@ func SetProxies(hh *hardhat.Hardhat, db vm.StateDB) error {
 	if err != nil {
 		return err
 	}
-	proxyAdmin, err := hh.GetDeployment("ProxyAdmin")
-	if err != nil {
-		return err
-	}
+	proxyAdminSlot := predeploys.ProxyAdminAddr.Hash()
```
smartcontracts, line 32: Nit: `proxyAdminSlotValue` might be more legible. Generally I assume the slot is the key and not the value which made this confusing to read at first.

maurelian: Nit on the nit:
`proxyAdminSlotValue` feels overload. Maybe:

- `proxyAdminValue`
- `proxyAdminPadded`

op-chain-ops/genesis/setters.go
```diff
@@ -29,23 +29,22 @@ func SetProxies(hh *hardhat.Hardhat, db vm.StateDB) error {
...
 
-	for i := uint64(0); i <= 2048; i++ {
+	for i := uint64(0); i < 2048; i++ {
```
maurelian, line 34: Where does 2048 come from?

packages/contracts-bedrock/src/constants.ts
```diff
@@ -19,6 +19,7 @@ export const predeploys = {
   LegacyERC20ETH: '0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000',
   WETH9: '0x4200000000000000000000000000000000000006',
   GovernanceToken: '0x4200000000000000000000000000000000000042',
+  ProxyAdmin: '0x4200000000000000000000000000000000000016',
```
maurelian, line 22: Comment only: it sucks that this has to be written 3 times. I have not solution for it though.

specs/predeploys.md
```diff
@@ -378,3 +380,9 @@ interface L1Block {
     ) external;
 }
 ```
+
+## ProxyAdmin
+
+The `ProxyAdmin` is the owner of all of the proxy contracts set at the
+predeploys. It is not behind a proxy itself. The owner of the `ProxyAdmin` will
+have the ability to upgrade any of the other predeploy contracts.
```
maurelian, line 388: Suggestions: 
1. Move this to the top. 
2. Add a section about upgradability, saying that all predeploys (with some exceptions) are actually an instance of our Proxy implementation, which has this interface: 

```
    /**
     * @notice An event that is emitted each time the implementation is changed. This event is part
     *         of the EIP-1967 specification.
     *
     * @param implementation The address of the implementation contract
     */
    event Upgraded(address indexed implementation);

    /**
     * @notice An event that is emitted each time the owner is upgraded. This event is part of the
     *         EIP-1967 specification.
     *
     * @param previousAdmin The previous owner of the contract
     * @param newAdmin      The new owner of the contract
     */
    event AdminChanged(address previousAdmin, address newAdmin);


    /**
     * @notice Sets the initial admin during contract deployment. Admin address is stored at the
     *         EIP-1967 admin storage slot so that accidental storage collision with the
     *         implementation is not possible.
     *
     * @param _admin Address of the initial contract admin. Admin as the ability to access the
     *               transparent proxy interface.
     */
    constructor(address _admin) {
        _changeAdmin(_admin);
    }

    /**
     * @notice Default functions which forward calls to the proxy.
     */
    receive() external payable;
    fallback() external payable;

    /**
     * @notice Set the implementation contract address. The code at the given address will execute
     *         when this contract is called.
     *
     * @param _implementation Address of the implementation contract.
     */
    function upgradeTo(address _implementation) external;

    /**
     * @notice Set the implementation and call a function in a single transaction. Useful to ensure
     *         atomic execution of initialization-based upgrades.
     *
     * @param _implementation Address of the implementation contract.
     * @param _data           Calldata to delegatecall the new implementation with.
     */
    function upgradeToAndCall(address _implementation, bytes calldata _data)
        external
        payable
        proxyCallIfNotAdmin
        returns (bytes memory)
    {
        _setImplementation(_implementation);
        (bool success, bytes memory returndata) = _implementation.delegatecall(_data);
        require(success, "Proxy: delegatecall to new implementation contract failed");
        return returndata;
    }

    /**
     * @notice Changes the owner of the proxy contract. Only callable by the owner.
     *
     * @param _admin New owner of the proxy contract.
     */
    function changeAdmin(address _admin) external proxyCallIfNotAdmin {
        _changeAdmin(_admin);
    }

    /**
     * @notice Gets the owner of the proxy contract.
     *
     * @return Owner address.
     */
    function admin() external proxyCallIfNotAdmin returns (address);

    /**
     * @notice Queries the implementation address.
     *
     * @return Implementation address.
     */
    function implementation() external proxyCallIfNotAdmin returns (address);

    /**
     * @notice Sets the implementation address.
     *
     * @param _implementation New implementation address.
     */
    function _setImplementation(address _implementation) internal;

    /**
     * @notice Changes the owner of the proxy contract.
     *
     * @param _admin New owner of the proxy contract.
     */
    function _changeAdmin(address _admin) internal;

    /**
     * @notice Performs the proxy call via a delegatecall.
     */
    function _doProxyCall() internal;

    /**
     * @notice Queries the implementation address.
     *
     * @return Implementation address.
     */
    function _getImplementation() internal view returns (address);

    /**
     * @notice Queries the owner of the proxy contract.
     *
     * @return Owner address.
     */
    function _getAdmin() internal view returns (address);
```

tynes: I want to clean up this entire page and delete all interfaces and instead have links to the contracts themselves so that we don't need to worry about the interfaces getting out of date

packages/contracts-bedrock/src/constants.ts
tynes: Yeah... solidity, go and js

op-chain-ops/genesis/setters.go
tynes: Random number that @mslipper chose that is large enough to allow for many predeploys but small enough to make the genesis file not be massive

op-chain-ops/genesis/config.go
tynes: I was going to do this in a different PR, was thinking of adding the following config values:

- `l1ProxyAdminOwner`
- `l2ProxyAdminOwner`

Then those values would be passed through to here

op-chain-ops/genesis/config.go
```diff
@@ -174,5 +181,8 @@ func NewL2StorageConfig(config *DeployConfig, block *types.Block, proxyL1Standar
 		// TODO: this should be set to the MintManager
 		"_owner": common.Address{},
 	}
+	storage["ProxyAdmin"] = state.StorageValues{
+		"owner": l2Addrs.ProxyAdmin,
```
smartcontracts, line 185: Is this making the owner of the proxy admin be itself? Shouldn't the owner be `ProxyAdminOwner`?

tynes: I can update the variable naming

tynes: I've updated the variable name, it was always the owner

packages/contracts-bedrock/contracts/libraries/Predeploys.sol
```diff
@@ -78,4 +78,9 @@ library Predeploys {
      *         L2ToL1MessagePasser contract instead.
      */
     address internal constant LEGACY_MESSAGE_PASSER = 0x4200000000000000000000000000000000000000;
+
+    /**
+     * @notice Address of the ProxyAdmin predeploy.
+     */
+    address internal constant PROXY_ADMIN = 0x4200000000000000000000000000000000000017;
```
maurelian, line 85: Should this be a predeploy? Will we also make the ProxyAdmin upgradeable?

Could we just deploy it normally, even before state surgery, and then set the owner of the predeploys to that address?

tynes: We could do that, we talked about this awhile ago on office hours and agreed to this approach but then it never got merged. I don't really care which approach we take.

The proxy admin predeploy is not upgradable, along with the legacy eth and gov token

maurelian: Right. OK, so worst case of making it a predeploy is that we might replace it eventually and then we just have a legacy predeploy floating around that doesn't do much. 

Not ideal, but also not worth a bunch of rework.

op-chain-ops/genesis/config.go
```diff
@@ -127,6 +132,9 @@ func NewL2StorageConfig(config *DeployConfig, block *types.Block, proxyL1Standar
 	if block.BaseFee() == nil {
 		return storage, errors.New("block base fee not set")
 	}
+	if l2Addrs == nil {
+		return storage, errors.New("must pass L1 address info")
```
trianglesphere, line 136:
```suggestion
		return storage, errors.New("must pass L2 address info")
```
Typo?

op-chain-ops/genesis/config.go
```diff
@@ -103,22 +104,26 @@ func NewDeployConfigWithNetwork(network, path string) (*DeployConfig, error) {
 
 // NewL2ImmutableConfig will create an ImmutableConfig given an instance of a
 // Hardhat and a DeployConfig.
-func NewL2ImmutableConfig(config *DeployConfig, block *types.Block, proxyL1StandardBridge common.Address, proxyL1CrossDomainMessenger common.Address) (immutables.ImmutableConfig, error) {
+func NewL2ImmutableConfig(config *DeployConfig, block *types.Block, l2Addrs *L2Addresses) (immutables.ImmutableConfig, error) {
 	immutable := make(immutables.ImmutableConfig)
 
+	if l2Addrs == nil {
+		return immutable, errors.New("must pass L1 contract addresses")
```
trianglesphere, line 111: ditto

trianglesphere (approved):

mslipper (approved):

PR #3303, opened 2022-08-24
op-chain-ops: add helper function

tynes:
**Description**

Deduplicates some code for a common pattern.
Callers will usually pass a network and a path
to a deploy config directory. This comes from
the way that `hardhat-deploy-config` works.
The config directory must include a JSON file
that is named after the network.


protolambda (approved):

tuxcanfly (approved):
OK.

PR #3304, opened 2022-08-24
op-chain-ops: Fix config marshal/unmarshal

mslipper:
Hive specifies the entire deployment config - even default values. This exposed a bug where the deployment config JSON specified here was not properly handling bigints/uint64s specified as hex as per Geth's genesis.json. This PR adds support for those values, as well as a test using a real Hive deploy config to assert that everything works.


op-chain-ops/genesis/config.go
```diff
@@ -18,50 +19,247 @@ import (
...
+	L2GenesisBlockGasLimit           hexutil.Uint64  `json:"l2GenesisBlockGasLimit"`
+	L2GenesisBlockDifficulty         string          `json:"l2GenesisBlockDifficulty"`
+	L2GenesisBlockMixHash            common.Hash     `json:"l2GenesisBlockMixHash"`
+	L2GenesisBlockCoinbase           common.Address  `json:"l2GenesisBlockCoinbase"`
+	L2GenesisBlockNumber             hexutil.Uint64  `json:"l2GenesisBlockNumber"`
+	L2GenesisBlockGasUsed            hexutil.Uint64  `json:"l2GenesisBlockGasUsed"`
+	L2GenesisBlockParentHash         common.Hash     `json:"l2GenesisBlockParentHash"`
+	L2GenesisBlockBaseFeePerGas      string          `json:"l2GenesisBlockBaseFeePerGas"`
+	L1GenesisBlockTimestamp          hexutil.Uint64  `json:"l1GenesisBlockTimestamp"`
+	L1GenesisBlockNonce              hexutil.Uint64  `json:"l1GenesisBlockNonce"`
```
tynes, line 105: It's a lot of fields to maintain 2 separate structs, I wouldn't be opposed to using hexutil types in the main struct if it lets us only maintain 1 struct. What do you think?

tynes:
Also is there anything we have to do to get semgrep to run on the chain ops package? I looked for config and didn't see anything

op-chain-ops/genesis/config.go
mslipper: IMO I'd rather have a bunch of additional fields than leak details about how the struct is serialized. We either maintain lots of fields or force callers to type cast/call unnecessary methods everywhere this struct is used, and I prefer the former.

The nice thing about the test is that it'll complain if the structs fall out of sync.

mslipper:
It is running on the chain ops package - see above. I enabled our Go linting tooling, which caught some additional issues which I fixed in this PR.

mslipper:
@tynes I'm going to revert the deleted test artifacts and put them back in the `op-chain-ops/testdata` package. We can't rely on artifacts existing in the `packages/contracts-bedrock` directory because CI won't necessarily run the contracts-bedrock build. I don't want to introduce a dependency on that build since it'd increase CI runtime by ~5 minutes and require a monorepo build for all Go PRs.

tynes:
I'm not sure what you mean by revert, but I'm cool with copying a set of artifacts into the testdata and then removing the monorepo build prereq

mslipper:
I seem to remember seeing some test artifacts in here that later got deleted - my bad if I'm misremembering.

tynes:
We should be sure to update the testdata artifacts once in awhile so that we can be sure nothing changes in the contracts such that it breaks the pipeline. There are certain solidity types that the state package won't work with, it's unlikely we will hit that problem, itll also break in linked libraries, we don't use any of those

mslipper:
Roger. Hmm, looks like the Go code doesn't support defaults within the deploy config either - will implement those.

tynes:
> I seem to remember seeing some test artifacts in here that later got deleted - my bad if I'm misremembering.

It was part of the rename PR, there isn't a commit to revert that'll fix it, it happened bc they were targeted by gitignore and I had to manually readd them with `git add -f` and that was overlooked originally. I believe the dependency on the monorepo build came elsewhere

mslipper:
OK, I looked into adding the artifacts into the `testdata` directory. The genesis tests require the entire contract state to be there, including build info and dependencies. This is a >150MB tree, so to get around that problem I included a tarball of the deployment artifacts as of this evening. The tarball gets untarred into a temporary directory at test time.

op-chain-ops/genesis/testdata/test-deploy-config-full.json
```diff
@@ -0,0 +1,47 @@
...
+  "l1GenesisBlockNonce": "0x0",
+  "cliqueSignerAddress": "0x3c44cdddb6a900fa2b585dd299e03d12fa4293bc",
+  "l1GenesisBlockGasLimit": "0xe4e1c0",
+  "l1GenesisBlockDifficulty": "0x1",
+  "l1GenesisBlockMixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "l1GenesisBlockCoinbase": "0x0000000000000000000000000000000000000000",
+  "l1GenesisBlockNumber": "0x0",
+  "l1GenesisBlockGasUsed": "0x0",
+  "l1GenesisBlockParentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "l1GenesisBlockBaseFeePerGas": "0x3b9aca00",
```
tynes, line 28: I think a more simple solution would be just use numbers instead of hex strings. The go code can handle them and they will end up as the correct type by the time the `genesis.json` is serialized. If you want me to implement this in the hh tasks first to show that it can work, I am happy to do so

tynes:
> OK, I looked into adding the artifacts into the `testdata` directory. The genesis tests require the entire contract state to be there, including build info and dependencies. This is a >150MB tree, so to get around that problem I included a tarball of the deployment artifacts as of this evening. It's still large (12MB), but much smaller than the alternative. The tarball gets untarred into a temporary directory at test time.

Are you sure that we would need to build the monorepo for each time any go tests run? I thought they would only need to run if the chain-ops package tests needed to run

op-chain-ops/genesis/testdata/test-deploy-config-full.json
mslipper: Done - though some of these are bignums, and need to stay hex. The gas values are an example of this.

op-chain-ops/solc/types.go
```diff
@@ -55,7 +55,7 @@ type StorageLayoutEntry struct {
 	Label    string `json:"label"`
 	Offset   uint   `json:"offset"`
 	Slot     uint   `json:"slot,string"`
-	Type     string `json"type"`
+	Type     string `json:"type"`
```
tuxcanfly, line 58: Nice catch.

tuxcanfly (approved):
OK.

op-chain-ops/genesis/layer_two_test.go
```diff
@@ -77,3 +88,45 @@ func TestBuildOptimismGenesis(t *testing.T) {
...
+				return err
+			}
+			continue
+		}
+
+		file, err := os.OpenFile(path, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, info.Mode())
+		if err != nil {
+			return err
+		}
+		defer file.Close()
```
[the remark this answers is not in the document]
mslipper: Nice catch. Fixed.

op-chain-ops/genesis/config.go
```diff
@@ -18,50 +19,260 @@ import (
...
+	L2GenesisBlockExtraData          hexutil.Bytes          `json:"l2GenesisBlockExtraData"`
+	L2GenesisBlockGasLimit           uint64                 `json:"l2GenesisBlockGasLimit"`
+	L2GenesisBlockDifficulty         *big.Int               `json:"l2GenesisBlockDifficulty"`
+	L2GenesisBlockMixHash            common.Hash            `json:"l2GenesisBlockMixHash"`
+	L2GenesisBlockCoinbase           common.Address         `json:"l2GenesisBlockCoinbase"`
+	L2GenesisBlockNumber             uint64                 `json:"l2GenesisBlockNumber"`
+	L2GenesisBlockGasUsed            uint64                 `json:"l2GenesisBlockGasUsed"`
+	L2GenesisBlockParentHash         common.Hash            `json:"l2GenesisBlockParentHash"`
+	L2GenesisBlockBaseFeePerGas      *big.Int               `json:"l2GenesisBlockBaseFeePerGas"`
+	L1GenesisBlockTimestamp          uint64                 `json:"l1GenesisBlockTimestamp"`
```
protolambda, line 56: I'm ok-ish with this as-is, but it maybe less verbose if we use `hexutil.Uint64` and `hexutil.Big` as field types.

And if we are using `deployConfigJSON`, then the fields in this type don't need json annotations. Either one struct + typed fields, or the custom function for encoding.
Maybe we just copy over the Hive definition of the deploy config to this package. And then import that into Hive. So we only need 1 version of the deploy config Go type.

PR #3305, opened 2022-08-24
contracts-bedrock: simplify `GasPriceOracle`

tynes:
**Description**

Remove the ability to set the decimals field of the scalar.
This is a simplification of the `GasPriceOracle` in preparation
for making it `CrossDomainOwnable`. It is currently set to 6 decimals
which is plenty of precision. We have never changed this value
and removing the ability to change it will make the L1 portion of
updating the gas config on L2 much more simple.

To verify that the value hardcoded in this PR is correct:

```bash
$ cast call \
    --rpc-url https://mainnet.optimism.io \
    0x420000000000000000000000000000000000000F \
    'decimals()(uint256)'
6
```

The decimals are read from state on L2 during the state transition
function, this is a backwards compatible change because the same
number is still set in storage and its still available via a public
function. If we really need to set this value we could always upgrade
to a `GasPriceOracle` contract with a setter and then set the value
but there is no world where I see that ever being useful.


packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
```diff
@@ -42,7 +42,7 @@ contract GasPriceOracle is Ownable, Semver {
     /**
      * @notice Number of decimals used in the scalar.
      */
-    uint256 public decimals;
+    uint256 public decimals = 6;
```
smartcontracts, line 45: Might as well make this a `constant`

tynes: If we make it a constant, will it be in the bytecode or in the state?

tynes: It must be in the state for this to be backwards compatible

tynes: It appears as if making it a constant does not place it in state, so we cannot do this unfortunately

maurelian: what is the backwards compatibility issue?

tynes: The decimals are read from state in the L2 geth state transition function

maurelian: as is written, I believe that this syntax depends on running initcode to set the value...

tynes: To clarify, the same scheme is use in the bedrock optigeth as old l2geth

maurelian (approved):

packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
smartcontracts: Got it. Let's leave it in state then.

smartcontracts (approved):

PR #3306, opened 2022-08-24
Add op-exporter feature to collect baseFee

optimisticben:
Added collection of the `baseFee` field of a `eth_getBlockByNumber` response for bedrock deployments.

This feature is behind the flag `--gaseBaseFee.enable`.

Rollup price collection no longer applies to bedrock deployments, so it now requires the flag `--rollUpGasPrices.enable`.

Closes ENG-2604


protolambda (approved):
Would be nice to exit the polling loop more gracefully, but it does work and is consistent with the other op exporter functions, so LGTM.

mslipper (approved):

PR #3309, opened 2022-08-24
contracts-bedrock: CrossDomainOwnable

tynes:
**Description**

Adds 2 new contracts to the `L2` directory. Both contracts
extend OZ `Ownable` such that the `onlyOwner` modifer logic is
overwritten.

`CrossDomainOwnable` is designed for ownership of contracts on L2
when the communication goes directly through the `OptimismPortal`.
This is not safe for when the communication goes through the
`CrossDomainMessenger` system, as `msg.sender` would be the
address of the `L2CrossDomainMessenger` and if that is the owner
of the L2 contract then anybody going through the messenger system
would have owner permissions on the contract.

`CrossDomainOwnable2` is designed for ownership of contracts on L2
when the communication goes through the messenger system. It checks
for the L1 message sender value that is set in the
`L2CrossDomainMessenger` contract.

Both of these contracts include test coverage and are relatively simple.


packages/contracts-bedrock/contracts/L2/CrossDomainOwnable2.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Predeploys } from "../libraries/Predeploys.sol";
+import { L2CrossDomainMessenger } from "./L2CrossDomainMessenger.sol";
+import { AddressAliasHelper } from "../vendor/AddressAliasHelper.sol";
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+
+/**
```
smartcontracts, line 9: Why do we need this second version? `xDomainMessageSender` is not aliased.

tynes: Good catch, I fixed this issue

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable2.sol
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * @notice Overrides the implementation of the `onlyOwner` modifier
+     *         to check that the unaliased `xDomainMessageSender`
+     *         is the owner of the contract. This value is set
+     *         to the caller of the L1CrossDomainMessenger.
+     */
+    function _checkOwner() internal view override {
+        require(
+            owner() == messenger.xDomainMessageSender(),
```
smartcontracts, line 28: Need to check `msg.sender == messenger` as well

smartcontracts (approved):

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable2.sol
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * @title CrossDomainOwnable2
+ * @notice This contract extends the OpenZeppelin `Ownable` contract
+ *         for L2 contracts to be owned by contracts on L1. Note that
+ *         this contract is meant to be used with systems that use the
+ *         CrossDomainMessenger system. It will not work if the OptimismPortal
+ *         is used directly.
+ */
+abstract contract CrossDomainOwnable2 is Ownable {
+    L2CrossDomainMessenger internal messenger =
```
maurelian, line 17: Should this be immutable? There's no affordance made for modifying it, and the address is unlikely to change.

tynes: I'll make it a constant, good call

packages/contracts-bedrock/contracts/test/CrossDomainOwnable.t.sol
```diff
@@ -0,0 +1,107 @@
...
+        );
+
+        // from is indexed and the first argument to the event.
+        bytes32 _from = log.topics[1];
+        address from = Bytes32AddressLib.fromLast20Bytes(_from);
+
+        assertEq(
+              AddressAliasHelper.undoL1ToL2Alias(from),
+              alice
+        );
```
maurelian, line 94: Nice. This could be useful elsewhere. I've thought a fair bit about how there's not much reason not to test both components from both L1 and L2 on one chain.

maurelian (approved):

PR #3319, opened 2022-08-26
op-node: handle special engine RPC user-input error codes

protolambda:
**Description**

Testnet bugfix:
* The testnet was started without the parent-hash checking (#3221 )
* Reorgs happened (on L1 Goerli testnet, 20+ block deep, well past confirmation distance)
* Frame from data tx was missed again (twice actually). Due to valid reasons: it was just too old, from pre-reorg multiple minutes ago, likely bad L1 data in there too. 
* Batches *after* that frame were still derived, and not dropped, and insertion into engine was attempted
* Nonce errors were hit, but this time we hit a different error path: we received back an RPC error code that we were not properly handling. So the nice payload-attributes validity error we got was assumed to be temporary.
* And then it got stuck retrying those attributes over and over again. At least we have a backoff, but it essentially got stuck due to bad error handling

This PR:
- Adds the missing error type codes that we need to handle. See https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md
- Transforms the RPC error into an `InputError` so we can safely wrap it without losing the ability to distinguish errors
- Handles forkchoice-update errors pre-block-insertion as a reason to reset to regain consistency, not a reason to drop sequencer transactions to try to force forward when the forkchoice is incoherent already.
- Handles input-errors: transactions can be bad, and the payload processing status may not be reached if something is wrong about the input before EVM execution. Nonce values are one example of this.

Depends on #3318 for lint issue in CI.

**Metadata**

Fix ENG-2695

### DIFF
--- a/op-node/eth/types.go
+++ b/op-node/eth/types.go
@@ -20,8 +20,32 @@ import (
 type ErrorCode int
 
 const (
-	UnavailablePayload ErrorCode = -32001
+	UnknownPayload           ErrorCode = -32001 // Payload does not exist / is not available.
+	InvalidForkchoiceState   ErrorCode = -38002 // Forkchoice state is invalid / inconsistent.
+	InvalidPayloadAttributes ErrorCode = -38003 // Payload attributes are invalid / inconsistent.
 )
+
+// InputError distinguishes an user-input error from regular rpc errors,
+// to help the (Engine) API user divert from accidental input mistakes.
+type InputError struct {
+	Inner error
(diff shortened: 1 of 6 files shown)


mslipper (approved):

tuxcanfly (approved):
OK.

protolambda:
Will rebase to include the linting fix on develop.

protolambda:
And fixed go-imports issue

PR #3320, opened 2022-08-26
op-node: fix sync heads starting point [bedrock]

protolambda:
**Description**

This PR fixes the way we find the L2 block that we start syncing on top of.

#### Off by one fix in safe head

As I was working on tests for the finalization, I found that the safe-head sync algorithm has an off-by-one error:
* We reset back a full sequence window, but stop at the L2 block that has a L1 origin exactly n blocks from the common point, inclusive. This means that the safe head we're left with after completing the reset, is the one block with sequence nr 0, within the scope of the window. Which can point to an older L1 origin, but where the inclusion is within the window.
* Meanwhile, if this exact batch wasn't included anymore after a deep reorg, or e.g. if the L1 chain we reorged to has no batches at all for a full window, then an alternative (or empty L2 block) would be there.
We need to reset one L2 block deeper, to force all previous safe L1 blocks (only those that we can't be sure of with inclusion even though their L1 origin is canonical) to be reconciled against L1, to ensure they are still included on the new canonical L1 chain.

#### Safe block RPC label usage

Currently we're also just doing the above starting at the unsafe head instead of safe head, meaning the off-by-one error can be higher if we happen to have a long unsafe chain that still references canonical L1 data but isn't actually submitted on-chain.
This is a long-due fix for a workaround that couldn't be fixed due to the lack of support for the `safe` block label in the geth RPC.

#### Finalized block check

The initialization of the finalized info is now part of the sync start algorithm, since we need to check if we don't revert it accidentally.

**Metadata**

Fix ENG-2671


mslipper:
This has been replaced.

PR #3322, opened 2022-08-26
fix(ct): bug in L2OutputOracle constructor

smartcontracts:
**Description**
Fixes a small bug in the L2OutputOracle constructor where an assertion
was being made about the wrong value. Unlikely that this would've caused
any significant issues but worth fixing anyway.


maurelian (approved):

maurelian:
I approve this, but now that I think about it we really need to have the discipline of writing a test every time we fix a bug. even when it's a trivial one like this. would really appreciate if you'd be so kind as to oblige me on that.

smartcontracts:
SGTM!

smartcontracts:
> I approve this, but now that I think about it we really need to have the discipline of writing a test every time we fix a bug. even when it's a trivial one like this. would really appreciate if you'd be so kind as to oblige me on that.

Done

mslipper (approved):

mslipper:
@smartcontracts mind updating the gas snapshot + bindings?

tynes (approved):

tynes:
This needs code to handle the L1 genesis creation

mslipper:
This is causing a bug somewhere:

```
Failed
=== RUN   TestBuildL1DeveloperGenesis
    layer_one_test.go:32: 
        	Error Trace:	/root/project/op-chain-ops/genesis/layer_one_test.go:32
        	Error:      	Received unexpected error:
        	            	no code found for L2OutputOracle
        	Test:       	TestBuildL1DeveloperGenesis
--- FAIL: TestBuildL1DeveloperGenesis (0.01s)
```

mslipper:
This is breaking tests because the Geth simulated backend that deploys the contracts on L1 in the devnet sets the block timestamp to zero. We can fix this, but it'll require changes in `op-geth`.

PR #3346, opened 2022-09-07
op-node: buffer unsafe payloads with priority queue [bedrock]

protolambda:
**Description**

Pop blocks by lowest number to maintain max size, and do not drop if the first payload is in the future.
This way we keep the buffer around until block derivation from L1 makes progress, and the unsafe payloads can be processed (or dropped if they do not connect with parent hash).

This should enabled nodes to stay on the canonical unsafe L2 chain, the blocks by the sequencer ahead of what's submitted to L1, more reliably.

Testing work in progress.

**Metadata**

Fix ENG-2511


op-node/metrics/metrics.go
```diff
@@ -150,6 +153,17 @@ func NewMetrics(procName string) *Metrics {
 		SequencingErrors: NewEventMetrics(registry, ns, "sequencing_errors", "sequencing errors"),
 		PublishingErrors: NewEventMetrics(registry, ns, "publishing_errors", "p2p publishing errors"),
 
+		UnsafePayloadsBufferLen: promauto.With(registry).NewGauge(prometheus.GaugeOpts{
```
mslipper, line 156: Nice 🎉

mslipper (approved):

PR #3348, opened 2022-09-07
Generate codecoverage and upload to CodeCov

maurelian:
Changes CI to upload code coverage results generated by the go modules and contracts-bedrock package.

This job doesn't run on this PR because of the Check Changed step, but it was tested in previous iterations of this branch by disabling that step. The results can be seen [here](https://app.codecov.io/gh/ethereum-optimism/optimism/tree/temp-getcoverage). 

Improvements which are not included in this PR but will be added later: 
1. Moving the codecov uploader installation to the ci-builder image.
2. Failing in CI when the coverage threshold decreases.
3. Adding a coverage badge to the README.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,6 @@ version: 2.1
 
 orbs:
   go: circleci/go@1.5.0
-
 jobs:
   yarn-monorepo:
     docker:
@@ -33,6 +32,9 @@ jobs:
       - run:
           name: Install dependencies
           command: yarn --frozen-lockfile
+      - run:
+          name: print forge version
+          command: forge --version
       - save_cache:
           key: v2-cache-yarn-install-{{ checksum "yarn.lock" }}
           paths:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -52,11 +54,23 @@ jobs:
       - run:
           name: Build monorepo
           command: yarn build
+      - run:
+          name: Download and verify Codecov uploader
+          command: |
+            curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import # One-time step
```
mslipper, line 60: Let's put the codecov downloader inside `ci-builder` directly. This way we avoid having to download and install it on every build.

mslipper (changes requested):

.circleci/config.yml
maurelian: OK, I've added that in a [new commit](https://github.com/ethereum-optimism/optimism/pull/3348/commits/2b6e5773494eec3c23f9f0e6a74718e86611f6eb). 

To avoid failures caused by running CI before a new release is made, I've also made it so that the upload step [will always pass](https://github.com/ethereum-optimism/optimism/pull/3348/commits/8c902ba18e142f340d700418c0356bbe09fb9796).

mslipper (approved):

PR #3350, opened 2022-09-07
indexer: scan eth/standard bridge for deposits

tuxcanfly:
**Description**
This PR fixes an issue with the bedrock indexer where L1 chain was not being scanned for deposits. The issue was that bridges needed to be instantiated and passed to the indexing service.

**Tests**
This is a hotfix so that the indexer works with bedrock. Tests will be added in a upcoming PR.


protolambda (approved):

indexer/services/l1/bridge/bridge.go
```diff
@@ -25,59 +26,62 @@ type Bridge interface {
...
 	name string
 	impl string
-	addr common.Address
+	addr string
 }
 
-var addrs map[string]common.Address
+var defaultBridgeCfgs = []*implConfig{
+	{"Standard", "StandardBridge", predeploys.DevL1StandardBridge},
+	{"ETH", "ETHBridge", predeploys.DevL1StandardBridge},
```
mslipper, line 34: These are L1 predeploys for the devnet - you can't use them against a real network. Is there any way to set these at runtime/via config?

mslipper (changes requested):

indexer/services/l1/bridge/bridge.go
tuxcanfly: Typically these should be configured _once_ after deployment on the network and mostly never change after that. Not sure if it's worth it to make it configurable. Instead I've updated to map chainID to the bridge addresses like we're doing with custom bridges below.

mslipper (approved):

PR #3365, opened 2022-09-09
op-node: Properly return error on bad hex data

mslipper:


trianglesphere:
The candidate linters are `nilnil` or `nilerr`. Looking into them.

op-node/cmd/p2p/cmd.go
```diff
@@ -16,7 +16,7 @@ import (
 func Priv2PeerID(r io.Reader) (string, error) {
 	b, err := readHexData(r)
 	if err != nil {
-		return "", nil
+		return "", err
 	}
 
 	p, err := crypto.UnmarshalSecp256k1PrivateKey(b)
```
trianglesphere, line 22: Further down in `Pub2PeerID` we have the same issue.

trianglesphere (commented):
`nilerr` is the lint we want to use to find these issues.
It found two instances (only one of which is currently fixed in this PR) in the `cmd/p2p/cmd.go` file.

protolambda (approved):

PR #3367, opened 2022-09-09
op-node,op-proposer: move rollup client bindings

protolambda:
move rollup bindings into op-node to avoid unnecessary op-proposer dependency in tooling / batcher.

- left `rollupclient` with alias type / constructor call for backwards compatibility.
- `SyncStatus` was moved to eth package to avoid import cycles with bindings and driver.


op-node/rollup/driver/state.go
```diff
@@ -16,31 +16,8 @@ import (
...
-	UnsafeL2 eth.L2BlockRef `json:"unsafe_l2"`
-	// SafeL2 points to the L2 block that was derived from the L1 chain.
-	// This point may still reorg if the L1 chain reorgs.
-	SafeL2 eth.L2BlockRef `json:"safe_l2"`
-	// FinalizedL2 points to the L2 block that was derived fully from
-	// finalized L1 information, thus irreversible.
-	FinalizedL2 eth.L2BlockRef `json:"finalized_l2"`
-}
+// Deprecated: use eth.SyncStatus instead.
+type SyncStatus = eth.SyncStatus
```
tynes, line 20: What sort of deprecation policy do we have for things like this? We are pre 1.0.0 and its likely that only we import our packages, so if this is left to keep a smaller diff thats cool but would be nice to delete this eventually

op-node/sources/rollupclient.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"math/big"
+
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+type RollupClient struct {
```
tynes, line 13: Is this meant to be exhaustive in the RPCs it can call?

tynes (approved):

op-node/rollup/driver/state.go
mslipper: I'd like us to do this kind of thing iteratively. There are downstream dependencies of this code that use the tagged Go modules. To avoid breaking those, we should add a deprecated alias like this, upgrade tags, update downstream dependencies to remove the alias, then remove the alias and retag.

mslipper:
Can you update the Make target that updates packages? Or do we not have to update it because we have the alias?

mslipper (changes requested):
LGTM pending response to the requested `Makefile` changes.

mslipper (approved):

op-node/sources/rollupclient.go
protolambda: those are in the p2p package in op node

PR #3370, opened 2022-09-09
fix: simplify erc20 factory

tynes:
**Description**

Updates the `OptimismMintableERC20Factory` to have an L1 and L2 version. This simplifies the system
as the L2 contract can have a hardcoded predeploy value for the bridge and the L1 version can be deployed
with the deployed address of the L1 bridge


packages/contracts-bedrock/contracts/L2/L2OptimismMintableERC20Factory.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { OptimismMintableERC20Factory } from "../universal/OptimismMintableERC20Factory.sol";
+import { Predeploys } from "../libraries/Predeploys.sol";
+
+/**
+ * @custom:proxied
+ * @custom:predeployed 0x4200000000000000000000000000000000000012
+ * @title OptimismMintableERC20Factory
```
smartcontracts, line 10:
```suggestion
 * @title L2OptimismMintableERC20Factory
```

And same below

mslipper:
I updated the bindings - still seeing a bunch of failing tests. Looks like `op-chain-ops` and the `sdk` haven't been completely migrated over to use the new contracts.

mslipper (changes requested):
See above comment.

PR #3373, opened 2022-09-09
feat(indexer): Add wsteth to mainnet indexer

roninjin10:
This branch needs to be deployed as a hot-change canary release.  This is because indexer HEAD is currently built for bedrock.


roninjin10:
Need to do a hot release of this pre bedrock

PR #3377, opened 2022-09-09
fix: M-08 Upgradeability inconsistencies

maurelian:
**Description**

From OZ-08. 
This PR:
1. Makes the use of storage gaps consistent across contracts which are inherited into proxied contracts.
2. Introduces a new `@custom:upgradeable` tag to identify contracts which include a storage gap and are initializable.
3. Removes the storage gap from the Portal. We don't need gaps on concrete contracts which are not mean to be inherited.

Note that Semver.sol does not have a storage gap. I don't think we're likely to ever need storage in that contract.

**Tests**

No tests have been added, but pre-existing tests have been updated.


packages/contracts-bedrock/README.md
```diff
@@ -61,6 +61,8 @@ Some basic rules:
 We also have the following custom tags:
 
 - `@custom:proxied`: Add to a contract whenever it's meant to live behind a proxy.
+- `@custom:upgradeable`: Add to a contract whenever it's meant to be used in an upgradeable contract.
+- `@custom:semver`: Add to a constructor to indicate the version of a contract.
```
maurelian, line 65: This is slightly unrelated, but I noticed it wasn't documented.

tynes (approved):

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -97,6 +98,11 @@ abstract contract CrossDomainMessenger is
      */
     address public otherMessenger;
 
+    /**
+     * @notice Reserve extra slots (to a total of 50) in the storage layout for future upgrades.
+     */
+    uint256[46] private __gap;
```
smartcontracts, line 104: Why not put this gap underneath `blockedSystemAddresses`?

maurelian: ah, good catch.

smartcontracts (approved):

PR #3382, opened 2022-09-09
Safe head fix 2

protolambda:
**Description**

This replaces #3320, preserving the FindHeads function, with the recently discussed requirements:
- [x] Off By one fix: full sequence window, beyond the start
- [x] Prev safe head minimum:
  -  if the L1 origins referenced by unsafe L2 blocks are canonical for a full sequence window, then we can accept the previous safehead without going back further.
  - if the above is not the case, we end up selecting a block before the safe head. And since it is one loop, we can still optimize this like the above
- [x] Finalized head: make sure we never go back further than the finalized head, and that we fall back to genesis if nothing has been finalized yet.
- [x] Genesis checks: make sure the engine and the L1 source are on the correct chain. Never start syncing between different chain data.
- [x] Contiguous L1 origin: when we are not ahead of the L1 chain, we should check the L1 origin parent hashes all match. And always check basic consistency between L1 origins in L2 blocks we traverse.
- [x] Port over tests from original PR. Including adjustments to engine queue finalization PR, since it needs to account for the safe head changes / lavel usage.
- [x] Change function signature: we have rpc labels to go by, and shouldn't rely on any previous derivation state from the rollup node itself.

**Metadata**

Fix ENG-2671


op-node/rollup/sync/start.go
```diff
@@ -40,164 +27,191 @@ import (
...
+type FindHeadsResult struct {
+	Unsafe    eth.L2BlockRef
+	Safe      eth.L2BlockRef
+	Finalized eth.L2BlockRef
+}
+
+func currentHeads(ctx context.Context, cfg *rollup.Config, l2 L2Chain) (*FindHeadsResult, error) {
+	finalized, err := l2.L2BlockRefByLabel(ctx, eth.Finalized)
+	// geth rpc bug: returning an error instead of nil result when finalized block cannot be found.
+	if errors.Is(err, ethereum.NotFound) || (err != nil && strings.Contains(err.Error(), "block not found")) {
```
tynes, line 64: Perhaps this line could be turned into a helper function, its used twice

op-node/rollup/sync/start.go
```diff
@@ -40,164 +27,192 @@ import (
...
-	} else {
-		canonical := canonical.Hash == block.Hash
-		return canonical, canonical, nil
+type FindHeadsResult struct {
+	Unsafe    eth.L2BlockRef
+	Safe      eth.L2BlockRef
+	Finalized eth.L2BlockRef
+}
+
+func currentHeads(ctx context.Context, cfg *rollup.Config, l2 L2Chain) (*FindHeadsResult, error) {
```
trianglesphere, line 61: needs godoc comment

op-node/rollup/sync/start.go
```diff
@@ -40,164 +27,192 @@ import (
...
 			}
+			l1Block = b
+			ahead = notFound
 		}
 
-		// Don't walk past genesis. If we were at the L2 genesis, but could not find its L1 origin,
-		// the L2 chain is building on the wrong L1 branch.
-		if n.Hash == genesis.L2.Hash || n.Number == genesis.L2.Number {
-			return eth.L2BlockRef{}, eth.L2BlockRef{}, WrongChainErr
+		if n.Number == cfg.Genesis.L2.Number {
```
trianglesphere, line 134: why cut the comment?

op-node/rollup/sync/start.go
```diff
@@ -40,164 +27,192 @@ import (
...
 
-		// Found an L2 block whose L1 origin is the start of the sequencing window.
-		// Note: We also ensure that we are on the block number with the 0 seq number.
-		// This is a little hacky, but kinda works. The issue is about where the
-		// batch queue should start building.
-		if depth == seqWindowSize && n.SequenceNumber == 0 {
-			return highestPlausibleCanonicalOrigin, n, nil
+		// If the L2 block is at least as old as the previous safe head, and we have seen at least a full sequence window worth of L1 blocks to confirm
+		// TODO discuss with Josh
+		if n.Number <= result.Safe.Number && n.Number+cfg.SeqWindowSize < highestL2WithCanonicalL1Origin.Number && n.SequenceNumber == 0 {
```
trianglesphere, line 179: seqNumber check almost seems like a return check over finding the safehead check.
I also need to think more if we actually need the full sequence window check here. Maybe, but it's a weird to me way of thinking about it.

op-node/rollup/sync/start_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/common"
```
trianglesphere, line 7: goland imports

trianglesphere (changes requested):
Where is the off by one error & where is it fixed? I'm having a hard time finding it.

op-node/rollup/sync/start.go
protolambda: moved it into the rpc bindings package to share between safe/finalized heads. Also put a copy in L1 however, and keeping that a copy since L1 and L2 nodes may update/fix this at different times.

op-node/rollup/sync/start.go
protolambda: There are new more descriptive comments below within the if statement, and more descriptive errors too.

protolambda:
> Where is the off by one error & where is it fixed? I'm having a hard time finding it.

Commit [56c6093](https://github.com/ethereum-optimism/optimism/pull/3382/commits/56c60931c84ad533478c42dd8607f51f41b5b70e) fixes it.

op-node/rollup/sync/start.go
```diff
@@ -41,163 +28,191 @@ import (
...
```
trianglesphere, line 179: `result.Safe` is the original safe head, right?

protolambda: yes

op-node/rollup/sync/start.go
```diff
@@ -41,163 +28,197 @@ import (
...
-	reorgDepth := 0
+	n := result.Unsafe
 
-	// Blockhash of L1 origin hash for the L2 block during the previous iteration, 0 for first
-	// iteration. When this changes as we walk the L2 chain backwards, it means we're seeing a different
-	// (earlier) epoch.
-	var prevL1OriginHash common.Hash
+	var highestL2WithCanonicalL1Origin eth.L2BlockRef
+	var l1Block eth.L1BlockRef
+	var ahead bool
```
trianglesphere, line 119: Comments on what `ahead` means + what the loop invariants / stages are would be helpful

op-node/rollup/sync/start.go
trianglesphere: Two reasons to keep it: it fits in the style of having a comment over each if stmt block. The second is that the reason we don't return on the case of `n == cfg.Genesis` is that `finalized = cfg.Genesis` in the case that we don't having a finalized block & we do a return on hitting `n == finalized` later on.

op-node/rollup/sync/start.go
```diff
@@ -41,163 +28,197 @@ import (
...
+			// keep the unsafe head if we can't tell if its L1 origin is canonical or not yet.
+		} else if l1Block.Hash == n.L1Origin.Hash {
+			// if L2 matches canonical chain, even if unsafe,
+			// then we can start finding a span of L1 blocks to cover the sequence window,
+			// which may help avoid rewinding the existing safe head unnecessarily.
+			if highestL2WithCanonicalL1Origin == (eth.L2BlockRef{}) {
+				highestL2WithCanonicalL1Origin = n
+			}
+		} else {
+			// L1 origin nor ahead of L1 head nor canonical, discard previous candidate and keep looking.
```
trianglesphere, line 174:
```suggestion
			// L1 origin not ahead of L1 head nor canonical, discard previous candidate and keep looking.
```

op-node/rollup/sync/start.go
```diff
@@ -41,163 +28,197 @@ import (
...
-		return true, false, nil
-	} else if canonical, err := l1.L1BlockRefByNumber(ctx, block.Number); err != nil {
-		return false, false, err
-	} else {
-		canonical := canonical.Hash == block.Hash
-		return canonical, canonical, nil
+type FindHeadsResult struct {
+	Unsafe    eth.L2BlockRef
+	Safe      eth.L2BlockRef
+	Finalized eth.L2BlockRef
```
trianglesphere, line 57: we don't strictly need the `Finalized` head b/c we never actually use it.

op-node/testutils/mock_l1.go
```diff
@@ -12,6 +13,7 @@ type MockL1Source struct {
 }
 
 func (m *MockL1Source) L1BlockRefByLabel(ctx context.Context, label eth.BlockLabel) (eth.L1BlockRef, error) {
+	fmt.Println("L1BlockRefByLabel", label)
```
trianglesphere, line 16: Is this intentional?

op-node/rollup/sync/start.go
protolambda: We do use it, we need to make sure we don't go past it, and that we use the current engine value instead of a past value in rollup node. So retrieving it when we retrieve safe/unsafe works well.

op-node/rollup/sync/start.go
protolambda: nice catch, copied the comment over and didn't see the typo.

op-node/testutils/mock_l1.go
protolambda: it was quite useful for debugging, but I should probably put it behind a logger. Will drop it from this PR.

trianglesphere (approved):
goimports is failing in the mocks

PR #3394, opened 2022-09-10
feat(ctb): make StandardBridge vars immutable

smartcontracts:
**Description**
Updates the StandardBridge to use immutable variables instead of state variables. Primary reason for this change is to guarantee that the storage layout will not be a problem during the upgrade. Plus, there's really no reason why these variables shouldn't be immutable.

**Tests**
Yes, tests have been updated accordingly.


maurelian (approved):

packages/contracts-bedrock/.gitignore
```diff
@@ -7,3 +7,4 @@ coverage.out
 broadcast
 genesis.json
 src/contract-artifacts.ts
+tmp-artifacts
```
tynes, line 10: This can be deleted

tynes (approved):

PR #3403, opened 2022-09-10
op-node/batcher/proposer: Standardize goimports

trianglesphere:
**Description**

This standardizes go imports across the three go codebases. This removes
empty lines between import sections & relies on `goimports` to sort the rest.
Imports from the `ethereum-optimism` and `go-ethereum` repositories are
treated as local imports.

**Additional Information**

I used the following script to implement this change:
```
#!/bin/bash                                                                      
                                                                                 
# remove all blank lines in go 'imports' statements,                             
# then sort with goimports                                                       
# Usage over a codebase: find . -name "*.go" -exec ./goimports2.sh {} \;
                                                                                 
if [ $# != 1 ] ; then                                                            
  echo "usage: $0 <filename>"                                                    
  exit 1                                                                         
fi                                                                               

# Remove empty lines inside the import statement                                                                                 
sed -i '' '/^import/,/^)/ {/^$/d;}' $1                                                                             

#goimports -w $1
#goimports -local "github.com/ethereum-optimism/optimism" -w $1
goimports -local "github.com/ethereum/go-ethereum","github.com/ethereum-optimism/optimism" -w $1
```

### DIFF
--- a/op-batcher/batch_submitter.go
+++ b/op-batcher/batch_submitter.go
@@ -21,7 +21,8 @@ import (
 	opmetrics "github.com/ethereum-optimism/optimism/op-service/metrics"
 	oppprof "github.com/ethereum-optimism/optimism/op-service/pprof"
 	oprpc "github.com/ethereum-optimism/optimism/op-service/rpc"
-	"github.com/ethereum/go-ethereum/rpc"
+	hdwallet "github.com/miguelmota/go-ethereum-hdwallet"
+	"github.com/urfave/cli"
 
 	"github.com/ethereum-optimism/optimism/op-batcher/sequencer"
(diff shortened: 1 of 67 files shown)


protolambda:
LGTM, can you put the script in CI? Like the other diff checks, we can run it, and then check if there's a diff to confirm the contents are formatted correctly.

trianglesphere:
> can you put the script in CI? Like the other diff checks, we can run it, and then check if there's a diff to confirm the contents are formatted correctly.

There's two parts of this: killing the newlines between the different import sections & doing goimports. We already have goimports, but this mainly does does the killing newlines. Unfortunately the sed script is not reliable enough when there is a single import line to run.

protolambda (approved):

tynes (approved):

PR #3404, opened 2022-09-10
feat(ctb): remove blockedSystemAddresses

smartcontracts:
**Description**
Removes the blockedSystemAddresses in favor of a more flexible approach that also removes most of the need for the initializer. We will be following this up with a second PR that removes the initializer entirely.

**Tests**
No tests required, same logic as before is being tested.

### DIFF
--- a/packages/contracts-bedrock/hardhat.config.ts
+++ b/packages/contracts-bedrock/hardhat.config.ts
@@ -45,6 +45,11 @@ const config: HardhatUserConfig = {
       url: process.env.L1_RPC || '',
       accounts: [process.env.PRIVATE_KEY_DEPLOYER || ethers.constants.HashZero],
     },
+    'alpha-1': {
+      chainId: 5,
+      url: process.env.L1_RPC || '',
+      accounts: [process.env.PRIVATE_KEY_DEPLOYER || ethers.constants.HashZero],
+    },
     deployer: {
       chainId: Number(process.env.CHAIN_ID),
       url: process.env.L1_RPC || '',


packages/contracts-bedrock/tasks/genesis-l2.ts
```diff
@@ -248,9 +248,7 @@ task('genesis-l2', 'create a genesis config')
         _owner: deployConfig.l2CrossDomainMessengerOwner,
         xDomainMsgSender: '0x000000000000000000000000000000000000dEaD',
         msgNonce: 0,
-        otherMessenger: ProxyL1CrossDomainMessenger.address,
-        // TODO: handle blockedSystemAddresses mapping
-        // blockedSystemAddresses: [{key: '', value: ''}],
+        otherMessenger: ProxyL1CrossDomainMessenger.address
```
tynes, line 251: This file was deleted in a PR that was merged

tynes:
I like this approach because it makes things cheaper and also resolves issues with storage slots since it moves a variable to being an immutable, we no longer need to worry about a storage slot collision.

@smartcontracts We should set the storage slot that is no longer being used to 0 as part of the L1 upgrade

smartcontracts:
> @smartcontracts We should set the storage slot that is no longer being used to 0 as part of the L1 upgrade

`blockedSystemAddresses` was a new storage slot, so removing this would not require adding a spacer.

maurelian (approved):

maurelian:
This looks safe to me. 
Regarding storage slot spacing etc, while these contracts are continuing to evolve I would propose that we continue pursuing the ideal design, and only worry about fixing any storage slot issues after the design is frozen.

PR #3422, opened 2022-09-11
op-node: limit reorg depth with sanity check based on L1 origin and sequence window size

protolambda:
Avoids conflicts between seq window config param and the previous constant max depth


trianglesphere (approved):

mslipper (approved):

mslipper:
Looks like there's some failing unit tests.

PR #3430, opened 2022-09-12
indexer: filter withdrawals by finalization

tuxcanfly:
**Description**

This PR updates the `/withdrawals` endpoint to filter withdrawals by finalization status as requested by the frontend team.

**Tests**

There will be a follow up PR with tests.

**Additional context**

This change affects the frontend API. The backed will only return ready to finalize withdrawals by default. Finalized withdrawals can be query separately by using the `?finalized=true` query string. cc @roninjin10 

**Metadata**
- Fixes #ENG-2670

### DIFF
--- a/indexer/db/db.go
+++ b/indexer/db/db.go
@@ -450,7 +450,7 @@ func (d *Database) GetWithdrawalStatus(hash common.Hash) (*WithdrawalJSON, error
 
 // GetWithdrawalsByAddress returns the list of Withdrawals indexed for the given
 // address paginated by the given params.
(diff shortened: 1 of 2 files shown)


roninjin10:
Instead of filtering them out by default can we have following behavior?

- By default don't filter
- if finalized=true is passed then we fitler for only finalized
- if finalized=false is passed in than we filter for non-finalized

tynes:
Would it be easy to add some tests for this functionality?

tuxcanfly:
Going to rework to address feedback from @roninjin10

PR #3431, opened 2022-09-12
feat(ctb): remove blockedSystemAddresses

smartcontracts:
**Description**
Removes the blockedSystemAddresses in favor of a more flexible approach that also removes most of the need for the initializer. We will be following this up with a second PR that removes the initializer entirely.

I broke my last PR


maurelian (approved):
Love this change.

tynes (approved):

PR #3432, opened 2022-09-12
fix(ct): patch build error

smartcontracts:
**Description**
Fixes a build error where files were being referenced that were not included in the NPM build. CI did not catch this because CI operates on what's in the monorepo, which included the relevant files.

**Tests**
We need to figure out a way to test stuff like this, but my concern is that we would need to significantly update the CI pipeline to do so. I have not added any tests for this. I will create a canary release and test it directly that way.


packages/contracts/scripts/generate-artifacts.ts
```diff
@@ -15,6 +15,15 @@ const main = async () => {
 
   const artifactPaths = glob
     .sync(`${contractArtifactsFolder}/**/*.json`)
+    .filter((match) => {
+      return [
+        'artifacts/contracts/chugsplash',
```
smartcontracts, line 20: These are now the only artifact folders contained in the NPM build. Please reference `package.json` to confirm this.

tynes: Looking here: https://github.com/ethereum-optimism/optimism/blob/59e41374800dedec872bd4ef4211c057c8badd29/packages/contracts/package.json#L11

Appears to match

tynes (approved):

maurelian (approved):

packages/contracts/scripts/generate-artifacts.ts
maurelian: resolving to get on the merge train

PR #3436, opened 2022-09-12
indexer: verify flags with chainid instead of hash

tuxcanfly:
**Description**

This PR updates the indexer flags and removes the need to pass L1 / L2 block hashes to start indexing.

This simplifies ops for the indexer.


tuxcanfly:
Even though this is ready, this would be a good candidate for using mock rpc server for testing, so holding off to add that.

PR #3438, opened 2022-09-12
feat(ctb): Make spacers private

maurelian:
**Description**

Spacer variables are changed from `internal` to `private` so that they cannot be accessed in child contracts.

**Tests**

No tests have been added. I don't think it worth adding more test code to maintain in this case.


tynes:
It looks like a diff in the metadata hash in weth9, we set that off in the `foundry.toml` so not sure why that is happening. Its as if the config is not being applied correctly or something

smartcontracts (approved):

PR #3441, opened 2022-09-13
op-bindings: make deterministic _more.go files

tynes:
**Description**

The astids are non deterministic for some reason
in our toolchain but the storage layouts are always the same. This adds some code to remove the nondeterminism but we lose the ability to work with astids. This is fine for the way that we are using the storage layouts.


mslipper:
Didn't you mention that the AST IDs were referenced by other fields in the struct?

maurelian (approved):
LGTM

maurelian:
Actually, I think we can achieve the same thing just be deleting this line: 
https://github.com/ethereum-optimism/optimism/blob/develop/op-bindings/solc/types.go#L53

tynes:
> Actually, I think we can achieve the same thing just be deleting this line: https://github.com/ethereum-optimism/optimism/blob/develop/op-bindings/solc/types.go#L53

The problem is that the ast id is appended to the contract type, which is why there is extra code that modifies the type struct. I have a preference of not deleting the ast id from the struct if possible since that would limit other usecases of this code

tynes:
> Didn't you mention that the AST IDs were referenced by other fields in the struct?

This code handles it:
```go
for key, value := range storage.Types {
```

Issue #3372, opened 2022-09-09
Typo after abi formatting using interface.format function


Kharabet:
**Describe the bug**
I was trying to get l1BaseFee using `OVMGasPriceOracle` contract. I have initialized it using the following code:

```
  const OVMGasPriceOracle = getContractFactory('OVM_GasPriceOracle').attach(
    predeploys.OVM_GasPriceOracle
  )
  const formatted = OVMGasPriceOracle.interface.format(ethers.utils.FormatTypes.json) as string
  const abi = JSON.parse(formatted)
  return new web3.eth.Contract(abi, OVMGasPriceOracle.address)
```

In `formatted`  there is a typo in abi structure: `ouput` instead of `output` which doesn't allow to parse the result of the method call

![image](https://user-images.githubusercontent.com/14793248/189368201-7e33f767-e157-4ff0-a8f6-f125eade8dee.png)

**Additional context**
The issue comes actually from ethers.js formatter and they have fixed it here https://github.com/ethers-io/ethers.js/issues/1275 in version 5.0.29 

In package json I see ethers.js v 5.0.0 as a dependency. Upgrading it to v 5.0.29 would resolve the issue

smartcontracts:
The `@eth-optimism/contracts` package has `ethers@^5` as a peer dependency, so the version of ethers will depend on the version you have installed in your project. If you install the latest version of ethers within your project, this should work.

PR #3444, opened 2022-09-13
deps: bump ethers, remove old oz deps

tynes:
**Description**

Bumps ethers to the latest version and removes unused openzeppelin deps that were in the root of the project. There was no need to have those older versions of openzeppelin around, I do not believe they were being used at all but we should be sure that the exact version of openzeppelin is compiled into the contracts.

Closes https://github.com/ethereum-optimism/optimism/issues/3372


tynes:
New versions of ethers break smock, cc @smartcontracts

packages/contracts/package.json
```diff
@@ -115,6 +115,6 @@
     "typescript": "^4.6.2"
   },
   "peerDependencies": {
-    "ethers": "^5"
+    "ethers": "^5.7.0"
```
smartcontracts, line 118: Needs to be ^5

smartcontracts:
> New versions of ethers break smock, cc @smartcontracts

Need to update smock to 2.3.0 everywhere, will fix this

tynes:
@smartcontracts will take over this PR

tynes:
Replaced by https://github.com/ethereum-optimism/optimism/pull/3500

PR #3445, opened 2022-09-13
fix(dtl): optimization for mainnet

smartcontracts:
**Description**
Optimizes the DTL by making the assumption that CTC and SCC addresses in the AddressManager will NOT change before Bedrock rolls around. This finally makes it possible to properly sync off of a Geth node where events are not indexed. You can still sync from Erigon as an alternative to this commit.

**Tests**
Ah... the DTL. You know how it goes. I'll sync a node to test this.


Inphi:
Does this resolve #2653?

smartcontracts:
I am now syncing a node with this change

smartcontracts:
Confirmed that this syncs mainnet properly

tynes (approved):

mslipper (approved):

PR #3446, opened 2022-09-13
feat(sdk): Add token list to sdk

roninjin10:
This adds cloning the token list as a build step.  

**Why**
- We want to use token list as source of truth instead of duplicating things like token address constants and bridge adapters https://docs.google.com/document/d/1oGoXwKIh3YD3BdviT3OuHILG0iZSDRTB31eut22Z1WI/edit?usp=sharing
- Versioning the token list makes our build more hermetic on the frontend.   We previously have had the UI break out from under us.

I held off on updating the adapters for now because that is blocked by this pr adding all the existing adapters https://github.com/ethereum-optimism/ethereum-optimism.github.io/pull/186


nickbalestra:
What are the benefits of fetching it over adding it as a git submodule during the build step?

roninjin10:
@nickbalestra 

1. More convenient.  For this the token list is always up to date without having to change a hash
2. @smartcontracts would not approve the pr if I used git submodules because he thinks they are gross
3. I don't like them either

The downside is the build still isn't hermetic with this change.   We can make it hermetic though later by publishing the token list as an npm package directly and having it as a sub dependency here.

packages/sdk/src/utils/token-list-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+    optimismBridgeAddress?: string
+    ens?: string
+  }
+}
+
+export const fetchTokenList = async ({
+  uri = OPTIMISM_TOKENLIST_URI,
+  chainId,
+  filterTerm,
+}: Options = {}): Promise<TokenList> => {
```
nickbalestra, line 34: Just to make sure I am not missing something, what is the reason for having those options if are not needed/used in this PR?

packages/sdk/scripts/generateTokenList.ts
```diff
@@ -0,0 +1,24 @@
+import fs, { rmdirSync } from 'fs'
+import path from 'path'
+
+import { fetchTokenList } from '../src/utils/token-list-utils'
+
+const generatedDirectory = path.join(__dirname, '..', 'src', 'generated')
```
nickbalestra, line 6: should this be in `/misc-constants` ?

nickbalestra (changes requested):
I would suggest to avoid adding options we don't need yet (see comment) 
Curious on what made you go with fetching vs adding git module during build step

packages/sdk/src/utils/token-list-utils.ts
roninjin10: They are going to be used in the gateway that's where this came from.  I'm exporting this as a common util for anybody to use 
image

packages/sdk/scripts/generateTokenList.ts
roninjin10: misc-constants imports it from src/generated

nickbalestra (approved):

nickbalestra:
Thanks for the clarifications @roninjin10! (Btw, I'm pretty aligned with @smartcontracts on this :) - Approved

packages/sdk/package.json
```diff
@@ -9,8 +9,9 @@
   ],
   "scripts": {
     "all": "yarn clean && yarn build && yarn test && yarn lint:fix && yarn lint",
+    "prebuild": "npx ts-node scripts/generateTokenList.ts",
```
roninjin10, line 12: The indexer in future will use the standard token list to find custom bridges it needs to index

packages/sdk/package.json
```diff
@@ -9,8 +9,9 @@
...
     "build": "tsc -p tsconfig.json",
-    "clean": "rimraf dist/ ./tsconfig.tsbuildinfo",
+    "clean": "rimraf dist/ ./tsconfig.tsbuildinfo && rimraf generated/",
```
tynes, line 14: I don't think you need to call `rimraf` twice here

packages/sdk/src/utils/misc-constants.ts
```diff
@@ -0,0 +1,2 @@
+export const OPTIMISM_TOKENLIST_URI =
+  'https://static.optimism.io/optimism.tokenlist.json' as const
```
tynes, line 2: I don't love that there are dots in the filename here

roninjin10:
Going to change this to make sdk not hardcode any tokens.   Instead it will take tokens as a constructor arg

packages/sdk/package.json
roninjin10: Good catch

PR #3449, opened 2022-09-13
fix(ctb): N-03 Assert for impossible condition

maurelian:
**Description**

Uses assert rather than a require statements to check for conditions we believe are unreachable.

This is more semantically explicit, and should enable us to more effectively use some advanced analysis methods in our testing.

IMO it's also more readable.

**Tests**

We don't have a good harness in place for verifying that code is unreachable. Making use of Solidity's SMTChecker would be good to do eventually.


tynes:
100% on getting the SMT checker going. Do you know of anything better than the solidity docs for an intro on getting going with it?

maurelian:
This looks interesting, though not directly useful without installing dapp: https://fv.ethereum.org/2021/12/01/smtchecker-dapptools/

For what little I've used it in the past I've just referred to the official docs.

smartcontracts (approved):

tynes:
This looks like it needs a rebase before merge, but the code looks good to me

tynes (approved):

PR #3450, opened 2022-09-13
fix: Emit DepositFailed in L2 Bridge

maurelian:
**Description**

The `L2StandardBridge` contract did not emit a `DepositFailed` event when a
`finalizeDeposit` call fails and instead always emitted a `DepositFinalized` event.

This was fixed by adding a boolean return value to `StandardBridge.finalizeBridgeERC20`,  and emitting the correct event based on the return value.

The L1 bridge is not affected by this issue because it did not previously have a failure case which could occur in the minting of the deposited token. 

**Tests**

Existing test cases have been extended to validate that the proper events are emitted.


packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol
```diff
@@ -151,10 +151,22 @@ contract L2StandardBridge is StandardBridge, Semver {
...
+            bool finalized = finalizeBridgeERC20(
+                _l2Token,
+                _l1Token,
+                _from,
+                _to,
+                _amount,
+                _extraData
+            );
+            if (finalized) {
+                emit DepositFinalized(_l1Token, _l2Token, _from, _to, _amount, _extraData);
```
tynes, line 165: The abstract contract emits a failed event so I'm hesitant with accepting this change. Curious what the thinking is here

maurelian: This particular event is emitted by the existing L2 bridge. It's for backwards compatibility.

smartcontracts (approved):

tynes:
Will approve after snapshot and bindings are rebuilt

tynes (approved):

PR #3462, opened 2022-09-14
op-chain-ops: cross domain message

tynes:
**Description**

Implement encoding and hashing of `CrossDomainMessage` structs. This includes some tests for both v0 and v1 that were generated using solidity.

This is required for the migration of withdrawal hashes.


op-chain-ops/crossdomain/encoding.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var (
+	// NonceMask represents a mask used to extract version bytes from the nonce
+	NonceMask, _ = new(big.Int).SetString("0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", 16)
+	// relayMessage0ABI represents the v0 relay message encoding
+	relayMessage0ABI = "[{\"inputs\":[{\"internalType\":\"address\",\"name\":\"_target\",\"type\":\"address\"},{\"internalType\":\"address\",\"name\":\"_sender\",\"type\":\"address\"},{\"internalType\":\"bytes\",\"name\":\"_message\",\"type\":\"bytes\"},{\"internalType\":\"uint256\",\"name\":\"_messageNonce\",\"type\":\"uint256\"}],\"name\":\"relayMessage\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"}]"
```
tuxcanfly, line 15: Not a blocker but maybe ABIs can be a part of `op-bindings` so we don't have to rewrite them next time they're reqiured.

op-chain-ops/crossdomain/encoding.go
```diff
@@ -0,0 +1,72 @@
...
+	// relayMessage1 represents the ABI of relay message v1
+	relayMessage1 abi.ABI
+)
+
+// Create the required ABIs
+func init() {
+	var err error
+	relayMessage0, err = abi.JSON(strings.NewReader(relayMessage0ABI))
+	if err != nil {
+		panic(err)
```
tuxcanfly, line 29: Normally not a fan of panic, but here it will only fail if the ABI is wrong, so I'm OK.

op-chain-ops/crossdomain/hashing.go
```diff
@@ -0,0 +1,42 @@
...
+	sender *common.Address,
+	data []byte,
+	nonce *big.Int,
+) (common.Hash, error) {
+	encoded, err := EncodeCrossDomainMessageV0(target, sender, data, nonce)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	hash := crypto.Keccak256(encoded)
+	return common.BytesToHash(hash), nil
```
tuxcanfly, line 23: Could skip allocating `hash` but it's just 256 bytes, so not worth it.

tuxcanfly (approved):
OK.

op-chain-ops/crossdomain/message_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+		require.Equal(t, uint64(1), msg.Version())
+
+		hash, err := msg.Hash()
+		require.Nil(t, err)
+
+		expect := common.HexToHash("0x09bbda7f59cdaccab5c41cab4600bd458b2bd7d9f8410f13316fe07e5f4237cc")
+		require.Equal(t, expect, hash)
+	})
+}
```
maurelian, line 100: What do you think about writing a test in op-e2e that compares the output of the go with that of the solidity?
I suppose we'd need to write a simple mock contract to expose the libraries in Encoding.sol and Hashing.sol

tynes: If we want to do that kind of testing, I would prefer to do with via forge ffi so that we can have coverage between solidity, js and go

maurelian (approved):

PR #3466, opened 2022-09-14
docs(fd): minor readme update

smartcontracts:
**Description**
Minor README update after questions from a third party.

**Tests**
README change, not tests required.


tynes (approved):

packages/fault-detector/README.md
```diff
@@ -20,3 +20,40 @@ Once your environment variables have been set, run the service via:
...
 ```
+
+## What this service does
+
+The `fault-detector` detects differences between the transaction results generated by your local Optimism node and the transaction results actually published to Ethereum.
+Currently, transaction results take the form of [the root of the Optimism state trie](https://medium.com/@eiki1212/ethereum-state-trie-architecture-explained-a30237009d4e).
+For every Optimism block, the state root of the block is published to the [`StateCommitmentChain`](https://github.com/ethereum-optimism/optimism/blob/39b7262cc3ffd78cd314341b8512b2683c1d9af7/packages/contracts/contracts/L1/rollup/StateCommitmentChain.sol) contract on Ethereum.
+We can therefore detect differences by, for each block, checking the state root of the given block as reported by an Optimism node and the state root as published to Ethereum.
+
+We export a series of Prometheus metrics that you can use to trigger alerting when issues are detected.
```
tynes, line 31: Perhaps listing the default port of the metrics server could be useful

optimisticben: Something like metrics are exposed at `FAULT_DETECTOR__HOSTNAME `:`FAULT_DETECTOR__PORT`/metrics

optimisticben (approved):
lgtm

packages/fault-detector/README.md
smartcontracts: Done

PR #3467, opened 2022-09-14
fix(ctb): use seaport style for gov token

smartcontracts:
**Description**
Updates GovernanceToken to use seaport style comments. Minor fix, but might as well be consistent across the codebase.

**Tests**
Not changing any logic. No tests required.


tynes (approved):
Sorry for missing that

maurelian (approved):

PR #3468, opened 2022-09-14
maint(ctb): clean up config check task

smartcontracts:
**Description**
Cleans up the config check task so it's slightly easier to maintain.


tynes:
We prob want to make this exhaustive eventually so that we can run this against a network while its staging

tynes (approved):

maurelian (approved):

PR #3469, opened 2022-09-14
op-chain-ops: implement withdrawal hashing

tynes:
**Description**

Implement both the new hashing scheme and the legacy hashing for withdrawals so that the withdrawals can be migrated.
This only includes tests for legacy style hashing, will follow up with an additional PR to add tests for bedrock style withdrawal hashing to prevent this PR from growing too large.

Closes ENG-2718


tynes:
Can get test vectors from here: https://optimistic.etherscan.io/address/0x4200000000000000000000000000000000000000#internaltx

op-chain-ops/crossdomain/withdrawal.go
```diff
@@ -0,0 +1,112 @@
...
+		{Name: "data", Type: BytesType},
+	}
+	enc, err := args.Pack(w.Nonce, w.Sender, w.Target, w.Value, w.GasLimit, w.Data)
+	if err != nil {
+		return nil, err
+	}
+	return enc, nil
+}
+
+// EncodeLegacy will serialze the Withdrawal in the legacy format so that it
```
tynes, line 63: Perhaps this should be `LegacyEncode` to be the same as the other legacy functions (where legacy is the prefix) or the others should be updated such that legacy is the suffix

trianglesphere (commented):
LGTM - one Q about some test setup. I really like how it compares the computed slot vs what get set in the trace.

mslipper:
Recusing myself from review here; others will have more insightful feedback.

tynes:
> LGTM - one Q about some test setup. I really like how it compares the computed slot vs what get set in the trace.

Good call on catching that error, I have updated it to panic on error of abi deserialization

tynes:
I've added some additional functionality to this PR:
- Split the withdrawal type into `LegacyWithdrawal` and `Withdrawal`
- encoding + decoding with roundtrip fuzz tests of both withdrawal types
- Test coverage of the storage slot generation for the `Withdrawal` type
  - Test vectors generated using foundry, this is important for the storage slot migration

op-chain-ops/crossdomain/message.go
```diff
@@ -67,6 +68,26 @@ func (c *CrossDomainMessage) Hash() (common.Hash, error) {
 	case 1:
 		return HashCrossDomainMessageV1(c.Nonce, c.Sender, c.Target, c.Value, c.GasLimit, c.Data)
 	default:
-		return common.Hash{}, fmt.Errorf("unknown nonce version %d", version)
+		return common.Hash{}, fmt.Errorf("unknown version %d", version)
+	}
+}
+
+// ToWithdrawal will turn a CrossDomainMessage into a Withdrawal.
```
smartcontracts, line 75: I'm still a little unclear about what this function is supposed to do

tynes: V0 cross domain messages have all of the information required to create the corresponding withdrawal type, so this is a helper to get access to the the methods on the withdrawal given the same data in the cross domain message

smartcontracts: Oh ok I see, thanks for the clarification!

smartcontracts (approved):

op-chain-ops/crossdomain/legacy_withdrawal.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	out := make([]byte, len(enc)+len(predeploys.L2CrossDomainMessengerAddr.Bytes()))
+	copy(out, enc)
+	copy(out[len(enc):], predeploys.L2CrossDomainMessengerAddr.Bytes())
+	return out, nil
+}
+
+// Decode will decode a serialized LegacyWithdrawal
+func (w *LegacyWithdrawal) Decode(data []byte) error {
```
trianglesphere, line 52: I'm not sure how critical this is given that this isn't exposed to much hostile input, but this will panic if `data` is not long enough.

op-chain-ops/crossdomain/withdrawal.go
```diff
@@ -0,0 +1,116 @@
...
+		{Name: "gasLimit", Type: Uint256Type},
+		{Name: "data", Type: BytesType},
+	}
+	decoded, err := args.Unpack(data)
+	if err != nil {
+		return err
+	}
+
+	sender, ok := decoded[1].(common.Address)
+	if !ok {
```
trianglesphere, line 74: Why are sender/target done with the `ok` syntax (no panics), but the rest are done a in a manner that could panic (if the type was not correct in the result of `args.Unpack`).

op-chain-ops/crossdomain/legacy_withdrawal.go
tynes: Ah good catch, might as well add checks for that

op-chain-ops/crossdomain/withdrawal.go
trianglesphere: Ok. I agree that this is very unlikely to cause a problem, I was curious why there was a difference.

tynes:
> 

i have addressed the issues, thanks for the review

op-chain-ops/crossdomain/legacy_withdrawal.go
```diff
@@ -0,0 +1,132 @@
...
+	if len(data) < 4 {
```
trianglesphere, line 53: The length check should be that is greater than 4 + len(address). This is because on line 71 we slice from `4` to `len(data) - len(address)`. We have only guaranteed that `len(data) >= len(address) >= 4`. That means that if `len(data) == len(address)`, slicing from `4:0` on line 71 would panic.

trianglesphere (changes requested):
There's a small bug in the length check code. Fix that and it's good to go.

op-chain-ops/crossdomain/legacy_withdrawal.go
tynes: I've addressed this in https://github.com/ethereum-optimism/optimism/commit/dd274000bcbdc3325f11e84918d6381ced10786c

trianglesphere (approved):

tynes:
Not sure why this is failing in CI but not locally for me

mslipper:
It fails for me locally as well.

trianglesphere:
@tynes is there an issue with the files checked in / something around the path where it's not able to load in the test cases?

tynes:
> @tynes is there an issue with the files checked in / something around the path where it's not able to load in the test cases?

Found the source of the bug, it was due to the way that the L2ToL1MessagePasser is now being handled. Instead of migrating the state at `address(0x42..00)` we are introducing a new predeploy at `address(0x42..16)` that will receive the migrated hashes and the proof on L1 will work against that account.

PR #3470, opened 2022-09-14
feat(ctb): standardize spacers

smartcontracts:
**Description**
Cleans up spacers to all be private, introduces types to the custom spacer natspec, and renames spacer variables to include slot, offset, and length so we can automate checks.


packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -24,10 +24,10 @@ import { Encoding } from "../libraries/Encoding.sol";
 contract CrossDomainMessengerLegacySpacer {
     /**
      * @custom:legacy
-     * @custom:spacer libAddressManager
+     * @custom:spacer address libAddressManager
      * @notice Spacer for backwards compatibility.
      */
-    address internal spacer0;
```
tynes, line 30: How exactly does the naming scheme work? Perhaps add to readme

smartcontracts: Added to README

maurelian (approved):
We still need to write a harness to test that the location of these spaces in the layout matches the name, but this LGTM for now.

tynes (approved):

PR #3471, opened 2022-09-14
proxyd: Support per-RPC rate limits

mslipper:
Adds support for custom rate limits on a per-RPC method basis. We need this in order to limit abuse of expensive requests like `eth_getLogs`.


proxyd/server.go
```diff
@@ -197,22 +213,37 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
...
+		if lim == nil {
+			return false
 		}
+
+		_, _, _, ok, _ := lim.Take(ctx, xff)
+		return !ok
 	}
-	if !ok {
+
+	if isLimited("") {
```
tynes, line 246: Is the empty string here correct?

tynes: This is saying that if the non method override is limited correct?

proxyd/server.go
```diff
@@ -347,6 +378,16 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isB
 			continue
 		}
 
+		// Take rate limit for specific methods.
+		// NOTE: eventually, this should apply to all batch requests. However,
+		// since we don't have data right now on the size of each batch, we
+		// only apply this to the methods that have an additional rate limit.
```
Inphi, line 384: Should it matter if rate limiting isn't applied to batch requests that don't contain method overrides? I could be wrong, but isn't the global rate limiter meant only for downstream requests? Which we already do in `handleRPC`.

proxyd/server.go
mslipper: Yes - this selects the global limiter.

Inphi:
excellent!

Inphi (approved):

tynes (approved):

PR #3477, opened 2022-09-14
maint(ct): no full ethers in entrypoint

smartcontracts:
**Description**
Updates the contracts package so that the files imported by the entrypoint do not use the full ethers package and only use submodules. Useful because this allows people to only import specific submodules of ethers and not have to bundle the entire ethers package.

**Tests**
Should just work with same tests as before.


tynes (approved):

tynes:
Why is this tagged with `do-not-merge`?

smartcontracts:
> Why is this tagged with `do-not-merge`?

I guess it doesn't have to be

PR #3483, opened 2022-09-15
fix(sdk): Make source maps work for sdk

roninjin10:
Source maps don't currently work because they expect src to be included
image

fix: include src


smartcontracts (approved):
Makes sense

roninjin10:
Alternatively we could make the sourcemaps inline source maps

tynes (approved):

PR #3486, opened 2022-09-15
fix(fd): bug for short FPWs

smartcontracts:
**Description**
Fixes a bug in the fault detector that would cause errors when running on testnets with very short fault proof windows. Now handles this case explicitly.


optimisticben (approved):
lgtm

tynes (approved):
This makes sense

PR #3491, opened 2022-09-15
nft: L-13 and M-08

sam-goldman:
L-13: Refactors a lot of the functionality in the NFT bridges into a single base contract.

M-08: Fixes upgradeability inconsistencies in the ERC721 bridge contracts.

The third commit replaces `CrossDomainEnabled` with `CrossDomainMessenger` in the ERC721 bridge contracts because `CrossDomainEnabled` is not upgradeable.

The fourth commit makes `otherBridge` and `messenger` in the NFT bridge contracts immutable as requested [here](https://github.com/ethereum-optimism/optimism/pull/3459).

PR #3493, opened 2022-09-15
nft: N-08 Easily bypass-able requirement

sam-goldman:


smartcontracts (approved):

tynes:
It feels like we could dedupe this by putting it into the base contract natspec and then using `@inheritdoc`

tynes (approved):

maurelian (approved):

sam-goldman:
> It feels like we could dedupe this by putting it into the base contract natspec and then using `@inheritdoc`

Did that in #3491

PR #3501, opened 2022-09-17
EIP-4844: experimental support [ethberlin hackathon]

protolambda:
Not pretty code yet, since the Blobs API from beacon-node is very limited, but it works (tm)

### DIFF
--- a/op-node/rollup/derive/blob_data_source.go
+++ b/op-node/rollup/derive/blob_data_source.go
@@ -1,0 +1,119 @@
+package derive
+
+import (
+	"context"
+	"fmt"
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type BlobsSidecar struct {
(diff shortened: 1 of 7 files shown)


op-node/rollup/derive/blob_data_source.go
```diff
@@ -0,0 +1,119 @@
...
+			expectedHashes = append(expectedHashes, tx.BlobVersionedHashes()...)
+		}
+	}
+	if len(computedHashes) != len(expectedHashes) {
+		return nil, fmt.Errorf("got %d data hashes for beacon block %s (%d), but expected %d hashes from blob txs in execution block %s",
+			len(computedHashes), sidecar.BeaconBlockRoot, sidecar.BeaconBlockSlot, len(expectedHashes), id)
+	}
+	for i, h := range computedHashes {
+		if h != expectedHashes[i] {
+			return nil, fmt.Errorf("data hash mismatch %d (beacon block %s, slot %d): hash %d does not match expected hash from el block %s: %s",
```
rootulp, line 73: [question] does "el" in this context mean execution layer? If so, proposal to capitalize (EL) or replace with the more explicit 

```suggestion
			return nil, fmt.Errorf("data hash mismatch %d (beacon block %s, slot %d): hash %d does not match expected hash from execution block %s: %s",
```

motivation: conforms to the error message above:
```go
...  hashes from blob txs in execution block %s"
```

PR #3502, opened 2022-09-17
op-node: beacon source

tynes:


op-node/node/node.go
```diff
@@ -160,6 +164,16 @@ func (n *OpNode) initL2(ctx context.Context, cfg *Config, snapshotLog log.Logger
 	return nil
 }
 
+func (n *OpNode) initBeacon(ctx context.Context, cfg *Config) error {
+	var err error
+	n.beaconSource, err = sources.NewBeaconClient(cfg.BeaconAddress)
```
tynes, line 169: TODO: parse + assign the new config flag

tynes:
@protolambda what do you think of merging this into the `eip4844` branch?

op-node/sources/beacon_client.go
```diff
@@ -0,0 +1,48 @@
...
+
+// "/eth/v1/blobs/sidecar/{block_id}"
+func (cl *BeaconClient) FetchSidecar(ctx context.Context, slot uint64) (*derive.BlobsSidecar, error) {
+	url := url.JoinPath(cl.BeaconAddress, "/eth/v1/blobs/sidecar/", slot)
+	res, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	body, err := io.ReadAll(res.Body)
+	res.Body.Close()
```
protolambda, line 36: A sidecar can be up to 2MB large, 4MB of hexified json strings. It may be better to feed the body into the json decoder directly. It's fine like this though.

protolambda (approved):

op-node/sources/beacon_client.go
tynes: Pushed a commit that changes this

PR #3506, opened 2022-09-19
Remove the progress API from the derivation pipeline

trianglesphere:
**Description**

This is a stack of PRs to incrementally refactor the derivation pipeline with the goal of removing the progress API. This does this by recognizing that the progress API is used for two different things: 1) Recognizing when the a stage will not get any more data from a lower level stage and 2) Tracking where the stage is. There is a third function of advancing the L1 origin of the pipeline by a complex relationship between the progress and step APIs.

The new API is as follows:
- Each stage retains a reference to the previous stage
- Each stage asks the previous stage for more data. If a stage is out of data, it returns io.EOF. If a stage may produce more data but is currently out, it returns derive.NotEnoughData.
- Once the final stage (the engine queue) returns io.EOF, the pipeline advances the L1 traversal stage
- There is a single origin for the entire pipeline. It is set the the origin required of the channel bank. As such, the batch queue needs to reject some batches that if pulls from earlier in the pipeline.

The goal of this refactor (removing the progress API) + defining a new API between stages is to make it easier to test each stage in isolation as well as make it easier to incrementally improve each stage in the future.

**Tests**

Every commit in this sequence passes the op-e2e tests, however no effort was made to make the unit tests pass.

**Additional context**

Each commit in this stack is meant to be able to be reviewed as it's own PR. I've combined them together to show what each PR is driving towards.

### DIFF
--- a/op-node/rollup/derive/calldata_source.go
+++ b/op-node/rollup/derive/calldata_source.go
@@ -2,7 +2,6 @@ package derive
 
 import (
 	"context"
-	"fmt"
 	"io"
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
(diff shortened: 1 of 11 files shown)


op-node/rollup/derive/calldata_source.go
```diff
@@ -15,21 +14,60 @@ import (
...
+	cfg     *rollup.Config // TODO: `DataFromEVMTransactions` should probably not take the full config
+	fetcher L1TransactionFetcher
+	log     log.Logger
+}
 
-func (ds *DataSlice) Next(ctx context.Context) (eth.Data, error) {
-	if len(*ds) == 0 {
+func NewCalldataSourceImpl(ctx context.Context, log log.Logger, cfg *rollup.Config, fetcher L1TransactionFetcher, block eth.BlockID) *CalldataSourceImpl {
+	_, txs, err := fetcher.InfoAndTxsByHash(ctx, block.Hash)
+	if err != nil {
```
tynes, line 39: Couldn't this get stuck in an infinite loop if the block hash doesn't exist? Perhaps want to check for `ethereum.NotFound`

trianglesphere: Yea, this can. This PR serves as an example of the structure rather than being ready to go, but this specific code will need to be beefed up before being merge.

protolambda, line 39: There should be a difference between temporary RPC errors and persistent not-found type of errors. Creating a calldata source with no data (if it doesn't error, but just returns nothing?) may result in accidental missed data, which could become a divergence. IMHO leaving the calldata as-is, and putting it in a separate PR or reducing changes is nice, since it's not quite the same as a derivation stage.

protolambda (commented):
Looks pretty good! I like how the explicit L1 traversal means the `io.EOF` of stages can now be regarded as having fully read the current origin, instead of immediately traversing further. And it looks much more like an io.Reader interface, which is nice.

I'm more hesitant on the `Reset` logic: making sure we reproduce the same buffer-state in each stage is very important to keep nodes that started syncing at different points actually in sync. E.g. the channel bank may emit bad data while it hasn't fully rebuilt its own buffer, and the next stages may be polluted with that inconsistent data. E.g. a batch may have been encoded in a data tx that would have been pruned if the channel bank was full, instead of the artificial empty buffer. That batch should not get buffered/processed, but dropped. Keeping track of origins per stage aimed to avoid these cross-stage interactions, dropping inconsistent data as soon as possible, while a stage is getting consistent again after resetting it back.

It would help if we can write tests for the resetting, or include in the specs how we make sure we can sync from any older point by using the reset + sequence window properties. Let me know if/how I can help with that.

And I don't want to scope-creep this work, but it would be nice to discuss the possible DoS vectors, maybe in a separate PR: making sure that e.g. the ChannelBank can never grow too large is difficult, but maybe less so if we think about that in this refactor.

trianglesphere:
> I'm more hesitant on the Reset logic: making sure we reproduce the same buffer-state in each stage is very important to keep nodes that started syncing at different points actually in sync

The only stage that actually prunes data is the channel bank. For that stage it is important that the order of operations (insert, remove, prune) remains the same and that we start far enough back that we end up in the same state as if we had been going for the start.

The other important stage is the batch queue. Here we probably want to be careful to not have extra batches, but my suspicion is that it's more ok to put a mix of batches in here because of the batch pruning rules.

trianglesphere:
> it would be nice to discuss the possible DoS vectors

What new DoS vectors are you worried about here?

trianglesphere:
Closing this PR to do this incrementally. The first PR is https://github.com/ethereum-optimism/optimism/pull/3532

PR #3512, opened 2022-09-20
ctb: Emit a refunded event in StandardBridge

maurelian:
**Description**

This follows the lead of a change made to the ERC721 bridge contracts so that when a refund is created:
1. the `_from` address is set to the current Bridge (ie. the bridge which initiates the return transfer)
2. a new `ERC20BridgeRefunded` event is emitted on the other Bridge when the refund is finalized.

**Tests**

I've added a new test to both L1 and L2 bridges. This also required updating the expected event arguments in other existing tests.

Although it does add some new branching logic, this change also makes the contracts more testable, by allowing us to identify when a `finalizeBridgeX` call is a refund.


packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -344,18 +362,34 @@ abstract contract StandardBridge {
...
+            } else {
+                emit ERC20BridgeFinalized(
+                    _localToken,
+                    _remoteToken,
+                    _from,
+                    _to,
+                    _amount,
+                    _extraData
+                );
+                return true;
```
maurelian, line 380: it would be better to move these duplicate return statements to the end of the if/else block

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -344,18 +362,34 @@ abstract contract StandardBridge {
...
+            }
         } catch {
             // Something went wrong during the bridging process, return to sender.
             // Can happen if a bridge UI specifies the wrong L2 token.
-            // We reverse the to and from addresses to make sure the tokens are returned to the
-            // sender on the other chain and preserve the accuracy of accounting based on emitted
+            // Since the _to address never received the L2 token, we set this bridge contract
```
smartcontracts, line 385: I'm a little confused by this comment. The original flow makes the accounting look like:

(person A bridged X amount of token Y to person B from chain 1 to chain 2) 
(person B bridged X amount of token Y to person A from chain 2 to chain 1)

Which seems very accurate. I guess the difference here is the explicit acknowledgement that the user did NOT actually transfer the funds back and instead there was a refund because of an error on the other chain.

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
smartcontracts: Yes plz

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -344,18 +362,34 @@ abstract contract StandardBridge {
...
+            // as the sender address for the refund below, and set the recipient as the _from
+            // address. This preserves the accuracy of accounting based on emitted
             // events.
             _initiateBridgeERC20Unchecked(
```
smartcontracts, line 389: What happens if the return-to-sender fails for some reason? Now both `_from` and `_to` are set to `address(this)` and I guess the funds just get stuck forever?

smartcontracts: At least with the original design you get a loop that can be resolved by fixing one side or the other

smartcontracts: We could change the interface of `finalizeBridgeERC20` to include a flag `isRefund` instead of doing this

smartcontracts: And instead of adding a new event we could just emit the flag as part of the `ERC20BridgeFinalized` and `ERC20BridgeFailed` events, which are both new events and don't need to be backwards-compatible

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
maurelian: Correct. It's not as thought Person B actually sent anything.

maurelian:
@smartcontracts even prior to reading your comments, and moreso now, I think we just don't need this new event.
I do like setting `from` on the events to the bridge which initiates the refund, and that logic is easy to change, but I really don't like adding a new if/else into the already fairly complex `finalizeBridgeERC20` flow. 
And I think we should also remove the new event from the ERC721 Bridge.

smartcontracts:
> @smartcontracts even prior to reading your comments, and moreso now, I think we just don't need this new event. I do like setting `from` on the events to the bridge which initiates the refund, and that logic is easy to change, but I really don't like adding a new if/else into the already fairly complex `finalizeBridgeERC20` flow. And I think we should also remove the new event from the ERC721 Bridge.

My preference would still be to add a flag instead of using the sender as a flag. Messing with senders seems bug prone. Adding a flag is still explicit enough for accounting purposes.

maurelian:
Hmmm, OK. I can put a bool on it.

tynes:
This needs a rebase, will review afterwards

maurelian:
Closing for now, will reopen a new PR which applies this change to the NFT bridge at the same time.

PR #3514, opened 2022-09-21
maint(ct): no full ethers in entrypoint

smartcontracts:
**Description**
Updates the contracts package so that the files imported by the entrypoint do not use the full ethers package and only use submodules. Useful because this allows people to only import specific submodules of ethers and not have to bundle the entire ethers package.


tynes:
Looks like an ethers peer dep problem with smock

```
1) L1ChugSplashProxy
       fallback
         should throw an error if the owner has signalled an upgrade:
     TypeError: abi.map is not a function
      at new Interface (/root/project/node_modules/ethers/node_modules/@ethersproject/abi/src.ts/interface.ts:100:47)
      at Object.ethersInterfaceFromSpec (/root/project/node_modules/@defi-wonderland/smock/src/factories/ethers-interface.ts:24:12)
      at Sandbox.fake (/root/project/node_modules/@defi-wonderland/smock/src/sandbox.ts:42:13)
      at Object.fake (/root/project/node_modules/@defi-wonderland/smock/src/index.ts:12:24)
      at async Context.<anonymous> (test/contracts/chugsplash/L1ChugSplashProxy.spec.ts:169:21)
```

packages/data-transport-layer/package.json
```diff
@@ -52,7 +52,7 @@
     "ethers": "^5.7.0",
     "express": "^4.17.1",
     "express-prom-bundle": "^6.3.6",
-    "level": "^6.0.1",
+    "level6": "npm:level@^6.0.1",
```
smartcontracts, line 55: @tynes this PR is marked as "protocol critical" because it touches the DTL, but this is the only change being made. I had to use this alias because:
1. We updated `smock` to 2.3.0 which now imports some new hardhat package.
2. The new hardhat package being imported itself imports level 8.
3. Our monorepo is picking up level 8 instead of level 6 for the DTL and the interface was broken so it throws a TSC error.

I resolved this issue by explicitly telling the package to use level6. This should NOT have any impact on the behavior of the DTL since it's using the exact same version of level it was using before.

yarn.lock
```diff
@@ -2933,6 +2924,138 @@
...
+    "@nomicfoundation/ethereumjs-block" "^4.0.0"
+    "@nomicfoundation/ethereumjs-common" "^3.0.0"
+    "@nomicfoundation/ethereumjs-ethash" "^2.0.0"
+    "@nomicfoundation/ethereumjs-rlp" "^4.0.0"
+    "@nomicfoundation/ethereumjs-trie" "^5.0.0"
+    "@nomicfoundation/ethereumjs-util" "^8.0.0"
+    abstract-level "^1.0.3"
+    debug "^4.3.3"
+    ethereum-cryptography "0.1.3"
+    level "^8.0.0"
```
smartcontracts, line 2953: Here you can see where the new level 8 dependency is being added

smartcontracts:
Getting some weird access error in CI...

tynes:
The build error is happening in `contracts-governance` which is pretty strange. Perhaps a hardhat versioning thing?

tynes:
Still trying to get this merged @smartcontracts ?

smartcontracts:
Going to put this on hold until after I get back from devcon

mslipper:
Closing this - please reopen once you've had a chance to return to this.

PR #3520, opened 2022-09-21
fix(ctb): standardize storage spacers

smartcontracts:
**Description**
Standardizes storage spacer names and makes sure that storage spacers are all using their original variable types.


maurelian (approved):

tynes (approved):
Looks good to me, it makes sense to use the same types altho I feel like I haven't seen it before

PR #3521, opened 2022-09-21
feat(ctb): move message passer to new address

smartcontracts:
**Description**
Moves the L2ToL1MessagePasser to a new address and introduces the LegacyMessagePasser to live at the old address.

**Tests**
Original tests still apply for the `L2ToL1MessagePasser`. I have also added a test for the single `passMessageToL1` function inside the `LegacyMessagePasser` contract.


packages/contracts-bedrock/contracts/legacy/LegacyMessagePasser.sol
```diff
@@ -0,0 +1,33 @@
...
+     */
+    constructor() Semver(0, 0, 1) {}
+
+    /**
+     * @notice Passes a message to L1.
+     *
+     * @param _message Message to pass to L1.
+     */
+    function passMessageToL1(bytes memory _message) external {
+        sentMessages[keccak256(abi.encodePacked(_message, msg.sender))] = true;
```
tynes, line 31: We should add back the event that never was here
https://github.com/ethereum-optimism/optimism/blob/009939e0143e15c42919521bd9004cb69a05ebbe/packages/contracts/contracts/L2/predeploys/iOVM_L2ToL1MessagePasser.sol#L12
lol, since it was left out previously I'm ok with it being a different event

smartcontracts: I would prefer not to add the event. We're never going to use this contract again and we never emitted it previously. I don't want to change the behavior of a legacy contract unless strictly necessary.

smartcontracts:
Need to make some SDK changes

packages/contracts-bedrock/contracts/libraries/Predeploys.sol
```diff
@@ -67,8 +73,9 @@ library Predeploys {
...
 
     /**
-     * @notice Address of the GasPriceOracle predeploy. Includes fee information
-     *         and helpers for computing the L1 portion of the transaction fee.
+     * @custom:legacy
+     * @notice Address of the LegacyMessagePasser predeploy. Deprecate. Use the updated
+     *         L2ToL1MessagePasser contract instead.
      */
-    address internal constant GAS_PRICE_ORACLE = 0x420000000000000000000000000000000000000F;
+    address internal constant LEGACY_MESSAGE_PASSER = 0x4200000000000000000000000000000000000016;
```
maurelian, line 80: did you mean this?
```suggestion
    address internal constant LEGACY_MESSAGE_PASSER = 0x4200000000000000000000000000000000000000;
```

smartcontracts: Oops, thanks!

packages/contracts-bedrock/contracts/legacy/LegacyMessagePasser.sol
smartcontracts: Actually, this is a good point. Maybe legacy contracts don't need semver.

smartcontracts: `LegacyERC20ETH` isn't getting Semver which is inconsistent

smartcontracts: Ok started discussion in slack, can figure it out there

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1220,7 +1220,7 @@ export class CrossChainMessenger {
     // compatible without WithdrawalInitiatedExtension1
     const logs: Partial<{ number: WithdrawalEntry }> = {}
     for (const [i, log] of Object.entries(receipt.logs)) {
-      if (log.address === predeploys.OVM_L2ToL1MessagePasser) {
+      if (log.address === this.contracts.l2.BedrockMessagePasser.address) {
```
tynes, line 1223: Nit: I don't think this should be bedrock message passer because we likely wont migrate message passers for the release that comes after bedrock

smartcontracts: What do you mean?

smartcontracts: Also I made it `BedrockMessagePasser` because people will expect `L2ToL1MessagePasser` to refer to the old message passer until the Bedrock upgrade goes through. Then once we ship Bedrock I'll make a new PR to rename it to `L2ToL1MessagePasser` and have the old contract be `LegacyMessagePasser`.

tynes (approved):

maurelian (approved):

PR #3522, opened 2022-09-21
ctp: Remove ERC721Refunded events

maurelian:
**Description**

Simplifies branching logic in the ERC721 bridge by removing the `ERC20Refunded` event. 

This has the downside of making it slightly more difficult for indexers to differentiate between a transfer from Bob to Alice vs. a refund resulting from a failed transfer from Alice to Bob. However: 
1. This is most likely to happen in case where we don't care about one of the two tokens. 
2. A proper indexer should be watching both bridges, and can detect the `ERC721BridgeFailed` event.

**Tests**

I removed a couple tests that became redundant after removing this functionality.

**Additional context**

Yay for code deletion.


smartcontracts (approved):

packages/contracts-periphery/contracts/L2/L2ERC721Bridge.sol
```diff
@@ -115,16 +102,18 @@ contract L2ERC721Bridge is ERC721Bridge, Semver {
...
+            // Note that supportsInterface makes a callback to the _localToken address
+            // which is user provided.
             ERC165Checker.supportsInterface(_localToken, type(IOptimismMintableERC721).interfaceId),
             "L2ERC721Bridge: local token interface is not compliant"
         );
         require(
             _remoteToken == IOptimismMintableERC721(_localToken).remoteToken(),
             "L2ERC721Bridge: wrong remote token for Optimism Mintable ERC721 local token"
         );
-        require(_localToken != address(this), "L2ERC721Bridge: local token cannot be self");
```
maurelian, line 117: apologies for the slight larger diff, I also reordered the checks here to be more consistent between the L1 and L2 bridges.

smartcontracts (approved):

tynes (approved):

PR #3524, opened 2022-09-21
chore: Update op-geth to latest version

mslipper:
Update op-geth to the latest version on the `optimism-history` branch. This version exposes the customizable simulator backend that we need to merge https://github.com/ethereum-optimism/optimism/pull/3322.

Also adds a small script to automatically update the op-geth version in the future.

### DIFF
--- a/go.work.sum
+++ b/go.work.sum
@@ -27,6 +27,8 @@ github.com/envoyproxy/go-control-plane v0.9.9-0.20210217033140-668b12f5399d/go.m
 github.com/envoyproxy/go-control-plane v0.9.10-0.20210907150352-cf90f659a021/go.mod h1:AFq3mo9L8Lqqiid3OhADV3RfLJnjiw63cSpi+fDTRC0=
 github.com/ethereum-optimism/op-geth v0.0.0-20220904174542-4311f9d2cead/go.mod h1:EYFyF19u3ezGLD4RqOkLq+ZCXzYbLoNDdZlMt7kyKFg=
 github.com/ethereum-optimism/op-geth v0.0.0-20220906163738-712cb0a1c140/go.mod h1:EYFyF19u3ezGLD4RqOkLq+ZCXzYbLoNDdZlMt7kyKFg=
+github.com/ethereum-optimism/op-geth v0.0.0-20220909213840-e6575c0168f1 h1:W/ZU6BZH7ilTrpdoJOF9N4OReqXbpeRtUB6klIpEdMA=
+github.com/ethereum-optimism/op-geth v0.0.0-20220909213840-e6575c0168f1/go.mod h1:/6CsT5Ceen2WPLI/oCA3xMcZ5sWMF/D46SjM/ayY0Oo=
 github.com/fjl/gencodec v0.0.0-20220412091415-8bb9e558978c/go.mod h1:AzA8Lj6YtixmJWL+wkKoBGsLWy9gFrAzi4g+5bCKwpY=
 github.com/frankban/quicktest v1.11.3/go.mod h1:wRf/ReqHper53s+kmmSZizM8NamnL3IM0I9ntUbOk+k=
 github.com/garslo/gogen v0.0.0-20170306192744-1d203ffc1f61/go.mod h1:Q0X6pkwTILDlzrGEckF6HKjXe48EgsY/l7K7vhY4MW8=
--- a/op-chain-ops/deployer/deployer.go
+++ b/op-chain-ops/deployer/deployer.go
@@ -5,14 +5,13 @@ import (
 	"fmt"
 	"math/big"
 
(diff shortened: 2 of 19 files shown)


op-chain-ops/go.mod
```diff
@@ -58,4 +58,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-replace github.com/ethereum/go-ethereum v1.10.21 => github.com/ethereum-optimism/op-geth v0.0.0-20220904174542-4311f9d2cead
+replace (
+	github.com/ethereum/go-ethereum v1.10.21 => github.com/ethereum-optimism/op-geth v0.0.0-20220904174542-4311f9d2cead
```
trianglesphere, line 62: Why are we requiring two different versions of geth here?

trianglesphere (commented):
I think there is one double import, but overall looking good. Having the script is super nice rather than doing it by hand.

op-chain-ops/go.mod
mslipper: Nice catch. We shouldn't have to different versions. Fixing.

mslipper:
Tests will run once CircleCI fixes its outage.

ops/scripts/update-op-geth.sh
```diff
@@ -0,0 +1,15 @@
...
+export GOPRIVATE="github.com/ethereum-optimism"
+
+for value in $PROJECTS
+do
+	echo "Updating $value"
+	cd "./$value"
+	go mod edit -replace "github.com/ethereum/go-ethereum@$GETH_VERSION=github.com/ethereum-optimism/op-geth@optimism-history"
+	go mod tidy
+	cd ../
+done
```
protolambda, line 15: Let's add a `go work edit -replace ...........`

protolambda (changes requested):
`go.work` also contains a reference to op-geth. The script should include that too.

ops/scripts/update-op-geth.sh
mslipper: ah good call, always forget that.

protolambda (approved):

go.work.sum
```diff
@@ -27,6 +27,8 @@ github.com/envoyproxy/go-control-plane v0.9.9-0.20210217033140-668b12f5399d/go.m
 github.com/envoyproxy/go-control-plane v0.9.10-0.20210907150352-cf90f659a021/go.mod h1:AFq3mo9L8Lqqiid3OhADV3RfLJnjiw63cSpi+fDTRC0=
 github.com/ethereum-optimism/op-geth v0.0.0-20220904174542-4311f9d2cead/go.mod h1:EYFyF19u3ezGLD4RqOkLq+ZCXzYbLoNDdZlMt7kyKFg=
 github.com/ethereum-optimism/op-geth v0.0.0-20220906163738-712cb0a1c140/go.mod h1:EYFyF19u3ezGLD4RqOkLq+ZCXzYbLoNDdZlMt7kyKFg=
```
tynes, line 29: Are there supposed to be multiple versions of `op-geth` here?

ops/scripts/update-op-geth.py
```diff
@@ -0,0 +1,61 @@
...
+def main():
+	for project in ('op-service', 'op-node', 'op-proposer', 'op-batcher', 'op-bindings', 'op-chain-ops', 'op-e2e'):
+		print(f'Updating {project}...')
+		update_mod(project)
+
+	# pull the replacer from one of the modules above, since
+	# go work edit -replace does not resolve the branch name
+	# into a pseudo-version.
+	pv = None
+	with open('./op-service/go.mod') as f:
```
tynes, line 22: This only needs to run in `op-service`?

op-chain-ops/deployer/deployer.go
```diff
@@ -39,14 +38,17 @@ type Deployment struct {
 type Deployer func(*backends.SimulatedBackend, *bind.TransactOpts, Constructor) (common.Address, error)
 
 func NewBackend() *backends.SimulatedBackend {
-	return backends.NewSimulatedBackendWithCacheConfig(
```
trianglesphere, line 42: Is this a regression with deployments?

tynes (approved):

trianglesphere (approved):
How much of this tooling will we get to delete if we are using tags rather than commit hashes/branches?

op-chain-ops/deployer/deployer.go
mslipper: Nope, the API was updated in `op-geth`.

mslipper:
> How much of this tooling will we get to delete if we are using tags rather than commit hashes/branches?

The bit that resolves the branch to the pseudo-version in the updater script. We'll still need the rest.

PR #3527, opened 2022-09-22
contracts-bedrock: fuzz L2ToL1MessagePasser

tynes:
**Description**

Fuzzes the L2ToL1MessagePasser for state changes to its storage. This specifically tests that the storage slot that the proofs are generated against is computed correctly. This is important to test as any changes that break the proof would also break this test.

The fuzz runs were used to generate test cases in
https://github.com/ethereum-optimism/optimism/pull/3469 for off chain computation of the storage slots for withdrawals.

### DIFF
--- a/.changeset/hot-panthers-reply.md
+++ b/.changeset/hot-panthers-reply.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/contracts': patch
+---
+
+Expose the deployments in the deployer image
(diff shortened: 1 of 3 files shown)


maurelian (approved):

tynes:
I've resolved the merge conflict, this should be good to review @smartcontracts

smartcontracts (approved):

PR #3528, opened 2022-09-22
feat: withdrawal migration

tynes:
**Description**

Implement withdrawal hash migration functions with tests. The tests reuse the receipts in `op-chain-ops/crossdomain/testdata`.
The branch coverage does hit both the `finalizeETHWithdrawal` case and the other case.

To really know if this migration code works, we need to do end to end tests.

Closes ENG-2743


op-chain-ops/crossdomain/migrate.go
```diff
@@ -0,0 +1,112 @@
...
+	return nil
+}
+
+// MigrateWithdrawal will turn a LegacyWithdrawal into a bedrock
+// style Withdrawal.
+func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger, l1StandardBridge *common.Address) (*Withdrawal, error) {
+	value := new(big.Int)
+
+	isFromL2StandardBridge := *withdrawal.Sender == predeploys.L2StandardBridgeAddr
+	isToL1StandardBridge := *withdrawal.Target == *l1StandardBridge
```
trianglesphere, line 55: Is target guaranteed to be non-nil?

op-chain-ops/crossdomain/migrate.go
```diff
@@ -0,0 +1,112 @@
...
+		new(big.Int),
+		withdrawal.Data,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot abi encode relayMessage: %w", err)
+	}
+
+	w := NewWithdrawal(
+		withdrawal.Nonce,
+		&predeploys.L2CrossDomainMessengerAddr,
```
trianglesphere, line 105: Are all legacy withdrawals from the L2 xdm?

op-chain-ops/crossdomain/migrate.go
```diff
@@ -0,0 +1,112 @@
...
+		if err != nil {
+			return err
+		}
+
+		legacyValue := db.GetState(predeploys.LegacyMessagePasserAddr, legacySlot)
+		if legacyValue != abiTrue {
+			return fmt.Errorf("%w: %s", errLegacyStorageSlotNotFound, legacyValue)
+		}
+
+		// TODO: pass in args
```
trianglesphere, line 33: what's the plan to resolve this?

op-chain-ops/crossdomain/migrate.go
```diff
@@ -0,0 +1,112 @@
...
+
+// MigrateWithdrawal will turn a LegacyWithdrawal into a bedrock
+// style Withdrawal.
+func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger, l1StandardBridge *common.Address) (*Withdrawal, error) {
+	value := new(big.Int)
+
+	isFromL2StandardBridge := *withdrawal.Sender == predeploys.L2StandardBridgeAddr
+	isToL1StandardBridge := *withdrawal.Target == *l1StandardBridge
+
+	if isFromL2StandardBridge && isToL1StandardBridge {
```
trianglesphere, line 57: Is this ever not true or is it possible to withdraw eth by another method?

op-chain-ops/crossdomain/migrate.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/vm"
+)
+
+var (
+	abiTrue                      = common.Hash{31: 0x01}
+	errLegacyStorageSlotNotFound = errors.New("cannot find storage slot")
+)
+
+// This takes a state db and a list of withdrawals
```
trianglesphere, line 20: doc comment could be a bit more descriptive of what the method does instead of the param name/types.

trianglesphere (commented):
Code is pretty good. A lot of my questions are about making sure that I have enough context about the system.

op-chain-ops/crossdomain/migrate.go
tynes: Assuming there is no adversarial input, yes target will always be non-nil. I can double check it

trianglesphere (approved):
lgtm

op-chain-ops/crossdomain/migrate.go
tynes: @smartcontracts Can you clarify what you mean by this? Based on the way that we collect the withdrawals, we will not get any that do not come from the L2CrossDomainMessenger, since we rely on that event.

Any storage slots that exist in the LegacyMessagePasser set by a contract call where `msg.sender != address(L2CrossDomainMessenger)` doesn't need to be upgraded since it isn't part of our proof system anyways.

smartcontracts (approved):

PR #3530, opened 2022-09-22
op-node: Fix OPB-04

mslipper:
The length of the ExtraData field was not being checked, which could lead to panics in the MarshalSSZ function.

### DIFF
--- a/op-node/eth/ssz.go
+++ b/op-node/eth/ssz.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


op-node/eth/ssz.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/ethereum-optimism/optimism/l2geth/common/math"
```
tynes, line 7: This should not import `l2geth` but instead upstream geth

tynes (approved):

mslipper:
Test is failing because it allocates more memory than is available.

PR #3531, opened 2022-09-22
feat(ct): governance token in predeploys

smartcontracts:
**Description**
Moves the governance token address into predeploys rather than futurePredeploys now that it's been deployed.

### DIFF
--- a/.changeset/olive-taxis-bake.md
+++ b/.changeset/olive-taxis-bake.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/contracts': patch
+---
+
+GovernanceToken now available in the predeploys constant.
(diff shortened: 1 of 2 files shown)


.changeset/olive-taxis-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@eth-optimism/contracts': minor
```
tynes, line 2: I don't think this necessarily needs to be a minor change but not super opinionated

tynes (approved):

.changeset/olive-taxis-bake.md
smartcontracts: Made patch instead

mslipper (approved):

smartcontracts:
Closing because something subtle is broken and this is not important enough for me to figure out why it's broken

PR #3532, opened 2022-09-22
op-node: Cleanup calldata source API

trianglesphere:
**Description**

This provides a new set of objects which provide the following API guarantees:
- The opening of a data for a new L1 block will never fail.
- If it failed to get transactions, it does so in the calls to `Next`.

This greatly simplifies usage of this object when constructing new data. If the node did not keep track of the internal state, the external users of this API would have to keep track of this state in a more complex way.

**Tests**

**Additional context**

**Metadata**
- Fixes ENG-2746


op-node/rollup/derive/l1_retrieval.go
```diff
@@ -2,37 +2,24 @@ package derive
...
 
+type CalldataProvider interface {
+	OpenData(ctx context.Context, id eth.BlockID) DataIter
+}
+
 type L1Retrieval struct {
 	log     log.Logger
-	dataSrc DataAvailabilitySource
+	dataSrc CalldataProvider
 	next    L1SourceOutput
```
protolambda, line 23: Can we keep the name more general here? It's meant to be swapped out for EIP-4844 and other data sources easily, which are not calldata.

protolambda (commented):
LGTM, but please rename the interface in the retrieval stage back

protolambda (approved):

mslipper (approved):

PR #3533, opened 2022-09-22
feat(ctb): spacer validation task

smartcontracts:
**Description**
Introduces a new task for validating the correctness of spacer variables. Prevents future developers from accidentally moving spacers.


packages/contracts-bedrock/tasks/validate-spacers.ts
```diff
@@ -0,0 +1,37 @@
...
+            const [, slot, offset] = variable.label.split('_')
+            if (slot !== variable.slot) {
+              throw new Error(
+                `${contractName} ${variable.label} is in slot ${variable.slot} but should be in ${slot}`
+              )
+            }
+            if (parseInt(offset, 10) !== variable.offset) {
+              throw new Error(
+                `${contractName} ${variable.label} is at offset ${variable.offset} but should be at ${offset}`
+              )
```
maurelian, line 30: I noticed that you're not checking that the length given in the space name matches the type...
which could be done easily enough given all spacers are currently either a mapping, uint or address.

That would be a slight improvement, but I'm fine either way. Will approve now and you can resolve the comment if you just want to merge it.

maurelian (approved):

packages/contracts-bedrock/tasks/validate-spacers.ts
smartcontracts: Oops, good catch. Added an simple version of this to the script.

maurelian:
@mergify bump

tynes (approved):
Easy to read and understand

PR #3534, opened 2022-09-22
docs(ctb): add versioning rules

smartcontracts:
**Description**
Adds versioning rules to the Bedrock contracts package.


maurelian (approved):

packages/contracts-bedrock/README.md
```diff
@@ -138,3 +138,19 @@ Unless explicitly discussed otherwise, you MUST include the following basic upgr
...
+All (non-library and non-abstract) contracts MUST extend the `Semver` base contract which exposes a `version()` function that returns a semver-compliant version string.
+During the Bedrock development process the `Semver` value for all contracts SHOULD return `0.0.1` (this is not particularly important, but it's an easy standard to follow).
+When the initial Bedrock upgrade is released, the `Semver` value MUST be updated to `1.0.0`.
+
+After the initial Bedrock upgrade, contracts MUST use the following versioning scheme:
+
+- `patch` releases are to be used only for changes that do NOT modify contract bytecode (such as updating comments).
+- `minor` releases are to be used for changes that modify bytecode OR changes that expand the contract ABI provided that these changes do NOT break the existing interface.
+- `major` releases are to be used for changes that break the existing contract interface.
+
```
tynes, line 153: I think we should enforce semver bumps for any redeployment.

Patch releases could also include the same bytecode with altered immutables

Minor is changes to the executable bytecode, which includes expanding the abi, but the security model stays the same

Major can break the abi or alter the security model

packages/contracts-bedrock/README.md
```diff
@@ -138,3 +138,19 @@ Unless explicitly discussed otherwise, you MUST include the following basic upgr
 2. Include a `uint8 public constant VERSION = X` at the TOP of your contract.
 3. Include a function `initialize` with the modifier `reinitializer(VERSION)`.
 4. In the `constructor`, set any `immutable` variables and call the `initialize` function for setting mutables.
+
```
tynes, line 141: I have a slight preference for adding this to the specs over the readme

smartcontracts: Can do that later, I think it's fine here for now

packages/contracts-bedrock/README.md
smartcontracts: I would consider altered immutables to be a change to bytecode and therefore `minor`.

smartcontracts: My preference is that `patch` be reserved for changes that do not modify any executable paths

smartcontracts: I added your comment re: major changes

tynes: I feel like `patch` upgrades won't happen much at all with this kind of scheme. I suppose i'm ok with that

smartcontracts: I think that's fine. `patch` should be relatively rare in the context of a smart contract.

tynes (approved):

PR #3535, opened 2022-09-23
feat(ctb): no refunds

smartcontracts:
**Description**
Removes refunds from the StandardBridge flow.

**Tests**
Removes tests related to refund logic.


packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -418,6 +370,11 @@ abstract contract StandardBridge {
         uint32 _minGasLimit,
         bytes memory _extraData
     ) internal {
+        require(
```
smartcontracts, line 373: Sorry couldn't resist making this a little more defensive

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -470,39 +423,8 @@ abstract contract StandardBridge {
...
-            _localToken,
-            _remoteToken,
-            _from,
-            _to,
-            _amount,
-            _minGasLimit,
-            _extraData
-        );
-    }
+        emit ERC20BridgeInitiated(_localToken, _remoteToken, _from, _to, _amount, _extraData);
```
smartcontracts, line 426: Checks, effects, interactions. Should've been above the sendMessage call.

packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol
```diff
@@ -191,15 +160,11 @@ contract L2StandardBridge is StandardBridge, Semver {
     ) internal {
         address l1Token = OptimismMintableERC20(_l2Token).l1Token();
         if (_l2Token == Predeploys.LEGACY_ERC20_ETH) {
-            require(
```
smartcontracts, line 194: Cleaner to do this check inside of `_initiateBridgeETH`

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
smartcontracts, line 373: Couldn't resist making this a little more defensive.

tynes:
Needs a regenerated snapshot

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -313,10 +286,13 @@ abstract contract StandardBridge {
         uint256 _amount,
         bytes calldata _extraData
     ) public payable onlyOtherBridge {
-        require(msg.value == _amount, "StandardBridge: amount sent does not match amount required");
-        require(_to != address(this), "StandardBridge: cannot send to self");
```
tynes, line 317: Did you intend to remove `cannot send to self` check?

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
tynes: Do we have test coverage that this will not break replayability?

smartcontracts: I believe this line will never actually be triggered in our current codebase, so maybe it's better to use `assert`. Also would not break replayability because the ETH value comes out of the `L2CrossDomainMessenger` either way. If the first message fails then the ETH is still stuck in the `L2CrossDomainMessenger` to be sent again on replay.

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
smartcontracts: Reverting this change and adding tests for the behavior

maurelian: > I guess we could keep the check here to prevent people from accidentally bricking funds.

Two reactions: 
1. If so, we should do the same in `finalizeBridgeERC20` below.
2. I don't think this check prevents people from losing funds, if they're calling `finalizeBridgeETH` they already lost funds when the initiate bridging on the other side.

maurelian: I guess suggested path forward is that we should either:
1. Not do this check, or
2. Do it on the initiating side.

maurelian (approved):
lgtm

tynes:
This is going to need a rebase, will review afterwards

tynes:
Will need a follow up PR that removes this functionality from the NFT bridge

tynes:
Looks like another PR was merged, needs another rebase

mslipper:
I'll take care of it

tynes (approved):

zhiqiangxu:
What's the rationale behind the decision to remove refunds out of the bridge?

PR #3536, opened 2022-09-23
feat(ctb): remove SecureMerkleTrie wrapper

smartcontracts:
**Description**
Removes the SecureMerkleTrie wrapper and cleans up the structure of the trie and related libraries.

**Tests**
Same tests as before should work just fine.


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -302,12 +303,12 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
         );
 
-        return
-            SecureMerkleTrie.verifyInclusionProof(
-                abi.encode(storageKey),
-                hex"01",
-                _proof,
-                _storageRoot
-            );
+        (bool exists, bytes memory value) = MerkleTrie.get(
```
tynes, line 306: I feel like the previous API is less bug prone

tynes: An example here is the first argument should be a bytes32 to enforce that a hashed value is passed in. There is no way for the caller to know that a hashed value should be passed in. I like the older API more because its more clear how to use it

tynes (changes requested):
Please sell me on why this is a good idea :)

tynes:
@smartcontracts Planning on working on this still?

smartcontracts:
Not right now, let's scrap this for the moment

PR #3541, opened 2022-09-23
contracts-periphery: remove ownable upgradable from erc721 factory

tynes:
**Description**

The ownable upgradable modifier was not used anywhere in the contract. The erc20 factory does not follow that pattern. This change simplifies things by moving
values to immutables instead of being in storage.

This is a cleanup PR before migrating these contracts to `contracts-bedrock`.


smartcontracts (approved):
Nice, thanks

maurelian (approved):
Great changes

PR #3542, opened 2022-09-23
Use uint64 for arithmetic in XDM's baseGas

maurelian:
**Description**

This PR updates the CrossDomainMessengers' `baseGas()` method to return a `uint64` and also for internal arithmetic. This fixes an issue that would have cause a revert with any `_minGasLimit` above ~4.2 MM.

**Additional Notes:**

Internal casting is used, so the argument type to `baseGas` was unchanged from `uint32`, however the return value was modified to be `uint64`. Interestingly the only usage of that return value in the `_sendMessage()` function was already silently upcasting it to `uint64` anyways.

**Tests**

I added a simple forge fuzz test that calls `baseGas()` with any input and another unit test that calls it with max possible value.


tynes:
I like the idea here, are we sure that its safe to go over a uint32 though? I think so since relay message uses a uint256 for the min gas limit

tynes:
I hit this problem when writing tests in go, was getting a `Panic()` for overflow and had to reduce the gaslimit i was setting

maurelian:
> I like the idea here, are we sure that its safe to go over a `uint32` though? I think so since relay message uses a `uint256` for the min gas limit

As implemented, the user input is still limited to `uint32`, it's only upcast within the `baseGas` function. 

Also worth noting that everywhere the output of `baseGas` is used was already accepting a `uint64`, so I'm not sure why `uint32` was chosen.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -369,13 +369,16 @@ abstract contract CrossDomainMessenger is
      *
      * @return Amount of gas required to guarantee message receipt.
      */
-    function baseGas(bytes calldata _message, uint32 _minGasLimit) public pure returns (uint32) {
+    function baseGas(bytes calldata _message, uint32 _minGasLimit) public pure returns (uint64) {
```
maurelian, line 372: Note: 
The inputs are unchanged, but the internal math and return value are now `uint64`. Everywhere its used is already accepting a `uint64` value though: 

ie. in `_sendMessage`: 

https://github.com/ethereum-optimism/optimism/blob/4bc75740ad830b51cf0ad61bce63e4bb37158736/packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol#L410

Which is implemented on L1 as: 
https://github.com/ethereum-optimism/optimism/blob/4bc75740ad830b51cf0ad61bce63e4bb37158736/packages/contracts-bedrock/contracts/L1/L1CrossDomainMessenger.sol#L45-L52

and on L2 as:

https://github.com/ethereum-optimism/optimism/blob/4bc75740ad830b51cf0ad61bce63e4bb37158736/packages/contracts-bedrock/contracts/L2/L2CrossDomainMessenger.sol#L51-L60

maurelian: A more general question though is "what happens when either of these functions get a gas limit greater than the block gas limit". 
I think that requires an iTest.

tynes, line 372: I think you should make `_minGasLimit` a `uint64` as an argument here

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Messenger_Initializer, Reverter, CallerCaller } from "./CommonTest.t.sol";
```
tynes, line 4: `Reverter` and `CallerCaller` are not used

tynes:
Will re-review after a rebase

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
smartcontracts: The reason for uint32 was because `sendMessage` takes uint32 as input: https://github.com/ethereum-optimism/optimism/blob/e59b44965256a0d3a20166dabccc523a74a10d14/packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol#L219-L223

smartcontracts: This is a holdover from the old `sendMessage` interface: https://github.com/ethereum-optimism/optimism/blob/e59b44965256a0d3a20166dabccc523a74a10d14/packages/contracts/contracts/L1/messaging/L1CrossDomainMessenger.sol#L137-L141

smartcontracts: To be honest I don't know why `_gasLimit` was changed from a `uint32` to a `uint64`. `uint32` was sufficient.

smartcontracts: If anything my preference would actually be to go in the other direction and change the new interface to use `uint32` instead of `uint64` and keep everything more consistent.

smartcontracts: There's no point in having `uint64` in `_sendMessage` if the public `sendMessage` only accepts `uint32`.

tynes: The reason why I think this change makes sense is that the entire set of `uint32`s is not valid inputs to `sendMessage` because `_baseGas` will pad `_gasLimit` to a larger value. By making it a `uint64`, all `uint32`s are valid inputs. Users should never be getting `Panic(0x11)` overflow errors back by setting `_gasLimit` too high

maurelian:
converting to draft until I can: 
1. understand the tradeoffs
2. get consensus

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
maurelian: @tynes where does the `Panic(0x11)` come from?

maurelian: > By making it a `uint64`, all `uint32s` are valid inputs. Users should never be getting `Panic(0x11)` overflow errors back by setting _gasLimit too high

In practice, the amount required to cause an overflow is going to be very close to `uint32` (1000/1016ths), so about 4 Billion gas.


@tynes how do you feel about [this approach](https://github.com/ethereum-optimism/optimism/blob/6b75037ca3545b8c60eef252badfb43d66a6cb59/packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol#L376-L386) which would throw an `Arithmetic: overflow/underflow` error on the downcast step?

If we want a more descriptive message, we could add an explicit check, something like:
`require(output < type(uint32).max, "The amount of gas supplied would lead to an excessively large gas limit")`

tynes: > @tynes where does the `Panic(0x11)` come from?

`solc` returns this on overflow/underflow

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Messenger_Initializer } from "./CommonTest.t.sol";
+
+// CrossDomainMessenger_Test is for testing functionality which is common to both the L1 and L2
+// CrossDomainMessenger contracts. For simplicity, we use the L1 Messenger as the test contract.
+contract CrossDomainMessenger_Test is Messenger_Initializer {
+    // Ensure that baseGas passes for the max value of _minGasLimit.
+    function test_baseGas() external {
```
tynes, line 10: I'm curious about the max size of the message before it will revert given max uint32

maurelian: Agree, that's a good question. I'll modify these tests a bit to make sure they're clearly showing where the boundaries are.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
maurelian: resolving this lengthy thread in favor of discussion about the types elsewere.

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
maurelian: @tynes I added a test showing that the message length is not bounded by arithmetic overflow/underflow, but rather by the block gas limit, meaning it's not possible to submit a large enough message to cause a revert here.

https://github.com/ethereum-optimism/optimism/blob/10c12d6173ef7028269f218a8debc1038f6ff25f/packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol#L20

maurelian:
Ater some though, I think the way to go here is to keep the interface of `baseGas` unchanged, with `uint32` for both the input and output gas amounts, ie. 
```
function baseGas(bytes calldata _message, uint32 _minGasLimit) public pure returns (uint32)
```

Reasoning: 
1. For consistency, `baseGas()` should have the same input type as `sendMessage()`, and we need to maintain the [pre-existing](https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts/contracts/L1/messaging/L1CrossDomainMessenger.sol#L140) interface for `sendMessage()` which has a `uint32`.
2. The output type should be the same as the output type, otherwise it's awkward. 

In order to prevent users from receiving a `Panic(0x11)` message, we will detect an overflow and return a better error message. 

Final changes will follow shortly.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
maurelian: Decided on `uint32` based on reasoning [here](https://github.com/ethereum-optimism/optimism/pull/3542#issuecomment-1266331518). LMK what you think.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -370,14 +370,20 @@ abstract contract CrossDomainMessenger is
...
-            // Dynamic overhead
-            ((_minGasLimit * MIN_GAS_DYNAMIC_OVERHEAD_NUMERATOR) /
-                MIN_GAS_DYNAMIC_OVERHEAD_DENOMINATOR) +
-            // Calldata overhead
-            (uint32(_message.length) * MIN_GAS_CALLDATA_OVERHEAD) +
-            // Constant overhead
+        // We peform the following math on uint64s to avoid overflow errors. Multiplying by
+        // MIN_GAS_DYNAMIC_OVERHEAD_NUMERATOR would otherwise limit the _minGasLimit to
+        // approximately 4.2 MM.
+        // First component is dynamic overhead
```
tynes, line 376: Can we put this in an `unchecked` block since we now have the require at the end?

tynes:
Nice tests

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
maurelian: Yeah, good call out

tynes:
```
warning[6321]: Warning: Unnamed return variable can remain unassigned when the function is called when "L1CrossDomainMessenger" is the most derived contract. Add an explicit return with value to all non-reverting code paths or name the variable.
   --> contracts/universal/CrossDomainMessenger.sol:372:89:
    |
372 |     function baseGas(bytes calldata _message, uint32 _minGasLimit) public pure returns (uint32) {
    |     
```

tynes:
The bridge erc20 deposit timed out, wondering if this change is breaking the deposit flow or if its just a flake

tynes (approved):

tynes:
This looks like it needs a rebase @maurelian

PR #3544, opened 2022-09-23
proxyd: Add customizable whitelist error message

mslipper:
Alchemy is asking for this so that we can include their affiliate link.


tynes:
Generally looks good to me, how did you confirm this works as expected?

mslipper:
It's tested as part of the integration tests. The itests boot up the server, which modifies the messages.

mslipper:
Just realized that the whitelist custom message case was not tested - just fixed that.

Inphi (approved):

PR #3550, opened 2022-09-24
feat(cmn): keep raw body in requests

smartcontracts:
**Description**
Minor modification that puts the raw body as a string on the request object. Useful for applications where the raw body is required rather than the parsed JSON body as given by bodyParser. For example, this is useful when verifying GitHub webhooks.


smartcontracts:
I need this specifically so that the bot I'm writing for the token list repository can verify that webhooks are coming from GitHub and not somewhere else

mslipper (approved):

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -386,9 +386,17 @@ export abstract class BaseServiceV2<
       const app = express()
 
       // Body parsing.
-      app.use(bodyParser.json())
       app.use(bodyParser.urlencoded({ extended: true }))
 
+      // Keep the raw body around in case the application needs it.
+      app.use(
+        bodyParser.json({
```
tynes, line 393: This still will parse the json body as before?

mslipper: Yes - the parsing will continue as long as `verify` does not through an error.

tynes (approved):

PR #3555, opened 2022-09-25
fix: turn nft bridge into predeploy

tynes:
**Description**

Migrates the nft bridge to `contracts-bedrock`.

This should not be merged until the code can be frozen. We should decide on whether or not we want to hardcode known addresses or instead keep the constructors the way that they are. If we alter the way that the constructors work, then the old integration tests will no longer work since they run on `ops` which doesn't have the nft bridge as a predeploy

Closes ENG-2793


packages/contracts-bedrock/contracts/L2/L2ERC721Bridge.sol
```diff
@@ -0,0 +1,173 @@
+// SPDX-License-Identifier: MIT
```
maurelian, line 1: Shouldn't this just appear as a file rename from the periphery package?

packages/contracts-bedrock/contracts/L2/L2ERC721Bridge.sol
```diff
@@ -0,0 +1,173 @@
...
+            /// In either case, we stop the process here and construct a withdrawal in which we
+            // flip the to and from addresses. This ensures that event-based accounting
+            // will indicate net-zero transfer to the recipient. The ERC721BridgeFailed event
+            // emitted below can also be used to identify this occurence.
+            bytes memory message = abi.encodeWithSelector(
+                L1ERC721Bridge.finalizeBridgeERC721.selector,
+                _remoteToken,
+                _localToken,
+                _to,
+                _from, // Refund the NFT to the original owner on the remote chain.
```
maurelian, line 73: thought this got taken out

tynes:
Turning this into a draft until I have more time to work on it

packages/contracts-bedrock/contracts/L2/L2ERC721Bridge.sol
tynes: I've rebased on the latest changes

op-bindings/predeploys/addresses.go
```diff
@@ -3,35 +3,39 @@ package predeploys
...
+	L2StandardBridge              = "0x4200000000000000000000000000000000000010"
+	SequencerFeeVault             = "0x4200000000000000000000000000000000000011"
+	OptimismMintableERC20Factory  = "0x4200000000000000000000000000000000000012"
+	L1BlockNumber                 = "0x4200000000000000000000000000000000000013"
+	GasPriceOracle                = "0x420000000000000000000000000000000000000F"
+	L1Block                       = "0x4200000000000000000000000000000000000015"
+	GovernanceToken               = "0x4200000000000000000000000000000000000042"
+	LegacyMessagePasser           = "0x4200000000000000000000000000000000000000"
+	L2ERC721Bridge                = "0x4200000000000000000000000000000000000014"
+	OptimismMintableERC721Factory = "0x4200000000000000000000000000000000000018"
```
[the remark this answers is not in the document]
tynes: 14 was reserved and nothing was at it. It is currently the ERC721Bridge on mainnet. I fixed this to not skip 17, the erc721 factory is now at 17

packages/contracts-bedrock/contracts/L2/L2ERC721Bridge.sol
tynes: git is not catching this for some reason, there must be a large enough diff in the file - I used `diff` on each to double check that the only diffs are in the imports

tynes:
I've rebased this PR and marked it ready for review again

op-bindings/predeploys/dev_addresses.go
```diff
@@ -10,6 +10,7 @@ const (
 	DevOptimismMintableERC20Factory = "0x6900000000000000000000000000000000000004"
 	DevAddressManager               = "0x6900000000000000000000000000000000000005"
 	DevProxyAdmin                   = "0x6900000000000000000000000000000000000006"
+	DevL1ERC721Bridge               = "0x6900000000000000000000000000000000000007"
```
smartcontracts, line 13: Why do we use these dev addresses?

packages/contracts-bedrock/contracts/L1/L1ERC721Bridge.sol
```diff
@@ -0,0 +1,111 @@
...
+ */
+contract L1ERC721Bridge is ERC721Bridge, Semver {
+    /**
+     * @notice Mapping of L1 token to L2 token to ID to boolean, indicating if the given L1 token
+     *         by ID was deposited for a given L2 token.
+     */
+    mapping(address => mapping(address => mapping(uint256 => bool))) public deposits;
+
+    /**
+     * @custom:semver 1.0.0
```
smartcontracts, line 23: Should this be 1.0.0? Or 0.0.1?

op-bindings/predeploys/dev_addresses.go
tynes: These dev addresses are used for deploying the L1 contracts to know addresses when building the L1 developer genesis

packages/contracts-bedrock/contracts/L1/L1ERC721Bridge.sol
tynes: It is deployed with `1.0.0`, but willing to revert if you have a strong preference

smartcontracts (approved):

mslipper (approved):

PR #3559, opened 2022-09-26
ci: Docker login on push

mslipper:
Increases the reliability of Docker publish jobs by logging in to Docker Hub first. The secret used is read-only, and scoped to public repositories only.

Also locks the nameservers used by the builder to Google's to prevent DNS issues arising from AWS's use of DNS-based load balancing.

Note that the `l1ops` Hive test will continue to fail until https://github.com/ethereum-optimism/hive/pull/43 is merged.


.circleci/config.yml
```diff
@@ -85,13 +85,21 @@ jobs:
       image: ubuntu-2204:2022.07.1
       resource_class: xlarge
     steps:
+      # BElow is CircleCI recommended way of specifying nameservers on an Ubuntu box:
```
tynes, line 88: slight nit: `BElow` -> `Below`

tynes (approved):

maurelian (approved):

PR #3499, opened 2022-09-16
feat(ctp): merge with contracts-governance

smartcontracts:
**Description**
Removes the contracts-governance package and merges the remaining scripts, contracts, and tests into contracts-periphery.

PR #3561, opened 2022-09-26
contracts-governance: mint manager deploy

tynes:
**Description**

Add a deploy script for the new mint manager. To review this, check the `cast` calls in the mint manager implementation deploy script. The owners of the `MintManager` are preserved, double check that in the `deploy-config` directory against what the cast calls return.

To deploy this, the new implementation needs to be deployed and then the existing owner of the `MintManager` needs to send a transaction to the `MintManager` calling `upgrade(address)` to the new implementation of the `MintManager`.

We are in the process of eliminating `contracts-governance` as a package, I added these changes to `contracts-governance` for simplicity. This will cause https://github.com/ethereum-optimism/optimism/pull/3499 to need updates, it already requires a rebase


packages/contracts-governance/deploy/000-mint-manager-impl.ts
```diff
@@ -0,0 +1,77 @@
...
+      console.log(`GovernanceToken.owner() is still old implementation`)
+    } else if (getAddress(owner) === getAddress(newImpl)) {
+      console.log(`GovernanceToken.owner() upgraded to new implementation!`)
+      break
+    } else {
+      throw new Error(
+        `GovernanceToken.owner() upgraded to unknown address ${owner}`
+      )
+    }
+  }
```
maurelian, line 71: This loop is basically meant to replicate what happens with the address dictator based appointment right? If so, maybe it should use more similar text for guidance.

tynes: Yes its basically meant to poll for changes in the owner. If i understand correctly, you are suggesting it should print off the instructions for what kind of tx to send? I see what you mean

mslipper:
Recusing myself from review.

tynes:
@maurelian I've updated this PR with instructions to finish the deployment

maurelian (approved):

tynes:
cc @smartcontracts

packages/contracts-governance/deploy/000-mint-manager-impl.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (getAddress(owner) === getAddress(oldImpl)) {
+      console.log(`GovernanceToken.owner() is still old implementation`)
+    } else if (getAddress(owner) === getAddress(newImpl)) {
+      console.log(`GovernanceToken.owner() upgraded to new implementation!`)
+      break
+    } else {
+      throw new Error(
+        `GovernanceToken.owner() upgraded to unknown address ${owner}`
+      )
+    }
```
smartcontracts, line 85: Looks like a bug, `await sleep(5000)` should be inside this loop

smartcontracts (dismissed):

packages/contracts-governance/deploy/000-mint-manager-impl.ts
tynes: ah good catch

tynes:
@smartcontracts I've fixed the bug, thanks for catching it

smartcontracts (approved):

PR #3562, opened 2022-09-26
op-chain-ops: address aliasing helpers

tynes:
**Description**

Implement address aliasing helpers in go. Fuzz them to make sure that roundtrip aliasing is reliable. Add foundry fuzz tests that are used to generate test vectors.

Note that the JS implementation does not take into account the modular arithmetic, I do not believe that it is used anywhere important currently.
See:
https://github.com/ethereum-optimism/optimism/blob/develop/packages/core-utils/src/optimism/alias.ts

Perhaps we should move our differential fuzzing to use go instead of typescript or along with the typescript because most utilities that we wrote in JS are now also written in go.


protolambda (approved):

op-chain-ops/crossdomain/alias.go
```diff
@@ -0,0 +1,33 @@
...
+var (
+	uint160Max, _ = uint256.FromHex("0xffffffffffffffffffffffffffffffffffffffff")
+	offset        = new(uint256.Int).SetBytes(common.HexToAddress("0x1111000000000000000000000000000000001111").Bytes())
+)
+
+// ApplyL1ToL2Alias will apply the alias applied to L1 to L2 messages when it
+// originates from a contract address
+func ApplyL1ToL2Alias(address *common.Address) *common.Address {
+	input := new(uint256.Int).SetBytes(address.Bytes())
+	output := new(uint256.Int).AddMod(input, offset, uint160Max)
```
protolambda, line 18: The specs currently don't define if the addition is on a `uint256` (and clipped) or `uint160` (and modulo'd). I see it's the latter, maybe worth updating the specs? Not blocking the PR, but would be nice.

trianglesphere: I looked at the address aliasing contract and I think it needs to be 160 modulo math or you might lose information and might not be able to invert it.

trianglesphere: hmm, maybe this works with truncation. I need to think about this more to be confident either way.

trianglesphere (approved):

op-chain-ops/crossdomain/alias.go
protolambda: it's a power of 2, truncation/modulo should probably work the same. Curious if there's a cleaner, uint256 + clipping implementation now.

PR #3566, opened 2022-09-26
contracts-governance: make the governance token immutable

tynes:
**Description**

Slight improvement to the `MintManager`. Making
the reference to the `governanceToken` an immutable is just better than placing it in storage.


smartcontracts (approved):

packages/contracts-governance/contracts/MintManager.sol
```diff
@@ -15,7 +15,7 @@ contract MintManager is Ownable {
     /**
      * @notice The GovernanceToken that the MintManager can mint tokens
      */
-    GovernanceToken public governanceToken;
+    GovernanceToken public immutable governanceToken;
```
maurelian, line 18: I see the `mintPermittedAfter` is declared after this, so we probably need a spacer here?

tynes: No we don't - its not deployed behind a proxy. We need to call `upgrade()` on the existing mint manager with the address the new mint manager is deployed to

maurelian: Ah, gotcha.

maurelian (approved):

PR #3569, opened 2022-09-26
op-chain-ops: use PoS consensus in simulated backend in deployer and genesis

protolambda:
**Description**

This PR adds PoS consensus to the chain-ops package:
- For use in the simulated backend during deploying. See op-geth PR for more details. This will improve genesis preparation time.
- For use in the L1 genesis, to unblock action testing, since we can now build L1 blocks with more control (like a CL node) + latest L1 changes.

Depends on https://github.com/ethereum-optimism/op-geth/pull/8
So we'll need to merge that, and update the geth reference.

Fix ENG-2755


tynes:
Looks like an old version of geth is being referenced someplace? Undefined `backends.WithConsensus`

mslipper (approved):

mslipper:
LGTM. Geth PR is merged, so we can update that now.

trianglesphere (approved):

PR #3570, opened 2022-09-26
ctp: fixup deploy scripts for the nft bridge

tynes:
**Description**

The most important one is the deploy script for the `Proxy` that will end up at the predeploy on L2. There are specific checks for the correct deployer account being used. Will likely require some changes to the hardhat config when doing the actual deployment.

### DIFF
--- a/packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
+++ b/packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
@@ -1,13 +1,33 @@
 /* Imports: External */
 import { DeployFunction } from 'hardhat-deploy/dist/types'
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
 import '@eth-optimism/hardhat-deploy-config'
 import '@nomiclabs/hardhat-ethers'
 import 'hardhat-deploy'
+import fetch from 'node-fetch'
+
+// Handle the `ops` deployment
+const getL1CrossDomainMessengerProxyDeployment = async (
+  hre: HardhatRuntimeEnvironment
+) => {
+  const network = hre.network.name
+  if (network === 'ops-l1') {
+    const res = await fetch(
+      'http://localhost:8080/deployments/local/Proxy__OVM_L1CrossDomainMessenger.json'
+    )
+    return res.json()
+  } else {
+    return hre.deployments.get('Proxy__OVM_L1CrossDomainMessenger')
+  }
+}
 
 const deployFn: DeployFunction = async (hre) => {
   const { deployer } = await hre.getNamedAccounts()
   const { deploy } = hre.deployments
   const { getAddress } = hre.ethers.utils
+
+  console.log(`Deploying L1ERC721Bridge to ${hre.network.name}`)
+  console.log(`Using deployer ${deployer}`)
 
   const Deployment__L1ERC721BridgeProxy = await hre.deployments.get(
     'L1ERC721BridgeProxy'
@@ -17,18 +37,18 @@ const deployFn: DeployFunction = async (hre) => {
     'Proxy',
     Deployment__L1ERC721BridgeProxy.address
   )
-  const admin = await L1ERC721BridgeProxy.admin()
+  const admin = await L1ERC721BridgeProxy.callStatic.admin()
   if (getAddress(admin) !== getAddress(deployer)) {
     throw new Error('deployer is not proxy admin')
   }
 
   // Get the address of the currently deployed L1CrossDomainMessenger.
   // This should be the address of the proxy
-  const Deployment__L1CrossDomainMessenger = await hre.deployments.get(
-    'Proxy__OVM_L1CrossDomainMessenger'
-  )
+  const Deployment__L1CrossDomainMessengerProxy =
+    await getL1CrossDomainMessengerProxyDeployment(hre)
+
   const L1CrossDomainMessengerAddress =
-    Deployment__L1CrossDomainMessenger.address
+    Deployment__L1CrossDomainMessengerProxy.address
 
   const predeploy = '0x4200000000000000000000000000000000000014'
   // Deploy the L1ERC721Bridge. The arguments are
@@ -76,13 +96,16 @@ const deployFn: DeployFunction = async (hre) => {
   {
     if (
       hre.network.name === 'optimism' ||
-      hre.network.name === 'optimism-goerli'
+      hre.network.name === 'optimism-goerli' ||
+      hre.network.name === 'ops-l2'
     ) {
       let newAdmin: string
       if (hre.network.name === 'optimism') {
         newAdmin = '0x2501c477D0A35545a387Aa4A3EEe4292A9a8B3F0'
       } else if (hre.network.name === 'optimism-goerli') {
         newAdmin = '0xf80267194936da1E98dB10bcE06F3147D580a62e'
+      } else {
+        newAdmin = deployer
       }
       const tx = await L1ERC721BridgeProxy.changeAdmin(newAdmin)
       const receipt = await tx.wait()
--- a/packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
+++ b/packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
@@ -6,6 +6,13 @@ import '@nomiclabs/hardhat-ethers'
 import 'hardhat-deploy'
 import fetch from 'node-fetch'
 
+import {
+  isTargetL1Network,
+  predeploy,
+  getProxyAdmin,
+  validateERC721Bridge,
+} from '../../src/nft-bridge-deploy-helpers'
+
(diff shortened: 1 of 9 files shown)


tynes:
Going to be testing this against `ops`

tynes:
Tagging `do-not-merge` until i have it e2e working with `ops`

packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
```diff
@@ -1,15 +1,96 @@
...
+  const L1CrossDomainMessengerAddress =
+    Deployment__L1CrossDomainMessenger.address
+
+  const predeploy = '0x4200000000000000000000000000000000000014'
+  // Deploy the L1ERC721Bridge. The arguments are
+  // - messenger
+  // - otherBridge
+  // Since this is the L1ERC721Bridge, the otherBridge is the
+  // predeploy address
+  await deploy('L1ERC721Bridge', {
```
maurelian, line 39: I'd like it if we can postfix this with Impl to remove ambiguity.

packages/contracts-periphery/deploy/nft-bridge/OptimismMintableERC721FactoryImplementation.ts
```diff
@@ -1,15 +1,81 @@
...
+  }
+
+  {
+    if (
+      hre.network.name === 'optimism' ||
+      hre.network.name === 'optimism-goerli'
+    ) {
+      let newAdmin: string
+      if (hre.network.name === 'optimism') {
+        newAdmin = '0x2501c477D0A35545a387Aa4A3EEe4292A9a8B3F0'
```
maurelian, line 78: This feels error prone and hard to verify, could we put it in a well commented file similar to [predeploy.ts](https://github.com/ethereum-optimism/optimism/blob/0088ec6070d2e1eb7098d880a47d7abb52a96cf6/packages/contracts/src/predeploys.ts#L1)

packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
```diff
@@ -1,15 +1,119 @@
...
+
+  {
+    // Upgrade the Proxy to the newly deployed implementation
+    const tx = await L1ERC721BridgeProxy.upgradeTo(
+      Deployment__L1ERC721Bridge.address
+    )
+    const receipt = await tx.wait()
+    console.log(`L1ERC721BridgeProxy upgraded: ${receipt.transactionHash}`)
+  }
+
```
smartcontracts, line 95: Please also check the `messenger` and `otherBridge` addresses on the Proxy.

packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
```diff
@@ -1,15 +1,119 @@
...
+    if (
+      hre.network.name === 'optimism' ||
+      hre.network.name === 'optimism-goerli' ||
+      hre.network.name === 'ops-l2'
+    ) {
+      let newAdmin: string
+      if (hre.network.name === 'optimism') {
+        newAdmin = '0x2501c477D0A35545a387Aa4A3EEe4292A9a8B3F0'
+      } else if (hre.network.name === 'optimism-goerli') {
+        newAdmin = '0xf80267194936da1E98dB10bcE06F3147D580a62e'
```
smartcontracts, line 106: Where did you get this address?

packages/contracts-periphery/deploy/nft-bridge/L2ERC721BridgeImplementation.ts
```diff
@@ -1,18 +1,110 @@
...
+    // Upgrade the implementation of the proxy to the newly deployed
+    // L2ERC721Bridge
+    const tx = await L2ERC721BridgeProxy.upgradeTo(
+      Deployment__L2ERC721Bridge.address
+    )
+    const receipt = await tx.wait()
+    console.log(
+      `Upgraded the implementation of the L2ERC721BridgeProxy: ${receipt.transactionhash}`
+    )
+  }
```
smartcontracts, line 76: Same as above, please check the two conditions for the Proxy too

packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
```diff
@@ -1,15 +1,119 @@
...
+    // Upgrade the Proxy to the newly deployed implementation
+    const tx = await L1ERC721BridgeProxy.upgradeTo(
+      Deployment__L1ERC721Bridge.address
+    )
+    const receipt = await tx.wait()
+    console.log(`L1ERC721BridgeProxy upgraded: ${receipt.transactionHash}`)
+  }
+
+  {
+    if (
```
smartcontracts, line 97: Shouldn't this only be deployed to L1?

packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
```diff
@@ -1,15 +1,119 @@
...
+      hre.network.name === 'optimism' ||
+      hre.network.name === 'optimism-goerli' ||
+      hre.network.name === 'ops-l2'
+    ) {
+      let newAdmin: string
+      if (hre.network.name === 'optimism') {
+        newAdmin = '0x2501c477D0A35545a387Aa4A3EEe4292A9a8B3F0'
```
smartcontracts, line 104: I think this logic is slightly off. Should be `if (hre.network.name === 'mainnet') { // mainnet multisig }` and similar for goerli.

tynes: Ah good catch

packages/contracts-periphery/deploy/nft-bridge/L2ERC721BridgeImplementation.ts
```diff
@@ -1,18 +1,95 @@
...
+  console.log(`Using deployer ${deployer}`)
+
+  const L2ERC721BridgeProxy = await hre.ethers.getContractAt('Proxy', predeploy)
+
+  // Check to make sure that the admin of the proxy is the deployer.
+  // The deployer of the L2ERC721Bridge should be the same as the
+  // admin of the L2ERC721BridgeProxy so that it is easy to upgrade
+  // the implementation. The admin is then changed depending on the
+  // network after the L2ERC721BridgeProxy is upgraded to the implementation
+  const admin = await L2ERC721BridgeProxy.callStatic.admin()
```
tynes, line 34: if `callStatic` isn't used, it'll send a transaction

packages/contracts-periphery/scripts/deploy-nft-bridge.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
tynes, line 2: This script just works:

- run `make up` in `ops`
- run `./scripts/deploy-nft-bridge.sh`

Would like to sub out the networks and use this for the actual deployments

packages/contracts-periphery/src/nft-bridge-deploy-helpers.ts
```diff
@@ -0,0 +1,63 @@
+import { utils } from 'ethers'
+
```
tynes, line 2: TODO: include etherscan links to these

packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
tynes: this is the owner of the L1 contracts on goerli

packages/contracts-periphery/scripts/deploy-nft-bridge.sh
smartcontracts: You can use tags in the deployment scripts to make this simpler. I would add `L2ERC721BridgeDeployment` as a tag to the L2 contracts then `L1ERC721BridgeDeployment` as a tag to the L1 contracts. Then you can just do `npx hardhat deploy --tags L2ERC721BridgeDeployment` and it'll deploy everything with that tag, respecting dependencies.

smartcontracts: But honestly if this works, I don't care enough to make you edit this

smartcontracts (approved):
Ok I'm reasonably happy with this. We can clean it up later once we move it into predeploys.

packages/contracts-periphery/deploy/nft-bridge/L1ERC721BridgeImplementation.ts
tynes: We don't do this with `contracts-bedrock`, would prefer to be similar to that, then when using hardhat to do things like `hre.artifacts.readArtifact` you will need to know if it has the impl suffix or not

maurelian (approved):

PR #3571, opened 2022-09-26
feat(ctb): more defensive value check

smartcontracts:
**Description**
Moves the msg.value == amount check when bridging ETH deeper into the call stack. Has the effect of making the function less of a footgun for future developers.


tynes (approved):

maurelian (approved):
LGTM. I agree that having the check in the child contract was inconsistent and that we should stick to this pattern in the future.

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -418,6 +418,11 @@ abstract contract StandardBridge {
         uint32 _minGasLimit,
         bytes memory _extraData
     ) internal {
+        require(
+            msg.value == _amount,
+            "L2StandardBridge: ETH withdrawals must include sufficient ETH value"
```
maurelian, line 423:
```suggestion
            "StandardBridge: Bridging ETH must include sufficient ETH value"
```

maurelian (changes requested):

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
tynes: ah good catch

packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol
```diff
@@ -191,15 +191,11 @@ contract L2StandardBridge is StandardBridge, Semver {
     ) internal {
         address l1Token = OptimismMintableERC20(_l2Token).l1Token();
         if (_l2Token == Predeploys.LEGACY_ERC20_ETH) {
-            require(
-                msg.value == _amount,
-                "L2StandardBridge: ETH withdrawals must include sufficient ETH value"
-            );
-
             _initiateBridgeETH(_from, _to, _amount, _minGasLimit, _extraData);
```
maurelian, line 194: I think it would also be less error prone if this used `msg.value` instead of amount, as we do in the [L1 Bridge](https://github.com/ethereum-optimism/optimism/blob/d98811e8424db2e1c99337600d759e8b476c29ae/packages/contracts-bedrock/contracts/L1/L1StandardBridge.sol#L267)

smartcontracts: Doing this wouldn't give you the feedback that `_amount != msg.value` and would let you provide any arbitrary value

maurelian: Hmmm, good point... 
In the L1 bridge there is no `_amount` argument when sending ETH using the old interface. So this check helps to enforce agreement between these redundant values.

maurelian: That said, if we need to be checking agreement between `_amount` and `msg.value`, then we should have a similar check in `_initiateBridgeERC20` because there is a `payable` path to it via `L2StandardBridge.withdraw()` and `withdrawTo()`.

smartcontracts: Hmmm I guess that might help prevent some user error but I wouldn't really worry too much about it, all the tooling will use the new methods

smartcontracts: Actually yeah this is a good point. I will add the msg.value == 0 check as part of a second PR.

maurelian (approved):

PR #3573, opened 2022-09-27
ci: Use workspace rather than cache

mslipper:
Updates CircleCI to use workspaces rather than the build cache to pass around node modules and build artifacts. I've found this to save between 1-2 minutes on the yarn install job.


tynes (approved):
Hard to know with confidence that these changes work without seeing the CI jobs themselves triggered, but assuming you saw it working

protolambda (approved):

PR #3574, opened 2022-09-27
chore: Disable PR commenting by codecov bot

maurelian:
**Description**

Prevents the codecov bot from commenting on PRs. We may wish to enable this again later when the configuration is more mature.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,4 @@
+comment: false
 ignore:
   - "l2geth"
   - "**/*.t.sol"


tynes:
Were the comments that were being added not very useful?

tynes:
An example https://github.com/ethereum-optimism/optimism/pull/3535#issuecomment-1256840591

maurelian:
> Were the comments that were being added not very useful?

I don't think they're very useful yet. IMO there's going to be some observing and tweaking to be done in the codecov settings until we get it setup to be really valuable, so I don't want to irritate people with yet another bot until it's in a better state. 

> An example https://github.com/ethereum-optimism/optimism/pull/3535#issuecomment-1256840591

That's a good example actually, as the comment was posted before we started ignoring the `op-bindings` module in coverage reports.

Generally, I think that a bot still implies a fairly manual process. Just failing in CI if coverage decreases would more strongly reinforce the habit of adding tests for new code.

tynes (approved):
I'd like bot comments eventually, but ok with disabling them until its set up nicely

optimisticben (approved):
lgmt

mslipper (approved):

PR #3578, opened 2022-09-27
ci: Remove publish dependencies

mslipper:
The publish tasks don't need to depend on upstream tests. This will let us run Hive tests earlier in the pipeline, and shave off additional CI runtime.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,13 @@ jobs:
             - "packages/*/node_modules"
             - "packages/*/dist"
             - "packages/*/artifacts"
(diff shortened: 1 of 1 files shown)


maurelian (approved):

trianglesphere (approved):
How does the `deployer-bedrock` image get built now?

mslipper:
> How does the deployer-bedrock image get built now?

It's not currently used anywhere - we can resurrect it as the need arises.

mslipper:
I stand corrected - the regentool uses it. However, cloud build still creates that image so we're still covered as long as the builds are for commits to develop.

PR #3579, opened 2022-09-27
Feat: --rpc.evmtimeout

jhawk28:
**Description**

Add the ability to set the timeout for eth_call. Was set to only 5 seconds

**Tests**

**Additional context**

**Metadata**


tynes:
There is a lot of unnecessary diff in this PR like changing around the ordering of imports. Can you please remove all diffs besides the bare minimum?

l2geth/cmd/utils/flags.go
```diff
@@ -1660,6 +1665,9 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *eth.Config) {
 	if ctx.GlobalIsSet(RPCGlobalGasCap.Name) {
 		cfg.RPCGasCap = new(big.Int).SetUint64(ctx.GlobalUint64(RPCGlobalGasCap.Name))
 	}
+	if ctx.GlobalIsSet(RPCGlobalEVMTimeoutFlag.Name) {
```
tynes, line 1668: Is this line triggered always so that the default value is set? If not then it will change functionality, the default will be 0

l2geth/eth/api_backend.go
```diff
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"time"
```
tynes, line 24: Is this `time` import required?

l2geth/cmd/utils/flags.go
tynes: Ok I see the default value is set in another file

l2geth/internal/ethapi/api.go
```diff
@@ -659,10 +659,10 @@ func (s *PublicBlockChainAPI) GetHeaderByHash(ctx context.Context, hash common.H
 }
 
 // GetBlockByNumber returns the requested canonical block.
-// * When blockNr is -1 the chain head is returned.
-// * When blockNr is -2 the pending chain head is returned.
-// * When fullTx is true all transactions in the block are returned, otherwise
-//   only the transaction hash is returned.
+//   - When blockNr is -1 the chain head is returned.
```
tynes, line 662: This diff isn't required

tynes:
Can you add a changeset? `yarn changeset` at the root of the repo and then select l2geth and commit the changeset afterwards

l2geth/cmd/utils/flags.go
jhawk28: I tried to keep the pattern from Geth

l2geth/internal/ethapi/api.go
jhawk28: gofmt is changing this, I can manually fix it if you want

l2geth/eth/api_backend.go
jhawk28: `func (b *EthAPIBackend) RPCEVMTimeout() time.Duration`

tynes:
I think this PR is good, tagging @protolambda for a double check

l2geth/eth/api_backend.go
```diff
@@ -400,6 +401,10 @@ func (b *EthAPIBackend) RPCGasCap() *big.Int {
 	return b.eth.config.RPCGasCap
 }
 
+func (b *EthAPIBackend) RPCEVMTimeout() time.Duration {
```
protolambda, line 404: Adding the `0 == infinite` note to a doc comment here would be helpful

l2geth/eth/config.go
```diff
@@ -169,6 +170,9 @@ type Config struct {
 	// RPCGasCap is the global gas cap for eth-call variants.
 	RPCGasCap *big.Int `toml:",omitempty"`
 
+	// RPCEVMTimeout is the global timeout for eth-call.
+	RPCEVMTimeout time.Duration
```
protolambda, line 174: Same here

l2geth/les/api_backend.go
```diff
@@ -316,6 +317,10 @@ func (b *LesApiBackend) RPCGasCap() *big.Int {
 	return b.eth.config.RPCGasCap
 }
 
+func (b *LesApiBackend) RPCEVMTimeout() time.Duration {
```
protolambda, line 320: Same here

protolambda (approved):
Some doc comments about the `0 == infinite` case would be good, LGTM otherwise.

l2geth/les/api_backend.go
jhawk28: no docs are on any of the other methods do you still want one here?

l2geth/eth/api_backend.go
jhawk28: no docs are on any of the other methods do you still want one here?

jhawk28: I added the (0=infinite) to the docs for the interface

tynes:
One more thing and I can merge this, sorry I should have mentioned earlier but do you mind changing the changeset to be a patch release?

```
'@eth-optimism/l2geth': minor
```

Should be:

```
'@eth-optimism/l2geth': patch
```

tynes (approved):

tynes:
Thank you @jhawk28 !

mslipper:
Pushing this up so CI tests can run.

tynes:
Will need to trigger a release after this merges, cc @mslipper

PR #3581, opened 2022-09-27
ENG-2835 websocket monitoring daemon (endpoint-monitor)

zhwrd:
**Description**

Adds websocket monitoring daemon for edge-proxyd and downstream infra providers. This first iteration is pretty specific to one method: eth_subscribe. I opted to re-use the eth client instead of rolling my own websocket client since I understand we mainly only allow the subscribe method... Also in general, think we'd be better off using a vendor for most synthetic testing on our edge proxies since it will let us ping the endpoint from several regions without having to deploy there...

In a loop, runs eth_subscribe over websocket for each configured provider and reports metrics to prometheus on any errors encountered.

**Metadata**
- [ENG-2835](https://linear.app/optimism/issue/ENG-2835/create-websocket-monitoring-service-for-edge-proxyd-and-infra)


mslipper (approved):
Looks good. A couple nits:

- We generally use `urfave/cli` elsewhere in the codebase to do env var parsing/CLI management. It's nice because it lets you specify command line args as well. It also integrates with the `op-service` package, which contains generalized utilities for passing in metric configs and serving them.
- We also use `op-service` to create instances of the Prometheus metrics collector rather than serving a global. See https://github.com/ethereum-optimism/optimism/tree/develop/op-service/metrics for the package in question.
- We usually do manual signal handling so we can gracefully shut things down on `SIGINT`.

These aren't blocking, though, given the scope + goals of this project.

Lastly, this PR is under the size SLA when you discount the `go.sum` file.

zhwrd:
>

zhwrd:
> Looks good. A couple nits:
> 
> * We generally use `urfave/cli` elsewhere in the codebase to do env var parsing/CLI management. It's nice because it lets you specify command line args as well. It also integrates with the `op-service` package, which contains generalized utilities for passing in metric configs and serving them.
> * We also use `op-service` to create instances of the Prometheus metrics collector rather than serving a global. See https://github.com/ethereum-optimism/optimism/tree/develop/op-service/metrics for the package in question.
> * We usually do manual signal handling so we can gracefully shut things down on `SIGINT`.
> 
> These aren't blocking, though, given the scope + goals of this project.
> 
> Lastly, this PR is under the size SLA when you discount the `go.sum` file.

cool, will go ahead and adopt the op-service standard stuff, looks useful!

zhwrd:
Think this looks good now... Don't think i can gracefully shutdown the metrics server in this case because don't have access to the underlying Server from the op-server/metrics package but correct me if i'm wrong?

endpoint-monitor/config.go
```diff
@@ -0,0 +1,97 @@
...
+			EnvVar: opservice.PrefixEnvVar(envPrefix, "CHECK_DURATION"),
+		},
+	}
+	flags = append(flags, opmetrics.CLIFlags(envPrefix)...)
+	flags = append(flags, oplog.CLIFlags(envPrefix)...)
+	return flags
+}
+
+type Config struct {
+	Providers     []string      `envconfig:"PROVIDERS" required:"true"`
```
mslipper, line 53: Do these `envconfig` tags carry meaning anymore?

mslipper (approved):

PR #3583, opened 2022-09-27
op-node: Switch L1 Traversal to a pull based model

trianglesphere:
**Description**

The L1 Retrieval stage is now responsible for pulling data from the L1 Traversal stage. In addition, the pipeline is responsible for advancing the state of the L1 Traversal stage.

The L1 Traversal stage only provides access to the current L1 block once - it pretends to be a queue that is consumed from.

**Tests**

I had to modify some of the unit tests, but they are passing.

**Metadata**
- Fixes ENG-2747


trianglesphere:
The hive p2p test seems to be failing on this PR, but it is fixed in the next PR + is working on the final PR.

protolambda (approved):
LGTM, nice work. After all stage refactor PRs land we can revisit the pipeline function itself to get rid of the separate types of resets (old resets and new pull based resets).

mslipper (approved):

PR #3584, opened 2022-09-27
op-node: Switch L1 Retrieval to pull based

trianglesphere:
**Description**

This makes the L1 Retrieval a purely pull based stage. This commit required large modifications to the channel bank stage in order for the channel bank to maintain its own progress.

**Tests**

L1 Retrieval unit tests pass. I skipped the channel bank unit tests & fixed them up in the next PR (would have been lots of extra work to fix them twice).

**Metadata**
- Fixes ENG-2748

### DIFF
--- a/op-node/rollup/derive/channel_bank.go
+++ b/op-node/rollup/derive/channel_bank.go
@@ -38,18 +38,20 @@ type ChannelBank struct {
 	progress Progress
 
 	next ChannelBankOutput
+	prev *L1Retrieval
 }
 
 var _ Stage = (*ChannelBank)(nil)
 
 // NewChannelBank creates a ChannelBank, which should be Reset(origin) before use.
-func NewChannelBank(log log.Logger, cfg *rollup.Config, next ChannelBankOutput) *ChannelBank {
+func NewChannelBank(log log.Logger, cfg *rollup.Config, next ChannelBankOutput, prev *L1Retrieval) *ChannelBank {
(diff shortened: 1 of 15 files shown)


trianglesphere:
Tests are failing at the check changed script. I think it's because it's not pointing to develop as the base commit.

op-node/rollup/derive/channel_bank.go
```diff
@@ -38,18 +38,20 @@ type ChannelBank struct {
...
 
 	next ChannelBankOutput
+	prev *L1Retrieval
 }
 
 var _ Stage = (*ChannelBank)(nil)
 
 // NewChannelBank creates a ChannelBank, which should be Reset(origin) before use.
-func NewChannelBank(log log.Logger, cfg *rollup.Config, next ChannelBankOutput) *ChannelBank {
+func NewChannelBank(log log.Logger, cfg *rollup.Config, next ChannelBankOutput, prev *L1Retrieval) *ChannelBank {
```
protolambda, line 47: Will focus on the retrieval code for this PR, the channel bank can change later. But would be nice to separate each stage by interface instead of concrete types.

protolambda (approved):

op-node/rollup/derive/channel_bank.go
trianglesphere: fyi the channel bank takes an interface rather than a concrete type in the next PR

PR #3586, opened 2022-09-28
op-e2e: e2eutils package for new action testing setup

protolambda:
**Description**

I wrote the first iteration of these utils as part of the Hive testing, but then ported it over to use more of the chainops code from Mark and use it in the experimental action testing branch. Now the utils are in a separate package, so we can share them with op-e2e long term.

Action-testing itself will be a sub-package of the op-e2e module, and import these test utils.

**Tests**

Includes some basic tests to make sure the test setup itself behaves like expected.

**Additional context**

This test setup code can later also be used by op-e2e.

**Metadata**

Fix ENG-2836


op-e2e/e2eutils/setup.go
```diff
@@ -0,0 +1,228 @@
...
+		BatchSenderAddress:     addresses.Batcher,
+
+		L2OutputOracleSubmissionInterval: 6,
+		L2OutputOracleStartingTimestamp:  -1,
+		L2OutputOracleProposer:           addresses.Proposer,
+		L2OutputOracleOwner:              common.Address{}, // tbd
+
+		L1BlockTime:                 15,
+		L1GenesisBlockNonce:         0,
+		CliqueSignerAddress:         common.Address{}, // proof of stake, no clique
```
tynes, line 76: Is it worth it to just delete `CliqueSignerAddress`?

tynes (approved):
Looks good to me, good work. Being able to run tests against the genesis creation code on L2 is going to be amazing

op-e2e/e2eutils/setup.go
protolambda: That's part of the chainops, and other tests like hive may depend on that. We can remove it later, it's currently an optional field. And I like being explicit about not using clique, instead of omitting the field altogether for the default zero definition.

mslipper (approved):
This is great, thanks Proto. We can use a bunch of these in Hive too.

trianglesphere (commented):
Really nice work. I really like all the comments

PR #3587, opened 2022-09-28
e2eutils: collect addresses from test setup for easy fuzzing/test usage

protolambda:
**Description**

Simple function to get all addresses involved in the test setup, including their L1/L2 aliases, so we can use those for fuzzing.

**Tests**

Test that the function works and outputs at least one expected address.

**Additional context**

Working on fuzz testing as part of action-testing.

**Metadata**

Fix ENG-2837

Draft PR, depends on action tests setup


op-e2e/e2eutils/addresses.go
```diff
@@ -0,0 +1,65 @@
...
+	out = append(out,
+		sd.L1Cfg.Coinbase,
+		sd.L2Cfg.Coinbase,
+		sd.RollupCfg.P2PSequencerAddress,
+		sd.RollupCfg.FeeRecipientAddress,
+		sd.RollupCfg.BatchInboxAddress,
+		sd.RollupCfg.BatchSenderAddress,
+		sd.RollupCfg.DepositContractAddress,
+	)
+	//  - precompiles
```
tynes, line 49: Perhaps also predeploys?

op-e2e/e2eutils/addresses.go
```diff
@@ -0,0 +1,65 @@
...
+	for i := 0; i <= 0xff; i++ {
+		out = append(out, common.Address{19: byte(i)})
+	}
+	//  - masked L2 version of all the original addrs
+	original := out[:]
+	for _, addr := range original {
+		masked := *crossdomain.ApplyL1ToL2Alias(&addr)
```
tynes, line 56: Do you think we should make this function not return a pointer?

op-e2e/e2eutils/addresses.go
protolambda: Those are part of genesis alloc

op-e2e/e2eutils/addresses.go
protolambda: Yes

protolambda:
Hmm, squash merging of the base PR caused this to need a rebase. So we can't really stack small PRs nicely without breaking line count. cc @mslipper

mslipper:
I'll update the Mergify rules.

op-e2e/e2eutils/addresses.go
```diff
@@ -0,0 +1,65 @@
...
+	})
+	return out
+}
+
+// CollectAddresses constructs a lists of addresses that may be used as fuzzing corpora
+// or random address selection.
+func CollectAddresses(sd *SetupData, dp *DeployParams) (out []common.Address) {
+	// This should be seeded with:
+	//  - reserve 0 for selecting nil (contract creation)
+	out = append(out, common.Address{})
```
trianglesphere, line 30: why put in the empty address twice?

op-e2e/e2eutils/addresses.go
trianglesphere: there's arguably some reason for the function to take a pointer, but it might be easier to switch both the input + output to be values.

protolambda: Opened a PR for this: https://github.com/ethereum-optimism/optimism/pull/3617

op-e2e/e2eutils/addresses.go
protolambda: I'm thinking we reserve index 0 to be "nil", to handle the contract creation case (no destination address), while the other is really zero.

trianglesphere: ok. Does this imply that this is going to be migrated to `[]*common.Address` as the return type.

protolambda: Just `[]common.Address`, no pointer. Only when used as destination address it's useful to make "none" a part of the options. Having a duplicate zeroed as fallback is fine. Or we can remove the `nil` case altogether, and fuzz that separately.

trianglesphere: so calling code that selects an address from this list would turn the zero address into a nil pointer if the index it selected was 0?

protolambda: that was the idea yes. Not perfect, but that would make it easy to select the nil case as part of all possible addresses when creating a fuzzer.

protolambda:
rebased on develop, and using the new non-pointer alias functions now

op-e2e/e2eutils/addresses.go
trianglesphere: Ok. That's a little non intuitive, but it won't cause problems. It's a good reminder for when fuzzing gets implemented that if a function takes a pointer, we should be passing `nil`.

trianglesphere (approved):

tynes (approved):

PR #3588, opened 2022-09-28
op-e2e/actions: action tests base test util

protolambda:
**Description**

Adds core types / base stuff for action testing. Mostly just types and testing wrapper.

Depends on #3586 

**Metadata**

Fix ENG-2834

### DIFF
--- a/op-e2e/actions/action.go
+++ b/op-e2e/actions/action.go
@@ -27,8 +27,12 @@ type Action func(t Testing)
 type ActionStatus uint
 
 const (
+	// ActionOK indicates the action is valid to apply
 	ActionOK ActionStatus = iota
+	// ActionInvalid indicates the action is not applicable, and a different next action may taken.
 	ActionInvalid
+	// More action status types may be used to indicate e.g. required rewinds,
+	// simple skips, or special cases for fuzzing.
 )
 
 // defaultTesting is a simple implementation of Testing that takes standard Go testing framework,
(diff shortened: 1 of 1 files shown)


op-e2e/actions/action.go
```diff
@@ -0,0 +1,62 @@
...
+	InvalidAction(format string, args ...any)
+}
+
+// Action is a function that may change the state of one or more actors or check their state.
+// Action definitions are meant to be very small building blocks,
+// and then composed into larger patterns to write more elaborate tests.
+type Action func(t Testing)
+
+// ActionStatus defines the state of an action, to make a basic distinction between InvalidAction() and other calls.
+type ActionStatus uint
```
trianglesphere, line 27: Will there be more `ActionStatus` types than just `Ok` or `Invalid`

op-e2e/actions/action.go
```diff
@@ -0,0 +1,62 @@
...
+	e2eutils.TestingBase
+	ctx   context.Context
+	state ActionStatus
+}
+
+func (st *defaultTesting) Ctx() context.Context {
+	return st.ctx
+}
+
+func (st *defaultTesting) InvalidAction(format string, args ...any) {
```
trianglesphere, line 47: Needs a docstring on usage (`format` + args)

op-e2e/actions/action.go
protolambda: I've a bunch of ideas; rewinding an action, skipping, or doing other special things for fuzzing. For now this is all I need though.

op-e2e/actions/action.go
protolambda: Added more docstrings

trianglesphere: Specifically here it'd be helpful to describe that format gets passed to Golang's fmt and must match the provided args. When directly working with fmt.Printf/Errorf there's good tooling to make sure that the args match the format string, but I'm not sure what will happen when there's a layer of indirection.

protolambda: It behaves the same, if it's wrong it will also behave the same. Tooling might not pick up on it, but I don't see a better way of doing this while also tracking invalid calls for fuzzing purposes.

trianglesphere (approved):

mslipper (approved):

PR #3590, opened 2022-09-28
contracts-bedrock: cleanup imports, fix storage layout

tynes:
**Description**

We don't need to use the upgradable initializable in the `OptimismPortal`, also fix the sizes of the storage layout spacers so that they are in sets of 50


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -63,6 +63,11 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
      */
     mapping(bytes32 => bool) public finalizedWithdrawals;
 
+    /**
+     * @notice Reserve extra slots (to a total of 50) in the storage layout for future upgrades.
+     */
+    uint256[48] private __gap;
```
maurelian, line 69: I previously took this out, I don't think we need them in our concrete contracts which are not inherited by anything in production. If we put this one here, then we should have one in the `L2OutputOracle` too.

It's fine either way, but a bit cleaner not to IMO.

tynes: Ok i'm down to remove it

tynes: I've addressed your comment @maurelian

tynes:
Looking for review @maurelian @smartcontracts

maurelian (approved):

tynes:
This needs a rebase

tynes:
I've rebased this, can i get a review @smartcontracts ?

smartcontracts (approved):

PR #3596, opened 2022-09-28
op-node: Switch channel bank and channel in reader to a pull based stage

trianglesphere:
**Description**

This is two commits because the intermediate commit was failing a test & the second commit fixes it.

* op-node: Switch channel bank to be pull based
    
    This again requires a fair amount of changes to channel_in_reader.go
    for the channel in reader to maintain its progress state.

* op-node: Switch channel in reader to a pull based stage
    Like the rest of the changes, this also required modifications to
    the next stage - the batch queue in order for it to manage the
    progress API. This commit required even more changes than usual.
    I changed the pipeline to be reset to a common starting point
    and now use the L2SafeHead block to filter out adding batches &
    L1 blocks to the batch queue.

**Metadata**

- Fixes ENG-2749

### DIFF
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -32,12 +32,18 @@ type BatchQueueOutput interface {
 	SafeL2Head() eth.L2BlockRef
 }
 
+type NextBatchProvider interface {
+	Origin() eth.L1BlockRef
+	NextBatch(ctx context.Context) (*BatchData, error)
+}
+
 // BatchQueue contains a set of batches for every L1 block.
 // L1 blocks are contiguous and this does not support reorgs.
 type BatchQueue struct {
 	log      log.Logger
 	config   *rollup.Config
 	next     BatchQueueOutput
+	prev     NextBatchProvider
 	progress Progress
 
 	l1Blocks []eth.L1BlockRef
@@ -47,11 +53,12 @@ type BatchQueue struct {
 }
 
 // NewBatchQueue creates a BatchQueue, which should be Reset(origin) before use.
-func NewBatchQueue(log log.Logger, cfg *rollup.Config, next BatchQueueOutput) *BatchQueue {
+func NewBatchQueue(log log.Logger, cfg *rollup.Config, next BatchQueueOutput, prev NextBatchProvider) *BatchQueue {
 	return &BatchQueue{
 		log:    log,
 		config: cfg,
 		next:   next,
+		prev:   prev,
 	}
 }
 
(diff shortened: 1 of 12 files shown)


op-node/rollup/driver/state.go
```diff
@@ -440,6 +440,10 @@ func (s *state) eventLoop() {
 			} else if err != nil && errors.Is(err, derive.ErrCritical) {
 				s.log.Error("Derivation process critical error", "err", err)
 				return
+			} else if err != nil && errors.Is(err, derive.NotEnoughData) {
+				stepAttempts = 0 // don't do a backoff for this error
+				reqStep()
+				continue
```
protolambda, line 446: This is quite subtle, but preserves one of the best things from the old design I think was most undervalued in redesign work.

While the pipeline is deriving, a stage can choose to stop and return this error, handing back control to the caller (driver). This keeps the derivation pipeline nice and synchronous for Cannon usage, while handling cases where there's a slow background fetching thing or something requiring repeated smaller updates (like here with the channels). And thus we can put a short and strict timeout on the pipeline work, while keeping the whole thing snappy and simple.

The alternative would be to make the pipeline blocking and essentially halt until we pull either a batch or an EOF, but then it would hold the Engine control captive unless we decouple that further (and deal with queuing and async stuff there without breaking Cannon compatibility), which I think isn't worth it (but maybe with a good design can be done eventually at a later point, since it's not a breaking protocol change).

cc @norswap this is related to thing we started to discuss.

op-node/rollup/derive/channel_bank.go
```diff
@@ -143,73 +138,37 @@ func (ib *ChannelBank) Read() (data []byte, err error) {
...
+// consistency around channel bank pruning which depends upon the order
+// of operations.
+func (ib *ChannelBank) NextData(ctx context.Context) ([]byte, error) {
 
-	if data, err := ib.prev.NextData(ctx); err == io.EOF {
-		outOfData = true
+	// Do the read from the channel bank first
+	data, err := ib.Read()
+	if err == io.EOF {
+		// continue - We will attempt to load data into the channel bank
```
protolambda, line 151: maybe the below code can be nested here? I'm not sure what's best; flat code or easier coverage checking.

op-node/rollup/derive/channel_bank.go
```diff
@@ -78,10 +75,8 @@ func (ib *ChannelBank) prune() {
 // IngestData adds new L1 data to the channel bank.
 // Read() should be called repeatedly first, until everything has been read, before adding new data.\
 func (ib *ChannelBank) IngestData(data []byte) {
```
protolambda, line 77: The `ib` receiver name is legacy, if you're refactoring it anyway we can rename it to `cb` or something. Optional change

op-node/rollup/derive/batch_queue.go
```diff
@@ -60,19 +67,58 @@ func (bq *BatchQueue) Progress() Progress {
 }
 
 func (bq *BatchQueue) Step(ctx context.Context, outer Progress) error {
-	if changed, err := bq.progress.Update(outer); err != nil {
-		return err
-	} else if changed {
-		if !bq.progress.Closed { // init inputs if we moved to a new open origin
+
+	originBehind := bq.progress.Origin.Number < bq.next.SafeL2Head().L1Origin.Number
```
protolambda, line 71: Assuming the pipeline reset logic is walking back the L1 header chain starting from the L2 safe head and checking all L1 origins parent hashes this should be safe, but it wouldn't hurt to do a consistency check to make sure the safe head data is consistent with the L1 origin when we need it to, to avoid weird reorg edge cases (I hope to capture these soon in the action testing). Outside of the scope of this PR if we focus on channel bank + channel reader here, but worth looking at.

protolambda (approved):
LGTM. Left some comments about the interesting bits (to me), but nothing blocking merge, nice work :+1:

op-node/rollup/derive/channel_bank.go
trianglesphere: done

op-node/rollup/derive/channel_bank.go
trianglesphere: it can be. Nice catch. I might do this cleanup later, but it's a really good callout

op-node/rollup/driver/state.go
trianglesphere: This error was mainly introduced to simplify some of the code (so it doesn't need to do a loop inside a function). Honestly, I'm really skeptical about the benefit of yielding back to the caller - it tends to introduce a lot of extra internal state to be able to resume and I'm not sure what the benefit of this is. Having a block pipeline is really simple and easy to understand. In addition, because the fetcher is just an interface, it's really easy to sub out for cannon.

protolambda: The pipeline blocking means we can't interact with the engine from the driver without knowing if we interfere with the engine state or if things change while we use the engine, handing back control to the caller to not lock the engine is important IMHO. Agreed on the blocking being easier, but I don't believe we should block on something too long when it wraps a resource as important as the engine.

mslipper (approved):

PR #3598, opened 2022-09-28
op-node: Switch attributes queue to be pull based

trianglesphere:
**Description**

The progress API is very nearly removed from the engine queue stage.

**Metadata**
- Fixes ENG-2751

### DIFF
--- a/op-node/rollup/derive/attributes_queue.go
+++ b/op-node/rollup/derive/attributes_queue.go
@@ -22,75 +22,60 @@ import (
 // This stage can be reset by clearing it's batch buffer.
 // This stage does not need to retain any references to L1 blocks.
 
-type AttributesQueueOutput interface {
-	AddSafeAttributes(attributes *eth.PayloadAttributes)
-	SafeL2Head() eth.L2BlockRef
-	StageProgress
-}
-
 type AttributesQueue struct {
-	log      log.Logger
-	config   *rollup.Config
-	dl       L1ReceiptsFetcher
-	next     AttributesQueueOutput
-	prev     *BatchQueue
-	progress Progress
-	batches  []*BatchData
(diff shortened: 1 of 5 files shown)


op-node/rollup/derive/attributes_queue.go
```diff
@@ -22,75 +22,60 @@ import (
...
+}
+
+// createNextAttributes transforms a batch into a payload attributes. This sets `NoTxPool` and appends the batched transactions
+// to the attributes transaction list
+func (aq *AttributesQueue) createNextAttributes(ctx context.Context, batch *BatchData, l2SafeHead eth.L2BlockRef) (*eth.PayloadAttributes, error) {
 	// sanity check parent hash
-	if batch.ParentHash != safeL2Head.Hash {
-		return NewCriticalError(fmt.Errorf("valid batch has bad parent hash %s, expected %s", batch.ParentHash, safeL2Head.Hash))
+	if batch.ParentHash != l2SafeHead.Hash {
+		return nil, NewCriticalError(fmt.Errorf("valid batch has bad parent hash %s, expected %s", batch.ParentHash, l2SafeHead.Hash))
```
protolambda, line 72: I think we can downgrade this to a reset error, since that should resolve it, if it ever goes wrong? I'm worried that if we don't properly check all reorg edge cases when determining the safe head that we might hit this (even though we should never). Working on testing too.

protolambda (approved):
LGTM. One nit about an error case we may want to adjust

op-node/rollup/derive/attributes_queue.go
trianglesphere: yea, a reset error might fix this, I'll downgrade it

mslipper (approved):

PR #3605, opened 2022-09-29
op-node: ParseWithdrawalInitiated improvements

mslipper:
`ParseWithdrawalInitiated` used to pass every log to `abi.EventByID`. Since `abi.EventByID` was bound to the `L2ToL1MessagePasser` contract, any logs on the receipt from a different contract would cause this function to return an error. This breaks support for parsing withdrawal events initiated by the `L2StandardBridge`, since the bridge adds many events of its own. This PR updates `ParseWithdrawalInitiated` to use the hex-encoded withdrawal intiated topic to determine which log messages to use instead.


protolambda (commented):
Not immediately sure if we can even have a log without it, but checking topic 0 exists doesn't hurt. And added a suggestion about topic readability.

op-node/withdrawals/utils.go
```diff
@@ -21,6 +21,9 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
+var WithdrawalInitiatedTopic = common.HexToHash("0x87bf7b546c8de873abb0db5b579ec131f8d0cf5b14f39933551cf9ced23a6136")
+var WithdrawalInitiatedExtension1Topic = common.HexToHash("0x2ef6ceb1668fdd882b1f89ddd53a666b0c1113d14cf90c0fbf97c7b1ad880fbb")
```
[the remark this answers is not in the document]
tynes: I wish that these hashes were easily accessible via the generated bindings

op-node/withdrawals/utils.go
```diff
@@ -264,23 +267,17 @@ func ParseWithdrawalInitiated(receipt *types.Receipt) (*bindings.L2ToL1MessagePa
...
 		return nil, err
 	}
-	abi, err := bindings.L2ToL1MessagePasserMetaData.GetAbi()
-	if err != nil {
-		return nil, err
-	}
 
 	for _, log := range receipt.Logs {
-		event, err := abi.EventByID(log.Topics[0])
-		if err != nil {
```
tynes, line 274: I handled this problem like this:

https://github.com/ethereum-optimism/optimism/blob/5a5e6ef15ee272af99da20e1b3a5716c97a26dee/op-chain-ops/crossdomain/legacy_withdrawal_test.go#L216-L218

Where when the abi didn't know about the event, it would return `nil` for the event. 

Its unlikely that the event will change so i think its fine to hardcode the topic

mslipper: This suppresses the error returned by `abi.EventByID`. It'll return an `event not found` error when it encounters a topic it doesn't know about.

op-node/withdrawals/utils.go
```diff
@@ -264,23 +267,17 @@ func ParseWithdrawalInitiated(receipt *types.Receipt) (*bindings.L2ToL1MessagePa
 	if err != nil {
```
trianglesphere, line 267: two lines above needs a doc comment (though it was probably me that forgot to write it in the first place).

op-node/withdrawals/utils.go
```diff
@@ -264,23 +267,17 @@ func ParseWithdrawalInitiated(receipt *types.Receipt) (*bindings.L2ToL1MessagePa
...
 		return nil, err
 	}
-	abi, err := bindings.L2ToL1MessagePasserMetaData.GetAbi()
-	if err != nil {
-		return nil, err
-	}
 
 	for _, log := range receipt.Logs {
```
trianglesphere, line 271: this still only parses the first withdrawal that is done (& would second multiple withdrawals if that is possible to do in a single transaction). Probably fine for now, but good to keep track of.

op-node/withdrawals/utils.go
mslipper: Done.

op-node/withdrawals/utils.go
mslipper: Added in docstring, and tracking as ENG-2852.

trianglesphere (approved):

trianglesphere:
devnet test failure is just a 503

mslipper:
Will re-run and let Mergify handle it. Will also incorporate your suggestions above re: Bigints.

mslipper:
@mergifyio refresh

PR #3611, opened 2022-09-29
Initial refactor of L1StdBridge test suite

maurelian:
**Description**

Breaks the test suite up into distinct test contracts. Each state changing
method has its own contract for happy tests and another for sad
tests.

This is a 'weather baloon' PR with a minimal diff, affecting on the L1 Bridge.
Once accepted I'll extend the changes to the rest of the forge tests.

**Tests**

I added one test case to cover getters. Otherwise the only changes are renaming test methods, and breaking them up into a smaller contracts.

Having both a happy and sad test contract for each method makes it much easier to see where gaps in coverage are.

**Additional context**

Naming conventions were heavily inspired by [Maple Finance's suite](https://github.com/maple-labs/revenue-distribution-token/blob/be9592fd72bfa7142a217507f2d5500a7856329e/contracts/test/RevenueDistributionToken.t.sol#L1).


packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -10,14 +10,17 @@ import { AddressAliasHelper } from "../vendor/AddressAliasHelper.sol";
 import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import { stdStorage, StdStorage } from "forge-std/Test.sol";
 
-contract L1StandardBridge_Test is Bridge_Initializer {
-    using stdStorage for StdStorage;
-
-    function setUp() public override {
-        super.setUp();
+contract L1StandardBridge_Getter_Test is Bridge_Initializer {
```
smartcontracts, line 13: Does `--match` account for contract names or just function names? This format might make it annoying to test individual functions. We could work around this by having the test names still include the contract name.

maurelian: Just function names for `--match`. Though there are a bunch of variants too that you can use:

```
--match-test <REGEX>                   Only run test functions matching the specified regex pattern [aliases: mt]
--match-contract <REGEX>               Only run tests in contracts matching the         --match-path <GLOB>                    Only run tests in source files matching the specified glob pattern [aliases: mp]
```

And you can combine them, ie. this runs all L1 Bridge tests with the word 'deposit' in the name: `ft --mc L1StandardBridge --mt deposit`. 

Given those options, I think the current style is cleaner.

tynes: Using `--match` is deprecated, shoudl use `--match-test` and `--match-contract`

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -33,10 +36,14 @@ contract L1StandardBridge_Test is Bridge_Initializer {
             Predeploys.L2_STANDARD_BRIDGE
         );
     }
+}
+
+contract L1StandardBridge_Initialize_TestFail is Bridge_Initializer {}
```
tynes, line 41: This test should be filled out

tynes:
So the empty contracts, the idea is to fill them out later? I notice there is multiple

tynes (approved):

maurelian:
> So the empty contracts, the idea is to fill them out later? I notice there is multiple

Yeah, this is more of a PoC, trying to keep the diff small to get approval before I embark on this journey.
The number of empty contracts is pretty interesting in and of itself tho.

smartcontracts (approved):

maurelian:
@mergifyio refresh

PR #3610, opened 2022-09-29
op-node: move derivation test metrics to test utils package

protolambda:
**Description**

Move derivation test metrics to test utils package, so they can be used in action testing.

PR #3612, opened 2022-09-29
op-e2e: Action testing L2 engine (without block building)

protolambda:
**Description**

Implements the L2 Engine as a test actor. The L2 rollup node would be the actor that starts/ends block building, sharing the codepaths with the rollup driver, as well as syncing from L1 etc. The engine just serves the Engine API, but instrumented in a way where we can manipulate the block building process. The goal is not to test the Engine API, but rather to expose the L2 block building in a synchronous instrumented way for testing of the rollup  node.

This excludes the block building part, which I'm separating into a smaller separate PR.

Depends on #3610 and #3607 
 
**Tests**

This test actor is probably the most complex of them all, since it implements the Geth Engine API backend in a more instrumented way. If the geth separation of API view and model was better we could reuse more code there.
To test the test-utils, I added a unit test that does the important things like block insertion and a reorg.

**Metadata**

Fix ENG-2764


trianglesphere:
I think this PR needs a rebase / double check on the base branch

protolambda:
Rebased

op-e2e/actions/l2_engine_api.go
```diff
@@ -0,0 +1,215 @@
...
+	if state.HeadBlockHash == (common.Hash{}) {
+		ea.log.Warn("Forkchoice requested update to zero hash")
+		return STATUS_INVALID, nil
+	}
+	// Check whether we have the block yet in our database or not. If not, we'll
+	// need to either trigger a sync, or to reject this forkchoice update for a
+	// reason.
+	block := ea.l2Chain.GetBlockByHash(state.HeadBlockHash)
+	if block == nil {
+		// TODO: syncing not supported yet
```
trianglesphere, line 62: should we panic / return another error here?

trianglesphere (approved):

op-e2e/actions/l1_miner.go
```diff
@@ -0,0 +1,142 @@
...
+		s.l1BuildingState = statedb
+		s.l1Receipts = make([]*types.Receipt, 0)
+		s.l1Transactions = make([]*types.Transaction, 0)
+		s.pendingIndices = make(map[common.Address]uint64)
+
+		s.l1GasPool = new(core.GasPool).AddGas(header.GasLimit)
+	}
+}
+
+// ActL1IncludeTx includes the next tx from L1 tx pool from the given account
```
tynes, line 85: This includes a single tx from a particular sender that is in the mempool. Is `TxPool().ContentFrom` deterministic?

op-e2e/actions/l1_miner.go
```diff
@@ -0,0 +1,142 @@
...
+	l1GasPool        *core.GasPool             // track gas used of ongoing building
+	pendingIndices   map[common.Address]uint64 // per account, how many txs from the pool were already included in the block, since the pool is lagging behind block mining.
+	l1Transactions   []*types.Transaction      // collects txs that were successfully included into current block build
+	l1Receipts       []*types.Receipt          // collect receipts of ongoing building
+	l1Building       bool
+	l1TxFailed       []*types.Transaction // log of failed transactions which could not be included
+}
+
+// NewL1Miner creates a new L1Replica that can also build blocks.
+func NewL1Miner(log log.Logger, genesis *core.Genesis) *L1Miner {
```
tynes, line 32: Not super opinionated but should this be "block producer"?

op-e2e/actions/l1_replica.go
```diff
@@ -0,0 +1,188 @@
...
+// The other node always has the canonical chain.
+// May be nil if there is nothing to sync from
+type L1CanonSrc func(num uint64) *types.Block
+
+// L1Replica is an instrumented in-memory L1 geth node that:
+// - can sync from the given canonical L1 blocks source
+// - can rewind the chain back (for reorgs)
+// - can provide an RPC with mock errors
+type L1Replica struct {
+	log log.Logger
```
tynes, line 33: Does the miner need to be different than the `L1Replica`? It feels like you could make an `L1Node` that has a miner and the `L1Replica` logic

tynes: Ah nevermind, the miner embeds `L1Node`

op-e2e/actions/l1_miner.go
protolambda: The Miner name was used in the code before the Merge, haha :sweat_smile: . It's also making them canonical, so maybe it's just L1Sequencer ? I still like the distinction with the L2 naming.

tynes:
Is this PR waiting for this to be implemented?

```
	// TODO proto - block building PR
```

protolambda:
@tynes see #3613 that adds that functionality. I isolated it because this PR would otherwise be too large.

op-e2e/actions/l1_miner.go
protolambda: The mempool order per account is deterministic, it keeps the txs with the lowest nonce value around, in order. And separates pending txs from queued ones (when it can't be processed due to nonce or fee etc.)

op-e2e/actions/l1_miner.go
protolambda: Think I'll leave it as is-is for now and avoid rebasing other in-flight work for the rename. But happy to rename later if we have a better name for this actor.

protolambda:
Will rebase to clean up the diff

op-e2e/actions/l1_miner.go
protolambda: Also, this is part of the other action testing which was merged in develop, but the mergify squashing caused this stacked PR to contain old commits.

protolambda:
Ready for another review @tynes

tynes (approved):

mslipper:
Will make a merge commit to avoid conflicts.

PR #3613, opened 2022-09-29
op-e2e: Action testing L2 Engine Block building

protolambda:
**Description**

This adds block building functionality to the L2 engine actor in action testing.

This will help us write specific test cases where we include certain L2 transactions in specific order, while also running the driver and derivation pipeline (when L2 verifier actor is implemented).

The actual action tests will be less verbose. The testing here is verbose because it tests the implementation of the test utils. If that's tested, we can compose and repeat things.

Depends on #3612 

**Tests**

Test that the L2 engine can build blocks, and that we can instrument the tx inclusion.

**Metadata**

Fix ENG-2850


mslipper (approved):

trianglesphere (approved):

protolambda:
Rebased to clean up the diff. The target branch is another PR (#3612) which is about ready. We should merge that first (can use mergify there). I'll then rebase this once more, and then mergify can pick this one up after I remove the label.

tynes (approved):

PR #3614, opened 2022-09-29
Changsets: nag on .sol but not t.sol files

maurelian:
**Description**

Changes Mergify's changeset behavior to:
1. nag when a `.sol` file is changed (it seems this was missing until now?)
3. not nag when a `.t.sol` file is changed

I tested the changes here:
- [before](https://regex101.com/r/WT6isQ/1) (note that the `h.sol` file is not matched).
- [after](https://regex101.com/r/CdvVB2/1)


tynes (approved):

mslipper (approved):

maurelian:
@mergifyio refresh

PR #3615, opened 2022-09-29
contracts-periphery: erc721 factory fixes

tynes:
**Description**

These changes make the `OptimismMintableERC721Factory` in line with the bedrock version of the `OptimismMintableERC20Factory`. We might as well follow what we are doing with the new erc20 factory to keep things in line with each other.

These changes are very small, they just rename a function to not follow the deprecated naming scheme, the function now returns an address which makes testing/integration a bit easier and also does a slight refactor to prevent the need to cast a type 3 times.

### DIFF
--- a/packages/contracts-periphery/contracts/universal/op-erc721/OptimismMintableERC721Factory.sol
+++ b/packages/contracts-periphery/contracts/universal/op-erc721/OptimismMintableERC721Factory.sol
@@ -14,8 +14,13 @@ contract OptimismMintableERC721Factory is Semver {
      *
      * @param localToken  Address of the token on the this domain.
      * @param remoteToken Address of the token on the remote domain.
+     * @param deployer    Address of the initiator of the deployment
      */
-    event OptimismMintableERC721Created(address indexed localToken, address indexed remoteToken);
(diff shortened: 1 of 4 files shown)


sam-goldman (approved):

packages/contracts-periphery/contracts/universal/op-erc721/OptimismMintableERC721Factory.sol
```diff
@@ -14,8 +14,13 @@ contract OptimismMintableERC721Factory is Semver {
      *
      * @param localToken  Address of the token on the this domain.
      * @param remoteToken Address of the token on the remote domain.
+     * @param deployer    Address of the initiator of the deployment
      */
-    event OptimismMintableERC721Created(address indexed localToken, address indexed remoteToken);
+    event OptimismMintableERC721Created(
+        address indexed localToken,
+        address indexed remoteToken,
+        address deployer
```
maurelian, line 22: Comment only: It feels like this should've been indexed in the ERC20 factory, but better to keep consistent for now.

maurelian (approved):
this LGTM

smartcontracts (approved):

PR #3628, opened 2022-09-30
op-e2e: action testing user

protolambda:
**Description**

This PR introduces:
- `L1Bindings` and `L2Bindings`: contracts backed by an ethclient and other typed client bindings for users.
- a `BasicUserEnv` which holds common user environment properties (i.e. can be shared between users). This is using generics, so we can embed either L1 or L2 bindings.
- a `BasicUser` actor that can make txs and randomize their tx input choices with actions.
- a `CrossLayerUser` actor that has a L1 and L2 `BasicUser` side and combines them to enable deposits and later withdrawal actions (see #3670).

Some of the actions to randomize and create tx contents are still a bit unpolished; we can fix that when when we use the User actor in more tests.

This also fixes minor bugs in action-testing:
- The L2 engine block building wasn't preparing the state db for the tx, causing receipts to have no event logs.
- The L1 miner had the same problem as the above.
- The sequencer state was resetting the payload ID to zero value in the outer sequencer function, instead of the complete building function, causing it to be missed when not calling the outer function.

**Tests**

Create `alice` and make all types of transactions in very few lines of code: these actions can be reordered/repeated to build different L1/L2 chains (and we can do so with more users too).

**Additional context**

Some of the action functions are curried so we can make an action with the same parameters that is repeatable.

**Metadata**

Fix ENG-2765


mslipper (approved):

trianglesphere (changes requested):
This is a cool framework, but having more comments would be really helpful for understanding the actions.

op-e2e/actions/user.go
```diff
@@ -0,0 +1,322 @@
...
+	if s.L2.txToAddr == nil {
+		isCreation = true
+	} else {
+		toAddr = *s.L2.txToAddr
+	}
+	depositTransferValue := s.L2.TxValue()
+	depositGas := s.L2.txOpts.GasLimit
+	if s.L2.txOpts.GasLimit == 0 {
+		// estimate gas used by deposit
+		gas, err := s.L2.env.EthCl.EstimateGas(t.Ctx(), ethereum.CallMsg{
```
tynes, line 281: Perhaps optionally add a buffer, this is known to produce faulty gas estimation

op-e2e/actions/user.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// Send the actual tx (since tx opts don't send by default)
+	err = s.L1.env.EthCl.SendTransaction(t.Ctx(), tx)
+	require.NoError(t, err, "must send tx")
+	s.lastL1DepositTxHash = tx.Hash()
+}
+
+func (s *CrossLayerUser) ActCheckDepositStatus(l1Success, l2Success bool) Action {
+	return func(t Testing) {
+		s.CheckDepositTx(t, s.lastL1DepositTxHash, 0, l1Success, l2Success)
```
tynes, line 303: Hardcoding 0 makes this less flexible but simple

op-e2e/actions/user.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+}
+
+func (s *CrossLayerUser) CheckDepositTx(t Testing, l1TxHash common.Hash, index int, l1Success, l2Success bool) {
+	depositReceipt := s.L1.CheckReceipt(t, l1Success, l1TxHash)
+	if depositReceipt == nil {
+		require.False(t, l1Success)
+		require.False(t, l2Success)
+	} else {
+		reconstructedDep, err := derive.UnmarshalDepositLogEvent(depositReceipt.Logs[index])
```
tynes, line 313: Bounds check here?

op-e2e/actions/user_test.go
```diff
@@ -0,0 +1,82 @@
...
+		AddressCorpora: addresses,
+		Bindings:       NewL2Bindings(t, l2Cl, withdrawalsCl),
+	}
+
+	alice := NewCrossLayerUser(log, dp.Secrets.Alice, rand.New(rand.NewSource(1234)))
+	alice.L1.SetUserEnv(l1UserEnv)
+	alice.L2.SetUserEnv(l2UserEnv)
+
+	// regular L2 tx, in new L2 block
+	alice.L2.ActResetTxOpts(t)
```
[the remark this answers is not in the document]
tynes: We can make a function that chains common actions together and reuse that?

op-node/rollup/driver/sequencer.go
```diff
@@ -91,6 +91,7 @@ func (d *Sequencer) CompleteBuildingBlock(ctx context.Context) (*eth.ExecutionPa
 	if err != nil {
 		return nil, fmt.Errorf("failed to complete building on top of L2 chain %s, error (%d): %w", d.buildingOnto.HeadBlockHash, errTyp, err)
 	}
+	d.buildingID = eth.PayloadID{}
```
tynes, line 94: Should this be part of its own PR?

op-e2e/actions/user.go
```diff
@@ -0,0 +1,322 @@
...
+
+	txOpts bind.TransactOpts
+
+	txToAddr   *common.Address
+	txCallData []byte
+
+	lastTxHash common.Hash
+}
+
+func NewBasicUser[B any](log log.Logger, priv *ecdsa.PrivateKey, rng *rand.Rand) *BasicUser[B] {
```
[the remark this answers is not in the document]
protolambda: RNR is to randomize some tx inputs easily.

op-e2e/actions/user.go
```diff
@@ -0,0 +1,322 @@
...
+
+type BasicUser[B any] struct {
+	log log.Logger
+	rng *rand.Rand
+	env *BasicUserEnv[B]
+
+	account *ecdsa.PrivateKey
+	address common.Address
+
+	txOpts bind.TransactOpts
```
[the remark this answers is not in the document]
protolambda: So we can set the user preferences separate from the tx sending. This way we can compose the tx configuration as actions, and reuse the tx settings easily.

op-e2e/actions/user.go
protolambda: The action can run just in time before the block creation, so should never be a bad estimate (and I'm curious when that estimate doesn't work). Alternatively we can set the gaslimit to non-0 with a separate action, and implement any special estimation there.

op-e2e/actions/user.go
protolambda: Exactly, this is a good example of when the higher level action (checking a deposit) omits some lower level edge case (checking tx success). When we do need the lower level detail in a test, we can use the underlying lower-level actions.

op-node/rollup/driver/sequencer.go
protolambda: The difference is clear with the tests in this PR, but behavior is equivalent outside of action testing, so I think we just keep it here.

trianglesphere (approved):
lgtm

PR #3629, opened 2022-10-01
op-node: Add wrapper to support EthSubscribe in HTTP clients

mslipper:
The Geth RPC client does not support subscriptions for HTTP endpoints. As a result, users who set an HTTP endpoint for the op-node were unable to sync since the driver never got any L1 head updates. This PR adds a wrapper around `client.RPC` that makes `newHeads` subscriptions work over HTTP. It does this by layering a notifications system that mimicks what Geth does under the hood on top of an HTTP polling loop.

The interface is the same as the one exposed by the raw RPC client. I tried my best to match Geth's subscription semantics, including:

- Closing the client causes all subscribers to get a nil error on their error channels.
- Unsubscribing closes each subscriber's error channel.
- Calling `Unsubscribe` multiple times is safe.

The poll rate is configurable. It default to 250ms, which should be fine for both L1 and L2. I tried at first to use Geth's underlying libraries, but found that it was impossible to do so without making changes to Geth itself to export more fields. This system is also simpler, and gives us room to instrument the polling loop in the future.

The wrapper is not integrated with the rest of the system in this PR. I'll stack those changes on top of this one so we can review this in isolation.

Totally open to different solutions here - marking as draft to reflect the fact that if there's a better way to do this, I'm happy o close this + the dependent PRs and go a different route.

Meta:

- Fixes ENG-2490

### DIFF
--- a/op-node/client/polling.go
+++ b/op-node/client/polling.go
@@ -1,0 +1,193 @@
+package client
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+var ErrSubscriberClosed = errors.New("subscriber closed")
+
+// PollingClient is an RPC client that provides newHeads subscriptions
+// via a polling loop. It's designed for HTTP endpoints, but WS will
+// work too.
+type PollingClient struct {
+	c        RPCGeneric
+	lgr      log.Logger
+	pollRate time.Duration
+	ctx      context.Context
+	cancel   context.CancelFunc
+	currHead *types.Header
+	subID    int
+
+	// pollReqCh is used to request new polls of the upstream
+	// RPC client.
+	pollReqCh chan struct{}
+
+	mtx sync.RWMutex
+
+	subs map[int]chan *types.Header
+
+	closedCh chan struct{}
+}
+
+type WrappedHTTPClientOption func(w *PollingClient)
+
+// WithPollRate specifies the rate at which the PollingClient will poll
+// for new heads. Setting this to zero disables polling altogether,
+// which is useful for testing.
+func WithPollRate(duration time.Duration) WrappedHTTPClientOption {
+	return func(w *PollingClient) {
+		w.pollRate = duration
+	}
+}
+
+// NewPollingClient returns a new PollingClient. Canceling the passed-in context
+// will close the client. Callers are responsible for closing the client in order
+// to prevent resource leaks.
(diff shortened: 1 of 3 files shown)


mslipper:
* develop
  * **PR #3629** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3629" target="_blank">Graphite</a>  👈
    * **PR #3630** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3630" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3629?utm_source=stack-comment).

protolambda (commented):
Code looks ok, but there's quite some complexity in the event handling, while Geth subscriptions have util funcs that might be useful to reduce that. In particular, we already have an [`op-node/eth.PollBlockChanges`](https://github.com/ethereum-optimism/optimism/blob/48aa3d14288624a372839fca563ee4ec10083d8f/op-node/eth/heads.go#L56) function which opens a subscription backed by a polling loop, using geth utils.
To solve the http heads source issue I would have expected to use that instead, or something like it if we need more functionality.

mslipper:
One issue I see with `PollBlockChanges` is if we call `EthSubscribe` multiple times it'll kick off multiple polling loops. Will find a solution that uses the Geth utils, however, to simplify the notifications code.

op-node/client/polling.go
```diff
@@ -0,0 +1,186 @@
...
+// for new heads. Setting this to zero disables polling altogether,
+// which is useful for testing.
+func WithPollRate(duration time.Duration) WrappedHTTPClientOption {
+	return func(w *PollingClient) {
+		w.pollRate = duration
+	}
+}
+
+// NewPollingClient returns a new PollingClient. Canceling the passed-in context
+// will close the client.
```
trianglesphere, line 51: Should comment that is starts to polling routine and the client must be closed to avoid leaks.

op-node/client/polling.go
```diff
@@ -0,0 +1,186 @@
...
+		opt(res)
+	}
+	go res.pollHeads()
+	return res
+}
+
+// Close closes the PollingClient and the underlying RPC client it talks to.
+func (w *PollingClient) Close() {
+	w.cancel()
+	w.c.Close()
```
trianglesphere, line 73: This might need a wait group / channel to synchronize the shutdown of the polling loop.

mslipper: By synchronize do you mean make `Close()` block until the polling loop is completely closed?

trianglesphere: yea

trianglesphere (approved):

trianglesphere:
**Graphite Merge Job**

Current status: 🔄 Waiting for checks to pass

This pull request is currently waiting for CI and mergeability checks to pass before being merged.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3629?utm_source=stack-comment).


Job Reference: djnQcRhFGqf86aZ5mLsM

PR #3632, opened 2022-10-03
ci: Parallelize op-e2e

mslipper:
`op-e2e` can easily be run across multiple executors. To make this work, we need to:

1. Make various helper utilities in `op-e2e` public and in the actual `op-e2e` package rather than the tests so that other packages can use them.
2. Enable test parallelism in `circleci.yml`.
3. Split the tests in `system_test.go` up into discrete packages so that CircleCI can detect timing/test split information for them.

This PR does 1 and 2, and starts 3 with `TestFinalize`. Unfortunately, CircleCI's test splitting mechanism isn't smart enough to split tests within the same Go package.

Ideally, this should reduce our `op-e2e` test runtime from ~3m to ~30s.

I also made the coverage generation a bit more intelligent by excluding things that aren't related to the op-node from the coverage report. This saves additional time, since the report took a while to generate.

Let me know if y'all are in favor of the approach. I'll be stacking the rest of the tests in 3 above as PRs on top of this one.

PR #3633, opened 2022-10-03
op-e2e: Migrate withdrawals and fees into their own package

mslipper:
Continues the process of migrating e2e tests into discrete packages so they can be parallelized.

### DIFF
--- a/op-e2e/setup.go
+++ b/op-e2e/setup.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"math/big"
+	"net"
 	"os"
 	"path"
 	"strings"
@@ -37,6 +38,7 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
 	hdwallet "github.com/miguelmota/go-ethereum-hdwallet"
+	"github.com/stretchr/testify/require"
 )
 
 // Init testing to enable test flags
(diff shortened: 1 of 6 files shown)


mslipper:
* develop
  * **PR #3632** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3632" target="_blank">Graphite</a> 
    * **PR #3633** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3633" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3633?utm_source=stack-comment).

op-e2e/setup.go
```diff
@@ -754,3 +756,17 @@ func CalcL1GasUsed(data []byte, overhead *big.Int) *big.Int {
...
+
+// GetFreePort gets an open, random port.
+func GetFreePort(t *testing.T) int {
+	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
+	require.NoError(t, err)
+	l, err := net.ListenTCP("tcp", addr)
+	require.NoError(t, err)
+	defer l.Close()
+	return l.Addr().(*net.TCPAddr).Port
+}
```
protolambda, line 772: IMO the preferred way would be to just configure 0 as listening port, and then retrieve the port that is used from the rollup node itself. This only works if we don't start the rollup nodes in parallel (since the port is released after returning it).

op-node/node/node.go
```diff
@@ -60,6 +60,7 @@ func New(ctx context.Context, cfg *Config, log log.Logger, snapshotLog log.Logge
 
 	err := n.init(ctx, cfg, snapshotLog)
 	if err != nil {
+		fmt.Println(err)
```
protolambda, line 63: Do we need this here?

protolambda (commented):
Looks great, although I think we might be able to speed up op-e2e without splitting it into independent CI jobs. Since this PR is not in draft mode I'm not approving, since it would auto-merge into the base PR otherwise (Single reviewer for PRs into non-develop branch? Or did we disable that?)

PR #3635, opened 2022-10-03
ci: Parallelize op-e2e

mslipper:
Replaces my previous PRs since `t.Parallel()` works fine. Also exposes the listener on the op-node, so that we can use a port of zero and allow the kernel to allocate a port.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -288,7 +288,7 @@ jobs:
       - run:
           name: test op-e2e
           command: |
-            gotestsum --junitfile /test-results/op-e2e.xml -- -coverpkg=github.com/ethereum-optimism/optimism/... -coverprofile=coverage.out -covermode=atomic ./...
+            gotestsum --format standard-verbose --junitfile /test-results/op-e2e.xml -- -coverpkg=github.com/ethereum-optimism/optimism/... -coverprofile=coverage.out -covermode=atomic ./...
           working_directory: op-e2e
       - run:
           name: test op-service
--- a/op-e2e/actions/l1_miner.go
+++ b/op-e2e/actions/l1_miner.go
@@ -29,11 +29,11 @@ type L1Miner struct {
 }
 
 // NewL1Miner creates a new L1Replica that can also build blocks.
-func NewL1Miner(log log.Logger, genesis *core.Genesis) (*L1Miner, func() error) {
-	rep, done := NewL1Replica(log, genesis)
+func NewL1Miner(log log.Logger, genesis *core.Genesis) *L1Miner {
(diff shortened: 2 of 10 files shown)


.circleci/config.yml
```diff
@@ -288,7 +288,7 @@ jobs:
       - run:
           name: test op-e2e
           command: |
-            gotestsum --junitfile /test-results/op-e2e.xml -- -coverpkg=github.com/ethereum-optimism/optimism/... -coverprofile=coverage.out -covermode=atomic ./...
+            gotestsum --format standard-verbose --junitfile /test-results/op-e2e.xml -- -coverpkg=github.com/ethereum-optimism/optimism/... -coverprofile=coverage.out -covermode=atomic ./...
```
tynes, line 291: Should it be `-coverpkg=github.com/ethereum-optimism/optimism/op-e2e/...` ?

mslipper: Should be Optimism, since we're covering other Optimism packages with these tests.

op-e2e/actions/l1_miner.go
```diff
@@ -29,11 +29,11 @@ type L1Miner struct {
...
 
 // NewL1Miner creates a new L1Replica that can also build blocks.
-func NewL1Miner(log log.Logger, genesis *core.Genesis) *L1Miner {
-	rep := NewL1Replica(log, genesis)
+func NewL1Miner(log log.Logger, genesis *core.Genesis) (*L1Miner, func() error) {
+	rep, done := NewL1Replica(log, genesis)
 	return &L1Miner{
 		L1Replica: *rep,
-	}
+	}, done
```
protolambda, line 36: Maybe we can use the `testing.Cleanup()` functionality to close the geth node, instead of passing around a close function?

mslipper: This is neat, didn't know that was a feature. Will look into that.

protolambda:
The L2 engine actor may also need to be closed (my bad for not thinking of that in original PR)

tynes (approved):

.circleci/config.yml
protolambda, line 291: Does this pass through flags to the go testing tool? In that case we might need to add `-parallel`

protolambda (commented):
code LGTM, but wondering about CLI flags, expected `-parallel`.

.circleci/config.yml
mslipper: It does pass through flags, but it's already running the tests in parallel. You can also see that parallel tests are being picked up in the CI job - the e2e tests are taking 30s rather than 3m.

PR #3638, opened 2022-10-03
sdk: add hh task for finalizing withdrawal

tynes:
**Description**

Now we no longer need to use one off scripts for testing this kind of thing. The hh task will check to see the status of a transaction hash and then finalize the withdrawal on L1 if the message status is ready for relay. It does this by creating a `CrossChainMessenger`.

### DIFF
--- a/packages/sdk/tasks/finalize-withdrawal.ts
+++ b/packages/sdk/tasks/finalize-withdrawal.ts
@@ -1,0 +1,96 @@
+import { task, types } from 'hardhat/config'
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
+import { Wallet, providers } from 'ethers'
+import { predeploys } from '@eth-optimism/contracts-bedrock'
+import 'hardhat-deploy'
+import '@nomiclabs/hardhat-ethers'
+
+import {
+  CrossChainMessenger,
+  StandardBridgeAdapter,
+  MessageStatus,
+} from '../src'
+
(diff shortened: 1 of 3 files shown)


packages/sdk/tasks/finalize-withdrawal.ts
```diff
@@ -0,0 +1,96 @@
...
+    const l2Signer = new Wallet(hre.network.config.accounts[0], l2Provider)
+
+    let Deployment__L1StandardBridgeProxy = await hre.deployments.getOrNull(
+      'L1StandardBridgeProxy'
+    )
+    if (Deployment__L1StandardBridgeProxy === undefined) {
+      Deployment__L1StandardBridgeProxy = await hre.deployments.getOrNull(
+        'Proxy__OVM_L1StandardBridge'
+      )
+    }
```
maurelian, line 46: So, if we have both `contracts-bedrock` and `contracts` deployments available, we default to `contracts-bedrock`?
Isn't the main Goerli deployment still on `contracts`, but in the `hardhat.config.ts` file above we've made both available?

maurelian (commented):
Just a question for now.

packages/sdk/tasks/finalize-withdrawal.ts
tynes: Both are available so that this script can work with networks that are not legacy

maurelian (approved):

mslipper (approved):

PR #3639, opened 2022-10-03
op-node: Simplify Receipts Fetching API

trianglesphere:
**Description**

The previous api would return a receipts fetcher that the caller was responsible for iterating through and then calling `Result` on. This moves this inside the `Fetch` function. This simplifies usage of the API and does not result in performance regression because the receipts hash check requires all receipts to be fetched before

**Tests**

No added tests, but all pass.

### DIFF
--- a/op-node/rollup/derive/attributes.go
+++ b/op-node/rollup/derive/attributes.go
@@ -3,7 +3,6 @@ package derive
 import (
 	"context"
 	"fmt"
-	"io"
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
@@ -15,7 +14,7 @@ import (
 // L1ReceiptsFetcher fetches L1 header info and receipts for the payload attributes derivation (the info tx and deposits)
 type L1ReceiptsFetcher interface {
 	InfoByHash(ctx context.Context, hash common.Hash) (eth.BlockInfo, error)
-	Fetch(ctx context.Context, blockHash common.Hash) (eth.BlockInfo, types.Transactions, eth.ReceiptsFetcher, error)
+	Fetch(ctx context.Context, blockHash common.Hash) (eth.BlockInfo, types.Receipts, error)
 }
 
 // PreparePayloadAttributes prepares a PayloadAttributes template that is ready to build a L2 block with deposits only, on top of the given l2Parent, with the given epoch as L1 origin.
@@ -32,7 +31,7 @@ func PreparePayloadAttributes(ctx context.Context, cfg *rollup.Config, dl L1Rece
 	// case we need to fetch all transaction receipts from the L1 origin block so we can scan for
 	// user deposits.
 	if l2Parent.L1Origin.Number != epoch.Number {
-		info, _, receiptsFetcher, err := dl.Fetch(ctx, epoch.Hash)
+		info, receipts, err := dl.Fetch(ctx, epoch.Hash)
 		if err != nil {
 			return nil, NewTemporaryError(fmt.Errorf("failed to fetch L1 block info and receipts: %w", err))
 		}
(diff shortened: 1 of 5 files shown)


tynes:
Generally seems like a sane change that makes using the `L1ReceiptsFetcher` easier, can't really think of a good reason to defer fetching the receipts like how the old code was architected

tynes (approved):

protolambda (changes requested):
Do we really want this? Sure it simplifies the API (yay, good thinking), but I expect it makes things quite unreliable when syncing from a L1 source: receipts are not cached individually, and the progress of the fetching is lost, since the receipt fetcher itself isn't cached anymore either.

Note that there's 1 receipt request per L1 tx. And on average Mainnet L1 has 10x as many txs per block as Goerli. And our op-e2e tx throughput is both lower and unrealistically stable (pretty much never a failed receipt retrieval).

Maybe we should do a mainnet test: without deploying any contracts, we can create a ghost rollup, that essentially functions the same like a rollup without active sequencer. Verifiers will run through all L1 blocks and receipts, and hit the bottlenecks that I think are fragile enough to warrant per-request caching/continuation on errors. If that works fine we can merge this, but I highly doubt it (especially when syncing from a rate-limited L1 endpoint, like an average user might...)

trianglesphere:
> Do we really want this?

It's not an easy API to use and doesn't document that it relies on future calls to `(receiptsFetcher).Fetch` to work after a previous one didn't because `Fetch` hits a catch

> receipts are not cached individually

That's not that hard to do.

trianglesphere:
@protolambda thanks for pointing out that the receipts fetcher cache is important. I've added a commit to this PR which now caches the receipts fetcher instead of the set of receipts. It attempts to used a cached receipts fetcher instead of making a new one. This means that this new code is the essentially the same as the old code except for the fact that the iteration is done inside of the `Fetch` call rather than outside.

How do you feel about this change? I think the performance should be the same and the API is a bit easier to use.

op-node/rollup/derive/attributes.go
```diff
@@ -41,17 +40,7 @@ func PreparePayloadAttributes(ctx context.Context, cfg *rollup.Config, dl L1Rece
...
-		for {
-			if err := receiptsFetcher.Fetch(ctx); err == io.EOF {
-				break
-			} else if err != nil {
-				return nil, NewTemporaryError(fmt.Errorf("failed to fetch more receipts: %w", err))
-			}
-		}
-		receipts, err := receiptsFetcher.Result()
-		if err != nil {
-			return nil, NewResetError(fmt.Errorf("fetched bad receipt data: %w", err))
```
protolambda, line 53: With the new change we don't distinguish between a temporary RPC error and a corrupted (or malicious) L1 endpoint providing bad receipt data. I think it's fine as it's probably not the right approach to reorg anyway if we're getting bad L1 data, and we'll follow the right chain by hash eventually.

op-node/rollup/derive/attributes.go
```diff
@@ -15,7 +14,7 @@ import (
 // L1ReceiptsFetcher fetches L1 header info and receipts for the payload attributes derivation (the info tx and deposits)
 type L1ReceiptsFetcher interface {
 	InfoByHash(ctx context.Context, hash common.Hash) (eth.BlockInfo, error)
-	Fetch(ctx context.Context, blockHash common.Hash) (eth.BlockInfo, types.Transactions, eth.ReceiptsFetcher, error)
+	Fetch(ctx context.Context, blockHash common.Hash) (eth.BlockInfo, types.Receipts, error)
```
protolambda, line 17: If we're removing the txs output we should maybe just rename it to `FetchReceipts`, but I'm happy to drop the unused output either way.

protolambda (approved):
LGTM!

Hopefully the `debug_getRawReceipts` RPC method becomes more accessible than debug namespace some day, then the receipts fetcher stuff would be easier for us both. (And might still even be worth adding as option for those that sync from their own node now that I think of it)

op-node/rollup/derive/attributes.go
trianglesphere: Yea, the other thing is that the iterative batch call will nuke it's local state if it errs on the a failed consistency check & will re-fetch the data. That serves as a better retry logic for when one of the receipts is invalid.

trianglesphere:
gonna merge by hand b/c mergify messed up the update merge.

PR #3640, opened 2022-10-03
op-node: Remove BlockIDRange

trianglesphere:
**Description**

This function was not being used and could be safely removed.


tynes:
Perhaps the hive test is a flake?

trianglesphere:
> Perhaps the hive test failure is a flake?

I think so. Currently re-running. It didn't fail at build time.

protolambda (approved):
Nice catch, forgot we kept this around, indeed dead code.

tynes (approved):

PR #3641, opened 2022-10-03
op-node: split start of payload building and confirmation of payload

protolambda:
**Description**

This splits the payload building process into two steps. The payload building is common between verifier and sequencer, but a sequencer may want to give the Engine more time before sealing the block that is built, so it can fill the block, and optimize tx choice. Similarly, in action testing, as L2 sequencer actor, we want to enact the building of the payload as Engine actor to reach different edge-cases, and thus also need to breathe before sealing the payload.

This PR does not change any functionality, it merely refactors a function into two parts, and adds docs.


trianglesphere (approved):
lgtm. I like the new comments about the BlockInsert error codes.

tynes (approved):
Looks good to me, test failure looks like a flake

PR #3647, opened 2022-10-04
op-node: split driver state/model into L1 state, L1 origin selector, L2 sequencer, L2 derivation

protolambda:
**Description**

TLDR of driver refactor design doc:
- `Driver` should not mix state with event handling so much. Separate with interfaces for testing instead.
- Sequencing needs to be separated as a model, if we want to test it separate from the event handling, like in action-testing
- L1 origin selection can be tricky, separating it will enable us to unit test it

Changes:
- No more `Driver struct { state }` that proxies all calls. We just define a `Driver` struct.
- Separate L1 head/safe/finalized, with handling funcs, from driver state
- Separate L1 origin finding from driver state
- Separate sequencing from driver state, and deprecate the all-in-one-go sequencing in favor of a two-step process which we can adopt later, to enable the sequencer to build the best blocks and maybe even respond to reorgs before completing a block.
- Cleaned up unused arg in `Start` function

**Tests**

No functionality changed, it's rewiring to make testing much more viable. In the L2 sequencer action-testing PR I plan to plug in all these state components (origin selection, l1 state, sequencing). We can wait with this PR if we're unsure.

**Metadata**

Fix ENG-2864


op-node/rollup/driver/l1_state.go
```diff
@@ -0,0 +1,68 @@
...
+	s.l1Safe = safe
+}
+
+func (s *L1State) HandleNewL1FinalizedBlock(finalized eth.L1BlockRef) {
+	s.Log.Info("New L1 finalized block", "l1_finalized", finalized)
+	s.Metrics.RecordL1Ref("l1_finalized", finalized)
+	s.l1Finalized = finalized
+}
+
+func (s *L1State) L1Head() eth.L1BlockRef {
```
tynes, line 58: Do we need a rwlock here?

protolambda: It's only used synchronously currently, we can leave it out for now

op-node/rollup/driver/origin_selector.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
+)
+
+type L1Blocks interface {
+	derive.L1BlockRefByHashFetcher
+	derive.L1BlockRefByNumberFetcher
+}
+
+type L1OriginSelector struct {
+	Cfg                 *rollup.Config
```
trianglesphere, line 19: again, fields should tend to be private by default unless the struct is simply a collection of data.

op-node/rollup/driver/sequencer.go
```diff
@@ -0,0 +1,101 @@
...
+	payload, errTyp, err := derive.ConfirmPayload(ctx, d.Log, d.L2, d.buildingOnto, d.buildingID, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to complete building on top of L2 chain %s, error (%d): %w", d.buildingOnto.HeadBlockHash, errTyp, err)
+	}
+	return payload, nil
+}
+
+// createNewBlock sequences a L2 block with immediate building and sealing.
+// Deprecated: the sequencer should build the block in two steps, adjusted to maximize building time, within the block-time span, or faster when catching up.
+func (d *Sequencer) createNewBlock(ctx context.Context, l2Head eth.L2BlockRef, l2SafeHead eth.BlockID, l2Finalized eth.BlockID, l1Origin eth.L1BlockRef) (eth.L2BlockRef, *eth.ExecutionPayload, error) {
```
trianglesphere, line 87: For this file, I'm fine leaving the function signatures as is, but I'd suggest that the `start/stop building` functions to be private and this function to be public with the deprecated warning removed. The reason for this is that the `CompleteBuildingBlock` function reuses the forkchoice update struct. This means that if there are other updates to `safeHead` or `finalizedHead` the forkchoice update in `completeBuildingBlock` will be out of date. To avoid this the `Completion` API could be changed or it needs to be noted that new modifications to the l2 execution engine state should be made in between the `start` and `complete` calls.

op-node/rollup/driver/origin_selector.go
```diff
@@ -0,0 +1,69 @@
...
+	Log                 log.Logger
+	L1                  L1Blocks
+	SequencingConfDepth uint64
+}
+
+// FindL1Origin determines what the next L1 Origin should be.
+// The L1 Origin is either the L2 Head's Origin, or the following L1 block
+// if the next L2 block's time is greater than or equal to the L2 Head's Origin.
+func (los *L1OriginSelector) FindL1Origin(ctx context.Context, l1Head eth.L1BlockRef, l2Head eth.L2BlockRef) (eth.L1BlockRef, error) {
```
trianglesphere, line 28: I'm starting to wonder if this should be merged in with `L1 State`. Outside of `syncStatus`, this is the only usage of the `l1Head`.

op-node/rollup/driver/sequencer.go
```diff
@@ -0,0 +1,101 @@
...
+type Downloader interface {
+	InfoByHash(ctx context.Context, hash common.Hash) (eth.BlockInfo, error)
+	Fetch(ctx context.Context, blockHash common.Hash) (eth.BlockInfo, types.Transactions, eth.ReceiptsFetcher, error)
+}
+
+// Sequencer implements the sequencing interface of the driver: it starts and completes block building jobs.
+type Sequencer struct {
+	L1     Downloader
+	L2     derive.Engine
+	Log    log.Logger
```
trianglesphere, line 26: private fields

trianglesphere (changes requested):
The biggest thing is that I don't think we're ready to migrate over to the `start/complete` block building process, but it is good that we're setting it up to go that way.

op-node/rollup/driver/origin_selector.go
protolambda: Do you prefer `NewXYZ(...)` and private fields? I tend to prefer structs with public fields when the initializer would basically be a pass-through and mutating the fields doesn't really affect the other private fields of the struct.

op-node/rollup/driver/origin_selector.go
protolambda: Good point about `l1Head`. The safe/finalized L1 data however is not part of this function, so I think it's better to isolate, to keep tests small and logic more encapsulated.

op-node/rollup/driver/sequencer.go
protolambda: Will add a note about the forkchoice state usage and consistency between start and complete. I do need to make the start/complete functions public so I can call into them from the op-e2e action testing package, or we end up with duplicate code in the action-testing that doesn't cover the actual driver code.

op-node/rollup/driver/origin_selector.go
protolambda: Made the fields private and added `New...` constructors so I can still cover these in action testing

protolambda:
Will rebase to fix merge conflicts, and include the deadlock fix

op-node/rollup/driver/origin_selector.go
trianglesphere: other than sync status (which is nice to have), we don't really use the rest of the L1 State. It's not an immediate change, but I think it might be a possible simplification down the road.

trianglesphere (approved):
nice work, every looks great

tynes (approved):

PR #3649, opened 2022-10-04
op-e2e: Adopt op-chain-ops

mslipper:
This PR updates op-e2e to use op-chain-ops. Adopting op-chain-ops in op-e2e improves the following:

1. Transactions now route through the correct proxy contracts.
2. There's no need to wait for a contract deployment since conracts exist in genesis.
3. We can export the test helpers in op-e2e to enable us to run similar end-to-end tests in other applications like the indexer


trianglesphere:
@mslipper I think the secrets have already been ported over to the `e2eutils` folder: https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/e2eutils/secrets.go

mslipper:
oh nice! closing this then will restack.

tynes (approved):

mslipper:
I rolled #3650 in here since the secrets util was already included - will push once tests pass locally.

mslipper:
Removing dead code made this a hair over PR size SLA.

op-e2e/setup.go
```diff
@@ -178,135 +236,78 @@ func (cfg SystemConfig) start() (*System, error) {
...
-	batchSubmitterAddr := crypto.PubkeyToAddress(bssPrivKey.PublicKey)
 
-	p2pSignerPrivKey, err := wallet.PrivateKey(accounts.Account{
-		URL: accounts.URL{
-			Path: cfg.P2PSignerHDPath,
-		},
-	})
-	if err != nil {
-		return nil, err
+	for addr, amount := range cfg.Premine {
```
tynes, line 244: Is this premine still necessary? Can we dedupe this code and add something to the chain ops package and not set anything here?

mslipper: It is - we sometimes want to override balances in the dev accounts, so having a way to do that without specifying all of them is nice.

We can certainly upgrade the chain ops package, but I'd rather minimize the diff here and do that separately.

tynes (approved):

tynes:
being able to adopt this code and have it used as the test infra is nice, gives more confidence in using it for the actual state surgery

op-e2e/e2eutils/secrets.go
```diff
@@ -43,8 +45,6 @@ type MnemonicConfig struct {
 	Mallory string
 }
 
-// Secrets computes the private keys for all mnemonic paths,
```
trianglesphere, line 46: why are these doc comments getting removed?

op-e2e/setup.go
```diff
@@ -178,135 +236,78 @@ func (cfg SystemConfig) start() (*System, error) {
...
+		existing, ok := l1Genesis.Alloc[addr]
+		if ok {
```
trianglesphere, line 246:
```suggestion
		if existing, ok := l1Genesis.Alloc[addr]; ok {
```

trianglesphere (commented):
2 small things, but I'm excited for this

op-e2e/e2eutils/secrets.go
mslipper: I coped this from Hive - can add it back.

op-e2e/e2eutils/secrets.go
```diff
@@ -94,9 +101,9 @@ func (m *MnemonicConfig) Secrets() (*Secrets, error) {
 	}, nil
 }
 
-// Secrets bundles secp256k1 private keys for all common rollup actors for testing purposes.
 type Secrets struct {
```
trianglesphere, line 104:
```suggestion
// Secrets bundles secp256k1 private keys for all common rollup actors for testing purposes.
type Secrets struct {
```

op-e2e/e2eutils/secrets.go
```diff
@@ -131,9 +139,9 @@ func (s *Secrets) Addresses() *Addresses {
 	}
 }
 
-// Addresses bundles the addresses for all common rollup addresses for testing purposes.
 type Addresses struct {
```
trianglesphere, line 142:
```suggestion
// Addresses bundles the addresses for all common rollup addresses for testing purposes.
type Addresses struct {
```

tynes:
I'd like to get this merged so I can use it to write better tests for https://github.com/ethereum-optimism/optimism/pull/3528
Realizing that the tests that I'd like to write belong more in `op-e2e` than in `op-chain-ops`

mslipper:
I'm debugging a test timeout. Stand by.

trianglesphere (approved):

mslipper:
Ok, this one was really interesting - the timeout was caused by a deadlock in the `SyncStatus` function. Here's the implementation:

```
func (s *state) SyncStatus(ctx context.Context) (*eth.SyncStatus, error) {
	respCh := make(chan eth.SyncStatus)
	select {
	case <-ctx.Done():
		return nil, ctx.Err()
	case s.syncStatusReq <- respCh:
		select {
		case <-ctx.Done():
			return nil, ctx.Err()
		case resp := <-respCh:
			return &resp, nil
		}
	}
}
```

In some cases, the e2e test would cancel the passed-in context. This caused the function to return. If the timing as correct, the context would be canceled _before_ the driver loop had a chance to send a response to `respCh`. This caused closing the rollup node to hang, since the listener for the `done` channel in `Close()` is inside the big driver loop.

The solution was to make `respCh` buffered, so that the driver can still send responses to the channel after the `SyncStatus` context is closed.

PR #3652, opened 2022-10-04
ctb: Rename event in the L2ToL1MessagePasser

maurelian:
**Description**

The new name MessagePassed, is less confusing since the
name 'WithdrawalInitiated' is used in the L2 Bridge.

**Tests**

Existing tests are updated to match the implementation.


tynes:
Going to need to deeply grep to get all of the changes required for this one

smartcontracts (approved):

packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
```diff
@@ -51,11 +51,11 @@ contract L2ToL1MessagePasser is Semver {
 
     /**
      * @notice Emitted any time a withdrawal is initiated. An extension to
-     *         WithdrawalInitiated so that the interface is maintained.
+     *         MessagePassed to allow for a 4th indexed argument.
      *
      * @param hash The hash of the withdrawal
      */
-    event WithdrawalInitiatedExtension1(bytes32 indexed hash);
+    event MessagePassedExtension1(bytes32 indexed hash);
```
maurelian, line 58: Comment to stop mergify

tynes: I think we can consolidate these events, the workflows are going to involve more getting the event by user who is withdrawing and then pulling the hash from the event

tynes (approved):

packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
tynes: I'm open to this via a followup PR

maurelian: yeah, agree that's better

PR #3656, opened 2022-10-05
CI Builder: install slither==0.9.0

maurelian:
This reverts commit b979d861879c74d6d38a0a7abdd75d267c1a0912, which installed a specific slither commit. 

A new release of Slither was announced today ([0.9.0](https://github.com/crytic/slither/releases/tag/0.9.0)) which includes the fixes I wanted. 

**Tests**

I updated and ran slither locally to make sure it works on all our contracts without having to delete the test files.

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


tynes (approved):

tynes:
I do think we want a changeset to trigger a new build. Can you confirm that when installing from pip, it gives you the latest version?

maurelian:
> I do think we want a changeset to trigger a new build. Can you confirm that when installing from pip, it gives you the latest version?

@tynes Based on local testing, yes `pip install slither-analyzer` installs `0.9.0`, but I've made it explicit in the `Dockerfile`, which is probably a good idea given we've had things break with slither in the past. 

Also added a changeset.

tynes (approved):
Ok great, thank you

mslipper (approved):

PR #3658, opened 2022-10-05
indexer: Restore some bindings

mslipper:
To support Bedrock and non-Bedrock within the same indexer, we need to include the SCC bindings. This PR adds them back under a `legacy/scc` directory.


mslipper:
* develop
  * **PR #3658** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3658" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3658?utm_source=stack-comment).

indexer/Makefile
```diff
@@ -7,6 +7,8 @@ LDFLAGSSTRING +=-X main.GitDate=$(GITDATE)
 LDFLAGSSTRING +=-X main.GitVersion=$(GITVERSION)
 LDFLAGS := -ldflags "$(LDFLAGSSTRING)"
 
+SCC_ABI_ARTIFACT = ../packages/contracts/artifacts/contracts/L1/rollup/StateCommitmentChain.sol/StateCommitmentChain.json
```
trianglesphere, line 10: is this file checked in?

mslipper: No it's not - it's an artifact generated during compilation of the legacy contracts.

trianglesphere: Ok, can you leave a comment in the makefile to that effect? That means that running this make command on a fresh repo would fail if they haven't already been built. Also, do we ever intend to change these?

tynes (approved):

roninjin10 (approved):

roninjin10:
Clicking this button
image

indexer/Makefile
tynes: Unlikely they will ever change, we used this pattern in the past. There will come a time when we delete the `contracts` package and rename `contracts-bedrock` to `contracts`. I wonder if a solidity interface can be used to generate bindings, if so we could add the scc interface to `contracts-bedrock` and generate the bindings that way but then we won't have the deployment functionality, which i suppose we need for testing

PR #3659, opened 2022-10-05
indexer: Update DB package

mslipper:
Updates the indexer's DB package to support both Bedrock and the legacy system. 90% of of this is a revert back to Mark's last commit on the indexer pre-Bedrock.

### DIFF
--- a/indexer/db/db.go
+++ b/indexer/db/db.go
@@ -193,20 +193,10 @@ func (d *Database) AddIndexedL1Block(block *IndexedL1Block) error {
 
 	const insertDepositStatement = `
 	INSERT INTO deposits
-		(guid, from_address, to_address, l1_token, l2_token, amount, tx_hash, log_index, l1_block_hash, data)
+		(guid, from_address, to_address, l1_token, l2_token, amount, tx_hash, log_index, block_hash, data)
 	VALUES
 		($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)
 	`
-
-	const insertWithdrawalStatement = `
-	INSERT INTO withdrawals
-		(guid, from_address, to_address, l1_token, l2_token, amount, tx_hash, log_index, l1_block_hash, data)
-	VALUES
-		($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)
-	ON CONFLICT (tx_hash)
-		DO UPDATE SET l1_block_hash = $9;
-	`
-
 	return txn(d.db, func(tx *sql.Tx) error {
 		_, err := tx.Exec(
 			insertBlockStatement,
@@ -242,29 +232,6 @@ func (d *Database) AddIndexedL1Block(block *IndexedL1Block) error {
 			}
 		}
 
-		if len(block.Withdrawals) == 0 {
-			return nil
-		}
-
-		for _, withdrawal := range block.Withdrawals {
-			_, err = tx.Exec(
-				insertWithdrawalStatement,
-				NewGUID(),
-				withdrawal.FromAddress.String(),
-				withdrawal.ToAddress.String(),
-				withdrawal.L1Token.String(),
-				withdrawal.L2Token.String(),
-				withdrawal.Amount.String(),
-				withdrawal.TxHash.String(),
-				withdrawal.LogIndex,
-				block.Hash.String(),
-				withdrawal.Data,
-			)
-			if err != nil {
-				return err
-			}
-		}
-
 		return nil
 	})
 }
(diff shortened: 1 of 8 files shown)


mslipper:
* develop
  * **PR #3658** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3658" target="_blank">Graphite</a> 
    * **PR #3659** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3659" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3659?utm_source=stack-comment).

indexer/db/db.go
```diff
@@ -282,9 +249,9 @@ func (d *Database) AddIndexedL2Block(block *IndexedL2Block) error {
 
 	const insertWithdrawalStatement = `
 	INSERT INTO withdrawals
-		(guid, from_address, to_address, l1_token, l2_token, amount, tx_hash, log_index, l2_block_hash, data)
+		(guid, from_address, to_address, l1_token, l2_token, amount, tx_hash, log_index, block_hash, data, br_withdrawal_hash)
```
tynes, line 252: Why does this need to be tying bedrock into the withdrawal hash? It seems like its not very future proof

mslipper: It delineates which fields are legacy, and which are not. This field is only used for Bedrock withdrawals. We can remove the prefixes after launch once we’re no longer supporting the legacy system.

indexer/db/db.go
```diff
@@ -401,68 +399,79 @@ func (d *Database) GetDepositsByAddress(address common.Address, page PaginationP
...
-		l2_blocks.number, l2_blocks.timestamp
+		l2_blocks.number, l2_blocks.timestamp, withdrawals.br_withdrawal_hash
 	FROM withdrawals
-		INNER JOIN l2_blocks ON withdrawals.l2_block_hash=l2_blocks.hash
+		INNER JOIN l2_blocks ON withdrawals.block_hash=l2_blocks.hash
 		INNER JOIN l2_tokens ON withdrawals.l2_token=l2_tokens.address
-	WHERE withdrawals.from_address = $1 ORDER BY l2_blocks.timestamp LIMIT $2 OFFSET $3;
-	`
+	WHERE withdrawals.from_address = $1 %s ORDER BY l2_blocks.timestamp LIMIT $2 OFFSET $3;
+	`, FinalizationStateAny.SQL())
```
tynes, line 474: This is safe from sql injection attacks?

mslipper: Yes. FinalizationState is an int that returns a SQL fragment. In later PRs it’ll be used by the API to request withdrawals in different states. For not it just supports the one.

mslipper: It doesn’t consume any user input.

indexer/db/withdrawal.go
```diff
@@ -17,6 +17,7 @@ type Withdrawal struct {
 	Amount      *big.Int
 	Data        []byte
 	LogIndex    uint
+	BedrockHash *common.Hash
```
tynes, line 20: This is specifically a `StandardBridgeWithdrawal` since it corresponds to a token withdrawal. A general withdrawal is here: https://github.com/ethereum-optimism/optimism/blob/d04d81e8615f4551d438a8e1fd0780749b1c6569/op-chain-ops/crossdomain/withdrawal.go#L15

The scope of this is only token withdrawals?

mslipper: Token and eth withdrawals are the same in the indexer. Eth is represented in the db as a token withdrawal with a special address.

tynes:
Is the plan to test this using the `op-e2e` utils once we get more stuff cleaned up?

There are currently linting errors, will review again once fixed

indexer/db/withdrawal.go
mslipper: The hash is calculated from the message passer WithdrawalInitiated event, which I believe is emitted for both.

mslipper:
Correct, the last PR in the chain will have op-e2e tests. I’ll fix the linter.

indexer/db/withdrawal.go
tynes: Got it - the bridge handles eth in the same way, `address(0)` and `address(0x..dead)`

I meant more like arbitrary L2 to L1 messages, which i'm getting the feeling its no

mslipper: Ah got it. Yeah it only supports standard bridge deposits/withdrawals, not arbitrary L1 messages. This is the same as the previous implementation. We only care about the deposits and withdrawals users can see in the gateway UI, so this is OK.

indexer/db/db.go
tynes: Ok, i see that there is a switch statement that determines the possible strings that can be returned here

indexer/db/withdrawal.go
roninjin10: There is a wish for the UI to use the lower level contract (apologies for forgetting the name) to do deposits/withdrawals because it's cheaper to use.  But @zainbacchus and others determined that it's not a priority for our OKRs at this time.

roninjin10 (approved):
I d

tynes:
> Correct, the last PR in the chain will have op-e2e tests. I’ll fix the linter.

I'm realizing there isn't a good way to include legacy contracts in `op-e2e`, is this something you would like to do?

tynes (approved):

PR #3663, opened 2022-10-06
ctb: Format forge tests with prettier

maurelian:
**Description**

Thus far the forge test suite has not been formatted with prettier. This PR applies prettier formatting to our test files. 

Concerning readability, I think the one downside is that long `assertEq()` calls are definitely less readable now. 
image. 
The upside for readability is that the LoC is significantly reduced: ` 27 files changed, 368 insertions(+), 698 deletions(-)`

We should also just favor standardization and remove the need to make judgement calls or nits wrt style.

**Tests**

This does heavily modify the test files, but the absence of a diff in the gas-snapshot should be evidence that none of the logic has changed.


tynes:
Thoughts on `forge fmt`?

maurelian:
> Thoughts on `forge fmt`?

It broke our beautiful seaport style commenting last time I tried it, but that was a while ago, they might have more config options now? 
Worth considering for a future PR.

smartcontracts:
Strongly prefer that we do not change any formatting style before going to Bedrock. We have enough to worry about and the format is pretty good for now.

smartcontracts (approved):

packages/contracts-bedrock/contracts/test/RLP.t.sol
```diff
@@ -13,19 +13,58 @@ library LibRLP {
...
+        if (nonce <= 0x7f)
+            return
+                keccak256(abi.encodePacked(bytes1(0xd6), bytes1(0x94), deployer, uint8(nonce)))
+                    .fromLast20Bytes();
 
         // Nonces greater than 1 byte all follow a consistent encoding scheme, where each value is preceded by a prefix of 0x80 + length.
-        if (nonce <= type(uint8).max)  return keccak256(abi.encodePacked(bytes1(0xd7), bytes1(0x94), deployer, bytes1(0x81), uint8(nonce))).fromLast20Bytes();
-        if (nonce <= type(uint16).max) return keccak256(abi.encodePacked(bytes1(0xd8), bytes1(0x94), deployer, bytes1(0x82), uint16(nonce))).fromLast20Bytes();
-        if (nonce <= type(uint24).max) return keccak256(abi.encodePacked(bytes1(0xd9), bytes1(0x94), deployer, bytes1(0x83), uint24(nonce))).fromLast20Bytes();
+        if (nonce <= type(uint8).max)
```
tynes, line 26: oof

tynes (approved):

packages/contracts-bedrock/contracts/test/RLP.t.sol
maurelian: Agree 🥺

PR #3664, opened 2022-10-06
Add balance metrics to op-propose & op-batcher

trianglesphere:
**Description**

This adds a utility function to the `op-service` package which launches a go rountine
to check the balance of a given account. This is helpful for in process monitoring of
the balance of an account for a specific service (like the batcher or proposer).

**Tests**

Tested locally with devnet.

**Metadata**
- Fixes ENG-2713

### DIFF
--- a/op-service/metrics/balance.go
+++ b/op-service/metrics/balance.go
@@ -1,0 +1,59 @@
+package metrics
+
+import (
+	"context"
+	"math/big"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// weiToEther divides the wei value by 10^18 to get a number in ether as a float64
+func weiToEther(wei *big.Int) float64 {
+	num := new(big.Rat).SetInt(wei)
+	denom := big.NewRat(params.Ether, 1)
+	num = num.Quo(num, denom)
+	f, _ := num.Float64()
+	return f
+}
+
+// LaunchBalanceMetrics fires off a go rountine that queries the balance of the supplied account & periodically records it
+// to the balance metric of the namespace. The balance of the account is recorded in Ether (not Wei).
+// Cancel the supplied context to shut down the go routine
+func LaunchBalanceMetrics(ctx context.Context, log log.Logger, r *prometheus.Registry, ns string, client *ethclient.Client, account common.Address) {
+	go func() {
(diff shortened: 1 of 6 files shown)


op-service/metrics/balance.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// 10^18
+var etherScalar = new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
```
tynes, line 16: This constant is defined in many places throughout the monorepo and also in geth itself

trianglesphere: Do you know the names of the other definitions? There's one with the same name (also private) in the op-e2e package. In geth there is a private one in `cmd/faucet`.

mslipper (approved):

op-service/metrics/balance.go
protolambda: e2eutils or action testing has one too iirc. We should probably just put a global in the `op-node/eth` package

tynes: https://github.com/ethereum/go-ethereum/blob/b628d7276624c2d8ea7dd97d2259a2c2fce7d3cc/params/denomination.go

protolambda (approved):

op-service/metrics/balance.go
```diff
@@ -0,0 +1,61 @@
...
+// Cancel the supplied context to shut down the go routine
+func LaunchBalanceMetrics(ctx context.Context, log log.Logger, r *prometheus.Registry, ns string, client *ethclient.Client, account common.Address) {
+	go func() {
+		balanceGuage := promauto.With(r).NewGauge(prometheus.GaugeOpts{
+			Namespace: ns,
+			Name:      "balance",
+			Help:      "balance (in ether) of account " + account.String(),
+		})
+
+		ticker := time.NewTicker(10 * time.Second)
```
tynes, line 38: I feel like 10 seconds is a bit fast, I feel like 1 minute might be plenty, its just a lot of requests. The cost of running a node decreases the amount of users running nodes

trianglesphere: will do. I don't think we actually need that fine grain data b/c we can set alerts based on projected balances (using a linear prediction in grafana).

op-service/metrics/balance.go
trianglesphere: Oh I see, it's 1E18 there. I'm probably gonna use that + `big.NewRat` rather than going through `big.Ints` here.

tynes:
Will approve pending changes implemented

PR #3670, opened 2022-10-07
Action-testing: extend user actor with withdrawal txs

protolambda:
**Description**

Withdrawals depend on the proposer working in action-testing. So this is now split from the main user-actor PR (#3628) to integrate with the proposer actor.

Update the withdrawals RPC client input to use minimal interfaces, and refactor to not tightly couple geth and eth client functionality. This enables the use of an instrumented `client.RPC` and mocks.

~~Depends on #3628~~

~~Depends on #3805~~

**Tests**

Test creates a withdrawal tx on L2, includes it, then makes proposal on L1 for the L2 blocks including the withdrawal, includes those on L1, checks the withdrawal period, and then completes the withdrawal.

Fix ENG-2763

### DIFF
--- a/indexer/integration_tests/bedrock_test.go
+++ b/indexer/integration_tests/bedrock_test.go
@@ -11,6 +11,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/ethclient/gethclient"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/stretchr/testify/require"
+
 	"github.com/ethereum-optimism/optimism/indexer"
 	"github.com/ethereum-optimism/optimism/indexer/db"
 	"github.com/ethereum-optimism/optimism/indexer/services/l1"
@@ -20,12 +29,6 @@ import (
 	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils"
 	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
 	"github.com/ethereum-optimism/optimism/op-node/withdrawals"
-	"github.com/ethereum/go-ethereum/accounts/abi/bind"
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/params"
-	"github.com/ethereum/go-ethereum/rpc"
-	"github.com/stretchr/testify/require"
 
 	_ "github.com/lib/pq"
 )
@@ -196,8 +199,9 @@ func TestBedrockIndexer(t *testing.T) {
 
 		rpcClient, err := rpc.Dial(sys.Nodes["sequencer"].HTTPEndpoint())
 		require.NoError(t, err)
-		proofClient := withdrawals.NewClient(rpcClient)
-		wParams, err := withdrawals.FinalizeWithdrawalParameters(context.Background(), proofClient, wdTx.Hash(), finHeader)
+		proofCl := gethclient.New(rpcClient)
+		receiptCl := ethclient.NewClient(rpcClient)
+		wParams, err := withdrawals.FinalizeWithdrawalParameters(context.Background(), proofCl, receiptCl, wdTx.Hash(), finHeader)
 		require.NoError(t, err)
 
 		l1Opts.Value = big.NewInt(0)
--- a/op-e2e/actions/user.go
+++ b/op-e2e/actions/user.go
@@ -46,16 +46,16 @@ func NewL1Bindings(t Testing, l1Cl *ethclient.Client, deployments *e2eutils.Depl
 type L2Bindings struct {
 	L2ToL1MessagePasser *bindings.L2ToL1MessagePasser
 
-	WithdrawalsClient *withdrawals.Client
+	ProofClient withdrawals.ProofClient
 }
 
-func NewL2Bindings(t Testing, l2Cl *ethclient.Client, withdrawalsCl *withdrawals.Client) *L2Bindings {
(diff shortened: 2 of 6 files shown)


op-e2e/actions/user.go
```diff
@@ -317,6 +319,90 @@ func (s *CrossLayerUser) CheckDepositTx(t Testing, l1TxHash common.Hash, index i
...
+			Data:     params.Data,
+		},
+		params.BlockNumber,
+		params.OutputRootProof,
+		params.WithdrawalProof,
+	)
+	require.NoError(t, err)
+
+	// Send the actual tx (since tx opts don't send by default)
+	err = s.L1.env.EthCl.SendTransaction(t.Ctx(), tx)
```
tynes, line 405: Should this wait for the receipt and ensure no revert?

op-e2e/actions/user.go
```diff
@@ -317,6 +319,90 @@ func (s *CrossLayerUser) CheckDepositTx(t Testing, l1TxHash common.Hash, index i
...
 }
 
+func (s *CrossLayerUser) ActStartWithdrawal(t Testing) {
+	targetAddr := common.Address{}
+	if s.L1.txToAddr != nil {
+		targetAddr = *s.L2.txToAddr
+	}
+	tx, err := s.L2.env.Bindings.L2ToL1MessagePasser.InitiateWithdrawal(&s.L2.txOpts, targetAddr, new(big.Int).SetUint64(s.L1.txOpts.GasLimit), s.L1.txCallData)
+	require.NoError(t, err, "create initiate withdraw tx")
+	err = s.L2.env.EthCl.SendTransaction(t.Ctx(), tx)
```
tynes, line 329: Should this wait on receipt?

protolambda:
Will rebase this onto the proposer work so we can change it from draft to ready

op-e2e/actions/user.go
protolambda: The receipt check is a separate action, so we can cover the unhappy case of failed txs too

op-e2e/actions/user.go
protolambda: Also a separate action, so we can send multiple txs and do other actions, before the block is sealed and receipts are available

protolambda:
Rebased onto `proposer-cleanup` (#3805), we can now test withdrawals once the proposer actor makes its output root proposals.

indexer/integration_tests/bedrock_test.go
```diff
@@ -196,8 +199,9 @@ func TestBedrockIndexer(t *testing.T) {
 
 		rpcClient, err := rpc.Dial(sys.Nodes["sequencer"].HTTPEndpoint())
 		require.NoError(t, err)
-		proofClient := withdrawals.NewClient(rpcClient)
-		wParams, err := withdrawals.FinalizeWithdrawalParameters(context.Background(), proofClient, wdTx.Hash(), finHeader)
+		proofCl := gethclient.New(rpcClient)
+		receiptCl := ethclient.NewClient(rpcClient)
+		wParams, err := withdrawals.FinalizeWithdrawalParameters(context.Background(), proofCl, receiptCl, wdTx.Hash(), finHeader)
```
protolambda, line 204: Bedrock testing modification to indexer, no functionality changes. Some minor refactor to pass the client bindings individually, instead of bundling them in another client binding that was getting too bloated for testing elsewhere.

mslipper (approved):

op-node/withdrawals/utils.go
```diff
@@ -163,9 +145,9 @@ type FinalizedWithdrawalParameters struct {
 // FinalizeWithdrawalParameters queries L2 to generate all withdrawal parameters and proof necessary to finalize an withdrawal on L1.
 // The header provided is very important. It should be a block (timestamp) for which there is a submitted output in the L2 Output Oracle
 // contract. If not, the withdrawal will fail as it the storage proof cannot be verified if there is no submitted state root.
-func FinalizeWithdrawalParameters(ctx context.Context, l2client ProofClient, txHash common.Hash, header *types.Header) (FinalizedWithdrawalParameters, error) {
+func FinalizeWithdrawalParameters(ctx context.Context, proofCl ProofClient, l2ReceiptCl ReceiptClient, txHash common.Hash, header *types.Header) (FinalizedWithdrawalParameters, error) {
```
trianglesphere, line 148: Half of this PR seems to be about changing the interface of this function.

op-e2e/actions/user_test.go
```diff
@@ -6,10 +6,10 @@ import (
...
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/stretchr/testify/require"
 
 	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils"
 	"github.com/ethereum-optimism/optimism/op-node/testlog"
-	"github.com/ethereum-optimism/optimism/op-node/withdrawals"
 )
 
 func TestCrossLayerUser(gt *testing.T) {
```
trianglesphere, line 15: Describing what this tests would be super helpful

op-node/withdrawals/utils.go
mslipper: TBH I like this interface better - the old one required an RPC client to be instantiated, which was annoying because Geth doesn't expose the underlying RPC from the `ethclient.Client` and we always had to instantiate a fresh one :(

trianglesphere: fair, though making a new `gethclient` (the concrete type that satisifies `ProofClient`) still is usually done with an RPC client.

protolambda: The old coupled client made it impossible to instrument (required the concrete type instead of client.RPC interface) and didn't fit into testing for the same reason, hence the changes.

trianglesphere: So could a `client.RPC` have been used in the ProofClient struct?

trianglesphere (approved):
PR description still doesn't mention the refactoring in here.

protolambda:
@trianglesphere updated, thanks

op-node/withdrawals/utils.go
protolambda: Not in the `Client` struct, since we need the proof and receipt methods. Instead this now uses the bare minimum interface, and separates the proof and receipt sources so the user can choose the implementation (for instrumentation, mocking, whatever that may be).

PR #3672, opened 2022-10-07
op-e2e: action testing batcher actor

protolambda:
**Description**

Introduces a batcher as actor in op-e2e.

The batcher state/logic should really be refactored out of the batcher loop to reuse the same model in the batcher actor.
For now I implemented just the bare minimum (buffer and submitting action) to emulate what the batcher would do, so we can at least test the derivation pipeline in action-testing with batches (incl. l1 reorgs etc. too).

This PR depends on #3671

Added `mergify-ignore` label so the dependency PR can be merged into `develop` first.

**Tests**

Simple test that makes a verifier sync a L2 chain from an L1 that included a batch tx with a L2 tx from alice.

**Metadata**

Fix ENG-2862


mslipper (approved):

op-e2e/actions/l2_batcher.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	if err := s.l2ChannelOut.AddBlock(block); err != nil { // should always succeed
+		t.Fatalf("failed to add block to channel: %v", err)
+	}
+}
+
+// ActL2BatchSubmit constructs a batch tx from previous buffered L2 blocks, and submits it to L1
+func (s *L2Batcher) ActL2BatchSubmit(t Testing) {
+	// Don't run this action if there's no data to submit
+	if s.l2ChannelOut == nil || s.l2ChannelOut.ReadyBytes() == 0 {
```
trianglesphere, line 133: I'm surprised that this works. In particular `ReadyBytes()` could be 0 until `l2ChannelOut.Close()` is called.

op-e2e/actions/l2_batcher.go
```diff
@@ -0,0 +1,175 @@
...
+		t.InvalidAction("need to buffer data first, cannot batch submit with empty buffer")
+		return
+	}
+	require.NoError(t, s.l2ChannelOut.Close(), "must close channel before submitting it")
```
trianglesphere, line 137: This will fail if the channel is larger than a single transaction b/c Close returns an error if the channel has already be closed & this function does not have an internal loop to wait until all frames in the channel are submitted.

trianglesphere (commented):
This is fine, but highlights that fact that we're maintaining two different batchers right now until we can directly import the batcher code into this. I'm also a little concerned with the value of this test relative to the setup + framework code.

op-e2e/actions/l2_batcher.go
protolambda: Nice catch. I think it worked because it writes the compression header. I think we can just allow it to continue with an empty l2ChannelOut and then hit the `Close`.

op-e2e/actions/l2_batcher.go
protolambda: I'll separate the `Close` into its own action, so we can run it once, and test/fuzz when we do close it in bad ways.

protolambda: Although soon we may be able to just get rid of the internal logic here, and call an interface of the op-batcher channel manager.

trianglesphere (approved):
lgtm

PR #3674, opened 2022-10-07
op-node: Fillbytes size check in l1 block info

maurelian:
**Description**

Prevents unhandled panics in the case that unexpectedly large values are passed to `FillBytes`.

**Tests**

No new tests, it's not clear that these errors are actually reachable.

**Additional context**

**Metadata**
- Fixes ENG-2613

### DIFF
--- a/op-node/rollup/derive/deposit_log_test.go
+++ b/op-node/rollup/derive/deposit_log_test.go
@@ -23,7 +23,10 @@ func TestUnmarshalLogEvent(t *testing.T) {
 				LogIndex:    uint64(rng.Intn(10000)),
 			}
 			depInput := testutils.GenerateDeposit(source.SourceHash(), rng)
-			log := MarshalDepositLogEvent(MockDepositContractAddr, depInput)
+			log, err := MarshalDepositLogEvent(MockDepositContractAddr, depInput)
+			if err != nil {
+				t.Fatal(err)
+			}
 
 			log.TxIndex = uint(rng.Intn(10000))
 			log.Index = uint(source.LogIndex)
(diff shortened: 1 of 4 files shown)


op-node/rollup/derive/deposit_log_test.go
```diff
@@ -47,8 +50,10 @@ type receiptData struct {
 	DepositLogs []bool
 }
 
-func makeReceipts(rng *rand.Rand, blockHash common.Hash, depositContractAddr common.Address, testReceipts []receiptData) (receipts []*types.Receipt, expectedDeposits []*types.DepositTx) {
+func makeReceipts(rng *rand.Rand, blockHash common.Hash, depositContractAddr common.Address, testReceipts []receiptData) ([]*types.Receipt, []*types.DepositTx, error) {
 	logIndex := uint(0)
+	receipts := []*types.Receipt{}
+	expectedDeposits := []*types.DepositTx{}
```
maurelian, line 56: I realize this is kind of ugly, but the alternative would've been to name the error variable in the function declaration, which seemed like it went against convention in the rest of the codebase.

trianglesphere: we've done it, but I actually prefer explicit rather than named returns

trianglesphere (approved):
lgtm

op-node/rollup/derive/deposit_log_test.go
maurelian: Yeah, this instance of named returns seemed unconventional as well.

mslipper (approved):

PR #3675, opened 2022-10-08
fix(ctb): remove unnecessary L2OO init param

smartcontracts:
**Description**
Removes an unnecessary parameter from the L200 initialization function since this parameter is already set in the constructor.


packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -138,7 +138,27 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
...
+    }
+
+    /**
+     * @notice Initializer.
+     *
+     * @param _genesisL2Output     The initial L2 output of the L2 chain.
+     * @param _proposer            The address of the proposer.
+     * @param _owner               The address of the owner.
+     */
+    function initialize(
```
smartcontracts, line 151: Moving this up to align with other contracts where initialize is underneath the constructor

tynes (approved):

maurelian (approved):

PR #3676, opened 2022-10-08
feat(ctb): simplfy deploy scripts with dictator

smartcontracts:
**Description**
Simplifies the deployment scripts by breaking them out into individual files. Also introduces a new FreshSystemDictator contract which is responsible for executing upgrades for fresh deployments.


packages/contracts-bedrock/contracts/deployment/FreshSystemDictator.sol
```diff
@@ -0,0 +1,110 @@
...
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+import { ProxyAdmin } from "../universal/ProxyAdmin.sol";
+import { L2OutputOracle } from "../L1/L2OutputOracle.sol";
+import { OptimismPortal } from "../L1/OptimismPortal.sol";
+import { L1CrossDomainMessenger } from "../L1/L1CrossDomainMessenger.sol";
+import { L1StandardBridge } from "../L1/L1StandardBridge.sol";
+import { OptimismMintableERC20Factory } from "../universal/OptimismMintableERC20Factory.sol";
+
+/**
+ * @custom:proxied
```
tynes, line 13: shouldn't be proxied

packages/contracts-bedrock/deploy/003-L1CrossDomainMessengerProxy.ts
```diff
@@ -0,0 +1,19 @@
+import { DeployFunction } from 'hardhat-deploy/dist/types'
+import { assertContractVariable, deployAndVerifyAndThen } from '../src/deploy-utils'
+
+const deployFn: DeployFunction = async (hre) => {
+  const { address: proxyAdmin } = await hre.deployments.get('ProxyAdmin')
```
tynes, line 5: Use the `getDeploymentAddress` helper function

packages/contracts-bedrock/contracts/deployment/FreshSystemDictator.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    Config public config;
+
+    constructor(Config memory _config) Ownable() {
+        config = _config;
+        _transferOwnership(config.globalConfig.controller);
+    }
+
+    function step1() external onlyOwner {
+        config.globalConfig.proxyAdmin.upgradeAndCall(
```
tynes, line 63: Can you add comments to these calls? It will make it much easier to read

smartcontracts: Sure

tynes:
Is something in the `.dockerignore` that shouldn't be?

mslipper (approved):

packages/contracts-bedrock/contracts/deployment/FreshSystemDictator.sol
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    Config public config;
+
+    constructor(Config memory _config) Ownable() {
+        config = _config;
+        _transferOwnership(config.globalConfig.controller);
+    }
+
+    function step1() external onlyOwner {
```
tynes, line 61: I believe we do not need to upgrade the L1ERC721Bridge as part of this, just mentioning it to double check

packages/contracts-bedrock/contracts/deployment/FreshSystemDictator.sol
```diff
@@ -0,0 +1,107 @@
...
+        address finalOwner;
+    }
+
+    struct ProxyAddressConfig {
+        address l2OutputOracleProxy;
+        address optimismPortalProxy;
+        address l1CrossDomainMessengerProxy;
+        address l1StandardBridgeProxy;
+        address optimismMintableERC20FactoryProxy;
+    }
```
tynes, line 31: This will need the `SystemConfig` contract as well once it is complete

packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -0,0 +1,298 @@
...
+  await hre.ethers.provider.waitForTransaction(result.transactionHash)
+
+  if (result.newlyDeployed) {
+    if (!(await isHardhatNode(hre)) && hre.network.config.live !== false) {
+      // Verification sometimes fails, even when the contract is correctly deployed and eventually
+      // verified. Possibly due to a race condition. We don't want to halt the whole deployment
+      // process just because that happens.
+      try {
+        console.log('Verifying on Etherscan...')
+        await hre.run('verify:verify', {
```
tynes, line 44: We should double check that this works with the foundry setup by doing a testnet deployment, was getting some weird errors at one point that didn't make sense so i made a hh task for verifications using our foundry setup and it works

packages/contracts-bedrock/contracts/deployment/FreshSystemDictator.sol
```diff
@@ -0,0 +1,107 @@
...
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+import { ProxyAdmin } from "../universal/ProxyAdmin.sol";
+import { L2OutputOracle } from "../L1/L2OutputOracle.sol";
+import { OptimismPortal } from "../L1/OptimismPortal.sol";
+import { L1CrossDomainMessenger } from "../L1/L1CrossDomainMessenger.sol";
+import { L1StandardBridge } from "../L1/L1StandardBridge.sol";
+import { OptimismMintableERC20Factory } from "../universal/OptimismMintableERC20Factory.sol";
+
+/**
```
tynes, line 12: The flow of this contract makes sense to me. Do all of the proxy upgrades in a single step, then we have time to do an offchain check that they worked as expected. We can call `step1` again if anything went wrong. 

One failure case that sticks out to me is if we deploy this contract with the incorrect config. There is no way out of a misconfigured `finalOwner` without redeploying the entire set of contracts to L1. Maybe this isn't a big deal with the fresh dictator since we will not be running it in prod and op-chains should get their config correct at deploy time

packages/contracts-bedrock/deploy/000-ProxyAdmin.ts
```diff
@@ -0,0 +1,24 @@
...
+} from '../src/deploy-utils'
+
+const deployFn: DeployFunction = async (hre) => {
+  const { deployer } = await hre.getNamedAccounts()
+  await deployAndVerifyAndThen({
+    hre,
+    name: 'ProxyAdmin',
+    args: [deployer],
+    postDeployAction: async (contract) => {
+      // Owner is temporarily set to the deployer. We transfer ownership of the ProxyAdmin to the
```
tynes, line 15: I see there is no legacy config of the `ProxyAdmin` which i suppose isn't needed here since this is a fresh deploy

packages/contracts-bedrock/deploy/013-FreshSystemDicator.ts
```diff
@@ -0,0 +1,104 @@
...
+        l2OutputOracleConfig: {
+          l2OutputOracleGenesisL2Output:
+            hre.deployConfig.l2OutputOracleGenesisL2Output,
+          l2OutputOracleProposer: hre.deployConfig.l2OutputOracleProposer,
+          l2OutputOracleOwner: hre.deployConfig.l2OutputOracleOwner,
+        },
+      },
+    ],
+    postDeployAction: async () => {
+      // TODO: Assert all the config was set correctly.
```
tynes, line 82: Saving this todo for a follow up PR?

tynes (approved):
Looks good to me, I looked through it pretty thoroughly, my comments can be resolved, they are more thoughts that I had during review than requests for changes

tynes:
I've had to clone a fresh repo and rebuild in the past to get the bindings to match for some reason, that might solve the issue if you cant get them to match 1:1

packages/contracts-bedrock/deploy/013-FreshSystemDicator.ts
smartcontracts: Yep

tynes:
Looking at the docker build issue

```
#44 [ethereumoptimism/l2geth:latest internal] load build context
#44 transferring context: 10.12MB 1.7s done
#44 ERROR: rpc error: code = Unavailable desc = transport is closing

#26 [ethereumoptimism/batch-submitter-service:latest builder 1/5] FROM docker.io/library/golang:1.18.0-alpine3.15@sha256:a2ca4f4c0828b1b426a3153b068bf32a21868911c57a9fc4dccdc5fbb6553b35
#26 CANCELED

#45 [ethereumoptimism/hardhat-node:latest 4/6] RUN yarn install && yarn cache clean
#0 0.705 yarn install v1.22.19
#0 0.727 info No lockfile found.
#0 0.755 [1/4] Resolving packages...
#45 CANCELED
------
 > [ethereumoptimism/l2geth:latest internal] load build context:
------
------
 > [ethereumoptimism/fault-detector:latest base  2/26] RUN apk --no-cache add curl     jq     python3     ca-certificates     git     make     gcc     musl-dev     linux-headers     bash     build-base     gcompat:
------
failed to solve: failed to compute cache key: extract requires an applier
```

tynes:
There might be something in the `.dockerignore` that matches something that is trying to be copied into the image? The exact error message has no hits on google

PR #3678, opened 2022-10-09
fix typo in spec

xenowits:
**Description**

Fix typo in architecture overview docs.

**Tests**

No tests required.

**Additional context**

None.

**Metadata**

None.

### DIFF
--- a/specs/overview.md
+++ b/specs/overview.md
@@ -135,7 +135,7 @@ Optimism's block derivation function is designed such that it:
 #### Epochs and the Sequencing Window
 
 The rollup chain is subdivided into epochs. There is a 1:1 correspondence between L1 block numbers and epoch numbers.
-For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived a "sequencing windows" has
+For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived after "sequencing windows" has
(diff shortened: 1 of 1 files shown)


tynes (approved):

specs/overview.md
```diff
@@ -135,7 +135,7 @@ Optimism's block derivation function is designed such that it:
 #### Epochs and the Sequencing Window
 
 The rollup chain is subdivided into epochs. There is a 1:1 correspondence between L1 block numbers and epoch numbers.
-For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived a "sequencing windows" has
+For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived after "sequencing windows" has
```
maurelian, line 138: I'm not sure of this phrasing ether: 

```suggestion
For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived after the "sequencing window" has
```

maurelian: @norswap wdyt here?

tynes: We don't want this to go stale @norswap - I think this working is good @maurelian

norswap: I would suggest:

```suggestion

For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived after a *sequencing window* worth of block has
```

PR #3679, opened 2022-10-09
proxyd: Custom rate limiter implementation

mslipper:
Our current proxyd deployment does not share rate limit state across multiple servers within a backend group. This means that rate limits on the public endpoint are artifically high.

This PR adds a Redis-based rate limiter to fix this problem. While our current rate limiting library (github.com/sethvargo/go-limiter) _does_ support Redis, the client library it uses is not type safe, is less performant, and would require us to update the other places we use Redis. To avoid these issues, I created a simple rate limiting interface with both Redis and memory backend.

Note that this PR only adds the new implementations - it does not integrate them with the rest of the codebase. I'll do that in a separate PR to make review easier.


mslipper:
* develop
  * **PR #3679** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3679" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3679?utm_source=stack-comment).

proxyd/frontend_rate_limiter.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return incr.Val()-1 < int64(max), nil
+}
+
+// truncateNow truncates the current timestamp
+// to the specified duration.
+func truncateNow(dur time.Duration) int64 {
+	// pointless to truncate to seconds since
+	// the UNIX timestamp is already in seconds
+	if dur == time.Second {
```
Inphi, line 122: nit: it is nice though to have branchless code.

proxyd/frontend_rate_limiter.go
```diff
@@ -0,0 +1,127 @@
...
+
+type FrontendRateLimiter interface {
+	// Take consumes a key, and a maximum number of requests
+	// per time interval. It returns a boolean denoting if
+	// the limit could be taken, or an error if a failure
+	// occurred in the backing rate limit implementation.
+	//
+	// No error will be returned if the limit could not be taken
+	// as a result of the requestor being over the limit.
+	Take(ctx context.Context, key string, max int) (bool, error)
```
Inphi, line 20: Is there's a usecase that allows a different max value for the same key? If not, we can simplify this interface with a simple `Take(ctx)` by having a mapping of keys to FrontendRateLimiters (w/ a configured max value).

proxyd/frontend_rate_limiter.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// MemoryFrontendRateLimiter is a rate limiter that stores
+// all rate limiting information in local memory. It works
+// by storing a limitedKeys struct that references the
+// truncated timestamp at which the struct was created. If
+// the current truncated timestamp doesn't match what's
+// referenced, the limit is reset. Otherwise, values in
+// a map are incremented to represent the limit.
+type MemoryFrontendRateLimiter struct {
```
Inphi, line 58: Alternatively, you can use https://github.com/kevinms/leakybucket-go. It's an in-memory rate limit used by prysm to rate limit peers and it can work here. The library also supports dynamic rate limits per keys.

Inphi (approved):
Have some comments to improve the implementation. But they aren't blockers.

proxyd/frontend_rate_limiter.go
mslipper: Good callout. There's no use case for this right now - fixed in next PR in the stack.

proxyd/frontend_rate_limiter.go
mslipper: Will check out `leakybucket` - might simplify this. If I think it's suitable I'll open a separate PR.

proxyd/frontend_rate_limiter.go
mslipper: I was concerned with the performance of this, but turns out it doesn't really matter:

```
BenchmarkTrunc
BenchmarkTrunc/trunc
BenchmarkTrunc/trunc-8         	33962488	        35.38 ns/op
BenchmarkTrunc/now
BenchmarkTrunc/now-8           	28988336	        40.55 ns/op
PASS
``

mslipper: Ironically the `trunc` is slightly faster 🤷

mslipper: Updated to remove the branching.

PR #3680, opened 2022-10-09
fix(ctb): move inits under constructors

smartcontracts:
**Description**
Standardizes the location of initializers to be under constructors.

### DIFF
--- a/.changeset/friendly-rats-scream.md
+++ b/.changeset/friendly-rats-scream.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/contracts-bedrock': patch
+---
+
+Moves initializers underneath constructors always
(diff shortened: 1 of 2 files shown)


op-bindings/bindings/addressmanager.go
```diff
@@ -156,11 +157,11 @@ func NewAddressManagerFilterer(address common.Address, filterer bind.ContractFil
 
 // bindAddressManager binds a generic wrapper to an already deployed contract.
 func bindAddressManager(address common.Address, caller bind.ContractCaller, transactor bind.ContractTransactor, filterer bind.ContractFilterer) (*bind.BoundContract, error) {
-	parsed, err := abi.JSON(strings.NewReader(AddressManagerABI))
```
tynes, line 159: This looks like a new version of abigen was used locally, we should update in our build

smartcontracts: Hmm... I'll just use the old version of abigen for now.

tynes (approved):

mslipper (approved):

PR #3683, opened 2022-10-10
fix(proxyd): Fix compliance with JSON-RPC 2.0 spec by adding optional RPCError.Data

chriswessels:
**Description**

The JSON-RPC 2.0 specification indicates that the error object may contain an optional data field. Proxyd currently drops this field from any responses. This PR fixes that.

https://www.jsonrpc.org/specification#error_object

**Tests**

I added a test to check that if a data field is present in the response from a backend, it will be added to the response from proxyd. The existing test should validate that if data field is not present in the response from the backend, the data key should not be present in the JSON response from proxyd.

### DIFF
--- a/.changeset/angry-wombats-marry.md
+++ b/.changeset/angry-wombats-marry.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/proxyd': minor
+---
+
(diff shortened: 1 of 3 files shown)


mslipper (approved):
Nice work, thank you!

PR #3694, opened 2022-10-12
op-node: Fix reset bug in batch queue

trianglesphere:
**Description**

The batch queue was using the previous origin when computing if the origin was behind or not. It would then immediately advance it's internal origin to the next origin (confusingly coming from the `prev` stage). It would take action on the current origin based on out of date data. Specifically, it would not save an origin that should have been saved into the l1Blocks array. This causes errors because the L1 Origin of the safe head would not be present when attempting to derive the next batch after a reset.

**Tests**

Not tested either in a full reproduction or with unit tests. 

**Metadata**
- Fixes ENG-2906

### DIFF
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -59,7 +59,11 @@ func (bq *BatchQueue) Origin() eth.L1BlockRef {
 }
 
 func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) (*BatchData, error) {
-	originBehind := bq.origin.Number < safeL2Head.L1Origin.Number
+	// Note: We use the origin that we will have to determine if it's behind. This is important
+	// because it's the future origin that gets saved into the l1Blocks array.
+	// We always update the origin of this stage if it is not the same so after the update code
+	// runs, this is consistent.
+	originBehind := bq.prev.Origin().Number < safeL2Head.L1Origin.Number
 
 	// Advance origin if needed
 	// Note: The entire pipeline has the same origin


packages/contracts-bedrock/deploy/001-InitImplementations.deploy.ts
```diff
@@ -7,7 +7,7 @@ import '@eth-optimism/hardhat-deploy-config'
 
 const upgradeABIs = {
   L2OutputOracleProxy: async (deployConfig) => [
-    'initialize(bytes32,uint256,address,address)',
```
protolambda, line 10: Is this a related change / fix? Seems worth a comment in PR description or a separate PR. And is it required for CI?

protolambda (commented):
LGTM. A test to reproduce it would be nice to have (or an issue for future test, I think we can cover it with action-testing). And there's an unrelated change in the PR?

packages/contracts-bedrock/deploy/001-InitImplementations.deploy.ts
trianglesphere: I would like to write a unit test for this. I was waiting on doing a full deploy to make sure that this fixed the issue. This specific line was blocking a full deploy - we updated the ABI but did not change this so hardhat would choke on deployments.

trianglesphere: @mslipper when did this get fixed on develop? This is pretty up to date & the PR that broke it is about a week old

trianglesphere: Oh, I found the old PR. This commit is dropped.

PR #3695, opened 2022-10-12
feat(commonts): Make base service v2 api typesafe by default with trpc

roninjin10:
I add trpc which is the best way to create APIs with typescript that are meant to be consumed by other typescript services or UI.

1. Create an API object with BaseServiceV2

``` typescript
const trpcApi = BaseServiceV2.createApi()
```

2. Register a unique endpoint on it
``` typescript
const trpcApi = BaseServiceV2.query('user', {
})
```
3. Each query takes in json object as parameter.  Use zod to validate it and to automatically infer typescript types

``` typescript
const trpcApi = BaseServiceV2.createAPI().query('user', {
  input: z.object({
    id: z.string(),
  }),
})
```
4. Add a resolver and you are done!  You created an endpoint that can be consumed on the frontend with no further boilerplate!   The entire feature will be typesafe e2e.
``` typescript
const trpcApi = BaseServiceV2.createAPI().query('user', {
  /**
   * Zod will validate input and also allow typescript to infer types
   */
  input: z.object({
    id: z.string(),
  }),
  // Resolve the endpoint here
  resolve: async (req) => {
    // this request will be typesafe
    return users.getFromDb<User>(req.id)
  },
})
```

The client can then import the Router type.  This only imports the typescript type and no actual server code is included.   From the typescript type, the client can safely call the API with the correct params and be guaranteed to be in sync with the server.   This will make it stupidly quick and simple for any dev to build a full stack feature.


packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -525,13 +564,6 @@ export abstract class BaseServiceV2<
    */
   protected init?(): Promise<void>
 
-  /**
-   * Initialization function for router.
-   *
-   * @param router Express router.
-   */
-  protected routes?(router: ExpressRouter): Promise<void>
```
roninjin10, line 533: bumped major version because I removed this

roninjin10:
No rush to review this can wait til after devcon

roninjin10:
trpc version 10 though in beta is much better than version 9 and has a lot of breaking changes.  Putting this in draft because we should start with version 10 instead

PR #3700, opened 2022-10-13
proxyd: Allow disabling backend rate limiting

mslipper:
The backend rate limiter is in place to protect upstreams like the sequencer. However, in many cases it isn't needed and it causes unnecessary requests to Redis. This PR allows this to be disabled, and disables this by default.


zhwrd (approved):

tynes:
Looks like tests are failing

tynes (approved):

PR #3703, opened 2022-10-13
op-e2e: Transaction validity tests

trianglesphere:
**Description**

This PR adds two transaction validity tests. One around invalid deposits that enforces existing behavior (should be migrated to hive when we start properly support multiple execution engines). The second highlights an issue I found it block production.

**FCU vs New Payload Discrepancy**

Fork Choice Update uses the mining code to create the block. The
mining code does not share the EVM context when building the block
(unlike the Process code which is used by the NewPayload call). I
was able to trigger a discrepancy between these two codepaths by
updating the GPO parameters in a normal L2 transaction.

By decreasing the GPO cost in a normal L2 tx that is just before
a simple transfer, the mining code uses the new GPO parameters
while the NewPayload verification uses the old parameters. This
means that the block that is produced by the miner includes an
invalid transaction.

I have not done this, but the inverse could be done, by increasing
the L1 cost past what an account is able to afford, you can construct
a L2 block that would not be created by ForkChoiceUpdated but would
be validated by NewPayload. I think that this case is more dangerous
because NewPayload is used when receiving transactions from the
sequencer but ForkChoiceUpdated is used when creating the block
locally (as in a replica). Thus the sequencer could construct a block
that unconnected replicas would not be able to create (but connected
replicas could accept).

Note that reconciliation would still be fine because it only looks at
the block bodies an assumes that the blocks produced by the mining
code will always be valid.

**Additional context**

Depends upon
- https://github.com/ethereum-optimism/optimism/pull/3702
- https://github.com/ethereum-optimism/op-geth/pull/14

Once the op-geth PR is merged, I can update the version of geth that this uses
and the tests will pass expecting a status `VALID` instead of expecting `INVALID`.

### DIFF
--- a/op-e2e/go.mod
+++ b/op-e2e/go.mod
@@ -3,6 +3,7 @@ module github.com/ethereum-optimism/optimism/op-e2e
 go 1.18
 
 require (
+	github.com/ethereum-optimism/optimism/gas-oracle v0.0.0-20221013175259-1108348a742a
 	github.com/ethereum-optimism/optimism/op-batcher v0.8.10
 	github.com/ethereum-optimism/optimism/op-bindings v0.8.10
 	github.com/ethereum-optimism/optimism/op-chain-ops v0.8.10
--- a/op-e2e/op_geth_test.go
+++ b/op-e2e/op_geth_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ethereum-optimism/optimism/gas-oracle/bindings"
 	"github.com/ethereum-optimism/optimism/op-bindings/predeploys"
 	"github.com/ethereum-optimism/optimism/op-chain-ops/genesis"
 	"github.com/ethereum-optimism/optimism/op-node/client"
@@ -15,8 +16,12 @@ import (
 	"github.com/ethereum-optimism/optimism/op-node/sources"
 	"github.com/ethereum-optimism/optimism/op-node/testlog"
 
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/log"
 	gn "github.com/ethereum/go-ethereum/node"
 	"github.com/ethereum/go-ethereum/params"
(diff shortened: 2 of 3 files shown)


op-e2e/op_geth_test.go
```diff
@@ -0,0 +1,268 @@
...
+	// We also need to setup a L1 Genesis to create the rollup genesis.
+	log := testlog.Logger(t, log.LvlCrit)
+	cfg := DefaultSystemConfig(t)
+	cfg.DeployConfig.FundDevAccounts = false
+
+	l1Genesis, err := genesis.BuildL1DeveloperGenesis(cfg.DeployConfig)
+	require.Nil(t, err)
+	l1Block := l1Genesis.ToBlock()
+
+	l2Addrs := &genesis.L2Addresses{
```
tynes, line 46: I updated the impl of `BuildL2DeveloperGenesis` so you can pass in a `nil` value for `l2Addrs` and it'll use all the defaults, this is useful if `L2Addresses` changes as a struct

op-e2e/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/ethereum-optimism/optimism/op-e2e
 go 1.18
 
 require (
+	github.com/ethereum-optimism/optimism/gas-oracle v0.0.0-20221013175259-1108348a742a
```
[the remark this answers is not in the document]
tynes: The `GasPriceOracle` contract should also be in the `op-bindings` package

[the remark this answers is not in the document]
trianglesphere: hmm, I'm not sure. I'll look into it.

op-e2e/op_geth_test.go
```diff
@@ -0,0 +1,268 @@
...
+	payload, err := client.GetPayload(ctx, *res.PayloadID)
+	require.Nil(t, err)
+	require.NotNil(t, payload)
+	require.Equal(t, payload.Transactions, attrs.Transactions) // Ensure we don't drop the transactions
+
+	status, err := client.NewPayload(ctx, payload)
+	require.Nil(t, err)
+	require.Equal(t, eth.ExecutionInvalid, status.Status)
+
+	// fc.HeadBlockHash = payload.BlockHash
```
tynes, line 263: I'm guessing this needs a little more cleanup before its ready?

trianglesphere: So this line isn't strictly necessary, but is there for completeness. The part that does need to be updated when the op-geth PR is merged is the above line that changes from `ExecutionInvalid` to `ExecutionValid`.

op-e2e/op_geth_test.go
```diff
@@ -0,0 +1,268 @@
...
+	res, err = client.ForkchoiceUpdate(ctx, &fc, nil)
+	require.Nil(t, err)
+	require.Equal(t, eth.ExecutionValid, res.PayloadStatus.Status)
+
+}
+
+// TestFCUVSNewPayloadDiscrepency attempts to exploit a difference in block creation between
+// FCU and NewPayload. It specifically looks at the fact that NewPayload uses a shared EVMContext
+// for the full block while FCU (and mining code) uses a new EVMContext for each transaction.
+func TestFCUVSNewPayloadDiscrepency(t *testing.T) {
```
protolambda, line 143: Maybe this can be split up in another PR? The other test is useful and can be merged, but this GPO update test may not be applicable if we continue with the system config approach to updating GPO params.

trianglesphere:
Closing because we are changing how we do GPO updates which makes this discrepancy impossible to exploit.

PR #3705, opened 2022-10-15
op-batcher: Concurrently send transactions

trianglesphere:
**Description**

This is a simple hack to enable significantly higher throughput for lattice with op-craft.

**Tests**

I tested this locally by modifying the max tx size argument in the op-e2e tests & confirming that I see expected transaction sending behavior (by looking at the logs).

**Metadata**
- Fixes ENG-2898

### DIFF
--- a/op-batcher/batch_submitter.go
+++ b/op-batcher/batch_submitter.go
@@ -354,7 +354,7 @@ mainLoop:
 				// Query for the submitter's current nonce.
 				walletAddr := crypto.PubkeyToAddress(l.cfg.PrivKey.PublicKey)
 				ctx, cancel = context.WithTimeout(l.ctx, time.Second*10)
-				nonce, err := l.cfg.L1Client.NonceAt(ctx, walletAddr, nil)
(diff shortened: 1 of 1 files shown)


protolambda (approved):

tynes:
We are guaranteed that the transactions will never revert if they make it into the mempool since there is no code at the account we are sending to. If we decide to add a contract that does the owner management instead of being able to dynamically + deterministically change the batch sender in the op-node, there is some complexity regarding reverted transactions, ie how do you know to resend the same batch if it happens to revert?

trianglesphere:
> If we decide to add a contract that does the owner management instead of being able to dynamically + deterministically change the batch sender in the op-node, there is some complexity regarding reverted transactions, ie how do you know to resend the same batch if it happens to revert?

There probably is, but we're ignoring it for now. This is only getting applied to the lattice branch and won't be brought into our production environment

trianglesphere:
@mslipper

trianglesphere:
Turning into a draft to hide this until opcraft is over or we actually need this.

trianglesphere:
no longer needed.

PR #3707, opened 2022-10-16
op-node: Add P2P metrics

mslipper:
Adds the following peer-to-peer metrics:

- Peer count
- Stream count
- Number of gossip events by type
- Peer-to-peer bandwidth usage by direction

I used libp2p's existing interfaces wherever I could. Note that we still need a backgorund goroutine to pull banwdiwth stats out of the libp2p bandwidth counter.

Meta:

- Fixes ENG-2338
- Fixes ENG-2336
- Fixes ENG-2329

### DIFF
--- a/op-node/metrics/metrics.go
+++ b/op-node/metrics/metrics.go
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/libp2p/go-libp2p-core/metrics"
+	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -66,6 +68,12 @@ type Metrics struct {
 	L1ReorgDepth prometheus.Histogram
 
 	TransactionsSequencedTotal prometheus.Counter
+
+	// P2P Metrics
+	PeerCount         prometheus.Gauge
+	StreamCount       prometheus.Gauge
+	GossipEventsTotal *prometheus.CounterVec
+	BandwidthTotal    *prometheus.GaugeVec
 
 	registry *prometheus.Registry
 }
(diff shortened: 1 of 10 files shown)


op-node/metrics/metrics.go
```diff
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/libp2p/go-libp2p-core/metrics"
```
protolambda, line 13: It may be helpful to rename this to `p2pmetrics` or something that is less easily confused with the local package name.

op-node/p2p/gossip.go
```diff
@@ -115,7 +116,7 @@ func BuildGlobalGossipParams(cfg *rollup.Config) pubsub.GossipSubParams {
 	return params
 }
 
-func NewGossipSub(p2pCtx context.Context, h host.Host, cfg *rollup.Config) (*pubsub.PubSub, error) {
+func NewGossipSub(p2pCtx context.Context, h host.Host, cfg *rollup.Config, metrics *metrics.Metrics) (*pubsub.PubSub, error) {
```
protolambda, line 119: Would prefer to use an interface here to keep it more testable in isolation

op-node/p2p/gossip.go
```diff
@@ -441,3 +443,11 @@ func LogTopicEvents(ctx context.Context, log log.Logger, evHandler *pubsub.Topic
...
 	}
 }
+
+type gossipTracer struct {
+	m *metrics.Metrics
+}
+
+func (g *gossipTracer) Trace(evt *pb.TraceEvent) {
+	g.m.RecordGossipEvent(int32(*evt.Type))
+}
```
protolambda, line 453:
```suggestion
func (g *gossipTracer) Trace(evt *pb.TraceEvent) {
    if g.m != nil {
	    g.m.RecordGossipEvent(int32(*evt.Type))
	}
}
```
We could make the metrics optional like this.

op-node/p2p/host_test.go
```diff
@@ -132,7 +132,7 @@ func TestP2PFull(t *testing.T) {
 	// TODO: maybe swap the order of sec/mux preferences, to test that negotiation works
 
 	logA := testlog.Logger(t, log.LvlError).New("host", "A")
-	nodeA, err := NewNodeP2P(context.Background(), &rollup.Config{}, logA, &confA, &mockGossipIn{})
+	nodeA, err := NewNodeP2P(context.Background(), &rollup.Config{}, logA, &confA, &mockGossipIn{}, metrics.NewMetrics(""))
```
protolambda, line 135: See above comment about optional metrics. We can then pass `nil` in unit tests when metrics don't interest us.

op-node/p2p/notifications.go
```diff
@@ -11,6 +12,7 @@ import (
 
 type notifications struct {
 	log log.Logger
+	m   *metrics.Metrics
```
protolambda, line 15: Interface and optionality of metrics would be nice here.

op-node/p2p/gossip.go
protolambda: And maybe not in this PR, but based on the event type we can decide to collect other metrics. E.g. if a peer subscribes to a topic, we can track the number of peers on that topic. Or e.g. update a gauge with number of peers in our mesh.

trianglesphere (approved):

protolambda (commented):
Looks great. Some nits on interfaces/optionality, then good to merge.

protolambda (approved):

Issue #3343, opened 2022-09-07
Include instructions to install abigen globally


smartcontracts:
**Describe the bug**
Monorepo expects that you have abigen installed if you want to rebuild bindings (which is necessary for certain contract changes). I think it'd be ideal if the monorepo explicitly instructed you to install abigen if you intend to modify smart contract code.

PR #3714, opened 2022-10-17
ci-builder: update geth version

tynes:
Updates the version of `abigen` that is used in the `ci-builder` so that it produces bindings that do not parse JSON every time that an instance of the binding is created.

The change in geth is here: https://github.com/ethereum/go-ethereum/pull/25574

Fixes the additional diff found here: https://github.com/ethereum-optimism/optimism/pull/3680/files

Right now we do not have a guarantee that local devs use a specific version of `abigen`. After this is merged, all devs would need to use `v1.10.25`. Devs could use higher versions but risk running into problems if there are more changes to `abigen` in the future.


tynes:
> This is a cool feature, but depends on a very recent version of geth & abigen. Getting this merged seems worth it, but it needs to be documented in where we list setup requirements & communicated to the whole team.

Good call, I was trying to fix #3680 

I can add something to the docs

smartcontracts:
We really need to add this to the docs. #3343 talks about this issue too.

tynes:
I documented how to install `abigen` as well as the specific version that we should use after this PR is merged and the new `ci-builder` image is released

trianglesphere (approved):
lgtm

maurelian (approved):

PR #3715, opened 2022-10-17
wip: batch portal

tynes:
**Description**

Simple huff contract for the batch portal. It is `ownable` and allows us to later upgrade to an offchain leader election mechanism.
Tests don't build yet because of a missing dep

### DIFF
--- a/packages/contracts-bedrock/contracts/L1/BatchPortal.huff
+++ b/packages/contracts-bedrock/contracts/L1/BatchPortal.huff
@@ -1,0 +1,57 @@
+/// @title BatchPortal
+/// @notice SPDX-License-Identifier: MIT
+
+// Interface
+#define function transferOwnership(address) nonpayable returns ()
+
+// Events
+#define event OwnershipTransferred(address indexed previousOwner, address indexed newOwner)
+
+// Storage Slots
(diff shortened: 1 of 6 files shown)


packages/contracts-bedrock/contracts/L1/BatchPortal.huff
```diff
@@ -0,0 +1,57 @@
...
+  caller                      // [from, newOwner]
+  __EVENT_HASH(OwnerUpdated)  // [sig, from, newOwner]
+  0x00 0x00                   // [0, 32, sig, from, newOwner]
+  log3                        // []
+
+  stop
+}
+
+#define macro FALLBACK() = takes (0) returns (0) {
+  IS_OWNER()
```
maurelian, line 46: So the sequencer is the owner? But the owner account can also call `transferOwner`, so in the event of the sequencer key being stolen, this won't help us much. There should be a separate account and auth (ie. `IS_ADMIN`) which would probably be held by the multisig.

tynes: Good call

tynes:
I think we want to check the gas cost diff between this impl and a yul based impl

mslipper:
Closing this since we've since moved on to a different architecture.

PR #3717, opened 2022-10-17
op-node: Initialize L1State at driver startup

trianglesphere:
**Description**

This avoids issues with the confirmation depth.
If this returns an error, it exits via CLI.

**Metadata**

Fixes ENG-2920

### DIFF
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -75,6 +75,26 @@ type Driver struct {
 func (s *Driver) Start() error {
 	s.derivation.Reset()
 
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	l1Head, err := s.l1.L1BlockRefByLabel(ctx, eth.Unsafe)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


mslipper (approved):

trianglesphere:
Converted to draft b/c it looks like this is using the L1 client before it's fully available or the L1 client does not have a head block at startup.

protolambda (commented):
Can we not handle the initial zero values when we read them? It would be nice to instantly start up, without failure edge cases, by avoiding these RPC calls.

trianglesphere:
> Can we not handle the initial zero values when we read them? It would be nice to instantly start up, without failure edge cases, by avoiding these RPC calls.

There is an underlying problem that this interacts with the `conf_depth` to reject valid getBlockByNumber which causes problems with the initial find sync start (particularly with large . On solution is to use a non-conf depth fetcher in the right places, another is to add a fourth condition here: https://github.com/ethereum-optimism/optimism/blob/e97a9adec87fbcd9f4326ee34edbd90b1a4d5e03/op-node/rollup/driver/conf_depth.go#L33-L36 It is possible to write code with the assumption that there is a zero value, but I think it adds a fair bit of complexity.

PR #3719, opened 2022-10-17
op-e2e: user action allow set calldata

tynes:
**Description**

This commit adds a function to the `BasicUser` that allows a more complex flow where the calldata can be set. This allows for the user to make specific contract calls.

### DIFF
--- a/op-e2e/actions/user.go
+++ b/op-e2e/actions/user.go
@@ -149,6 +149,13 @@ func (s *BasicUser[B]) ActRandomTxToAddr(t Testing) {
 	s.txToAddr = to
 }
 
+func (s *BasicUser[B]) ActSetTxCalldata(calldata []byte) Action {
+	return func(t Testing) {
+		require.NotNil(t, calldata)
+		s.txCallData = calldata
+	}
+}
+
 func (s *BasicUser[B]) ActSetTxToAddr(to *common.Address) Action {
 	return func(t Testing) {
 		s.txToAddr = to


trianglesphere (approved):

op-e2e/actions/user.go
```diff
@@ -149,6 +149,13 @@ func (s *BasicUser[B]) ActRandomTxToAddr(t Testing) {
 	s.txToAddr = to
 }
 
+func (s *BasicUser[B]) ActSetTxCalldata(calldata []byte) Action {
+	return func(t Testing) {
+		require.NotNil(t, calldata)
+		s.txCallData = calldata
```
trianglesphere, line 155: two things I just found by reading this code. I don't think we actually send `s.txCalldata` (line 205/212 - it seems to only be used in the gas estimation). Second is that if we require non-nill calldata we should have a way to reset the calldata b/c a basic user can send multiple transactions.

mslipper (approved):

op-e2e/actions/user.go
trianglesphere: Upon further review, we set use txCalldata with deposits for a crossLayer user but not for a basic user & still don't have a way to zero it (other than length 0 but non nil data array).

tynes: Good catch - I was going crazy because things were not going as expected when writing the tests and couldn't figure out why

tynes:
@protolambda This now sends the calldata in the generated tx and then resets the calldata to being empty. I feel like the same place where we reset the calldata we should reset the call value, but that is out of scope of this PR

protolambda (approved):
Not sure yet what the best pattern for testing / fuzzing is: prepare options once, and reuse + reset explicitly. Or prepare once, and reset back to defaults implicitly (at end of action).

Either way is fine for now though, so LGTM.

PR #3725, opened 2022-10-17
op-batcher: Limit number of blocks per channel

trianglesphere:
**Description**

This limits the number of L2 blocks submitted per channel when the batcher
is submitting a large number of L2 blocks at once (like when it is catching
up to a far ahead safe head).


op-batcher/batch_submitter.go
```diff
@@ -313,7 +313,14 @@ mainLoop:
 				l.ch = ch
 			}
 			prevID := l.lastSubmittedBlock
-			for i := l.lastSubmittedBlock.Number + 1; i <= syncStatus.UnsafeL2.Number; i++ {
+			maxBlocksPerChannel := uint64(100)
```
protolambda, line 316: Should we make this configurable? Or increase it? Blocks can be really small.

trianglesphere: It's a would be nice, but this PR was focused on fixing the short term issue & we're going to do a bunch of work on this in a short bit as well.

mslipper (approved):

protolambda (approved):

PR #3731, opened 2022-10-18
specs,op-node: Clarify Max RLP Bytes Per Channel

trianglesphere:
**Description**

This pulls it out into it's own constant inside the op-node. This also clarifies where the limit is applied and the implications that the limit has on what batches can be accepted.

**Metadata**
- Fixes ENG-2882

### DIFF
--- a/op-node/rollup/derive/channel.go
+++ b/op-node/rollup/derive/channel.go
@@ -138,7 +138,7 @@ func BatchReader(r io.Reader, l1InclusionBlock eth.L1BlockRef) (func() (BatchWit
 	if err != nil {
 		return nil, err
 	}
-	rlpReader := rlp.NewStream(zr, 10_000_000)
+	rlpReader := rlp.NewStream(zr, MaxRLPBytesPerChannel)
 	// Read each batch iteratively
 	return func() (BatchWithL1InclusionBlock, error) {
 		ret := BatchWithL1InclusionBlock{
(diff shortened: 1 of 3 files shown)


tynes (approved):

op-node/rollup/derive/channel.go
```diff
@@ -138,7 +138,7 @@ func BatchReader(r io.Reader, l1InclusionBlock eth.L1BlockRef) (func() (BatchWit
 	if err != nil {
 		return nil, err
 	}
-	rlpReader := rlp.NewStream(zr, 10_000_000)
+	rlpReader := rlp.NewStream(zr, MaxRLPBytesPerChannel)
```
protolambda, line 141: Clarification looks good to me, but can we maybe apply this limit to the writer too?

In here when it writes the batch to the compression stream:
https://github.com/ethereum-optimism/optimism/blob/936b8ba5fe354f73c8e99fc38275d1175a1ff547/op-node/rollup/derive/channel_out.go#L78

We could count the bytes that are written to this stream, and return a special error if there are too many bytes, so we never create channels that are too large, and start new channels instead. I can open a separate PR for this if you like, but I think it integrates well with this one.

trianglesphere: I rather have it be part of a separate PR. I'm happy to go make the new PR as well..

protolambda: Ok, will open an issue in linear for this then, thanks.

protolambda (approved):

PR #3734, opened 2022-10-19
ops: Add bedrock-devnet script

mslipper:
Adds a bedrock-devnet script that deploys a local network using the Hardhat deploy scripts. I opted to write this in Python rather than Bash because we're starting to add complexity to our devnet setup, and I find Python much easier and safer to maintain.

Over time, I'll migrate the other devnet-up script to use this Python script as well.

To invoke this script, run `make devnet-up-deploy`. It's included in CI already as a separate testing job.

### DIFF
--- a/packages/sdk/tasks/deposit-erc20.ts
+++ b/packages/sdk/tasks/deposit-erc20.ts
@@ -1,3 +1,5 @@
+import { promises as fs } from 'fs'
+
 import { task, types } from 'hardhat/config'
 import { HardhatRuntimeEnvironment } from 'hardhat/types'
 import '@nomiclabs/hardhat-ethers'
@@ -8,7 +10,13 @@ import {
 } from '@eth-optimism/contracts-bedrock'
 import { Event, Contract, Wallet, providers, utils } from 'ethers'
 
-import { CrossChainMessenger, MessageStatus, CONTRACT_ADDRESSES } from '../src'
+import {
+  CrossChainMessenger,
+  MessageStatus,
+  CONTRACT_ADDRESSES,
+  OEContractsLike,
+  DEFAULT_L2_CONTRACT_ADDRESSES,
+} from '../src'
 
 const deployWETH9 = async (
   hre: HardhatRuntimeEnvironment,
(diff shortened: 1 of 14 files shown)


packages/sdk/tasks/deposit-erc20.ts
```diff
@@ -127,7 +141,14 @@ task('deposit-erc20', 'Deposits WETH9 onto L2.')
     )
 
     const l2ChainId = await l2Signer.getChainId()
-    const contractAddrs = CONTRACT_ADDRESSES[l2ChainId]
+    let contractAddrs = CONTRACT_ADDRESSES[l2ChainId]
+    if (args.l1ContractsJsonPath) {
+      const data = await fs.readFile(args.l1ContractsJsonPath)
```
tynes, line 146: You can `require` a json file right into an object

packages/sdk/tasks/deposit-eth.ts
```diff
@@ -70,7 +84,14 @@ task('deposit-eth', 'Deposits WETH9 onto L2.')
     )
 
     const l2ChainId = await l2Signer.getChainId()
-    const contractAddrs = CONTRACT_ADDRESSES[l2ChainId]
+    let contractAddrs = CONTRACT_ADDRESSES[l2ChainId]
```
tynes, line 87: If its easier to not use the hh deploy artifacts I understand but then this becomes very tied to the way we do it in CI instead of being able to work against live networks as well

mslipper: I don't see why this breaks against live networks. All this code does is let me pass in a JSON with addresses in it rather than using the contract addresses hardcoded in the SDK. We'd have to traverse the HH deploy dir and create this JSON anyway to get the `contracts` arg the SDK expects.

packages/sdk/tasks/deposit-erc20.ts
mslipper: Hmm, require doesn't seem to work in this case - reverting to use `fs` like before.

tynes (approved):

PR #3742, opened 2022-10-19
l2geth: Add OVM ETH and message passer state dump

mslipper:
This is necessary in order to perform the mainnet OVM ETH and withdrawal hashes migration. Operates using a L2GETH_STATE_DUMP_PATH environment variable that specifies the file that addresses should be dump to. They are not deduplicated. The format of the file is:

```
ETH|<ovm eth address>
MSG|<sender>|<msg>
```

### DIFF
--- a/l2geth/core/state/statedb.go
+++ b/l2geth/core/state/statedb.go
@@ -24,6 +24,8 @@ import (
 	"sort"
 	"time"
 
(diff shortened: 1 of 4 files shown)


tynes (approved):

mslipper:
@smartcontracts reviewed this live with me. Merging.

Issue #3606, opened 2022-09-29
Optimism - Release linux/arm64 builds


andreclaro:
## Description
Release`linux/arm64` builds for optimism services, such as, batch-submitter-service, data-transport-layer, deployer, fault-detector, gas-oracle, hardhat, l2geth and message-relayer,

## Why this feature is important
- Take advantages of running the nodes on top Arm CPU architecture machines, such as, AWS Graviton2/3. The main benefits are costs reduction and performance increase. For further details consult: https://aws.amazon.com/ec2/graviton and https://about.gitlab.com/blog/2021/08/05/achieving-23-cost-savings-and-36-performance-gain-using-gitlab-and-gitlab-runner-on-arm-neoverse-based-aws-graviton2-processor
" Our latest cost analysis and performance benchmarks show that customers can realize cost savings of up to 23% and performance gains of up to 36% by deploying the GitLab application and GitLab Runner on the Arm-based Graviton2 when compared to the x86 based M5/C5 EC2 instances." 

- Enable developers to run `stellar-core` on their MacBook Pro M1/M2 laptops.

michaelneuder:
Taking a look at this! 

I see two main options for building linux/arm64 compatible containers:

1. We add a new job that is similar to `docker-publish` to the circleci config (call it like `docker-arm-publish`) https://github.com/ethereum-optimism/optimism/blob/98035730167fd5698d0436f1a170586c4625823b/.circleci/config.yml#L87-L110 The difference is that we use the arm execution environment https://circleci.com/docs/using-arm/. If we set resource class to `resource_class: arm.xlarge` then we should be provisioning a GCE instance that is compatible with ARM (i think that is what circle is doing under the hood?). I tried this with a test `t2a-standard-4` instance and was able to create a test `hardhat-node` image which is listed as `linux/arm64/v8` arch: https://hub.docker.com/layers/mikeneuder/hardhat-node/1.0/images/sha256-08d3fb84f0c4717155abf1fc88eceb094e35a051cc8fa6952225236649a67c9d?context=repo. 
2. We use `dockerx` in the CI workflow to build cross architecture images: https://www.docker.com/blog/getting-started-with-docker-for-arm-on-linux/. IIUC this builds a single container that can run on either architecture. This doesn't really feel like the problem we are trying to solve, but I am not sure. 

IMO option 1 sounds more like what we want to do, but I wanted to touch base here.

BlocksOnAChain:
closing since it's no longer an active issue.

PR #3757, opened 2022-10-21
Adding support for publishing a linux/arm64 build to docker hub.

michaelneuder:
**Description**

https://github.com/ethereum-optimism/optimism/issues/3606 requests that linux/arm64 versions of the docker containers are published to docker hub. This PR modifies the `docker-publish` job template to be parameterized by a linux image and a a resource. It also creates a new job called `op-node-publish-arm-dev` which builds and publishes an arm64 version of the `op-node` container. 

This is option 1 as described in https://github.com/ethereum-optimism/optimism/issues/3606#issuecomment-1283153169. See https://circleci.com/docs/using-arm/ for details on the arm execution environments provided by circle (namely note that there are fewer images with arm support https://circleci.com/docs/using-arm/#images-with-arm-support, which requires us to use a different ubuntu image than is used for the amd64 build). 

**Tests**

I am not sure how the circleci workflow is triggered, which is why I only added the `op-node` container job (just to sanity check the method). If everything WAI, we can add the jobs for building arm versions of the batcher and proposer containers.

**Additional context**

**Metadata**
https://github.com/ethereum-optimism/optimism/issues/3606


zhwrd:
Hey @michaelneuder, thanks for putting this pull request up! For us to feel comfortable with releasing arm64 support, we'd like to ensure our test pipelines run against both amd64 and arm64 platforms as well.

Unfortunately because CircleCI only offers VM executors for arm64 while our existing test jobs use docker images, this is not straightforward to re-tool.

The team is heads down on getting the bedrock release out right now and won't be able to take this on for quite some time. Again, thanks for your contribution and we will revisit this in the future.

PR #3762, opened 2022-10-21
fix(l2g): bug in state dumper

smartcontracts:
**Description**
Fixes a bug in the state dumper. os.OpenFile was using the wrong file descriptors and that would break l2geth.

**Tests**
Going to do canary release and test this.


tynes (approved):

smartcontracts:
Successfully dumping state with this change

trianglesphere (approved):

PR #3728, opened 2022-10-18
op-e2e: test orphan L1 block with batch and replay batch + keep engine fc state in sync [bedrock]

protolambda:
**Description**

This PR tests that a verifier can handle a L1 reorg of 1-block (a.k.a. orphan block) that includes a batch, and then tests what happens if it is then replayed onto a new L1 chain, and if it stays in sync with the sequencer.

**Tests**

See above.

**Metadata**

Fix ENG-2770

PR #3765, opened 2022-10-21
op-e2e: flip-flop test

protolambda:
**Description**

Depends on #3728 

**Tests**

This tests we can flip-flop as verifier between two L1 chains, with replayed (and temporarily invalid during flip-flop) batches.

TODO: the engine adds one valid previously seen block on top of the unsafe chain, even though the rollup node does not ask it to. It's consolidated fine later on, but this forkchoice state should be synced (or it's a minor geth bug where it's to eager to preserve blocks, which can't happen in PoS where this chain extension is always the head).

**Metadata**

Fix ENG-2766


trianglesphere (commented):
test itself lgtm

trianglesphere (approved):

mslipper (approved):

PR #3767, opened 2022-10-24
op-chain-ops: implement migration script

tynes:
**Description**

Implement the cli command for doing a database migration. This script has a lot of inputs so isn't straight forward to test, need to create some canonical data that we can test this against.

This script parses the inputs to the migration but does not yet use them fully for validation. It performs the migration but still requires the input to be present.

It also adds a wrapper for the legacy l2geth statedb so that it can be passed into functions that expect a geth statedb, making the types match up correctly.

### DIFF
--- a/op-chain-ops/genesis/migration.go
+++ b/op-chain-ops/genesis/migration.go
@@ -2,6 +2,7 @@ package genesis
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 
 	"github.com/ethereum-optimism/optimism/op-chain-ops/crossdomain"
@@ -14,7 +15,7 @@ import (
 // `LegacyMessagePasser`. The `who` should always be the
 // `L2CrossDomainMessenger` and the `msg` should be an abi encoded
 // `relayMessage(address,address,bytes,uint256)`
-type SentMessageJSON struct {
+type SentMessage struct {
 	Who common.Address `json:"who"`
 	Msg hexutil.Bytes  `json:"msg"`
 }
@@ -22,13 +23,13 @@ type SentMessageJSON struct {
 // NewSentMessageJSON will read a JSON file from disk given a path to the JSON
 // file. The JSON file this function reads from disk is an output from the
 // `migration-data` package.
-func NewSentMessageJSON(path string) ([]*SentMessageJSON, error) {
+func NewSentMessage(path string) ([]*SentMessage, error) {
 	file, err := os.ReadFile(path)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot find sent message json at %s: %w", path, err)
 	}
 
-	var j []*SentMessageJSON
+	var j []*SentMessage
 	if err := json.Unmarshal(file, &j); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


op-chain-ops/genesis/migration.go
```diff
@@ -0,0 +1,52 @@
...
+// ToLegacyWithdrawal will convert a SentMessageJSON to a LegacyWithdrawal
+// struct. This is useful because the LegacyWithdrawal struct has helper
+// functions on it that can compute the withdrawal hash and the storage slot.
+func (s *SentMessageJSON) ToLegacyWithdrawal() (*crossdomain.LegacyWithdrawal, error) {
+	data := make([]byte, 0, len(s.Who)+len(s.Msg))
+	copy(data, s.Msg)
+	copy(data[len(s.Msg):], s.Who[:])
+
+	var w crossdomain.LegacyWithdrawal
+	if err := w.Decode(data); err != nil {
```
smartcontracts, line 48: This will throw if `who` is not the L2CrossDomainMessenger: https://github.com/ethereum-optimism/optimism/blob/bba641d00366cd14616bee185477ec260bfa1038/op-chain-ops/crossdomain/legacy_withdrawal.go#L62

smartcontracts: I think this is fine for now, we'll have to address this later. I just wanted to make you aware of it if not already.

smartcontracts (approved):

mslipper:
@tynes is this ready for review? Saw a `wip` commit. Happy to review if it is.

tynes:
> @tynes is this ready for review? Saw a `wip` commit. Happy to review if it is.

Just converted it to a draft, sorry about that - I pushed some additional code to it and need to double check on the validation logic before its really ready for review.

Planning on a different PR that modularizes the balance migration stuff so that it could be used as part of this script, or we could just leave the balance migration stuff as is and run 2 scripts to do the full migration

op-chain-ops/genesis/migration.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, err
+	}
+	return &w, nil
+}
+
+// Addresses represents a list of addresses that interacted with
+// the ERC20 representation of ether in the pre-bedrock system.
+type Addresses []common.Address
```
tynes, line 57: Turn this into a mapping instead of a slice

op-chain-ops/genesis/migration.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Allowance represents the allowances that were set in the
+// legacy ERC20 representation of ether
+type Allowance struct {
+	From common.Address `json:"fr"`
+	To   common.Address `json:"to"`
+}
+
+// NewAllowances will read the ovm-allowances.json from the file system.
+func NewAllowances(path string) ([]*Allowance, error) {
```
tynes, line 82: Turn the return of this into a mapping instead of a slice

mslipper (approved):

op-chain-ops/genesis/migration.go
tynes: Thanks for the heads up here

PR #3769, opened 2022-10-24
specs,op-node: Clarify max bytes per channel comparison

trianglesphere:
**Description**

The exact requirement is that the input length is less than or equal to the max bytes per channel. This comparison is done with inputLimit in the RLP reader.

**Metadata**
- Fixes ENG-2926

### DIFF
--- a/op-node/rollup/derive/channel_out_test.go
+++ b/op-node/rollup/derive/channel_out_test.go
@@ -1,10 +1,12 @@
 package derive
 
 import (
+	"bytes"
 	"math/big"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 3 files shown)


op-node/rollup/derive/channel_out_test.go
```diff
@@ -1,9 +1,11 @@
 package derive
 
 import (
+	"bytes"
 	"math/big"
 	"testing"
 
+	"github.com/ethereum-optimism/optimism/l2geth/rlp"
```
tynes, line 8: This should use the upstream geth instead of l2geth

tynes (approved):

protolambda (approved):

PR #3771, opened 2022-10-25
feat(ctb): explicit throw in trie get

smartcontracts:
**Description**
Small refactor that has MerkleTrie.get throw explicitly when an element does not exist, rather than returning a non-existence boolean. Makes the code much cleaner. We need to review this change very carefully.

**Tests**
Needs tests.


tynes:
How required are these changes?

packages/contracts-bedrock/contracts/test/MerkleTrie.t.sol
```diff
@@ -0,0 +1,367 @@
+// SPDX-License-Identifier: MIT
```
tynes, line 1: Where did you get these test vectors?

PR #3773, opened 2022-10-25
feat(docs): update readme

smartcontracts:
**Description**
I made the monorepo prettier and added some more detail. A good readme makes the monorepo gods happy.


smartcontracts:
look at that nice big logo that actually uses the correct wordmark

tynes (approved):

mslipper (approved):

PR #3781, opened 2022-10-25
bedrock: make fee recipient be SequencerFeeVault

tynes:
**Description**

The fee recipient or `block.coinbase` is now set to a hardcoded value of the predeploy `SequencerFeeVault`. This reduces configuration that is consensus criticial.

The config value `OptimismL2FeeRecipient` is removed in favor of a hardcoded predeploy. The fees already accumulate to this address, so it is in line with the way that the current system works.

A bug is also fixed for the tests of the L1 fee recipient, the value was never being set correctly in the L2 chain config, the value is now set as expected and the correct address is asserted against in the tests

### DIFF
--- a/op-chain-ops/genesis/genesis.go
+++ b/op-chain-ops/genesis/genesis.go
@@ -52,6 +52,7 @@ func NewL2Genesis(config *DeployConfig, block *types.Block) (*core.Genesis, erro
 		TerminalTotalDifficulty:       big.NewInt(0),
 		TerminalTotalDifficultyPassed: true,
 		Optimism: &params.OptimismConfig{
+			L1FeeRecipient:     config.OptimismL1FeeRecipient,
 			BaseFeeRecipient:   config.OptimismBaseFeeRecipient,
 			EIP1559Denominator: eip1559Denom,
 			EIP1559Elasticity:  eip1559Elasticity,
--- a/op-chain-ops/genesis/genesis.go
+++ b/op-chain-ops/genesis/genesis.go
@@ -53,7 +53,6 @@ func NewL2Genesis(config *DeployConfig, block *types.Block) (*core.Genesis, erro
 		TerminalTotalDifficultyPassed: true,
 		Optimism: &params.OptimismConfig{
 			BaseFeeRecipient:   config.OptimismBaseFeeRecipient,
-			L1FeeRecipient:     config.OptimismL2FeeRecipient,
 			EIP1559Denominator: eip1559Denom,
 			EIP1559Elasticity:  eip1559Elasticity,
 		},
--- a/op-chain-ops/genesis/config.go
+++ b/op-chain-ops/genesis/config.go
@@ -30,7 +30,6 @@ type DeployConfig struct {
 	SequencerWindowSize       uint64         `json:"sequencerWindowSize"`
 	ChannelTimeout            uint64         `json:"channelTimeout"`
 	P2PSequencerAddress       common.Address `json:"p2pSequencerAddress"`
-	OptimismL2FeeRecipient    common.Address `json:"optimismL2FeeRecipient"`
 	BatchInboxAddress         common.Address `json:"batchInboxAddress"`
 	BatchSenderAddress        common.Address `json:"batchSenderAddress"`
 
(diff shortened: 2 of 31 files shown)


mslipper:
Seeing this in the devnet + yarn monorepo tests:

```
tasks/rollup-config.ts(47,11): error TS2741: Property 'fee_recipient_address' is missing in type '{ genesis: { l1: { hash: string; number: number; }; l2: { hash: string; number: number; }; l2_time: number; }; block_time: any; max_sequencer_drift: any; seq_window_size: any; channel_timeout: any; l1_chain_id: any; ... 4 more ...; deposit_contract_address: string; }' but required in type 'OpNodeConfig'.
```

The fees test in op-e2e is failing as a result of this as well. LGTM once these are fixed.

tynes:
```
=== CONT  TestFees
    system_test.go:1003: 
        	Error Trace:	/root/project/op-e2e/system_test.go:1003
        	Error:      	Not equal: 
        	            	expected: 32816
        	            	actual  : 210000
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -3,3 +3,3 @@
        	            	  abs: (big.nat) (len=1) {
        	            	-  (big.Word) 32816
        	            	+  (big.Word) 210000
        	            	  }
        	Test:       	TestFees
        	Messages:   	l1 fee mismatch
```

protolambda (commented):
LGTM, but these are breaking changes w.r.t. current testnets (alpha 1, beta 1, opcraft). With this + derivation fixes/tweaks from Josh + GPO/batcher key + dynamic gas limit I think it may make sense to see if we can merge them earlier than end of next week?

mslipper (approved):

tynes:
Not sure why the devnet with deployed contracts is breaking here, will need to run locally to debug

protolambda (approved):
LGTM, but in a follow-up PR we should instantiate different fee vaults for each type of L2 fee. I think we are sending it to EOAs in bedrock now? So instead of `predeploys.SequencerFeeVault`, we can have `L2FeeRecipient`, `L1FeeRecipient`, `BaseFeeRecipient` or something like that.

op-chain-ops/genesis/genesis.go
```diff
@@ -52,8 +52,8 @@ func NewL2Genesis(config *DeployConfig, block *types.Block) (*core.Genesis, erro
 		TerminalTotalDifficulty:       big.NewInt(0),
 		TerminalTotalDifficultyPassed: true,
 		Optimism: &params.OptimismConfig{
+			L1FeeRecipient:     config.OptimismL1FeeRecipient,
```
trianglesphere, line 55: Maybe we should also be updating the config name here - the `L1FeeRecipient` should receive all fees collected on L2 except for the basefee (which goes to the `BaseFeeRecipient`

trianglesphere (approved):
this is good to go once the devnet test is fixed. Weird that it's failing.

trianglesphere:
> we should instantiate different fee vaults for each type of L2 fee. I think we are sending it to EOAs in bedrock now? So instead of `predeploys.SequencerFeeVault`, we can have `L2FeeRecipient`, `L1FeeRecipient`, `BaseFeeRecipient` or something like that.

We decide a while ago that we shouldn't split it into three streams, but two (basefee + L1 fee) & (the miner tip). https://www.notion.so/oplabs/L2-Fee-Payout-8b3485ce8e61475ca31bfc38994a5b06. The addresses are supposed to be contracts rather than EOAs, but they were normal contracts rather than pre-deploys.

tynes:
> LGTM, but in a follow-up PR we should instantiate different fee vaults for each type of L2 fee. I think we are sending it to EOAs in bedrock now? So instead of `predeploys.SequencerFeeVault`, we can have `L2FeeRecipient`, `L1FeeRecipient`, `BaseFeeRecipient` or something like that.

I have a PR that i'm almost ready to push that does this, will have it up soon

tynes:
Unclear to me why this is failing, need to investigate more:

```
timeout 5m npx hardhat deposit-erc20 --network devnetL1 --l1-contracts-json-path ../../.devnet/sdk-addresses.json
```

tynes:
https://github.com/ethereum-optimism/optimism/pull/3827 fixes the linting error in this PR

tynes:
> Unclear to me why this is failing, need to investigate more:
> 
> ```
> timeout 5m npx hardhat deposit-erc20 --network devnetL1 --l1-contracts-json-path ../../.devnet/sdk-addresses.json
> ```

This does not fail for me locally, will try again with a freshly cloned repo such that its closer to what runs in CI

mslipper:
The devnet issues are unrelated - they work in CI when I activated SSH. I'm going to merge this in so that this doesn't cause conflicts. I'll look into the devnet issues.

PR #3782, opened 2022-10-25
op-e2e: Add invalid deposit test

trianglesphere:
**Description**

This test ensure that invalid deposits are stil included in the block. If these are somehow dropped, it would cause serious problems for consistency and potentially a loss of funds.

**Metadata**

Fixes ENG-2921


op-e2e/op_geth_test.go
```diff
@@ -0,0 +1,133 @@
...
+	l1Genesis, err := genesis.BuildL1DeveloperGenesis(cfg.DeployConfig)
+	require.Nil(t, err)
+	l1Block := l1Genesis.ToBlock()
+
+	l2Addrs := &genesis.L2Addresses{
+		ProxyAdmin:                  predeploys.DevProxyAdminAddr,
+		L1StandardBridgeProxy:       predeploys.DevL1StandardBridgeAddr,
+		L1CrossDomainMessengerProxy: predeploys.DevL1CrossDomainMessengerAddr,
+	}
+	l2Genesis, err := genesis.BuildL2DeveloperGenesis(cfg.DeployConfig, l1Block, l2Addrs)
```
tynes, line 46: nit: pass `nil` as the `l2Addrs` arg for the dev addrs to be used

op-e2e/op_geth_test.go
```diff
@@ -0,0 +1,133 @@
...
+		sources.EngineClientDefaultConfig(&rollup.Config{Genesis: rollupGenesis}),
+	)
+	require.Nil(t, err)
+
+	// Create the test data (L1 Info Tx and then always failing deposit)
+	l1Info, err := derive.L1InfoDepositBytes(1, l1Block)
+	require.Nil(t, err)
+
+	fromAddr := cfg.Secrets.Addresses().Alice
+	badDepositTx := types.NewTx(&types.DepositTx{
```
tynes, line 87: What makes this invalid?

trianglesphere: Alice has no funds on L2 (`cfg.DeployConfig.FundDevAccounts = false`) way above.
I manually checked it looking at op-geth logging, but can add an explicit check that the balance of alice is 0 on L2.

protolambda (approved):
LGTM. Would be nice to add that 0-balance check however.

trianglesphere:
> LGTM. Would be nice to add that 0-balance check however.

Added the 0-balance check.

mslipper (approved):

op-e2e/op_geth_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	status, err := client.NewPayload(ctx, payload)
+	require.Nil(t, err)
+	require.Equal(t, eth.ExecutionValid, status.Status)
+
+	fc.HeadBlockHash = payload.BlockHash
+	res, err = client.ForkchoiceUpdate(ctx, &fc, nil)
+	require.Nil(t, err)
+	require.Equal(t, eth.ExecutionValid, res.PayloadStatus.Status)
+
```
tynes, line 135: nit: extra newline here

op-e2e/op_geth_test.go
```diff
@@ -0,0 +1,136 @@
...
+	badDepositTx := types.NewTx(&types.DepositTx{
+		// TODO: Source Hash
+		From:                fromAddr,
+		To:                  &fromAddr, // send it to ourselves
+		Value:               big.NewInt(params.Ether),
+		Gas:                 25000,
+		IsSystemTransaction: false,
+	})
+	badDeposit, err := badDepositTx.MarshalBinary()
```
tynes, line 98: Is this deposit included in L2 but reverts?

trianglesphere: It's included, but not technically a revert. If you try to send more ETH than you have, you get `ErrInsufficientFundsForTransfer`. This normally prevents your tx from ever getting included in the first place (it's not a revert error, but a an intrinsic validity check). We require that it is still included & that the mint succeeds (we test that + something similar to this in `TestMintOnRevertedDeposit`).

PR #3787, opened 2022-10-26
op-node: SystemConfig in derivation

protolambda:
**Description**

Introduce `SystemConfig`: a small struct of configurable data that carries over in each L2 block, and can be modified based on events from L1.

When derivation starts (after a reset), the stages are reset to use the system config of the L1 origin they start with, which is read from the last block in the L2 chain with that origin.

The stages then update their system config with each L1 block that passes through the stage: this is isolated, and even though it may look like more work, it ensures we don't have to share data between stages, and the receipts cache avoids any unnecessary repeated data retrieval from L1.

The L1 traversal stage can thus update the system config as soon as it moves to the next L1 block, and this config can then be used to filter the data transactions in the retrieval stage.

And the attributes stage can update the system config to include new GPO param updates in the first L2 block attributes that introduce the L1 origin. Similarly to GPO param updates, this could also include other L2 parameter changes in the future easily, like e.g. gas limit updates.

This is part 1/5 to introduce batcher-key and GPO param configuration:
1. Derivation aware of system config (with placeholders for later actual contract/event changes) - #3787
2. Contract changes: system config on L1, and updated L2 contracts - #3788
3. Update bindings, update chainops to use updated bindings and dev-deploy system config, and misc integration like log parsing - #3789
4. Action-test the batcher key change and GPO params change. - #3790
5. Spec updates - #3831

**Metadata**

Fix ENG-2940


op-e2e/system_test.go
```diff
@@ -864,41 +874,12 @@ func TestFees(t *testing.T) {
...
-
-	receipt, err = waitForTransaction(tx.Hash(), l2Verif, 10*time.Duration(cfg.DeployConfig.L1BlockTime)*time.Second)
-	require.Nil(t, err, "waiting for gpo scalar update tx")
-	require.Equal(t, receipt.Status, types.ReceiptStatusSuccessful, "transaction failed")
-
 	overhead, err := gpoContract.Overhead(&bind.CallOpts{})
 	require.Nil(t, err, "reading gpo overhead")
-	decimals, err := gpoContract.Decimals(&bind.CallOpts{})
-	require.Nil(t, err, "reading gpo decimals")
+	decimals := big.NewInt(6) // protocol constant
```
tynes, line 882: I've realized that this is still available via a public getter on the contract, altho it is now inlined in the bytecode instead of in state. On upgraded systems, the value will still be in state. We will need to add to the specs that this is a protocol constant

mslipper (approved):

protolambda:
rebased again, fixed merge conflict + missing sys config argument in new go test

tynes:
Looks like this now needs a rebase

op-e2e/system_test.go
```diff
@@ -851,6 +856,11 @@ func TestFees(t *testing.T) {
 	}
 
 	cfg := DefaultSystemConfig(t)
+	// TODO: after we have the system config contract and new op-geth L1 cost utils,
+	// we can pull in l1 costs into every e2e test and account for it in assertions easily etc.
+	cfg.DeployConfig.GasPriceOracleOverhead = 2100
+	cfg.DeployConfig.GasPriceOracleScalar = 1000_000
+	cfg.DeployConfig.GasPriceOracleDecimals = 6
```
tynes, line 863: I feel like the decimals should no longer be a config param and we should make it into a protocol constant, it could go in `protocol_params.go` in `op-geth` or live in a constants in `op-node`. I believe this config previously would set the storage in the gas oracle contract but now the decimals value is a `constant` so its inlined in the bytecode and not in storage

op-e2e/system_test.go
tynes, line 882: Perhaps this test would be better to call the `decimals()(uint256)` getter and assert the result against a constant value that we can use across the codebase

op-node/eth/types.go
```diff
@@ -314,3 +314,12 @@ func (f FetchedReceipts) Result() (types.Receipts, error) {
 }
 
 var _ ReceiptsFetcher = (FetchedReceipts)(nil)
+
+// SystemConfig represents the rollup system configuration that carries over in every L2 block,
+// and may be changed through L1 system config events.
```
tynes, line 319: Its a bit error prone to have the initial config in the rollup config and passed to the contract constructor, i don't think there is anything preventing them from being different

tynes: It seems complex to have to read the initial config from the contract

op-node/eth/types.go
```diff
@@ -314,3 +314,12 @@ func (f FetchedReceipts) Result() (types.Receipts, error) {
...
+// The initial SystemConfig at rollup genesis is embedded in the rollup configuration.
+type SystemConfig struct {
```
tynes, line 321: Should be versioned?

op-node/rollup/derive/attributes_queue_test.go
```diff
@@ -47,7 +49,21 @@ func TestAttributesQueue(t *testing.T) {
 		Transactions: []eth.Data{eth.Data("foobar"), eth.Data("example")},
 	}}
 
-	l1InfoTx, err := L1InfoDepositBytes(safeHead.SequenceNumber+1, l1Info)
+	parentL1Cfg := eth.SystemConfig{
+		BatcherAddr: common.Address{42},
+		Overhead:    [32]byte{},
```
tynes, line 54: Is the main benefit here of keeping the overhead and scalar as `[32]byte` not needing to deserialize it? I suppose its never used in any arithmetic here so no need to change into a `big.Int`

op-node/rollup/derive/attributes_test.go
```diff
@@ -46,11 +56,19 @@ func TestPreparePayloadAttributes(t *testing.T) {
 		l1Fetcher := &testutils.MockL1Source{}
 		defer l1Fetcher.AssertExpectations(t)
 		l2Parent := testutils.RandomL2BlockRef(rng)
+		l1CfgFetcher := &testutils.MockL2Client{}
+		l1Cfg := eth.SystemConfig{
```
tynes, line 60: Perhaps a function `mockSystemConfig` to reduce the boilerplate

op-node/rollup/derive/calldata_source.go
```diff
@@ -100,7 +104,7 @@ func (ds *DataSource) Next(ctx context.Context) (eth.Data, error) {
 // DataFromEVMTransactions filters all of the transactions and returns the calldata from transactions
 // that are sent to the batch inbox address from the batch sender address.
 // This will return an empty array if no valid transactions are found.
-func DataFromEVMTransactions(config *rollup.Config, txs types.Transactions, log log.Logger) []eth.Data {
```
tynes, line 103: I don't think this function needs to take the entire `rollup.Config` anymore since it just uses `config.BatchInboxAddress`, could just pass in that instead of the entire config

op-node/rollup/derive/engine_queue.go
```diff
@@ -483,22 +489,32 @@ func (eq *EngineQueue) Reset(ctx context.Context, _ eth.L1BlockRef) error {
 			safe, safe.Time, l1Origin, l1Origin.Time))
 	}
 
-	pipelineNumber := l1Origin.Number - eq.cfg.ChannelTimeout
-	if l1Origin.Number < eq.cfg.ChannelTimeout {
-		pipelineNumber = 0
+	// Walk back L2 chain to find the L1 origin that is old enough to start buffering channel data from.
+	pipelineL2 := safe
+	for pipelineL2.Number > eq.cfg.Genesis.L2.Number && pipelineL2.L1Origin.Number > eq.cfg.Genesis.L1.Number && pipelineL2.L1Origin.Number+eq.cfg.ChannelTimeout > l1Origin.Number {
```
tynes, line 494: I don't love how long this condition is, perhaps could use a for loop with a break and assign each part of the condition to a variable like this:

```go
for {
    isPipelineNumberGreater := pipelineL2.Number > eq.cfg.Genesis.L2.Number
    isPipelineOriginGreater := pipelineL2.L1Origin.Number > eq.cfg.Genesis.L1.Number
   // other condition
   if isPipelineNumberGreater && isPipelineOriginGreater {
      // do logic
    }
}
```

op-node/rollup/derive/l1_retrieval_test.go
```diff
@@ -117,6 +133,7 @@ func TestL1RetrievalNextData(t *testing.T) {
 		{
 			name:         "fail to open data",
 			prevBlock:    a,
+			sysCfg:       eth.SystemConfig{BatcherAddr: common.Address{42}},
```
tynes, line 136: Not sure if `Address{42}` is the best test account bc it is a predeploy, the `LegacyMessagePasser` contract

op-node/rollup/derive/l1_traversal_test.go
```diff
@@ -87,16 +110,31 @@ func TestL1TraversalAdvance(t *testing.T) {
...
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			src := &testutils.MockL1Source{}
 			src.ExpectL1BlockRefByNumber(test.startBlock.Number+1, test.nextBlock, test.fetcherErr)
+			info := &testutils.MockBlockInfo{
+				InfoHash:       test.nextBlock.Hash,
+				InfoParentHash: test.nextBlock.ParentHash,
+				InfoNum:        test.nextBlock.Number,
+				InfoTime:       test.nextBlock.Time,
+				// TODO: don't need full L1 info in receipts fetching API maybe?
```
tynes, line 125: Is this todo something you would like to address?

op-node/rollup/derive/engine_queue.go
trianglesphere, line 494: Ideally this change would not be part of this PR.

op-node/rollup/types_test.go
```diff
@@ -27,14 +28,23 @@ func randConfig() *Config {
 			L1:     eth.BlockID{Hash: randHash(), Number: 424242},
 			L2:     eth.BlockID{Hash: randHash(), Number: 1337},
 			L2Time: uint64(time.Now().Unix()),
+			SystemConfig: eth.SystemConfig{
+				BatcherAddr: randAddr(),
+				Overhead:    randHash(),
+				Scalar:      randHash(),
```
tynes, line 34: If these are used in tests that have fees, could see some weird results if random hashes are used as the values

op-node/rollup/derive/l1_block_info.go
```diff
@@ -49,10 +54,17 @@ func (info *L1BlockInfo) MarshalBinary() ([]byte, error) {
 	copy(data[offset:offset+32], info.BlockHash.Bytes())
 	offset += 32
 	binary.BigEndian.PutUint64(data[offset+24:offset+32], info.SequenceNumber)
+	//offset += 32
+	//copy(data[offset+12:offset+32], info.BatcherAddr[:])
```
trianglesphere, line 58: In general I think we should be moving to a reader based API for marshalling/unmarshalling because these offset calculations are usually pretty tricky.

trianglesphere (commented):
Again this is doing a lot. I'd like to see merge conflicts + tests for this fixed. Maybe hive is being flakier than I've expected, but we need to either treat these tests seriously or get rid of them.

op-node/eth/types.go
protolambda: To load an initial configuration from L1 we need to add more code, and discuss how we can read it securely, even when the systemconfig has already changed after genesis. Let's do that in a separate PR, and maybe write a short design doc together how we load the initial config? Opened ENG-2963 to track this.

op-node/rollup/derive/attributes_queue_test.go
protolambda: It's never used for arithmetic indeed, so it avoids redundant error edge cases by just not deserializing it.

op-node/eth/types.go
protolambda: We don't need to version the struct as a whole, the individual config fields can have versions, and we can extend it with more fields later quite easily.

op-node/rollup/derive/calldata_source.go
protolambda: It also needs the chain ID to construct a signer with to get transaction sender information.

op-node/rollup/derive/l1_traversal_test.go
protolambda: Opened ENG-2964, it's low priority.

op-node/rollup/types_test.go
protolambda: It's only a basic json encoding test, the values don't matter that much in this case. We can fuzz the rollup.json encoding later, but it seems low risk since we only deal with one such config in a network.

op-node/rollup/derive/engine_queue.go
protolambda: We need the L2 block that references the L1 origin we reset to, not just the L1 origin, to be able to recover the initial system config. I'm updating it to be more readable based on suggestions from Mark.

op-node/rollup/derive/l1_block_info.go
protolambda: Opened ENG-2965 to track this

op-e2e/system_test.go
protolambda: Later PRs in the stack get rid of it, here I just try to minimize config changes.

op-e2e/system_test.go
protolambda: See other comment, later PRs in the stack deal with remaining integration/simplification of the GPO params.

protolambda: The decimals thing is removed from deploy config in later PR in the stack, and I think we can tweak the usage of the decimals GPO variable in a separate PR, this one is getting unwieldy.

tynes (approved):

tynes:
```
CRIT [11-03|20:08:37.989] Application failed                       message="SystemConfig: no contract code after deployment"
```

Looks like the systemconfig is reverting during deployment?

protolambda:
@tynes I've seen that error before if the constructor fails due to the `initialize` call, but thought I fixed it. I'll check the `initialize`

protolambda:
@tynes I think it only fails if the owner is set to the zeroed address. Some configuration is not set somewhere

op-e2e/system_test.go
```diff
@@ -670,6 +691,9 @@ func calcL1GasUsed(data []byte, overhead *big.Int) *big.Int {
 	return new(big.Int).Add(l1Gas, overhead)
 }
 
+// TODO: temp until TestFees is unskipped
```
tynes, line 694: We can prob undo this now

op-e2e/system_test.go
```diff
@@ -1018,3 +1022,6 @@ func TestFees(t *testing.T) {
 func safeAddBig(a *big.Int, b *big.Int) *big.Int {
 	return new(big.Int).Add(a, b)
 }
+
+// TODO: temp until TestFees is unskipped
```
tynes, line 1026: We can prob undo this now

tynes:
> @tynes I think it only fails if the owner is set to the zeroed address. Some configuration is not set somewhere

I'll take a look

op-chain-ops/genesis/layer_one.go
```diff
@@ -65,12 +67,35 @@ func BuildL1DeveloperGenesis(config *DeployConfig) (*core.Genesis, error) {
 	if err != nil {
 		return nil, err
 	}
+	sysCfgABI, err := bindings.SystemConfigMetaData.GetAbi()
+	if err != nil {
+		return nil, err
+	}
+	data, err := sysCfgABI.Pack(
+		"initialize",
+		config.SystemConfigOwner,
```
tynes, line 76: It looks like this value could be `address(0)`

protolambda: yep, updated the deploy configs in hardhat, that should fix the devnet. We'll also have to add it to hive setup.

tynes:
Dockerhub seems to be down

```
> [ethereumoptimism/batch-submitter-service:latest internal] load metadata for docker.io/library/golang:1.18.0-alpine3.15:
------
failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/library/golang/blobs/sha256:c04fbda3c7471a7bc6f9057295058a5c55c5c21e51680007085bdb4ff2dfb2dd: 400 Bad Request
```

op-e2e/actions/system_config_test.go
```diff
@@ -0,0 +1,290 @@
...
+	miner.ActL1IncludeTx(dp.Addresses.Batcher)(t)
+	miner.ActL1EndBlock(t)
+
+	sequencer.ActL2PipelineFull(t)
+	verifier.ActL2PipelineFull(t)
+	require.Equal(t, uint64(3+6+1), verifier.L2Safe().L1Origin.Number, "sync new L1 chain, while key change is reorged out")
+	require.Equal(t, sequencer.L2Unsafe(), verifier.L2Unsafe(), "verifier synced")
+}
+
+func TestGPOParamsChange(gt *testing.T) {
```
trianglesphere, line 175: ditto test desc.

op-node/rollup/derive/system_config.go
```diff
@@ -0,0 +1,102 @@
...
+	version := ev.Topics[1]
+	if version != ConfigUpdateEventVersion0 {
+		return fmt.Errorf("unrecognized L1 sysCfg update event version: %s", version)
+	}
+	// indexed 1
+	updateType := ev.Topics[2]
+	// unindexed data
+	switch updateType {
+	case SystemConfigUpdateBatcher:
+		if len(ev.Data) != 32*3 {
```
trianglesphere, line 72: Ditto comments in a previous PR. I would be good split this into two dedicated parsing functions, one for each type & then have unit + fuzz tests for them.

trianglesphere (commented):
some comments where pulled from previous PRs. This has gotten pretty large & is pretty hard to review.

op-e2e/actions/l2_batcher.go
```diff
@@ -105,16 +108,24 @@ func (s *L2Batcher) ActL2BatchBuffer(t Testing) {
 		s.l2BufferedBlock = syncStatus.SafeL2.ID()
 		s.l2ChannelOut = nil
 	}
+	// Add the next unsafe block to the channel
+	if s.l2BufferedBlock.Number >= syncStatus.UnsafeL2.Number {
+		if s.l2BufferedBlock.Number > syncStatus.UnsafeL2.Number || s.l2BufferedBlock.Hash != syncStatus.UnsafeL2.Hash {
+			s.log.Error("detected a reorg in L2 chain vs previous buffered information, resetting to safe head now", "safe_head", syncStatus.SafeL2)
```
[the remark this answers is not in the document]
protolambda: iirc the action-test that covers batcher key changes hits this

op-e2e/actions/l2_sequencer.go
```diff
@@ -102,3 +102,17 @@ func (s *L2Sequencer) ActBuildToL1Head(t Testing) {
 		s.ActL2EndBlock(t)
 	}
 }
+
+// ActBuildToL1HeadExcl builds empty blocks until (excl.) the L1 head becomes the L2 origin
```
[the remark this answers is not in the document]
protolambda: ActBuildToL1Head = build L2 blocks until there is a L2 block with a L1 origin that matches the L1 head.
ActBuildToL1HeadExcl = same as above, except that last L2 block with the L1 origin that matches the L1 head.

op-node/rollup/derive/system_config.go
protolambda: Sounds good, but those previous PRs have been reviewed sufficiently, and we should do further refactors in separate PRs. This PR got so large because the whole stack was reviewed enough that we could merge it into one. If we are going to make code changes then let's isolate those into new reviewable PRs.

PR #3788, opened 2022-10-26
contracts-bedrock: Sys config contract and L2 contract updates

protolambda:
**Description**

This introduces the `SystemConfig` contract and L2 contract changes written by @tynes, with some minor modifications to use `initialize` for integration with proxy etc. things in the chain-ops PR.

This is part 2/5 to introduce batcher-key and GPO param configuration:
1. Derivation aware of system config (with placeholders for later actual contract/event changes) - #3787
2. Contract changes: system config on L1, and updated L2 contracts - #3788
3. Update bindings, update chainops to use updated bindings and dev-deploy system config, and misc integration like log parsing - #3789
4. Action-test the batcher key change and GPO params change. - #3790
5. Spec updates - #3831

Do not merge: this is a breaking change, and the other parts of the PR stack should be reviewed first.

Note: CI will fail on op-bindings, these bindings get their own PR.

Changes:
- L1 Block info contract now holds on to GPO params and batcher. We should maybe just rename it to L1 info contract, the scope is a bit expanded.
- GPO contract just reads the block info contract - this is for backwards compatibility. The gas cost is lower to keep storage in one contract
- SystemConfig on L1, owner by some account/contract, can change the batcher key and GPO params. On a change an event is emitted that can then be picked up by L2 derivation.


packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -0,0 +1,59 @@
...
+import { Semver } from "../universal/Semver.sol";
+
+/**
+ * @title SystemConfig
+ * @notice This contract is used to update L2 configuration via L1
+ */
+contract SystemConfig is OwnableUpgradeable, Semver {
+    uint256 public constant VERSION = 0;
+
+    uint256 public overhead;
```
tynes, line 16: I think its simple to have these values be `uint256`s but I don't think that we really need them to be. The `GasPriceOracle` on L2 has them as `uint256` but making them smaller wouldn't break anything since its restricting the possible values, plus we haven't set them that large

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {
+    OwnableUpgradeable
+} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import { Semver } from "../universal/Semver.sol";
+
+/**
+ * @title SystemConfig
```
tynes, line 10: I'm ok with a follow up PR that adds devdocs and unit tests for this contract, its relatively straight forward and the e2e tests technically do cover the functionality here, but that isn't an excuse for not having unit tests

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -0,0 +1,59 @@
...
+        GAS_CONFIG
+    }
+
+    constructor(uint256 _overhead, uint256 _scalar, address _batcher) Semver(0, 0, 1) {
+        overhead = _overhead;
+        scalar = _scalar;
+        batcher = _batcher;
+    }
+
+    /**
```
tynes, line 37: Would you consider this to be the canonical way we do upgradable ownable contracts?
cc @maurelian @smartcontracts

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -0,0 +1,59 @@
...
+
+    enum UpdateType {
+        BATCHER,
+        GAS_CONFIG
+    }
+
+    constructor(uint256 _overhead, uint256 _scalar, address _batcher) Semver(0, 0, 1) {
+        overhead = _overhead;
+        scalar = _scalar;
+        batcher = _batcher;
```
tynes, line 34: This should also take the owner and call `initialize` so that the impl isn't uninitialized

tynes:
Do we want to emit the same events that are on the GPO to make it easy for data analytics? It will now require some abi decoding to get that data

- in the system config contract?
- in the L1Block contract but only if there is a diff?


```
    /**
     * @notice Emitted when the overhead value is updated.
     */
    event OverheadUpdated(uint256 overhead);

    /**
     * @notice Emitted when the scalar value is updated.
     */
    event ScalarUpdated(uint256 scalar);
```

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
protolambda: Changing them may affect contract storage layout - not a problem for this new contract, but there may be some other cases to consider, like updating op-geth to read the packed storage.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
protolambda: Yep, need some help with contract testing, although not urgent (it'll will be a while until we can land this breaking change). Maybe you or @maurelian have time end this cycle, or start next cycle?

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
protolambda: isn't `initialize` called by the proxy / deployment setup? If the constructor calls it, what's the point of having it in a separate function?

protolambda:
@tynes 

> in the system config contract?

There's no current dependency on such L1 events afaik, so I wouldn't add it there. And we do have the generic events, which analytics would want to support anyway, since it captures all kinds of updates.

> in the L1Block contract but only if there is a diff?

Since the tx receipts now get the exact fee data already I don't think analytics solutions actually have to rely on the L2 events?
Do we ever need to log-filter for L1 fee changes on L2, when the data is in transaction receipts already?

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -0,0 +1,60 @@
...
+
+    enum UpdateType {
+        BATCHER,
+        GAS_CONFIG
+    }
+
+    constructor(address _owner, uint256 _overhead, uint256 _scalar, bytes32 _batcherHash) Semver(0, 0, 1) {
+        overhead = _overhead;
+        scalar = _scalar;
+        initialize(_owner, _batcherHash);
```
tynes, line 34: We should be setting `overhead` and `scalar` in the `initialize` function. Perhaps via the functions so that events are emitted instead of setting them directly. If we don't go through `initialize`, then the values will not be set in the proxy

protolambda: I'm following the pattern of l2 ouput oracle here where protocol params are set in the constructor, and addresses/permissions are set in initialize. Is the output oracle set up wrong?

tynes: Exactly, immutables must be set in the constructor and not the `initialize` function

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
protolambda: ENG-2949 is tracking this now

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
protolambda: In slack there was some soft agreement that following the l2 output oracle pattern works for now. We can revisit it in a separate PR.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
protolambda: Updated

protolambda:
@tynes can you help update the gas-snapshot? (Locally it fails when running diff fuzz tests with some typescript module errors, not sure what's wrong). PR is ready for review after that. And we have ENG-2949 to track solidity testing. The sysconfig functionality is covered in Go action testing with the other PRs, so I think we can split out the solidity testing to unblock the other PRs.

tynes:
I'm updating the gas snapshot, I'm also not opposed to changing the base that these commits are being opened as a PR against, I feel like the contracts changes are self contained to the point where they can be opened against `develop`

trianglesphere (approved):
lgtm once build works

packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
```diff
@@ -164,7 +156,7 @@ contract GasPriceOracle is Ownable, Semver {
                 total += 16;
             }
         }
-        uint256 unsigned = total + overhead;
+        uint256 unsigned = total + overhead();
```
maurelian, line 159: This adds a fair bit of gas overhead, but it's only meant to be called off-chain right?

packages/contracts-bedrock/.storage-layout
```diff
@@ -152,6 +150,12 @@
...
 | sequenceNumber | uint64  | 3    | 0      | 8     | contracts/L2/L1Block.sol:L1Block |
+|----------------+---------+------+--------+-------+----------------------------------|
+| batcher        | address | 3    | 8      | 20    | contracts/L2/L1Block.sol:L1Block |
+|----------------+---------+------+--------+-------+----------------------------------|
+| l1FeeOverhead  | uint256 | 4    | 0      | 32    | contracts/L2/L1Block.sol:L1Block |
+|----------------+---------+------+--------+-------+----------------------------------|
+| l1FeeScalar    | uint256 | 5    | 0      | 32    | contracts/L2/L1Block.sol:L1Block |
 +----------------+---------+------+--------+-------+----------------------------------+
 
 =======================
```
maurelian, line 161: I wonder why the `SystemConfig` contract wasn't added to this file?

tynes: I believe that we have a list of contracts to be included, its not just every contract

maurelian: oh, that sounds right. ALso something we can address in a subsequent PR.

tynes: Will fix

packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
tynes: Yes its only meant to be called offchain

maurelian (approved):
LGTM. 
There is some clean up to be done, which can be addressed in a subsequent PR.

packages/contracts-bedrock/contracts/L2/L1Block.sol
```diff
@@ -74,5 +96,8 @@ contract L1Block is Semver {
         basefee = _basefee;
         hash = _hash;
         sequenceNumber = _sequenceNumber;
+        batcherHash = _batcherHash;
```
tynes, line 99: Why did this become `batcherHash` instead of just `address _batcher`?

packages/contracts-bedrock/contracts/deployment/BaseSystemDictator.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity 0.8.15;
 
 import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
-import { SystemConfig } from "./SystemConfig.sol";
+import { SystemConfig } from "./DeployConfig.sol";
```
tynes, line 5: I renamed the file `deployment/SystemConfig.sol` so that there are not 2 files with the same name

tynes:
Unclear why the hive p2p tests are failing here, must be a flake

packages/contracts-bedrock/contracts/L2/L1Block.sol
protolambda: Because we are doing the versioned batcher hash as per the design doc, we wanted it to be forward compatible with different batcher identification.

tynes:
Fighting the tests here a bit, we need a plan of action with merging this stuff because I think the tests here are failing due to needing changes in a different PR that is in this set of PRs, cc @protolambda

protolambda:
Rebased + updated gas snapshot and bindings during merge conflicts where necessary

protolambda:
Hmm, the fuzz-op-node one fails because the fuzzing test updates are in the next PR of the stack. Might pull those updates into this PR, will see how isolated those change are first, it may just have to stay in that next PR.

mslipper:
@protolambda it's OK to leave in the next PR if it's fixed in that part of the stack. We'll merge these in quick succession.

tynes (approved):

PR #3791, opened 2022-10-26
integration-tests: update readme, try to fix ci build

tynes:
**Description**

integration tests are failing to build in ci due to a cache error, try to fix that here


tynes:
Added a diff to the `integration-tests` readme so it would trigger them running, not seeing the build issue anymore now that caching is turned off. Not sure if it fixes the problem though, would need to see the other PRs include this change

mslipper:
This looks good; we already know this works because the itests ran. I'll admin merge this to unblock other stuff.

PR #3792, opened 2022-10-26
op-chain-ops: read l2 addrs from hh deploy artifacts

tynes:
**Description**

The addresses required for the migration are now read from the hh deployments. This adds some additional CLI params to the script


tynes:
Converting this to a draft, will update to use hh deploy artifacts instead of reading l2addrs from json

mslipper (approved):

maurelian (approved):

PR #3793, opened 2022-10-26
op-node: Add frame validity checks 

trianglesphere:
**Description**

This ensures that the following frame validity checks are implemented:
* Reject duplicate frames
* Reject frames that attempt to close an already closed channel
* Reject frames with a number higher than the closing frame number
* Prune frames with a number higher than the closing frame number when we receive a closing frame.

This also updates the specs with these changes.

**Tests**

Added a tests for all of the above checks & ensured that we are properly pruning the channel.

**Metadata**
- Fixes ENG-2878


op-node/rollup/derive/channel.go
```diff
@@ -58,27 +53,49 @@ func (ch *Channel) AddFrame(frame Frame, l1InclusionBlock eth.L1BlockRef) error
...
+			if id >= uint64(ch.endFrameNumber) {
+				idsToPrune = append(idsToPrune, id)
+			}
+			ch.size -= uint64(len(prunedFrame)) + frameOverhead
+		}
+		for _, id := range idsToPrune {
+			delete(ch.inputs, id)
+		}
+		ch.highestFrameNumber = ch.endFrameNumber
+
```
tynes, line 86: nit: extra newline

trianglesphere: FYI you can make a code suggestion to remove newlines

op-node/rollup/derive/channel.go
```diff
@@ -28,10 +25,8 @@ type Channel struct {
 	// true if we have buffered the last frame
 	closed bool
 
-	// TODO: implement this check
 	// highestFrameNumber is the highest frame number yet seen.
-	// This must be equal to `endFrameNumber`
-	// highestFrameNumber uint16
+	highestFrameNumber uint16
```
tynes, line 29: Could a malicious sequencer in theory overflow this value?

trianglesphere: No, it's a uint16 on the wire: https://github.com/ethereum-optimism/optimism/blob/d734f9d93ccbe12c2fa51d0a9fab772764493fcc/op-node/rollup/derive/frame.go#L75

op-node/rollup/derive/channel.go
```diff
@@ -58,27 +53,48 @@ func (ch *Channel) AddFrame(frame Frame, l1InclusionBlock eth.L1BlockRef) error
...
+	// Update highest seen frame number after pruning
+	if frame.FrameNumber > ch.highestFrameNumber {
+		ch.highestFrameNumber = frame.FrameNumber
+	}
+
 	if ch.highestL1InclusionBlock.Number < l1InclusionBlock.Number {
 		ch.highestL1InclusionBlock = l1InclusionBlock
 	}
 	ch.inputs[uint64(frame.FrameNumber)] = frame.Data
 	ch.size += uint64(len(frame.Data)) + frameOverhead
```
tynes, line 96: Perhaps a helper function like `Frame.Size()` would be useful and replace the inline casting + overhead

op-node/rollup/derive/channel_test.go
```diff
@@ -0,0 +1,85 @@
...
+			shouldErr: []bool{true},
+			sizes:     []uint64{0},
+		},
+		{
+			name: "double close",
+			frames: []Frame{
+				{ID: id, FrameNumber: 2, IsLast: true, Data: []byte("four")},
+				{ID: id, FrameNumber: 1, IsLast: true}},
+			shouldErr: []bool{false, true},
+			sizes:     []uint64{204, 204},
```
tynes, line 54: How did you compute these sizes? Feels like hardcoding them instead of computing them dynamically isn't ideal

op-node/rollup/derive/channel_test.go
```diff
@@ -0,0 +1,85 @@
...
+		},
+		{
+			name: "duplicate frame",
+			frames: []Frame{
+				{ID: id, FrameNumber: 2, Data: []byte("four")},
+				{ID: id, FrameNumber: 2, Data: []byte("seven__")}},
+			shouldErr: []bool{false, true},
```
tynes, line 61: Is a duplicate frame an error or is the frame supposed to just be dropped? Will this prevent further parsing of the batch data?

op-node/rollup/derive/channel_test.go
trianglesphere: The size of a single frame is `200 + len(frame.Data)` (Note 200 is the frame overhead). The size of the channel is the sum of the sizes of the frames.

TBH I rather hardcode the value in the tests than dynamically compute it. If I dynamically compute it, all I'm asserting is that the function returns the same thing (which is still a useful property), but not that I've computed it properly to what I expect.

op-node/rollup/derive/channel_test.go
trianglesphere: Frame is supposed to be dropped

trianglesphere: `IngestFrame` (which uses `AddFrame` is an infallible function. It simply logs the error (+ also skips the CB prune, but this is fine b/c the size would not have grow if err != nil & the CB would have been in compliance with it's size limit prior to entering IngestFrame)

https://github.com/ethereum-optimism/optimism/blob/e081a971958157191ea9b5ad8264b8c12d58cd7c/op-node/rollup/derive/channel_bank.go#L96-L100

op-node/rollup/derive/channel.go
trianglesphere: added as a free function. I did have to change how some data was stored to keep a clean usage of it.

tynes (approved):

op-node/rollup/derive/channel.go
```diff
@@ -58,26 +53,47 @@ func (ch *Channel) AddFrame(frame Frame, l1InclusionBlock eth.L1BlockRef) error
...
+		for id, prunedFrame := range ch.inputs {
+			if id >= uint64(ch.endFrameNumber) {
+				idsToPrune = append(idsToPrune, id)
+			}
+			ch.size -= frameSize(prunedFrame)
+		}
+		for _, id := range idsToPrune {
+			delete(ch.inputs, id)
+		}
+		ch.highestFrameNumber = ch.endFrameNumber
```
protolambda, line 85: In Go it's save to delete from a map while iterating the map, so we can avoid allocating idsToPrune here I think.

```suggestion
		// Do a linear scan over saved inputs instead of ranging over ID numbers
		for id, prunedFrame := range ch.inputs {
			if id >= uint64(ch.endFrameNumber) {
				delete(ch.inputs, id)
			}
			ch.size -= frameSize(prunedFrame)
		}
		ch.highestFrameNumber = ch.endFrameNumber
```

op-node/rollup/derive/channel.go
```diff
@@ -121,11 +137,11 @@ func (ch *Channel) IsReady() bool {
 func (ch *Channel) Reader() io.Reader {
 	var readers []io.Reader
 	for i := uint64(0); i <= uint64(ch.endFrameNumber); i++ {
-		data, ok := ch.inputs[i]
+		frame, ok := ch.inputs[i]
 		if !ok {
 			panic("dev error in channel.Reader. Must be called after the channel is ready.")
 		}
-		readers = append(readers, bytes.NewBuffer(data))
+		readers = append(readers, bytes.NewBuffer(frame.Data))
```
protolambda, line 144:
```suggestion
		readers = append(readers, bytes.NewReader(frame.Data))
```
We don't need the buffer functionality here I think, so maybe a regular bytes reader is better?

op-node/rollup/derive/channel_test.go
```diff
@@ -0,0 +1,85 @@
...
+		{
+			name: "double close",
+			frames: []Frame{
+				{ID: id, FrameNumber: 2, IsLast: true, Data: []byte("four")},
+				{ID: id, FrameNumber: 1, IsLast: true}},
+			shouldErr: []bool{false, true},
+			sizes:     []uint64{204, 204},
+		},
+		{
+			name: "duplicate frame",
```
protolambda, line 57: Adding a duplicate frame where both are closing frames would be a nice test to have.

protolambda (commented):
Looks good. Added some suggestions to fix some nits.

op-node/rollup/derive/channel_test.go
trianglesphere: Done

PR #3799, opened 2022-10-27
Natspec legacy contracts update

cyborgdennett:
**Description**

I added a couple of `@param` and `@return` where there were none, and added `@inheritdoc` where there was an interface contract available. I only took a look at legacy contracts.


maurelian (approved):
LGTM thank you!

packages/contracts/contracts/L2/messaging/L2CrossDomainMessenger.sol
```diff
@@ -33,6 +33,9 @@ contract L2CrossDomainMessenger is IL2CrossDomainMessenger {
      * Constructor *
      ***************/
 
+     /**
```
smartcontracts, line 36: I think there's an extra space here

smartcontracts (commented):
This is great, thanks! Dropped one quick comment

packages/contracts/contracts/L2/messaging/L2CrossDomainMessenger.sol
cyborgdennett: Good catch! 👁️

tynes (approved):

tynes:
CI will fail due to permission errors but this should be fine to merge since its just changing comments

PR #3809, opened 2022-10-27
op-chain-ops: ether migration

tynes:
**Description**

Updates the migration script to also migrate
eth balances from their legacy erc20 format.
It will find the balances given the input generated from `l2geth` and the JS package responsible for parsing that input and then migrate the balance and delete the balance in the erc20 contract.

It also adds some additional sanity checks to ensure that the data is completely accounted for so that
the script cannot be ran with insufficient or malicious inputs.


op-chain-ops/genesis/db_migration.go
```diff
@@ -26,28 +92,55 @@ func MigrateDB(db vm.StateDB, config *DeployConfig, l1Block *types.Block, l2Addr
...
+		return nil
+	}
+
+	root, err := db.Commit(true)
+	if err != nil {
 		return err
 	}
 
-	// TODO: use migration data to double check things
+	// Create the bedrock transition block
```
tynes, line 114: The config here could cause problems, I feel like some validation must be done, ie the timestamp could go backwards

op-chain-ops/genesis/db_migration.go
```diff
@@ -26,28 +92,55 @@ func MigrateDB(db vm.StateDB, config *DeployConfig, l1Block *types.Block, l2Addr
...
+		ReceiptHash: types.EmptyRootHash,
+		Bloom:       types.Bloom{},
+		Difficulty:  (*big.Int)(config.L2GenesisBlockDifficulty),
+		Number:      new(big.Int).Add(header.Number, common.Big1),
+		GasLimit:    (uint64)(config.L2GenesisBlockGasLimit),
+		GasUsed:     (uint64)(config.L2GenesisBlockGasUsed),
+		Time:        uint64(config.L2OutputOracleStartingTimestamp),
+		Extra:       config.L2GenesisBlockExtraData,
+		MixDigest:   config.L2GenesisBlockMixHash,
+		//Nonce:       (types.BlockNonce)(config.L1GenesisBlockNonce),
```
tynes, line 130: TODO: wasn't sure how to inline this, going to need to binary.WriteUint64 i think

op-chain-ops/ether/cli.go
```diff
@@ -108,8 +99,11 @@ func Migrate(dataDir, outDir string, genesis *core.Genesis, addrLists, allowance
 		}
 		logProgress := ProgressLogger(10000, "read allowance")
 		err = IterateAllowanceList(f, func(owner, spender common.Address) error {
-			addressesToMigrate[owner] = true
-			storageSlotsToMigrate[CalcAllowanceStorageKey(owner, spender)] = 2
+			allowance := &migration.Allowance{
```
tynes, line 102: Need to double check this is the correct order

op-chain-ops/genesis/db_migration.go
```diff
@@ -26,28 +93,57 @@ func MigrateDB(db vm.StateDB, config *DeployConfig, l1Block *types.Block, l2Addr
...
+	if err != nil {
+		return err
 	}
-	for _, msg := range migrationData.EvmMessages {
-		wd, err := msg.ToLegacyWithdrawal()
-		if err != nil {
-			return err
-		}
-		messages = append(messages, wd)
+
```
tynes, line 100: TODO: need to pass in the chain id correctly, now its hard coded to 0

op-chain-ops/genesis/db_migration.go
```diff
@@ -26,28 +93,57 @@ func MigrateDB(db vm.StateDB, config *DeployConfig, l1Block *types.Block, l2Addr
...
 
-	if err := crossdomain.MigrateWithdrawals(messages, db, &l2Addrs.L1CrossDomainMessengerProxy, &l2Addrs.L1StandardBridgeProxy); err != nil {
+	root, err := db.Commit(true)
+	if err != nil {
 		return err
 	}
 
-	// TODO: use migration data to double check things
+	nonce := [8]byte{}
+	binary.LittleEndian.PutUint64(nonce[:], (uint64)(config.L1GenesisBlockNonce))
```
tynes, line 116: TODO: double check little endian

op-chain-ops/ether/migrate.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/trie"
+)
+
```
tynes, line 18: set `totalSupply` to 0

op-chain-ops/genesis/db_migration.go
trianglesphere: Do we need this block to have the state transition? I was kinda thinking there wouldn't be one, but this makes sense

trianglesphere (approved):
Generally looks good to me. I think when this is ready for the test migration, doing an internal audit would be a good idea - I think I'm missing some of the context to fully understanding everything as it's getting built.

op-chain-ops/genesis/db_migration.go
tynes: The root is used to build the genesis output commitment right?

trianglesphere: hmm, probably. I haven't thought much about it

mslipper (approved):

PR #3814, opened 2022-10-28
op-node,contracts-bedrock: dynamic gas limit via SystemConfig

protolambda:
**Description**

Dynamic gas limit: we extend the SystemConfig on L1 with a setter that emits gas limit changes, which can then be picked up during derivation, and applied to the payload attributes for block building.

Changes:
- extend SystemConfig contract with gaslimit setter, and require minimum gas limit for deposit / chain safety
- update chainops to deploy L1 contracts with gas limit
- new action test to test gas limit update via system config contract
- minor e2e / action test updates to handle new gas limit attribute and minimum gas limit
- update engine API types with new gas limit field
- extend system config with gas limit field
- check gas limit field when comparing payload attributes and block
- handle gas limit update events
- spec updates to describe config value, derivation and engine API usage.
- bindings update
- system config storage layout update
- minor test/utils updates to handle new field

This PR depends on https://github.com/ethereum-optimism/op-geth/pull/22

Once the geth PR is merged we can update the geth dependencies, and marge this PR. A temporary `go.work` entry points to the geth PR for now to make CI run.

**Tests**

Added an action test to cover the derivation changes, and the block body validation changes in the engine.

**Metadata**

Fix ENG-2935
Fix ENG-2957


protolambda:
This PR depends on https://github.com/ethereum-optimism/op-geth/pull/22

I can pull in that commit in the go.work, but I'd prefer to first resolve the dependency stack of this PR, which also depend on open geth changes.

trianglesphere:
hey @protolambda, it seems like there's a lot of system config changes, this PR should just be the dynamic gas limit

protolambda:
@trianglesphere yea, I was planning to rebase this once the sys config changes land (so close, just need op-geth PR). Github just exploded the diff here because the base branch changes. The actual changes are much smaller. I'll rebase it again

op-e2e/actions/l2_engine_api.go
```diff
@@ -42,13 +42,13 @@ func computePayloadId(headBlockHash common.Hash, params *eth.PayloadAttributes)
 	_ = binary.Write(hasher, binary.BigEndian, params.Timestamp)
```
protolambda, line 42: Extending the compute payload ID method is kind of weird, since the engine API spec defines it to be a random ID, but geth still uses a legacy method of hashing the payload attributes. I opened ENG-2976 to fix this upstream. Then we won't need this diff, nor would we have to have all this complexity in the action testing.

op-e2e/actions/l2_engine_api.go
```diff
@@ -42,13 +42,13 @@ func computePayloadId(headBlockHash common.Hash, params *eth.PayloadAttributes)
...
+	_ = binary.Write(hasher, binary.BigEndian, params.NoTxPool)
+	_ = binary.Write(hasher, binary.BigEndian, uint64(len(params.Transactions)))
 	for _, tx := range params.Transactions {
 		_ = binary.Write(hasher, binary.BigEndian, uint64(len(tx))) // length-prefix to avoid collisions
 		hasher.Write(tx)
 	}
-	if params.NoTxPool {
-		hasher.Write([]byte{1})
-	}
+	_ = binary.Write(hasher, binary.BigEndian, *params.GasLimit)
```
tynes, line 51: Do you need to do a nil check here?

specs/system_config.md
```diff
@@ -34,6 +35,12 @@ to enable more extensive redundancy and/or rotation configurations.
 The L1 fee parameters, also known as Gas Price Oracle (GPO) parameters,
 are updated in conjunction and apply new L1 costs to the L2 transactions.
 
+### `gasLimit` (`uint256`)
```
tynes, line 38: This is a `uint64`

specs/system_config.md
```diff
@@ -65,6 +73,7 @@ The contained log events are filtered and processed as follows:
   and encodes the configuration update. In version `0` the following types are supported:
   - type `0`: `batcherHash` overwrite, as `bytes32` payload.
   - type `1`: `l1FeeOverhead` and `l1FeeScalar` overwrite, as two packed `uint256` entries.
+  - type `2`: `gasLimit` overwrite, as `uint256` payload.
```
tynes, line 76: `uint64`

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -40,13 +50,16 @@ contract SystemConfig is OwnableUpgradeable, Semver {
         address _owner,
         uint256 _overhead,
         uint256 _scalar,
-        bytes32 _batcherHash
+        bytes32 _batcherHash,
+        uint64 _gasLimit
     ) public initializer {
+        require(_gasLimit >= MINIMUM_GAS_LIMIT, "SystemConfig: gas limit too low");
```
tynes, line 56: Wondering if we should just call the methods with each value passed in to remove this require check

specs/exec-engine.md
```diff
@@ -106,6 +107,10 @@ if there is an error processing the transactions. It must return `STATUS_VALID`
 be executed without error. **Note**: The state transition rules have been modified such that deposits will never fail
 so if `engine_forkchoiceUpdatedV1` returns `STATUS_INVALID` it is because a batched transaction is invalid.
 
+The `gasLimit` is optional w.r.t. compatibility with L1, but required when used as rollup.
```
tynes, line 110: Could we add test coverage of this?

op-e2e/actions/l2_engine_api.go
tynes: I didn't realize that hashing was a legacy way of computing the payload id, is the preferred way just getting random bytes?

protolambda: yes

op-e2e/actions/l2_engine_api.go
protolambda: it's always L2 here, and just quick testing code while we wait for the random ID. In op geth I have the nil check.

specs/exec-engine.md
protolambda: Figured out how/where to add the test. Bindings do support the nil gas limit value, thought they did not previously.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
protolambda: Not sure what's better here. Calling the methods would also emit log events, which I think is a bad pattern during constructors, since the events are not emitted when embedding the contract in a state, so we can't rely on that.

protolambda:
Hmm, devnet is failing on contract deployment, meaning it's failing to deploy the system config. Likely hitting the new require statement on minimum gaslimit. What gas limit do we start it with?

protolambda:
Found the issue: the L2 gas limit is set to 0, the genesis generator then uses 15M when it's unspecified, but does not implement this same fallback when inserting the gaslimit value into the initialize args.

tynes (approved):

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -16,21 +16,31 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
     uint256 public scalar;
     bytes32 public batcherHash;
+    uint64 public gasLimit;
+
+    /**
+     * @notice Minimum gas limit. This should not be lower than the maximum deposit
+     *         gas resource limit in the ResourceMetering contract used by OptimismPortal,
+     *         to ensure the L2 block always has sufficient gas to process deposits.
+     */
+    uint64 public constant MINIMUM_GAS_LIMIT = 8_000_000;
```
trianglesphere, line 26: How easy is it to tie this value to the resource metering contract value? It's good to have a check in this contract rather than not messing it up during deploys, but this can get out of sync with the other contract.

trianglesphere (approved):
This is looking pretty good. One contracts question to see if we can limit the foot guns with setting the L2 gas limit.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
protolambda: We would have to tightly couple the contracts, which complicates set up. It's a constant, so maybe we add a test that the minimum gas limit is always equal or larger than the maximum L1 resource meter gas in the portal? I'll add it to the solidity testing issue.

mslipper:
Merging so we can fix Hive.

PR #3816, opened 2022-10-28
op-batcher: Restructure source code

trianglesphere:
**Description**

This moves everything into a `batcher` directory. It's maybe not the perfect name yet, but this is getting split into enough components that having all of the source files in the top level directory was becoming annoying.

**Additional context**

I think the code is getting large enough to make this change, but I don't love the stutter with `op-batcher/batcher`. If people have others ideas here I'd love to hear them.

### DIFF
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -1,4 +1,4 @@
-package op_batcher
+package batcher
 
 import (
 	"bytes"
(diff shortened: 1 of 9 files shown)


protolambda (approved):

tynes:
Looks like a package in `op-e2e` needs an update as well

trianglesphere:
Closing this because it's actually pretty tricky with the `go.work` file. `go mod tidy` does not respect `go.work` use statements. This causes problems because `op-e2e` depends on a newer version of `op-batcher` than is published.

FYI - running tests in op-e2e fails if `GOWORK=off` fails because it depends on newer version of `op-*` than what we've cut as releases.

PR #3817, opened 2022-10-29
ci: Fail if new bedrock contract code is not 100% tested

maurelian:
**Description**
 Two changes to the codecov config:

1. Fails CI if bedrock contracts test coverage is reduced by a PR. 
2. Turns on the commenting bot.


mslipper (approved):

mslipper:
Thanks for this!

PR #3818, opened 2022-10-29
op-node: catch up L2 sequencing based on time, not L1 origins

protolambda:
**Description**

Make sure to produce a L2 chain up to and including the latest wall clock timestamp. The L1 origin that we adopt doesn't matter in the catchup logic. When producing the block, and older L1 origin can be selected based on the conf depth, but the L2 chain should still be produced.

**Tests**

Fix based on alpha-1 testnet syncing occurence. Can extend testing in e2e for this later.


tynes:
Any plan for this PR to be actually merged?

protolambda:
@tynes yes, this is an important fix, thank you for reminding me.

tynes (approved):

trianglesphere (approved):
lgtm

Issue #3819, opened 2022-10-31
getDepositsByAddress and getWithdrawalsByAddress Bug


guoshijiang:
**Describe the bug**
use getWithdrawalsByAddress to get withdrawals and use getDepositsByAddress to get deposits，but retrun erros

1、the address had Withdrawal return error:
const withdrawals = await sdk.getWithdrawalsByAddress(address)
error: Error: input is not a hex string: undefined

2、the eoa address unwithdrawal return []
const withdrawals = await sdk.getWithdrawalsByAddress(address)

The same goes for getting deposits

**To Reproduce**
1、the address had Withdrawal return error:
const withdrawals = await sdk.getWithdrawalsByAddress(address)
error: Error: input is not a hex string: undefined

2、the eoa address unwithdrawal return []
const withdrawals = await sdk.getWithdrawalsByAddress(address)
**Expected behavior**

**Screenshots**

**System Specs:**
 - OS:
 - Package Version (or commit hash):

**Additional context**

guoshijiang:
This issure is not exist in bedrock,  It exist in old version optimist

PR #3820, opened 2022-10-31
getDepositsByAddress and getWithdrawalsByAddress Bug

guoshijiang:
**Description**
use getWithdrawalsByAddress to get withdrawals and use getDepositsByAddress to get deposits，but retrun erros

1、the address had Withdrawal return error:
const withdrawals = await sdk.getWithdrawalsByAddress(address)
error: Error: input is not a hex string: undefined

2、the eoa address unwithdrawal return []
const withdrawals = await sdk.getWithdrawalsByAddress(address)

The same goes for getting deposits

**Tests**

**Additional context**

**Metadata**
- Fixes #[https://github.com/ethereum-optimism/optimism/issues/3819]


tynes:
This was known to work in the past, the ABIs must have changed breaking things. Tagging @smartcontracts

guoshijiang:
It is not exist in bedrock, so I close it.

PR #3821, opened 2022-10-31
change  pint log ways

guoshijiang:
If you use github.com/ethereum-optimism/optimism/l2geth/log, the log will not be printed, because the log initialization uses github.com/ethereum/go-ethereum/log

**Description**

If you use github.com/ethereum-optimism/optimism/l2geth/log, the log will not be printed, because the log initialization uses github.com/ethereum/go-ethereum/log

**Tests**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


tynes:
Closing this PR as its changing legacy code that is known to work

PR #3825, opened 2022-10-31
feat: L1 fee + base fee vault predeploys

tynes:
**Description**

Adds 2 new predeploy addresses that become protocol constants. The L1 portion of the fee will be paid out to a particular predeploy and the basefee will be paid out to a particular predeploy instead of being burnt. This allows for the way that the fees are handled to be moved to the application layer, these contracts can be upgraded after the fact with alternative logic. Right now they are simple implementations of the way that the `SequencerFeeVault` works already

This PR is large because all of the changes in `op-bindings`

A companion PR can be found here: https://github.com/ethereum-optimism/op-geth/pull/23
Once this is merged into `op-geth` and the latest commit is added as a dep here, then we can further reduce config


tynes:
On the deployed contracts devnet

```
Error: expected 1 message, got 0
    at CrossChainMessenger.toCrossChainMessage (/home/circleci/project/packages/sdk/src/cross-chain-messenger.ts:517:15)
```

packages/contracts-bedrock/.storage-layout
```diff
@@ -259,11 +259,10 @@
...
 =======================
 
-+-------------+---------+------+--------+-------+------------------------------------------------------+
-| Name        | Type    | Slot | Offset | Bytes | Contract                                             |
-+======================================================================================================+
-| l1FeeWallet | address | 0    | 0      | 20    | contracts/L2/SequencerFeeVault.sol:SequencerFeeVault |
-+-------------+---------+------+--------+-------+------------------------------------------------------+
++------+------+------+--------+-------+----------+
+| Name | Type | Slot | Offset | Bytes | Contract |
++================================================+
```
tynes, line 264: Perhaps I should be adding a spacer here as to not change the storage layout

op-bindings/Makefile
```diff
@@ -108,7 +116,7 @@ more:
 	go run ./gen/main.go \
 		-artifacts ../packages/contracts-bedrock/artifacts \
 		-out ./bindings \
-		-contracts OptimismMintableERC20Factory,L2StandardBridge,L1BlockNumber,LegacyMessagePasser,DeployerWhitelist,Proxy,OptimismPortal,L2ToL1MessagePasser,L2CrossDomainMessenger,GasPriceOracle,SequencerFeeVault,L1Block,LegacyERC20ETH,WETH9,GovernanceToken,L1CrossDomainMessenger,L2ERC721Bridge,OptimismMintableERC721Factory,ProxyAdmin \
```
tynes, line 111: able to remove the `SequencerFeeVault` because we no longer need its storage layout, the recipient is now an immutable which is easier to manage since its behind a proxy

packages/contracts-bedrock/contracts/universal/FeeVault.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Semver } from "../universal/Semver.sol";
```
tynes, line 4: Semver is a dead import here

packages/contracts-bedrock/contracts/universal/FeeVault.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @notice Minimum balance before a withdrawal can be triggered.
+     */
+    uint256 public immutable MIN_WITHDRAWAL_AMOUNT;
+
+    /**
+     * @notice Wallet that will receive the fees on L1.
+     */
+    address public immutable recipient;
```
maurelian, line 25: These should be consistently cased.

packages/contracts-bedrock/contracts/universal/FeeVault.sol
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * @notice Triggers a withdrawal of funds to the L1 fee wallet.
+     */
+    function withdraw() external {
+        require(
+            address(this).balance >= MIN_WITHDRAWAL_AMOUNT,
+            "FeeVault: withdrawal amount must be greater than minimum withdrawal amount"
+        );
+
+        uint256 value = address(this).balance;
```
maurelian, line 51: nit, but this could just be inlined below.

packages/contracts-bedrock/contracts/universal/FeeVault.sol
```diff
@@ -0,0 +1,61 @@
...
+
+        L2StandardBridge(payable(Predeploys.L2_STANDARD_BRIDGE)).bridgeETHTo{ value: value }(
+            recipient,
+            20000,
+            bytes("")
+        );
```
maurelian, line 57: This is really an ops decision, but why do we need to have the funds on L1?

packages/contracts-bedrock/contracts/test/FeeVault.t.sol
```diff
@@ -0,0 +1,33 @@
...
+contract BFeeVault_Test is Bridge_Initializer {
+    BaseFeeVault baseFeeVault = BaseFeeVault(payable(Predeploys.BASE_FEE_VAULT));
+    L1FeeVault l1FeeVault = L1FeeVault(payable(Predeploys.L1_FEE_VAULT));
+
+    address constant recipient = address(0x10000);
+
+    function setUp() public override {
+        super.setUp();
+        vm.etch(Predeploys.BASE_FEE_VAULT , address(new BaseFeeVault(recipient)).code);
+        vm.etch(Predeploys.L1_FEE_VAULT , address(new L1FeeVault(recipient)).code);
```
maurelian, line 21: is this spacing before the `,` intentional? I'm surprised the formatter doesn't remove it, unless it's not actually operating on the test files despite #3663?

packages/contracts-bedrock/contracts/universal/FeeVault.sol
tynes: The event needs the value from before the transfer

packages/contracts-bedrock/contracts/test/FeeVault.t.sol
tynes: That is a typo, good catch, will fix

packages/contracts-bedrock/contracts/universal/FeeVault.sol
maurelian: Ah, OK.
That said, moving the event before the call would be more CEI-compliant anyways, but fine with caching the actual value if you prefer the style.

tynes: Ok i'm not super opinionated, can move the event before due to cei

tynes:
Need to regenerate the more files

packages/contracts-bedrock/contracts/universal/FeeVault.sol
tynes: Will make this all caps to follow the recent decision that constants are all caps

tynes:
I've addressed all feedback on this PR @maurelian @smartcontracts

packages/contracts-bedrock/contracts/universal/FeeVault.sol
maurelian: thanks :)

packages/contracts-bedrock/contracts/universal/FeeVault.sol
```diff
@@ -0,0 +1,60 @@
...
+pragma solidity 0.8.15;
+
+import { L2StandardBridge } from "../L2/L2StandardBridge.sol";
+import { Predeploys } from "../libraries/Predeploys.sol";
+
+/**
+ * @title FeeVault
+ * @notice The FeeVault contract has the base logic for handling transaction fees.
+ */
+contract FeeVault {
```
maurelian, line 11: Nit, I'll approve anyways.

```suggestion
abstract contract FeeVault {
```

maurelian (approved):

mslipper (approved):

PR #3827, opened 2022-11-01
contracts: lint

tynes:
**Description**

Lints the SCC contract, which recently started failing linting. Unclear why it was not previously failing linting


maurelian (approved):
Weird. 
LGTM tho

PR #3830, opened 2022-11-01
Foundry test suite clean up

maurelian:
**Description**

**Test organizing principles**

1. Uses `contract`s to organize the test suite similar to how mocha uses `describe`. 
1. Every non-trivial state changing function should have a contract for happy and sad path tests. This helps to make it very obvious where there are not sad tests (I haven't added any new tests, so there are quite a few empty placeholder contracts). 
1. Simpler functions like getters and setters are grouped together into test contracts.
2. All contracts and functions should be consistently named following a convention. 

**Test naming convention**

Test function names are split by underscores, into 3 or 4 parts. An example function name is `test_onlyOwner_callerIsNotOwner_reverts()`. 

The parts are: `[method]_[FunctionName]_[reason]_[success]`, where:
1. `[method]` is either `test` or `testFuzz`
2. `[FunctionName]` is the name of the function or higher level behavior being tested.
4. `[reason]` is an optional description for the behavior being tested.
3. `[success]` must be one of `success` or `reverts` or `_fails`. 

**Name validation**

I also wrote a quick and dirty script which can be run to validate function names and identify ones which are not yet compliant. From within the bedrock package:
`ts-node scripts/forges-test-names.ts`. 

It will error, because there are still test files which haven't yet been cleaned up.


packages/contracts-bedrock/contracts/test/CrossDomainOwnable.t.sol
```diff
@@ -38,7 +34,15 @@ contract CrossDomainOwnable_Test is CommonTest {
     }
 }
 
-contract CrossDomainOwnableThroughPortal_Test is Portal_Initializer {
+contract CrossDomainOwnable_OnlyOwner_TestFail is CrossDomainOwnable_TestInit {
+    // Check that the revert message is correct
+    function test_onlyOwner_callerIsNotOwner_revert() external {
```
smartcontracts, line 39:
```suggestion
    function test_onlyOwner_callerIsNotOwner_reverts() external {
```

smartcontracts (approved):
I really like this refactor. It makes holes in test coverage much more obvious. Thanks!

packages/contracts-bedrock/contracts/test/Encoding.t.sol
```diff
@@ -5,19 +5,31 @@ import { CommonTest } from "./CommonTest.t.sol";
...
 import { Encoding } from "../libraries/Encoding.sol";
 
-contract Encoding_Test is CommonTest {
+contract Encoding_TestInit is CommonTest {
     function setUp() external {
         _setUp();
     }
+}
+
+contract Encoding_EncodeVersionedNonce_TestFail is Encoding_TestInit {
```
tynes, line 14: Why is this included if there are none?

packages/contracts-bedrock/scripts/forges-test-names.ts
```diff
@@ -0,0 +1,77 @@
...
+    ['reverts', 'fails'].includes(parts[parts.length - 1]) &&
+    parts.length < 4
+  ) {
+    throw new Error(
+      `Invalid test name: ${name}. Failure tests should have 4 parts. The third part should indicate the reason for failure.`
+    )
+  }
+}
+
+// Todo
```
tynes, line 49: What is this todo?

packages/contracts-bedrock/scripts/forges-test-names.ts
```diff
@@ -0,0 +1,77 @@
+import fs from 'fs'
```
tynes, line 1: Should this script run as part of CI?

tynes:
Generally makes sense, I think we need to document the guidelines in the readme that you posted here

Also this PR needs a rebase to fix the conflicts with the more files

packages/contracts-bedrock/contracts/test/Encoding.t.sol
maurelian: The intention is to express that I've actually thought about whether or not there are any sad path tests that could be written here, and determined that there are none.
I like having it, but I'm not attached to it.

packages/contracts-bedrock/scripts/forges-test-names.ts
maurelian: Updated this to say "Todo: Implement a check to validate test contract names"

packages/contracts-bedrock/scripts/forges-test-names.ts
maurelian: eventually yes, but I haven't yet gone through all the test files.

maurelian: It would fail if run in CI now.

maurelian:
@tynes 
I documented the testing conventions here: [`a61feaa` (#3830)](https://github.com/ethereum-optimism/optimism/pull/3830/commits/a61feaa7e5455a192749cddcd73e567efeb888e9)

maurelian:
just realized this is still open here. Will rebase and fix up by EOD.

PR #3790, opened 2022-10-26
op-e2e: system config action tests

protolambda:
**Description**

This PR introduces:
- an action test for batcher key rotation, with reorging, old-key batches, etc.
- an action test for GPO params updating, with L2 tx on block on boundary and L1 fee in receipts
- introduces some minor fixes/helpers in the action-testing

This is part 4/5 to introduce batcher-key and GPO param configuration:
1. Derivation aware of system config (with placeholders for later actual contract/event changes) - #3787
2. Contract changes: system config on L1, and updated L2 contracts - #3788
3. Update bindings, update chainops to use updated bindings and dev-deploy system config, and misc integration like log parsing - #3789
4. Action-test the batcher key change and GPO params change. - #3790
5. Spec updates - #3831

**Tests**

Testing-only, covers the derivation changes in part 2 and 3 of the PR stack.

PR #3789, opened 2022-10-26
op-bindings,op-chain-ops,op-node: System config bindings and integration

protolambda:
**Description**

This PR:
- updates the Go bindings to the contracts
- updates op-chain-ops to use new bindings and deploy the system config
- update op-node to format the L1 block info tx in the new way, and fill derivation placeholders (apply receipts from system config to the sys config in derivation)

This is part 3/5 to introduce batcher-key and GPO param configuration:
1. Derivation aware of system config (with placeholders for later actual contract/event changes) - #3787
2. Contract changes: system config on L1, and updated L2 contracts - #3788
3. Update bindings, update chainops to use updated bindings and dev-deploy system config, and misc integration like log parsing - #3789
4. Action-test the batcher key change and GPO params change. - #3790
5. Spec updates - #3831

Do not merge: this is a breaking change, and the other parts of the PR stack should be reviewed first.

**Tests**

See part 4 with action tests for key rotation and GPO params updating.

PR #3831, opened 2022-11-01
specs: system configuration contract and derivation [bedrock]

protolambda:
**Description**

Define the system configuration updates, reading, derivation and corresponding L1 block info changes in the specification.

This specifies the system configuration changes in #3787, #3788, #3789, #3790.

### DIFF
--- a/specs/system_config.md
+++ b/specs/system_config.md
@@ -13,7 +13,7 @@
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 The `SystemConfig` is a contract on L1 that can emit rollup configuration changes as log events.
-The rollup [block derivation process](./derivation.md) picks up on these log events, and applies the changes.
+The rollup [block derivation process](./derivation.md) picks up on these log events and applies the changes.
 
 ## System config contents (version 0)
 
@@ -39,7 +39,7 @@ are updated in conjunction and apply new L1 costs to the L2 transactions.
 The `SystemConfig` contract applies authentication to all writing contract functions,
 the configuration management can be configured to be any type of ethereum account or contract.
 
-On write, an event is emitted for the change to be picked up by the L2 system,
+On a write, an event is emitted for the change to be picked up by the L2 system,
 and a copy of the new written configuration variable is retained in L1 state to read with L1 contracts.
 
 ## Reading the system config
@@ -53,7 +53,7 @@ A rollup node initializes its derivation process by finding a starting point bas
   - other future variables may also be retrieved from other contents of the L2 block, such as the header.
 
 After preparing the initial system configuration for the given L1 starting input,
-with each new L1 block the system configuration is updated by processing all receipts from the block.
+the system configuration is updated by processing all receipts from each new L1 block.
 
 The contained log events are filtered and processed as follows:
 
--- a/specs/system_config.md
+++ b/specs/system_config.md
@@ -4,7 +4,7 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**
 
-- [System config contents](#system-config-contents)
+- [System config contents (version 0)](#system-config-contents-version-0)
   - [`batcherHash` (`bytes32`)](#batcherhash-bytes32)
   - [`l1FeeOverhead` and `l1FeeScalar` (`uint256,uint256`)](#l1feeoverhead-and-l1feescalar-uint256uint256)
 - [Writing the system config](#writing-the-system-config)
@@ -15,7 +15,9 @@
 The `SystemConfig` is a contract on L1 that can emit rollup configuration changes as log events.
 The rollup [block derivation process](./derivation.md) picks up on these log events, and applies the changes.
 
-## System config contents
+## System config contents (version 0)
+
+Version 0 of the system configuration contract defines the following parameters:
 
 ### `batcherHash` (`bytes32`)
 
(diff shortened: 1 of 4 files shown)


specs/system_config.md
```diff
@@ -0,0 +1,68 @@
...
+
+- When started from L2 genesis, the initial system configuration is retrieved from the rollup chain configuration.
+- When started from an existing L2 chain, a previously included L1 block is determined as derivation starting point,
+  and the system config can thus be retrieved from the last L2 block that referenced the L1 block as L1 origin:
+  - `batcherHash`, `l1FeeOverhead` and `l1FeeScalar` are retrieved from the L1 block info transaction.
+  - other future variables may also be retrieved from other contents of the L2 block, such as the header.
+
+After preparing the initial system configuration for the given L1 starting input,
+with each new L1 block the system configuration is updated by copying the configuration
+and then applying any emitted log events from the L1 `SystemConfig` contract:
```
trianglesphere, line 55: Need to specify that the system config is updated by the receipts of a block before scanning the block's transactions & the GPO parameters take effect in the first L2 block  whose L1 origin is the block in which it was changed.

specs/system_config.md
```diff
@@ -0,0 +1,68 @@
...
+- [Reading the system config](#reading-the-system-config)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+The `SystemConfig` is a contract on L1 that can emit rollup configuration changes as log events.
+The rollup [block derivation process](./derivation.md) picks up on these log events, and applies the changes.
+
+## System config contents
+
+### `batcherHash` (`bytes32`)
```
tynes, line 20: Got it this makes sense, although we have multiple levels of versioning going on, since there is a top level config version and now the `batcherHash` itself is versioned and the first byte specifies the version.

Does this versioned batcher scheme only apply for system config version 0? That is where I think there is ambiguity here

specs/system_config.md
```diff
@@ -0,0 +1,68 @@
...
+  - [`batcherHash` (`bytes32`)](#batcherhash-bytes32)
+  - [`l1FeeOverhead` and `l1FeeScalar` (`uint256,uint256`)](#l1feeoverhead-and-l1feescalar-uint256uint256)
+- [Writing the system config](#writing-the-system-config)
+- [Reading the system config](#reading-the-system-config)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+The `SystemConfig` is a contract on L1 that can emit rollup configuration changes as log events.
+The rollup [block derivation process](./derivation.md) picks up on these log events, and applies the changes.
+
```
tynes, line 17: I think the following values should be below a system config version 0 section and we should mention that the schema of the system config can be updated in the future

specs/system_config.md
tynes: I think it makes sense to make each of the defined values properties of system config version 0 and then we can define system config version 1 later and if we need to use batcher hash version 1 before that then thats fine, and we can decide the semantics around the batcher hash with system config version 1 when it comes time

protolambda: > Does this versioned batcher scheme only apply for system config version 0?

Maybe, that's not been discussed yet. In the design doc phase we agreed on formatting it as bytes32. If you believe the event versioning is sufficient we can revert it back to just a bytes20 after discussing with design doc reviewers. I'm fine with it being either type.

specs/system_config.md
protolambda: Renamed the section of constants and added a description to define this as version 0.

tynes (approved):

specs/system_config.md
```diff
@@ -0,0 +1,71 @@
...
+A rollup node initializes its derivation process by finding a starting point based on its past L2 chain:
+
+- When started from L2 genesis, the initial system configuration is retrieved from the rollup chain configuration.
+- When started from an existing L2 chain, a previously included L1 block is determined as derivation starting point,
+  and the system config can thus be retrieved from the last L2 block that referenced the L1 block as L1 origin:
+  - `batcherHash`, `l1FeeOverhead` and `l1FeeScalar` are retrieved from the L1 block info transaction.
+  - other future variables may also be retrieved from other contents of the L2 block, such as the header.
+
+After preparing the initial system configuration for the given L1 starting input,
+with each new L1 block the system configuration is updated by processing all receipts from the block.
```
trianglesphere, line 56:
```suggestion
the system configuration is updated by processing all receipts from each new L1 block.
```

specs/system_config.md
```diff
@@ -0,0 +1,71 @@
...
+
+The contained log events are filtered and processed as follows:
+
+- the log event contract address must match the rollup `SystemConfig` deployment
+- the first log event topic must match the ABI hash of `ConfigUpdate(uint256,uint8,bytes)`
+- the second topic determines the version. Unknown versions are critical derivation errors.
```
trianglesphere, line 62: Should we be rejecting invalid events instead of blowing up the op-node?

trianglesphere (commented):
lots of grammar & wording nits, content is good.

specs/system_config.md
protolambda: Good point, ignoring the events is better, since a malicious/bad L1 system config contract upgrade can't harm the chain as much. Although when the system config is emitting bad things we might want to prefer safety over liveness. I'll open a ticket for this.

protolambda: Opened ENG-2974 and tackling next cycle. As long as we don't emit a bad event, handling it one way or another is compatible, so it does not block a testnet. But I would like to discuss more + update spec/implementation with the safety vs liveness decision.

trianglesphere (approved):

PR #3832, opened 2022-11-01
Optimize `_verifyWithdrawalInclusion` in `OptimismPortal.sol`

clabby:
# Overview

**Description**

Optimizes the hashing process of the `storageKey` in `OptimismPortal`'s `_verifyWithdrawalInclusion` function.

**Tests**

Existing tests have not been modified.

**Additional context**

See the gas snapshot diff.

**Metadata**
n/a


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -303,19 +303,42 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
+            // as a 32-byte 0 value (the slot of the withdrawals mapping).
+            mstore(0x40, _withdrawalHash)
+
+            // Hash 64 bytes of memory between [0x40:0x80) and store the digest
+            // within the first word of `storageKey`.
+            mstore(add(storageKey, 0x20), keccak256(0x40, 0x40))
+
+            // SAFETY: Restore the free memory pointer
+            mstore(0x40, ptr)
+        }
```
maurelian, line 340: This is really amazing, but I think we should heavily prioritize readability over efficiency _for now_, and I'd prefer not to introduce new assembly blocks. :)

tynes: This code path is fuzzed in the tests so I do feel confident in these changes but there is a big tradeoff between readability and gas savings. This saves ~100 gas which may not be worth the maintainability tradeoff

tynes:
I think for low level operations, hiding the assembly behind a library function could make it easier to read

PR #3833, opened 2022-11-01
ctb: Use an assert for unreachable code in Hashing.sol

maurelian:
**Description**

Uses an assert with comment to indicate code believed unreachable. 

**Tests**

No new tests.

### DIFF
--- a/packages/contracts-bedrock/contracts/libraries/Hashing.sol
+++ b/packages/contracts-bedrock/contracts/libraries/Hashing.sol
@@ -71,9 +71,9 @@ library Hashing {
             return hashCrossDomainMessageV0(_target, _sender, _data, _nonce);
         } else if (version == 1) {
             return hashCrossDomainMessageV1(_nonce, _sender, _target, _value, _gasLimit, _data);
-        } else {
-            revert("Hashing: unknown cross domain message version");
         }
+        // Should be unreachable as the version number is hardcoded into the CrossDomainMessenger.
+        assert(false);
     }
 
     /**


packages/contracts-bedrock/contracts/libraries/Hashing.sol
```diff
@@ -71,9 +71,9 @@ library Hashing {
             return hashCrossDomainMessageV0(_target, _sender, _data, _nonce);
         } else if (version == 1) {
             return hashCrossDomainMessageV1(_nonce, _sender, _target, _value, _gasLimit, _data);
-        } else {
-            revert("Hashing: unknown cross domain message version");
```
smartcontracts, line 75: I don't think this is unreachable. A user might just provide a bogus message that triggers this.

tynes:
Not sure if this function is in the hot code path anymore

packages/contracts-bedrock/contracts/libraries/Hashing.sol
maurelian: Yeah, I was thinking this was still used by one of the production contracts which would constrain the input, but the only usage of this high-level function is in the tests. Closing.

maurelian: Yeah, I was thinking this was still used by one of the production contracts which would constrain the input, but the only usage of this high-level function is in the tests. Closing.

maurelian:
> Not sure if this function is in the hot code path anymore

You're right. Looks like it's just a utility in the tests now.

PR #3847, opened 2022-11-02
feat(bedrock): Add input fuzzer for `MerkleTrie` tests

clabby:
# Overview
**Description**

Adds an input fuzzer program for bedrock's `MerkleTrie` tests for use in #3771.

*Checklist*
- [x] Happy path mode
- [ ] Modes for generating various error cases
- [ ] Ensure that the `patricia` binary is built in the CI
- [ ] Add information to the README.
- [ ] Add build script for the binary in the `package.json`

**Tests**

Adds the `testFuzzTrie_validProofs` test.
*todo- error case fuzzing*

**Additional context**

n/a

**Metadata**
n/a


packages/contracts-bedrock/scripts/merkle-trie-fuzzer/README.md
```diff
@@ -0,0 +1,22 @@
+# `merkle-trie-fuzzer`
+
+A lightweight, efficient Merkle Trie fuzzer for use in Optimism bedrock's [Merkle Trie Verifier](https://google.com).
```
tynes, line 3: Looks like a wrong link?

tynes (commented):
Before we can merge this I think we will need to talk during office hours about introducing rust into the codebase. I do think for fuzzing its appropriate, we currently use JS for fuzzing and its really slow. The fuzzing exists in JS currently because we had implementations of primitives in JS but since then, I've implemented a lot of our primitives in go so i was thinking of porting the fuzzing to go.

If we have consensus on adding rust, to merge this we will need to add the rust toolchain to the `ci-builder` docker image that runs our tests as well as something to the `package.json` so that CI will run these tests. There is a `build:differential` command that would be appropriate to include building the fuzzer

packages/contracts-bedrock/scripts/merkle-trie-fuzzer/README.md
clabby: Oops- yep, forgot to fill the placeholder. Updated 🤝

clabby:
> Before we can merge this I think we will need to talk during office hours about introducing rust into the codebase. I do think for fuzzing its appropriate, we currently use JS for fuzzing and its really slow. The fuzzing exists in JS currently because we had implementations of primitives in JS but since then, I've implemented a lot of our primitives in go so i was thinking of porting the fuzzing to go.
> 
> If we have consensus on adding rust, to merge this we will need to add the rust toolchain to the `ci-builder` docker image that runs our tests as well as something to the `package.json` so that CI will run these tests. There is a `build:differential` command that would be appropriate to include building the fuzzer

Definitely- will bring this up during one of our meetings today. Just pushed a TS version of the fuzzer that I wrote yesterday between classes, it does introduce a significant slowdown:
Screen Shot 2022-11-02 at 4 40 39 PM

Spoke with Matt about this yesterday, adding the Rust toolchain into the CI builder may be more trouble than its worth, though. We could:
* Pre-build a binary for the CI's arch so that it does not need to build on each run.
* Add an ENV var that enables/disables these tests.

If that's not a route we want to take, making a Go impl might be a nice middle-ground.

mslipper:
Closing this in response to our discussion above.

PR #3853, opened 2022-11-02
feat(ctb): introduce layout lock

smartcontracts:
**Description**
Introduces layout-lock.json which describes specific variables that should exist in specific contracts. Keeps the same logic that will also check that these variables are in the correct positions.


packages/contracts-bedrock/contracts/deployment/BaseSystemDictator.sol
```diff
@@ -1,14 +1,76 @@
...
+import { PortalSender } from "./PortalSender.sol";
 import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
-import { SystemConfig } from "./SystemConfig.sol";
 
 /**
  * @title BaseSystemDictator
  * @notice The BaseSystemDictator is a base contract for SystemDictator contracts.
  */
 contract BaseSystemDictator is Ownable {
+    /**
```
smartcontracts, line 20: Moved this into here because otherwise I would get an error because `SystemConfig.sol` doesn't have a buildInfo file.

tynes: Ok i'm into the idea of moving them into this file

packages/contracts-bedrock/tasks/validate-spacers.ts
```diff
@@ -12,10 +16,37 @@ task(
       continue
     }
 
+    // Some files may not have buildInfo (certain libraries). We can safely skip these because we
+    // make sure that everything is accounted for anyway.
     const buildInfo = await hre.artifacts.getBuildInfo(name)
+    if (buildInfo === undefined) {
```
tynes, line 22: Perhaps logging here would be useful as to not have a silent error go without us knowing

packages/contracts-bedrock/tasks/validate-spacers.ts
```diff
@@ -12,10 +16,37 @@ task(
...
+      continue
+    }
+
     for (const source of Object.values(buildInfo.output.contracts)) {
       for (const [contractName, contract] of Object.entries(source)) {
         const storageLayout = (contract as any).storageLayout
+
+        // Check that the layout lock is respected.
+        if (layoutLock[contractName]) {
+          const removed = layoutLock[contractName].filter((variable: any) => {
```
tynes, line 32: This is kind of hard to read, very nested

packages/contracts-bedrock/layout-lock.json
```diff
@@ -0,0 +1,20 @@
+{
```
tynes, line 1: Can you document how this file is supposed to work?

packages/contracts-bedrock/tasks/validate-spacers.ts
smartcontracts: It's not really an error. Since we do "accounting" based on the lock file it doesn't really matter even if there were an error here.

smartcontracts: Adding a log anyway

packages/contracts-bedrock/tasks/validate-spacers.ts
smartcontracts: Updated

packages/contracts-bedrock/layout-lock.json
smartcontracts: Added to README

packages/contracts-bedrock/layout-lock.json
```diff
@@ -0,0 +1,60 @@
...
+  "GasPriceOracle": {
+    "spacer_1_0_32": {
```
smartcontracts, line 3: This file looks weird because it's only being used for spacers, but slot/offset/length will be important if we add other variables to this file.

mslipper:
@tynes if this LGTY I can update the bindings conflicts.

tynes (approved):

PR #3854, opened 2022-11-02
fix(ctb): remove AddressManager from config

smartcontracts:
**Description**
Removes AddressManager from deploy config, not necessary since we can just pull it from the deployments folder.

### DIFF
--- a/packages/contracts-bedrock/deploy-config/mainnet-forked.json
+++ b/packages/contracts-bedrock/deploy-config/mainnet-forked.json
@@ -28,7 +28,5 @@
   "finalizationPeriodSeconds": 2,
 
   "deploymentWaitConfirmations": 1,
-  "fundDevAccounts": false,
-
-  "addressManager": "0xdE1FCfB0851916CA5101820A69b13a4E276bd81F"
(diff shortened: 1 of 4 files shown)


tynes:
For this to work on all networks, we will need to have this part of the hardhat config be updated for each network:
https://github.com/ethereum-optimism/optimism/blob/8fccc4334eb15c6d105e36680363eeef367ae5b0/packages/contracts-bedrock/hardhat.config.ts#L352

tynes (approved):

tynes:
Something is up with the newly merged mergify rules

```
Rule: Automatic merge on approval (queue) — Cannot use `speculative_checks` with queue action.
```

```
Cannot use `speculative_checks` with queue action.
warning The subscription needs to be updated to enable this feature.
```

tynes (changes requested):
Could you add actual mainnet to the hh config for the hh deploy deployments entry? So we don't forget then

smartcontracts:
> Could you add actual mainnet to the hh config for the hh deploy deployments entry? So we don't forget then

Done

tynes (approved):

PR #3856, opened 2022-11-02
feat(ctb): break out FSD steps

smartcontracts:
**Description**
Breaks out the FreshSystemDictator execution steps into a separate deployment stage and adds new post-deploy checks. Also fixes a bug in the L1CrossDomainMessenger initializer that only impacted fresh deployments.


tynes (approved):

tynes:
with a bindings rebuild this should be good to merge

packages/contracts-bedrock/contracts/deployment/BaseSystemDictator.sol
```diff
@@ -79,14 +79,14 @@ contract BaseSystemDictator is Ownable {
     /**
      * @notice Current step;
      */
-    uint256 public currentStep = 1;
+    uint8 public currentStep = 1;
```
maurelian, line 82: What does the `uint8` get us?

maurelian (approved):
one question, but lgtm

packages/contracts-bedrock/contracts/deployment/BaseSystemDictator.sol
smartcontracts: Dumb reason, means the result from `ethers` isn't a BigNumber so I don't need to do `.toNumber()`. And we'll never have 255 steps here (dear lord I hope we never have 255 steps) so not a problem.

PR #3859, opened 2022-11-02
op-batcher: Rewrite ChannelManager

trianglesphere:
**Description**

The channel manager is now responsible for tracking in flight frames and properly handling different error conditions.

A channel is a collection of frames. We have a single frame per L1 transaction.
When frames are submitted to L1 even if all frames are successfully included, their
inclusion may not be timely. The tx confirmation method handles resubmitting frames
if the transaction fails or voiding the entire channel if the channel times out.

It has hook points for how large a range of blocks to select, but does not use them
yet. It simply chooses up to 100 blocks (but it will also not submit channels that
are too large).

**Tests**

No unit tests yet. op-e2e still passing.

### DIFF
--- a/op-batcher/channel_manager.go
+++ b/op-batcher/channel_manager.go
@@ -116,6 +116,11 @@ func (s *channelManager) pendingChannelIsTimedOut() bool {
 	if s.pendingChannel == (derive.ChannelID{}) {
 		return false // no channel to be timed out
 	}
+	// No confirmed transactions => not timed out
+	if len(s.confirmedTransactions) == 0 {
+		return false
+	}
+	// If there are confirmed transactions, find the first + last confirmed block numbers
 	min := uint64(math.MaxUint64)
 	max := uint64(0)
 	for _, inclusionBlock := range s.confirmedTransactions {
--- a/op-batcher/channel_manager.go
+++ b/op-batcher/channel_manager.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"errors"
 	"io"
+	"math"
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
 	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
 )
 
 var ErrReorg = errors.New("block does not extend existing chain")
@@ -22,10 +24,42 @@ type taggedData struct {
 	id   txID
 }
 
+// channelManager stores a contiguous set of blocks & turns them into channels.
+// Upon receiving tx confirmation (or a tx failure), it does channel error handling.
+//
+// For simplicity, it only creates a single pending channel at a time & waits for
+// the channel to either successfully be submitted or timeout before creating a new
+// channel.
+// Functions on channelManager are not safe for concurrent access.
 type channelManager struct {
-	// All blocks since the last request for new tx data
+	log            log.Logger
+	channelTimeout uint64
+
+	// All blocks since the last request for new tx data.
 	blocks []*types.Block
 	datas  []taggedData
+
+	// Pending data returned by TxData waiting on Tx Confirmed/Failed
+	// channelPending is true if there is an unconfirmed channel
+	channelPending bool
+	// id of the pending channel
+	pendingChannel derive.ChannelID
+	// list of blocks in the channel. Saved in case the channel must be rebuilt
+	pendingBlocks []*types.Block
+	// Set of unconfirmed txID -> frame data. For tx resubmission
+	pendingTransactions map[txID][]byte
+	// Set of confirmed txID -> inclusion block. For determining if the channel is timed out
+	confirmedTransactions map[txID]eth.BlockID
+}
+
+func NewChannelManager(log log.Logger, channelTimeout uint64) *channelManager {
+	return &channelManager{
+		log:                   log,
+		channelTimeout:        channelTimeout,
+		channelPending:        false,
+		pendingTransactions:   make(map[txID][]byte),
+		confirmedTransactions: make(map[txID]eth.BlockID),
+	}
 }
 
 // Clear clears the entire state of the channel manager.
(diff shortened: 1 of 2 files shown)


op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
-	if len(s.blocks) == 0 {
-		return nil, txID{}, io.EOF
+func (s *channelManager) pendingChannelIsFullySubmitted() bool {
+	if !s.channelPending {
+		return false // todo: can decide either way here. Nonsensical answer though
 	}
+	return len(s.pendingTransactions)+len(s.datas) == 0
+}
+
+func (s *channelManager) TxFailed(id txID) {
```
tynes, line 129: Perhaps the happy path should log in this function?

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
+		return s.nextTxData()
+	}
+	// If we have no saved blocks, we will not be able to create valid frames
+	if len(s.blocks) == 0 {
+		return nil, txID{}, io.EOF
+	}
+
+	// Select range of blocks
+	end := len(s.blocks)
+	if end > 100 {
```
tynes, line 205: Seems like a sane constant, at some point we will want to measure and optimize this

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
+		end = 100
+	}
+	blocks := s.blocks[:end]
+	s.blocks = s.blocks[end:]
+
+	chID, frames, leftOverBlocks, err := blocksToFrames(blocks, 120_000)
+	// If the range of blocks serialized to be too large, restore
+	// blocks that could not be included inside the channel
+	if len(leftOverBlocks) != 0 {
+		s.blocks = append(leftOverBlocks, s.blocks...)
```
tynes, line 215: This looks right, I think a test case hitting this particular branch would be nice to see

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
+	if err != nil {
+		return nil, txID{}, err
+	}
+
+	var t []taggedData
+	for i, data := range frames {
+		t = append(t, taggedData{data: data, id: txID{chID: chID, frameNumber: uint16(i)}})
+	}
+
+	// Load up pending state. Note: pending transactions is taken care of by nextTxData
```
tynes, line 226: I don't think any of these values are accessed concurrently, just a thought because a lot of state is being mutated and there is no lock

op-batcher/driver.go
```diff
@@ -234,7 +234,7 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
 	if l.lastStoredBlock == (eth.BlockID{}) {
 		l.log.Info("Starting batch-submitter work at safe-head", "safe", syncStatus.SafeL2)
 		l.lastStoredBlock = syncStatus.SafeL2.ID()
-	} else if l.lastStoredBlock.Number <= syncStatus.SafeL2.Number {
+	} else if l.lastStoredBlock.Number < syncStatus.SafeL2.Number {
```
tynes, line 237: Ok this change matches the behavior stated in the logline right below

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
-	if len(s.datas) != 0 {
-		r := s.datas[0]
-		s.datas = s.datas[1:]
-		return r.data, r.id, nil
+func (s *channelManager) pendingChannelIsTimedOut() bool {
+	if !s.channelPending {
+		return false // no channel to be timed out
+	}
+	min := uint64(math.MaxUint64)
+	max := uint64(0)
```
tynes, line 110: I wish there was a cheap way to determine this by looking at the chain, having non persistent statefulness isn't ideal

When there are no confirmed txs:
```
0 - math.MaxUint64 >= s.channelTimeout
```

This will prob always be false?

op-batcher/driver.go
trianglesphere: yea, this used to be hit on every calcL2BlockRange

op-batcher/channel_manager.go
trianglesphere: yea, I should probably short circuit if there are no confirmed transactions. I don't like taking my chances on the wrapping behavior here.

op-batcher/channel_manager.go
trianglesphere: line 33 :)

op-batcher/channel_manager.go
trianglesphere: yea, this code also worried me a bit

op-batcher/channel_manager.go
tynes: Got it

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
+	chID, frames, leftOverBlocks, err := blocksToFrames(blocks, 120_000)
+	// If the range of blocks serialized to be too large, restore
+	// blocks that could not be included inside the channel
+	if len(leftOverBlocks) != 0 {
+		s.blocks = append(leftOverBlocks, s.blocks...)
+	}
+	if err != nil {
+		return nil, txID{}, err
+	}
+
```
tynes, line 220: Perhaps sanity check on the length of frames being less than `math.MaxUint16` to prevent an overflow with the typecast, it should never happen but it would be a nice sanity check

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
+	} else {
 		return nil, txID{}, io.EOF // TODO: not enough data error instead
 	}
+}
+
+// TxData returns the next tx.data that should be submitted to L1.
+// It is very simple & currently ignores the l1Head provided (this will change).
+// It may buffer very large channels as well.
+func (s *channelManager) TxData(l1Head eth.L1BlockRef) ([]byte, txID, error) {
+	s.log.Trace("Requested tx data")
```
tynes, line 191: Would be nice to log the blockref data here in this trace

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
-	s.datas = append(s.datas, t...)
-	// Say i = 0, 1 are added to the channel, but i = 2 returns ErrTooManyRLPBytes. i remains 2 & is inclusive, so this works.
-	// Say all blocks are added, i will be len(blocks) after exiting the loop (but never inside the loop).
-	s.blocks = s.blocks[i:]
-
-	if len(s.datas) == 0 {
+// nextTxData pops off s.datas & handles updating the internal state
+func (s *channelManager) nextTxData() ([]byte, txID, error) {
+	if len(s.datas) != 0 {
+		r := s.datas[0]
```
tynes, line 178: Perhaps useful to have a trace log whenever global state is mutated

op-batcher/driver.go
```diff
@@ -271,16 +271,20 @@ func (l *BatchSubmitter) loop() {
...
-				data, _, err := l.state.TxData(eth.L1BlockRef{})
+				data, id, err := l.state.TxData(eth.L1BlockRef{})
 				if err == io.EOF {
+					l.log.Trace("no transaction data available")
 					break // local for loop
 				} else if err != nil {
 					l.log.Error("unable to get tx data", "err", err)
 					break
 				}
 				// Drop receipt + error for now
```
tynes, line 282: This comment is dead

op-batcher/driver.go
```diff
@@ -271,16 +271,20 @@ func (l *BatchSubmitter) loop() {
...
-				if _, err := l.txMgr.SendTransaction(l.ctx, data); err != nil {
+				if receipt, err := l.txMgr.SendTransaction(l.ctx, data); err != nil {
 					l.log.Error("Failed to send transaction", "err", err)
+					l.state.TxFailed(id)
+				} else {
+					l.state.TxConfirmed(id, eth.BlockID{Number: receipt.BlockNumber.Uint64(), Hash: receipt.BlockHash})
```
tynes, line 287: An info logline could be useful here

tynes (changes requested):
Will re-review after comments addressed

tynes:
Looks like not all comments have been addressed yet

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
...
+	} else {
+		s.log.Warn("marked transaction as failed despite having no record of it.")
+	}
+}
 
-	// Add all pending blocks to a channel
+// blocksToFrames turns a set of blocks into a set of frames inside a channel.
+// It will only create a single channel which contains up to `MAX_RLP_BYTES`. Any
+// blocks not added to the channel are returned. It uses the max supplied frame size.
+func blocksToFrames(blocks []*types.Block, maxFrameSize uint64) (derive.ChannelID, [][]byte, []*types.Block, error) {
```
protolambda, line 141: Why immediately produce the frames, instead of holding on to the `derive.ChannelOut`? You can produce frames with the exact size we want later that way, and pack more efficiently?

trianglesphere:
> Looks like not all comments have been addressed yet

Nope. I didn't re-request a review. The force push was a rebase.

protolambda (commented):
Changes look good, but I would prefer to defer the creation of frames until we include them into transactions, so we can better utilize the packing, instead of committing to a specific size of frame before we know the tx we can put them in.

And long-term I would like if we always pack frames from one *or two* channels per transaction: if a channel leaves data-tx space underutilized, we can start encoding the next channel in the same tx. Maybe it's too deep of a change for this PR, but we can keep the option more open to do so later? Outputting the frames early and using single channel `txID`s makes that quite hard.

trianglesphere:
> Maybe it's too deep of a change for this PR, but we can keep the option more open to do so later? Outputting the frames early and using single channel txIDs makes that quite hard

It is too large a change for this PR. It's tricky enough to handle multiple frames per channel with a single channel (you have to keep track of which block were submitted where).

In terms of the IDs, they're opaque IDs on purpose & when it becomes worth it pack multiple frames into a single L1 transaction we will be able to do that by only updating the channel manager.

op-batcher/channel_manager.go
```diff
@@ -22,10 +24,42 @@ type taggedData struct {
 	id   txID
 }
```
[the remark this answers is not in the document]
trianglesphere: Added for txID, don't really want to add for taggedData (logging bytes is pretty verbose).

op-batcher/channel_manager.go
trianglesphere: Gonna leave a TODO comment about this. It's a good idea, but has some trickiness when implementing it.

op-batcher/channel_manager.go
```diff
@@ -35,90 +69,174 @@ func (s *channelManager) Clear() {
 	s.datas = s.datas[:0]
 }
 
-// func (s *channelManager) TxConfirmed(id txID, inclusionBlock eth.BlockID) {
-// 	// todo: implement
-// }
+func (s *channelManager) TxConfirmed(id txID, inclusionBlock eth.BlockID) {
+	if _, ok := s.pendingTransactions[id]; !ok {
+		s.log.Info("marked transaction as confirmed despite having no record of it")
```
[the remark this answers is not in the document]
trianglesphere: manually did these

tynes (approved):

mslipper (approved):

mslipper:
@trianglesphere LGTM - you'll need to fix a misspelling in order to get the go tests to pass. Otherwise once @protolambda has a look I'll merge this while we update + fix hive.

trianglesphere:
I fixed the lint

tynes:
```
    driver.go:284:         ERROR[11-04|22:16:53.453] Failed to send transaction               role=batcher   err="failed to create tx: failed to get suggested gas tip cap: context canceled"
```

trianglesphere:
> ```
>     driver.go:284:         ERROR[11-04|22:16:53.453] Failed to send transaction               role=batcher   err="failed to create tx: failed to get suggested gas tip cap: context canceled"
> ```

gah, I guess the first attempt at fixing this didn't work. I force pushed something that's trying to fix it.

protolambda (approved):
LGTM. Can merge after CI passes (except hive, which I'll work on fixing as soon as the geth rebase lands on develop)

PR #3861, opened 2022-11-02
op-node: Prefer following seq drift instead of conf depth

trianglesphere:
**Description**

We found a log on a testnet where a batch was rejected because it's timestamp was greater than the origin timestamp + sequencer drift. I looked into origin selection and determined we produce empty blocks if the block time is ahead of the origin + seq drift; however if there is a next origin, it is not enough to produce an empty block, we must change the origin.

I hypothesize that this occurred because the L1 head that the node was looking at was lagging. That would cause the node to keep using an old origin even if the timestamp was past the max drift. I added a unit test which shows this case & would fail without this fix being applied.

**Tests**

Unit tested with a regression test.

**Metadata**
- Fixes ENG-2923

### DIFF
--- a/op-node/rollup/driver/origin_selector.go
+++ b/op-node/rollup/driver/origin_selector.go
@@ -47,6 +47,12 @@ func (los *L1OriginSelector) FindL1Origin(ctx context.Context, l1Head eth.L1Bloc
 		return eth.L1BlockRef{}, err
 	}
 
+	// We know there is at least one block & we have hit the sequencer drift, we need to advance.
+	if l2Head.Time+los.cfg.BlockTime > currentOrigin.Time+los.cfg.MaxSequencerDrift {
+		log.Info("Must try to advance origin") // TODO: Full log
+		return los.l1.L1BlockRefByNumber(ctx, currentOrigin.Number+1)
+	}
+
 	if currentOrigin.Number+1+los.sequencingConfDepth > l1Head.Number {
 		// TODO: we can decide to ignore confirmation depth if we would be forced
 		//  to make an empty block (only deposits) by staying on the current origin.
--- a/op-node/rollup/driver/origin_selector.go
+++ b/op-node/rollup/driver/origin_selector.go
@@ -47,13 +47,18 @@ func (los *L1OriginSelector) FindL1Origin(ctx context.Context, l1Head eth.L1Bloc
 		return eth.L1BlockRef{}, err
 	}
 
-	// We know there is at least one block & we have hit the sequencer drift, we need to advance.
-	if l2Head.Time+los.cfg.BlockTime > currentOrigin.Time+los.cfg.MaxSequencerDrift {
-		log.Info("Must try to advance origin") // TODO: Full log
-		return los.l1.L1BlockRefByNumber(ctx, currentOrigin.Number+1)
+	// If we are past the sequencer depth, we may want to advance the origin, but need to still
+	// check the time of the next origin.
+	pastSeqDrift := l2Head.Time+los.cfg.BlockTime > currentOrigin.Time+los.cfg.MaxSequencerDrift
+	if pastSeqDrift {
+		log.Info("Next L2 block time is past the sequencer drift + current origin time",
+			"current", currentOrigin, "current_time", currentOrigin.Time,
+			"l1_head", l1Head, "l1_head_time", l1Head.Time,
+			"l2_head", l2Head, "l2_head_time", l2Head.Time,
+			"depth", los.sequencingConfDepth)
 	}
 
-	if currentOrigin.Number+1+los.sequencingConfDepth > l1Head.Number {
+	if !pastSeqDrift && currentOrigin.Number+1+los.sequencingConfDepth > l1Head.Number {
 		// TODO: we can decide to ignore confirmation depth if we would be forced
 		//  to make an empty block (only deposits) by staying on the current origin.
 		log.Info("sequencing with old origin to preserve conf depth",
@@ -68,6 +73,8 @@ func (los *L1OriginSelector) FindL1Origin(ctx context.Context, l1Head eth.L1Bloc
 	// the current origin block.
 	nextOrigin, err := los.l1.L1BlockRefByNumber(ctx, currentOrigin.Number+1)
 	if err != nil {
+		// TODO: this could result in a bad origin being selected if we are past the seq
+		// drift & should instead advance to the next origin.
 		log.Error("Failed to get next origin. Falling back to current origin", "err", err)
 		return currentOrigin, nil
 	}
--- a/op-node/rollup/driver/origin_selector_test.go
+++ b/op-node/rollup/driver/origin_selector_test.go
@@ -177,3 +177,42 @@ func TestOriginSelectorRespectsMaxSeqDrift(t *testing.T) {
 	require.Nil(t, err)
 	require.Equal(t, b, next)
 }
+
+// TestOriginSelectorSeqDriftRespectsNextOriginTime
+//
+// There are 2 L1 blocks at time 20 & 100. The L2 Head is at time 27.
+// The next L2 time is 29. Even though the next L2 time is past the seq
+// drift, the origin should remain on block `a` because the next origin's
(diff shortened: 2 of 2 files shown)


trianglesphere:
@mergifyio refresh

op-node/rollup/driver/origin_selector.go
```diff
@@ -47,6 +47,12 @@ func (los *L1OriginSelector) FindL1Origin(ctx context.Context, l1Head eth.L1Bloc
 		return eth.L1BlockRef{}, err
 	}
 
+	// We know there is at least one block & we have hit the sequencer drift, we need to advance.
+	if l2Head.Time+los.cfg.BlockTime > currentOrigin.Time+los.cfg.MaxSequencerDrift {
+		log.Info("Must try to advance origin") // TODO: Full log
```
tynes, line 52: Want to knock out this todo before merge?

trianglesphere: yea, good catch

protolambda (changes requested):
What happens if the next L1 origin time is more than the sequencer time drift away from the current L1 origin? It should keep using the same L1 origin and force the blocks to be empty iirc.

But with this check, we can't produce a L2 block with the an old origin anymore, and get stuck with sequencing I think, until block derivation produces the block.

And when the next L1 origin time is within the time drift span from the current origin, we can stay on the old origin safely I believe?

trianglesphere:
> What happens if the next L1 origin time is more than the sequencer time drift away from the current L1 origin?

Then it should compare the next L2 time vs the child block of the head L2 block's L1 Origin. Sequencer drift is only about the L2 block time vs the L1 origin time of that block.
This code is a little tricky in that at the point that it does the check for the sequencer drift, it knows that there is at least one more L1 block, but that L1 block may not be known to be confirmed to the node's satisfaction. This means that the LOS will not attempt to advance to a not found block at the sequencer drift check.

Note: If we lose connection to L1 for too long (more than the sequencer drift), the sequencer may produce blocks when it should not have. (It must produce empty block for all L2 times from l1Origin.Time + seq drift to lOriginNext.Time). However, if the sequencer produced empty blocks past l1OriginNext.Time because it did not have l1OriginNext, ignore those batches.

protolambda:
@trianglesphere 

Say if we have:
```
L2 chain blocks:
A, num=200, time=10, origin=X

L1 chain blocks;
X, num=100, time=1
Y, num=101, time=50

L2 block time = 2
L2 seq time drift = 10
```

Then with this PR:
```
if l2Head.Time+los.cfg.BlockTime > currentOrigin.Time+los.cfg.MaxSequencerDrift { select next origin }

if 10 + 2 > 1 + 10 { select next origin }
```
But the next origin time is way ahead: `50 > 10 + 2`, the next L2 block cannot have this origin yet, or it breaks the time invariant that L2 time is >= L1 origin time.
And so unless L1 reorgs out the block, the L2 sequencer gets stuck because the origin it selects cannot actually be used for a new L2 block?

norswap:
Comments & clarifications. I'd need to do a deeper dive on the code to see what happens there, but I already have a lot to say based on the discussion alone:

> I looked into origin selection and determined we produce empty blocks if the block time is ahead of the origin + seq drift; however if there is a next origin, it is not enough to produce an empty block, we must change the origin.

Confirming this, but reframing:

If we're ahead of origin + seq drift, the only allowed move is to create a new block with a new origin (since that moves the L2 time 2s but the moves L1 time at least 12s, hence reducing the drift).

Though not strictly necessary (in terms of properties), it's probably a good move to indeed make these blocks empty — the spec mandates it as we'll see.

> What happens if the next L1 origin time is more than the sequencer time drift away from the current L1 origin? It should keep using the same L1 origin and force the blocks to be empty iirc.

Can this happen — that so many slots fail that two successive L1 blocks are very far away from one another?

But anyway, if I understand correctly, the issue here would be that that new origin's timestamp is actually now *ahead* of the L2 head, meaning we do need to output more empty blocks to catch up. This is definitely an edge case to consider.

In general, the algorithm should then be:

- if `nextOrigin.timestamp <= L2head.timestamp` → create L2 block with next origin
- otherwise, create empty blocks until `nextOrigin.timestamp <= L2head.timestamp` then create an L2 block with the next origin

> And when the next L1 origin time is within the time drift span from the current origin, we can stay on the old origin safely I believe?

We can, but I think the right move is to adopt new origins as fast as possible to reduce the drift.

--------------------------------------------------------------------------------

Btw, these conversations should be related back to the spec. There are two relevant snippets:

[First one][spec1]:

- `min_l2_timestamp <= block.timestamp < max_l2_timestamp`, where
  - all these values are denominated in seconds
  - `min_l2_timestamp = prev_l2_timestamp + l2_block_time`
    - `prev_l2_timestamp` is the timestamp of the last L2 block of the previous epoch
    - `l2_block_time` is a configurable parameter of the time between L2 blocks (on Optimism, 2s)
  - `max_l2_timestamp = max(l1_timestamp + max_sequencer_drift, min_l2_timestamp + l2_block_time)`
    - `l1_timestamp` is the timestamp of the L1 block associated with the L2 block's epoch
    - `max_sequencer_drift` is the most a sequencer is allowed to get ahead of L1

[Second one][spec2]:

> - `batch.timestamp > batch_origin.time + max_sequencer_drift` -> `drop`: i.e.
>   a batch that does not adopt the next L1 within time will be dropped, in favor
>   of an empty batch that can advance the L1 origin.
>
>  ... // elided som more rules
>
> If no batch can be accept-ed, and the stage has completed buffering of all batches that can fully be read from the L1 block at height epoch.number + sequence_window_size, and the next_epoch is available, then an empty batch can be derived with the following properties:
>
> - parent_hash = safe_l2_head.hash
> - timestamp = next_timestamp
> - transactions is empty, i.e. no sequencer transactions. Deposited transactions may be added in the next stage.
> - If next_timestamp < next_epoch.time: the current L1 origin is repeated, to preserve the L2 time invariant.
>   - epoch_num = epoch.number
>   - epoch_hash = epoch.hash
> - Otherwise,
>    - epoch_num = next_epoch.number
>    - epoch_hash = next_epoch.hash

where `batch_origin == batch.origin.number`, with caveat if the origin isn't
known, which I guess we should think about in the current case.

[spec1]: https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#overview
[spec2]: https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#batch-buffering

The first snippet seems correct: it allows us to go past the drift if we have to — though that should only be allowed if we reduce the drift.

I guess that's something it doesn't say, and also "`max_sequencer_drift` is the most a sequencer is allowed to get ahead of L1" is de facto false, so there's room for improvement there!

The second snippet seems correct as well! If we're past the drift, generate an empty block, and since we have a drift, we need to adopt the next origin (`next_epoch`).

op-node/rollup/driver/origin_selector.go
```diff
@@ -62,6 +73,8 @@ func (los *L1OriginSelector) FindL1Origin(ctx context.Context, l1Head eth.L1Bloc
 	// the current origin block.
 	nextOrigin, err := los.l1.L1BlockRefByNumber(ctx, currentOrigin.Number+1)
 	if err != nil {
+		// TODO: this could result in a bad origin being selected if we are past the seq
```
tynes, line 76: What is the deal with this todo?

trianglesphere: This would occur if `pastSeqDrift` is true & we fail to get the potential next L1 origin. In the case that the next L1 origin block time would have been valid (i.e. the next L2 time >= nextOrigin.Time) we would incorrectly return the currentOrigin instead of the next origin. This is a bit tricky because we don't know if the nextOrigin is valid until after we get it, so only part of the time the currentOrigin is invalid & part of the time it is valid.

protolambda (approved):
Change LGTM, and fixes part of the problem, but I think the discussion here + the new added todo highlights something that's still possible to go wrong on a testnet, and we should fix too. Let's discuss it on Monday.

mslipper (approved):

PR #3862, opened 2022-11-03
Update forge-std and use bound instead of vm.assume

maurelian:
**Description**

1. Updates `forge-std` to the 1.0.0 release https://github.com/foundry-rs/forge-std/releases/tag/v1.0.0

2. Replace `vm.assume` with `bound` where possible (mostly for `uint` type arguments). The benefit of this is that rather than discard test cases which don't meet the condition in `vm.assume(condition)`, values will now be shifted to fit within 
the min/max specified in `bound(value, min, max)`. ([see impl](https://github.com/foundry-rs/forge-std/blob/17656a2fa5453f495d8c1302a0cedded912457eb/src/StdUtils.sol#L10))


tynes:
```
Compiler run failed
error[6275]: ParserError: Source "node_modules/forge-std/lib/ds-test/src/test.sol" not found: File not found. Searched the following locations: "/root/project/packages/contracts-bedrock", "/root/project/packages/contracts-bedrock/contracts".
 --> node_modules/forge-std/src/StdAssertions.sol:4:1:
  |
4 | import "../lib/ds-test/src/test.sol";
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Looks like its not packaged with `ds-test`

packages/contracts-bedrock/contracts/test/Hashing.t.sol
```diff
@@ -21,19 +21,20 @@ contract Hashing_Test is CommonTest {
...
         address _sender,
         address _target,
         uint256 _value,
         uint256 _gasLimit,
         bytes memory _data
     ) external {
-        // Discard any fuzz tests with an invalid version
-        (, uint16 version) = Encoding.decodeVersionedNonce(_nonce);
-        vm.assume(version < 2);
+        // Ensure the version is valid
```
smartcontracts, line 32: @tynes can you check if this looks good?

smartcontracts (commented):
Will approve except for Mark's review on the one line that I'm unsure about

packages/contracts-bedrock/contracts/test/Hashing.t.sol
maurelian: Yeah, it's worth expanding on what it does because it's cool. 

The code on the left generates random nonces, and then throws out anything where the version number is not either 0 or 1. So if the inputs are naively random, then theoretically we'd be throwing out all but 2 out of every 65536 inputs. 

The code on the right will take the randomly generated `uint16 version`, and modulo it into the range `[0,1]`.

packages/contracts-bedrock/contracts/test/Hashing.t.sol
```diff
@@ -21,19 +21,20 @@ contract Hashing_Test is CommonTest {
...
+        uint16 version = uint16(bound(uint256(_version), 0, 1));
```
tynes, line 33: I'd like to double check here that the final arg is inclusive

maurelian: Yes:
https://github.com/foundry-rs/forge-std/blob/33d4895fb629aabf12e0f60740367bd03e702bd1/src/StdUtils.sol#L15

I also [instrumented the code](https://github.com/maurelian/optimism-1/compare/develop...maurelian:optimism-1:vm-bound-counts?expand=1#diff-2e77d00d31481bd3b8012d5c4c74bb334d95ec7c74485825aa437af34357c53cR35-R38) to see the version values that were used, you [can see](https://github.com/maurelian/optimism-1/blob/b5eab76b34cefe19ba39bd2b125d1d057a7b4bb7/packages/contracts-bedrock/runs.txt) that it includes both 0s and 1s.

tynes (approved):

smartcontracts (approved):

PR #3870, opened 2022-11-03
optimism: update op-geth to the predeploy commit

tynes:
**Description**

Includes https://github.com/ethereum-optimism/op-geth/pull/23

Pulling these changes into the monorepo so that things can be integrated with `op-e2e`

Will follow up with PRs that clean up `op-e2e` so that they don't get too large

### DIFF
--- a/go.work
+++ b/go.work
@@ -20,7 +20,7 @@ use (
 	./teleportr
 )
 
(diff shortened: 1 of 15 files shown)


protolambda (approved):
LGTM.

Note: This should be closed if we first merge the sys config stack with op-geth update, to not accidentally revert the op-geth version back.

trianglesphere (approved):

PR #3875, opened 2022-11-04
op-node,op-e2e: update libp2p libs, update libp2p gossip settings

protolambda:
**Description**

PR changes:
- update `github.com/libp2p/go-libp2p`
- remove `github.com/libp2p/go-libp2p-core` since it was deprecated and moved into `github.com/libp2p/go-libp2p/core`
- remove `https://github.com/libp2p/go-libp2p-peerstore` since it was deprecated and moved into ` `github.com/libp2p/go-libp2p/peerstore`
- update gossip params to deduplicate a constant, and match the L1 value
- update gossip params to use the `WithSeenMessagesTTL` I upstreamed a few months ago, to not have to configure it with a global anymore
- update the libp2p crypto dependencies
- go mod tidy to fix the dependencies

**Metadata**

Fix ENG-2708


trianglesphere (approved):
lgtm. Thanks for taking care of all of the updates.

tynes:
Looks good to me, will approve when the deps are fixed

protolambda:
go mod tidy is reverting a crypto lib update, since geth (and maybe the mnemonics lib too?) use the older version and they get priority somehow, even though it compiles fine with the new version. We're starting to update both of those today though, so then I'll update this PR with a rebase + new go mod tidy to get it sorted out.

tynes:
This should help: https://github.com/ethereum-optimism/optimism/pull/3880

trianglesphere:
Go prefers using the oldest rather than newest version: https://go.dev/ref/mod#minimal-version-selection

protolambda:
Will resolve merge conflicts once the op-geth PR lands (#3892 )

trianglesphere:
> Will resolve merge conflicts once the op-geth PR lands (#3892 )

#3892 has just landed @protolambda

tynes:
Hopefully the build issues are fixed now, feels good to be on modern geth as a dep

protolambda:
Rebased on develop and cleaned up the dependency issues. We can remove the replace statements in `op-e2e/go.mod` only after we make releases of our own modules, to get rid of the stale dependency of the old `github.com/miguelmota/go-ethereum-hdwallet` which still uses an older `github.com/btcsuite/btcd/btcec` version.

op-node/p2p/discovery.go
```diff
@@ -11,22 +11,32 @@ import (
...
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
 )
 
+// force to use the new chainhash module, and not the legacy chainhash package btcd module
+const _ = chainhash.HashSize
+
+// force to use the btcd module, while the keycard dependency in geth still depends on it,
```
tynes, line 36: https://github.com/status-im/keycard-go/issues/14

tynes (approved):

mslipper (approved):

PR #3876, opened 2022-11-04
batch-submitter: update go-ethereum to v1.10.26

tynes:
**Description**

Updates the batch submitter's geth dep to v.10.26 which includes a fix to the batch RPC call that could cause a halt


trianglesphere:
@tynes this is switching the dependency from `op-geth` to `go-ethereum` b/c we when we do the replace statement from `go-ethereum` to `op-geth`, it is done per version. It turns out the batcher has a dependency on `op-geth` rather than just upstream geth so I think we should wait on updating `op-geth` & updating the whole monorepo to use that version.

PR #3880, opened 2022-11-04
deps: replace deprecated package with updated package

tynes:
**Description**

https://github.com/miguelmota/go-ethereum-hdwallet is no longer maintained and is using older versions of dependencies. There are dependencies clashing when trying to update code elsewhere in the codebase.

https://github.com/ethereum-optimism/go-ethereum-hdwallet is a fork with updated deps such that we can more easily update other deps in the codebase and drop old deps.

Fixes ENG-2980


mslipper:
Can you move the fork into `ethereum-optimism`?

tynes:
> Can you move the fork into `ethereum-optimism`?

Updated

trianglesphere (approved):
lgtm. Let's just remember to remove the replace statements in op-e2e/go.mod when we're done.

tynes:
@protolambda I've rebased this PR after a recent `op-node` update

protolambda (approved):
Reminder to self: Need to remove the replace statements after the op-geth PR lands.

PR #3892, opened 2022-11-05
bedrock: update op-geth dependency

protolambda:
**Description**

Update op-geth dependency, rebased to latest `master` branch. See `optimism-history` branch on op-geth repo. The previous `optimism-history` can be found as `optimism-history-nov-4` branch.

Note: the upstream miner code changed, the background job logic for updating a block specifically. Now that the first tx pool work is fully in the background the sequencer code path needs more room to breathe, or would retrieve empty blocks consistently if it asks for the block building result instantly.

This is something we can optimize later: more time is good, to build better blocks. But less time is better when we are catching up to the latest wallclock time (i.e. when we don't have as much time to wait for production).
For now I increased a few e2e tx inclusion timeouts and added some unconditional breathing time.

Other than that the `trie` interface changed slightly yet again, and we now use eip1559 params via the chain config.

-----

Note: endpoint monitor and indexer were also updated from geth v1.10.23 to v1.10.26 to keep them in sync with the upstream base we use for bedrock.


tynes:
Do you mind adding a changeset for the `batch-submitter` and also the monitor if its hooked up to the old docker publish flow?

tynes:
I think we may still have to do releases, but maybe you can delete these lines?

```
// temporary, until we do another set of releases
replace (
	github.com/ethereum-optimism/optimism/op-batcher v0.9.0 => ../op-batcher
	github.com/ethereum-optimism/optimism/op-proposer v0.9.0 => ../op-proposer
)
```

protolambda:
@tynes need releases for removing those replace statements first, I still get:
```
go: finding module for package github.com/btcsuite/btcd/btcec
github.com/ethereum-optimism/optimism/op-e2e imports
        github.com/ethereum-optimism/optimism/op-batcher imports
        github.com/miguelmota/go-ethereum-hdwallet imports
        github.com/btcsuite/btcutil/hdkeychain imports
        github.com/btcsuite/btcd/btcec: module github.com/btcsuite/btcd@latest found (v0.23.3), but does not contain package github.com/btcsuite/btcd/btcec
```
pointing to the old library, instead of the new one you introduced in #3880

mslipper (approved):

tynes (approved):

mslipper:
The devnet issue is a real one - the withdrawal transaction reverts.

PR #3897, opened 2022-11-07
ctb: Add system config deployment

mslipper:
Adds a Hardhat deployment for the system config. I followed the patterns I saw elsewhere in the migration to make this work, please let me know if I missed anything.

PR #3898, opened 2022-11-07
op-chain-ops: Updates for the migration

mslipper:
Updates `op-chain-ops` a bit to make it work with the migration. The comments below outline exactly what was changed.


mslipper:
* develop
  * **PR #3897** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3897" target="_blank">Graphite</a> 
    * **PR #3898** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3898" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/3898?utm_source=stack-comment).

op-chain-ops/crossdomain/migrate.go
```diff
@@ -27,7 +27,12 @@ func MigrateWithdrawals(withdrawals []*LegacyWithdrawal, db vm.StateDB, l1CrossD
 
 		legacyValue := db.GetState(predeploys.LegacyMessagePasserAddr, legacySlot)
 		if legacyValue != abiTrue {
-			return fmt.Errorf("%w: %s", errLegacyStorageSlotNotFound, legacySlot)
+			// TODO: Re-enable this once we have the exact data we need on mainnet.
```
mslipper, line 30: I deliberately commented this out so that the migration can actually run using our test data.

op-chain-ops/ether/cli.go
```diff
@@ -1,169 +1,13 @@
...
-//
-//   - dataDir:        A Geth data dir.
-//   - outDir:         A directory to output the migrated database to.
-//   - genesis:        The new chain's genesis configuration.
-//   - addrLists:      A list of address list file paths. These address lists are used to populate
-//     balances from previous regenesis events.
-//   - allowanceLists: A list of allowance list file paths. These allowance lists are used
-//     to calculate allowance storage slots from previous regenesis events.
-//   - chainID:        The chain ID of the chain being migrated.
-func Migrate(dataDir, outDir string, genesis *core.Genesis, addrLists, allowanceLists []string, chainID, levelDBCacheSize, levelDBHandles int) error {
```
mslipper, line 67: This is all dead code from the old CLI.

op-chain-ops/ether/migrate.go
```diff
@@ -34,10 +54,19 @@ func MigrateLegacyETH(db ethdb.Database, addresses []common.Address, allowances
...
-		// TODO: double check ordering here
 		storageSlotsToMigrate[CalcAllowanceStorageKey(allowance.From, allowance.To)] = 2
 	}
 
+	if chainID == 1 {
+		// Some folks sent money to this address ages ago, permanently locking it
+		// there. This contract never transacted on a modern network, so hardcode
+		// this to ensure that all storage slots are accounted for.
+		// This address was once the OVM_SequencerEntrypoint contract.
+		seqEntryAddr := common.HexToAddress("0x4200000000000000000000000000000000000005")
```
mslipper, line 65: This was our last missing state slot.

op-chain-ops/ether/migrate.go
```diff
@@ -194,5 +222,21 @@ func MigrateLegacyETH(db ethdb.Database, addresses []common.Address, allowances
...
+	}
+
+	log.Info("committing state DB")
+	newRoot, err := stateDB.Commit(true)
+	if err != nil {
+		return common.Hash{}, err
+	}
+
+	log.Info("committing trie DB")
+	if err := stateDB.Database().TrieDB().Commit(newRoot, true, nil); err != nil {
```
mslipper, line 237: This was missing previously, which meant that no state data was being committed to disk.

op-chain-ops/ether/params.go
```diff
@@ -20,13 +15,8 @@ type Params struct {
 
 var ParamsByChainID = map[int]*Params{
 	1: {
-		// These storage keys were unaccounted for in the genesis state of regenesis 5.
-		map[common.Hash]bool{
-			common.HexToHash("0x8632b3478ce27e6c2251f16f71bf134373ff9d23cff5b8d5f95475fa6e52fe22"): true,
```
mslipper, line 25: All these keys are accounted for now.

op-chain-ops/genesis/config.go
```diff
@@ -8,76 +8,110 @@ import (
...
-
-	GasPriceOracleOwner    common.Address `json:"gasPriceOracleOwner"`
-	GasPriceOracleOverhead uint64         `json:"gasPriceOracleOverhead"`
-	GasPriceOracleScalar   uint64         `json:"gasPriceOracleScalar"`
-
-	DeploymentWaitConfirmations int `json:"deploymentWaitConfirmations"`
-
-	EIP1559Elasticity  uint64 `json:"eip1559Elasticity"`
-	EIP1559Denominator uint64 `json:"eip1559Denominator"`
+	JSONDeployConfig
```
mslipper, line 23: I needed the ability to serialize the `rpc.BlockNumberOrHash` struct to JSON. It's not supported in the Geth library - it outputs an object that then cannot be read. We've run into this issue before and never solved it, so I tried to solve it here without introducing a huge struct with lots of duplicate fields. Open to other solutions, but we need a way to serialize this weird `BlockNumberOrHash` thing to JSON for the migration tests.

op-chain-ops/genesis/db_migration.go
```diff
@@ -19,96 +22,142 @@ import (
...
 		ParentHash:  header.Hash(),
 		UncleHash:   types.EmptyUncleHash,
 		Coinbase:    config.L2GenesisBlockCoinbase,
-		Root:        root,
+		Root:        newRoot,
 		TxHash:      types.EmptyRootHash,
 		ReceiptHash: types.EmptyRootHash,
 		Bloom:       types.Bloom{},
-		Difficulty:  (*big.Int)(config.L2GenesisBlockDifficulty),
+		Difficulty:  common.Big0,
```
mslipper, line 101: This must be zero, otherwise the opnode breaks.

op-chain-ops/genesis/db_migration.go
```diff
@@ -19,96 +22,142 @@ import (
...
 		GasLimit:    (uint64)(config.L2GenesisBlockGasLimit),
-		GasUsed:     (uint64)(config.L2GenesisBlockGasUsed),
+		GasUsed:     0,
 		Time:        uint64(config.L2OutputOracleStartingTimestamp),
-		Extra:       config.L2GenesisBlockExtraData,
-		MixDigest:   config.L2GenesisBlockMixHash,
-		Nonce:       types.EncodeNonce((uint64)(config.L1GenesisBlockNonce)),
+		Extra:       []byte("BEDROCK"),
+		MixDigest:   common.Hash{},
+		Nonce:       types.BlockNonce{},
```
mslipper, line 108: This must be empty otherwise the opnode fails block validation.

op-chain-ops/genesis/db_migration.go
```diff
@@ -19,96 +22,142 @@ import (
...
+	rawdb.WriteHeadBlockHash(ldb, bedrockBlock.Hash())
+	rawdb.WriteHeadFastBlockHash(ldb, bedrockBlock.Hash())
+	rawdb.WriteHeadHeaderHash(ldb, bedrockBlock.Hash())
+
+	// Make the first Bedrock block a finalized block.
+	rawdb.WriteFinalizedBlockHash(ldb, bedrockBlock.Hash())
+
+	// We need to pull the chain config out of the DB, and update
+	// it so that the latest hardforks are enabled.
+	genesisHash := rawdb.ReadCanonicalHash(ldb, 0)
```
mslipper, line 136: This is necessary, otherwise Geth doesn't activate 1559 and other hardforks.

op-chain-ops/genesis/db_migration.go
```diff
@@ -147,7 +196,7 @@ func CheckWithdrawals(db vm.StateDB, withdrawals []*crossdomain.LegacyWithdrawal
 	for slot := range knownSlots {
 		_, ok := slots[slot]
 		if !ok {
-			return fmt.Errorf("Unknown input message: %s", slot)
+			//return nil, fmt.Errorf("Unknown input message: %s", slot)
```
mslipper, line 199: This was deliberately commented out for similar reasons re: our available data as stated above.

op-chain-ops/genesis/migration/legacy.go
```diff
@@ -0,0 +1,95 @@
...
+		return err
+	}
+	r.Logs = make([]*types.Log, len(stored.Logs))
+	for i, log := range stored.Logs {
+		r.Logs[i] = (*types.Log)(log)
+	}
+	return nil
+}
+
+func ReadLegacyReceipts(db ethdb.Reader, hash common.Hash, number uint64) ([]*LegacyReceipt, error) {
```
mslipper, line 83: This gets around the fact that legacy receipts are not deserializable in `op-geth` right now. We can remove all this code once Mark gets around to adding legacy receipt support to `op-geth`.

op-chain-ops/genesis/migration_action/action.go
```diff
@@ -0,0 +1,110 @@
...
+	EVMMessagesPath       string
+	Network               string
+	HardhatDeployments    []string
+	L1URL                 string
+	StartingL1BlockNumber uint64
+	L2DBPath              string
+	DryRun                bool
+}
+
+func Migrate(cfg *Config) (*genesis.MigrationResult, error) {
```
mslipper, line 30: This is identical to what we had in CLI, but extracted into a standalone method so that the migration can call it directly.

op-chain-ops/genesis/config.go
tynes: Perhaps we can open a PR to geth that makes that type json serializable

op-chain-ops/genesis/db_migration.go
tynes: Both the nonce and mix digest?

op-chain-ops/genesis/config.go
tynes: It looks like there is test coverage of this here: https://github.com/ethereum/go-ethereum/blob/d629e0204799f2f7109fd52aa245e852c2e4f921/rpc/types_test.go#L127

Perhaps we are on an older version of geth here?

tynes: This isn't that bad of a diff to be honest, i think its ok leaving as is

op-chain-ops/genesis/migration_action/action.go
```diff
@@ -0,0 +1,110 @@
...
+	l1CrossDomainMessengerProxyDeployment, err := hh.GetDeployment("Proxy__OVM_L1CrossDomainMessenger")
+	if err != nil {
+		return nil, err
+	}
+	l1ERC721BridgeProxyDeployment, err := hh.GetDeployment("L1ERC721BridgeProxy")
+	if err != nil {
+		return nil, err
+	}
+
+	l2Addrs := genesis.L2Addresses{
```
tynes, line 103: In rebasing this, we can get rid of `L2Addresses` and reading from `hh`, instead we can pass `hh` to `config.GetDeploymentAddresses` and it will populate it correctly

op-chain-ops/genesis/config.go
mslipper: I fixed it. Will update this in a moment.

op-chain-ops/crossdomain/migrate.go
mslipper: Let's punt this until we completely wrap the migration script up?

op-chain-ops/genesis/db_migration.go
tynes: Longer term we will want to remove anything from the config that is not required

tynes (approved):

trianglesphere (approved):
lgtm

op-chain-ops/genesis/db_migration.go
protolambda: zero block nonce is the right thing yes, we don't want any PoW legacy to be non-zero. The uncles root should be set to the empty-uncles hash. See https://eips.ethereum.org/EIPS/eip-3675 for more details.

protolambda: And extra-data should be no longer than 32 bytes, but may be shorter. The `BEDROCK` value looks good to me.

op-chain-ops/ether/params.go
```diff
@@ -20,13 +15,8 @@ type Params struct {
...
-			common.HexToHash("0x47c25b07402d92e0d7f0cd9e347329fa0d86d16717cf933f836732313929fc1f"): true,
-			common.HexToHash("0x2acc0ec5cc86ffda9ceba005a317bcf0e86863e11be3981e923d5b103990055d"): true,
-		},
-		// Regenesis 4 contained a supply bug.
-		new(big.Int).SetUint64(1637102600003999992),
+		// Regenesis 4 contained a supply bug, and the Saurik bug likely
```
smartcontracts, line 18: Worth expanding on this slightly. Something along the lines of "Regenesis 4 (Nov 11 2021) contained a supply bug such that the total OVM ETH supply was 1.637... ETH greater than the sum balance of every account migrated during the regenesis. A further 0.0012 ETH was incorrectly not removed from the total supply by accidental invocations of the Saurik bug (link)."

PR #3901, opened 2022-11-07
feat(ctb): SystemConfig comments

smartcontracts:
**Description**
Adds comments to SystemConfig functions and variables.


protolambda (approved):
LGTM, thank you for improving this, sorry about not commenting the contract earlier.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -25,14 +34,42 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
-        GAS_CONFIG,
-        GAS_LIMIT
-    }
+    /**
+     * @notice Dynamic L2 gas overhead.
+     */
+    uint256 public scalar;
+
+    /**
+     * @notice Identifier for the batcher. Currently a padded address.
```
tynes, line 48: Perhaps the docstring could mention the versioned hash

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -25,14 +34,42 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+    /**
+     * @notice Emitted when configuration is updated
+     *
+     * @param version    SystemConfig version.
+     * @param updateType Type of update.
+     * @param data       Encoded update data.
+     */
+    event ConfigUpdate(uint256 indexed version, UpdateType indexed updateType, bytes data);
+
+    /**
```
tynes, line 66: This is missing the semver tag

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -44,7 +81,13 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
-     * @notice Initializer;
+     * @notice Initializer.
+     *
+     * @param _owner       Initial owner of the contract.
+     * @param _overhead    Initial overhead value.
+     * @param _scalar      Initial scalar value.
+     * @param _batcherHash Initial batcher hash.
+     * @param _gasLimit    Initial gas limit.
      */
     function initialize(
```
tynes, line 92: @smartcontracts What do you think about calling the methods instead of setting the values directly?

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -77,6 +132,11 @@ contract SystemConfig is OwnableUpgradeable, Semver {
         emit ConfigUpdate(VERSION, UpdateType.GAS_CONFIG, data);
     }
 
+    /**
+     * @notice Updates the L2 gas limit.
+     *
+     * @param _gasLimit New gas limit.
```
tynes, line 138: Thoughts on having a max gas limit set at deploy time and requiring that the new gas limit is less than or equal to the max gas limit? Would prevent a runaway sequencer

tynes:
Generally looks good to me, could use a little more polish with some of the docstrings, but ok with merging as is. Left a few questions that I have been thinking about regarding tweaks to this contract

tynes (approved):

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
smartcontracts: I think that makes sense but let's not change that here

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
smartcontracts: Eh, I think that's complex. Multisig has other ways it could mess with the system.

smartcontracts:
@mergifyio requeue

smartcontracts:
@mergifyio requeue

PR #3908, opened 2022-11-07
op-node: set initial system config gaslimit, check initial values

protolambda:
**Description**

Some networks are not properly getting initialized. The contracts default to 15M, but the rollup config didn't, so I added a hotfix for filling the 15M.

The rollup config now checks the genesis system config values, and we check if the rollup config we output is valid, before writing the configs to the final genesis configuration output.


trianglesphere:
@protolambda it seems we missed setting the initial gas limit in the op-e2e tests as well

protolambda:
@trianglesphere looks like the system tests actually rely on overhead and scalar being 0 as default in most cases, before the gaslimit changes even, to not apply L1 fees to the transactions. I'll update it to use some sane defaults there, using 0 is oversimplified testing IMO.

tynes (approved):

protolambda:
Fixed the op-e2e l1 fee usage. This should also fix the indexer tests. Now the `== 0` `Check` will pass there, and should only catch real issues. I updated the op-node genesis command to init with gas + check the rollup config. CI should point out any issues if I missed anything.

trianglesphere (approved):
lgtm. Might need an update b/c the devnet is acting weird

mslipper (approved):

trianglesphere:
I tried the devnet deposit-er20 locally and it was working. I'm not sure why it failed in CI

tynes:
> I tried the devnet deposit-er20 locally and it was working. I'm not sure why it failed in CI

This has been happening for awhile now, unclear why it fails in CI but not locally.

PR #3915, opened 2022-11-07
op-e2e: remove temporary go.mod workaround

protolambda:
We no longer need these `replace` statements now that the dependencies can be properly resolved with the new go module releases.


trianglesphere (approved):

tynes (approved):
I see the bedrock-go tests passing

PR #3916, opened 2022-11-08
op-bindings: Canonicalize bindings

mslipper:
Adds a function to canonicalize AST IDs in storage layouts so that they don't cause unnecessary conflicts/diffs. The implementation is not particularly efficient, but is plenty fast enough for our purposes.

It works in two passes:

1. First, it finds all AST IDs in storage and types, and builds a map to replace them in the second pass.
2. The second pass performs the replacement.

The inefficiency comes from `replaceType`, which performs a linear search of all replacements when performing substring matches of composite types. In practice this doesn't matter - we can still generate bindings for all contracts in our system in under 30 seconds.


trianglesphere (approved):

mslipper:
@mergifyio refresh

PR #3920, opened 2022-11-08
op-e2e: Use head rather than pending block for gas estimation

trianglesphere:
**Description**

This should hopefully avoid issues with not being able to find the pending block.


tynes (approved):

tynes:
Seems like a sane change to me

protolambda (approved):

PR #3921, opened 2022-11-08
fix: more config cleanup

tynes:
**Description**

Now all of the config lives in the `DeployConfig`
and there is a method to read dynamic config values
from disk given a hh or the dev addresses can be
added to the config with `GetDeveloperDeployedAddresses`.

This is useful to deduplicate code that has been changing
a lot, resulting in code changes in many places to get
tests passing again.

This will break hive when we do another release of the go
packages.

Fixes ENG-2993

### DIFF
--- a/op-chain-ops/genesis/config.go
+++ b/op-chain-ops/genesis/config.go
@@ -249,8 +249,8 @@ func (d *DeployConfig) GetDeployedAddresses(hh *hardhat.Hardhat) error {
 	return nil
 }
 
-// GetDeveloperDeployedAddresses will set the dev addresses on the DeployConfig
-func (d *DeployConfig) GetDeveloperDeployedAddresses() error {
+// InitDeveloperDeployedAddresses will set the dev addresses on the DeployConfig
+func (d *DeployConfig) InitDeveloperDeployedAddresses() error {
 	d.L1StandardBridgeProxy = predeploys.DevL1StandardBridgeAddr
 	d.L1CrossDomainMessengerProxy = predeploys.DevL1CrossDomainMessengerAddr
 	d.L1ERC721BridgeProxy = predeploys.DevL1ERC721BridgeAddr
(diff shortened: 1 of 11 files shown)


op-chain-ops/genesis/config.go
```diff
@@ -249,6 +249,16 @@ func (d *DeployConfig) GetDeployedAddresses(hh *hardhat.Hardhat) error {
 	return nil
 }
 
+// GetDeveloperDeployedAddresses will set the dev addresses on the DeployConfig
+func (d *DeployConfig) GetDeveloperDeployedAddresses() error {
```
protolambda, line 253: Maybe call it `InitDeveloperDeployedAddresses` instead? It's not quite a regular getter method

op-node/cmd/genesis/cmd.go
```diff
@@ -175,13 +175,14 @@ var Subcommands = cli.Commands{
...
-	portalAddr common.Address,
-	sysConfigAddr common.Address,
-) *rollup.Config {
+func makeRollupConfig(config *genesis.DeployConfig, l1StartBlock *types.Block, l2Genesis *core.Genesis) (*rollup.Config, error) {
+	if config.OptimismPortalProxy == (common.Address{}) {
+		return nil, errors.New("OptimismPortalProxy cannot be address(0)")
+	}
+	if config.SystemConfigProxy == (common.Address{}) {
+		return nil, errors.New("SystemConfigProxy cannot be address(0)")
+	}
```
protolambda, line 184: shouldn't this be part of the deploy config `Check` function already?

protolambda (approved):
LGTM, but I have some minor nits on duplicate checks and naming. Thank you for hardening the deploy config system.

op-node/cmd/genesis/cmd.go
tynes: Yes it is, I could delete these checks but I wanted to be very sure that we don't build the rollup config incorrectly, ie a developer forgets to use a config file that includes these or they do not call the method to get the dev addresses or the addresses from disk using hardhat

op-chain-ops/genesis/config.go
tynes: I agree that name is better, I used `Get` so it would be similar to the other method for getting deployed addresses, will update

tynes: I've updated the name

mslipper (approved):

PR #3927, opened 2022-11-08
op-chain-ops: add logging to db migration

tynes:
**Description**

This commit adds additional logging to the database migration codepath. This is helpful for understanding what is happening in the code, to understand how much of the migration has already happened.

It may also help catch programmer/configuration errors, although we should have testing to catch that and not rely on the programmer watching the logs to catch issues.

### DIFF
--- a/op-chain-ops/crossdomain/migrate.go
+++ b/op-chain-ops/crossdomain/migrate.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/log"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


protolambda (approved):
Changes LGTM, but it would be nice to not use the global logger as much. In op-e2e testing it has been very useful to decorate more localized loggers with different names like verifier/sequencer etc. when we have different instances running. And we can direct the log output with a test-logger, instead of having the unorganized std-out throughout parallel tests.

tynes:
> Changes LGTM, but it would be nice to not use the global logger as much. In op-e2e testing it has been very useful to decorate more localized loggers with different names like verifier/sequencer etc. when we have different instances running. And we can direct the log output with a test-logger, instead of having the unorganized std-out throughout parallel tests.

This is a good point, I didn't want to change the interfaces in a bunch of places to pass through a logger. Would require a bit of a refactor to achieve what you are saying

trianglesphere (approved):

mslipper:
@mergifyio requeue

PR #3929, opened 2022-11-08
feat(ctb): assert dictator config

smartcontracts:
**Description**
Asserts `Dictator` configuration after deployment.


tynes (approved):
Generally looks good to me, could modularize the creation of the config object into a function and reuse it between the two and then have a second argument determine the differences in the config creation

maurelian (approved):

PR #3931, opened 2022-11-08
ci: Split go-bedrock tests & use circle CI parallelism for op-e2e

trianglesphere:
**Description**

This first splits up the go-bedrock job into smaller more discrete jobs to ensure that we are using the same linters & test runner options for all packages. This also improves performance by parallelizing trivially paralleliziable jobs.

The second big change is to use `circlci tests split` to enable parallelism in CI instead of using `t.Parallel`. It is configured with an env var so tests will remain fast with local development, but we cannot use `t.Parallel` in CI because `gotestsum` currently does not do a good job of reporting the specific test that failed (because of the interleaved output) and instead reports multiple tests as failing when there is only one that failed. This improves the experience of using CI and debugging failed tests.

The last change is to add and option to disable colored logs in the testlogger. Somewhere in the reporting pipeline the color escape codes are getting mangled & the logs are significantly harder to read because of the colors.

To retain compatibility with github required checks, all of the new go jobs feed into `bedrock-go-tests` which will simply return true if all of it's required jobs succeed.

**Metadata**
- Fixes ENG-3007
- Fixes ENG-2999

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -262,101 +262,6 @@ jobs:
           name: Upload coverage
           command: codecov --verbose --clean --flags <<parameters.coverage_flag>>
 
-  bedrock-go-tests:
-    docker:
-      - image: ethereumoptimism/ci-builder:latest
-    resource_class: xlarge
-    steps:
-      - checkout
-      - run:
-          name: Check if we should run
-          command: |
-            shopt -s inherit_errexit
-            CHANGED=$(check-changed "op-(batcher|bindings|e2e|node|proposer|chain-ops)" || echo "TRUE")
-            if [[ "$CHANGED" = "FALSE" ]]; then
-              circleci step halt
-            fi
-      - run:
-          name: lint op-bindings
-          command: |
-            golangci-lint run -E goimports,sqlclosecheck,bodyclose,asciicheck,misspell ./...
-          working_directory: op-bindings
-      - run:
-          name: lint op-node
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -435,6 +340,76 @@ jobs:
...
+        description: Go Module Name
+        type: string
+    docker:
+      - image: cimg/go:1.19
+    # name: Lint <<parameters.module>>
+    steps:
+      - checkout
+      - run:
+          name: run lint
+          # command: golangci-lint run -E goimports,sqlclosecheck,bodyclose,asciicheck,misspell,errorlint -e "errors.As" -e "errors.Is" ./...
```
trianglesphere, line 355: Not all modules (op-node + op-e2e) pass this lint yet.

op-e2e/system_test.go
```diff
@@ -31,6 +32,8 @@ import (
 	"github.com/ethereum-optimism/optimism/op-node/withdrawals"
 )
 
+var enableParallelTesting bool = true
```
protolambda, line 35: Why not use the `-p 1` go testing flag to limit parallelism?

trianglesphere: `-p 1` does not stop tests from being run in parallel, it simply multiplexes them over a single process.

protolambda (approved):

mslipper (approved):

PR #3834, opened 2022-11-01
feat(ctb): use SCREAMING_SNAKE_CASE for immutables

smartcontracts:
**Description**
Makes sure we use SCREAMING_SNAKE_CASE for immutables everywhere. Adds legacy getters where necessary.

**Tests**
Existing tests should cover.

PR #3932, opened 2022-11-09
contracts-bedrock: OptimismMintableERC20 immutables

tynes:
**Description**

Turn the constructor args into immutables. There isn't a real reason for them to be in storage. The legacy tokens can have the constructor args held in storage and it shouldn't make a difference as post bedrock, this contract will be used on any newly created tokens. This just would mean that newly created tokens are a bit cheaper to use.


packages/contracts-bedrock/contracts/universal/OptimismMintableERC20.sol
```diff
@@ -16,12 +16,12 @@ contract OptimismMintableERC20 is IOptimismMintableERC20, ILegacyMintableERC20,
...
      * @notice Address of the corresponding version of this token on the remote chain.
      */
-    address public remoteToken;
+    address immutable public remoteToken;
 
     /**
      * @notice Address of the StandardBridge on this network.
      */
-    address public bridge;
+    address immutable public bridge;
```
maurelian, line 24: This should be SCREAMING_SNAKE a la #3834.

tynes: I don't love making these screaming snake case to be honest, it feels a bit annoying to keep both interfaces

maurelian: This code hasn't run in production so I don't think we need to keep both interfaces. (Assuming that by "both interfaces" you mean [like this](https://github.com/ethereum-optimism/optimism/pull/3834/files#diff-952f0a0004b0405b194415ffab201f3d55ec3f0937edcab0ce08fd38dbbf0232R59-R60).

tynes: Offchain services operating on the tokens may be running in production and we shouldn't break them

maurelian: You're right, so we need to maintain the interface with camelcase getters :)

But it's not clear to me why these immutables are a special case different from (for example) `OptimismMintableERC721` where we recently made the same change:

https://github.com/ethereum-optimism/optimism/blob/87702c741226ad4eb8a561dba349694e37c465a6/packages/contracts-bedrock/contracts/universal/OptimismMintableERC721.sol#L95-L104

tynes: Ok, you are right, updating and pushing code that follows the screaming snake case pattern now

maurelian: TY!

maurelian (approved):

mslipper (approved):

PR #3937, opened 2022-11-09
op-node: add ToB tests

maurelian:
**Description**

Adds tests created by Trail of Bits in the `op-node` package.

The added tests are named:
```
TestBedrockIndexer
TestCanonicalize
TestBuildL2DeveloperGenesis
TestBuildL2DeveloperGenesisDevAccountsFunding
TestBuildL2DeveloperGenesisDevAccountsFunding
TestMissingGasLimit
TestInvalidDepositInFCU
TestRejectCreateBlockBadTimestamp
FuzzBatchRoundTrip
FuzzDeriveDepositsRoundTrip
FuzzDeriveDepositsBadVersion
FuzzParseL1InfoDepositTxDataValid
FuzzParseL1InfoDepositTxDataBadLength
FuzzRejectCreateBlockBadTimestamp
```

**Reviewers please consider the following:** 

- Are the tests operating on correct assumptions about the system?
- Are they a net value add? 

Note that there are some 'todos' from the authors. I have not tried to resolve them and have kept them for posterity. Please flag any why are a high priority.

### DIFF
--- a/op-node/Makefile
+++ b/op-node/Makefile
@@ -29,6 +29,12 @@ fuzz:
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzUnmarshallLogEvent ./rollup/derive
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzParseFrames ./rollup/derive
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzFrameUnmarshalBinary ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzBatchRoundTrip ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzDeriveDepositsRoundTrip ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzDeriveDepositsBadVersion ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzParseL1InfoDepositTxDataValid ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzParseL1InfoDepositTxDataBadLength ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzRejectCreateBlockBadTimestamp ./rollup/driver
 
 
 .PHONY: \
--- a/op-node/Makefile
+++ b/op-node/Makefile
@@ -35,6 +35,7 @@ fuzz:
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzParseL1InfoDepositTxDataValid ./rollup/derive
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzParseL1InfoDepositTxDataBadLength ./rollup/derive
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzRejectCreateBlockBadTimestamp ./rollup/driver
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzDecodeDepositTxDataToL1Info ./rollup/driver
 
 
 .PHONY: \
--- a/op-node/rollup/derive/deposit_log_tob_test.go
+++ b/op-node/rollup/derive/deposit_log_tob_test.go
@@ -1,0 +1,211 @@
+package derive
+
+import (
+	"github.com/ethereum-optimism/optimism/op-node/testutils"
+	"github.com/ethereum-optimism/optimism/op-node/testutils/fuzzerutils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	fuzz "github.com/google/gofuzz"
+	"github.com/stretchr/testify/require"
+	"math/big"
+	"testing"
+)
+
+// fuzzReceipts is similar to makeReceipts except it uses the fuzzer to populate DepositTx fields.
+func fuzzReceipts(typeProvider *fuzz.Fuzzer, blockHash common.Hash, depositContractAddr common.Address) (receipts []*types.Receipt, expectedDeposits []*types.DepositTx) {
+	// Determine how many receipts to generate (capped)
+	var receiptCount uint64
+	typeProvider.Fuzz(&receiptCount)
+
+	// Cap our receipt count otherwise we might generate for too long and our fuzzer will assume we hung
+	if receiptCount > 0x10 {
+		receiptCount = 0x10
+	}
+
+	// Create every receipt we intend to
+	logIndex := uint(0)
+	for i := uint64(0); i < receiptCount; i++ {
+		// Obtain our fuzz parameters for generating this receipt
+		var txReceiptValues struct {
+			GoodReceipt bool
+			DepositLogs []bool
+		}
+		typeProvider.Fuzz(&txReceiptValues)
+
+		// Generate a list of transaction receipts
+		var logs []*types.Log
+		status := types.ReceiptStatusSuccessful
+		if txReceiptValues.GoodReceipt {
+			status = types.ReceiptStatusFailed
+		}
+
+		// Determine if this log will be a deposit log or not and generate it accordingly
+		for _, isDeposit := range txReceiptValues.DepositLogs {
+			var ev *types.Log
+			if isDeposit {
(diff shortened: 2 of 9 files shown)


op-node/rollup/derive/l1_block_info_tob_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+		// Get our info from out deposit tx
+		res, err := L1InfoDepositTxData(depTx.Data)
+		require.NoError(t, err, "expected valid deposit info")
+
+		// Verify all parameters match in our round trip deriving operations
+		assert.Equal(t, res.Number, l1Info.NumberU64())
+		assert.Equal(t, res.Time, l1Info.Time())
+		assert.True(t, res.BaseFee.Sign() >= 0)
+		assert.Equal(t, res.BaseFee.Bytes(), l1Info.BaseFee().Bytes())
```
trianglesphere, line 64: we generally use `require` instead of `assert`. In addition, we can use `go-cmp` with big ints or do a direct `a.Cmp(b) == 0`.

Ex for `go-cmp`. https://github.com/ethereum-optimism/optimism/blob/78826a131c022fa2527d350271f56fa42cb20ce8/op-node/rollup/derive/fuzz_parsers_test.go#L285-L287

op-node/rollup/derive/deposit_log_tob_test.go
```diff
@@ -0,0 +1,217 @@
...
+					Gas:                 fuzzedDepositInfo.Gas,
+					Data:                fuzzedDepositInfo.Data,
+					Mint:                fuzzedDepositInfo.Mint,
+					IsSystemTransaction: false,
+				}
+
+				// Marshal our actual log event
+				ev, err = MarshalDepositLogEvent(depositContractAddr, dep)
+				if err != nil {
+					// fmt.Println(err)
```
trianglesphere, line 77:
```suggestion
```

op-node/rollup/derive/deposit_log_tob_test.go
```diff
@@ -0,0 +1,217 @@
...
+					panic(err)
+				}
+
+				// If we have a good version and our tx succeeded, we add this to our list of expected deposits to
+				// return.
+				if status == types.ReceiptStatusSuccessful {
+					expectedDeposits = append(expectedDeposits, dep)
+				}
+
```
trianglesphere, line 86:
```suggestion
```

trianglesphere (commented):
Generally looking pretty good. There's a bit of an open question about `assert` vs `require`. I think we probably want to switch over to `require` to be consistent, but we kinda just adopted `require` rather than `assert` without thinking about it much.

op-node/rollup/derive/l1_block_info_tob_test.go
maurelian: I changed these to `require`. My feeling was that the `cmp` approach would be pretty verbose for the number of checks being done here.

op-node/rollup/derive/l1_block_info_tob_test.go
```diff
@@ -0,0 +1,83 @@
...
+			Scalar:      [32]byte{},
+		}
+		depTx, err := L1InfoDeposit(seqNr, l1Info, testSysCfg)
+		require.NoError(t, err)
+
+		// Get our info from out deposit tx
+		res, err := L1InfoDepositTxData(depTx.Data)
+		require.NoError(t, err, "expected valid deposit info")
+
+		// Verify all parameters match in our round trip deriving operations
```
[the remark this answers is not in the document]
maurelian: How's this?
https://github.com/ethereum-optimism/optimism/blob/c8aa6af73f598ccd19359bfc2ba5312d4e11b2cb/op-node/rollup/derive/l1_block_info_tob_test.go#L64-L66

op-node/rollup/derive/l1_block_info_tob_test.go
```diff
@@ -0,0 +1,84 @@
...
+		require.Equal(t, res.Number, l1Info.NumberU64())
+		require.Equal(t, res.Time, l1Info.Time())
+		require.True(t, res.BaseFee.Sign() >= 0)
+		require.Equal(t, res.BaseFee.Bytes(), l1Info.BaseFee().Bytes())
+		require.Equal(t, res.BlockHash, l1Info.Hash())
+		require.Equal(t, res.SequenceNumber, seqNr)
+		l1CfgFetcher := &testutils.MockL2Client{}
+		l1CfgFetcher.ExpectSystemConfigByL2Hash(res.BlockHash, testSysCfg, nil)
```
protolambda, line 66: After setting up expectations in a mock object, and running code, the expectations of the mock should be asserted with a `l1CfgFetcher.AssertExpectations(t)` call.
It doesn't look like this expected call that is set up is ever actually hit or asserted.

op-node/rollup/driver/state_tob_test.go
```diff
@@ -0,0 +1,253 @@
...
+	return l.retval, nil
+}
+
+// fixes "undefined: derive.Progress"
+// func (d TestDummyDerivationPipeline) Progress() derive.Progress {
+// 	return derive.Progress{
+// 		Origin: eth.L1BlockRef{},
+// 		Closed: false,
+// 	}
+// }
```
protolambda, line 79: This can be removed

op-node/rollup/driver/state_tob_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// Create a new block
+	// - L2Head's L1Origin, its timestamp should be greater than L1 genesis.
+	// - L2Head timestamp + BlockTime should be greater than or equal to the L1 Time.
+	err := s.createNewL2Block(ctx)
+
+	// Verify the L1Origin's timestamp is greater than L1 genesis in our config.
+	if l2l1OriginBlock.Number < s.config.Genesis.L1.Number {
+		assert.Nil(t, err)
+		return
```
protolambda, line 140: `require.NoError(t, err, "message why we don't expect an error here")` would be more idiomatic. The `require` essentially asserts and fails the test there, instead of having to return to stop execution.

op-node/rollup/driver/state_tob_test.go
```diff
@@ -0,0 +1,253 @@
...
+		assert.Nil(t, err)
+
+		// If we expected the outputter to error, capture that here
+		if outputProvider.willError {
+			assert.NotNil(t, err, "outputInterface failed to createNewBlock, so createNewL2Block should also have failed")
+			return
+		}
+
+		// Otherwise we should have no error.
+		assert.Nil(t, err)
```
protolambda, line 251: ditto here: `require` instead of `assert`, and `NoError` instead of `Nil` when checking errors.

protolambda (commented):
Some nits on the testing style I'd like to resolve before merging, LGTM otherwise.

op-node/rollup/derive/l1_block_info_tob_test.go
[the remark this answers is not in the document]
maurelian: How's this?
https://github.com/ethereum-optimism/optimism/blob/70123d6c89c9aa6713743c219afec966d2899857/op-node/rollup/derive/l1_block_info_tob_test.go#L44-L46

op-node/Makefile
```diff
@@ -29,6 +29,12 @@ fuzz:
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzUnmarshallLogEvent ./rollup/derive
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzParseFrames ./rollup/derive
 	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzFrameUnmarshalBinary ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzBatchRoundTrip ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzDeriveDepositsRoundTrip ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzDeriveDepositsBadVersion ./rollup/derive
+	go test -run NOTAREALTEST -v -fuzztime 10s -fuzz FuzzParseL1InfoDepositTxDataValid ./rollup/derive
```
protolambda, line 35: The new `FuzzDecodeDepositTxDataToL1Info` is missing here

protolambda (approved):
LGTM. But need to add new fuzz test to fuzzing makefile list.

op-node/Makefile
maurelian: Good catch, added it below.

trianglesphere (approved):
lgtm

PR #3944, opened 2022-11-09
ctb: Add echidna tests

maurelian:
**Description**

Adds an Echidna Fuzz testing harness written by Trail of Bits during our latest engagement.

Changes: 
1. Echidna added to the CI builder image. I've tested the build locally.
2. Adds new solidity files to `packages/contracts-bedrock/contracts/echidna`. These files implement the test harness itself. 
3. Adds `./scripts/echidna_fuzz.sh` which runs the fuzzer with a significantly reduced number of fuzzing runs from the default.

The script can be run with `ECHIDNA_TESTLIMIT=99 ECHIDNA_SEQLEN=9 ./scripts/echidna-fuzz.sh`.

This PR does not yet change the CCI config to actually run the fuzzing. That should come in a separate PR so that we can properly evaluate how to minimize the run time before imposing it on our dev process.

Metadata:
ENG-2956


packages/contracts-bedrock/hardhat.config.ts
```diff
@@ -365,9 +365,6 @@ const config: HardhatUserConfig = {
       },
     ],
     settings: {
-      metadata: {
-        bytecodeHash: 'none',
-      },
```
maurelian, line 370: Not sure how we feel about this, for some reason it was a requirement of echidna. 
Perhaps we can set it up to only disable this setting for echidna runs.

packages/contracts-bedrock/foundry.toml
```diff
@@ -13,7 +13,6 @@ remappings = [
   'ds-test/=node_modules/ds-test/src'
 ]
 extra_output = ['devdoc', 'userdoc', 'metadata', 'storageLayout']
-bytecode_hash = 'none'
```
maurelian, line 16: Not sure how we feel about this, for some reason it was a requirement of echidna. 
Perhaps we can set it up to only disable this setting for echidna runs.

packages/contracts-bedrock/hardhat.config.ts
mslipper: I'm concerned this will cause more conflicts with the `more` + bindings files. You should be able to use an env var in the hardhat config, and a `sed` command to replace the `bytecode_hash` value in the `foundry.toml` prior to running echidna.

maurelian: Fixed with this: d38cef9a9236bd2a5c0eb1b3493665192ef3f490

.circleci/config.yml
```diff
@@ -236,6 +236,7 @@ jobs:
           working_directory: packages/contracts-bedrock
       - run:
           name: test and generate coverage
+          no_output_timeout: 20m
```
maurelian, line 239: Intended to fix [this failure](https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/9873/workflows/1e515c93-6384-49e1-842c-7eb958f2e0b8/jobs/230614?invite=true#step-107-15).

maurelian:
Closing for now, will re-open with smaller PRs

maurelian:
closing will break up into smaller prs

PR #3950, opened 2022-11-09
feat(fd): event cache for Geth L1s

smartcontracts:
**Description**
Introduces a new event cache to the fault detector. Requires for running the fault detector against Geth as the L1 node and with certain L1 node providers.


smartcontracts:
Reported by users who couldn't use fault detector when connected to Geth L1

mslipper (approved):

PR #3952, opened 2022-11-09
op-node: Provide explicit presence check for L1 State

trianglesphere:
**Description**

The L1 State struct is not initialized at startup because initialization can fail (because it has to go over the network). This provides an explicit indication that a value is not present & checks for the value or accepts the default value.

**Metadata**
- Fixes ENG-2920

### DIFF
--- a/op-node/rollup/driver/l1_state.go
+++ b/op-node/rollup/driver/l1_state.go
@@ -20,6 +20,10 @@ type L1State struct {
 	l1Head      eth.L1BlockRef
 	l1Safe      eth.L1BlockRef
 	l1Finalized eth.L1BlockRef
+
+	haveL1Head      bool
+	haveL1Safe      bool
+	haveL1Finalized bool
 }
 
 func NewL1State(log log.Logger, metrics L1Metrics) *L1State {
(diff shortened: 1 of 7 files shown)


op-node/rollup/driver/state.go
```diff
@@ -135,7 +135,12 @@ func (s *Driver) createNewL2Block(ctx context.Context) error {
 	l2Finalized := s.derivation.Finalized()
 
 	// Figure out which L1 origin block we're going to be building on top of.
-	l1Origin, err := s.l1OriginSelector.FindL1Origin(ctx, s.l1State.L1Head(), l2Head)
+	l1Head, ok := s.l1State.L1Head()
+	if !ok {
```
tynes, line 139: Does comparing against `eth.L1BlockRef{}` not work here?

op-node/rollup/driver/l1_state.go
```diff
@@ -20,6 +20,10 @@ type L1State struct {
 	l1Head      eth.L1BlockRef
```
tynes, line 20: Could these be pointers and then we could check for `nil`?

op-node/rollup/driver/l1_state.go
```diff
@@ -20,6 +20,10 @@ type L1State struct {
...
 	l1Safe      eth.L1BlockRef
 	l1Finalized eth.L1BlockRef
+
+	haveL1Head      bool
+	haveL1Safe      bool
+	haveL1Finalized bool
```
protolambda, line 26: I still like to use the zeroed values for the have checks, instead of introducing these booleans in the state. The boolean in the return parameters is only there to force us to not assume the value isn't zero. We can check that in the getters too. If we introduce separate state for that, then it might create more edge cases where the value is zeroed but the haveL1X is true.

trianglesphere:
Closing in favor of #4035

PR #3973, opened 2022-11-10
feat(ctb): remove historicalTotalBlocks

smartcontracts:
**Description**
Removes historicalTotalBlocks from the L2OutputOracle.


tynes:
I'm into this but unsure if any offchain services use this, cc @trianglesphere @protolambda

protolambda (commented):
Not sure about the exact original reasoning for having both a `_historicalTotalBlocks` and `_startingBlockNumber`. I imagine there may be some value if the bedrock hardfork is at some non-stylish L2 block time, but if we want the proposals to start at a specific pretty block time, like making the first proposal align with the exact hour mark (and with fixed L2 block time and 1 hour proposal delta we get many predictable proposals). I'm not concerned about that timing if it simplifies the contract nicely though, nor do I think we are testing that type of complexity now, so removing it is nice.

There are no Go bedrock services usages of this parameter that I can think of. But in testing I think we need to remove https://github.com/ethereum-optimism/optimism/blob/74e054dd5f1918045803b359bb57a2c4fd1a5d07/op-e2e/setup.go#L165 (`L2OOContractConfig` as a whole is unused and can be removed, I think since it was changed to use op-chain-ops instead)

maurelian (approved):
Approving this on the basis of the contract changes looking good to me. It looks like (at least) these locations in the go code still need to be changed to get CI passing.

smartcontracts:
Ah yeah sorry, need to fix the go side of things. @protolambda do the changes look good to you other than the remaining go fixes?

maurelian:
fwiw I'm testing what I think are those remaining go changes here: https://github.com/ethereum-optimism/optimism/pull/3986

protolambda:
@smartcontracts 

> . @protolambda do the changes look good to you other than the remaining go fixes?

Yes

protolambda (approved):

smartcontracts:
@mergifyio requeue

PR #3975, opened 2022-11-11
fix(ctb): zero more addresses in MSD

smartcontracts:
**Description**
Has MSD zero more addresses in the AddressManager during migration.


tynes (approved):

mslipper (approved):

smartcontracts:
@mergifyio requeue

smartcontracts:
@mergifyio requeue

smartcontracts:
Actually this will break the DTL. Going to investigate which addresses we can remove an which we need to keep.

PR #3986, opened 2022-11-11
Go code: remove historicalTotalBlocks

maurelian:


op-e2e/setup.go
```diff
@@ -162,7 +162,6 @@ func writeDefaultJWT(t *testing.T) string {
 
 type L2OOContractConfig struct {
 	SubmissionFrequency   *big.Int
-	HistoricalTotalBlocks *big.Int
 }
```
protolambda, line 165: The whole struct can be removed I believe, it's not used.

PR #3993, opened 2022-11-14
Fix yarn.lock diff

maurelian:
**Description**

The `yarn.lock` file is dirty after running `yarn` in the root of the monorepo on my machine. This PR adds the diff.

I'm not sure how this got past `yarn --frozen-lockfile` in [the build](https://github.com/ethereum-optimism/optimism/blob/89526c42d45a98387578b71a7c8d016bdbc40797/.circleci/config.yml#L64) which should[ostensibly](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile) "not generate a lockfile and fail if an update is needed".

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,14 @@
 # yarn lockfile v1
 
 
+"@ampproject/remapping@^2.1.0":
+  version "2.2.0"
+  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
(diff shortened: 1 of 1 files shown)


maurelian:
OK, so _now_ this is [failing](https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/9533/workflows/0dd61000-24e9-46e0-a400-9c0824ecf89b/jobs/214374?invite=true#step-103-2) as expected. Presumably something is wrong on my system but leaving this open here for the moment.

PR #4001, opened 2022-11-14
fix(ctb): tweak addresses for MSD to wipe

smartcontracts:
**Description**
Tweaks the addresses that the MigrationSystemDictator will zero out inside of the AddressManager. The Proxy addresses should not be deleted since some external services may rely on them as part of the current system. The BondManager should be wiped so that no more state batches can be appended.


tynes (approved):

tynes:
```
/root/project/packages/contracts-bedrock/deploy/018-MigrationSystemDictatorSteps.ts
  209:22  warning  Insert `,`  prettier/prettier
```

maurelian (approved):

PR #4008, opened 2022-11-15
feat(ctb): MerkleTrie fuzzing

clabby:
# Overview

*WIP*

Adds a fuzz test input generator for testing Bedrock's `MerkleTrie.sol`.

Go version of #3847 

**Tests**

Adds fuzz tests for `MerkleTrie`

**Additional context**

n/a

**Metadata**
n/a


packages/contracts-bedrock/scripts/merkle-trie-fuzzer/go.mod
```diff
@@ -0,0 +1,11 @@
+module oplabs.co/merkle-trie-fuzzer
```
tynes, line 1: Wondering where this should live, we would want the module name to be `github.com/ethereum-optimism/optimism/contracts-bedrock/<path-to-directory>`

A few options:
- `contracts-bedrock/fuzz` or `contracts-bedrock/fuzzer`
- `contracts-bedrock/scripts/fuzzer`

I'm leaning towards it being a top level package named `fuzz` or `fuzzer`

Any opinions @trianglesphere @protolambda @maurelian @smartcontracts?

packages/contracts-bedrock/scripts/merkle-trie-fuzzer/README.md
```diff
@@ -0,0 +1,3 @@
+# `merkle-trie-fuzzer`
```
tynes, line 1: Perhaps we can make this not specific to merkle tree fuzzing and expand it over time to include all of the different fuzz targets by calling it `fuzzer` or something more general

packages/contracts-bedrock/package.json
```diff
@@ -16,19 +16,20 @@
   "scripts": {
     "build:forge": "forge build",
     "build:differential": "tsc scripts/differential-testing.ts --outDir dist --moduleResolution node --esModuleInterop",
+    "build:scripts": "(cd scripts/merkle-trie-fuzzer && go build main.go)",
```
tynes, line 19: I think you don't need the `cd`, you should be able to do `go -o <path-to-binary-ouput> build scripts/merkle-trie-fuzzer`

packages/contracts-bedrock/contracts/test/MerkleTrie.t.sol
```diff
@@ -358,10 +302,16 @@ contract MerkleTrie_Test is CommonTest {
...
         vm.expectRevert("MerkleTrie: value node must be last node in proof (leaf)");
-        MerkleTrie.get(
-            key,
-            proof,
-            root
-        );
+        MerkleTrie.get(key, proof, root);
+    }
+
+    /// @notice The `bytes4` parameter is to enable fuzz runs; it is ignored.
```
tynes, line 308: Adding a negative case would be great too, to fuzz invalid proofs passing because that would be really bad

packages/contracts-bedrock/scripts/merkle-trie-fuzzer/main.go
```diff
@@ -0,0 +1,122 @@
...
+	return hexutil.Encode(packed[32:])
+}
+
+func main() {
+	// Create an empty merkle trie
+	memdb := memorydb.New()
+	randTrie := trie.NewEmpty(trie.NewDatabase(memdb))
+
+	// Seed the random number generator with the current unix timestamp
+	rand.Seed(time.Now().UnixNano())
```
tynes, line 55: If this runs in parallel, is this enough entropy for it to be a good seed? Some runs might end up with the same time. I know that foundry fuzzing is good but not the greatest in the world when it comes to generating input values, perhaps taking the input value produced by foundry and hashing it with the time could produce better randomness?

packages/contracts-bedrock/package.json
```diff
@@ -16,19 +16,20 @@
...
-    "build": "hardhat compile && yarn autogen:artifacts && yarn build:ts && yarn typechain",
+    "build": "hardhat compile && yarn autogen:artifacts && yarn build:ts && yarn typechain && yarn build:scripts",
     "build:ts": "tsc -p tsconfig.json",
     "autogen:artifacts": "ts-node scripts/generate-artifacts.ts",
     "deploy": "hardhat deploy",
-    "test": "yarn build:differential && forge test",
-    "coverage": "yarn build:differential && forge coverage",
-    "coverage:lcov": "yarn build:differential && forge coverage --report lcov",
-    "gas-snapshot": "yarn build:differential && forge snapshot",
+    "test": "yarn build:differential && yarn build:scripts && forge test",
```
maurelian, line 25: I haven't tested, but it feels like this might slow down the test run if you have to do this every time? If so, perhaps we could just throw an error if the script isn't built yet.

packages/contracts-bedrock/scripts/merkle-trie-fuzzer/go.mod
maurelian: Agree with "top level" (in `packages/contracts-bedrock`).

Maybe call it `go-fuzz`, as I expect there to be other fuzzing approaches as well

packages/contracts-bedrock/scripts/merkle-trie-fuzzer/main.go
clabby: Good spot- changed from `math/rand` to `crypto/rand` rather than passing hashed values as an argument to the binary.

packages/contracts-bedrock/package.json
clabby: At the moment, compilation of the binary takes less than a second, so it doesn't introduce much of a slowdown. Once this package is expanded, it might be a better idea to go this route.

packages/contracts-bedrock/scripts/merkle-trie-fuzzer/go.mod
clabby: Got it- moved to top level as `go-fuzzers`

packages/contracts-bedrock/contracts/test/MerkleTrie.t.sol
clabby: Will do today :smile:

trianglesphere (commented):
Really nice work. I've left a bunch of comments, but they're primarily focused on go tips & tricks and go style.

packages/contracts-bedrock/go-fuzz/trie/trie.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	// Remove the pointer and encode the packed bytes as a hex string
+	return hexutil.Encode(packed[32:])
+}
+
+// Helper that generates a cryptographically secure random 64-bit integer
+// between the range [min, max]
+func randRange(min int64, max int64) int64 {
+	r, err := rand.Int(rand.Reader, new(big.Int).Sub(new(big.Int).SetInt64(max), new(big.Int).SetInt64(min)))
```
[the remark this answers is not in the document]
clabby: The idea behind switching to `crypto/rand` was to allow for randomness while executing these tests in parallel. Originally, I had passed the hash of a 32 byte fuzz input from the foundry test + the current time as the seed for `math/rand`, but felt this was a cleaner solution. See [mark's above comment](https://github.com/ethereum-optimism/optimism/pull/4008#discussion_r1023425921)

packages/contracts-bedrock/go-fuzz/trie/trie.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		log.Fatalf("Error packing TrieTestCase: %v", err)
+	}
+
+	// Remove the pointer and encode the packed bytes as a hex string
+	return hexutil.Encode(packed[32:])
+}
+
+// Helper that generates a cryptographically secure random 64-bit integer
+// between the range [min, max]
```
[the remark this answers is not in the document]
clabby: Good spot, changed!

packages/contracts-bedrock/go-fuzz/trie/trie.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Generate an abi-encoded `trieTestCase` of a specified variant
+func FuzzTrie(variant string) {
+	if len(variant) < 1 {
+		log.Fatal("Must pass a variant to the trie fuzzer!")
+	}
+
+	var testCase trieTestCase
+	switch variant {
+	case "valid":
```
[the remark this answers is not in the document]
clabby: Swapped over to using enums- went with string variants so that they are readable within foundry tests w/o having to reference the variant docs.

i.e.

`ffi.getMerkleTrieFuzzCase("valid")` reads better than `ffi.getMerkleTrieFuzzCase(0)`

packages/contracts-bedrock/go-fuzz/trie/trie.go
```diff
@@ -0,0 +1,193 @@
...
+	return (new(big.Int).Add(r, new(big.Int).SetInt64(min))).Int64()
+}
+
+// Represents a range between two 64-bit integers
+type intRange struct {
+	min int64
+	max int64
+}
+
+// Weird golang type coercion wizardry
```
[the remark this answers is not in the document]
clabby: Spoke with you about this in DM last week- huge thanks for the insight! Sticking with this for now given that it is not exported.

packages/contracts-bedrock/go-fuzz/trie/trie.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Generate proof for `key`'s inclusion in our trie
+	var proof proofList
+	if err := randTrie.Prove(key, 0, &proof); err != nil {
+		log.Fatal("Error creating proof for randomly selected key's inclusion in generated trie")
+	}
+
+	// Create our test case with the data collected
+	testCase := trieTestCase{
+		Root:  (*[32]byte)(randTrie.Hash().Bytes()),
```
tynes, line 140: I'm surprised that `randTrie.Hash` doesn't just work here if it returns a `common.Hash` which is just an alias to `[32]byte`

clabby:
Reopened in #4142 - certain CI tasks didn't like the branch name.

PR #4013, opened 2022-11-16
fix(ctb): allow same L2OO owner and proposer

smartcontracts:
**Description**
Tweaks the L2OutputOracle to allow owner and proposer addresses to be the same. Extra logic required to prevent this case added smart contract code to a critical contract and it's unlikely that this will be a problem in practice.


maurelian (approved):

tynes (commented):
Operationally, I think having them as separate accounts is more safe. The proposer must be a hot key, the owner should be a multisig of cold keys. Enforcing this prevents a footgun. What is your rationale for this change?

smartcontracts:
> Operationally, I think having them as separate accounts is more safe. The proposer must be a hot key, the owner should be a multisig of cold keys. Enforcing this prevents a footgun. What is your rationale for this change?

I fully agree that the keys should be separate accounts in practice. But I don't see the need to encode this separation in the smart contracts, especially if it means doing stuff like modifying the `transferOwnership` function.

tynes (approved):

smartcontracts:
@mergifyio requeue

smartcontracts:
@mergifyio requeue

smartcontracts:
@mergifyio requeue

smartcontracts:
@mergifyio requeue

smartcontracts:
@mergifyio requeue

PR #4014, opened 2022-11-17
feat(ctb): allow multiple output deletions

smartcontracts:
**Description**
Updates the L2OutputOracle to allow deletion of multiple outputs at the same time. This is a safety mechanism that prevents a compromised sequencer from forcing the multisig into deleting potentially thousands of bad outputs one by one.


packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -149,32 +145,29 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
...
+    function deleteL2Outputs(uint256 _l2BlockNumber) external onlyOwner {
+        // Simple check that accomplishes three things:
+        //   1. Prevents deleting anything from before the genesis block.
+        //   2. Prevents deleting anything after the latest block.
+        //   3. Prevents deleting anything other than a checkpoint block.
         require(
-            _proposal.outputRoot == outputToDelete.outputRoot,
-            "L2OutputOracle: output root to delete does not match the latest output proposal"
+            l2Outputs[_l2BlockNumber].outputRoot != bytes32(0),
+            "L2OutputOracle: cannot delete a non-existent output"
```
maurelian, line 162: In the case that an output has already been deleted, check number 2 above is not necessarily true. 

We also need to prevent deleting finalized outputs.

maurelian: > check number 2 above is not necessarily true.

To be clear, I think that's OK. We just need to make sure our comments are accurate.

smartcontracts: Fixed the first issue. I don't know how to best handle the second issue because the finalization period is part of the portal and not the output oracle. It may not actually make sense to do the check in the oracle because e.g., if we replaced the proof system with a zk proof then there is no sense of "finalization".

maurelian: Right, an output is either valid or not regardless of age. SGTM.

maurelian:
Will re-review once CI is passing.

maurelian (approved):

clabby (approved):

packages/contracts-bedrock/contracts/test/L2OutputOracle.t.sol
```diff
@@ -280,29 +280,24 @@ contract L2OutputOracleTest is L2OutputOracle_Initializer {
...
 
-    event OutputDeleted(
-        bytes32 indexed l2Output,
-        uint256 indexed l1Timestamp,
+    event OutputsDeleted(
         uint256 indexed l2BlockNumber
     );
 
-    function test_deleteOutput() external {
+    function test_deleteOutputs() external {
```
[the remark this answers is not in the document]
smartcontracts: Oh yeah that's reasonable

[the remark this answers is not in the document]
smartcontracts: Added :vulcan_salute:

PR #4015, opened 2022-11-17
feat(ctb): MSD exit point 1

smartcontracts:
**Description**
Introduces the first exit point for the MigrationSystemDictator.

### DIFF
--- a/packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
+++ b/packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
@@ -18,14 +18,24 @@ import { BaseSystemDictator } from "./BaseSystemDictator.sol";
  */
 contract MigrationSystemDictator is BaseSystemDictator {
     /**
+     * @notice Step after which exit 1 can no longer be used.
+     */
+    uint8 constant EXIT_1_NO_RETURN_STEP = 3;
+
+    /**
      * @notice Step where proxy ownership is transferred.
      */
-    uint8 constant PROXY_TRANSFER_STEP = 3;
+    uint8 constant PROXY_TRANSFER_STEP = 4;
 
     /**
      * @notice Whether or not the deployment is finalized.
      */
     bool public finalized;
+
+    /**
+     * @notice Address of the old L1CrossDomainMessenger implementation.
+     */
+    address public oldL1CrossDomainMessenger;
 
     /**
      * @param _config System configuration.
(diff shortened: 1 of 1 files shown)


packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
```diff
@@ -17,16 +17,26 @@ import { BaseSystemDictator } from "./BaseSystemDictator.sol";
...
+    /**
+     * @notice Step after which exit 1 can no longer be used.
+     */
+    uint8 constant EXIT_1_NO_RETURN_STEP = 3;
+
     /**
      * @notice Step where proxy ownership is transferred.
      */
-    uint8 constant PROXY_TRANSFER_STEP = 3;
+    uint8 constant PROXY_TRANSFER_STEP = 4;
```
maurelian, line 28: IMO this should also be referenced in the `step` modifier on `step4()`.

packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
```diff
@@ -63,12 +73,25 @@ contract MigrationSystemDictator is BaseSystemDictator {
      *         deposit halt flag to tell the Sequencer's DTL to stop accepting deposits.
      */
     function step2() external onlyOwner step(2) {
+        // Store the address of the old L1CrossDomainMessenger implementation. We will need this
+        // address in the case that we have to exit early.
+        oldL1CrossDomainMessenger = config.globalConfig.addressManager.getAddress(
+            "OVM_L1CrossDomainMessenger"
```
maurelian, line 79: Feels slightly odd here. Why not do this in the constructor?

maurelian (approved):

packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
```diff
@@ -229,4 +252,23 @@ contract MigrationSystemDictator is BaseSystemDictator {
 
         finalized = true;
     }
+
+    /**
+     * @notice First exit point, can only be called before step 3 is executed.
```
tynes, line 257: I don't think before is accurate here because of the require statement that checks the `currentStep == EXIT_1_NO_RETURN_STEP` - it must be called at an exact place

tynes: Should it be `<=` instead of `==`?

smartcontracts: It must be `==` because `DTL_SHUTOFF_BLOCK` and `oldL1CrossDomainMessenger` are not set until `EXIT_1_NO_RETURN_STEP`.

tynes (approved):

PR #4017, opened 2022-11-19
Fix: Adding proper debug output for L1 CTC Address env var in l2geth-exporter

epociask:
**Description**
Added proper debug output for `OVM_CTC_ADDRESS` to ensure output matched expected variable. 

**Tests**
Testing would be unnecessary here. 

**Additional context**
Given use and demand, it could make sense to use separate `Config` struct to store env var references. Furthermore an extraction function could be used for blank env var validation cases instead of copying the same code three times. 

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/l2geth-exporter/main.go
+++ b/l2geth-exporter/main.go
@@ -33,7 +33,7 @@ func main() {
 	}
 	ctcAddress := os.Getenv("OVM_CTC_ADDRESS")
 	if ctcAddress == "" {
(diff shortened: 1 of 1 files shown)


smartcontracts (approved):
Thanks! Looks good to me.

tynes (approved):
Looks good, the `l2geth-exporter` needs some cleanup for sure

tynes:
I added a changeset, this should be good to be merged

PR #4028, opened 2022-11-21
fix(sdk): test flakes

smartcontracts:
**Description**
Fixes test flakes in the SDK

### DIFF
--- a/packages/sdk/tasks/deposit-erc20.ts
+++ b/packages/sdk/tasks/deposit-erc20.ts
@@ -4,11 +4,12 @@ import { task, types } from 'hardhat/config'
 import { HardhatRuntimeEnvironment } from 'hardhat/types'
 import '@nomiclabs/hardhat-ethers'
 import 'hardhat-deploy'
+import { Event, Contract, Wallet, providers, utils } from 'ethers'
 import {
   predeploys,
   getContractDefinition,
 } from '@eth-optimism/contracts-bedrock'
-import { Event, Contract, Wallet, providers, utils } from 'ethers'
+import { sleep } from '@eth-optimism/core-utils'
 
 import {
   CrossChainMessenger,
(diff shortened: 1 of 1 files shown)


packages/sdk/tasks/deposit-erc20.ts
```diff
@@ -247,18 +248,35 @@ task('deposit-erc20', 'Deposits WETH9 onto L2.')
     await depositTx.wait()
     console.log(`ERC20 deposited - ${depositTx.hash}`)
 
-    const messageReceipt = await messenger.waitForMessageReceipt(depositTx)
-    if (messageReceipt.receiptStatus !== 1) {
-      throw new Error('deposit failed')
-    }
+    // Deposit might get reorged, wait 10s and also log for reorgs.
+    let prevBlockNumber = 0
+    for (let i = 0; i < 10; i++) {
```
smartcontracts, line 253: Literally just waiting for 10s to bypass deposit reorgs, I think this should fix the issue.

packages/sdk/tasks/deposit-erc20.ts
```diff
@@ -247,18 +248,35 @@ task('deposit-erc20', 'Deposits WETH9 onto L2.')
...
+      const messageReceipt = await messenger.waitForMessageReceipt(depositTx)
+      if (messageReceipt.receiptStatus !== 1) {
+        throw new Error('deposit failed')
+      }
+
+      if (
```
smartcontracts, line 259: Logging here just so we know a reorg happened

roninjin10 (approved):

packages/sdk/tasks/deposit-erc20.ts
trianglesphere: TBH I'm kinda surprised that just a block number difference would show up with reorgs, I'd expect to need a block hash check to catch every reorg.

trianglesphere (approved):

tynes (approved):

PR #4032, opened 2022-11-21
CI: Enable Hive to work with external contributors

trianglesphere:
**Description**

This PR modifies CI to let external contributors more easily contribute. I've split the docker-publish job into a docker-build and docker-publish job. Hive now only depends on the docker-build job (which can be run by external contributors).

This makes one change to the hive job to remove the `docker.pull=true` flag. This is because it loads the image locally but in the case of an external contributor the image will not be published. Docker will use the local image first & then pull other images if they are not present.

I've also removed the conditional step in docker-build. It was previously used for nightly builds, but we do not have those any more (removed in #3330).

**Testing**

I've tested that hive works from just the build step by disabling the publish step. With the docker login check, I think that this will work for external contributors, but I have not yet tried.

**Metadata**
- Fixes ENG-3033

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,8 +83,7 @@ jobs:
             - "packages/contracts/libraries"
             - "packages/contracts/standards"
 
-
-  docker-publish:
+  docker-build:
     environment:
       DOCKER_BUILDKIT: 1
     parameters:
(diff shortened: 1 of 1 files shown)


zhwrd (approved):

zhwrd:
Nice! Had planned to split these up at some point, thanks for taking care of it.

mslipper (approved):

mslipper:
Looks great. Will update the branch protection accordingle.

PR #4033, opened 2022-11-21
feat(ctb): move L2OO genesis setup to init

smartcontracts:
**Description**
Moves genesis output, timestamp, and block number setup to the initializer. This is necessary because we will not have this information at deploy time for the migration deployment.


maurelian (approved):

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -93,49 +93,54 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
...
-        bytes32 _genesisL2Output,
+        uint256 _l2BlockTime,
+        bytes32 _startingL2Output,
         uint256 _startingBlockNumber,
         uint256 _startingTimestamp,
-        uint256 _l2BlockTime,
         address _proposer,
         address _owner
     ) Semver(0, 0, 1) {
-        require(
```
tynes, line 112: I feel like this require was to prevent config issues, a bad value here could cause the opnode to break

smartcontracts: It's in `initialize` now

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -22,19 +22,19 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
...
      */
-    uint256 public immutable STARTING_BLOCK_NUMBER;
+    uint256 public immutable L2_BLOCK_TIME;
 
     /**
-     * @notice The timestamp of the first L2 block recorded in this contract.
+     * @notice The number of the first L2 block recorded in this contract.
      */
-    uint256 public immutable STARTING_TIMESTAMP;
+    uint256 public startingBlockNumber;
```
tynes, line 32: Why is this no longer immutable?

packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
```diff
@@ -143,14 +143,19 @@ contract MigrationSystemDictator is BaseSystemDictator {
      * @notice Upgrades and initializes proxy contracts.
      */
     function step5() external onlyOwner step(5) {
+        // Dynamic config must be set before we can initialize the L2OutputOracle.
+        require(dynamicConfigSet, "FreshSystemDictator: dynamic config not yet initialized");
```
tynes, line 147: This contract is called `MigrationSystemDictator`

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
tynes: I feel like we are just creating churn by changing these from immutables to being in storage

smartcontracts: It can't be immutable because it's set in the initializer and not in the constructor

packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
smartcontracts: Oops, copy/paste error

tynes (approved):

PR #4039, opened 2022-11-22
feat(ctb): unify fresh and migration deployments

smartcontracts:
**Description**
Removes the concept of a "Fresh" deployment and deploys some legacy contracts where necessary so that all new deployments will mirror the mainnet system. Although this enshrines some legacy code, it means that we won't have multiple conflicting deployments in the future.


tynes:
Looks like a flake in CI for the `op-node` tests, rerunning them

tynes (approved):

maurelian (approved):

PR #4041, opened 2022-11-22
ctb: Configurable verification for deployer network

mslipper:
Most ephemeral networks created via the regntool don't need verification. This PR allows verification to be enabled on a case-by-case basis by setting an environment variable.

### DIFF
--- a/packages/contracts-bedrock/hardhat.config.ts
+++ b/packages/contracts-bedrock/hardhat.config.ts
@@ -38,6 +38,7 @@ const config: HardhatUserConfig = {
       chainId: Number(process.env.CHAIN_ID),
       url: process.env.L1_RPC || '',
       accounts: [process.env.PRIVATE_KEY_DEPLOYER || ethers.constants.HashZero],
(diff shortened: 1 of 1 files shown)


packages/contracts-bedrock/hardhat.config.ts
```diff
@@ -38,6 +38,7 @@ const config: HardhatUserConfig = {
       chainId: Number(process.env.CHAIN_ID),
       url: process.env.L1_RPC || '',
       accounts: [process.env.PRIVATE_KEY_DEPLOYER || ethers.constants.HashZero],
+      live: process.env.VERIFY_CONTRACTS === 'true',
     },
```
tynes, line 42: I see that when live is false, it skips verification here: https://github.com/ethereum-optimism/optimism/blob/ea2cf94cc5940f9b91bc976a2095a167c5a61b61/packages/contracts-bedrock/src/deploy-utils.ts#L77

tynes (approved):

mslipper:
Merging this - I need it to deploy the beta. This is a low-risk change.

PR #4045, opened 2022-11-22
ctb: Change getL2Output to getL2OutputAfter

maurelian:
**Description**

Renames `getL2Output` to `getL2OutputAfter`, which is semantically more descriptive, and could prevent future issues due to misunderstandings. 

This change may break external code, and so is posted here for consideration and discussion. 

**Tests**

Tests are updated.


clabby (changes requested):
Looks good- will approve after the gas snapshot is updated.

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -231,7 +231,7 @@ export class CrossChainMessenger {
    *
    * @returns The L2OutputOracle parameters.
    */
-  public async getL2OutputOracleParameters(): Promise<L2OutputOracleParameters> {
+  public async getL2OutputAfterOracleParameters(): Promise<L2OutputOracleParameters> {
```
smartcontracts, line 234: Lawl I think this is an errant find/replace

maurelian: barf

clabby (approved):

smartcontracts (approved):

smartcontracts (changes requested):
Superseded by #4093

tynes:
Going to close this in favor of #4093

PR #4051, opened 2022-11-22
feat(dtl): handle deposit shutoff

smartcontracts:
**Description**
Updates the DTL to correctly handle the deposit shutoff flag. We put the flag inside the AddressManager because the DTL already had easy access to the AddressManager. We parse the address as a uint256 and check if it's non-zero. By checking this every loop we can smoothly recover if the shutoff flag is reset.

### DIFF
--- a/packages/data-transport-layer/src/services/l1-ingestion/service.ts
+++ b/packages/data-transport-layer/src/services/l1-ingestion/service.ts
@@ -4,7 +4,7 @@ import { BaseService, Metrics } from '@eth-optimism/common-ts'
 import { TypedEvent } from '@eth-optimism/contracts/dist/types/common'
 import { BaseProvider, StaticJsonRpcProvider } from '@ethersproject/providers'
 import { LevelUp } from 'levelup'
-import { constants } from 'ethers'
+import { BigNumber, constants } from 'ethers'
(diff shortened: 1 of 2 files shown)


packages/data-transport-layer/src/services/l1-ingestion/service.ts
```diff
@@ -230,14 +230,36 @@ export class L1IngestionService extends BaseService<L1IngestionServiceOptions> {
           targetL1Block,
         })
 
+        // We always try loading the deposit shutoff block in every loop! Less efficient but will
+        // guarantee that we don't miss the shutoff block being set and that we can automatically
+        // recover if the shutoff block is set and then unset.
```
tynes, line 235: Don't you think its pretty ugly to nest an await like that and then call `toNumber` on its own line?

tynes (approved):
Looks good to me besides style nit

packages/data-transport-layer/src/services/l1-ingestion/service.ts
smartcontracts: Meh I don't think it's a huge problem. We're gonna toss all this code soon anyway.

mslipper (approved):

tynes (approved):

PR #4066, opened 2022-11-23
feat(ctb): remove genesis output

smartcontracts:
**Description**
Removes the startingL2Output from the L2OutputOracle. There's no reason for this parameter to exist, there's nothing special about the startingL2Output. It only complicates the initialization process. Instead, the first output can be published by a proposer just like any other output.


packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -136,10 +132,9 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
             "L2OutputOracle: starting L2 timestamp must be less than current time"
         );
 
+        startingTimestamp = _startingTimestamp;
         startingBlockNumber = _startingBlockNumber;
         latestBlockNumber = _startingBlockNumber;
-        startingTimestamp = _startingTimestamp;
-        l2Outputs[startingBlockNumber] = Types.OutputProposal(_startingL2Output, block.timestamp);
```
smartcontracts, line 142: Note that now `getL2Output(startingBlockNumber)` will fail which is fine because there's no proposal for that block number.

maurelian (approved):

packages/contracts-bedrock/deploy/018-MigrationSystemDictatorSteps.ts
```diff
@@ -362,7 +349,40 @@ const deployFn: DeployFunction = async (hre) => {
   }
 
   for (let i = 1; i <= 6; i++) {
-    if ((await MigrationSystemDictator.currentStep()) === i) {
+    const currentStep = await MigrationSystemDictator.currentStep()
+    if (currentStep === i) {
+      if (
```
tynes, line 354: I'm sorry kelvin but the nested awaits in an if statement triggers me :P barf

packages/contracts-bedrock/deploy/018-MigrationSystemDictatorSteps.ts
```diff
@@ -362,7 +349,40 @@ const deployFn: DeployFunction = async (hre) => {
...
+              throw new Error(
+                `Cannot fetch block tag ${hre.deployConfig.l1StartingBlockTag}`
+              )
+            }
+            deployL2StartingTimestamp = l1StartingBlock.timestamp
+          }
+
+          await MigrationSystemDictator.updateL2OutputOracleDynamicConfig({
+            l2OutputOracleStartingBlockNumber:
+              hre.deployConfig.l2OutputOracleStartingBlockNumber,
```
tynes, line 378: Is there a relationship between `l1StartingBlockTag` and `l2OutputOracleStartingBlockNumber` ?

smartcontracts: nope

tynes:
This will greatly simplify the deployment

packages/contracts-bedrock/deploy/018-MigrationSystemDictatorSteps.ts
tynes: The unnecessary variables become necessary for readability :)

PR #4069, opened 2022-11-23
Fix: build and publish all bedrock docker images before tagging release

zhwrd:
**Description**

rebuild and push all images for every git tag and then tag the one image that corresponds to the git tag
this is somewhat wasteful but ensures that the tagging works for any tagging strategy. 
will revisit once we address simplified release process / versioning.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -955,8 +955,6 @@ workflows:
           docker_name: op-node
           docker_tags: <<pipeline.git.revision>>,<<pipeline.git.branch>>
           docker_context: .
-          context:
-            - gcr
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -955,8 +955,6 @@ workflows:
           docker_name: op-node
           docker_tags: <<pipeline.git.revision>>,<<pipeline.git.branch>>
           docker_context: .
-          context:
```
zhwrd, line 958: this context is no longer necessary since we've separated publish from build (also forked builds will now work)

mslipper (approved):

PR #4078, opened 2022-11-24
op-chain-ops: add bedrock transition receipts

tynes:
**Description**

To make withdrawals really easy, we need to add in receipts corresponding to the migrated withdrawals. This will allow the exact same flow to be used when creating the withdrawing transactions as normal
withdrawals.

Closes ENG-3012

### DIFF
--- a/op-chain-ops/crossdomain/withdrawal.go
+++ b/op-chain-ops/crossdomain/withdrawal.go
@@ -4,9 +4,20 @@ import (
 	"errors"
 	"math/big"
 
+	"github.com/ethereum-optimism/optimism/op-bindings/predeploys"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
+)
+
+var (
+	SentMessageEventABI               = "SentMessage(address,address,bytes,uint256)"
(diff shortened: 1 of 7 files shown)


op-chain-ops/crossdomain/withdrawal.go
```diff
@@ -130,3 +141,100 @@ func (w *Withdrawal) StorageSlot() (common.Hash, error) {
...
+		Topics: []common.Hash{
+			SentMessageEventABIHash,
+			w.Target.Hash(),
+		},
+		Data:        data,
+		BlockNumber: hdr.Number.Uint64(),
+		TxHash:      common.Hash{},
+		TxIndex:     0,
+		BlockHash:   hdr.Hash(),
+		Index:       0,
```
trianglesphere, line 182: I'm a little worried about using the log + transaction index.

tynes: I set the tx index to 0 as well as the index to 0 for all logs. It is unexpected behavior to have the index be the same for each log, but I don't think it will break anything. I could refactor things such that the index is non zero and increments

tynes: i've canonicalized the tx index + log index

trianglesphere: thanks for doing this

trianglesphere (approved):
lgtm

mslipper (approved):

PR #4084, opened 2022-11-24
Test Clean Up

maurelian:
**Description**

**Test organizing principles**

1. Uses `contract`s to organize the test suite similar to how mocha uses `describe`. 
1. Every non-trivial state changing function should have a contract for happy and sad path tests. This helps to make it very obvious where there are not sad tests (I haven't added any new tests, so there are quite a few empty placeholder contracts). 
1. Simpler functions like getters and setters are grouped together into test contracts.
2. All contracts and functions should be consistently named following a convention. 

**Test naming convention**

Test function names are split by underscores, into 3 or 4 parts. An example function name is `test_onlyOwner_callerIsNotOwner_reverts()`. 

The parts are: `[method]_[FunctionName]_[reason]_[success]`, where:
1. `[method]` is either `test` or `testFuzz`
2. `[FunctionName]` is the name of the function or higher level behavior being tested.
4. `[reason]` is an optional description for the behavior being tested.
3. `[success]` must be one of `success` or `reverts` or `_fails`. 

**Name validation**

I also wrote a quick and dirty script which can be run to validate function names and identify ones which are not yet compliant. From within the bedrock package:
`ts-node scripts/forges-test-names.ts`.


maurelian:
Closing will break up.

PR #4092, opened 2022-11-28
feat(ci): Proof of concept of nx

roninjin10:
Add nx to monorepo

- nx capable of handling go if we wanted
- can completely replace lerna
- much simpler and lightweight than lerna
- NX will cache builds and subdependencies remotely both locally and on CI
- Added nx to gateway repo here https://github.com/ethereum-optimism/gateway/pull/1211

image

- It also gives us an awesome vscode extension and cli 

image

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,0 +1,21 @@
+{
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "@nrwl/nx-cloud",
+      "options": {
+        "cacheableOperations": ["build"],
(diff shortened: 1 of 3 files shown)


nx.json
```diff
@@ -0,0 +1,21 @@
+{
```
roninjin10, line 1: pr generated with `npx add-nx-to-monorepo`

mslipper:
How much does this improve our build time?

PR #4093, opened 2022-11-28
feat(ctb): make SUBMISSION_INTERVAL modifiable

smartcontracts:
**Description**
Updates the L2OutputOracle so that SUBMISSION_INTERVAL can be modified. This is a relatively significant refactor that makes the l2Outputs mapping keyed by index rather than L2 block number.

**Tests**
Tests have been updated.

**Metadata**
ENG-3041


tynes:
Looks like `SUBMISSION_INTERVAL` is still `immutable`? So to change it, we would upgrade the implementation. Perhaps we should note that in the natspec of the contract

smartcontracts:
> Looks like `SUBMISSION_INTERVAL` is still `immutable`? So to change it, we would upgrade the implementation. Perhaps we should note that in the natspec of the contract

Yes it's still immutable but the variable can now be safely changed (by upgrading the implementation)

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -143,34 +128,26 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
...
-        );
-
-        // We're setting the latest block number back to the checkpoint block before the given L2
-        // block number. Next proposal will overwrite the deleted output and following proposals
-        // will delete any outputs after that.
-        latestBlockNumber = _l2BlockNumber - SUBMISSION_INTERVAL;
+        // Use assembly to delete the array elements because Solidity doesn't allow it.
+        assembly {
+            sstore(l2Outputs.slot, _l2OutputIndex)
+        }
```
tynes, line 148: Wondering if there is an off by 1 error here

smartcontracts: I think it's correct. Imagine if there are 2 elements in the array. If we want to delete the second element (element index 1) then we give `_l2OutputIndex = 1`. So we store `1` as the length and then the array becomes 1 element long, which deletes element 2.

tynes (approved):

tynes:
For the `indexer-tests`:

```
WARN [11-28|21:13:48.869] Served eth_call                          reqid=25 duration="217.352µs" err="execution reverted"                     errdata=0x4e487b710000000000000000000000000000000000000000000000000000000000000032
```

Looks like a `Panic(uint256)` with `0x32` which indicates

> 0x32: If you access an array, bytesN or an array slice at an out-of-bounds or negative index (i.e. x[i] where i >= x.length or i < 0).

Source: https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/

tynes:
Looks like this is failing:

```
WARN [11-28|22:40:48.160] Served eth_estimateGas                   reqid=47 duration=4.158602ms  err="execution reverted"                     errdata=0x4e487b710000000000000000000000000000000000000000000000000000000000000032
```

Which is `Panic(uint256)` with `0x32` so oob access again

tynes (changes requested):
Generally looks good but integration tests are now failing with out of bounds access on the array

op-e2e/system_test.go
```diff
@@ -842,8 +842,15 @@ func TestWithdrawals(t *testing.T) {
 	portal, err := bindings.NewOptimismPortal(predeploys.DevOptimismPortalAddr, l1Client)
 	require.Nil(t, err)
 
+	oracle, err := bindings.NewL2OutputOracleCaller(predeploys.DevL2OutputOracleAddr, l1Client)
+	require.Nil(t, err)
+
 	opts.Value = nil
 
+	// Get output index
```
tynes, line 850: slight nit, this should be above `opts.Value = nil`, would read a bit better

tynes:
Looks like there are now some merge conflicts that need to be fixed

mslipper:
GitHub requested a whole gang of reviewers because this PR touches the indexer/SDK/contracts. I'm fine merging this if Mark + Maurelian approve.

tynes:
I'm pretty confused about why the devnet is reorging, I feel like it never used to do that

tynes:
It looks like the storage snapshot needs to be regenerated cc @smartcontracts

tynes:
This is really annoying

```
ERC20 deposited - 0xd6040febb53bd1ae9cb65faa973039f38c5eb07568d508f334cf5ba1682fd1b0
Block hash changed from 0x3d36995c96dbf84ec22b30e993d5af6c2c4a19fba625ae59c7a875bc6330cadc to 0xc66669cf9ed697bef6e1e923570495f4c672ebd7325a8942d3bf05142ea5254b
Deposit failed, retrying...
Block hash changed from 0xc66669cf9ed697bef6e1e923570495f4c672ebd7325a8942d3bf05142ea5254b to 0x09e83c6a33e0bffeab62685621831c247d13fb79a73bc86c2906d6441452864a
```

mslipper:
Yarn error is a fluke, will re-run once the devnet jobs complete.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -45,36 +46,31 @@ contract L2OutputOracle is Initializable, Semver {
...
      * @param l2BlockNumber The L2 block number of the output root.
+     * @param l1Timestamp   The L1 timestamp when proposed.
      */
     event OutputProposed(
         bytes32 indexed outputRoot,
-        uint256 indexed l1Timestamp,
-        uint256 indexed l2BlockNumber
+        uint256 indexed l2OutputIndex,
+        uint256 indexed l2BlockNumber,
+        uint256 l1Timestamp
```
maurelian, line 65: I think `OutputDeleted` would be more descriptive now. 
Or keep the name, and make it 
`OutputsDeleted(uint256 indexed l2OutputIndex, uint256 numOutputsDeleted);

maurelian, line 65: I think `OutputDeleted` would be more descriptive now. 
Or keep the name, and make it 
`OutputsDeleted(uint256 indexed l2OutputIndex, uint256 numOutputsDeleted);

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -45,36 +46,31 @@ contract L2OutputOracle is Initializable, Semver {
...
     );
 
     /**
      * @notice Emitted when outputs are deleted.
      *
-     * @param l2BlockNumber First L2 block number deleted.
+     * @param l2OutputIndex First L2 output index deleted.
      */
-    event OutputsDeleted(uint256 indexed l2BlockNumber);
+    event OutputsDeleted(uint256 indexed l2OutputIndex);
```
maurelian, line 73: I think OutputDeleted would be more descriptive now.
Or keep the name, and make it
`OutputsDeleted(uint256 indexed l2OutputIndex, uint256 numOutputsDeleted);

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -204,59 +191,129 @@ contract L2OutputOracle is Initializable, Semver {
             // blockhash value, and delay submission until it is confident that the L1 block is
             // finalized.
             require(
-                blockhash(_l1BlockNumber) == _l1Blockhash,
+                blockhash(_l1BlockNumber) == _l1BlockHash,
                 "L2OutputOracle: blockhash does not match the hash at the expected height"
```
maurelian, line 195: Nit: The change of casing in the variable suggests that the error message should become
```suggestion
                "L2OutputOracle: block hash does not match the hash at the expected height"
```

maurelian (approved):
A couple nits, but LGTM.

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -204,59 +191,129 @@ contract L2OutputOracle is Initializable, Semver {
...
+     *
+     * @return The output at the given index.
      */
-    function getL2Output(uint256 _l2BlockNumber)
+    function getL2Output(uint256 _l2OutputIndex)
         external
         view
         returns (Types.OutputProposal memory)
     {
+        return l2Outputs[_l2OutputIndex];
```
tynes, line 223: Do we want our own error rather than `Panic(0x32)`?

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
tynes: from the pov of an offchain actor, getting `numOutputsDeleted` feels a bit less helpful than say getting the new height because getting the num outputs deleted requires context (the total length) to know what the new height is

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
smartcontracts: I think it's fine

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
smartcontracts: Did a different change but same idea

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -119,43 +116,36 @@ contract L2OutputOracle is Initializable, Semver {
...
-            _l2BlockNumber <= latestBlockNumber,
-            "L2OutputOracle: cannot delete outputs after the latest block number"
-        );
+        uint256 prevNextL2OutputIndex = nextOutputIndex();
 
-        // We're setting the latest block number back to the checkpoint block before the given L2
-        // block number. Next proposal will overwrite the deleted output and following proposals
-        // will delete any outputs after that.
-        latestBlockNumber = _l2BlockNumber - SUBMISSION_INTERVAL;
+        // Use assembly to delete the array elements because Solidity doesn't allow it.
```
trianglesphere, line 143: Is this less deleting the elements & more setting the length of the array?

smartcontracts: Yes, but it has the same result

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -119,43 +116,36 @@ contract L2OutputOracle is Initializable, Semver {
 
         startingTimestamp = _startingTimestamp;
```
tynes, line 117: Just realizing that we could pack `startinTimestamp` and `startingBlocknumber` into the same storage slot by making them uint128s, solidity will handle that automatically. I don't think it adds a ton of value really to be honest, just saves a storage slot

smartcontracts: Eh I don't think it really makes a difference

specs/proposals.md
```diff
@@ -124,7 +124,7 @@ function deleteL2Output(bytes32 _l2Output) external
 /**
```
tynes, line 124: We will probably need to go through the specs and update them more but that is fine to do in the future

tynes (approved):

PR #4094, opened 2022-11-28
op-chain-ops: remove setting owner in GovernanceToken

tynes:
**Description**

We do not want to set the `_owner` field in the `GovernanceToken`. This assumes that it is preset in the state, so it will not work for spinning up new networks.


tynes:
The question here is if we want to add the `governanceTokenOwner` to the config so that it can be set to a non zero value for alt networks. I think yes but I mostly would prefer moving the `MintManager` to being a predeploy to simplify this cc @smartcontracts @maurelian

PR #4095, opened 2022-11-28
feat(ctb): make L2OO addrs immutable

smartcontracts:
**Description**
Makes the L2OO proposer and challenger addresses immutable and gets rid of the concept of ownership in the L2OO.


tynes:
This looks like it depends on https://github.com/ethereum-optimism/optimism/pull/4093

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -26,6 +24,16 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
      */
     uint256 public immutable L2_BLOCK_TIME;
 
+    /**
```
tynes, line 27: We probably want to note that these can be changed by redeploying the contract and then changing the impl in the proxy

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -26,6 +24,16 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
      */
```
tynes, line 24: Is `L2_BLOCK_TIME` safe to change? I don't think so

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -97,36 +84,34 @@ contract L2OutputOracle is OwnableUpgradeable, Semver {
...
     }
 
     /**
      * @notice Initializer.
      *
      * @param _startingBlockNumber Block number for the first recoded L2 block.
      * @param _startingTimestamp   Timestamp for the first recoded L2 block.
-     * @param _proposer            The address of the proposer.
-     * @param _owner               The address of the owner.
      */
```
tynes, line 110: So much cleaner

packages/contracts-bedrock/contracts/test/L2OutputOracle.t.sol
```diff
@@ -190,12 +148,12 @@ contract L2OutputOracleTest is L2OutputOracle_Initializer {
      ***************************/
 
     // Test: proposeL2Output fails if called by a party that is not the proposer.
-    function testCannot_proposeL2OutputIfNotProposer() external {
+    function testCannot_proposeL2Outpu_ifNotProposer() external {
```
tynes, line 151: nit: typo, `proposeL2Output_if..`

op-chain-ops/genesis/layer_one.go
```diff
@@ -104,8 +104,6 @@ func BuildL1DeveloperGenesis(config *DeployConfig) (*core.Genesis, error) {
 		"initialize",
 		big.NewInt(0),
 		uint642Big(uint64(config.L1GenesisBlockTimestamp)),
-		config.L2OutputOracleProposer,
-		config.L2OutputOracleOwner,
```
tynes, line 108: This becomes a bit sticky because now the name of the config value should change, otherwise it is stale and can be confusing

tynes: If we change the config value name then its going to be a large diff

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
smartcontracts: It is not safe to change. I made a note of it in the other PR.

packages/contracts-bedrock/package.json
```diff
@@ -25,7 +25,7 @@
     "test": "yarn build:differential && forge test",
     "coverage": "yarn build:differential && forge coverage",
     "coverage:lcov": "yarn build:differential && forge coverage --report lcov",
-    "gas-snapshot": "forge snapshot --no-match-test 'differential|fuzz'",
+    "gas-snapshot": "yarn build:differential && forge snapshot --no-match-test 'differential|fuzz'",
```
maurelian, line 28: why do we need this back in?

smartcontracts: The finalizeWithdrawal tests rely on those utils existing, even for the non-differential tests. It's a pretty quick part of the task so not a big deal.

op-chain-ops/genesis/layer_one.go
smartcontracts: Ok I changed it

mslipper (approved):

smartcontracts:
@mergifyio requeue

PR #4103, opened 2022-11-29
op-node: Hardcode beta-1 config

mslipper:
Adds a hardcoded Beta-1 rollup config, and provides the plumbing for us to hardcode rollup configs in the future.

Fixes ENG-3053

### DIFF
--- a/op-node/chaincfg/chains.go
+++ b/op-node/chaincfg/chains.go
@@ -1,0 +1,61 @@
+package chaincfg
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var Beta1 = rollup.Config{
+	Genesis: rollup.Genesis{
+		L1: eth.BlockID{
+			Hash:   common.HexToHash("0x87ba22412f6d081a28ca0d8aafcf630d13a0d2fc16a7345eb3a0d5cd329f935e"),
+			Number: 7996739,
+		},
+		L2: eth.BlockID{
+			Hash:   common.HexToHash("0x76aac93f04b04b051c0232dc96cdfb2ebd150ce726aa9f776a4713c2ac524dc8"),
+			Number: 0,
+		},
+		L2Time: 1669088016,
+		SystemConfig: eth.SystemConfig{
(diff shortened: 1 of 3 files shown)


mslipper:
* develop
  * **PR #4103** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4103" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4103?utm_source=stack-comment).

trianglesphere (approved):
lgtm

op-node/chaincfg/chains.go
```diff
@@ -0,0 +1,49 @@
...
+			Number: 7996739,
+		},
+		L2: eth.BlockID{
+			Hash:   common.HexToHash("0x76aac93f04b04b051c0232dc96cdfb2ebd150ce726aa9f776a4713c2ac524dc8"),
+			Number: 0,
+		},
+		L2Time: 1669088016,
+		SystemConfig: eth.SystemConfig{
+			BatcherAddr: common.HexToAddress("0xc02551cde892e9716363b8e99d655298909e1a84"),
+			Overhead:    eth.Bytes32(common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000834")),
```
tynes, line 25: nit: overhead and scalar should be as human readable as possible here

op-node/flags/flags.go
```diff
@@ -33,6 +33,11 @@ var (
 		Usage:  "Rollup chain parameters",
 		EnvVar: prefixEnvVar("ROLLUP_CONFIG"),
 	}
+	Network = cli.StringFlag{
+		Name:   "network",
+		Usage:  "Predefined network selection",
```
tynes, line 38: We should list which possible choices there are here in the `Usage` field, it may get stale but its better ux than not knowing at all

tynes:
We should probably do the same thing for `op-geth` - wdyt @trianglesphere ?

mslipper:
We should do it for op-geth too, yes.

op-node/chaincfg/chains.go
mslipper: I think leaving this in hex is a bit clearer - otherwise, we'll have to either update the type to something numeric or convert the number into a hex-padded `bytes32`. Right now it matches the rollup config and can be easily verified.

mslipper:
@mergifyio refresh

PR #4112, opened 2022-11-30
maint(ctb): clean up deploy config

smartcontracts:
**Description**
Cleans up the deployment configuration structs for contracts-bedrock.


packages/contracts-bedrock/src/deploy-config.ts
```diff
@@ -0,0 +1,339 @@
...
+  l1GenesisBlockBaseFeePerGas: {
+    type: 'string', // uint256
+    default: ethers.BigNumber.from(1000_000_000).toHexString(), // 1 gwei
+  },
+  l2GenesisBlockNonce: {
+    type: 'string', // uint64
+    default: '0x0',
+  },
+  l2GenesisBlockExtraData: {
+    type: 'string', // important: in the case of L2, which uses post-Merge Ethereum rules, this must be <= 32 bytes.
```
tynes, line 276: If there was a `bytes32` type that would be helpful here but the comment is good enough

packages/contracts-bedrock/src/deploy-config.ts
```diff
@@ -0,0 +1,339 @@
...
+  p2pSequencerAddress: string
+
+  /**
+   * L1 address that batches are sent to.
+   */
+  batchInboxAddress: string
+
+  /**
+   * Acceptable batch-sender address, to filter transactions going into the batchInboxAddress on L1 for data.
+   * Warning: this address is hardcoded now, but is intended to become governed via L1.
```
tynes, line 70: I believe this is the initial batch sender address, used for deployment of the `SystemConfig`

packages/contracts-bedrock/src/deploy-config.ts
```diff
@@ -0,0 +1,339 @@
...
+  l2GenesisBlockExtraData: string
+  l2GenesisBlockGasLimit: string
+  l2GenesisBlockDifficulty: string
+  l2GenesisBlockMixHash: string
+  l2GenesisBlockCoinbase: string
+  l2GenesisBlockNumber: string
+  l2GenesisBlockGasUsed: string
+  l2GenesisBlockParentHash: string
+  l2GenesisBlockBaseFeePerGas: string
+  optimismBaseFeeRecipient: string
```
tynes, line 141: The recipients and owners are optional? I suppose `address(0)` will just be used

packages/contracts-bedrock/src/deploy-config.ts
```diff
@@ -0,0 +1,339 @@
...
+  optimismL1FeeRecipient: string
+  l2CrossDomainMessengerOwner: string
+  gasPriceOracleOverhead: number
+  gasPriceOracleScalar: number
```
tynes, line 145: Some of these values are technically optional but i'm not sure if they should be as it would degrade usage of the system

tynes (approved):

packages/contracts-bedrock/src/deploy-config.ts
smartcontracts: I can add that later

packages/contracts-bedrock/src/deploy-config.ts
smartcontracts: Yeah should probably be required but I can change that in another PR

packages/contracts-bedrock/src/deploy-config.ts
smartcontracts: I can make them required in another PR

PR #4114, opened 2022-11-30
feat: op-migrate docker image

tynes:
**Description**
```
Adds an `op-migrate` binary as well as a dockerfile for
building it. This `Dockerfile` should be accessed from
the root of the repository, similar to the other docker
files for the other bedrock packages.
```


op-chain-ops/docker.go.work
```diff
@@ -0,0 +1,6 @@
+go 1.18
+
+use (
+	./op-bindings
+    ./op-chain-ops
```
mslipper, line 5: [nit] indentation

mslipper (approved):

PR #4116, opened 2022-11-30
docs(ctb): Document the test naming convention

maurelian:
**Description**

This is a smaller version of #4084. 

This PR does not move or reorganize functions, it only renames them. 

It also:
1. Adds documentation for the naming convention to the README.
2. Adds a ts script which checks to ensure that tests conform to the convention. It can be run from within the contracts-bedrock package using `ts-node scripts/forges-test-names.ts`. 

This script should eventually be added to CI, once all function names have been updated.

Longer term the rules in the script should be added to a more robust set of internal rules, defined in slither or a similar tool.


packages/contracts-bedrock/contracts/test/AddressAliasHelper.t.sol
```diff
@@ -5,7 +5,7 @@ import { Test } from "forge-std/Test.sol";
 import { AddressAliasHelper } from "../vendor/AddressAliasHelper.sol";
 
 contract AddressAliasHelper_Test is Test {
-    function test_fuzz_roundtrip(address _address) external {
+    function testFuzz_applyAndUndo_succeeds(address _address) external {
```
maurelian, line 8: TBH I'm now leaning towards removing the `_succeeds` postfix, but I'd like to get these changes in as is, then we could easily cut it out in a subsequent PR.

smartcontracts:
I feel like this should be multiple PRs. Doc change itself could be one PR, then PR to add the script, then multiple PRs to update names to pass the script

maurelian:
@smartcontracts OK. 
See #4125 and #4126 for a start.

PR #4117, opened 2022-11-30
op-chain-ops: set initial message passer nonce

tynes:
**Description**

The `L2ToL1MessagePasser` nonce is set to `type(uint32).max` initially to prevent nonce collisions with legacy withdrawals that have been upgraded. The max value for the nonce is `type(uint256).max` so it is not a problem to start at a non zero nonce.


trianglesphere:
Changes lgtm. Once these types of changes are done it'd be nice to have a doc outlining them.

PR #4119, opened 2022-11-30
feat(sdk): support legacy withdraws after Bedrock

smartcontracts:
**Description**
Adds support for finalizing legacy withdrawals after the Bedrock migration. Cleans up some SDK code at the same time.


packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1296,92 +1393,23 @@ export class CrossChainMessenger {
       messageSlot
     )
 
-    // Sanity check that the value is set to 1 in the state
```
tynes, line 1299: Do you think this is a good sanity check? It would only break if we decided to start proving against a non bool in the state

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1261,33 +1369,22 @@ export class CrossChainMessenger {
...
+            withdrawal.sender,
+            withdrawal.target,
+            withdrawal.value,
+            withdrawal.minGasLimit,
+            withdrawal.message
+          ),
+        ]
+      )
     )
-    const isMessageSent =
```
tynes, line 1283: Do you think checking `isMessageSent` here is overkill?

tynes: I feel like it'll give a better error message than something downstream

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1261,33 +1369,22 @@ export class CrossChainMessenger {
...
-      resolved.transactionHash
-    )
-
-    const [withdrawal, withdrawalHash] = this.getWithdrawalFromReceipt(
-      receipt,
-      resolved
-    )
-
-    // Sanity check
-    if (withdrawal.MessagePassed.withdrawalHash !== withdrawalHash) {
```
tynes, line 1274: Slight preference to use the hash from the event instead of computing the hash client side, so that we can "just work" with protocol upgrades

tynes (approved):

tynes:
Looks like there is a problem in the codepath:

```
/home/circleci/project/packages/sdk/src/cross-chain-messenger.ts:419
      throw new Error(`can only convert L2 to L1 messages to low level`)
            ^
Error: can only convert L2 to L1 messages to low level
    at CrossChainMessenger.toLowLevelMessage (/home/circleci/project/packages/sdk/src/cross-chain-messenger.ts:419:13)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async CrossChainMessenger.getMessageStatus (/home/circleci/project/packages/sdk/src/cross-chain-messenger.ts:695:30)
    at async Timeout._onTimeout (/home/circleci/project/packages/sdk/tasks/deposit-erc20.ts:319:29)
```

tynes (changes requested):
test fail :(

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -356,6 +358,131 @@ export class CrossChainMessenger {
...
+        }
+      }
+
+      // Should not happen.
+      if (withdrawals.length === 0) {
+        throw new Error(`no withdrawals found in receipt`)
+      }
+
+      // TODO: Add support for multiple withdrawals.
+      if (withdrawals.length > 0) {
```
tynes, line 460: This should be `withdrawals.length > 1`

tynes:
Going to need to update this usage of `decodeVersionedNonce` as well: https://github.com/ethereum-optimism/optimism/blob/af9a5eed46cb45402c5894250c8688faaac691e9/packages/contracts-bedrock/scripts/differential-testing.ts#L25

packages/sdk/src/cross-chain-messenger.ts
smartcontracts: oops

mslipper (approved):

tynes (approved):

PR #4129, opened 2022-12-01
l2geth: cleanly shut down the syncservice

tynes:
**Description**

The ss was not previously being shut down cleanly, which can result in a corrupted db state. This makes the shutdown more clean.

### DIFF
--- a/l2geth/rollup/sync_service.go
+++ b/l2geth/rollup/sync_service.go
@@ -403,6 +403,7 @@ func (s *SyncService) IsSyncing() bool {
 // Stop will close the open channels and cancel the goroutines
 // started by this service.
 func (s *SyncService) Stop() error {
+	log.Info("Stopping sync service")
 	s.scope.Close()
 	s.chainHeadSub.Unsubscribe()
 	close(s.chainHeadCh)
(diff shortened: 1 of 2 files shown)


Inphi:
what's the extent of the db corruption? Does l2geth detect and fail to boot on a bad db?

mslipper:
> what's the extent of the db corruption? Does l2geth detect and fail to boot on a bad db?

We were seeing failures writing the chain tip hash, but this may have been a red herring. The node booted fine, but the migration failed since it needs the head hash to migrate the state.

Regardless, this code is helpful since it reduces the amount of time it takes for the node to shut down cleanly.

l2geth/rollup/sync_service.go
```diff
@@ -418,9 +419,15 @@ func (s *SyncService) VerifierLoop() {
...
+
+	for {
+		select {
+		case <-t.C:
+			if err := s.verify(); err != nil {
+				log.Error("Could not verify", "error", err)
+			}
+		case <-s.ctx.Done():
+			return
 		}
```
tynes, line 431: The call to `s.cancel` on line 411 will trigger the `Done` on 429

The problem that we faced was that the ticker polled again during shutdown and another round of `verify` happened while it was shutting down, resulting in the db being partially updated

trianglesphere (approved):

PR #4135, opened 2022-12-01
feat(ctb): add message versioning to MessagePasser

smartcontracts:
**Description**
Adds message versioning to the L2ToL1MessagePasser, using the same scheme as in the CrossDomainMessenger. Also renames the "nonce" variable to "messageNonce" in line with the naming within the CrossDomainMessenger contracts.


packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
```diff
@@ -96,7 +102,7 @@ contract L2ToL1MessagePasser is Semver {
     ) public payable {
         bytes32 withdrawalHash = Hashing.hashWithdrawal(
```
tynes, line 103: I think you should also add logic to `Hashing.hashWithdrawal` that observes the version

packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
```diff
@@ -107,9 +113,29 @@ contract L2ToL1MessagePasser is Semver {
 
         sentMessages[withdrawalHash] = true;
 
-        emit MessagePassed(nonce, msg.sender, _target, msg.value, _gasLimit, _data, withdrawalHash);
+        emit MessagePassed(
+            messageNonce(),
```
tynes, line 117: The migrated withdrawals will be a special case, where they do not have the version applied to the nonce, but they are hashing with version 1 style hashing

packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
smartcontracts: Feels unnecessary for now since we don't actually have a second version. It would be easy to add in the future if/when we have a second version.

packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol
smartcontracts: That's fine, we just say that version 0 and 1 have the same hashing scheme

tynes (approved):

tynes:
I don't think that there are any implications of this PR, but would like a review from @maurelian before merge just to get another set of eyes on this.

The contract didn't exist previously, there should be no nonce collisions with migrated withdrawals and the nonces should be guaranteed to be monotonic

maurelian (approved):

tynes:
@mergifyio requeue

PR #4138, opened 2022-12-01
ctb: Create Goerli deploy config

mslipper:
Creates a Goerli deploy config. Would like review from @trianglesphere, @tynes, and @smartcontracts on this. I'll respond in PR comments with specific values that give me pause.

### DIFF
--- a/packages/contracts-bedrock/deploy-config/goerli.json
+++ b/packages/contracts-bedrock/deploy-config/goerli.json
@@ -1,26 +1,35 @@
 {
-  "l1StartingBlockTag": "0xafce66a0a2446856112e4069b275ad32b1f4a607888f9c4c59eddf9be81f8670",
+  "finalSystemOwner": "DUMMY",
+  "controller": "DUMMY",
+
+  "l1StartingBlockTag": "DUMMY",
   "l1ChainID": 5,
-  "l2ChainID": 111,
+  "l2ChainID": 420,
   "l2BlockTime": 2,
 
-  "maxSequencerDrift": 1000,
-  "sequencerWindowSize": 120,
+  "maxSequencerDrift": 1200,
+  "sequencerWindowSize": 3600,
   "channelTimeout": 120,
-  "p2pSequencerAddress": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
-  "batchInboxAddress": "0xff00000000000000000000000000000000000002",
-  "batchSenderAddress": "0xa11d2b908470e17923fff184d48269bebbd9b2a5",
 
-  "systemConfigOwner": "0xa0Ee7A142d267C1f36714E4a8F75612F20a79720",
-
-  "l2OutputOracleSubmissionInterval": 6,
+  "p2pSequencerAddress": "DUMMY",
+  "batchInboxAddress": "0xff00000000000000000000000000000000000420",
+  "batchSenderAddress": "DUMMY",
+  "l2OutputOracleSubmissionInterval": 20,
   "l2OutputOracleStartingTimestamp": -1,
-  "l2OutputOracleProposer": "0x6c23a0dcdfc44b7a57bed148de598895e398d984",
-  "l2OutputOracleChallenger": "0x6925b8704ff96dee942623d6fb5e946ef5884b63",
+  "l2OutputOracleProposer": "DUMMY",
+  "l2OutputOracleChallenger": "DUMMY",
+  "finalizationPeriodSeconds": 2,
 
-  "baseFeeVaultRecipient": "0xf116a24056b647e3211d095c667e951536cdebaa",
-  "l1FeeVaultRecipient": "0xc731837b696ca3d9720d23336925368ceaa58f83",
-  "sequencerFeeVaultRecipient": "0xf116a24056b647e3211d095c667e951536cdebaa",
+  "systemConfigOwner": "DUMMY",
 
-  "deploymentWaitConfirmations": 1
+  "l2GenesisBlockGasLimit": "0x17D7840",
+  "l2GenesisBlockCoinbase": "0x4200000000000000000000000000000000000011",
+  "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
+
+  "l2CrossDomainMessengerOwner": "DUMMY",
+
+  "gasPriceOracleOverhead": 2100,
+  "gasPriceOracleScalar": 1000000,
+  "eip1559Denominator": 50,
+  "eip1559Elasticity": 10
 }


mslipper:
* develop
  * **PR #4138** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4138" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4138?utm_source=stack-comment).

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
   "channelTimeout": 120,
-  "p2pSequencerAddress": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
-  "batchInboxAddress": "0xff00000000000000000000000000000000000002",
-  "batchSenderAddress": "0xa11d2b908470e17923fff184d48269bebbd9b2a5",
-
-  "systemConfigOwner": "0xa0Ee7A142d267C1f36714E4a8F75612F20a79720",
 
-  "l2OutputOracleSubmissionInterval": 6,
+  "p2pSequencerAddress": "DUMMY",
+  "batchInboxAddress": "0xff00000000000000000000000000000000000000",
```
mslipper, line 15: This address can be anything - are we OK with making this `0xff00000000000000000000000000000000000000`?

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
-  "l2OutputOracleProposer": "0x6c23a0dcdfc44b7a57bed148de598895e398d984",
-  "l2OutputOracleChallenger": "0x6925b8704ff96dee942623d6fb5e946ef5884b63",
+  "l2OutputOracleProposer": "DUMMY",
+  "l2OutputOracleChallenger": "DUMMY",
+  "finalizationPeriodSeconds": 2,
+
+  "systemConfigOwner": "DUMMY",
+
+  "l2GenesisBlockGasLimit": "0x1C9C380",
+  "l2GenesisBlockCoinbase": "0x42000000000000000000000000000000000000f0",
```
mslipper, line 26: I pulled this address from one of our other configs, but I'm not sure where it comes from. What should this be set to? An address we generate, or the sequencer fee vault? @tynes

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
+  "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
+
+  "optimismBaseFeeRecipient": "DUMMY",
```
mslipper, line 29: Should this be the sequencer fee vault? @tynes

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
+  "optimismL1FeeRecipient": "DUMMY",
```
mslipper, line 30: Should this be the sequencer fee vault? @tynes

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
-  "baseFeeVaultRecipient": "0xf116a24056b647e3211d095c667e951536cdebaa",
-  "l1FeeVaultRecipient": "0xc731837b696ca3d9720d23336925368ceaa58f83",
-  "sequencerFeeVaultRecipient": "0xf116a24056b647e3211d095c667e951536cdebaa",
+  "l2CrossDomainMessengerOwner": "DUMMY",
 
-  "deploymentWaitConfirmations": 1
+  "gasPriceOracleOverhead": 2100,
+  "gasPriceOracleScalar": 1000000,
+  "gasPriceOracleDecimals": 6,
+  "eip1559Denominator": 50,
```
mslipper, line 37: @tynes @trianglesphere please verify that these 1559 params are OK.

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
   "l2OutputOracleStartingTimestamp": -1,
-  "l2OutputOracleProposer": "0x6c23a0dcdfc44b7a57bed148de598895e398d984",
-  "l2OutputOracleChallenger": "0x6925b8704ff96dee942623d6fb5e946ef5884b63",
+  "l2OutputOracleProposer": "DUMMY",
+  "l2OutputOracleChallenger": "DUMMY",
+  "finalizationPeriodSeconds": 2,
+
+  "systemConfigOwner": "DUMMY",
+
+  "l2GenesisBlockGasLimit": "0x1C9C380",
```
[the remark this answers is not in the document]
trianglesphere: We set the limit rather than the target here. With an elasticity of 10 a `30MM` limit implies at `3MM` target which is `1.5 mgas/sec`. I thought we wanted `1.25 mgas/sec` which is a limit of `25MM`, but a `30MM` limit is also fine.

packages/contracts-bedrock/deploy-config/goerli.json
trianglesphere: I think it's should be the fee vault b/c that's the address which will have all of the coinbase values in the future.

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
+  "optimismBaseFeeRecipient": "DUMMY",
+  "optimismL1FeeRecipient": "DUMMY",
 
-  "baseFeeVaultRecipient": "0xf116a24056b647e3211d095c667e951536cdebaa",
-  "l1FeeVaultRecipient": "0xc731837b696ca3d9720d23336925368ceaa58f83",
-  "sequencerFeeVaultRecipient": "0xf116a24056b647e3211d095c667e951536cdebaa",
+  "l2CrossDomainMessengerOwner": "DUMMY",
 
-  "deploymentWaitConfirmations": 1
+  "gasPriceOracleOverhead": 2100,
```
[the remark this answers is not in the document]
trianglesphere: I don't think we need the decimals any more

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
+
+  "l1StartingBlockTag": "DUMMY",
   "l1ChainID": 5,
-  "l2ChainID": 111,
+  "l2ChainID": 420,
   "l2BlockTime": 2,
 
-  "maxSequencerDrift": 1000,
-  "sequencerWindowSize": 120,
+  "maxSequencerDrift": 3600,
```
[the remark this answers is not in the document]
trianglesphere: Do you intend for the sequencer window size to be a day? Also, a sequencer drift of one hour (3600s) is very high.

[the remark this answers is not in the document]
mslipper: I thought a day would be reasonable in case of batcher problems, but I defer to you here. Can lower it if need be.

packages/contracts-bedrock/deploy-config/goerli.json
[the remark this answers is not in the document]
mslipper: Which value (`25MM` or `30MM`) do you think we should start with? Not married to either.

packages/contracts-bedrock/deploy-config/goerli.json
tynes: Its not really important what this value is, its an invalid state transition creating this block. I think that using the fee vault predeploy makes sense. We could also put graffiti in there

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
-  "l2OutputOracleChallenger": "0x6925b8704ff96dee942623d6fb5e946ef5884b63",
+  "l2OutputOracleProposer": "DUMMY",
+  "l2OutputOracleChallenger": "DUMMY",
+  "finalizationPeriodSeconds": 2,
+
+  "systemConfigOwner": "DUMMY",
+
+  "l2GenesisBlockGasLimit": "0x1C9C380",
+  "l2GenesisBlockCoinbase": "0x42000000000000000000000000000000000000f0",
+  "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
```
tynes, line 27: I am in favor of removing `l2GenesisBlockBaseFeePerGas` to simplify things, if the value is not sane then it will make eip1559 less stable after the migration block.

I propose we use this formula:

```
	gasUsed := (uint64)(config.L2GenesisBlockGasLimit) * config.EIP1559Elasticity
```

So that we hit this line here: https://github.com/ethereum-optimism/op-geth/blob/e383f9911586765c9792357da37a6c4020fcc7bf/consensus/misc/eip1559.go#L65

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: OK, will set it to sequencer fee vault.

packages/contracts-bedrock/deploy-config/goerli.json
tynes: The 3 values used are:

- `"baseFeeVaultRecipient"`
- `"l1FeeVaultRecipient"`
- `"sequencerFeeVaultRecipient"`

Each of them are L1 accounts that will receive the funds via a permissionless withdrawal. We should use accounts that we control

tynes: Meaning that `optimismBaseFeeRecipient` and `optimismL1FeeRecipient` are no longer valid config options

packages/contracts-bedrock/deploy-config/goerli.json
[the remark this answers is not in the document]
tynes: Decimals is now a protocol constant

tynes, line 34: In theory the overhead will be able to be much smaller, we never really did the work in computing the amount of gas each tx in a batch costs given an average batch size. This value is mutable and impacts end user gas prices, so we probably want to launch mainnet with the optimal value to pass savings on to users

packages/contracts-bedrock/deploy-config/goerli.json
maurelian: While it can be anything, I do think it should be consistent across systems. The monorepo has quite a mix of them in different configs, including: 

```
0xff00000000000000000000000000000000000002
0xff00000000000000000000000000000000000000
0x42000000000000000000000000000000000000ff
```

To minimize bikeshedding here, I'll strawman propose that we go with `0x42000000000000000000000000000000000000ff`, since I like keeping the `0x42` prefix as our calling card.

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
-  "batchInboxAddress": "0xff00000000000000000000000000000000000002",
-  "batchSenderAddress": "0xa11d2b908470e17923fff184d48269bebbd9b2a5",
-
-  "systemConfigOwner": "0xa0Ee7A142d267C1f36714E4a8F75612F20a79720",
 
-  "l2OutputOracleSubmissionInterval": 6,
+  "p2pSequencerAddress": "DUMMY",
+  "batchInboxAddress": "0xff00000000000000000000000000000000000000",
+  "batchSenderAddress": "DUMMY",
+  "l2OutputOracleSubmissionInterval": 20,
```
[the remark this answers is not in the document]
maurelian: Every 40 seconds. SGTM for a testnet.

packages/contracts-bedrock/deploy-config/goerli.json
maurelian: > I do think it should be consistent across systems

... or is it better to make them always unique as replay protection?

mslipper: I like keeping them different per network. How about we just use the chain ID?

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: Ok, will remove these.

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: OK. Is this good for now? If it's mutable then I think it's fine to start with this then change later.

packages/contracts-bedrock/deploy-config/goerli.json
maurelian: Chain ID would have to be the prefix to avoid precompiles. But that SGTM.

packages/contracts-bedrock/deploy-config/goerli.json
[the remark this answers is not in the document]
trianglesphere: I think I prefer 12 hours. When we did the opcraft testnet we had more issues with the super large sequencing window than we did with batch submission.

Also, I think 10-20 mins is a better range for the maxSequencerDrift (I prefer even lower - like 2x the seq conf depth, but that is pretty aggresive)

packages/contracts-bedrock/deploy-config/goerli.json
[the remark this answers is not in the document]
trianglesphere: 25MM. That's been the standard so far

packages/contracts-bedrock/deploy-config/goerli.json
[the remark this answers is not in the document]
mslipper: Cool, will update.

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: Hmm, I'm on the fence about this since it only matters for networks that we're migrating. Concerned that adding the formula will introduce hidden complexity.

mslipper: What do you think?

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: Put the chain ID at the end and went with `0xff00000000000000000000000000000000000420` to disambiguate from other `0x42` addresses.

packages/contracts-bedrock/deploy-config/goerli.json
tynes: Its good for now, we should be sure to measure on a live network and calculate what the mainnet value should be

packages/contracts-bedrock/deploy-config/goerli.json
tynes: I have a preference for using `params.InitialBaseFee` as the initial base fee across our networks to remove additional config that could be a footgun as done in https://github.com/ethereum-optimism/optimism/pull/4139/files

tynes: If you have a strong preference for manually configuring these values then we can do that, I have a preference to remove as much config as possible to reduce the chances of misconfiguration. I'd like the eip1559 stuff to be stable in the beginning

tynes: Either way, we will need to do the eip 1559 math under the hood with these values to ensure that its not extremely volatile at the start of the network

trianglesphere: why is it not the params.InitialBasefee? That seems like it should work automatically

packages/contracts-bedrock/deploy-config/goerli.json
maurelian, line 37: This is a constant value 

https://github.com/ethereum-optimism/optimism/blob/18928f9d3191e3f6f006b8464a02c3cb820ca589/packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol#L57

Is it being used anywhere?

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,26 +1,39 @@
...
 
-  "baseFeeVaultRecipient": "0xf116a24056b647e3211d095c667e951536cdebaa",
-  "l1FeeVaultRecipient": "0xc731837b696ca3d9720d23336925368ceaa58f83",
-  "sequencerFeeVaultRecipient": "0xf116a24056b647e3211d095c667e951536cdebaa",
+  "l2CrossDomainMessengerOwner": "DUMMY",
 
-  "deploymentWaitConfirmations": 1
+  "gasPriceOracleOverhead": 2100,
+  "gasPriceOracleScalar": 1000000,
+  "gasPriceOracleDecimals": 6,
```
maurelian, line 36: `gasPriceOracleDecimals` is a constant value in the contracts:

https://github.com/ethereum-optimism/optimism/blob/18928f9d3191e3f6f006b8464a02c3cb820ca589/packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol#L57

Is the config value actually being used anywhere?

maurelian: I see it has a default value here

[develop/packages/contracts-bedrock/hardhat.config.ts#L334](https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock/hardhat.config.ts#L334) perhaps it can eventually be used for post deployment validation?

mslipper: This is removed.

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: Ok, I'll remove this configuration in a subsequent PR. That'll span multiple projects so I don't want to do it here.

trianglesphere (approved):
lgtm

tynes (approved):

PR #4139, opened 2022-12-01
op-migrate: various sanity checks + cleanup

tynes:
**Description**

Adds cleanup to the migration script, ensures that config is sane.

We cannot allow for non monotonic block time. We also try to make eip 1159 stable at start and use a hardcoded initial base fee.

Add additional logging

### DIFF
--- a/op-chain-ops/genesis/db_migration.go
+++ b/op-chain-ops/genesis/db_migration.go
@@ -120,6 +120,9 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 	}
 	log.Info("Completed ERC20 ETH migration", "root", newRoot)
 
+	// Set the amount of gas used so that EIP 1559 starts off stable
+	gasUsed := (uint64)(config.L2GenesisBlockGasLimit) * config.EIP1559Elasticity
+
 	// Create the bedrock transition block
 	bedrockHeader := &types.Header{
 		ParentHash:  header.Hash(),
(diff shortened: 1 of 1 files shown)


op-chain-ops/genesis/db_migration.go
```diff
@@ -54,8 +54,21 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 		}, nil
 	}
 
-	if config.L2GenesisBlockBaseFeePerGas == nil {
-		return nil, errors.New("must configure L2 genesis block base fee per gas")
+	// Ensure EIP 1559 params are sane
+	if config.EIP1559Elasticity == 0 {
```
mslipper, line 58: Isn't there a `Check()` method for the deploy config? If so, should all these checks go in there?

op-chain-ops/genesis/db_migration.go
```diff
@@ -54,8 +54,21 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
...
+		return nil, errors.New("EIP 1559 Elasticity cannot be 0")
+	}
+	if config.EIP1559Denominator == 0 {
+		return nil, errors.New("EIP 1559 Denominator cannot be 0")
+	}
+	if config.L2GenesisBlockGasLimit == 0 {
+		return nil, errors.New("L2 genesis block gas limit cannot be 0")
+	}
+	// Ensure monotonic timestamps
+	if uint64(config.L2OutputOracleStartingTimestamp) <= header.Time {
```
mslipper, line 68: Nice catch!

mslipper (approved):
Dropped some non-blocking comments. Were you able to boot the sequencer from a migrated disk after making these changes?

op-chain-ops/genesis/db_migration.go
tynes: I'll consolidate things there, i was worried it would break other things

tynes:
> Dropped some non-blocking comments. Were you able to boot the sequencer from a migrated disk after making these changes?

Not yet, these are changes from statically analyzing the code, will continue trying the migration and will confirm when its successful

mslipper:
@mergifyio refresh

PR #4141, opened 2022-12-01
fix(ctb): various migration fixes

smartcontracts:
**Description**
Various fixes to the migration script as discovered during migration practice.

### DIFF
--- a/packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
+++ b/packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
@@ -5,6 +5,7 @@ import { L2OutputOracle } from "../L1/L2OutputOracle.sol";
 import { OptimismPortal } from "../L1/OptimismPortal.sol";
 import { L1CrossDomainMessenger } from "../L1/L1CrossDomainMessenger.sol";
 import { L1ChugSplashProxy } from "../legacy/L1ChugSplashProxy.sol";
+import { Proxy } from "../universal/Proxy.sol";
 import { ProxyAdmin } from "../universal/ProxyAdmin.sol";
 import { PortalSender } from "./PortalSender.sol";
 import { SystemConfig } from "../L1/SystemConfig.sol";
@@ -135,6 +136,11 @@ contract MigrationSystemDictator is BaseSystemDictator {
 
         // Transfer ownership of the L1StandardBridge to the ProxyAdmin.
         L1ChugSplashProxy(payable(config.proxyAddressConfig.l1StandardBridgeProxy)).setOwner(
+            address(config.globalConfig.proxyAdmin)
+        );
+
+        // Transfer ownership of the L1ERC721Bridge to the ProxyAdmin.
+        Proxy(payable(config.proxyAddressConfig.l1ERC721BridgeProxy)).changeAdmin(
             address(config.globalConfig.proxyAdmin)
         );
     }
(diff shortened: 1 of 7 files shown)


packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -69,38 +84,14 @@ export const deployAndVerifyAndThen = async ({
   await hre.ethers.provider.waitForTransaction(result.transactionHash)
 
   if (result.newlyDeployed) {
-    if (!(await isHardhatNode(hre)) && hre.network.config.live !== false) {
```
smartcontracts, line 72: Removing programmatic verification for now, because of the way the deployment works we can easily just verify after the fact. Breaks currently because we're using foundry under the hood.

packages/contracts-bedrock/deploy/018-MigrationSystemDictatorSteps.ts
```diff
@@ -298,10 +361,7 @@ const deployFn: DeployFunction = async (hre) => {
         await L1CrossDomainMessenger.xDomainMessageSender()
         assert(false, `L1CrossDomainMessenger was not initialized properly`)
       } catch (err) {
-        assert(
-          err.message.includes('xDomainMessageSender is not set'),
-          `L1CrossDomainMessenger was not initialized properly`
-        )
+        // Expected.
```
tynes, line 364: Perhaps log here?

packages/contracts-bedrock/src/deploy-utils.ts
tynes: there is a hh task that does the foundry verification with hh deploy

tynes (approved):
Generally looks good to me

packages/contracts-bedrock/deploy-config/goerli-forked.json
```diff
@@ -0,0 +1,29 @@
...
+  "sequencerWindowSize": 120,
+  "channelTimeout": 120,
+  "p2pSequencerAddress": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
+  "batchInboxAddress": "0xff00000000000000000000000000000000000002",
+  "batchSenderAddress": "0xa11d2b908470e17923fff184d48269bebbd9b2a5",
+
+  "systemConfigOwner": "0xa0Ee7A142d267C1f36714E4a8F75612F20a79720",
+
+  "l2OutputOracleSubmissionInterval": 6,
+  "l2OutputOracleStartingTimestamp": -1,
```
maurelian, line 20: THis is a uint256, how does `-1` work here?

packages/contracts-bedrock/contracts/deployment/MigrationSystemDictator.sol
```diff
@@ -273,6 +279,11 @@ contract MigrationSystemDictator is BaseSystemDictator {
             L1ChugSplashProxy(payable(config.proxyAddressConfig.l1StandardBridgeProxy)).setOwner(
                 address(config.globalConfig.finalOwner)
             );
+
+            // Transfer ownership of the L1ERC721Bridge to the final owner.
+            Proxy(payable(config.proxyAddressConfig.l1ERC721BridgeProxy)).changeAdmin(
+                address(config.globalConfig.finalOwner)
```
maurelian, line 285: SHouldn't the final owner be the proxy admin, which is owned by the final owner?

smartcontracts: This is an exit codepath that is meant to reset the system back to its original state. At this point the proxies have not yet been transferred to the proxy admin, so I figured it was best to return back to the original admin.

packages/contracts-bedrock/deploy-config/goerli-forked.json
smartcontracts: Special value, -1 is treated as "does not exist" but fwiw I don't even think this is being used directly anymore

smartcontracts:
@mergifyio requeue

maurelian (commented):
LGTM

PR #4163, opened 2022-12-02
Increase sequencer window & sequncer drift

trianglesphere:
**Description**

This should hopefully help some of the reorgs we've seen on the devnets.
I think the devnet was too faster for the sequencer window - particularly at startup.
In addition, I think the switch to using wallclock time may have aggravated the issue of the short sequencer window.

**Metadata**

- Fixes ENG-3036


tynes (approved):
Good looks

maurelian (approved):

mslipper (approved):

PR #4165, opened 2022-12-02
 CI: Echidna tests for Aliasing

maurelian:
This PR adds a simple echidna test which tests an address aliasing round trip.

It also does some basic scaffolding to lay the groundwork for forthcoming echidna tests.

PR #4166, opened 2022-12-02
ctb: Add echidna tests for Burn

maurelian:
Adds echidna fuzz tests for Burn.sol.


maurelian:
* develop
  * **PR #4165** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4165" target="_blank">Graphite</a> 
    * **PR #4166** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4166" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4166?utm_source=stack-comment).

packages/contracts-bedrock/contracts/echidna/FuzzBurn.sol
```diff
@@ -0,0 +1,59 @@
...
+     */
+    function testBurn(uint256 _value) public {
+        // cache the contract's eth balance
+        uint256 preBurnBalance = address(this).balance;
+
+        // execute a burn of _value eth
+        // (may way to add guardrails to this value rather than a truly unbounded uint256)
+        Burn.eth(_value);
+
+        // check that exactly _value eth was transfered from the contract
```
tynes, line 21: Do reverts for underflow not matter here?

maurelian: Oh, that's a great question. I think it's probably unreachable here, but regardless we should account for it.

maurelian: To address this I: 
1. Added a `bound()` to prevent overflow.
2. Put the invariant evaluation statements into `unchecked` blocks, and added an underflow check where necessary.

.circleci/config.yml
```diff
@@ -817,6 +822,15 @@ workflows:
           echidna_target: aliasing
           requires:
             - bedrock-echidna-build
+      - bedrock-echidna-run:
+          echidna_target: "burn:eth"
+          requires:
+            - bedrock-echidna-build
+      - bedrock-echidna-run:
+          echidna_target: "burn:gas"
+          size: 2xlarge
```
maurelian, line 831: the gas burn test seems to be a memory hog. This is the minimum size (32 gb) that will handle it.

packages/contracts-bedrock/contracts/echidna/FuzzBurn.sol
```diff
@@ -0,0 +1,71 @@
...
+     * @notice Takes an integer amount of eth to burn through the Burn library and
+     * updates the contract state if an incorrect amount of eth moved from the contract
+     */
+    function testBurn(uint256 _value) public {
+        // cache the contract's eth balance
+        uint256 preBurnBalance = address(this).balance;
+        uint256 value = bound(_value, 0, preBurnBalance);
+
+        // execute a burn of _value eth
+        Burn.eth(value);
```
smartcontracts, line 19: Not sure if this is easily testable here, but we *do* rely (in the actual contracts, not here) on the EVM behavior that `Burn.eth` does actually *burn* the ETH by removing it from the total supply rather than locking it into some other address.

packages/contracts-bedrock/contracts/echidna/FuzzBurn.sol
```diff
@@ -0,0 +1,71 @@
...
+    bool failedGasBurn;
+
+    /**
+     * @notice Takes an integer amount of gas to burn through the Burn library and
+     * updates the contract state if at least that amount of gas was not burned
+     * by the library
+     */
+    function testGas(uint256 _value) public {
+        // cap the value to the max resource limit
+        uint256 MAX_RESOURCE_LIMIT = 8_000_000;
```
smartcontracts, line 45: Implicit assumption here that the amount of gas available is at least 8m + some buffer to get to the end of execution safely.

packages/contracts-bedrock/contracts/echidna/FuzzBurn.sol
maurelian: yeah, I considered that but don't think we can test it. 
The best we could do, afaict, would be to change this function to return the address of `Burner`, then verify that that address is empty in a future block. But I don't think that's necessary.

https://github.com/ethereum-optimism/optimism/blob/0ac0098559bb4e918491af2bbc75adf1ddf1974a/packages/contracts-bedrock/contracts/libraries/Burn.sol#L14-L16

packages/contracts-bedrock/contracts/echidna/FuzzBurn.sol
maurelian: Well, I think the assumption is that it's at least `value` which is set below by modulating the argument into the range [0,MAX_RESOURCE_LIMIT].

Either way, if there's insufficient gas this whole thing will revert, which I think should be fine (echidna doesn't treat that as a failure).

smartcontracts (approved):

PR #4172, opened 2022-12-02
ctb: Add unit tests from SystemConfig

maurelian:
**Description**

Add System Config tests.

**Metadata**
- ENG-2949


mslipper:
Looks like we're missing the gas snapshot.

packages/contracts-bedrock/contracts/test/SystemConfig.t.sol
```diff
@@ -0,0 +1,117 @@
...
+
+        vm.expectEmit(true, true, true, true);
+        emit ConfigUpdate(
+            0,
+            SystemConfig.UpdateType.GAS_CONFIG,
+            abi.encode(newOverhead, newScalar)
+        );
+
+        vm.prank(alice);
+        sysConf.setGasConfig(newOverhead, newScalar);
```
clabby, line 87: Add check that the `overhead` + `scalar` was set correctly

```suggestion
        sysConf.setGasConfig(newOverhead, newScalar);
        assertEq(sysConf.overhead(), newOverhead);
        assertEq(sysConf.scalar(), newScalar);
```

clabby (changes requested):
Looks good. For completeness, may want to `vm.expectEmit` the `ConfigUpdate` event in the fuzz tests as well, but not necessary.

clabby (approved):

packages/contracts-bedrock/contracts/test/SystemConfig.t.sol
```diff
@@ -0,0 +1,94 @@
...
+
+contract SystemConfig_Initialize_TestFail is CommonTest {
+    function test_initialize_lowGasLimit_reverts() external {
+        vm.expectRevert("SystemConfig: gas limit too low");
+        new SystemConfig({
+            _owner: alice,
+            _overhead: 0,
+            _scalar: 0,
+            _batcherHash: bytes32(hex""),
+            _gasLimit: 7_999_999
```
tynes, line 29: I don't love how `7_999_999` is a magic value - it should have a comment as to how this value was picked

packages/contracts-bedrock/contracts/test/SystemConfig.t.sol
```diff
@@ -0,0 +1,94 @@
...
+
+    function testFuzz_setGasConfig_succeeds(uint256 newOverhead, uint256 newScalar) external {
+        vm.expectEmit(true, true, true, true);
+        emit ConfigUpdate(
+            0,
+            SystemConfig.UpdateType.GAS_CONFIG,
+            abi.encode(newOverhead, newScalar)
+        );
+
+        vm.prank(alice);
```
tynes, line 75: nit: `vm.prank(sysConf.owner())`

Makes it less contextual when reading and decouples the deploy constructor args

tynes (approved):

smartcontracts (approved):

packages/contracts-bedrock/contracts/test/SystemConfig.t.sol
maurelian: good call

packages/contracts-bedrock/contracts/test/SystemConfig.t.sol
maurelian: ok fixed.

PR #4191, opened 2022-12-02
[op-node] Set fallback UDP port

mdehoog:
**Description**

`op-node`'s `p2p.advertise.udp` flag says that it will fallback on `p2p.listen.udp` if not set. However in practice, this does not happen.

This PR adds this fallback.

See https://github.com/ethereum-optimism/optimism/blob/cba069eff6d8951ec3d58f72c089a0bb0b8e1f2d/op-node/flags/p2p_flags.go#L83

**Tests**

I've tested this behavior in an internal devnet. It causes the advertised ENR address to have a UDP key/value pair, as verified by https://enr-viewer.com/.


trianglesphere (approved):
Thanks for finding & fixing this. The test failure looks like a flake. I'll re-run it.

PR #4201, opened 2022-12-04
op-heartbeat: Project setup

mslipper:
Project setup (go.work, go.sum, etc.) for `op-heartbeat`. `op-heartbeat` is the server component that processes heartbeats from the `op-node`. This is used to keep track of which software versions are currently being used on the network. We've had the client side of this implemented in the `op-node` for a while, but never implemented the server. This PR kicks that off so that we can have all this in place once we migrate Goerli.

PR #4200, opened 2022-12-04
op-service: Exit HTTP server more cleanly

mslipper:
Updates `ListenAndServeContext` to prevent a goroutine leak and return `nil` if the context is canceled.

PR #4202, opened 2022-12-04
op-heartbeat: Add implementation

mslipper:
Adds implementation files for `op-heartbeat`. Will add CI integration in a separate PR. Notes:

- The implementation is pretty simple. All it does is receive heartbeats and log their fields to stdout.
- The server uses Prometheus metrics to keep track of the number of heartbeats keyed on chain ID and op-node version. This lets us see how quickly different versions of our software are adopted in production.
- To prevent DDoS attacks on our Prometheus metrics, a whitelist is used to restrict the number of chain IDs and versions that will be created as metrics.
- This package uses `op-service`, so it keeps track of HTTP metrics and support pprof automatically.

We can add additional functionality later; my goal here is to have something basic we can use to track version usage as the number of nodes on the network grows.

Heartbeats are totally optional, and can be disabled by setting `OP_NODE_HEARTBEAT_ENABLED` to false.

### DIFF
--- a/op-heartbeat/flags/flags.go
+++ b/op-heartbeat/flags/flags.go
@@ -1,0 +1,44 @@
+package flags
+
+import (
+	opservice "github.com/ethereum-optimism/optimism/op-service"
+	oplog "github.com/ethereum-optimism/optimism/op-service/log"
+	opmetrics "github.com/ethereum-optimism/optimism/op-service/metrics"
+	"github.com/urfave/cli"
+)
+
+const envPrefix = "OP_HEARTBEAT"
+
+const (
+	HTTPAddrFlagName        = "http.addr"
+	HTTPPortFlagName        = "http.port"
+	HTTPMaxBodySizeFlagName = "http.max-body-size"
+	AllowedChainIDsFlagName = "allowed-chain-ids"
+)
+
+var (
+	HTTPAddrFlag = cli.StringFlag{
+		Name:   HTTPAddrFlagName,
+		Usage:  "Address the server should listen on",
+		Value:  "0.0.0.0",
+		EnvVar: opservice.PrefixEnvVar(envPrefix, "HTTP_ADDR"),
+	}
+	HTTPPortFlag = cli.IntFlag{
+		Name:   HTTPPortFlagName,
+		Usage:  "Port the server should listen on",
+		Value:  8080,
+		EnvVar: opservice.PrefixEnvVar(envPrefix, "HTTP_PORT"),
+	}
+)
+
+var Flags []cli.Flag
+
+func init() {
+	Flags = []cli.Flag{
+		HTTPAddrFlag,
+		HTTPPortFlag,
+	}
+
+	Flags = append(Flags, oplog.CLIFlags(envPrefix)...)
+	Flags = append(Flags, opmetrics.CLIFlags(envPrefix)...)
+}
--- a/op-heartbeat/metrics.go
+++ b/op-heartbeat/metrics.go
@@ -1,0 +1,62 @@
+package op_heartbeat
+
+import (
+	"strconv"
+
+	"github.com/ethereum-optimism/optimism/op-node/heartbeat"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const MetricsNamespace = "op_heartbeat"
+
+type Metrics interface {
+	RecordHeartbeat(payload heartbeat.Payload)
+	RecordVersion(version string)
+}
+
+type metrics struct {
+	heartbeats *prometheus.CounterVec
+	version    *prometheus.GaugeVec
+}
+
+func NewMetrics(r *prometheus.Registry) Metrics {
+	m := &metrics{
+		heartbeats: promauto.With(r).NewCounterVec(prometheus.CounterOpts{
+			Namespace: MetricsNamespace,
+			Name:      "heartbeats",
+			Help:      "Counts number of heartbeats by chain ID",
+		}, []string{
+			"chain_id",
+			"version",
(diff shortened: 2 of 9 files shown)


mslipper:
* develop
  * **PR #4200** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4200" target="_blank">Graphite</a> 
    * **PR #4201** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4201" target="_blank">Graphite</a> 
      * **PR #4202** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4202" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4202?utm_source=stack-comment).

tynes (approved):

trianglesphere (approved):
lgtm. Do you want to add a codeowners for the `op-hearbeat` package?

op-heartbeat/service.go
```diff
@@ -0,0 +1,122 @@
...
+	return httputil.ListenAndServeContext(ctx, server)
+}
+
+func Handler(l log.Logger, maxBodySize int, metrics Metrics) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		innerL := l.New(
+			"xff", r.Header.Get("X-Forwarded-For"),
+			"user_agent", r.Header.Get("User-Agent"),
+			"remote_addr", r.RemoteAddr,
+		)
```
protolambda, line 98: Note to self: we're logging user-input here, but it should be safe, the type is a string and it'll be escaped when encoded for terminal/json. Correct?

op-heartbeat/service.go
```diff
@@ -0,0 +1,122 @@
...
+	metrics := NewMetrics(registry)
+	metrics.RecordVersion(version)
+	handler := Handler(l, cfg.HTTPMaxBodySize, metrics)
+	recorder := opmetrics.NewPromHTTPRecorder(registry, MetricsNamespace)
+	mw := opmetrics.NewHTTPRecordingMiddleware(recorder, handler)
+
+	server := &http.Server{
+		Addr:           net.JoinHostPort(cfg.HTTPAddr, strconv.Itoa(cfg.HTTPPort)),
+		MaxHeaderBytes: cfg.HTTPMaxBodySize,
+		Handler:        mw,
```
protolambda, line 86: We should set `WriteTimeout`, `IdleTimeout`, and `ReadTimeout` (and optionally `ReadHeaderTimeout`), or document that it needs to be put behind some proxy with proper timeouts.

op-heartbeat/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+	} else {
+		chainID = "unknown"
+	}
+	var version string
+	if AllowedVersions[payload.Version] {
+		version = payload.Version
+	} else {
+		version = "unknown"
+	}
+	m.heartbeats.WithLabelValues(chainID, version).Inc()
```
protolambda, line 57: This metric will be far off if one of the nodes starts spamming us at a different interval. Should we use the sender address data (or `X-Forwarded-For` if we use a proxy) to filter, maybe with an LRU of the last 10000 seen IPs, with their last seen timestamp, to check we're not getting spammed?

op-heartbeat/flags/flags.go
```diff
@@ -0,0 +1,51 @@
...
+		Usage:  "Port the server should listen on",
+		Value:  8080,
+		EnvVar: opservice.PrefixEnvVar(envPrefix, "HTTP_PORT"),
+	}
+	HTTPMaxBodySizeFlag = cli.IntFlag{
+		Name:   HTTPMaxBodySizeFlagName,
+		Usage:  "Maximum body size that will be accepted, in bytes",
+		Value:  1024 * 1024,
+		EnvVar: opservice.PrefixEnvVar(envPrefix, "HTTP_MAX_BODY_SIZE"),
+	}
```
protolambda, line 37: Does this really have to be a CLI flag instead of a constant? The data will always be just some version info and a chain-id, right?

op-heartbeat/service.go
```diff
@@ -0,0 +1,122 @@
...
+
+func Handler(l log.Logger, maxBodySize int, metrics Metrics) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		innerL := l.New(
+			"xff", r.Header.Get("X-Forwarded-For"),
+			"user_agent", r.Header.Get("User-Agent"),
+			"remote_addr", r.RemoteAddr,
+		)
+
+		defer r.Body.Close()
```
estensen, line 100: It's not necessary to close the request body

> 	// For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
https://pkg.go.dev/net/http#Request

op-heartbeat/metrics.go
mslipper: Mind if I add this separately? Don't want to make this too complex right now.

op-heartbeat/flags/flags.go
mslipper: Sure, I can simplify to just a constant.

op-heartbeat/service.go
mslipper: It'll be behind a proxy regardless (since it will live in k8s), but I've updated this to add these nonetheless.

trianglesphere:
@mslipper this branch needs an update to fix the devnet tests

mslipper:
Done

protolambda (approved):
LGTM, and I can help pick up the spam issue thing

PR #4206, opened 2022-12-05
feat(ctb): put dictator behind Proxy

smartcontracts:
**Description**
Puts the SystemDictator behind a standard Proxy contract. By putting the SystemDictator behind a proxy, we can work our way out of pretty much any error type. The owner of the proxy will be the controller as defined in the deploy config.


packages/contracts-bedrock/deploy/019-SystemDictatorInit.ts
```diff
@@ -0,0 +1,133 @@
...
+  getContractsFromArtifacts,
+} from '../src/deploy-utils'
+
+const deployFn: DeployFunction = async (hre) => {
+  const { deployer } = await hre.getNamedAccounts()
+
+  let controller = hre.deployConfig.controller
+  if (controller === ethers.constants.AddressZero) {
+    if (hre.network.config.live === false) {
+      console.log(`WARNING!!!`)
```
tynes, line 19: Perhaps a `printWarning(meta: string)` function would be useful

packages/contracts-bedrock/deploy/019-SystemDictatorInit.ts
```diff
@@ -0,0 +1,133 @@
...
+    await assertDictatorConfig(SystemDictator, config)
+  }
+
+  // Update the owner if necessary.
+  if (
+    (await SystemDictatorProxy.callStatic.admin({
+      from: ethers.constants.AddressZero,
+    })) !== controller
+  ) {
+    // Transfer ownership to the controller address.
```
tynes, line 113: A logline would be nice here

tynes:
Looks like something is wrong based on the failing devnet with deployed contracts

```
Error: cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ] (reason="execution reverted: L1ChugSplashProxy: implementation is not set yet", method="estimateGas", transaction={"from":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","to":"0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0","data":"0x58a997f6000000000000000000000000a85233c63b9ee964add6f2cffe00fd84eb32338f0000000000000000000000007c6b91d9be155a6db01f749217d76ff02a7227f20000000000000000000000000000000000000000000000000de0b6b3a76400000000000000000000000000000000000000000000000000000000000000030d4000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000000","accessList":null}, error={"name":"ProviderError","code":3,"_isProviderError":true,"data":"0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000304c314368756753706c61736850726f78793a20696d706c656d656e746174696f6e206973206e6f74207365742079657400000000000000000000000000000000"}, code=UNPREDICTABLE_GAS_LIMIT, version=providers/5.7.1)

```

Digging in:

> execution reverted: L1ChugSplashProxy: implementation is not set yet

Looking at the calldata

```bash
$ cast 4byte 0x58a997f6
depositERC20(address,address,uint256,uint32,bytes)
```

tynes:
Looks like the `L1StandardBridge` is no longer being configured correctly

packages/contracts-bedrock/deploy/019-SystemDictatorInit.ts
smartcontracts: Can do in another PR

tynes (approved):

PR #4211, opened 2022-12-05
Update op-geth

trianglesphere:
**Description**

Updates op-geth to a migration rehearsal ready version.
I had to modify the genesis creation tools to set the bedrock block number in the chain config for tests to pass as if the optimism config is set, the bedrock block should also
be set.


trianglesphere:
> It looks like somehow `eth_estimateGas` is trying to be forwarded to a historical node when it shouldn't be
> 
> > WARN [12-05|19:16:34.829] Served eth_estimateGas                   reqid=46 duration="39.395µs"  err="no historical RPC is available for this historical (pre-bedrock) execution request"

This was because the `optimism config` was being set inside the chain config, but the bedrock block was not. I added a fix for it & I'm re-running the tests

op-chain-ops/genesis/genesis.go
```diff
@@ -53,6 +53,7 @@ func NewL2Genesis(config *DeployConfig, block *types.Block) (*core.Genesis, erro
 		CancunBlock:                   nil,
 		TerminalTotalDifficulty:       big.NewInt(0),
 		TerminalTotalDifficultyPassed: true,
+		BedrockBlock:                  new(big.Int).SetUint64(uint64(config.L2GenesisBlockNumber)),
```
tynes, line 56: Hmm bedrock block isn't under `Optimism`?

tynes (approved):

op-chain-ops/genesis/genesis.go
trianglesphere: no, it's not. Easier to get to & it matches the other forks better.

protolambda: TBH I like it better to group optimism configurables, it'll make things more maintainable as diff. The consistency with other fork block numbers is nice too though.

mslipper (approved):

PR #4215, opened 2022-12-05
Set bedrock block in migration

trianglesphere:
**Description**

This sets the bedrock fork block number in the stored chain config during the migration.

**Tests**

Not tested.

**Metadata**

- Fixes ENG-3076

### DIFF
--- a/op-chain-ops/genesis/db_migration.go
+++ b/op-chain-ops/genesis/db_migration.go
@@ -200,6 +200,7 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 		EIP1559Denominator: config.EIP1559Denominator,
 		EIP1559Elasticity:  config.EIP1559Elasticity,
 	}
+	cfg.BedrockBlock = bedrockBlock.Number()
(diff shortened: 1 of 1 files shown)


tynes:
Good catch

tynes (approved):

tynes:
If the bedrock block is unset, i feel like `op-geth` should do no forwarding backwards

trianglesphere:
> If the bedrock block is unset, i feel like `op-geth` should do no forwarding backwards

`op-geth` does forwarding if the `optimism` config is active, and the specific block is not a bedrock block. Maybe it should default to bedrock if it's not set, but we'll have it set in the migration & set it at genesis.

mslipper (approved):

tynes:
> > If the bedrock block is unset, i feel like `op-geth` should do no forwarding backwards
> 
> `op-geth` does forwarding if the `optimism` config is active, and the specific block is not a bedrock block. Maybe it should default to bedrock if it's not set, but we'll have it set in the migration & set it at genesis.

I feel like `op-geth` should be able to act exactly like `geth` if the optimism config isn't set

PR #4229, opened 2022-12-05
op-migration: add extra L2OutoutOracle timestamp check

tynes:
**Description**

Errors if the L2OutoutOracle starting timestamp is greater than the L1 block time that is used as the initial L1 origin.

### DIFF
--- a/op-chain-ops/genesis/db_migration.go
+++ b/op-chain-ops/genesis/db_migration.go
@@ -67,6 +67,15 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 		)
 	}
 
+	if uint64(config.L2OutputOracleStartingTimestamp) > l1Block.Time() {
+		return nil, fmt.Errorf(
+			"L2OutputOracleStartingTimestamp (%d) cannot be greater than the L1 block time (%d)",
+			config.L2OutputOracleStartingTimestamp,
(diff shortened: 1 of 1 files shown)


op-chain-ops/genesis/db_migration.go
```diff
@@ -67,6 +67,15 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 		)
 	}
 
+	if uint64(config.L2OutputOracleStartingTimestamp) > l1Block.Time() {
```
trianglesphere, line 70: Is the L2 Output Oracle timestamp the time of the transition blocK? If so this check should be done the other way (transition block timestamp must be >= l1Block timestamp)

tynes: I think this is underspecified - I'm not 100% sure what makes the most sense here. I thought that the deploy param `startingL2Timestamp` for the `L2OutputOracle` should be the same value as the timestamp in the genesis block

tynes: This leads me to believe we shouldn't even have `L2OutputOracleStartingTimestamp` as part of the config

trianglesphere: We could check how it is used in the contract & see if we can relax the constraints there & pull from the L1 block time the L2 genesis time

PR #4230, opened 2022-12-05
op-chain-ops: update script to submit pending withdrawals

tynes:
**Description**

An old an unused script for fetching withdrawals
is updated to instead submit pending withdrawals.

Going to update this to:
- check the execution trace of the withdrawal after it was finalized

Closes ENG-2740


op-chain-ops/cmd/withdrawals/main.go
```diff
@@ -32,64 +43,244 @@ func main() {
...
+			// Get the finalization period
+			finalizationPeriod, err := portal.FINALIZATIONPERIODSECONDS(&bind.CallOpts{})
+			if err != nil {
+				return err
+			}
+
+			future := block.Time() + finalizationPeriod.Uint64() + 1
+			var result bool
+			// TODO: double check this is the correct RPC
+			err = l1RpcClient.Call(&result, "hardhat_setTimestamp", future)
```
tynes, line 254: We can remove this now

tynes: The finalization period should be 2 seconds, we should be able to just wait for it

PR #4232, opened 2022-12-05
contracts-bedrock: add post check L2 hh task

tynes:
**Description**

This task will check that the L2 node was brought up as expected. It can be ran with the following command:

```
$ npx hardhat check-l2 --network $NETWORK
```

The `devnetL2` hardhat network is added to the hardhat config so that this can be developed against a local network very easily.

Closes ENG-3091


packages/contracts-bedrock/tasks/check-l2.ts
```diff
@@ -0,0 +1,628 @@
...
+  // - is behind a proxy
+  // - owner is `address(0)`
+  DeployerWhitelist: async (hre: HardhatRuntimeEnvironment) => {
+    const DeployerWhitelist = await hre.ethers.getContractAt(
+      'DeployerWhitelist',
+      predeploys.DeployerWhitelist
+    )
+
+    const version = await DeployerWhitelist.version()
+    assertSemver(version, 'DeployerWhitelist')
```
smartcontracts, line 128: Nitpick but you could just pass the contract instance in here to avoid having to do `await contract.version()` every time

packages/contracts-bedrock/tasks/check-l2.ts
```diff
@@ -0,0 +1,628 @@
...
+    const OTHER_BRIDGE = await L2StandardBridge.OTHER_BRIDGE()
+    if (OTHER_BRIDGE === hre.ethers.constants.AddressZero) {
+      throw new Error('invalid OTHER_BRIDGE')
+    }
+    logLoud()
+    console.log(`  - OTHER_BRIDGE: ${OTHER_BRIDGE}`)
+    logLoud()
+
+    const MESSENGER = await L2StandardBridge.MESSENGER()
+    if (MESSENGER !== predeploys.L2CrossDomainMessenger) {
```
smartcontracts, line 262: You could use the `assertContractVariable` helper function in `src/deploy-utils` to simplify a few things

packages/contracts-bedrock/tasks/check-l2.ts
```diff
@@ -0,0 +1,628 @@
...
+  // - check REMOTE_TOKEN
+  // - totalSupply should be set to 0
+  LegacyERC20ETH: async (hre: HardhatRuntimeEnvironment) => {
+    const LegacyERC20ETH = await hre.ethers.getContractAt(
+      'LegacyERC20ETH',
+      predeploys.LegacyERC20ETH
+    )
+
+    const name = await LegacyERC20ETH.name()
+    if (name !== 'Ether') {
```
smartcontracts, line 365: General nitpick that node's built-in `assert` is perfect for this sort of stuff. `assert(name === 'Ether', 'name mismatch')`. Although for contract variables specifically you can use `assertContractVariable`.

packages/contracts-bedrock/tasks/check-l2.ts
```diff
@@ -0,0 +1,628 @@
...
+    if (symbol !== 'WETH') {
+      throw new Error('symbol misconfigured')
+    }
+    console.log(`  - symbol: ${symbol}`)
+
+    const decimals = await WETH9.decimals()
+    if (decimals !== 18) {
+      throw new Error('decimals misconfigured')
+    }
+    console.log(`  - decimals: ${decimals}`)
```
smartcontracts, line 424: Mind if we also log the total supply?

packages/contracts-bedrock/tasks/check-l2.ts
```diff
@@ -0,0 +1,628 @@
...
+    const ProxyAdmin = await hre.ethers.getContractAt(
+      'ProxyAdmin',
+      predeploys.ProxyAdmin
+    )
+
+    const owner = await ProxyAdmin.owner()
+    if (owner === hre.ethers.constants.AddressZero) {
+      throw new Error('misconfigured owner')
+    }
+    logLoud()
```
smartcontracts, line 524: Nitpick, based on the way `logLoud` is being used, you could have a function `yell` that looks like:

```ts
const yell = (msg: string) => {
   logLoud()
   console.log(msg)
   logLoud()
}
```

packages/contracts-bedrock/tasks/check-l2.ts
tynes: require vs custom errors lol

packages/contracts-bedrock/tasks/check-l2.ts
tynes: This is a nice util but i need to log the values

tynes:
I've addressed all feedback on this PR

tynes:
This script is currently failing due to some misconfiguration, https://github.com/ethereum-optimism/optimism/pull/4392 fixes the problem

mslipper (approved):

PR #4238, opened 2022-12-06
op-node: Replace interface{} with any

estensen:
**Description**

Replace interface{} with any to make code more readable. Shorter lines is nice too!

**Additional context**

Go itself [did this a while back](https://github.com/golang/go/commit/2580d0e08d5e9f979b943758d3c49877fb2324cb)

### DIFF
--- a/op-node/client/polling.go
+++ b/op-node/client/polling.go
@@ -78,7 +78,7 @@ func (w *PollingClient) Close() {
 	w.c.Close()
 }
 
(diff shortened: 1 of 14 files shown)


tynes:
Generally looks good to me, requiring a newer go version is no problem

estensen:
`go.mod` is already pinning version 1.18, which introduced the alias, so don't think it's necessary to use a newer version of Go

protolambda (approved):

PR #4263, opened 2022-12-06
contracts-bedrock: delete dead code

tynes:
**Description**

The `GasPriceOracle` contract no longer uses the events that were defined in it. This commit deletes the events and adds a comment noting that they existed before bedrock.


maurelian (approved):

packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
```diff
@@ -56,21 +62,6 @@ contract GasPriceOracle is Semver {
...
 
-    /**
-     * @notice Emitted when the overhead value is updated.
-     */
-    event OverheadUpdated(uint256 overhead);
-
-    /**
-     * @notice Emitted when the scalar value is updated.
-     */
-    event ScalarUpdated(uint256 scalar);
```
smartcontracts, line 67: Hmmm this is a weird case. It changes the ABI, meaning that if you want to scan for the legacy events then you need to use a legacy package that exports the correct ABI. I'm not super opinionated on this though. I lean on the side of deleting this code since it isn't being used anymore.

smartcontracts (approved):

packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
tynes: Acknowledging your thoughts, going to resolve this comment thread so this PR auto merges

PR #4275, opened 2022-12-06
Update withdrawal params for two step withdrawals

trianglesphere:
**Description**

This renames the L2 block field to the L2 Index. This more accurately matches what the L2 output oracle contract expects.

**Tests**

Tested in hive here: https://github.com/ethereum-optimism/hive/pull/66

**Additional context**

I think we need a design doc on the API design of the proof withdrawal utilities. I expect that the API might change again - having to pass in the contract is a little bit ugly.

Once this is merged, I'll have to cut a release `v0.10.3` and then update hive with that release.


tynes:
> I think we need a design doc on the API design of the proof withdrawal utilities. I expect that the API might change again - having to pass in the contract is a little bit ugly.

I agree with this, the sdk has a stable interface for typescript and the existing go utils aren't working well for my usecase of submitting migrated withdrawals

mslipper (approved):

PR #4276, opened 2022-12-06
Version binaries

trianglesphere:
**Description**

This adds two layers of versioning to builds. The first is providing git information to docker builds. We already pulled the information into the build via make & ldflags, but did not have it available because we did not have the `.git` folder inside docker.

The second is setting a reminder in the `tag-bedrock-go-modules.sh` script to set the version in the `op-node`, `op-batcher`, & `op-proposer`. This should try to correspond to the tag version. This might be able to be scripted, but scripting it is hard, so I chose to prompt the developer.

**Metadata**
- Fixes ENG-3081

### DIFF
--- a/ops/scripts/tag-bedrock-go-modules.sh
+++ b/ops/scripts/tag-bedrock-go-modules.sh
@@ -36,6 +36,9 @@ go get github.com/ethereum-optimism/optimism/op-service@$VERSION
 go get github.com/ethereum-optimism/optimism/op-chain-ops@$VERSION
 go mod tidy
 
+echo Please update the version to ${VERSION} in op-node/version/version.go
+read -p "Press [Enter] key to continue"
+
 git add .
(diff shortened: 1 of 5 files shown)


ops/scripts/tag-bedrock-go-modules.sh
```diff
@@ -36,6 +36,9 @@ go get github.com/ethereum-optimism/optimism/op-service@$VERSION
 go get github.com/ethereum-optimism/optimism/op-chain-ops@$VERSION
 go mod tidy
```
tynes, line 38: Should we make a script that autogens `version.go`?

trianglesphere: Probably, but we use a different type of versionning between op-node and op-proposer/op-batcher

zhwrd (approved):

protolambda (approved):

mslipper (approved):

PR #4280, opened 2022-12-07
feat(indexer): Support OP Portal's `WithdrawalProven` event

clabby:
# Overview

Adds support for indexing the `OptimismPortal`'s `WithdrawalProven`
event added in two step withdrwals.

*Metadata*
ENG-3075

### DIFF
--- a/indexer/db/db.go
+++ b/indexer/db/db.go
@@ -198,7 +198,12 @@ func (d *Database) AddIndexedL1Block(block *IndexedL1Block) error {
 		($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)
 	`
 
-	const updateWithdrawalStatement = `
+	const updateProvenWithdrawalStatement = `
+	UPDATE withdrawals SET (br_withdrawal_proven_tx_hash, br_withdrawal_proven_log_index) = ($1, $2)
+	WHERE br_withdrawal_hash = $3
+	`
+
+	const updateFinalizedWithdrawalStatement = `
 	UPDATE withdrawals SET (br_withdrawal_finalized_tx_hash, br_withdrawal_finalized_log_index, br_withdrawal_finalized_success) = ($1, $2, $3)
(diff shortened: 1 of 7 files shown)


indexer/db/db.go
```diff
@@ -198,7 +198,12 @@ func (d *Database) AddIndexedL1Block(block *IndexedL1Block) error {
 		($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)
 	`
 
-	const updateWithdrawalStatement = `
+	const updateProvenWithdrawalStatement = `
+	UPDATE withdrawals SET (br_withdrawal_proven_tx_hash, br_withdrawal_proven_log_index) = ($1, $2)
+	WHERE br_withdrawal_hash = $3
+	`
```
roninjin10, line 204: Will this automatically include joins on l1_blocks and l1_tokens?

indexer/db/withdrawal.go
```diff
@@ -40,6 +40,8 @@ type WithdrawalJSON struct {
 	TxHash                   string          `json:"transactionHash"`
 	Batch                    *StateBatchJSON `json:"batch"`
 	BedrockWithdrawalHash    *string         `json:"bedrockWithdrawalHash"`
+	BedrockProvenTxHash      *string         `json:"bedrockProvenTxHash"`
+	BedrockProvenLogIndex    *int            `json:"bedrockProvenLogIndex"`
```
roninjin10, line 44: Noob question, what is a log index?

roninjin10 (approved):
LGTM just a couple questions for my own understanding

indexer/db/withdrawal.go
clabby: The log index is the index of the event within the block- for ex., if there are 3 events emitted in the block and the `ProvenWithdrawal` is the 3rd, the log index would be `2`.

mslipper (approved):

PR #4284, opened 2022-12-07
op-wheel,op-service: create op-wheel, and extend op-service base functionality

protolambda:
**Description**

`op-wheel` is a new module that defines a CLI tool, used for Bedrock migration testing and debugging.

Now you can take the steering wheel of an ethereum execution engine, without CL, and apply cheats to the DB.

First PR in a stack of 4: base, cheats-logic, engine-logic, commands-frontend:
- #4284
- #4285
- #4286
- #4287

This PR also adds some minor additions to the op-service module, to help implement the later commands with:
- `CloseAction`: to define a CLI action function that respects signals and can close a background thing gracefully.
- `ReadLocalCLIConfig` variant of `ReadCLIConfig` for metrics and logging packages, since not all sub-commands need the metrics/logging functionality, but to share the same base functionality per sub-command where necessary.

**Metadata**

Fix ENG-3077

PR #4285, opened 2022-12-07
op-wheel: geth db cheats logic

protolambda:
**Description**

This PR adds a `cheats` package to `op-wheel` with all the backend logic for the cheating commands: a common pattern to open the database, open a blockchain/state around that, and to modify the head block with state mutations, and several state modifications that we later expose with commands.

Second PR in a stack of 4: base, cheats-logic, engine-logic, commands-frontend:
- #4284
- #4285
- #4286
- #4287

**Tests**

This is a testing tool used in the migration testing. I'm open to testing the backend functionality in CI somewhere, but don't like to cascade the testing with testing of the testing.

**Metadata**

Fix ENG-3078

### DIFF
--- a/op-wheel/cheat/cheat.go
+++ b/op-wheel/cheat/cheat.go
@@ -1,0 +1,340 @@
+package cheat
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"math/big"
+	"path/filepath"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/consensus/beacon"
+	"github.com/ethereum/go-ethereum/consensus/ethash"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/trie"
+)
+
+var HundredETH = big.NewInt(0).Mul(big.NewInt(100), big.NewInt(1000000000000000000))
+
+type Cheater struct {
+	// The database of the chain with the head block that we patch the state-root of, once the state is updated.
+	DB ethdb.Database
+	// Initialized chain, wrapping the database with in-memory presentation of headers and recent changes and such.
+	Blockchain *core.BlockChain
+	// The Cheater avoids making writes if this is set to True, and opens the DB as readonly.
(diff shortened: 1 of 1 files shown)


op-wheel/cheat/cheat.go
```diff
@@ -0,0 +1,340 @@
...
+	DB ethdb.Database
+	// Initialized chain, wrapping the database with in-memory presentation of headers and recent changes and such.
+	Blockchain *core.BlockChain
+	// The Cheater avoids making writes if this is set to True, and opens the DB as readonly.
+	ReadOnly bool
+}
+
+func OpenGethRawDB(dataDirPath string, readOnly bool) (ethdb.Database, error) {
+	// don't use readonly mode in actual DB, it doesn't work with Geth.
+	db, err := rawdb.NewLevelDBDatabaseWithFreezer(dataDirPath, 2048, 500, filepath.Join(dataDirPath, "ancient"), "", readOnly)
```
trianglesphere, line 38: That's a pretty high number of db handles and I think it's over the default limit on mac osx. I also don't think you get much more performance by having so many.

mslipper (approved):

mslipper:
Adding mergify ignore so @trianglesphere  can take a look

trianglesphere (approved):
One worry about db handles, but otherwise looks good

op-wheel/cheat/cheat.go
tynes: Perhaps using the default values of what geth uses makes sense

op-wheel/cheat/cheat.go
```diff
@@ -0,0 +1,340 @@
...
+
+	// Flush the whole batch into the disk, exit the node if failed
+	if err := batch.Write(); err != nil {
+		_ = ch.Close()
+		return fmt.Errorf("failed to update chain indexes and markers: %w", err)
+	}
+	// Technically there are more in-memory things to update in real geth,
+	// to which we don't even have public API access, but that's fine, we're done.
+	// *And we did update the finalized marker, which is flushed from memory to disk on shutdown in geth.
+	// bc.hc.SetCurrentHeader(block.Header())
```
tynes, line 147: Did you mean to leave these commented out?

op-wheel/cheat/cheat.go
trianglesphere: Geth uses `MakeDatabaseHandles` to actually request more handles.

protolambda: I found the 500 as default somewhere in geth, but in other places it's dynamic yes. I'll see if I can adjust it, or make geth pick the number.

trianglesphere: On macs you can raise the limit to almost anything IIRC, but the default limit is like 256

mslipper:
Geth issue is an unrelated flake. Merging.

PR #4286, opened 2022-12-07
op-wheel: execution engine logic

protolambda:
**Description**

This PR adds a `engine` package to `op-wheel` with all the backend logic for the engine commands: a common engine API dialing function with authentication, and functions to read or modify the engine API.

Third PR in a stack of 4: base, cheats-logic, engine-logic, commands-frontend:
- #4284
- #4285
- #4286
- #4287

**Tests**

This is a testing tool used in the migration testing. I'm open to testing the backend functionality in CI somewhere, but don't like to cascade the testing with testing of the testing.

**Metadata**

Fix #ENG-3084

### DIFF
--- a/op-wheel/engine/engine.go
+++ b/op-wheel/engine/engine.go
@@ -1,0 +1,278 @@
+package engine
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


trianglesphere (approved):
lgtm. If we want to test this, I think we could unit test it, but probably shouldn't do integration tests for this.

tynes (approved):

PR #4289, opened 2022-12-07
feat(ctb): use typescript config files

smartcontracts:
**Description**
Updates our deploy config files to TypeScript so we get type safety.


tynes:
This will break a test in `op-chain-ops`

roninjin10:
If we need json and want typesafety the way to get both is to create a json config, and export it

```
import jsonConfig from './devnetL1.json'

export default jsonConfig satisfies DeployConfig
```

smartcontracts:
> If we need json and want typesafety the way to get both is to create a json config, and export it
> 
> ```
> import jsonConfig from './devnetL1.json'
> 
> export default jsonConfig satisfies DeployConfig
> ```

Oh that's sick

PR #4290, opened 2022-12-07
fix(ctb): clean and fix test names script

smartcontracts:
**Description**
Cleans up the test names script and fixes a few errors.

### DIFF
--- a/packages/contracts-bedrock/scripts/forge-test-names.ts
+++ b/packages/contracts-bedrock/scripts/forge-test-names.ts
@@ -1,120 +1,86 @@
 import fs from 'fs'
-import path from 'path'
 
-type Check = {
-  check: (name: string, parts: string[]) => void
-  makeError: (name: string) => Error
-}
+import { glob } from 'glob'
 
 /**
- * Function name checks
+ * Series of function name checks.
  */
-const FunctionNameChecks: { [name: string]: Check } = {
-  invalidCase: {
-    check: (name: string, parts: string[]): void => {
-      parts.forEach((part) => {
-        if (part[0] !== part[0].toLowerCase()) {
-          throw FunctionNameChecks.invalidCase.makeError(name)
-        }
+const checks: Array<{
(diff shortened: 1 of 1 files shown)


packages/contracts-bedrock/scripts/forge-test-names.ts
```diff
@@ -1,120 +1,86 @@
...
-      ) {
-        throw FunctionNameChecks.noFailureReason.makeError(name)
-      }
+  {
+    error:
+      'failure tests should have 4 parts, third part should indicate the reason for failure',
+    check: (parts: string[]): boolean => {
+      return (
+        parts.length !== 4 ||
+        !['reverts', 'fails'].includes(parts[parts.length - 1])
```
maurelian, line 52:
```suggestion
      'failure tests should have 4 parts, third part should indicate the reason for failure',
    check: (parts: string[]): boolean => {
      return (
        parts.length !== 4 &&
        !['reverts', 'fails'].includes(parts[parts.length - 1])
```

maurelian, line 52:
```suggestion
        parts.length !== 4 &&
        !['reverts', 'fails'].includes(parts[parts.length - 1])
```

maurelian, line 52: I think this should be `&&`. If the second holds, then the first must as well.

maurelian (changes requested):
Local testing flagged pretty much every function with `failure tests should have 4 parts, third part should indicate the reason for failure`. I think I identified the bug here.

packages/contracts-bedrock/scripts/forge-test-names.ts
maurelian, line 52: Fixes a bug and more readable IMO
```suggestion
        !['reverts', 'fails'].includes(parts[parts.length - 1]) ||
        parts.length === 4
```

PR #4297, opened 2022-12-07
maint: update TypeScript version

smartcontracts:
**Description**
Updates TypeScript to the latest version so we get access to the satisfies keyword.


tynes:
I tried this awhile ago and it was painful, kept having build issues

tynes (approved):

maurelian:
> I tried this awhile ago and it was painful, kept having build issues

FWIW, the build worked for me.

mslipper (approved):

PR #4300, opened 2022-12-07
feat(ctb): type safety for deploy config

smartcontracts:
**Description**
Includes new minimal TypeScript files that guarantee type safety for deployment configuration files. I also had to update some eslint dependencies so that `eslint` and `prettier` wouldn't complain about the new `satisfies` keyword.


packages/contracts-bedrock/deploy-config/hivenet.json
```diff
@@ -1 +0,0 @@
-{}
```
smartcontracts, line 1: @mslipper said this was unused

packages/contracts-bedrock/package.json
```diff
@@ -19,7 +19,7 @@
     "build:differential": "tsc scripts/differential-testing.ts --outDir dist --moduleResolution node --esModuleInterop",
     "prebuild": "yarn ts-node scripts/verify-foundry-install.ts",
     "build": "hardhat compile && yarn autogen:artifacts && yarn build:ts && yarn typechain",
-    "build:ts": "tsc -p tsconfig.json",
+    "build:ts": "tsc -p tsconfig.build.json",
```
tynes, line 22: Why build.json here?

smartcontracts: If we don't include the scripts/deploy-config/tasks folders in tsconfig then vscode shows errors and it also will not automatically lint your files. However, if you do include those files then they will be compiled and included in the final build. The solution is to use two different files, one that gets used during development and includes all of your folders and one that gets used during the build and only included your source folder.

mslipper (approved):

maurelian (approved):

PR #4306, opened 2022-12-07
op-chain-ops: single database commit

tynes:
**Description**

Instead of having multiple statedbs created,
use a single statedb and make sure that it is
only committed once. Various APIs are required
for the full migration which wrap the low level
LevelDB at various abstractions. This will ensure
that changes to the statedb are committed to disk.

### DIFF
--- a/op-chain-ops/genesis/db_migration.go
+++ b/op-chain-ops/genesis/db_migration.go
@@ -227,7 +227,30 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 		"hash", bedrockHeader.Hash().String(),
 	)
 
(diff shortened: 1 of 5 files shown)


op-chain-ops/genesis/db_migration.go
```diff
@@ -216,9 +227,32 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 		"hash", bedrockHeader.Hash().String(),
 	)
 
+	postDB, err := state.New(newRoot, underlyingDB, nil)
```
tynes, line 230: Unsure if `underlyingDB` is safe to use here or not

trianglesphere (approved):
looks fine to me

PR #4316, opened 2022-12-08
op-chain-ops: Fix validation logic

mslipper:
Updaets `op-chain-ops` to correctly check for predeploy address implementations. Previously, every predeploy address in the namespace was being checked for an implementation. Only a select few of these will have implementations at genesis, so I pulled that logic out and instead iterated over the known predeploys to make sure _those_ have impementations. 
Additionally, I updated the logic to exclude certain checks for "special" predeploys like OVM ETH, the proxy admin, and the governance token.

### DIFF
--- a/op-chain-ops/genesis/check.go
+++ b/op-chain-ops/genesis/check.go
@@ -42,10 +42,12 @@ func CheckMigratedDB(ldb ethdb.Database) error {
 	if err := CheckPredeploys(db); err != nil {
 		return err
 	}
+	log.Info("checked predeploys")
 
 	if err := CheckLegacyETH(db); err != nil {
 		return err
 	}
+	log.Info("checked legacy eth")
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


op-chain-ops/genesis/check.go
```diff
@@ -64,24 +66,36 @@ func CheckPredeploys(db vm.StateDB) error {
...
-		if len(implCode) == 0 {
-			return fmt.Errorf("no code found at predeploy impl %s", addr)
-		}
-
 		// There must be an admin
 		admin := db.GetState(addr, AdminSlot)
 		adminAddr := common.BytesToAddress(admin.Bytes())
-		if adminAddr != predeploys.ProxyAdminAddr {
-			return fmt.Errorf("admin is %s when it should be % for %s", adminAddr, predeploys.ProxyAdminAddr, addr)
+		if addr != predeploys.ProxyAdminAddr && addr != predeploys.GovernanceTokenAddr && adminAddr != predeploys.ProxyAdminAddr {
```
protolambda, line 72: Sanity check: is it correct to not also skip `LegacyERC20ETHAddr` here? What's the exact difference with the `special` cases in `mapImplementationAddress`?

protolambda (commented):
LGTM, but commented with a sanity check about one line, I'm lacking context here.

tynes (approved):

op-chain-ops/genesis/check.go
mslipper: This loop only iterates over `0x42...0` - `0x42..0800`. The legacy ERC20 address is in `0xDead`, so won't be included.

mslipper: The `special` cases are those that aren't mapped to a proxy.

op-chain-ops/genesis/setters.go
```diff
@@ -136,3 +128,23 @@ func SetPrecompileBalances(db vm.StateDB) {
...
+	switch *addrP {
+	case predeploys.GovernanceTokenAddr:
+		addr = predeploys.GovernanceTokenAddr
+		special = true
+	case predeploys.LegacyERC20ETHAddr:
+		addr = predeploys.LegacyERC20ETHAddr
+		special = true
+	case predeploys.ProxyAdminAddr:
+		addr = predeploys.ProxyAdminAddr
+		special = true
```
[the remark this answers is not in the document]
tynes: this is true

tynes:
Lets get this merged so that we have everything ready for the next iteration

op-chain-ops/genesis/setters.go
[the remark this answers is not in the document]
mslipper: will write a unit test + fix nit later

PR #4326, opened 2022-12-08
fix: migration gaslimit

tynes:
**Description**

Set the withdrawal migration gas limit to
a non zero number.

### DIFF
--- a/op-chain-ops/crossdomain/migrate.go
+++ b/op-chain-ops/crossdomain/migrate.go
@@ -64,6 +64,8 @@ func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger *com
 	}
 
 	versionedNonce := EncodeVersionedNonce(withdrawal.Nonce, common.Big1)
+	// Encode the call to `relayMessage` on the `CrossDomainMessenger`.
+	// The minGasLimit can safely be 0 here.
 	data, err := abi.Pack(
 		"relayMessage",
 		versionedNonce,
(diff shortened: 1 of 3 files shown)


op-chain-ops/crossdomain/migrate.go
```diff
@@ -77,12 +79,15 @@ func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger *com
...
+	// Set the outer gas limit. This cannot be zero
+	gasLimit := uint64(len(data)*16 + 200_000)
+
 	w := NewWithdrawal(
 		withdrawal.Nonce,
 		&predeploys.L2CrossDomainMessengerAddr,
 		l1CrossDomainMessenger,
 		value,
-		new(big.Int),
+		new(big.Int).SetUint64(gasLimit),
```
trianglesphere, line 90: Is this the gas limit that is eventually passed to call which will be executed on L1?

tynes: Yup exactly

trianglesphere: Got it. I also just saw the doc you wrote up

op-chain-ops/crossdomain/migrate.go
```diff
@@ -77,12 +79,15 @@ func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger *com
 		return nil, fmt.Errorf("cannot abi encode relayMessage: %w", err)
 	}
 
+	// Set the outer gas limit. This cannot be zero
+	gasLimit := uint64(len(data)*16 + 200_000)
```
protolambda, line 83: Would be good to document this + determine if it's a sufficient gas limit. LGTM otherwise.

tynes: We have a notion doc explaining the rationale, the plan is to merge and then test e2e

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -351,10 +351,12 @@ export class CrossChainMessenger {
       }
     }
 
+    const minGasLimit = resolved.message.length * 16 + 200_000
```
smartcontracts, line 354: Message here isn't a buffer. You would need to convert to buffer first before getting length.

smartcontracts: Should be able to use `fromHexString`

smartcontracts (approved):

PR #4327, opened 2022-12-08
[RFC] Bedrock: rename "optimism" ENR key to "opstack"

mdehoog:
**Description**

Currently `op-node`'s P2P implementation uses a key called `optimism` to store the node's chain ID and version in its ENR address. However, there's nothing specific about optimism in this key to warrant the key name.

This PR genericizes the key to `opstack`.

**Tests**

This is an RFC; happy to add tests if there is alignment on this.

**Additional context**

This is primarily to make Bedrock more forkable / configurable. Some other ideas: make this key configurable via an op-node CLI parameter, or split into two separate ENR keys: `chainid` and `version`.

Some other potential areas for more generic naming are:
 - The `optimism` key used in the [L2 genesis configuration](https://github.com/ethereum-optimism/op-geth/commit/68bfa8b473bb8770216517f007b521fab41c5afa#diff-10ff00a15ef58da7485fedeca3906c73cb236fde4b143a9d61601c63cbf1ffe8R422). Perhaps could be `rollup` or similar.
 - The `optimism` key used in as the [prefix for the RPCs](https://github.com/ethereum-optimism/optimism/blob/8b70a0f3f51dc25af59ef500a4b138c9314ae02d/op-node/node/server.go#L38) on `op-node`. Could be made configurable, with `optimism` as the default.

### DIFF
--- a/op-node/p2p/discovery.go
+++ b/op-node/p2p/discovery.go
@@ -216,7 +216,7 @@ func FilterEnodes(log log.Logger, cfg *rollup.Config) func(node *enode.Node) boo
 		err := node.Load(&dat)
 		// if the entry does not exist, or if it is invalid, then ignore the node
 		if err != nil {
-			log.Debug("discovered node record has no optimism info", "node", node.ID(), "err", err)
+			log.Debug("discovered node record has no chain info", "node", node.ID(), "err", err)
 			return false
 		}
 		// check chain ID matches
--- a/op-node/p2p/discovery.go
+++ b/op-node/p2p/discovery.go
@@ -75,7 +75,7 @@ func (conf *Config) Discovery(log log.Logger, rollupCfg *rollup.Config, tcpPort 
 	} else {
 		return nil, nil, fmt.Errorf("no TCP port to put in discovery record")
 	}
-	dat := ChainENRData{
+	dat := OpStackENRData{
 		chainID: rollupCfg.L2ChainID.Uint64(),
 		version: 0,
 	}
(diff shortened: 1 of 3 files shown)


tynes:
Generally not opposed to this

smartcontracts:
Also not opposed, generally quite happy to accept PRs to make the stack more forkable. Will leave review to code owners though.

protolambda:
With more generic key-value pair names I'm afraid it collides with other chains or L1, and that we run into compatibility issues and parsing errors. The discv5 DHT is shared with all other discv5-using projects.

mdehoog:
In this case, collisions probably don't matter as we filter on the chainId and version in the value, and not just the key. The key is just used for lookup.

protolambda (changes requested):
- `chain` may be used by other protocols for discv5 tags already, we do not want to overlap with values with a different meaning.
- Our network specs need to be updated if we are going to change this.
- Comment in `DecodeRLP` still refers to an "optimism" tag, and that we might add more information later, which wouldn't work if we try to be as generic as any chain.

I'm open to renaming it to `opstack` or something that at least isolates it to the optimism tech domain to avoid compatibility issues with non-opstack networks. Wdyt?

tynes:
> * `chain` may be used by other protocols for discv5 tags already, we do not want to overlap with values with a different meaning.
> * Our network specs need to be updated if we are going to change this.
> * Comment in `DecodeRLP` still refers to an "optimism" tag, and that we might add more information later, which wouldn't work if we try to be as generic as any chain.
> 
> I'm open to renaming it to `opstack` or something that at least isolates it to the optimism tech domain to avoid compatibility issues with non-opstack networks. Wdyt?

+1 to `OpStack`

mdehoog:
sgtm!

tynes:
I believe this is a breaking change for the P2P network, we should just be mindful when merging it

protolambda:
@tynes it is, but it won't break anything as we currently don't have a hard dependency on discv5, we primarily use static peering currently anyway.

protolambda (approved):

tynes:
> @tynes it is, but it won't break anything as we currently don't have a hard dependency on discv5, we primarily use static peering currently anyway.

Ok i'm down to merge this whenever. One thing we don't have is automatic generation of changelogs, that would be a nice addition to our release process

PR #4329, opened 2022-12-08
feat(ctb): update portal gas amount

smartcontracts:
**Description**
Updates the amount of gas that can be sent to the target contract when sending transactions through the Optimism portal. Previously, the amount was specified as exactly the transaction gas limit. We now allow the withdrawal to be sent with more than the transaction gas limit.

Also updates the CrossDomainMessenger to include a new special codepath for gas estimation.


tynes:
Companion PR to https://github.com/ethereum-optimism/optimism/pull/4326

clabby (approved):

tynes (approved):

PR #4333, opened 2022-12-09
fix(cmn): BaseServiceV2 options parsing bug

smartcontracts:
**Description**
Fixes a bug in BaseServiceV2 where options would not be parsed correctly when passed directly into the constructor rather than via the environment or command line arguments.

### DIFF
--- a/packages/common-ts/src/base-service/base-service-v2.ts
+++ b/packages/common-ts/src/base-service/base-service-v2.ts
@@ -274,16 +274,12 @@ export abstract class BaseServiceV2<
     // names into lower case for the validation step. We'll turn the names back into their original
     // names when we're done.
     const cleaned = cleanEnv<TOptions>(
-      { ...config.env, ...config.args },
+      { ...config.env, ...config.args, ...(params.options || {}) },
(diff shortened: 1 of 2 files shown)


packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -274,17 +274,13 @@ export abstract class BaseServiceV2<
...
-      { ...config.env, ...config.args },
+      { ...config.env, ...config.args, ...(params.options || {}) },
       Object.entries(params.optionsSpec || {}).reduce((acc, [key, val]) => {
         acc[key.toLowerCase()] = val.validator({
           desc: val.desc,
           default: val.default,
         })
         return acc
-      }, {}) as any,
-      Object.entries(params.options || {}).reduce((acc, [key, val]) => {
```
smartcontracts, line 285: I misunderstood the interface of `cleanEnv`, the "options" being passed in here are options for the behavior of `cleanEnv` not the actual options that you want to clean. Just added the options field to the first param and it worked.

roninjin10 (approved):

tynes (approved):

smartcontracts:
@mergifyio requeue

PR #4334, opened 2022-12-09
feat(cmn): make logLevel a default option

smartcontracts:
**Description**
Updates BaseServiceV2 so that logLevel is a default option to any new instantiation of BaseServiceV2. Avoids people having to include hacky code to set the log level.

### DIFF
--- a/packages/common-ts/src/base-service/validators.ts
+++ b/packages/common-ts/src/base-service/validators.ts
@@ -13,6 +13,8 @@ import { Provider } from '@ethersproject/abstract-provider'
 import { Signer } from '@ethersproject/abstract-signer'
 import { ethers } from 'ethers'
 
+import { LogLevel, logLevels } from '../common'
+
 const provider = makeValidator<Provider>((input) => {
   const parsed = url()._parse(input)
   return new ethers.providers.JsonRpcProvider(parsed)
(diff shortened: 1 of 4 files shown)


packages/common-ts/src/base-service/validators.ts
```diff
@@ -39,6 +41,14 @@ const wallet = makeValidator<Signer>((input) => {
   }
 })
 
+const logLevel = makeValidator<LogLevel>((input) => {
+  if (!logLevels.includes(input as LogLevel)) {
+    throw new Error(`expected log level to be one of ${logLevels}`)
+  } else {
+    return input as LogLevel
+  }
+})
```
roninjin10, line 50: This is way cleaner with zod.  
Zod is generally useful.  Anytime you are taking unvalidated input and trying to validate it, zod is the best way to do it.   
``` typescript
const logLevelValidator =  z.union([
  z.literal('trace'), 
  z.literal('debug'),
  z.literal('info'),
  z.literal('warn'),
  z.literal('error'),
  z.literal('fatal'),
]);
// logLevel is typesafe now and will throw an error if it's not right type
const logLevel = logValidator.parse(input)

```

packages/common-ts/src/base-service/validators.ts
```diff
@@ -39,6 +41,14 @@ const wallet = makeValidator<Signer>((input) => {
   }
 })
 
+const logLevel = makeValidator<LogLevel>((input) => {
+  if (!logLevels.includes(input as LogLevel)) {
+    throw new Error(`expected log level to be one of ${logLevels}`)
```
roninjin10, line 46:
```suggestion
    throw new Error(`expected log level to be one of ${logLevels.join(', ')}`)
```

roninjin10 (approved):

packages/common-ts/src/base-service/validators.ts
smartcontracts: Interesting, gotta check that out for later

tynes (approved):

PR #4337, opened 2022-12-09
feat(ctb): add gas estimation logic to portal

smartcontracts:
**Description**
Adds the same special gas estimation logic that we just added to the CrossDomainMessenger into the OptimismPortal.


packages/contracts-bedrock/contracts/libraries/Constants.sol
```diff
@@ -0,0 +1,18 @@
...
+    /**
+     * @notice Special address to be used as the tx origin for gas estimation calls in the
+     *         OptimismPortal and CrossDomainMessenger calls. You only need to use this address if
+     *         the minimum gas limit specified by the user is not actually enough to execute the
+     *         given message and you're attempting to estimate the actual necessary gas limit. We
+     *         use address(1) because it's the ecrecover precompile and therefore guaranteed to
+     *         never have any code on any EVM chain.
+     */
+    address internal constant ESTIMATION_ADDRESS = address(1);
+}
```
maurelian, line 18: Kind of surprised we didn't already have this?

smartcontracts: Yeah. I'm gonna also make a PR to put `DEFAULT_L2_SENDER` in here.

maurelian (approved):

PR #4340, opened 2022-12-09
feat(ctb): Optimize `Bytes` library

clabby:
# Overview

Optimizes the `Bytes` library in the Bedrock contracts.

**Rationale**

The `Bytes` library is commonly used in inherently expensive functions such as the `MerkleTrie`'s `get`. As it stands, the implementation of the `Bytes` lib's `toNibbles` function is unoptimized. 

### Info for @maurelian 

**Gas Comparison**

With the Yul implementation, the `OptimismPortal`'s `proveWithdrawalTransaction` function saves $25,925$ gas ( $\Delta = -15.5$% ) with the inputs supplied in its [benchmark test](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/contracts/test/BenchmarkTest.t.sol#L116-L123).
```diff
- GasBenchMark_OptimismPortal:test_proveWithdrawalTransaction_benchmark() (gas: 167214)
+ GasBenchMark_OptimismPortal:test_proveWithdrawalTransaction_benchmark() (gas: 141289)
```

On average, across a range of $[0, 2048)$ byte inputs, the Yul implementation of `toNibbles` saves $470,870$ gas.
Screen Shot 2022-12-09 at 11 18 42 PM

**Invariants**
The `toNibbles` function has several invariants that must be held:
1. [✅  - [Length assertion test](https://github.com/ethereum-optimism/optimism/pull/4340/files#diff-7fac4081e056caa2cfcb5aca879f7c050c8eea05bdba051799037d12b98f0189R25)] The length of the resulting array of nibbles must always be two times the input array's length.
2. [✅ - [Static](https://github.com/ethereum-optimism/optimism/pull/4340/files#diff-7fac4081e056caa2cfcb5aca879f7c050c8eea05bdba051799037d12b98f0189R30-R47) & [Fuzz](https://github.com/ethereum-optimism/optimism/pull/4340/files#diff-7fac4081e056caa2cfcb5aca879f7c050c8eea05bdba051799037d12b98f0189R7-R27) Tests] The resulting array must contain **all** nibbles of the input array.
3. [✅ - [Test](https://github.com/ethereum-optimism/optimism/pull/4340/files#diff-7fac4081e056caa2cfcb5aca879f7c050c8eea05bdba051799037d12b98f0189R50-R55)] A zero length input should yield a zero length result.
4. Memory offsets outside of the bounds of the memory allocated for the `nibbles` array must not be written to.

**Memory Safety**
The memory for the `nibbles` array is allocated outside of an assembly block, and as such, is handled by solc. Within the assembly block, the only memory written to is within the bounds of the `nibbles` array's allocated memory. This can be seen in the loop:

```yul
for {
    let i := 0x00
} lt(i, bytesLength) {
    i := add(i, 0x01)
} {
    // Get the starting offset of the next 2 bytes in the nibbles array
    let offset := add(nibblesStart, shl(0x01, i))

    // Load the byte at the current index within the `_bytes` array
    let b := byte(0x00, mload(add(bytesStart, i)))

    // !!!
    // SAFETY: `offset` can never be larger than `nibblesStart + ((bytesLength - 1) * 2)`,
    // which ensures that memory outside of what has been allocated for `nibbles` is never
    // written to.
    // !!!

    // Pull out the first nibble and store it in the new array 
    mstore8(offset, shr(0x04, b))
    // Pull out the second nibble and store it in the new array
    mstore8(add(offset, 0x01), and(b, 0x0F))
}
```

**Metadata**
- Fixes ENG-3108


packages/contracts-bedrock/contracts/libraries/Bytes.sol
```diff
@@ -109,34 +109,53 @@ library Bytes {
...
-     * @param _other Second byte array to compare.
+     * @param a First byte array to compare.
+     * @param b Second byte array to compare.
      *
-     * @return True if the two byte arrays are equal, false otherwise.
+     * @return _eq True if the two byte arrays are equal, false otherwise.
      */
-    function equal(bytes memory _bytes, bytes memory _other) internal pure returns (bool) {
-        return keccak256(_bytes) == keccak256(_other);
+    function equal(bytes memory a, bytes memory b) internal pure returns (bool _eq) {
```
smartcontracts, line 156: I know it's a bit ugly but we deliberately block the use of named returns in our contracts because when not used carefully it can create confusing return values and obscure the flow of execution for auditors. Instead of mixing named returns and unnamed returns we currently have a blanket ban on named returns.

smartcontracts: We so currently enforce the underscore prefix convention for input variables, we might eventually change that but for now it's better to stay consistent.

smartcontracts (commented):
This seems like a pretty significant improvement. I'd love to see a benchmark test for `proveWithdrawalTransaction` added in a separate PR so that we can observe the actual difference. My intuition is that the `equal` change is probably fine but we'd need to be careful about the `toNibbles` change.

packages/contracts-bedrock/contracts/libraries/Bytes.sol
clabby: Swapped over to prefix names on input vars- will chat w/ you about the named return variable when we hop on gather in a few 😄

clabby:
> This seems like a pretty significant improvement. I'd love to see a benchmark test for `proveWithdrawalTransaction` added in a separate PR so that we can observe the actual difference. My intuition is that the `equal` change is probably fine but we'd need to be careful about the `toNibbles` change.

For sure- will create a benchmark / fuzz test against the old versions after the migration rehersal.

packages/contracts-bedrock/contracts/libraries/Bytes.sol
clabby: Removed named return vars- gets rid of some of the savings, but hear the convention piece.

tynes:
Perhaps we wanted to merge #4350 into develop first and then rebase this on top of develop. Sorry, I merged #4350

clabby:
> Perhaps we wanted to merge #4350 into develop first and then rebase this on top of develop. Sorry, I merged #4350

We could do this for sure, the `Bytes` lib tests in #4350 are effectively useless w/o these changes, though, given that they test the new impls of these functions against the old ones.

tynes:
If we could rebase this on `develop`, that would be amazing

packages/contracts-bedrock/contracts/test/Bytes.t.sol
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /// @dev Test that the `equals` function returns false when passed
+    /// non-equal byte arrays.
+    function testFuzz_equals_failure(bytes memory _a, bytes memory _b) public {
+        vm.assume(keccak256(_a) != keccak256(_b));
+
+        assertFalse(Bytes.equal(_a, _b));
+    }
+}
```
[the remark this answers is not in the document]
clabby: Done 🤝

packages/contracts-bedrock/contracts/libraries/Bytes.sol
mslipper: cc @maurelian - great thing to put in semgrep

packages/contracts-bedrock/contracts/libraries/Bytes.sol
```diff
@@ -131,12 +150,12 @@ library Bytes {
...
-     * @param _other Second byte array to compare.
+     * @param _a First byte array to compare.
+     * @param _b Second byte array to compare.
      *
      * @return True if the two byte arrays are equal, false otherwise.
      */
-    function equal(bytes memory _bytes, bytes memory _other) internal pure returns (bool) {
-        return keccak256(_bytes) == keccak256(_other);
+    function equal(bytes memory _a, bytes memory _b) internal pure returns (bool) {
+        return keccak256(_a) == keccak256(_b);
```
clabby, line 159: Optimization for `equal` removed- without named return parameters, the assembly version consumes more gas.

tynes (approved):
This looks good to me, good work and nice comments

clabby:
Closing per @maurelian's request to delay optimizations / non-trivial changes to the audited `contracts-bedrock` codebase. Optimizations will be performed in a larger-scope project while maintaining `contracts-bedrock` as a reference implementation.

PR #4351, opened 2022-12-09
feat(ctb): preserve legacy ETH invariant

smartcontracts:
**Description**
Preserves the invariant in LegacyERC20ETH that balanceOf(x) is always the same as address(x).balance. balanceOf is the only function that was not being blocked and contracts may exist that rely on this behavior.

**Invariants**
Preserves the invariant that `LegacyERC20ETH.balanceOf(x) == address(x).balance`.


mslipper:
I think this is important that we do this. One of Bedrock's guiding principles is backwards compatibility, so in the off chance someone is using this we should expose this functionality.

tynes:
This makes sense. Its a bit weird that `totalSupply == 0` while `balanceOf(...) > 0` can be true.
Would be nice to trace everything and see if the `balanceOf` interface is called. Any usage of `transfer` or `transferFrom` will revert, which could happen after a call to `balanceOf`.
Unclear if this prevents bugs or adds bugs

tynes:
@maurelian can you think of any issues with this?

maurelian:
What invariants need to be preserved in the OVM_ETH contract? 
1. You currently cannot call any state changing methods.

PR #4356, opened 2022-12-09
INF-21 signer service boilerplate

zhwrd:

PR #4357, opened 2022-12-09
INF-21 signer service api implementation

zhwrd:


zhwrd:
* develop
  * **PR #4356** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4356" target="_blank">Graphite</a> 
    * **PR #4357** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4357" target="_blank">Graphite</a>  👈
      * **PR #4358** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4358" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4357?utm_source=stack-comment).

signer/.air.toml
```diff
@@ -0,0 +1,37 @@
+root = "."
```
tynes, line 1: What exactly is this file used for?

zhwrd: its a [hot reloader for go](https://github.com/cosmtrek/air) i tend to use for go service development... i suppose don't have to commit this

signer/service/service.go
```diff
@@ -21,3 +41,124 @@ func (s *SignerService) RegisterAPIs(server *oprpc.Server) {
 		Service:   s,
 	})
 }
+
+type SignTransactionResponse struct {
+	Signature hexutil.Bytes `json:"signature"`
+}
+
+func (s *SignerService) SignTransaction(
```
tynes, line 49: What do you think about implementing the RPC `eth_signTransaction` so that existing tooling can "just work" with this service? This RPC is meant to work with a key that is remote from the client

The interface is here:
https://www.quicknode.com/docs/ethereum/eth_signTransaction

The main difference is that you don't pass in the raw transaction, you instead specify the fields. This is slightly less forwards compatible, ie we will manually need to add new transaction types

tynes: One interesting this about `eth_signTransaction` is that you can specify `from` so then we can manage multiple keys in a single deployment

signer/.air.toml
tynes: Going to leave this up to the team to decide :)

signer/service/service.go
zhwrd: Yeah we are going to support multiple keys, however we are going to select the key based on authenticated client making the request. We can verify it matches the expected from address in the payload, but it is somewhat redundant.

In terms of existing tooling, our authentication for this service will be using mTLS so as long as the the clients can support that then yeah think it would make sense to use this RPC

op-service/log/cli.go
```diff
@@ -64,7 +64,7 @@ func (cfg CLIConfig) Check() error {
 func NewLogger(cfg CLIConfig) log.Logger {
 	handler := log.StreamHandler(os.Stdout, Format(cfg.Format, cfg.Color))
 	handler = log.SyncHandler(handler)
-	log.LvlFilterHandler(Level(cfg.Level), handler)
+	handler = log.LvlFilterHandler(Level(cfg.Level), handler)
```
mslipper, line 67: D'oh. I missed that, nice catch.

signer/service/service.go
mslipper: One potential counterpoint to this is that we will eventually use this service to sign the P2P messages the sequencer sends over P2P.

I'm not married to either implementation, but want to ensure that we leave the door open for this functionality in the future.

PR #4361, opened 2022-12-09
feat(contracts-periphery): Add attestation contract

roninjin10:
Add attestation contract
- Has `attest` to make attestations

Add attestation tests
- Added just two tests since they were my first ever tests
- Looking for feedback if tests are good and if I should add more

Update foundry.toml
- Had some issues here
- Add bedrock contracts since we use SemVer
- Add open zepplin contracts

Update package.json
- If I added latest contract bedrock the module got hoisted.  I couldn't target the node module when it was hoisted because foundry said it was outside of project
- Because of this the hack I did was use 0.10.0 instead of 0.11.0
- Open to a less hacky way of doing this
- Remove unused contracts-upgradeable

Deployment script
- Adding in a follow up pr
- Hacky deployment script we were using is here https://github.com/ethereum-optimism/gateway/blob/develop/packages/contracts/scripts/deploy.ts


roninjin10:
* develop
  * **PR #4361** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4361" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4361?utm_source=stack-comment).

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,75 @@
...
+    mapping(address => mapping(address => mapping(bytes32 => bytes))) public attestations;
+
+    event AttestationCreated(
+        address indexed creator,
+        address indexed about,
+        bytes32 indexed key,
+        bytes val
+    );
+
+    constructor() Semver(0, 0, 1) {}
```
tynes, line 26: We should call `initialize()` in the `constructor` to prevent the implementation from being uninitialized

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @notice  Initialize the Optimist contract.
+     * @dev     Initializes the AttestationStation contract with msg.sender as owner
+     */
+    function initialize() public initializer {
+        __Ownable_init();
```
tynes, line 33: I believe we should transfer ownership here to the desired owner so that the `msg.sender` is not the owner

tynes: ie `initialize(address _owner)`

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
tynes: Should pass `owner` into the constructor and pass to `initialize`

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {
+    OwnableUpgradeable
+} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import { Semver } from "@eth-optimism/contracts-bedrock/contracts/universal/Semver.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+contract AttestationStation is Initializable, OwnableUpgradeable, Semver {
```
tynes, line 10: Why exactly is this ownable? I don't see `onlyOwner` used anywhere

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: ah makes sense

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
tynes: To be honest this doesn't need an `initialize` function because there are no side effects in the constructor, it should just work behind a proxy

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
tynes: See this comment before updating: https://github.com/ethereum-optimism/optimism/pull/4361#discussion_r1044821903

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
smartcontracts: Yep doesn't need to be ownable

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,75 @@
...
+        emit AttestationCreated(msg.sender, _about, _key, _val);
+    }
+
+    /**
+     * @notice  Attest to the given data.
+     * @dev     Attests to the given data from the sender.
+     * @param   _attestations  The array of attestation data.
+     */
+    function attestBulk(AttestationData[] memory _attestations) public {
+        for (uint256 i = 0; i < _attestations.length; ++i) {
```
tynes, line 61: You should be able to cache the length and iterate in an `unchecked` block

tynes:
```
uint256 length = _attenstations.length

for (uint256 i; i < length; ) {
  attest(..);
  unchecked { ++i };
}
```

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: Doesn't it need to be ownable so it can transfer ownership (which is onlyOwner)?  https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/access/OwnableUpgradeable.sol#L74

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * @notice  Reads an attestation
+     * @return  val bytes  The attestation
+     */
+    function readAttestation(
+        address _creator,
+        address _about,
+        bytes32 _key
+    ) public view returns (bytes memory val) {
+        val = this.attestations(_creator, _about, _key);
```
tynes, line 75: hmm no need for `this`, i think this should work:

```
return attestations[_creator][_about][_key]
```

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
tynes: There are no methods on the contract that require special access controls

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: oops javascript habits.  I nuked this entire function since it was unnecessary

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: got it.  So the deployment script will be what puts it behind a proxy

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: removed ownable

packages/contracts-periphery/package.json
```diff
@@ -55,7 +55,7 @@
   "devDependencies": {
     "@defi-wonderland/smock": "^2.0.7",
     "@eth-optimism/contracts": "^0.5.39",
-    "@eth-optimism/contracts-bedrock": "^0.11.0",
+    "@eth-optimism/contracts-bedrock": "0.10.0",
```
roninjin10, line 58: this is a hack to force yarn worskpaces to install this package in local node modules.   Otherwise it hoists both this and contract bedrock to ../../node_modules which foundry doesn't like and throws an error for if i try to remap outside of project

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: Awesome tip!

roninjin10:
@tynes @smartcontracts updated based on all the comments!  Thanks for the lightning fast review holy!  And while upgrading bedrock.

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,55 @@
...
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+contract AttestationStation is Initializable, Semver {
+    struct AttestationData {
+        address about;
+        bytes32 key;
+        bytes val;
+    }
+
+    mapping(address => mapping(address => mapping(bytes32 => bytes))) public attestations;
```
tynes, line 14: The event, mapping and struct need docs

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Semver } from "@eth-optimism/contracts-bedrock/contracts/universal/Semver.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+contract AttestationStation is Initializable, Semver {
```
tynes, line 7: This doesnt need `Initializable`

tynes: This contract needs a docstring `@notice`

roninjin10: just noticed this one

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,81 @@
...
+ * @dev   Contract for creating attestations.
+ * @notice This contract is not yet audited.
+ */
+contract AttestationStation is Semver {
+    /**
+     * @notice Struct representing data that is being attested
+     *
+     * @custom:field about                    Address being attested about (not creator/msg.sender)
+     * @custom:field key                      A bytes32 key for the attestation.
+     * @custom:field val                      The attestation as arbitrary bytes
```
roninjin10, line 18: I copied this comment originally from contracts-bedrock.   Should it be @custom?  What does @custom mean?

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Semver } from "@eth-optimism/contracts-bedrock/contracts/universal/Semver.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+/**
+ * @title AttestationStation
+ * @dev   Contract for creating attestations.
+ * @notice This contract is not yet audited.
```
roninjin10, line 10: should I have this notice here?

roninjin10 (commented):
@tynes comments added

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: Going to add more words to the @notice describing the contract like L1CrossDomainMessenger has

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
smartcontracts: Also nitpick but the descriptions should be separated by one space from the longest field name

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,83 @@
...
+     * @param   _about  The address of the attestation subject.
+     * @param   _key  The key of the attestation.
+     * @param   _val  The value of the attestation.
+     */
+    function _attest(
+        address _about,
+        bytes32 _key,
+        bytes memory _val
+    ) private {
+        attestations[msg.sender][_about][_key] = _val;
```
smartcontracts, line 65: I don't think this needs to be broken out into a separate function

smartcontracts: This is just my obsessive brain enjoying the idea of having an entire contract with only a single public function

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Semver } from "@eth-optimism/contracts-bedrock/contracts/universal/Semver.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
```
tynes, line 5: we can delete this import

packages/contracts-periphery/package.json
tynes: you can do this as well with the `nohoist` option in the root `package.json`

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: good catch

packages/contracts-periphery/package.json
tynes: it'll create a symlink, there are some examples in the `nohoist` field

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,73 @@
...
+        address about;
+        bytes32 key;
+        bytes val;
+    }
+
+    /**
+     * @notice Maps addresses to attestations
+     * @dev addresses map to attestations map of
+     *      `about` addresses to key/values
+     *      key/values are a map of bytes32 to bytes
```
tynes, line 31: nit: missing extra space

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
```diff
@@ -0,0 +1,73 @@
...
+     * @notice Emitted when Attestation is created
+     *
+     * @param creator Address that attested.
+     * @param about  Address attestation is about.
+     * @param key Key of the attestation.
+     * @param val Value of the attestation.
+     */
+    event AttestationCreated(
+        address indexed creator,
+        address indexed about,
```
tynes, line 45: slight nit: I feel like `topic` is a better term than `about` here

tynes:
Looks good to me

packages/contracts-periphery/contracts/universal/op-nft/AttestationStation.sol
roninjin10: nice I agree!

roninjin10: actually nvm don't want to change that since it's a breaking change to @gateway/contracts @gateway/backend and @gateway/frontend

tynes (approved):

packages/contracts-periphery/contracts/foundry-tests/AttestationStation.t.sol
```diff
@@ -0,0 +1,139 @@
...
+        // alice is going to attest about bob
+        AttestationStation.AttestationData memory attestationData = AttestationStation
+            .AttestationData({
+                about: bob,
+                key: bytes32("test-key:string"),
+                val: bytes("test-value")
+            });
+
+        // assert the attestation starts empty
+        assertEq(attestationStation.attestations(address(this), address(this), "test"), "");
```
roninjin10, line 46: Oops benign bug in the test (from copilot).  Will fix
```suggestion
        assertEq(attestationStation.attestations(alice_attestor, attestationData.about, attestationData.key), "");
```

PR #4367, opened 2022-12-10
op-node: continue sequencer cleanup

protolambda:
**Description**

This PR is a follow-up to #4236 to clean up the sequencing part of the driver more:
- Get rid of the old `startNewL2Block` and `completeNewBlock` methods
- Put the L1 origin selector in the `Sequencer`, the `Driver` doesn't need it.
- Simplify the `Driver` loop by:
  - removing the sequence error and sequence error time stuff, replaced by a cleaner version in `Sequencer`
  - simplifying it to just "sequencer action": the driver does not make branching sequencing logic. We can then test the sequencing logic better in isolation.
- Cleaning up logs/errors: no more duplicate wrapped errors or double sequence error metrics triggers

In another separate PR I would also want to analyze the L1 origin selection deeper: this is old code, with a likely bug based on what we've seen with sequencer time drift etc.
And the usage of the L1 head as seen by the Driver for shortcuts in the origin selection worries me: it's an abstraction leak, and we can simplify the sequencer interface even more by removing the `l1Head` argument.

**Tests**

Previously there were two very verbose, and a bit oversized, tests that covered just the timestamp selection. Showing how untestable the code really was. Since we don't expose the older more leaky interface anymore, these tests didn't make any sense. We should however cover the sequencer edge-cases.

I'm thinking we can refactor the `derive`  package calls to bundle them into an engine handle, or maybe attach them to the engine-queue, to try and unify the engine-interactions with the derivation pipeline in one place.
If these package calls were part of an interface, we could more easily mock them, and then unit-test the sequencing logic quite easily.

For now this PR is a draft, it's blocked by the above refactor work for testing, which we probably want to discuss and design before changing code.

**Metadata**

Fix ENG-3104


protolambda:
Closing in favor of new PR stack

PR #4374, opened 2022-12-10
feat(base-service-v2): Add trpc servers to base service v2

roninjin10:
Currently being used in the gateway backend: https://github.com/ethereum-optimism/gateway/tree/develop/packages/backend/src/routes

And it is also being used on the gateway frontend: https://github.com/ethereum-optimism/gateway/search?q=apiReact

TRPC allows a really easy api for creating full typesafe apis for typescript consumers.   It's the best way in terms of developer velocity safety to connect typescript services.  The typesafety does not require a build step.  https://trpc.io/

I added trpc as an option to base service v2 because for some use cases express would be better.

**Testing**

This had 100% test coverage in gateway repo but the tests were written in jest not mocha.  Won't have time to translate them here anytime soon.  This will get implicitly tested though via the gateway repo using it in it's tests

### DIFF
--- a/packages/common-ts/package.json
+++ b/packages/common-ts/package.json
@@ -49,7 +49,10 @@
     "pino-multi-stream": "^5.3.0",
     "pino-sentry": "^0.7.0",
     "prom-client": "^13.1.0",
-    "qs": "^6.10.5"
+    "qs": "^6.10.5",
+    "@trpc/server": "^10.4.3",
+    "superjson": "^1.10.1",
+    "zod": "^3.19.1"
   },
   "devDependencies": {
     "@ethersproject/abstract-provider": "^5.7.0",
--- a/packages/common-ts/src/base-service/base-service-v2.ts
+++ b/packages/common-ts/src/base-service/base-service-v2.ts
@@ -14,6 +14,8 @@ import morgan from 'morgan'
 import { Logger, LogLevel } from '../common/logger'
 import { Metric, Gauge, Counter } from './metrics'
 import { validators } from './validators'
+import { TrpcApi } from './trpc'
+import { Metrics } from '../common'
 
 export type Options = {
   [key: string]: any
@@ -51,6 +53,18 @@ export type MetricsSpec<TMetrics extends MetricsV2> = {
 }
 
 export type ExpressRouter = Router
+
+// need to make an example to infer the type
+class ExampleTRPCApi<
+  TMetrics extends MetricsV2,
+  TRoutes extends TrpcApi<TMetrics>['routes']
+> extends TrpcApi<TMetrics> {
+  routes: TRoutes = this.router({}) as any
+}
+type ExtendedTrpcApi<
+  TMetrics extends MetricsV2,
+  TRoutes
+> = typeof ExampleTRPCApi<TMetrics, TRoutes>
 
 /**
  * BaseServiceV2 is an advanced but simple base class for long-running TypeScript services.
@@ -426,6 +440,10 @@ export abstract class BaseServiceV2<
 
       // Register user routes.
       const router = express.Router()
+      if (this.TrpcApi) {
+        const trpc = new this.TrpcApi(this.metrics, this.logger)
+        router.use('trpc', trpc.createExpressMiddleware())
+      }
       if (this.routes) {
         this.routes(router)
       }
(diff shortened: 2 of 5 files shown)


packages/common-ts/package.json
```diff
@@ -49,7 +49,10 @@
     "pino-multi-stream": "^5.3.0",
     "pino-sentry": "^0.7.0",
     "prom-client": "^13.1.0",
-    "qs": "^6.10.5"
+    "qs": "^6.10.5",
+    "@trpc/server": "^10.4.3",
+    "superjson": "^1.10.1",
```
roninjin10, line 54: Superjson allows you to send dates and some other things

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -52,6 +54,18 @@ export type MetricsSpec<TMetrics extends MetricsV2> = {
...
+  TMetrics extends MetricsV2,
+  TRoutes extends TrpcApi<TMetrics>['routes']
+> extends TrpcApi<TMetrics> {
+  routes: TRoutes = this.router({}) as any
+}
+type ExtendedTrpcApi<
+  TMetrics extends MetricsV2,
+  TRoutes
+> = typeof ExampleTRPCApi<TMetrics, TRoutes>
+
```
roninjin10, line 68: This was the quickest way I could get the right type.   I need a type that is `class that extends TrpcApi`.    There is likely a better way to do this in typescript but this was easy and works

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -426,6 +440,10 @@ export abstract class BaseServiceV2<
 
       // Register user routes.
       const router = express.Router()
+      if (this.TrpcApi) {
+        const trpc = new this.TrpcApi(this.metrics, this.logger)
+        router.use('trpc', trpc.createExpressMiddleware())
+      }
```
roninjin10, line 446: add trpc apis to /api/trpc.

Might be better to add it to /trpc

packages/common-ts/src/base-service/trpc.ts
```diff
@@ -0,0 +1,133 @@
...
+ * // import only the type so you don't have to import any actual code
+ * import type {Api} from '../backend/api'
+ * import { createTRPCReact, httpBatchLink } from '@trpc/react'
+ *
+ * // the client can infer the api types from the routes
+ * export const apiClient = createTRPCReact<Api['routes']>()
+ *
+ * // you can then use it in a typesafe way in your client code
+ * apiClient.metadata.useQuery({address: userAddress})
+ */
```
roninjin10, line 51: These doc strings show up in editors when you hover over the variable

packages/common-ts/src/base-service/trpc.ts
```diff
@@ -0,0 +1,133 @@
...
+   *     metadata: this.metadataController,
+   *     nestedroute: this.router({
+   *        nestedRoute: this.nestedRouteHandler
+   *     })
+   *  })
+   *
+   */
+  public abstract readonly routes:
+    | BuildProcedure<any, any, any>
+    | CreateRouterInner<any, any>
```
roninjin10, line 117: Better api I can change to in future pr is createTrpcRoute that gets passed this object and must return routes.   This is a more direct port of the gateway code as is and reduces work needed to start using this in gateway via working like this to start though.

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -538,6 +556,21 @@ export abstract class BaseServiceV2<
...
+   * on typescript microservices and frontends with no build/generation
+   * step.
+   *
+   * @example
+   * // alternatively import from another file
+   * trpc: class MyTRPCApi extends TrpcApi {
+   *  // define your api here
+   * }
+   */
+  protected TrpcApi: ExtendedTrpcApi<TMetrics, StandardMetrics & Metrics>
```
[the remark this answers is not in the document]
roninjin10: Better api I can change to in future pr is createTrpcRoute that gets passed this object and must return routes. This is a more direct port of the gateway code as is and reduces work needed to start using this in gateway via working like this to start though.

PR #4375, opened 2022-12-11
feat(cmn): clean up BaseServiceV2 options

smartcontracts:
**Description**
Cleans up the system of options for BaseServiceV2. Includes some refactoring to reduce the size of base-service-v2.ts. Primary change from a user point of view is that standard options and regular options get merged. Makes it more clear how user inputted options will override defaults provided by the service.


yarn.lock
```diff
@@ -3207,11 +3207,6 @@
   resolved "https://registry.yarnpkg.com/@openzeppelin/contracts-upgradeable/-/contracts-upgradeable-4.7.3.tgz#f1d606e2827d409053f3e908ba4eb8adb1dd6995"
   integrity sha512-+wuegAMaLcZnLCJIvrVUDzA9z/Wp93f0Dla/4jJvIhijRrPabjQbZe6fWiECLaJyfn5ci9fqf9vTw3xpQOad2A==
```
tynes, line 3209: Why is this diff in here?

tynes (approved):

PR #4376, opened 2022-12-11
feat(cmn): BSv2 private options by default

smartcontracts:
**Description**
Makes BaseServiceV2 options "secret" by default. Services will have to explicitly mark options as "public" for those options to be logged and included in metrics. Much more sane than logging everything, especially for anyone using BaseServiceV2 who might not know that this happens under the hood.

### DIFF
--- a/.changeset/nervous-moose-bake.md
+++ b/.changeset/nervous-moose-bake.md
@@ -1,0 +1,8 @@
+---
+'@eth-optimism/common-ts': minor
+'@eth-optimism/drippie-mon': patch
+'@eth-optimism/fault-detector': patch
+'@eth-optimism/replica-healthcheck': patch
+---
+
+Updates BaseServiceV2 so that options are secret by default. Services will have to explicitly mark options as "public" for those options to be logged and included in the metadata metric.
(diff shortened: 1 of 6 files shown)


smartcontracts:
> Looks like the monorepo cannot build, the message relayer needs to be updated
> 
> ```
> lerna ERR! yarn run build exited 2 in '@eth-optimism/message-relayer'
> ```
> 
> Which to be honest, we haven't updated the message relayer for 2 step withdrawals so i wouldn't be opposed to just nuking it since its no longer used at all

Let's nuke it after Bedrock goes live

tynes:
> > Looks like the monorepo cannot build, the message relayer needs to be updated
> > ```
> > lerna ERR! yarn run build exited 2 in '@eth-optimism/message-relayer'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which to be honest, we haven't updated the message relayer for 2 step withdrawals so i wouldn't be opposed to just nuking it since its no longer used at all
> 
> Let's nuke it after Bedrock goes live

https://github.com/ethereum-optimism/optimism/pull/4387

tynes (approved):

smartcontracts:
@mergifyio requeue

PR #4387, opened 2022-12-12
message-relayer: delete dead code

tynes:
**Description**

This package is not used and has not been updated
to with with the new withdrawal scheme. There is no real reason to keep it around because we will need to continue to maintain it if we do keep it around.
It currently does not run in CI.


tynes:
Closing until after bedrock ships

PR #4389, opened 2022-12-12
feat(ctb): move gov contracts to top level folder

smartcontracts:
**Description**
Moves governance contracts to a new top-level folder named governance. Also copies the MintManager into contracts-bedrock in preparation to delete contracts-governance entirely. We no longer need our own version of the MerkleDistributor because Uniswap now ships a version of the MerkleDistributor that provides the functionality we needed.


yarn.lock
```diff
@@ -3207,11 +3207,6 @@
   resolved "https://registry.yarnpkg.com/@openzeppelin/contracts-upgradeable/-/contracts-upgradeable-4.7.3.tgz#f1d606e2827d409053f3e908ba4eb8adb1dd6995"
   integrity sha512-+wuegAMaLcZnLCJIvrVUDzA9z/Wp93f0Dla/4jJvIhijRrPabjQbZe6fWiECLaJyfn5ci9fqf9vTw3xpQOad2A==
 
-"@openzeppelin/contracts-upgradeable@4.8.0":
```
smartcontracts, line 3210: This diff just appeared by running yarn install?

packages/contracts-bedrock/contracts/governance/MintManager.sol
```diff
@@ -0,0 +1,88 @@
+// SPDX-License-Identifier: MIT
```
maurelian, line 1: Shouldn't this be a rename?

smartcontracts: I also cleaned up some stuff in the contract, namely error strings

tynes (approved):

PR #4399, opened 2022-12-12
ctb: Fix linter warnings in echidna contracts

maurelian:
**Description**

Fixes some solhint warnings

### DIFF
--- a/packages/contracts-bedrock/contracts/echidna/FuzzResourceMetering.sol
+++ b/packages/contracts-bedrock/contracts/echidna/FuzzResourceMetering.sol
@@ -5,16 +5,16 @@ import { Arithmetic } from "../libraries/Arithmetic.sol";
 import { StdUtils } from "forge-std/Test.sol";
 
 contract EchidnaFuzzResourceMetering is ResourceMetering, StdUtils {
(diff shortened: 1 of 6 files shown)


packages/contracts-bedrock/contracts/echidna/FuzzResourceMetering.sol
```diff
@@ -128,30 +128,37 @@ contract EchidnaFuzzResourceMetering is ResourceMetering, StdUtils {
 
     function _burnInternal(uint64 _gasToBurn) private metered(_gasToBurn) {}
 
+    // solhint-disable-next-line func-name-mixedcase
```
tynes, line 131: Is there a way to do this for the entire file instead of line by line?

tynes (approved):

packages/contracts-bedrock/contracts/echidna/FuzzResourceMetering.sol
maurelian: Talking to @smartcontracts we decided to just remove that rule from the solhint config, it's the kind of rule we're not going to unintentionally break anyways.

PR #4403, opened 2022-12-12
test(ctb): MintManager tests

smartcontracts:
**Description**
Ports old hardhat tests for MintManager to Foundry.


packages/contracts-bedrock/contracts/test/MintManager.t.sol
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    function test_constructor_succeeds() external {
+        assertEq(manager.owner(), owner);
+        assertEq(address(manager.governanceToken()), address(gov));
+    }
+
+    function test_mint_fromOwner_succeeds() external {
+        // Mint once.
+        vm.prank(owner);
```
maurelian, line 32: I got this suggestion recently and think it's easier to reason about/maintain.

```suggestion
        vm.prank(manager.owner());
```

packages/contracts-bedrock/contracts/test/MintManager.t.sol
```diff
@@ -0,0 +1,120 @@
...
+        manager.mint(owner, 100);
+
+        // Token balance increases.
+        assertEq(gov.balanceOf(owner), 100);
+    }
+
+    function test_mint_fromNotOwner_reverts() external {
+        // Mint from rando fails.
+        vm.prank(rando);
```
maurelian, line 41: I think this line can be deleted

packages/contracts-bedrock/contracts/test/MintManager.t.sol
```diff
@@ -0,0 +1,120 @@
...
+import { GovernanceToken } from "../governance/GovernanceToken.sol";
+
+contract MintManager_Test is CommonTest {
+    address constant owner = address(0x1234);
+    address constant rando = address(0x5678);
+    GovernanceToken internal gov;
+    MintManager internal manager;
+
+    function setUp() external {
+        vm.prank(owner);
```
maurelian, line 15: Could just use `alice` here, and then delete the state var.

packages/contracts-bedrock/contracts/test/MintManager.t.sol
smartcontracts: This creates problems with `vm.prank` because the call to `manager.owner()` gets pranked e.g. on line 57

packages/contracts-bedrock/contracts/test/MintManager.t.sol
smartcontracts: I prefer having the explicit variable name

maurelian (approved):

PR #4414, opened 2022-12-12
[Bedrock] op-batcher / op-proposer: allow customization of signer function

mdehoog:
**Description**

Supports passing a custom transaction signer function to the batcher's transaction manager and the proposer's driver.

**Tests**

No logic changes in this PR, simply threading through some configuration.

**Additional context**

This allows plugging in different implementations for transaction signing, which paves the way for more secure management of keys.

PR #4415, opened 2022-12-13
Add PrivateKeySignerFn method to op-service/crypto package

mdehoog:
**Description**

Adds a helper method `crypto.PrivateKeySignerFn` which generates a `bind.SignerFn` for signing transactions with a private key.

**Tests**

This is a commonly implemented method, tested in #4414.

**Additional context**

This is needed to support #4414, given `op-service` is versioned separately.

### DIFF
--- a/op-service/crypto/signature.go
+++ b/op-service/crypto/signature.go
@@ -1,0 +1,26 @@
+package crypto
+
+import (
+	"crypto/ecdsa"
+	"math/big"
+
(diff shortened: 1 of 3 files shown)


tynes:
Looks like the `go.mod` got out of date, I looked at the added packages and they look generally sane to me. Implementation also looks good to me

mslipper (approved):

mdehoog:
thanks @tynes, yeah I just ran `go mod tidy` in the `op-service` directory, which included the new deps

PR #4416, opened 2022-12-13
contracts-bedrock: fee vault track total value

tynes:
**Description**

Adds an accumulator for the total amount of fees
withdrawn from the system. This is really helpful for data analytics. I also think it could be used to build an on chain prediction market for economic activity. This code path isn't triggered too often, so an extra sload/sstore and checked math doesn't matter much.


packages/contracts-bedrock/.storage-layout
```diff
@@ -281,11 +281,33 @@
 ➡ contracts/L2/SequencerFeeVault.sol:SequencerFeeVault
 =======================
 
-+---------------+---------+------+--------+-------+------------------------------------------------------+
-| Name          | Type    | Slot | Offset | Bytes | Contract                                             |
-+========================================================================================================+
-| spacer_0_0_20 | address | 0    | 0      | 20    | contracts/L2/SequencerFeeVault.sol:SequencerFeeVault |
```
tynes, line 287: Note that the spacer for the legacy `l1FeeWallet` is in the same place. This value is in the storage of the `SequencerFeeVault` but the contract now uses an immutable to read that value. This allows for the admin of the proxy to upgrade the contract to change the withdrawal account

smartcontracts (approved):

packages/contracts-bedrock/.storage-layout
maurelian: I'm confused by why this is even showing a diff on these lines, can you see what's changed?

PR #4419, opened 2022-12-13
op-chain-ops: withdrawal tx helper

tynes:
**Description**

Adds a method to the `Withdrawal` that turns it into a `bindings.TypesWithdrawalTransaction` that is suitable for sending via an `ethclient` when sending a transaction. This will help to deduplicate code.

### DIFF
--- a/op-chain-ops/crossdomain/withdrawal.go
+++ b/op-chain-ops/crossdomain/withdrawal.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"math/big"
 
+	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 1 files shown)


op-chain-ops/crossdomain/withdrawal.go
```diff
@@ -139,3 +140,16 @@ func (w *Withdrawal) StorageSlot() (common.Hash, error) {
...
 	return common.BytesToHash(slot), nil
 }
+
+// WithdrawalTransaction will convert the Withdrawal to a type
+// suitable for sending a transaction.
+func (w *Withdrawal) WithdrawalTransaction() bindings.TypesWithdrawalTransaction {
+	return bindings.TypesWithdrawalTransaction{
+		Nonce:    w.Nonce,
+		Sender:   *w.Sender,
+		Target:   *w.Target,
```
protolambda, line 150: Think I'm missing some context here, but do withdrawals not support contract deployments? Target may be nil if contracts are deployed.

tynes: Withdrawals do not directly support contract deployments, although a withdrawal can `CALL` a factory that deploys a contract.

Note that `CALL` is used, not `CREATE`:
- https://github.com/ethereum-optimism/optimism/blob/764a9271eb68f70bf4c791662261ad0ba4e8e73f/packages/contracts-bedrock/contracts/L1/OptimismPortal.sol#L305
- https://github.com/ethereum-optimism/optimism/blob/764a9271eb68f70bf4c791662261ad0ba4e8e73f/packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol#L317

I don't think adding native `CREATE` via withdrawal is worth the complexity altho it would be nice for "completeness"

mslipper (approved):

PR #4425, opened 2022-12-13
op-chain-ops: configurable gov token

tynes:
**Description**

This adds the governance token config to the `DeployConfig`. The `name`, `symbol` and `_owner` can be set via new config options. We must set these values to be `Optimism`, `OP` and the address of the mint manager for the upgraded networks.

This change allows new networks to easily configure the gov token. This is important so that other networks do not spin up a chain that has a gov token that looks exactly the same as `OP`. This will enable scammers.

In my opinion, we should also make the `MintManager` a predeploy because it is the only L2 contract that we manage that is not a predeploy.


tynes:
> Can you hardcode the appropriate values for Goerli and mainnet? We shouldn't be able to create an invalid config for those networks.

Sure thing, was worried about having different config that was used for deployment being committed into the repo. Will follow up with another commit

tynes:
Followed up with updated config with the correct values. See the git commit message for how the value was obtained for the owner. This does not alter the owner in production networks, but it just allows for dynamic config of the owner for test networks

mslipper:
What do you think of ignoring the config values for mainnet and Goerli, and instead setting them from within `op-chain-ops` itself? I don't think we should even be looking at the deploy config for those networks. Thoughts?

maurelian (approved):
Deploy config changes LGTM

tynes:
> What do you think of ignoring the config values for mainnet and Goerli, and instead setting them from within `op-chain-ops` Go code ? I don't think we should even be looking at the deploy config for those networks. Thoughts?

I like where you are going with this thought but the problem is that the same deploy config is consumed by the typescript deploy scripts. We could hardcode into the Go a `MainnetDeployConfig` and a `GoerliDeployConfig` which contain the values that are required for the L2 portion, altho we would still need the ts versions and there may be overlapping values

mslipper (approved):

PR #4428, opened 2022-12-13
feat(ctp): clean AttestationStation comments

smartcontracts:
**Description**
Cleans up comments in the AttestationStation.


tynes (approved):
Should bump to semver `1.0.0` as well

tynes:
Perhaps also add a comment if this is supposed to live on L1 or L2

tynes:
flakey geth tests

PR #4432, opened 2022-12-13
Stop TestSystemMockP2P from flaking

trianglesphere:
**Description**

The root cause is that the elements were out of order. There isn't a strong reason that the order of the arrays would be the same so & the thing we care about is that the elements match so I switch the require statement to be that. In the failing test I manually verified that the elements were the same.

Last seen in https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/11438/workflows/e524ab48-9fff-445d-9813-8a4f76046012/jobs/321725.


mslipper (approved):

mslipper:
Nice detective work. So glad this is fixed.

PR #4434, opened 2022-12-14
contracts-periphery: run foundry tests in CI

tynes:
**Description**

The `yarn:coverage` command is ran in CI so add
`forge coverage` to that command so the foundry tests also run.

### DIFF
--- a/packages/contracts-periphery/contracts/foundry-tests/Transactor.t.sol
+++ b/packages/contracts-periphery/contracts/foundry-tests/Transactor.t.sol
@@ -50,7 +50,7 @@ contract TransactorTest is Transactor_Initializer {
         // Run CALL
         vm.prank(alice);
         vm.expectCall(address(callRecorded), data);
-        transactor.CALL(address(callRecorded), data, 200_000 wei, 420);
+        transactor.CALL(address(callRecorded), data, 200_000 wei);
     }
 
     // It should revert if called by non-owner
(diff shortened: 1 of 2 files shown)


packages/contracts-periphery/contracts/foundry-tests/Transactor.t.sol
```diff
@@ -78,7 +78,7 @@ contract TransactorTest is Transactor_Initializer {
         bytes memory data = abi.encodeWithSelector(reverter.doRevert.selector);
         // Run CALL
         vm.prank(bob);
-        transactor.DELEGATECALL(address(reverter), data, 200_000 wei);
```
maurelian, line 81: weird, all these tests were broken before?

maurelian (approved):

packages/contracts-periphery/contracts/foundry-tests/Transactor.t.sol
tynes: yes, they were not running in CI

tynes:
ty @maurelian <3

packages/contracts-periphery/contracts/foundry-tests/Transactor.t.sol
roninjin10: yea I've been `rm foundry-tests/Transactor.t.sol` everytime I run tests

PR #4421, opened 2022-12-13
test(ctb): port L2ERC721Bridge tests

smartcontracts:
**Description**
Ports the old hardhat tests for the L2ERC721Bridge to Foundry.

PR #4439, opened 2022-12-14
ctp: delete ERC721Bridge from ctp

tynes:
**Description**

The NFT bridge tests have been ported to foundry
as part of the `contracts-bedrock` package. We can now delete the NFT bridge code from `contracts-periphery`.

Note that the `OptimismMintableERC721Factory` will be a predeploy when upgrading to bedrock, so we do not need a deploy script for it. Its deploy artifact was migrated to `contracts-bedrock` so that the existing impl can be tracked. The deployment artifacts of the `L1ERC721Bridge` and its `Proxy` were also migrated to `contracts-bedrock`.

Depends on: https://github.com/ethereum-optimism/optimism/pull/4421

Also need to port tests for the factory to foundry


tynes (commented):
I cannot approve this PR as i opened it but looks good to me

maurelian (approved):

PR #4460, opened 2022-12-14
feat(ctb): inline verification function

smartcontracts:
**Description**
Inlines verifyInclusionProof because it was only being used once.


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -200,14 +200,23 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
             "OptimismPortal: withdrawal hash has already been proven"
         );
 
+        // Compute the storage slot of the withdrawal hash in the L2ToL1MessagePasser contract.
```
tynes, line 203: I could imagine in the future abstracting this behind a behind a function that pulls the version out of the withdrawal nonce and builds a particular input to `verifyInclusionProof`. Fine for now, but gives us a good upgrade path to verkle or mpt precompile

tynes (approved):

PR #4471, opened 2022-12-14
maint(ct): remove deprecated deployments

smartcontracts:
**Description**
Removes deprecated contract deployments to avoid confusing users who are looking at these deployment folders. Also updates the markdown generation script to remove an incorrect notice for Goerli.


tynes (approved):

roninjin10 (approved):
rip kovan

PR #4481, opened 2022-12-14
feat(ci): Make build fail if lockfile changes

roninjin10:
- --frozen-lockfile does not work on newer versions of yarn 1.x
- make a script copied from gateway repo that handles this

### DIFF
--- a/packages/contracts-periphery/package.json
+++ b/packages/contracts-periphery/package.json
@@ -25,6 +25,7 @@
     "gas-snapshot": "forge snapshot",
     "pretest:slither": "rm -f @openzeppelin && rm -f hardhat && ln -s node_modules/@openzeppelin @openzeppelin && ln -s ../../node_modules/hardhat hardhat",
     "posttest:slither": "rm -f @openzeppelin && rm -f hardhat",
+    "install:ci": "./scripts/yarn-lock-file.sh",
     "lint:ts:check": "eslint . --max-warnings=0",
     "lint:contracts:check": "yarn solhint -f table 'contracts/**/*.sol'",
     "lint:check": "yarn lint:contracts:check && yarn lint:ts:check",
--- a/packages/contracts-periphery/package.json
+++ b/packages/contracts-periphery/package.json
@@ -25,7 +25,7 @@
     "gas-snapshot": "forge snapshot",
     "pretest:slither": "rm -f @openzeppelin && rm -f hardhat && ln -s node_modules/@openzeppelin @openzeppelin && ln -s ../../node_modules/hardhat hardhat",
     "posttest:slither": "rm -f @openzeppelin && rm -f hardhat",
-    "install:ci": "./scripts/yarn-lock-file.sh",
+    "postinstall": "git diff --exit-code",
     "lint:ts:check": "eslint . --max-warnings=0",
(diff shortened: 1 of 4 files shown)


roninjin10:
* develop
  * **PR #4481** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4481" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4481?utm_source=stack-comment).

maurelian:
Could we accomplish the same thing just by doing something like this instead? 

https://github.com/ethereum-optimism/optimism/blob/1cac435d49960f2f95f5bdde19350e84fe41c6ea/.circleci/config.yml#L756

trianglesphere:
Either the script or `git diff --exit-code` work. One question is why there's the extra really large checked in thing.

smartcontracts:
`yarn install && git diff --exit-code` might be easier

trianglesphere (changes requested):
adding a reminder to not check in `.yarn/releases/yarn-1.22.19.cjs`

roninjin10:
@trianglesphere not sure how that got into my git commits but will revert

roninjin10:
@trianglesphere that file is the yarn runtime.    The way you upgrade yarn is you do `yarn set version 1.x` and it will update that file.   Would be a good idea to update it to latest but would rather do that in another pr

roninjin10:
@trianglesphere @smartcontracts I finally remembered why I did it this way and not using git.  Because sometimes git doesn't exist.   You can be in a docker container without git for example.   Also what is more important than the script itself is the giant comment explaining why it exists.  I think the pr as is is best.

roninjin10:
will change it though if anybody feels strongly

roninjin10:
The docker issue I don't think is an issue for this repo.  It was for us when I implemented https://github.com/ethereum-optimism/gateway/pull/485

roninjin10:
Looks like I copied this off internet from here https://github.com/yarnpkg/yarn/issues/5840#issuecomment-468782288 .  I was wondering why I wrote a bash script instead of a node script.

roninjin10:
my bad closed pr thinking another pr merged accomplishing this

trianglesphere:
> If ya'll go with the using git strategy, I'd recommend doing it as a CI check instead of making git a necessary dependency of the overall build pipeline.

git's already installed on most of the executors/docker images that we use. First step is usually to do a checkout

packages/contracts-periphery/package.json
```diff
@@ -25,6 +25,7 @@
     "gas-snapshot": "forge snapshot",
     "pretest:slither": "rm -f @openzeppelin && rm -f hardhat && ln -s node_modules/@openzeppelin @openzeppelin && ln -s ../../node_modules/hardhat hardhat",
     "posttest:slither": "rm -f @openzeppelin && rm -f hardhat",
+    "install:ci": "./scripts/yarn-lock-file.sh",
```
roninjin10, line 28:
```suggestion
    "postinstall": "git diff --exit-code",
```

roninjin10:
going to close for now I won't have time to really do anything here.  This is just a copy paste from gateway repo was only reason I was able to squeeze this pr.

PR #4493, opened 2022-12-14
op-node: Fix panic when p2p server fails to start

trianglesphere:
**Description**

I only set `OpNode.server` to the created server if it starts up successfully. This alleviates an error where if creation succeeds, but startup does not, we attempt to stop the server which panics.

This also adds logging to indicate that initialization failed. This logging primarily benefits users who call `node.New` directly rather than end users because there is already logging in the main function.

**Tests**

I tested this manually with op-e2e & was able to reproduce the error & confirm that it is fixed.

**Metadata**
- Fixes ENG-2863

### DIFF
--- a/op-node/node/node.go
+++ b/op-node/node/node.go
@@ -61,6 +61,7 @@ func New(ctx context.Context, cfg *Config, log log.Logger, snapshotLog log.Logge
 
 	err := n.init(ctx, cfg, snapshotLog)
 	if err != nil {
+		log.Error("Error intializing the rollup node", "err", err)
 		// ensure we always close the node resources if we fail to initialize the node.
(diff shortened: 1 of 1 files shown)


op-node/node/node.go
```diff
@@ -163,20 +164,21 @@ func (n *OpNode) initL2(ctx context.Context, cfg *Config, snapshotLog log.Logger
 
 func (n *OpNode) initRPCServer(ctx context.Context, cfg *Config) error {
 	var err error
```
tynes, line 166: Can probably remove this error declaration here now?

trianglesphere: yes, good catch

tynes (approved):

mslipper (approved):

PR #4494, opened 2022-12-15
op-node,op-e2e: optimize L1 receipts fetching by supporting different methods

protolambda:
**Description**

PR changes:
- Add "RPC provider kind: an enum that identifies different groups of L1 RPC providers, based on the different endpoints they offer
- Add "ReceiptsFetchingMethod": an enum, and bitfield, of the different receipts fetching methods that may be available.
- Make `sources.EthClient` initialize initially available methods based on the RPC kind specified by the user, and reduce those based on RPC errors. The user can specify "any" if they want the node to figure it out by trial and error. The default is "basic", which has the current behavior, so the user is not surprised by error/warning logs in their L1 node or op-node.
- Simplify `IterativeBatchCall` so that the `[]V -> result` conversion is separate, and simplify the receipts validation function signature (no need to return a function anymore)
- `receiptsFetchingJob` now performs the fetching work, and locks/caches/etc. per block like the receipts-fetcher did before, but now supports the different fetching methods.
- Add `l1.rpckind` flag to change the kind of RPC (options: alchemy, quicknode, infura, parity, nethermind, debug_geth, erigon, basic, any)

**Tests**

WIP. We can't test in CI against endpoint variants we don't have. I think it is thus sensible to default to what we can test in CI, and test the alternative methods by test-running against the different RPC providers.

I will change it from draft to review status once I've tested it against each kind of RPC provider.

**Metadata**

Fix ENG-2977

### DIFF
--- a/op-node/flags/flags.go
+++ b/op-node/flags/flags.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/ethereum-optimism/optimism/op-node/chaincfg"
+	"github.com/ethereum-optimism/optimism/op-node/sources"
 
 	"github.com/urfave/cli"
 )
@@ -63,11 +64,15 @@ var (
 		Usage:  "Trust the L1 RPC, sync faster at risk of malicious/buggy RPC providing bad or inconsistent L1 data",
 		EnvVar: prefixEnvVar("L1_TRUST_RPC"),
 	}
-	L1RPCProviderKind = cli.StringFlag{
-		Name:   "l1.rpckind",
-		Usage:  "The kind of RPC provider, used to inform optimal transactions receipts fetching, and thus reduce costs. Valid options: alchemy, quicknode, infura, parity, nethermind, debug_geth, erigon, basic, any.",
+	L1RPCProviderKind = cli.GenericFlag{
+		Name: "l1.rpckind",
+		Usage: "The kind of RPC provider, used to inform optimal transactions receipts fetching, and thus reduce costs. Valid options: " +
+			EnumString[sources.RPCProviderKind](sources.RPCProviderKinds),
 		EnvVar: prefixEnvVar("L1_RPC_KIND"),
-		Value:  "basic",
+		Value: func() *sources.RPCProviderKind {
+			out := sources.RPCKindBasic
+			return &out
+		}(),
 	}
 	L2EngineJWTSecret = cli.StringFlag{
 		Name:        "l2.jwt-secret",
--- a/op-node/flags/flags.go
+++ b/op-node/flags/flags.go
@@ -63,6 +63,12 @@ var (
 		Usage:  "Trust the L1 RPC, sync faster at risk of malicious/buggy RPC providing bad or inconsistent L1 data",
 		EnvVar: prefixEnvVar("L1_TRUST_RPC"),
 	}
+	L1RPCProviderKind = cli.StringFlag{
+		Name:   "l1.rpckind",
+		Usage:  "The kind of RPC provider, used to inform optimal transactions receipts fetching, and thus reduce costs. Valid options: alchemy, quicknode, infura, parity, nethermind, debug_geth, erigon, basic, any.",
+		EnvVar: prefixEnvVar("L1_RPC_KIND"),
+		Value:  "basic",
+	}
 	L2EngineJWTSecret = cli.StringFlag{
 		Name:        "l2.jwt-secret",
 		Usage:       "Path to JWT secret key. Keys are 32 bytes, hex encoded in a file. A new key will be generated if left empty.",
@@ -182,6 +188,7 @@ var optionalFlags = append([]cli.Flag{
 	RollupConfig,
 	Network,
 	L1TrustRPC,
+	L1RPCProviderKind,
 	L2EngineJWTSecret,
 	VerifierL1Confs,
 	SequencerEnabledFlag,
--- a/op-node/sources/eth_client.go
+++ b/op-node/sources/eth_client.go
@@ -67,7 +67,7 @@ func (c *EthClientConfig) Check() error {
 	if c.MaxRequestsPerBatch < 1 {
 		return fmt.Errorf("expected at least 1 request per batch, but max is: %d", c.MaxRequestsPerBatch)
 	}
-	if _, ok := RPCKinds[c.RPCProviderKind]; !ok {
+	if !ValidRPCProviderKind(c.RPCProviderKind) {
 		return fmt.Errorf("unknown rpc provider kind: %s", c.RPCProviderKind)
 	}
 	return nil
--- a/op-node/sources/eth_client.go
+++ b/op-node/sources/eth_client.go
@@ -3,7 +3,6 @@ package sources
 import (
 	"context"
 	"fmt"
-	"io"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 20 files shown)


op-node/sources/eth_client.go
```diff
@@ -106,14 +134,17 @@ func NewEthClient(client client.RPC, log log.Logger, metrics caching.Metrics, co
...
-		trustRPC:          config.TrustRPC,
-		log:               log,
-		receiptsCache:     caching.NewLRUCache(metrics, "receipts", config.ReceiptsCacheSize),
-		transactionsCache: caching.NewLRUCache(metrics, "txs", config.TransactionsCacheSize),
-		headersCache:      caching.NewLRUCache(metrics, "headers", config.HeadersCacheSize),
-		payloadsCache:     caching.NewLRUCache(metrics, "payloads", config.PayloadsCacheSize),
+		client:                  client,
+		maxBatchSize:            config.MaxRequestsPerBatch,
+		trustRPC:                config.TrustRPC,
+		mustBePostMerge:         config.MustBePostMerge,
```
protolambda, line 140: fwiw, `mustBePostMerge` was missing here, so I fixed it while adding `provKind`

protolambda:
Tested the alchemy + quicknode + debug-geth options with some beta-1 syncing, and after some fixes the fallbacks & optimization now work. Marking this PR ready for review. Lmk if anyone have ideas about automating the testing of the different RPC flavors in CI without rewriting the serving-side, I'm afraid we can't cover all options as easily.

op-node/sources/receipts.go
```diff
@@ -82,13 +91,387 @@ func makeReceiptRequest(txHash common.Hash) (*types.Receipt, rpc.BatchElem) {
...
+	RPCKindDebugGeth  RPCProviderKind = "debug_geth"
+	RPCKindErigon     RPCProviderKind = "erigon"
+	RPCKindBasic      RPCProviderKind = "basic" // try only the standard most basic receipt fetching
+	RPCKindAny        RPCProviderKind = "any"   // try any method available
+)
+
+func (kind RPCProviderKind) String() string {
+	return string(kind)
+}
+
```
tynes, line 125: You should make a type for `RPCKinds` and make it a stringer so that you can render it in the CLI `--help` command

op-node/flags/flags.go
```diff
@@ -63,6 +63,12 @@ var (
 		Usage:  "Trust the L1 RPC, sync faster at risk of malicious/buggy RPC providing bad or inconsistent L1 data",
 		EnvVar: prefixEnvVar("L1_TRUST_RPC"),
 	}
+	L1RPCProviderKind = cli.StringFlag{
+		Name:   "l1.rpckind",
+		Usage:  "The kind of RPC provider, used to inform optimal transactions receipts fetching, and thus reduce costs. Valid options: alchemy, quicknode, infura, parity, nethermind, debug_geth, erigon, basic, any.",
```
tynes, line 68: `RPCKinds.String()` would be helpful here to prevent this from getting out of date

op-node/sources/receipts.go
```diff
@@ -82,13 +91,387 @@ func makeReceiptRequest(txHash common.Hash) (*types.Receipt, rpc.BatchElem) {
...
-// NewReceiptsFetcher creates a receipt fetcher that can iteratively fetch the receipts matching the given txs.
-func NewReceiptsFetcher(block eth.BlockID, receiptHash common.Hash, txHashes []common.Hash, getBatch BatchCallContextFn, batchSize int) eth.ReceiptsFetcher {
-	return NewIterativeBatchCall[common.Hash, *types.Receipt, types.Receipts](
-		txHashes,
-		makeReceiptRequest,
-		makeReceiptsFn(block, receiptHash),
-		getBatch,
-		batchSize,
-	)
+// Cost break-down sources:
```
tynes, line 94: A lot of code here. I think we want mocks of each endpoint

mslipper: I think we should test this by having a set of optional tests that take environment variables for actual RPC endpoints from each provider. That way we can make real queries, and if the provider behavior changes we'll know quickly.

trianglesphere (commented):
Looking pretty good. It's pretty large, but does a lot. It did take a me a little bit to understand why a lot of the complexity is there (particularly around the bitsets)

op-node/sources/receipts.go
protolambda: implemented a util instead, I imagine we have more enums we may want to show in flags.

op-node/flags/flags.go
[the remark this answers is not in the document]
protolambda: using cli.GenericFlag now to make it parse/validate the value

op-node/sources/receipts.go
```diff
@@ -82,13 +91,387 @@ func makeReceiptRequest(txHash common.Hash) (*types.Receipt, rpc.BatchElem) {
...
+}
+
+// ReceiptsFetchingMethod is a bitfield with 1 bit for each receipts fetching type.
+// Depending on errors, tx counts and preferences the code may select different sets of fetching methods.
+type ReceiptsFetchingMethod uint64
+
+func (r ReceiptsFetchingMethod) String() string {
+	out := ""
+	x := ReceiptsFetchingMethod(1)
+	for r != 0 && x != 0 {
```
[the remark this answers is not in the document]
protolambda: updated

op-node/sources/receipts.go
```diff
@@ -82,13 +91,387 @@ func makeReceiptRequest(txHash common.Hash) (*types.Receipt, rpc.BatchElem) {
...
+
+	// Other:
+	//  - 250 credits, not supported, strictly worse than other options. In quicknode price-table.
+	// qn_getBlockWithReceipts - in price table, ? undocumented, but in quicknode "Single Flight RPC" description
+	// qn_getReceipts          - in price table, ? undocumented, but in quicknode "Single Flight RPC" description
+	// debug_getBlockReceipts  - ? undocumented, shows up in quicknode price table, not available.
+)
+
+// AvailableReceiptsFetchingMethods selects receipt fetching methods based on the RPC provider kind.
+// The results are ordered in priority. If any fetching types error frequently, other types may be attempted.
```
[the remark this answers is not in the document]
protolambda: it was combined into one method, but when I split out `PickBestReceiptsFetchingMethod` I forgot to update the doc, thanks.

op-node/sources/receipts.go
```diff
@@ -82,13 +91,387 @@ func makeReceiptRequest(txHash common.Hash) (*types.Receipt, rpc.BatchElem) {
...
+type ReceiptsRequester interface {
+	PickReceiptsMethod(txCount uint64) ReceiptsFetchingMethod
+	OnReceiptsMethodErr(m ReceiptsFetchingMethod, err error)
+}
+
+// runFetcher retrieves the result by continuing previous batched receipt fetching work,
+// and starting this work if necessary.
+func (job *receiptsFetchingJob) runFetcher(ctx context.Context) error {
+	if job.fetcher == nil {
+		// continue existing work
```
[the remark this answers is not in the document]
protolambda: yea, mixed up comments, my bad

mslipper (approved):
I think this looks good to merge. Would love to see some automation around each provider. If we mark the tests as optional, we can pass in real RPC URLs for each provider and integration test them. I don't think there's much value in mocking the functionality of each provider.

protolambda:
Will rebase on develop again to make CI work. And added tests + minor fixes in #4561 for further review.

op-node/sources/receipts.go
protolambda: Implemented the mocks / testing in #4561

PR #4502, opened 2022-12-15
test(ctb): OptimismMintableERC721 tests

smartcontracts:
**Description**
Ports tests for the OptimismMintableERC721 contract.


packages/contracts-bedrock/contracts/test/OptimismMintableERC721.t.sol
```diff
@@ -0,0 +1,126 @@
...
+        L2Token = new OptimismMintableERC721(
+            address(L2Bridge),
+            1,
+            address(L1Token),
+            "L2Token",
+            "L2T"
+        );
+
+        // Label the addresses for nice traces.
+        vm.label(address(L2Token), "L2Token");
```
tynes, line 34: Not sure if we use the same label for the erc20 L1/L2 tokens, might want to be more specific `L1ERC721Token`

tynes (approved):

PR #4506, opened 2022-12-15
fix(sdk): browser compatibility

smartcontracts:
**Description**
Updates the SDK for browser compatibility. Requires a few tweaks in core-utils.

### DIFF
--- a/.changeset/chilly-pets-rescue.md
+++ b/.changeset/chilly-pets-rescue.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/sdk': patch
+---
+
+Updates the SDK for full browser compatibility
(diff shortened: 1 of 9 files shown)


roninjin10 (commented):
Got a lot of questions that make it hard for me to approve this pr as making the code base better

1. Who is trying to do batch encoding in the browser?
2. Is the sdk actually meant to be used in the browser?   Is that really it's intention
3. Stuff like this is never done in source code.  Standard best practice is to use a build tool to pollyfill.
4. Who is complaining about browser compatibility and why aren't they polyfilling it themselves?  If this is an issue request I would simply close the issue as 'won't do'
5. In general, the crypto industry makes this mistake consistently of not doing a good job of seperating out browser from node.js as platforms.   This looks to me like another example of hacking a nodejs library to make it more usable in the browser.
6. This isn't a complete pr.  There are more libraries that don't work in the browser.   `http` `https` `buffer` `stream` `zlib` is as far as I'm aware the complete list

roninjin10:
Random other thing, frontend build tools even with this change will still fail to build because they throw an error early when they see the require.

PR #4508, opened 2022-12-15
op-node: guard against panics in the p2p gossip validator

protolambda:
**Description**

Add a simple util function to wrap a gossip validator function  with a guard that can recover from panics, log them, and reject the messages that caused them.

**Tests**

Added a test case to cover the new util function.

**Invariants**

Gossip validation should never unexpectedly panic, but if it does, we want to reject the message from the peer that caused the panic.

**Additional context**

This prevents the node from falling over in case of an unexpected problem in the gossip validation pipeline.

**Metadata**

Fix ENG-3114


trianglesphere (approved):
this is sweet

PR #4515, opened 2022-12-16
feat(ctb): `L2OutputOracle` monotonic block number increase invariant

clabby:
# Overview

Adds an echidna invariant test for the `L2OutputOracle` that ensures that the L2 block number of output root proposals monotonically increases.

### DIFF
--- a/packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
+++ b/packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
@@ -1,0 +1,113 @@
+pragma solidity 0.8.15;
+
+import { L2OutputOracle } from "../L1/L2OutputOracle.sol";
+import { Types } from "../libraries/Types.sol";
+
+/**
+ * @dev Modified version of `L2OutputOracle` that contains a less restrictive
+ * version `proposeL2Output` for use with echidna.
+ */
+contract ModL2OutputOracle is L2OutputOracle {
+    constructor(
+        uint256 _submissionInterval,
+        uint256 _l2BlockTime,
+        uint256 _startingBlockNumber,
+        uint256 _startingTimestamp,
+        address _proposer,
+        address _challenger
+    ) L2OutputOracle(
+        _submissionInterval,
+        _l2BlockTime,
+        _startingBlockNumber,
+        _startingTimestamp,
+        _proposer,
+        _challenger
+    ) {
+        // ...
+    }
+
+    /**
+     * @dev This function is a modified version of `proposeL2Output` that removes
+     * several checks which limit echidna.
+     *
+     * Because the purpose of this test is to check that the block number of the
+     * proposed outputs must monotonically increase, these checks are safe to remove.
+     */
+    function proposeLessChecks(
+        bytes32 _outputRoot,
+        uint256 _l2BlockNumber
+    ) external {
+        require(
+            _l2BlockNumber == nextBlockNumber(),
+            "L2OutputOracle: block number must be equal to next expected block number"
+        );
+
+        emit OutputProposed(_outputRoot, nextOutputIndex(), block.timestamp, _l2BlockNumber);
+
+        l2Outputs.push(
+            Types.OutputProposal({
+                outputRoot: _outputRoot,
+                timestamp: uint128(block.timestamp),
+                l2BlockNumber: uint128(_l2BlockNumber)
+            })
+        );
+    }
+}
+
+contract EchidnaFuzzL2OutputOracle {
+    // ENV
+    uint256 internal immutable submissionInterval = 1800;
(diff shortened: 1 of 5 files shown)


clabby:
* develop
  * **PR #4515** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4515" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4515?utm_source=stack-comment).

packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
```diff
@@ -0,0 +1,113 @@
...
+                timestamp: uint128(block.timestamp),
+                l2BlockNumber: uint128(_l2BlockNumber)
+            })
+        );
+    }
+}
+
+contract EchidnaFuzzL2OutputOracle {
+    // ENV
+    uint256 internal immutable submissionInterval = 1800;
```
tynes, line 59: I never knew that you could assign to an `immutable` outside of the constructor, thought it would have to be `constant`.

packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
```diff
@@ -0,0 +1,109 @@
+pragma solidity 0.8.15;
+
+import { L2OutputOracle } from "../L1/L2OutputOracle.sol";
+import { Types } from "../libraries/Types.sol";
+
+/**
+ * @dev Modified version of `L2OutputOracle` that contains a less restrictive
+ * version `proposeL2Output` for use with echidna.
+ */
+contract ModL2OutputOracle is L2OutputOracle {
```
maurelian, line 10: Let's stick, with a similar naming convention to the other echidna targets.

packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
clabby: Ah yeah this works- immutables can either be defined directly or within the constructor. Will swap over to constants to follow the convention used in the rest of the codebase.

packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
```diff
@@ -0,0 +1,109 @@
...
+            _proposer,
+            _challenger
+        )
+    {
+        // ...
+    }
+
+    /**
+     * @dev This function is a modified version of `proposeL2Output` that removes
+     * several checks which limit echidna.
```
maurelian, line 33: Could we restore the other checks that aren't based on `msg.sender` just to minimize the diff?

packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
```diff
@@ -0,0 +1,109 @@
...
+            Types.OutputProposal({
+                outputRoot: _outputRoot,
+                timestamp: uint128(block.timestamp),
+                l2BlockNumber: uint128(_l2BlockNumber)
+            })
+        );
+    }
+}
+
+contract EchidnaFuzzL2OutputOracle {
```
maurelian, line 56: oh, I see now (re: my comment above about naming).

trianglesphere (approved):
CI is good.

packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
```diff
@@ -0,0 +1,109 @@
...
+        if (oracle.latestBlockNumber() < previousLatestBlock) {
+            blockNumberMonotonicallyIncreases = false;
+        }
+    }
+
+    /**
+     * @dev Delete one or multiple L2 outputs from the L2 Output Oracle.
+     */
+    function deleteL2Output(uint256 _index) external {
+        oracle.deleteL2Outputs(_index);
```
maurelian, line 100: how does this get around the auth?

packages/contracts-bedrock/contracts/echidna/FuzzOracle.sol
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Invariant: The block number of output root proposals should monotonically increase.
+     */
+    function echidna_blockNumberMonotonicallyIncreases() public view returns (bool) {
+        return blockNumberMonotonicallyIncreases;
```
maurelian, line 107: comment only: I actually like this a lot more (ie. true if the invariant holds), than the opposite which is used in the other echidna tests. 
Would be nice to eventually move to that convention everywhere.

clabby:
Closing in favor of the `forge` version of this invariant test in #4573

PR #4518, opened 2022-12-16
op-bindings: add test coverage of the L1 fee constants

tynes:
**Description**

`TestL1BlockSlots` will ensure that the values in the committed storage layout matches the values that are hard coded in `op-geth`. This is important to handle because the storage layout may accidentally change in the future. This will make it very obvious is there are issues with the storage layout changing.


trianglesphere (approved):
nice test

PR #4527, opened 2022-12-18
ops: Better check-changed

mslipper:
Improves check-changed:

- `check-changed` is now just a Python script in `ops`, so it doesn't require a Docker build + release anymore to release changes to it.
- All changes to `.circleci` will trigger a complete rebuild.
- `check-changed` is now a standalone command in the CCI config, so it's easier to use elsewhere.
- The script now calls `circleci step halt` from within rather than as part of this weird `|| echo "FALSE"` thing we used to have.
- The script is wrapped in a global `try...except` handler. This means that errors in the script itself will cause a complete rebuild.
- If the GH token isn't present (e.g., on forked builds) it requires a complete rebuild.
- Random commits to branches that don't have an associated PR will no longer cause a full rebuild. Unfortunately, CCI doesn't let us do this natively - you can either build your default branch + PRs, or all commits. This change should save us some money. If you need a build, make a draft PR.


ops/check-changed/main.py
```diff
@@ -0,0 +1,126 @@
...
+
+    g = Github(gh_token)
+    try:
+        g.get_user()
+        repo = g.get_repo(os.getenv('CIRCLE_PROJECT_USERNAME') + '/' + os.getenv('CIRCLE_PROJECT_REPONAME'))
+    except Exception:
+        log.exception('Failed to get repo from GitHub')
+        exit_build()
+
+    pr = repo.get_pull(int(pr_urls[0].split('/')[-1]))
```
tynes, line 81: Could `pr` ever be `None` here? resulting in `pr.{base,head}` not existing? Or if `repo.get_pull` throws if it doesn't exist?

tynes: Ah i see the exception handler at the bottom that would catch this

tynes (approved):
Looks good to me

.circleci/config.yml
```diff
@@ -379,6 +356,8 @@ jobs:
       image: ubuntu-2204:2022.07.1
     steps:
       - checkout
+      - check-changed:
+          patterns: .*\.md$
```
trianglesphere, line 360: Would we want to use a check against the specs folder instead? This looks like it works, but IIRC the lints we use only apply to the specs folder (so we run just a tad more than intended).

trianglesphere (approved):
One small q, but otherwise lgtm

.circleci/config.yml
mslipper: Yeah good point, will update

trianglesphere:
It looks like test p2p discovery flaked. I've seen it flake once or twice before

mslipper:
Yeah. I'll re-run.

PR #4528, opened 2022-12-18
op-heartbeat: add missing golang-lru module

islishude:
Run `go mod tidy` in op-heartbeat package.


tynes (approved):
Confirming that running `go mod tidy` produces the same output on my machine

PR #4529, opened 2022-12-19
op-node: make p2p sequencer addr rollup runtime configurable

protolambda:
**Description**

To rotate the p2p block signing key we can load it from the system config contract on L1.
We can do so with a storage read, and verify it against the state-root of the block we choose to read from.
Ideally we couple this to block-derivation later, so the block matches the view of the L1 chain used to derive the L2 chain.

This PR:
- adds a runtime config struct to maintain this type of configurable in
- adds a tartup routine to try load the runtime config a few times
- exposes runtime config to p2p validation with interface
- adds storage-proof support to the account result verification of the proof method in our eth client bindings
- cleans up get-proof API method (argument order consistent with RPC definition and similar getStorageAt method)
- add get-storage-at client method as fallback (Erigon + Reth don't support getProof due to DB design choices, and difficulty serving proofs for older states than the latest)
- add convenience method to read storage, and verify if we don't trust the RPC
- add case to derivation to ignore the event, avoid derivation from hanging on this new systemconfig event
- remove p2p sequencer address from rollup config

**Tests**

- Added a unit-test for the account proof storage verification. 
- All e2e tests start the op-node and hit this code-path of loading the p2p key from system config upon startup

**Additional context**

Will open a PR with current draft of extension to use receipts to keep updating the runtime config after start-up, although it's a bit chunky/complex.

A maybe nicer alternative would be to just periodically reload the runtime-config by calling `Load(current derivation L1 block hash)`.

**Metadata**

Fix ENG-3130


op-e2e/e2eutils/setup.go
```diff
@@ -207,7 +208,6 @@ func Setup(t require.TestingT, deployParams *DeployParams, alloc *AllocParams) *
 		ChannelTimeout:         deployConf.ChannelTimeout,
 		L1ChainID:              new(big.Int).SetUint64(deployConf.L1ChainID),
 		L2ChainID:              new(big.Int).SetUint64(deployConf.L2ChainID),
-		P2PSequencerAddress:    deployConf.P2PSequencerAddress,
```
tynes, line 210: We need to consider backwards compatibility here: if we remove the ability to configure this, then we will need to upgrade the proxy of the `SystemConfig` contract on the L1 networks that are already running

op-node/eth/account_proof.go
```diff
@@ -21,24 +27,68 @@ type AccountResult struct {
 	CodeHash    common.Hash    `json:"codeHash"`
 	Nonce       hexutil.Uint64 `json:"nonce"`
 	StorageHash common.Hash    `json:"storageHash"`
-	// storageProof field is ignored, we only need to proof the account contents,
-	// we do not access any individual storage values.
+
+	// Optional
+	StorageProof []StorageProofEntry `json:"storageProof,omitempty"`
 }
```
tynes, line 34: You might be able to use a much lighter type than `memorydb` for the storage proof

```go
// Custom type to write the generated proof to
type proofList [][]byte

func (n *proofList) Put(key []byte, value []byte) error {
	*n = append(*n, value)
	return nil
}

func (n *proofList) Delete(key []byte) error {
	panic("not supported")
}
```

This pattern is used in geth

op-node/node/runtime_config.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/p2p"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+)
+
+var (
+	// UnsafeBlockSignerAddressSystemConfigStorageSlot is the storage slot identifier of the unsafeBlockSigner
```
tynes, line 17: I'd like a test similar to `TestL1BlockSlots` in the following PR https://github.com/ethereum-optimism/optimism/commit/b232cf41278752acb8933a710d5265780479a669

This will ensure storage layout never changes, resulting in a bug. I do like the storage proof, having an `eth_call` would decouple the storage layout of the system config from the network's policy. Perhaps adding an `eth_call` and checking the value against the storage proof value, this would be a runtime check for the storage layout changing

op-node/eth/account_proof.go
```diff
@@ -21,24 +27,68 @@ type AccountResult struct {
...
 
-// Verify an account proof from the getProof RPC. See https://eips.ethereum.org/EIPS/eip-1186
+// Verify an account (and optionally storage) proof from the getProof RPC. See https://eips.ethereum.org/EIPS/eip-1186
 func (res *AccountResult) Verify(stateRoot common.Hash) error {
```
trianglesphere, line 36: The other approach is to use MPT prove from geth & is the approach used in the withdrawals proof package.
https://github.com/ethereum-optimism/optimism/blob/48f3f508871cc0ca20149480de1c137788f880e0/op-node/withdrawals/proof.go#L43-L100)

op-node/node/node.go
```diff
@@ -78,6 +80,9 @@ func (n *OpNode) init(ctx context.Context, cfg *Config, snapshotLog log.Logger)
 	if err := n.initL1(ctx, cfg); err != nil {
 		return err
 	}
+	if err := n.initRuntimeConfig(ctx, cfg); err != nil {
+		return err
```
trianglesphere, line 84: Maybe an error log here to suggest that the user should check the supplied L1 RPC url + connection? That seems like the most likely reason for `initRuntimeConfig` to fail.

trianglesphere (commented):
I left a couple small comments but nothing major.

op-node/node/runtime_config.go
tynes: Or a better solution is to decouple to storage layout using a magic value

tynes: We should consider using something like:

```solidity
bytes32(uint256(keccack256("systemconfig.unsafeblocksigner") - 1))
```

Also why is there the `- 1` ? Perhaps we shouldn't subtract one for simplicity

op-e2e/e2eutils/setup.go
mslipper: If you're referring to the beta, I wouldn't worry about it. We've already introduced backwards incompatible changes to the beta.

op-node/node/runtime_config.go
mslipper: 👍 on the magic value. This will require a contract change but is much less brittle.

op-node/eth/account_proof.go
protolambda: ` Get(key []byte) ([]byte, error)` and `Has(key []byte) (bool, error)` also have to be implemented then, to satisfy the `KeyValueReader` required by `VerifyProof`. And at that point I'm writing the same thing, but without the map of memorydb, making lookups less light.

protolambda: Updated the proof verification based on comments from you/josh. The proof is now verified in isolation with `trie.VerifyProof`, but using the db from geth to avoid re-implementing the Get/Has functionality.

op-node/eth/account_proof.go
protolambda: Updated

op-node/node/node.go
protolambda: I think it's better if we separate the L1 identity check, and then we can check chain ID and rollup gensis block presence etc.
I've opened a linear issue to track it: ENG-3134

protolambda:
Updated proof verification based on review suggestions, and rebased on develop for CI. Ready for review again.

trianglesphere:
@protolambda hive test failure is expected & fine, but will need some followup work to fix that.

trianglesphere (approved):
LGTM

trianglesphere:
Devnet is failing. Maybe a config issue.

mslipper:
I rebased this branch and the tests passed locally - I think that if you rebase @protolambda the devnet should turn green.

protolambda:
rebasing now

mslipper:
Merged in latest develop. Will squash + merge once CI passes.

PR #4531, opened 2022-12-19
op-node: read runtime config updates from l1 block receipts

protolambda:
Extends #4529, adds update method to runtime config to apply changes from receipts. Possible direction after minimal solution of refreshing the p2p key only during startup. We'd still need to couple it to fire for every L1 event, which is error-prone. Thus considering to reload the full config from L1 state periodically instead; it's only a single storage-read anyway, and we don't have to reload it that often.


op-node/node/runtime_config.go
```diff
@@ -78,3 +83,97 @@ func (r *RuntimeConfig) Load(ctx context.Context, l1Ref eth.L1BlockRef) error {
...
+
+// parseReceipts iterates receipts of the SystemConfig to find runtime configuration updates
+func parseReceipts(cfg *rollup.Config, receipts types.Receipts, dest *runtimeConfigData) error {
+	var result error
+	for i, rec := range receipts {
+		if rec.Status != types.ReceiptStatusSuccessful {
+			continue
+		}
+		for j, log := range rec.Logs {
+			if log.Address == cfg.L1SystemConfigAddress && len(log.Topics) > 0 && log.Topics[0] == derive.ConfigUpdateEventABIHash {
```
tynes, line 133: slight preference for reducing the size of this `if` statement with a helper function, something like `isConfigUpdate(log *types.Log) bool`

op-node/node/runtime_config.go
```diff
@@ -78,3 +83,97 @@ func (r *RuntimeConfig) Load(ctx context.Context, l1Ref eth.L1BlockRef) error {
...
+					result = multierror.Append(result, fmt.Errorf("malformatted L1 system sysCfg log in receipt %d, log %d: %w", i, j, err))
+				}
+			}
+		}
+	}
+	return result
+}
+
+// parseEvent is handles event logs originating from the L1 system config to update the runtime configuration
+func parseEvent(ev *types.Log, dest *runtimeConfigData) error {
```
tynes, line 144: I think testing this function would be nice because of its byte slicing

protolambda:
Closing this PR, I am not quite as convinced this is the right approach anymore, and the minimal set up in base PR works well enough for short term.

PR #4532, opened 2022-12-19
op-e2e: action test conflicting sequencers

protolambda:
**Description**

Test that verifier follows the sequenced block that was first included on-chain, and does not sync the alternative L2 blocks which would reorg the chain, in both unsafe-block and batch input cases.

**Metadata**

Fix ENG-2781


trianglesphere (approved):
LGTM

PR #4542, opened 2022-12-19
op-batcher/op-proposer: New package structure

trianglesphere:
**Description**

This organizes the op-batcher & op-proposer files a bit better. It places them in a single shared package rather than being a collection of flat files & a weird mix of small packages.

I intend to come back later & clean up some of the duplicated config / driver code.

**Additional Changes**

I had to slightly modify the op-e2e tests because of some structs needed new names because of duplicate names.

Note: we need cut a release of go modules after this is merged to use non-local replaces in op-e2e

**Metadata**
- Fixes ENG-3128


tynes (approved):
Generally lgtm

mslipper (approved):

protolambda (approved):

PR #4543, opened 2022-12-20
ctb: Live migration utilities

mslipper:
Adds some utilities to make it easier to perform a live migration:

- If a live deployer isn't configured, the SystemDictator will populate transactions for each step and output a JSON containing the transaction's `to`, `data`, `value`, and `chainId` fields. This gives us the opportunity to simulate these transactions prior to executing them, and makes it easier to generate multisignature transactions.
- The dynamic oracle config step requires user input. To make this easier, I added a Hardhat task that configures the dynamic oracle config. It operates in two modes. In send mode it will send the transaction directly. This is useful for the Goerli migration. In no send mode it outputs the raw transaction hex for simulation/multisig purposes. The task will confirm all input before sending any transactions.


packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -276,3 +276,12 @@ export const getDeploymentAddress = async (
   const deployment = await hre.deployments.get(name)
   return deployment.address
 }
+
+export const jsonifyTransaction = (tx: ethers.PopulatedTransaction): string => {
+  return JSON.stringify({
+    to: tx.to,
+    data: tx.data,
+    value: '0',
```
tynes, line 284: While none of the transactions send value, hardcoding this value could cause bugs for our future selves if things change in unexpected ways

packages/contracts-bedrock/deploy/020-SystemDictatorSteps.ts
```diff
@@ -151,8 +158,15 @@ const deployFn: DeployFunction = async (hre) => {
       console.log(`Setting L1CrossDomainMessenger owner to MSD`)
```
tynes, line 158: The `if (isLiveDeployer)` case hits for devnets so it makes sense not actually send the transactions in the other case because the other case is for upgrading live networks

packages/contracts-bedrock/tasks/update-dynamic-oracle-config.ts
```diff
@@ -0,0 +1,78 @@
...
+      console.log(tx.data)
+    } else {
+      console.log(`Sending transaction...`);
+      const tx = await SystemDictator.updateL2OutputOracleDynamicConfig({
+        l2OutputOracleStartingBlockNumber: l2OutputOracleStartingBlockNumber,
+        l2OutputOracleStartingTimestamp: l2OutputOracleStartingTimestamp,
+      });
+      console.log(`Transaction sent with hash ${tx.hash}. Waiting for receipt...`);
+      await tx.wait(1)
+      console.log(`Transaction included in block ${tx.blockNumber}`);
```
tynes, line 62: Will the `blockNumber` be updated from withing `wait()`? I usually do it this way:

```ts
const receipt = await tx.wait(1)
console.log(receipt.blockNumber)
```

tynes:
Looks like a lot of linting errors

mslipper:
Lint is fixed

packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -276,3 +276,12 @@ export const getDeploymentAddress = async (
   const deployment = await hre.deployments.get(name)
   return deployment.address
 }
+
+export const jsonifyTransaction = (tx: ethers.PopulatedTransaction): string => {
```
tynes, line 280: do we want to stringify in a more human readable way, ie `JSON.stringify({}, nulll, 2)`?

packages/contracts-bedrock/tasks/update-dynamic-oracle-config.ts
```diff
@@ -0,0 +1,98 @@
+import readline from 'readline'
+
+import { task, types } from 'hardhat/config'
+import { ethers, Wallet } from 'ethers'
+
+import { getContractsFromArtifacts } from '../src/deploy-utils'
+
+task('update-dynamic-oracle-config', 'Updates the dynamic oracle config.')
```
tynes, line 8: Do we want a `--force` option to skip the prompt?

packages/contracts-bedrock/src/deploy-utils.ts
mslipper: Sure, I can do this

packages/contracts-bedrock/tasks/update-dynamic-oracle-config.ts
mslipper: I think the prompt is important. We'll never run this in CI, and it's important that these values are correct.

tynes: ie, running this script as part of our contract deployment to get test coverage. I think you can do `hre.run('update-dynamic-oracle-config')` from within another hh script to run it

tynes: In particular i mean the deployment script that deploys the system, this could replace the part of the script that sets the dynamic gas config

tynes (approved):

packages/contracts-bedrock/tasks/update-dynamic-oracle-config.ts
mslipper: Ah I see. Fair enough. Can update.

PR #4546, opened 2022-12-20
contracts-bedrock: deterministic storage

tynes:
**Description**

Makes the storage slot deterministic for the unsafe block signer. This is useful because the `op-node` uses a storage proof to fetch the value, so the `op-node` can compute the storage slot value locally. This decouples the way that `solc` lays out storage from the network policy, making bugs less likely to happen when changing the contracts.


smartcontracts (approved):

mslipper:
Added `mergify-ignore` so we can get Proto's PR in without breaking it

protolambda (approved):

mslipper:
Updated the storage slot in the op-node. This is now mergeable.

mslipper:
Hive will require the new bindings, so those failures are expected.

PR #4551, opened 2022-12-21
op-chain-ops: Fix state migration, add post-checks

mslipper:
This PR fixes a bug in `op-chain-ops` that could cause state to become erroneously erased during the L2 migration. While testing the Goerli migration in check mode, I encountered an issue where `db.GetState` was returning null storage slots for the legacy message passer during the withdrawals portion of the migration. Upon further inspection, I discovered that we were using `db.CreateAccount` in the `setProxies` method. `db.CreateAccount` will erase the state associated with the passed-in address, so I updated `setProxies` to check if the account exists first. I also added a map of predeploys that should _not_ be touched as part of the migration process at all. At the end of the migration process, I iterate over these "untouchable" contracts and make sure that their code matches a preset list of code hashes and sample up to 5,000 storage slots before and after the migration to make sure the match.

Additionally, I found a bug in the witness generation data within `l2geth`. The witness generation code for the message passer does not take reverts into account. There's no easy way to do this, since we don't have access to the call frame in which the revert occurred. To work around this, I added the ability to ignore reverted storage slots in the `ParamsByChainId` structure.

Lastly, I made a few miscellaneous changes to improve our confidence in the migration script:

- I added checks to the OVM ETH migration to ensure that state slots are properly preserved.
- I added a read cache to the underlying state DB so that repeated read calls are faster.

I have verified that this migration script works end-to-end with check mode enabled against a forked Goerli L1.

### DIFF
--- a/op-chain-ops/genesis/check.go
+++ b/op-chain-ops/genesis/check.go
@@ -1,12 +1,16 @@
 package genesis
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"math/big"
 
+	"github.com/ethereum-optimism/optimism/op-chain-ops/crossdomain"
+	"github.com/ethereum-optimism/optimism/op-chain-ops/genesis/migration"
+	"github.com/ethereum/go-ethereum/crypto"
+
 	"github.com/ethereum-optimism/optimism/op-bindings/predeploys"
-	"github.com/ethereum-optimism/optimism/op-chain-ops/ether"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/state"
@@ -16,8 +20,25 @@ import (
 	"github.com/ethereum/go-ethereum/trie"
 )
 
-// CheckMigratedDB will check that the migration was performed correctly
-func CheckMigratedDB(ldb ethdb.Database) error {
+// MaxSlotChecks is the maximum number of storage slots to check
+// when validating the untouched predeploys. This limit is in place
+// to bound execution time of the migration. We can parallelize this
+// in the future.
+const MaxSlotChecks = 5000
+
+var LegacyETHCheckSlots = map[common.Hash]common.Hash{
+	// Bridge
+	common.Hash{31: 0x06}: common.HexToHash("0x0000000000000000000000004200000000000000000000000000000000000010"),
+	// Symbol
+	common.Hash{31: 0x04}: common.HexToHash("0x4554480000000000000000000000000000000000000000000000000000000006"),
+	// Name
+	common.Hash{31: 0x03}: common.HexToHash("0x457468657200000000000000000000000000000000000000000000000000000a"),
+	// Total supply
+	common.Hash{31: 0x02}: {},
+}
+
+// PostCheckMigratedDB will check that the migration was performed correctly
+func PostCheckMigratedDB(ldb ethdb.Database, migrationData migration.MigrationData, l1XDM *common.Address, l1ChainID uint64) error {
 	log.Info("Validating database migration")
 
 	hash := rawdb.ReadHeadHeaderHash(ldb)
@@ -30,6 +51,13 @@ func CheckMigratedDB(ldb ethdb.Database) error {
 	header := rawdb.ReadHeader(ldb, hash, *num)
 	log.Info("Read header from database", "number", *num)
 
+	if !bytes.Equal(header.Extra, bedrockTransitionBlockExtraData) {
+		return fmt.Errorf("expected extra data to be %x, but got %x", bedrockTransitionBlockExtraData, header.Extra)
+	}
+
+	prevHeader := rawdb.ReadHeader(ldb, header.ParentHash, *num-1)
+	log.Info("Read previous header from database", "number", *num-1)
+
 	underlyingDB := state.NewDatabaseWithConfig(ldb, &trie.Config{
 		Preimages: true,
 	})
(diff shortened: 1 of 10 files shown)


op-chain-ops/genesis/setters.go
```diff
@@ -14,6 +14,26 @@ import (
...
+	predeploys.LegacyMessagePasserAddr: true,
+	predeploys.DeployerWhitelistAddr:   true,
+	predeploys.GovernanceTokenAddr:     true,
+	predeploys.L1BlockNumberAddr:       true,
+	predeploys.WETH9Addr:               true,
+}
+
+// UntouchableCodeHashes contains code hashes of all the contracts
+// that should not be touched by the migration process.
+var UntouchableCodeHashes = map[common.Address]common.Hash{
```
tynes, line 29: Did you pull these code hashes from goerli? There exists the possibility that mainnet has slightly different code. Unlikely but it's possible because I can't remember if the compiler version got changed

op-chain-ops/genesis/setters.go
```diff
@@ -14,6 +14,26 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
+// UntouchablePredeploys are addresses in the predeploy namespace
+// that should not be touched by the migration process.
```
tynes, line 18: I think this will break the check-l2 hh task because it checks for proxies at some of these accounts

op-chain-ops/genesis/setters.go
```diff
@@ -74,43 +103,32 @@ func SetImplementations(db vm.StateDB, storage state.StorageConfig, immutable im
...
-			log.Info("Setting deployed bytecode with immutables", "name", name, "address", addr)
-			db.SetCode(addr, bytecode)
-		} else {
-			depBytecode, err := bindings.GetDeployedBytecode(name)
-			if err != nil {
-				return err
-			}
-			log.Info("Setting deployed bytecode from solc compiler output", "name", name, "address", addr)
-			db.SetCode(addr, depBytecode)
+		// Proxy admin is a special case - it needs an impl set, but at its own address
```
tynes, line 119: Perhaps but this behind an exists check just in case, unlikely to be a problem

op-chain-ops/genesis/check.go
```diff
@@ -16,8 +18,25 @@ import (
...
+	// Symbol
+	common.Hash{31: 0x04}: common.HexToHash("0x4554480000000000000000000000000000000000000000000000000000000006"),
+	// Name
+	common.Hash{31: 0x03}: common.HexToHash("0x457468657200000000000000000000000000000000000000000000000000000a"),
+	// Total supply
+	common.Hash{31: 0x02}: {},
+}
+
+// PostCheckMigratedDB will check that the migration was performed correctly
+func PostCheckMigratedDB(ldb ethdb.Database, migrationData migration.MigrationData, l1XDM *common.Address, l1ChainID int) error {
```
tynes, line 39: Another good check here would be checking that the tip's blockheader extra data is the magic value we place there. There is a constant defined for what the value should be

op-chain-ops/genesis/check.go
```diff
@@ -39,22 +61,73 @@ func CheckMigratedDB(ldb ethdb.Database) error {
...
+	if err != nil {
+		return fmt.Errorf("cannot open StateDB: %w", err)
+	}
+
+	for addr := range UntouchablePredeploys {
+		// Check that the code is the same.
+		code := currDB.GetCode(addr)
+		hash := crypto.Keccak256Hash(code)
+		if hash != UntouchableCodeHashes[addr] {
+			return fmt.Errorf("expected code hash for %s to be %s, but got %s", addr, UntouchableCodeHashes[addr], hash)
```
tynes, line 100: Slightly worried this will need a per network basis

op-chain-ops/genesis/check.go
```diff
@@ -66,46 +139,101 @@ func CheckPredeploys(db vm.StateDB) error {
 			return fmt.Errorf("no code found at predeploy %s", addr)
 		}
 
+		if UntouchablePredeploys[addr] {
```
tynes, line 142: I think a trace log would be good here

op-chain-ops/genesis/check.go
```diff
@@ -66,46 +139,101 @@ func CheckPredeploys(db vm.StateDB) error {
...
 		}
 	}
 
 	// For each predeploy, check that we've set the implementation correctly when
 	// necessary and that there's code at the implementation.
 	for _, proxyAddr := range predeploys.Predeploys {
-		implAddr, special, err := mapImplementationAddress(proxyAddr)
-		if err != nil {
-			return err
+		if UntouchablePredeploys[*proxyAddr] {
```
tynes, line 157: Trace loglines on the continue statements

op-chain-ops/genesis/check.go
```diff
@@ -66,46 +139,101 @@ func CheckPredeploys(db vm.StateDB) error {
...
 		}
 
-		implCode := db.GetCode(implAddr)
+		implCode := db.GetCode(expImplAddr)
 		if len(implCode) == 0 {
 			return fmt.Errorf("no code found at predeploy impl %s", *proxyAddr)
 		}
+
+		if expImplAddr == predeploys.ProxyAdminAddr {
+			continue
```
tynes, line 182: Can this continue be hoisted to where the assignment happens instead of having this branch below?

op-chain-ops/genesis/migration/params.go
```diff
@@ -0,0 +1,51 @@
...
+		// / during the regenesis. A further 0.0012 ETH was incorrectly not removed from the
+		// total supply by accidental invocations of the Saurik bug (https://www.saurik.com/optimism.html).
+		new(big.Int).SetUint64(1627270011999999992),
+
+		map[common.Hash]bool{},
+	},
+	5: {
+		new(big.Int),
+
+		// The below reverted messages are current as of 12-21-2022. We may encounter more of these as
```
tynes, line 37: We probably want to update the witness generation scripts to handle this

op-chain-ops/genesis/setters.go
mslipper: I pulled from Goerli. Will check mainnet.

op-chain-ops/genesis/setters.go
mslipper: I'll update the check-l2 script to validate that implementations exist for these rather than proxies - this will make it match the real networks more.

op-chain-ops/genesis/migration/params.go
mslipper: This is removed now as per our discussion with Kelvin this morning.

op-chain-ops/genesis/setters.go
mslipper: This is done.

op-chain-ops/genesis/check.go
mslipper: Updated.

op-chain-ops/genesis/setters.go
mslipper: Updated to pull form both Goerli and mainnet.

op-chain-ops/genesis/check.go
```diff
@@ -39,22 +67,74 @@ func CheckMigratedDB(ldb ethdb.Database) error {
...
+		// Check that the code is the same.
+		code := currDB.GetCode(addr)
+		hash := crypto.Keccak256Hash(code)
+		expHash := UntouchableCodeHashes[addr][l1ChainID]
+		if hash != expHash {
+			return fmt.Errorf("expected code hash for %s to be %s, but got %s", addr, expHash, hash)
+		}
+		log.Info("checked code hash", "address", addr, "hash", hash)
+
+		// Sample storage slots to ensure that they are not modified.
```
maurelian, line 111: Could we also achieve this by checking the account's storage root?

op-chain-ops/genesis/check.go
```diff
@@ -66,46 +146,138 @@ func CheckPredeploys(db vm.StateDB) error {
...
+
+func CheckWithdrawalsAfter(db vm.StateDB, data migration.MigrationData, l1CrossDomainMessenger *common.Address) error {
+	wds, err := data.ToWithdrawals()
+	if err != nil {
+		return err
+	}
+
+	// First, make a mapping between old withdrawal slots and new ones.
+	// This list can be a superset of what was actually migrated, since
+	// some witness data may references withdrawals that reverted.
```
maurelian, line 223: What does it mean for a withdrawal to have reverted?

op-chain-ops/genesis/check.go
```diff
@@ -66,46 +146,138 @@ func CheckPredeploys(db vm.StateDB) error {
...
+			return fmt.Errorf("cannot compute migrated storage slot: %w", err)
+		}
+
+		oldToNew[legacySlot] = migratedSlot
+	}
+
+	// Now, iterate over each legacy withdrawal and check if there is a corresponding
+	// migrated withdrawal.
+	var innerErr error
+	err = db.ForEachStorage(predeploys.LegacyMessagePasserAddr, func(key, value common.Hash) bool {
```
maurelian, line 246: I think this might be an idea that came up today, but we could get some nice redundancy by verifying the state of the message pass are using the opposite approach, i.e.  precomputing the storage root of the MessagePasser by adding all the storage values into the trie then comparing, then comparing that to the post-migration root.

op-chain-ops/genesis/check.go
mslipper: Yeah, good idea. Will see about implementing that.

op-chain-ops/genesis/check.go
mslipper: If someone attempts a withdrawal, but the transaction reverts, the witness data will still be written. This means that the storage slot referenced by the witness data won't exist in the state, since the state changes get rolled back as a result of the revert.

op-chain-ops/genesis/check.go
mslipper: Yep, can do this. Will take a look.

op-chain-ops/genesis/check.go
mslipper: Added. I left sampling in there for added safety, but reduced the max samples.

op-chain-ops/genesis/check.go
maurelian: FWIW it feels like a fairly large expansion of scope, so it could also be done in a new PR.

smartcontracts (approved):
LGTM

op-chain-ops/genesis/check.go
mslipper: This is proving more difficult than anticipated - will punt on this for later.

PR #4553, opened 2022-12-21
feat(ctb): break out system dictator steps

smartcontracts:
**Description**
Breaks out the SystemDictator steps into two files so that the steps that get executed before the system is shut down can easily be separated from the steps that executed after the system gets spun back up. Idea here is that you would execute with the tag "steps1", wait for the legacy system to come down, migrate the L2, then execute with the tags "steps2" to complete the upgrade.

### DIFF
--- a/packages/contracts-bedrock/deploy/020-SystemDictatorSteps-1.ts
+++ b/packages/contracts-bedrock/deploy/020-SystemDictatorSteps-1.ts
@@ -1,0 +1,279 @@
+import assert from 'assert'
+
+import { ethers } from 'ethers'
+import { DeployFunction } from 'hardhat-deploy/dist/types'
+import { awaitCondition } from '@eth-optimism/core-utils'
+import '@eth-optimism/hardhat-deploy-config'
+import 'hardhat-deploy'
+import '@nomiclabs/hardhat-ethers'
+
+import {
+  assertContractVariable,
+  getContractsFromArtifacts,
+  getDeploymentAddress,
+  doStep,
+  jsonifyTransaction,
+} from '../src/deploy-utils'
+
+const deployFn: DeployFunction = async (hre) => {
+  const { deployer } = await hre.getNamedAccounts()
+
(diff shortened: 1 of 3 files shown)


mslipper (approved):

mslipper:
LGTM, just need to fix the linter

packages/contracts-bedrock/deploy/020-SystemDictatorSteps-1.ts
```diff
@@ -0,0 +1,279 @@
...
+    checks: async () => {
+      assert(
+        (await AddressManager.getAddress('OVM_L1CrossDomainMessenger')) ===
+          ethers.constants.AddressZero
+      )
+    },
+  })
+}
+
+deployFn.tags = ['SystemDictatorSteps', 'steps1']
```
maurelian, line 277: `steps1` is too similar to the function names in the dictator. I'd suggest something like `phase1`/`phase2`, or `pre-shutdown`/`post-spinup`.

packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -289,3 +289,61 @@ export const jsonifyTransaction = (tx: ethers.PopulatedTransaction): string => {
...
+ * Mini helper for executing a given step.
+ *
+ * @param opts Options for executing the step.
+ * @param opts.step Step to execute.
+ * @param opts.message Message to print before executing the step.
+ * @param opts.checks Checks to perform after executing the step.
+ */
+export const doStep = async (opts: {
+  isLiveDeployer?: boolean
+  SystemDictator: ethers.Contract
```
maurelian, line 317: these two are undocumented above.

packages/contracts-bedrock/deploy/020-SystemDictatorSteps-1.ts
smartcontracts: Good point

PR #4555, opened 2022-12-21
[op-batcher / op-proposer] Provide context to signer function

mdehoog:
**Description**
Passes the `context.Context` from the op-batcher / op-proposer into the `SignerFn`. This allows the SignerFn to make remote requests to external tx-signing services, and gives the ability for the caller to cancel those requests if needed.

**Tests**
No logic changes implemented, just passing the context through.


trianglesphere (approved):
LGTM. I think you're using the same context for signing as well as in the opts, but that's fine for now - I think we're going to clean up the txmgr a bit.

PR #4561, opened 2022-12-22
op-node: L1 rpc receipts fetching updates

protolambda:
**Description**

This PR:
- adds random block & receipts generation utils to the `testutils` package
- tweaks the receipts fetching logic:
  - gasused meta-data is now derived from cumulative gas used
  - added warning about contract address meta-data not being validated (we don't use it though)
  - remove parity methods from alchemy/quicknode fallbacks, since these methods are supposed to get deprecated.
  - make sure we maintain preference order for methods when using non-QN/alchemy RPC kinds
- implement receipt fetching tests:
  - create an in-process RPC server to serve requests with mock responses & assert that these requests happen
  - define test-cases for the happy cases and fallback cases of handling unavailable receipts methods

Also see #4494


trianglesphere:
> tweaks the receipts fetching logic:
gasused meta-data is now derived from cumulative gas used
added warning about contract address meta-data not being validated (we don't use it though)
remove parity methods from alchemy/quicknode fallbacks, since these methods are supposed to get deprecated.
make sure we maintain preference order for methods when using non-QN/alchemy RPC kinds

This are all really good changes, but shouldn't be in a tests PR

trianglesphere (commented):
testing changes lgtm. Need to look at the other tweaks

trianglesphere (approved):
lgtm

PR #4562, opened 2022-12-22
feat(ctb): check for relayed legacy messages

smartcontracts:
**Description**
Updates the CrossDomainMessenger base contract to check for relayed legacy messages. We originally specified that only unrelayed messages would be migrated as part of the bedrock migration, but we recently changed this to migrate all messages. As a result, we have to reintroduce checks into the CrossDomainMessenger contracts to block legacy messages that have already been relayed from being relayed again.


packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -263,15 +263,23 @@ abstract contract CrossDomainMessenger is
...
         (, uint16 version) = Encoding.decodeVersionedNonce(_nonce);
-
-        // Block any messages that aren't version 1. All version 0 messages have been guaranteed to
-        // be relayed OR have been migrated to version 1 messages. Version 0 messages do not commit
-        // to the value or minGasLimit fields, which can create unexpected issues for end-users.
         require(
-            version == 1,
-            "CrossDomainMessenger: only version 1 messages are supported after the Bedrock upgrade"
+            version == 0 || version == 1,
+            "CrossDomainMessenger: only version 0 or 1 messages are supported"
```
tynes, line 268: This could also be `version < 3`, its a `uint` so no need to worry about negative values

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -384,4 +385,302 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
...
+            0,
+            hex"1111"
+        );
+
+        // Set the value of op.l2Sender() to be the L2 Cross Domain Messenger.
+        vm.store(address(op), bytes32(senderSlotIndex), bytes32(abi.encode(sender)));
+
+        // Mark legacy message as already relayed.
+        bytes32 oldHash = Hashing.hashCrossDomainMessageV0(target, sender, hex"1111", 0);
+        bytes32 slot = keccak256(abi.encode(oldHash, uint256(203)));
```
tynes, line 450: Perhaps we turn the `203` into a constant variable or at least document it

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -384,4 +385,302 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
...
+        // Target should be called with expected data.
+        vm.expectCall(target, hex"1111");
+
+        // Expect RelayedMessage event to be emitted.
+        vm.expectEmit(true, true, true, true);
+        emit RelayedMessage(hash);
+
+        // Relay the message.
+        vm.prank(address(op));
+        L1Messenger.relayMessage(
```
tynes, line 416: We could update this call to use the object like syntax where its key value pairs, would remove the need to have `// value` and make it obvious what each arg is

tynes:
```solidity
L1Messenger.relayMessage({
  key: value,
  ...
})
```

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -384,4 +385,302 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
...
+        vm.prank(address(op));
+        L1Messenger.relayMessage{ value: value }(
+            Encoding.encodeVersionedNonce(0, 0), // nonce
+            sender,
+            target,
+            value,
+            0,
+            hex"1111"
+        );
+
```
tynes, line 508: Could add in an expect emit for `FailedRelayedMessage(versionedHash)` here

tynes: This could be added in each test where a reverter contract is used

tynes:
We will also need to make a change here:

https://github.com/ethereum-optimism/optimism/blob/509df58e90332dd6aa48866b3c69c1703fe3b8de/op-chain-ops/crossdomain/migrate.go#L66

We will no longer be setting the migrated withdrawals to version 1, they should be version 0

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
tynes: > Think you mean `version < 2`.

Yes this is what i meant here

tynes:
This generally looks good to me, I opened #4575 to accompany this change. We need to do end to end testing for this to be sure that it works but I think merging before then is fine

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
smartcontracts: Let's do that as part of a separate PR, it would be a style change we should be consistent about

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -384,4 +387,307 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
...
+            0,
+            0,
+            hex"1111"
+        );
+
+        // Set the value of op.l2Sender() to be the L2 Cross Domain Messenger.
+        vm.store(address(op), bytes32(senderSlotIndex), bytes32(abi.encode(sender)));
+
+        // Mark legacy message as already relayed.
+        uint256 successfulMessagesSlot = 203;
```
tynes, line 451: I've opened up an issue [here](https://github.com/foundry-rs/foundry/issues/4022) that would let us get rid of the magic numbers

maurelian:
I applied the `mergify-ignore` label as I think this should have 2 reviews prior to merging. I have looked at it and see no immediate issues, but I am planning to give this a more careful review in the morning.

tynes:
Looks like the bindings need to be regenerated again

tynes:
I've reviewed this pretty thoroughly and cannot find anything wrong with it

tynes:
I merged #4575 into this branch so be careful when force pushing @smartcontracts @maurelian

maurelian (approved):
LGTM

tynes (approved):

tynes:
This is ready for merge now

PR #4566, opened 2022-12-23
op-node: Fix too eager advance in batch queue

trianglesphere:
**Description**

This stops the batch queue from advancing to the next epoch before it should when it is generating blocks.

One result of the way I fixed this is that one of the assumptions in the batch queue - that the L2 safe head origin is the same as the origin of the batch queue is no longer valid.

**Tests**

This adds a regression test that I have confirmed fails without this fix. I had to modify several tests that where asserting incorrect behavior as well.

The fix also changed some assumptions which required some op-node test changes.

**Metadata**
- Fixes ENG-3145


op-node/rollup/derive/batch_queue.go
```diff
@@ -232,6 +232,10 @@ batchLoop:
 	// Fill with empty L2 blocks of the same epoch until we meet the time of the next L1 origin,
 	// to preserve that L2 time >= L1 time
 	if nextTimestamp < nextEpoch.Time {
+		if forceNextEpoch {
```
protolambda, line 235: isn't forceNextEpoch always true here? We already checked if it's false, and would have returned. Same for other checks below.

protolambda (commented):
Changes LGTM, nice work @trianglesphere and @clabby

op-e2e/actions/blocktime_test.go
```diff
@@ -12,6 +12,103 @@ import (
...
+		miner.ActL1StartBlock(4)(t)
+		miner.ActL1IncludeTx(sd.RollupCfg.Genesis.SystemConfig.BatcherAddr)(t)
+		miner.ActL1EndBlock(t)
+		sequencer.ActL1HeadSignal(t)
+		sequencer.ActL2PipelineFull(t)
+		makeL2BlockWithAliceTx()
+		makeL2BlockWithAliceTx()
+	}
+
+	verifyChainStateOnSequencer(8, 17, 8, 15, 7)
```
tynes, line 79: Would prefer to eliminate magic numbers here if possible. Perhaps assigning each to a nicely named variable with a comment would be nice

tynes: Creating variables seems a bit painful since this function is called multiple times, maybe this function should take a struct like python kwargs or at least add some comments as to why the values are correct

trianglesphere: I don't think variables or a struct would really help here. I do have some comments on what these values are, but could add more comments here.

trianglesphere: @tynes force pushed with better comments around this

protolambda (approved):

op-e2e/actions/blocktime_test.go
```diff
@@ -12,6 +12,108 @@ import (
...
+		miner.ActL1IncludeTx(sd.RollupCfg.Genesis.SystemConfig.BatcherAddr)(t)
+		miner.ActL1EndBlock(t)
+		sequencer.ActL1HeadSignal(t)
+		sequencer.ActL2PipelineFull(t)
+		makeL2BlockWithAliceTx()
+		makeL2BlockWithAliceTx()
+	}
+
+	// 8 L1 blocks with 17 L2 blocks is the unsafe state.
+	// Because wew consistently batch submitted we are one epoch behind the unsafe head with the safe head
```
tynes, line 80: typo `wew`

PR #4569, opened 2022-12-23
op-e2e: action test derivation with flaky L1 RPC

protolambda:
**Description**

Action-test that we can derive a L2 chain from a flaky L1 RPC (one that returns lots of temporary errors).

**Metadata**

Fix ENG-2776


trianglesphere (approved):
lgtm. This is sweet

PR #4570, opened 2022-12-23
op-node: allow sequencer to produce block with txs at sequencer time drift boundary

protolambda:
**Description**

The derivation batch filter allows batches with txs to pass through even for `origin.time + seq_time_drift == block.time`, so I updated the sequencer behavior to match + updated tests to handle updated sequencer behavior.
This PR makes no changes to the verifier derivation pipeline, so should be compatible with any existing testnet.

Specifically:
- derivation: https://github.com/ethereum-optimism/optimism/blob/c3493a023d250841e156dcee757a0e4874722953/op-node/rollup/derive/batches.go#L105
- sequencer: https://github.com/ethereum-optimism/optimism/blob/f1fc8e5c9856ffb84138eb460c9e379778402f47/op-node/rollup/driver/sequencer.go#L87

**Tests**

Updated action tests to reflect new sequencer behavior in the assertions.

**Metadata**

Fix ENG-3138


trianglesphere (approved):
This does match the verification logic laid out in the [batch validity rules](https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#batch-buffering); however, it is in conflict with [rules just below the overview](https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#overview).

In the interest of getting this through, I'm approving this. The specs will need to be reconciled, and if we choose to follow the rules in the overview this will need to be reverted.

PR #4573, opened 2023-01-03
feat(ctb): Forge invariant tests

clabby:
# Overview

Adds several simple invariant tests using forge's invariant fuzzer.

Bumps version of `contracts-bedrock` & `contracts-periphery`'s `forge-std` dep from `5bafa16b4a6aa67c503d96294be846a22a6f6efb` -> `53331f4cb2e313466f72440f3e73af048c454d02` to make use of the new `InvariantTest` contract.

**Metadata**
fixes ENG-3107


clabby:
* develop
  * **PR #4573** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4573" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4573?utm_source=stack-comment).

packages/contracts-bedrock/contracts/test/invariants/L2OutputOracle.t.sol
```diff
@@ -0,0 +1,89 @@
...
+            l2BlockTime,
+            startingBlockNumber,
+            startingTimestamp,
+            proposer,
+            owner
+        );
+    }
+
+    /**
+     * INVARIANT: The block number of the output root proposals should monotonically increase.
```
tynes, line 53: Do you mind updating this to use the custom tag from the other PR for consistency?

clabby: Will do after dinner! Must've not restacked- thanks for the spot 😄

packages/contracts-bedrock/contracts/test/invariants/L2OutputOracle.t.sol
```diff
@@ -0,0 +1,89 @@
...
+    address multisig = address(512);
+
+    // Test target
+    L2OutputOracle oracle;
+
+    L2ToL1MessagePasser messagePasser =
+        L2ToL1MessagePasser(payable(Predeploys.L2_TO_L1_MESSAGE_PASSER));
+
+    // Constructor arguments
+    address internal proposer = 0x000000000000000000000000000000000000AbBa;
```
tynes, line 19: These could technically all be constants

packages/contracts-bedrock/contracts/test/invariants/L2OutputOracle.t.sol
```diff
@@ -0,0 +1,89 @@
...
+        oracle = new L2OutputOracle(
+            submissionInterval,
+            l2BlockTime,
+            startingBlockNumber,
+            startingTimestamp,
+            proposer,
+            owner
+        );
+    }
+
```
tynes, line 51: Should we have a `targetSenders()([]address)` function that returns the proposer + challenger? (owner got renamed to challenger)

This way the permissioned actors are making the calls on the contract

tynes:
I don't think the foundrybook is updated with the special functions that invariant tests can have to configure them at runtime, i think `targetSenders` could be useful for us, are there any others that may be useful?

packages/contracts-bedrock/contracts/test/invariants/L2OutputOracle.t.sol
clabby: I actually had no idea this got merged yet- thanks for pointing it out! This is a game changer in the way we set these up- looking into the new features that maple upstreamed now 🚀

packages/contracts-bedrock/contracts/test/invariants/L2OutputOracle.t.sol
clabby: This was a direct copy of the `L2OutputOracle_Initializer` without the `CommonTest` inheritance- before we could target a specific contract (`InvariantTest.sol` in forge-std), it would run forever due to trying to hit the `FFIInterface` contract's functions. Should be able to remove this duplicated boilerplate, but can update the one in `CommonTest` 😄

packages/contracts-bedrock/contracts/test/invariants/OptimismPortal.t.sol
```diff
@@ -0,0 +1,165 @@
...
+    uint256 _proposedOutputIndex;
+    uint256 _proposedBlockNumber;
+    bytes32 _stateRoot;
+    bytes32 _storageRoot;
+    bytes32 _outputRoot;
+    bytes32 _withdrawalHash;
+    bytes[] _withdrawalProof;
+    Types.OutputRootProof internal _outputRootProof;
+
+    function setUp() public virtual override {
```
tynes, line 20: Just to make sure I understand

For each round:
- `setUp` is called
- A number of calls are made to the `targetContract`
- Each invariant is checked

tynes (approved):
Looks good to me pending my understanding is correct

packages/contracts-bedrock/package.json
```diff
@@ -77,7 +77,7 @@
     "dotenv": "^16.0.0",
     "ds-test": "https://github.com/dapphub/ds-test.git#9310e879db8ba3ea6d5c6489a579118fd264a3f5",
     "ethereum-waffle": "^3.0.0",
-    "forge-std": "https://github.com/foundry-rs/forge-std.git#5bafa16b4a6aa67c503d96294be846a22a6f6efb",
```
tynes, line 80: The `yarn.lock` wasn't updated?

packages/contracts-bedrock/contracts/test/invariants/OptimismPortal.t.sol
clabby: Yep! A high-level overview of how forge's invariant tests work behind the scenes can be found [here](https://github.com/mds1/solidity-sandbox/blob/main/test/4_InvariantNonceGoUp.t.sol#L19-L30)

packages/contracts-bedrock/package.json
clabby: It doesn't look like `contracts-bedrock` has a lockfile.

tynes: I though the global `yarn.lock` should be updated

clabby: Ah gotcha, wasn't updating automatically- manually bumped it 👍

tynes (approved):

tynes:
@mergifyio refresh

PR #4574, opened 2023-01-03
feat(ctb): Invariant docgen

clabby:
# Overview

Adds a script to generate invariant documentation for all `forge` & `echidna` invariant tests. 

**Metadata**
fixes ENG-3106

### DIFF
--- a/packages/contracts-bedrock/contracts/echidna/FuzzAddressAliasing.sol
+++ b/packages/contracts-bedrock/contracts/echidna/FuzzAddressAliasing.sol
@@ -24,7 +24,7 @@ contract EchidnaFuzzAddressAliasing {
     }
 
     /**
-     * INVARIANT: Address aliases are always able to be undone.
+     * @custom:inviariant Address aliases are always able to be undone.
      *
      * Asserts that an address that has been aliased with `applyL1ToL2Alias` can always
      * be unaliased with `undoL1ToL2Alias`.
--- a/packages/contracts-bedrock/contracts/echidna/FuzzAddressAliasing.sol
+++ b/packages/contracts-bedrock/contracts/echidna/FuzzAddressAliasing.sol
@@ -24,7 +24,10 @@ contract EchidnaFuzzAddressAliasing {
     }
 
     /**
-     * @notice Verifies that testRoundTrip(...) did not ever fail.
+     * INVARIANT: Address aliases are always able to be undone.
+     *
+     * Asserts that an address that has been aliased with `applyL1ToL2Alias` can always
+     * be unaliased with `undoL1ToL2Alias`.
      */
     function echidna_round_trip_aliasing() public view returns (bool) {
         // ASSERTION: The round trip aliasing done in testRoundTrip(...) should never fail.
--- a/packages/contracts-bedrock/contracts/echidna/FuzzAddressAliasing.sol
+++ b/packages/contracts-bedrock/contracts/echidna/FuzzAddressAliasing.sol
@@ -24,7 +24,7 @@ contract EchidnaFuzzAddressAliasing {
     }
 
     /**
-     * @custom:inviariant Address aliases are always able to be undone.
+     * @custom:invariant Address aliases are always able to be undone.
      *
      * Asserts that an address that has been aliased with `applyL1ToL2Alias` can always
      * be unaliased with `undoL1ToL2Alias`.
--- a/packages/contracts-bedrock/scripts/invariant-doc-gen.ts
+++ b/packages/contracts-bedrock/scripts/invariant-doc-gen.ts
@@ -1,8 +1,15 @@
 import fs from 'fs'
+import path from 'path'
 
-const BASE_INVARIANTS_DIR = `${__dirname}/../contracts/test/invariants`
-const BASE_ECHIDNA_DIR = `${__dirname}/../contracts/echidna`
-const BASE_DOCS_DIR = `${__dirname}/../invariant-docs`
+const BASE_INVARIANTS_DIR = path.join(
+  __dirname,
+  '..',
+  'contracts',
+  'test',
+  'invariants'
+)
+const BASE_ECHIDNA_DIR = path.join(__dirname, '..', 'contracts', 'echidna')
+const BASE_DOCS_DIR = path.join(__dirname, '..', 'invariant-docs')
 const BASE_ECHIDNA_GH_URL = '../contracts/echidna/'
 const BASE_INVARIANT_GH_URL = '../contracts/test/invariants/'
 const NATSPEC_INV = '@custom:invariant'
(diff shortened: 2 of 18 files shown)


clabby:
* clabby/ctb/invariant-tests
  * **PR #4574** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4574" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4574?utm_source=stack-comment).

packages/contracts-bedrock/contracts/echidna/FuzzAddressAliasing.sol
```diff
@@ -24,7 +24,10 @@ contract EchidnaFuzzAddressAliasing {
     }
 
     /**
-     * @notice Verifies that testRoundTrip(...) did not ever fail.
+     * INVARIANT: Address aliases are always able to be undone.
```
tynes, line 27: Perhaps we could use a custom tag here? `@custom:invariant`

Like this:
https://github.com/ethereum-optimism/optimism/blob/f7f6d76f882ec396c3d4ce89b2fc9cc26e7fc324/packages/contracts-bedrock/contracts/legacy/L1BlockNumber.sol#L11

clabby:
> This is an awesome idea, wondering if `forge doc` parses the custom tags as I know that other solidity tooling does parse them

Would be interesting to look into! I know `forge doc` is still [in progress](https://github.com/foundry-rs/foundry/pull/2701/files#diff-a0842ea73f3a116eff7958fbf833dd2280c366c0d8985aa05f8e1403e0f39273), but would be great to move towards using it once it's been merged.

It also focuses mainly on creating documentation for the contracts themselves rather than the tests, but I'm sure we could tweak it for our use cases without much hassle.

maurelian:
Love this. 
A couple thoughts: 
- We should run the docgen in ci and ensure there is no diff (similar to the gas snapshot). That shouldn't be too disruptive since new invariant tests won't be added all that often.
- Since most of the generated docs are quite short, WDYT about simplifying this by: 
	- generating a single file of invariants
	- moving the README content into main `contracts-bedrock` README. 
	
@tynes @smartcontracts would appreciate your thoughts on usability here.

clabby:
> Love this. A couple thoughts:
> 
> * We should run the docgen in ci and ensure there is no diff (similar to the gas snapshot). That shouldn't be too disruptive since new invariant tests won't be added all that often.
> * Since most of the generated docs are quite short, WDYT about simplifying this by:
>   
>   * generating a single file of invariants
>   * moving the README content into main `contracts-bedrock` README.
> 
> @tynes @smartcontracts would appreciate your thoughts on usability here.

Will add the first one 👍 

As for # 2:
* The only issue I see here is that it'd be a VERY long file eventually- having them split up is nice for quick indexing. Something we could do is auto-generate a table of contents with links to the section headers (each contract) if we want to go this route 😄 
* Can do! Could also retain the README in the `invariant-docs` folder and just add a small section to `contracts-bedrock`'s README that points towards it.

packages/contracts-bedrock/scripts/invariant-doc-gen.ts
```diff
@@ -0,0 +1,168 @@
+import fs from 'fs'
+
```
tynes, line 2: I'm fine with this not working on windows, but `path.join` could be useful to making this more portable. Everybody on the team develops with unix, altho there are some windows devs out there in the ethereum ecosystem

packages/contracts-bedrock/scripts/invariant-doc-gen.ts
```diff
@@ -0,0 +1,168 @@
...
+const docGen = (dir: string): void => {
+  // Grab all files within the invariants test dir
+  const files = fs.readdirSync(dir)
+
+  // Array to store all found invariant documentation comments.
+  const docs: Contract[] = []
+
+  for (const fileName of files) {
+    // Read the contents of the invariant test file.
+    const fileContents = fs.readFileSync(`${dir}/${fileName}`).toString()
```
tynes, line 42: This is a particular location that `path.join` makes sense

packages/contracts-bedrock/scripts/invariant-doc-gen.ts
```diff
@@ -0,0 +1,168 @@
...
+        }
+      }
+    }
+
+    // Add the contract to the array of docs
+    docs.push(contract)
+  }
+
+  for (const contract of docs) {
+    const fileName = `${BASE_DOCS_DIR}/${contract.name}.md`
```
tynes, line 107: Would it be annoying to use fully qualified names here? Ie `<filename>:<contractname>` or i'm fine with just enforcing that contract names must be unique

tynes: i think colons in filenames will break windows

tynes:
I'm fine with each contract getting its own file. Auto generating a table of contents would be a nice to have

I heard that dev work on `forge doc` is picking up again so wondering if we could migrate to using that for all of our docs  longer term and if it picks up on custom tags

tynes (approved):

packages/contracts-bedrock/scripts/invariant-doc-gen.ts
clabby: Ah yes- so the `contract`'s `name` is found through the file name containing the tests, which assumes that for `forge` invariants, the file is named `ContractName.t.sol` in the `contracts/test/invariants` dir, and for `echidna`, named `FuzzContractName.sol` in the `contracts/echidna` dir.

For example, w/ `forge` invariants, each invariant test file will likely have many contracts within it due to differing `setUp`s (see: [Portal Invariants](https://github.com/ethereum-optimism/optimism/blob/97b1f605d3770ab603fd27e7f08c8d313dff95fc/packages/contracts-bedrock/contracts/test/invariants/OptimismPortal.t.sol)), so separating these into different files would probably add a good deal of clutter. LMK if I'm misunderstanding 😄

clabby:
> I'm fine with each contract getting its own file. Auto generating a table of contents would be a nice to have
> 
> I heard that dev work on `forge doc` is picking up again so wondering if we could migrate to using that for all of our docs longer term and if it picks up on custom tags

Agreed- would love to migrate over to this once it's out. If it doesn't already, upstreaming support for custom natspec tags shouldn't be a big lift.

packages/contracts-bedrock/scripts/invariant-doc-gen.ts
```diff
@@ -0,0 +1,175 @@
...
+const docGen = (dir: string): void => {
+  // Grab all files within the invariants test dir
+  const files = fs.readdirSync(dir)
+
+  // Array to store all found invariant documentation comments.
+  const docs: Contract[] = []
+
+  for (const fileName of files) {
+    // Read the contents of the invariant test file.
+    const fileContents = fs.readFileSync(path.join(dir, fileName)).toString()
```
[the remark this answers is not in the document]
tynes: I think it's ok to ignore this semgrep warning

tynes (commented):
This looks great at quick glance, will do a more thorough review

zhwrd (approved):

tynes (approved):
Looks good to me!

PR #4575, opened 2023-01-03
op-chain-ops: make the migrated withdrawals version 0

tynes:
**Description**

Update the migrated withdrawals to be version 0 based on the new scheme that is introduced in https://github.com/ethereum-optimism/optimism/pull/4562. This is a safety measure to ensure that withdrawals cannot be double withdrawn and that ether stuck in the L1xdm cannot be stolen in the case of a withdrawal with value reverting.

### DIFF
--- a/op-chain-ops/crossdomain/migrate.go
+++ b/op-chain-ops/crossdomain/migrate.go
@@ -63,7 +63,9 @@ func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger *com
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


tynes:
Only merge this with #4562

smartcontracts (commented):
Can also be explicit and use versionedNonce on line 87

smartcontracts (approved):

mslipper (approved):

tynes:
> Can also be explicit and use versionedNonce on line 87

I've updated this so that the same versioning is used for both fields

PR #4577, opened 2023-01-04
bedrock: hardcode L2 genesis coinbase

tynes:
**Description**

Removes the config option for being able to set
the coinbase of the bedrock transition block.
It is now always set to the sequencer fee vault
predeploy. This is one less config option that needs to be set correctly. The coinbase is always set to the sequencer fee vault in bedrock blocks.


mslipper (approved):

trianglesphere (approved):

tynes:
Not sure why the linting job failed, CI is saying it timed out

maurelian (approved):

PR #4582, opened 2023-01-04
fix(ctb): correct GPO abi and layout

smartcontracts:
**Description**
Fixes some small errors in the GPO's API and storage layout that did not conform to the standards for how we are ugprading the L2 system.


packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
```diff
@@ -129,6 +136,15 @@ contract GasPriceOracle is Semver {
         return L1Block(Predeploys.L1_BLOCK_ATTRIBUTES).basefee();
     }
 
+    /**
+     * @notice Retrieves the number of decimals used in the scalar.
```
tynes, line 140: This should have `@custom:legacy`

tynes (approved):

maurelian (approved):

packages/contracts-bedrock/contracts/L2/GasPriceOracle.sol
smartcontracts: The others don't have this but happy to add in another PR

PR #4598, opened 2023-01-05
feat: make ProxyAdmin self-owned Proxy

smartcontracts:
**Description**
Turns the ProxyAdmin into a Proxy that owns itself. Result of this is that we can treat the ProxyAdmin as a Proxy which reduces some complexity in the migration. We can then also upgrade the ProxyAdmin on L2.


tynes:
The `check-l2` script needs to be updated to check for the admin + impl slots being set at the `ProxyAdmin` predeploy account

mslipper:
Confirming - the owner of the `ProxyAdmin` would be the multisig. The admin of the proxy in front of `ProxyAdmin` would be itself. This would let us use `ProxyAdmin` to self-upgrade by calling `ProxyAdmin` from the multisig. Is this correct?

smartcontracts:
> Confirming - the owner of the `ProxyAdmin` would be the multisig. The admin of the proxy in front of `ProxyAdmin` would be itself. This would let us use `ProxyAdmin` to self-upgrade by calling `ProxyAdmin` from the multisig. Is this correct?

That is correct.

mslipper:
This looks good to me. Would like a review from @tynes or @maurelian before merging.

mslipper:
Closing in favor of an alternate proposal.

packages/contracts-bedrock/tasks/check-l2.ts
```diff
@@ -577,6 +577,9 @@ const check = {
 
     const addressManager = await ProxyAdmin.addressManager()
     console.log(`  - addressManager: ${addressManager}`)
+
+    await checkProxy(hre, 'ProxyAdmin')
+    await assertProxy(hre, 'ProxyAdmin')
```
maurelian, line 582:
```suggestion
    await checkProxy(hre, 'ProxyAdmin', signer.provider)
    await assertProxy(hre, 'ProxyAdmin', signer.provider)
```

tynes: ^^ This is correct, will break when using a URL passed via CLI if not

op-chain-ops/genesis/setters.go
```diff
@@ -178,6 +172,7 @@ func SetImplementations(db vm.StateDB, storage state.StorageConfig, immutable im
 			return fmt.Errorf("code not set for %s", name)
 		}
 	}
+
```
tynes, line 175: nit: unnecessary diff

op-chain-ops/genesis/layer_two_test.go
```diff
@@ -92,5 +92,5 @@ func TestBuildL2DeveloperGenesisDevAccountsFunding(t *testing.T) {
 
 	gen, err := genesis.BuildL2DeveloperGenesis(config, block)
 	require.NoError(t, err)
-	require.Equal(t, 2320, len(gen.Alloc))
+	require.Equal(t, 2321, len(gen.Alloc))
```
tynes, line 95: There is now 1 more additional code address in the state

packages/contracts-bedrock/tasks/check-l2.ts
tynes: CI failing with this error:

```
tasks/check-l2.ts(581,11): error TS2554: Expected 3 arguments, but got 2.
```

maurelian:
Specs need a small change too: 

https://github.com/ethereum-optimism/optimism/blob/9a2ffd119e18a49e0264edcd29c6c6a6206755e1/specs/predeploys.md#L256-L257

maurelian (approved):

mslipper:
Running against a datadir now. Will approve once I see that work.

op-chain-ops/genesis/check.go
```diff
@@ -76,6 +76,10 @@ var (
 		predeploys.ProxyAdminAddr: {
 			// Slot 0x00 (0) is _owner. Requires custom check, so set to a garbage value
 			ProxyAdminOwnerSlot: common.HexToHash("0xbadbadbadbad0xbadbadbadbadbadbadbadbad0xbadbadbadbad0xbadbadbad0"),
```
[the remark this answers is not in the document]
mslipper: Could've sworn I'd fixed this as part of another PR - checking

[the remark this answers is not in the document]
mslipper: Ah, they weren't fully fixed there either. Fixing fully now.

mslipper:
Migration against the datadir worked. Merging, and will perform an end-to-end migration tonight.

mslipper (approved):

PR #4600, opened 2023-01-05
op-chain-ops: wipe predeploy storage

tynes:
**Description**

Makes the spec for the migration process much more simple by wiping the storage of the predeploys expect for special cases. The contracts that do not have their storage wiped are:

- GovernanceToken
- WETH9
- LegacyMessagePasser
- LegacyERC20ETH
- DeployerWhitelist

### DIFF
--- a/op-chain-ops/genesis/check.go
+++ b/op-chain-ops/genesis/check.go
@@ -26,16 +26,36 @@ import (
 // in the future.
 const MaxSlotChecks = 1000
 
-var LegacyETHCheckSlots = map[common.Hash]common.Hash{
-	// Bridge
-	common.Hash{31: 0x06}: common.HexToHash("0x0000000000000000000000004200000000000000000000000000000000000010"),
-	// Symbol
-	common.Hash{31: 0x04}: common.HexToHash("0x4554480000000000000000000000000000000000000000000000000000000006"),
-	// Name
-	common.Hash{31: 0x03}: common.HexToHash("0x457468657200000000000000000000000000000000000000000000000000000a"),
-	// Total supply
-	common.Hash{31: 0x02}: {},
-}
+var (
+	LegacyETHCheckSlots = map[common.Hash]common.Hash{
+		// Bridge
+		common.Hash{31: 0x06}: common.HexToHash("0x0000000000000000000000004200000000000000000000000000000000000010"),
+		// Symbol
+		common.Hash{31: 0x04}: common.HexToHash("0x4554480000000000000000000000000000000000000000000000000000000006"),
+		// Name
+		common.Hash{31: 0x03}: common.HexToHash("0x457468657200000000000000000000000000000000000000000000000000000a"),
+		// Total supply
+		common.Hash{31: 0x02}: {},
+	}
+
+	// ContractStorageCount is a map of predeploy addresses to the number of storage slots expected
+	// to be set in those predeploys after the migration. It does not include any predeploys that
+	// were not wiped. It also accounts for the 2 EIP-1967 storage slots in each contract.
+	ContractStorageCount = map[common.Address]int{
+		predeploys.L2CrossDomainMessengerAddr:        3,
+		predeploys.L2StandardBridgeAddr:              2,
+		predeploys.SequencerFeeVaultAddr:             2,
+		predeploys.OptimismMintableERC20FactoryAddr:  2,
+		predeploys.L1BlockNumberAddr:                 2,
+		predeploys.GasPriceOracleAddr:                2,
+		predeploys.L1BlockAddr:                       2,
+		predeploys.L2ERC721BridgeAddr:                2,
+		predeploys.OptimismMintableERC721FactoryAddr: 2,
+		predeploys.ProxyAdminAddr:                    3,
+		predeploys.BaseFeeVaultAddr:                  2,
+		predeploys.L1FeeVaultAddr:                    2,
+	}
+)
 
 // PostCheckMigratedDB will check that the migration was performed correctly
 func PostCheckMigratedDB(ldb ethdb.Database, migrationData migration.MigrationData, l1XDM *common.Address, l1ChainID uint64) error {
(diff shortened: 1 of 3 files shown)


mslipper (changes requested):
I'm generally in favor of erasing storage to simplify the migration. However, I have some concerns with the implementation here that I'd like to address before merging.

- Have we audited all of the predeploys to ensure that we're not accidentally erasing necessary state, or if we are that the state gets reset in subsequent migration operations? The concern I have is that we un-initialize something that used to be initialized, leading to something like the Nomad hack. Would love to see some bullets talking about which state gets erased for each predeploy, and whether that state needs to be reset later on in the migration.
- Have you been able to run this on a datadir, and confirmed that all the checks pass? I can set this up for you if it'd be helpful. Looking at the code here, I'm 80% sure that the storage for OVM_ETH will be zeroed out as well.

tynes:
Converting this to a draft until bullet points in https://github.com/ethereum-optimism/optimism/pull/4600#pullrequestreview-1236807650 are addressed

Perhaps another approach at this would be to zero out specific storage slots instead of zeroing them all out. It adds some complexity to the spec but there is less risk of an unknown storage slot being reset. @smartcontracts requested this while writing the migration spec

I did push a commit that skips resetting the storage of OVM_ETH. I haven't ran this on a datadir yet

mslipper:
Cool, let's jam tomorrow. I have a script you can use that will give you a clean datadir to run on. Should make the feedback loop for running the migration much faster.

smartcontracts:
> Have we audited all of the predeploys to ensure that we're not accidentally erasing necessary state, or if we are that the state gets reset in subsequent migration operations?

I've manually reviewed the predeploys and can confirm that only the contracts listed in `contracts/src/predeploys.ts` contain any code or state. We have also carefully analyzed every predeploy and can confirm that wiping everything EXCEPT LegacyMessagePasser, DeployerWhitelist, GovernanceToken, and LegacyERC20ETH is OK.

tynes:
> > Have we audited all of the predeploys to ensure that we're not accidentally erasing necessary state, or if we are that the state gets reset in subsequent migration operations?
> 
> I've manually reviewed the predeploys and can confirm that only the contracts listed in `contracts/src/predeploys.ts` contain any code or state. We have also carefully analyzed every predeploy and can confirm that wiping everything EXCEPT LegacyMessagePasser, DeployerWhitelist, GovernanceToken, and LegacyERC20ETH is OK.

I believe we also need WETH9 in that list as well?

op-chain-ops/genesis/setters.go
```diff
@@ -60,6 +78,37 @@ func SetL1Proxies(db vm.StateDB, proxyAdminAddr common.Address) error {
 	return setProxies(db, proxyAdminAddr, bigL1PredeployNamespace, 2048)
 }
 
+// WipeStorage will set every storage variable that is set to `bytes32(0)`.
+func WipeStorage(db vm.StateDB, addr common.Address) error {
+	err := db.ForEachStorage(addr, func(key, _ common.Hash) bool {
```
smartcontracts, line 83: In the interest of speed, let's do `db.CreateAccount` instead and call the function WipePredeploy

tynes: Updated to this suggestion

tynes:
I also added in the post check code so that we can have coverage that the storage has been wiped. There is a todo for building a map of the number of expected storage slots per contract and then asserting that we have exactly the correct number of storage slots

tynes:
I've changed the base onto https://github.com/ethereum-optimism/optimism/pull/4605, going to need to fix a merge conflict

op-chain-ops/genesis/check.go
```diff
@@ -208,6 +230,46 @@ func PostCheckPredeploys(db *state.StateDB) error {
...
+		}
+
+		log.Info("predeploy storage", "name", name, "address", *addr, "count", len(slots))
+		for key, value := range slots {
+			log.Debug("storage values", "key", key, "value", value)
+		}
+
+		// TODO: create a map with expected number of storage slots by contract
+		// and assert that the expected number of storage slots matches the
+		// observed amount
```
protolambda, line 265: Old todo?

op-chain-ops/genesis/check.go
```diff
@@ -208,6 +230,44 @@ func PostCheckPredeploys(db *state.StateDB) error {
...
+			return err
+		}
+
+		log.Info("predeploy storage", "name", name, "address", *addr, "count", len(slots))
+		for key, value := range slots {
+			log.Debug("storage values", "key", key, "value", value)
+		}
+
+		// Assert that the correct number of slots are present.
+		if ContractStorageCount[*addr] != len(slots) {
```
mslipper, line 264: Can we also check that the storage root is the empty storage root for those contracts that have zero storage slots?

mslipper (approved):
Approving pending answer to my comment above.

op-chain-ops/genesis/check.go
smartcontracts: I think this is guaranteed implicitly

op-chain-ops/genesis/check.go
```diff
@@ -26,16 +26,33 @@ import (
...
+		// Bridge
+		common.Hash{31: 0x06}: common.HexToHash("0x0000000000000000000000004200000000000000000000000000000000000010"),
+		// Symbol
+		common.Hash{31: 0x04}: common.HexToHash("0x4554480000000000000000000000000000000000000000000000000000000006"),
+		// Name
+		common.Hash{31: 0x03}: common.HexToHash("0x457468657200000000000000000000000000000000000000000000000000000a"),
+		// Total supply
+		common.Hash{31: 0x02}: {},
+	}
+
```
tynes, line 40: Each of these should should `+ 2` for admin + impl eip 1967 storage slots being set, if they are behind a proxy

op-chain-ops/genesis/check.go
```diff
@@ -26,16 +26,33 @@ import (
...
+	ContractStorageCount = map[common.Address]int{
```
tynes, line 41: slight nit to have a comment here describing this and saying it accounts for eip 1967 but lgtm otherwise

tynes:
`foundryup` failed causing devnet + genesis contracts to fail, i think the api is timing out

tynes:
I can't approve but looks good to me

PR #4601, opened 2023-01-05
specs: guaranteed gas market cleanup

tynes:
**Description**

Adds in an additional section explaining that the gas burn helps with sybil resistance, to prevent network spam on L2. It was not clear that this was the case when reading the doc. Also clean up the pseudocode.


specs/guaranteed-gas-market.md
```diff
@@ -56,9 +56,9 @@ at a multiple of the target, we enable deposits to temporarily use more L2 gas a
 # the ability of the account to afford this gas.
 BASE_FEE_MAX_CHANGE_DENOMINATOR = 8
 ELASTICITY_MULTIPLIER = 4
-MAX_RESOURCE_LIMIT = 8,000,000
+MAX_RESOURCE_LIMIT = 8_000_000
```
maurelian, line 59: There is another occurrence of the `,` delimiter in this document on line 45, and quite a few in other locations, so I think we should stick with that for now or change them all.

image

maurelian (approved):
LGTM once my comment is resolved.

specs/guaranteed-gas-market.md
tynes: Good catch, I think having a comma in pseudocode looks weird, I'd prefer to change it everywhere but open to leaving it if people don't want the diff

trianglesphere: We can check those numbers, but I personally prefer commas in prose.

maurelian: I see now, @tynes is right, ie. this is inside a code block which is not obvious from the diff. 
The other instances in the screenshot are prose.

Resolving this thread.

PR #4603, opened 2023-01-05
CI: Debug echidna

maurelian:
**Description**

A draft PR for debugging echidna in CI.


tynes:
@maurelian What is the status on this?

maurelian:
@tynes I think this should be fixed by the updated ci-builder. Waiting on that.

tynes:
> @tynes I think this should be fixed by the updated ci-builder. Waiting on that.

It should be released now

PR #4605, opened 2023-01-05
feat(ops): clean up migration process

smartcontracts:
**Description**
Cleans up the migration process so that it matches with the state surgery specification. Adds a significant number of very pedantic comments.


op-chain-ops/genesis/db_migration.go
```diff
@@ -52,98 +58,134 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
...
+	// First order of business is to convert all predeployed smart contracts into proxies so they
+	// can be easily upgraded later on. In the legacy system, all upgrades to predeployed contracts
+	// required hard forks which was a huge pain. Note that we do NOT put the GovernanceToken or
+	// WETH9 contracts behind proxies because we do not want to make these easily upgradable.
+	log.Info("Converting predeployed contracts to proxies")
+	if err := SetL2Proxies(db); err != nil {
+		return nil, fmt.Errorf("cannot set L2Proxies: %w", err)
 	}
 
+	// Next we wipe the storage of all predeployed contracts EXCEPT for the GovernanceToken, WETH9,
```
mslipper, line 143: I don't think `SetImplementations` will wipe the storage explicitly. There's a check to see if the account exists, and if so it leaves the existing state alone:

```
if !db.Exist(codeAddr) {
    db.CreateAccount(codeAddr)
}
```

mslipper: This call to `db.CreateAccount` was the call the caused the earlier issue with state being erased.

op-chain-ops/genesis/db_migration.go
```diff
@@ -155,16 +197,19 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 		Difficulty:  common.Big0,
 		Number:      new(big.Int).Add(header.Number, common.Big1),
 		GasLimit:    (uint64)(config.L2GenesisBlockGasLimit),
-		GasUsed:     gasUsed,
+		GasUsed:     (uint64)(config.L2GenesisBlockGasLimit) * config.EIP1559Elasticity,
 		Time:        uint64(config.L2OutputOracleStartingTimestamp),
```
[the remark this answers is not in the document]
smartcontracts: Ah mistake, thought I could get away with removing but I misread some code. Adding back in now...

op-chain-ops/genesis/db_migration.go
smartcontracts: Ah yeah sorry this was written under the assumption that the other PR that mark was working on would make it in

op-chain-ops/genesis/db_migration.go
```diff
@@ -155,16 +204,19 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 		Difficulty:  common.Big0,
 		Number:      new(big.Int).Add(header.Number, common.Big1),
 		GasLimit:    (uint64)(config.L2GenesisBlockGasLimit),
-		GasUsed:     gasUsed,
+		GasUsed:     (uint64)(config.L2GenesisBlockGasLimit) * config.EIP1559Elasticity,
```
protolambda, line 207: Possible bug: The elasticity should only be used to compute the maximum limit based on the target. This multiplication doesn't make sense. And it's higher than the above `GasLimit`, which it should never be.

protolambda (commented):
just noticed something wrong, didn't catch it in first review, sorry.

op-chain-ops/genesis/db_migration.go
smartcontracts: Interesting. Maybe it's meant to be a division rather than a multiplication?

protolambda: If the transition block doesn't include any transactions, I think we should just set `GasUsed` to 0.

op-chain-ops/ether/precheck.go
```diff
@@ -0,0 +1,156 @@
...
+		switch slotType {
+		case 1:
+			// Balance slot.
+			totalFound.Add(totalFound, slotsAct[slot].Big())
+		case 2:
+			// Allowance slot.
+			continue
+		default:
+			// Should never happen.
+			return nil, fmt.Errorf("unknown slot type: %d", slotType)
```
smartcontracts, line 114: Put this behind a nocheck flag

op-chain-ops/ether/migrate.go
```diff
@@ -33,221 +29,75 @@ var (
...
+				"supply mismatch",
+				"migrated", totalMigrated.String(),
+				"supply", totalSupply.String(),
+				"delta", delta.String(),
+				"exp_delta", params.ExpectedSupplyDelta.String(),
 			)
 		}
 	}
 
-	// Set the total supply to 0
```
smartcontracts, line 248: Undo this with comment

op-chain-ops/genesis/db_migration.go
tynes: Ah so the gas limit here is the limit and not the target. I was going for trying to have the eip1559 basefee computation be smooth. If we know that its going to mine a bunch of empty blocks at start then i'm fine with it just being set to 0 and save the complexity

op-chain-ops/crossdomain/precheck.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Build a mapping of the slots of all messages actually sent in the legacy system.
+	var count int
+	slotsAct := make(map[common.Hash]bool)
+	err := db.ForEachStorage(predeploys.LegacyMessagePasserAddr, func(key, value common.Hash) bool {
+		// When a message is inserted into the LegacyMessagePasser, it is stored with the value
+		// of the ABI encoding of "true". Although there should not be any other storage slots, we
+		// can safely ignore anything that is not "true".
+		if value != abiTrue {
+			return false
```
mslipper, line 35: Returning `false` from `ForEachStorage` stops iteration, which causes a bunch of these messages to be missed. I'll push a fix once I've successfully migrated a datadir with this code.

op-chain-ops/ether/precheck.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	// Build a mapping of every storage slot in the LegacyERC20ETH contract, except the list of
+	// slots that we know we can ignore (totalSupply, name, symbol).
+	var count int
+	slotsAct := make(map[common.Hash]common.Hash)
+	err = db.ForEachStorage(predeploys.LegacyERC20ETHAddr, func(key, value common.Hash) bool {
+		// We can safely ignore specific slots (totalSupply, name, symbol).
+		if ignoredSlots[key] {
+			return false
```
mslipper, line 75: Ditto here - this will stop iteration early.

op-chain-ops/genesis/check.go
```diff
@@ -26,16 +26,36 @@ import (
...
+		predeploys.L2CrossDomainMessengerAddr:        3,
+		predeploys.L2StandardBridgeAddr:              2,
+		predeploys.SequencerFeeVaultAddr:             2,
+		predeploys.OptimismMintableERC20FactoryAddr:  2,
+		predeploys.L1BlockNumberAddr:                 2,
+		predeploys.GasPriceOracleAddr:                2,
+		predeploys.L1BlockAddr:                       2,
+		predeploys.L2ERC721BridgeAddr:                2,
+		predeploys.OptimismMintableERC721FactoryAddr: 2,
+		predeploys.ProxyAdminAddr:                    3,
```
mslipper, line 54: The ProxyAdmin is not behind a proxy, so it only has the `owner` storage slot set.

op-chain-ops/ether/precheck.go
```diff
@@ -0,0 +1,160 @@
...
+	// For each known address, compute its balance key and add it to the list of addresses.
+	for _, addr := range addresses {
+		addrs = append(addrs, addr)
+		slotsInp[CalcOVMETHStorageKey(addr)] = 1
+	}
+
+	// For each known allowance, compute its storage key and add it to the list of addresses.
+	for _, allowance := range allowances {
+		addrs = append(addrs, allowance.From)
+		slotsInp[CalcAllowanceStorageKey(allowance.From, allowance.To)] = 2
```
tynes, line 42: We are 100% sure its `from, to` and not `to, from`?

op-chain-ops/genesis/check.go
```diff
@@ -26,19 +29,68 @@ import (
...
+
+var (
+	L2XDMOwnerSlot      = common.Hash{31: 0x33}
+	ProxyAdminOwnerSlot = common.Hash{}
+
+	LegacyETHCheckSlots = map[common.Hash]common.Hash{
+		// Bridge
+		common.Hash{31: 0x06}: common.HexToHash("0x0000000000000000000000004200000000000000000000000000000000000010"),
+		// Symbol
+		common.Hash{31: 0x04}: common.HexToHash("0x4554480000000000000000000000000000000000000000000000000000000006"),
```
tynes, line 42:
```
$ cast --to-ascii 0x455448
ETH
```

op-chain-ops/genesis/check.go
```diff
@@ -26,19 +29,68 @@ import (
...
+		// Name
+		common.Hash{31: 0x03}: common.HexToHash("0x457468657200000000000000000000000000000000000000000000000000000a"),
```
tynes, line 44:
```
$ cast --to-ascii 0x4574686572
Ether
```

op-chain-ops/genesis/check.go
```diff
@@ -26,19 +29,68 @@ import (
...
+	// ExpectedStorageSlots is a map of predeploy addresses to the storage slots and values that are
+	// expected to be set in those predeploys after the migration. It does not include any predeploys
+	// that were not wiped. It also accounts for the 2 EIP-1967 storage slots in each contract.
+	// It does _not_ include L1Block. L1Block is checked separately.
+	ExpectedStorageSlots = map[common.Address]StorageCheckMap{
+		predeploys.L2CrossDomainMessengerAddr: {
+			// Slot 0x00 (0) is a combination of spacer_0_0_20, _initialized, and _initializing
+			common.Hash{}: common.HexToHash("0x0000000000000000000000010000000000000000000000000000000000000000"),
+			// Slot 0x33 (51) is _owner. Requires custom check, so set to a garbage value
+			L2XDMOwnerSlot: common.HexToHash("0xbadbadbadbad0xbadbadbadbad0xbadbadbadbad0xbadbadbadbad0xbadbad00"),
```
tynes, line 58: Why does this contain additional `0x`s inside of it?

op-chain-ops/genesis/check.go
```diff
@@ -26,19 +29,68 @@ import (
...
+		predeploys.OptimismMintableERC20FactoryAddr: eip1967Slots(predeploys.OptimismMintableERC20FactoryAddr),
+		predeploys.L1BlockNumberAddr:                eip1967Slots(predeploys.L1BlockNumberAddr),
+		predeploys.GasPriceOracleAddr:               eip1967Slots(predeploys.GasPriceOracleAddr),
+		//predeploys.L1BlockAddr:                       eip1967Slots(predeploys.L1BlockAddr),
+		predeploys.L2ERC721BridgeAddr:                eip1967Slots(predeploys.L2ERC721BridgeAddr),
+		predeploys.OptimismMintableERC721FactoryAddr: eip1967Slots(predeploys.OptimismMintableERC721FactoryAddr),
+		// ProxyAdmin is not a proxy, and only has the _owner slot set.
+		predeploys.ProxyAdminAddr: {
+			// Slot 0x00 (0) is _owner. Requires custom check, so set to a garbage value
+			ProxyAdminOwnerSlot: common.HexToHash("0xbadbadbadbad0xbadbadbadbad0xbadbadbadbad0xbadbadbadbad0xbadbad00"),
```
tynes, line 78: Same as above, why does this contain additional `0x`s inside of it?

op-chain-ops/crossdomain/precheck.go
tynes: I thought that returning `false` would cause it to return an `error` - good catch here

op-chain-ops/cmd/op-migrate/main.go
```diff
@@ -179,7 +183,7 @@ func main() {
 
 			dryRun := ctx.Bool("dry-run")
 			noCheck := ctx.Bool("no-check")
-			// Perform the migration
+			//Perform the migration
```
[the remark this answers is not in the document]
smartcontracts: Seems like a mistake

op-chain-ops/crossdomain/precheck.go
```diff
@@ -0,0 +1,75 @@
...
+		slotsInp[slot] = wd
+	}
+
+	// Build a mapping of the slots of all messages actually sent in the legacy system.
+	var count int
+	slotsAct := make(map[common.Hash]bool)
+	err := db.ForEachStorage(predeploys.LegacyMessagePasserAddr, func(key, value common.Hash) bool {
+		// When a message is inserted into the LegacyMessagePasser, it is stored with the value
+		// of the ABI encoding of "true". Although there should not be any other storage slots, we
+		// can safely ignore anything that is not "true".
```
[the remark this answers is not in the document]
smartcontracts: Done

op-chain-ops/genesis/check.go
```diff
@@ -26,19 +29,68 @@ import (
...
-	common.Hash{31: 0x04}: common.HexToHash("0x4554480000000000000000000000000000000000000000000000000000000006"),
-	// Name
-	common.Hash{31: 0x03}: common.HexToHash("0x457468657200000000000000000000000000000000000000000000000000000a"),
-	// Total supply
-	common.Hash{31: 0x02}: {},
-}
+type StorageCheckMap = map[common.Hash]common.Hash
+
+var (
+	L2XDMOwnerSlot      = common.Hash{31: 0x33}
```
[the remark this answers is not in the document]
mslipper: It is checked - the value gets passed into the map below and the key and the value both get checked.

op-chain-ops/genesis/check.go
mslipper: It's a typo. It's in the map as a garbage value to fail the test if we forget to do the manual check. I'll fix the typo, however the actual test functionality is unaffected.

op-chain-ops/genesis/check.go
mslipper: See above

op-chain-ops/ether/precheck.go
```diff
@@ -0,0 +1,160 @@
...
+			if noCheck {
+				log.Error("ignoring unknown storage slot in state", "slot", slot)
+			} else {
+				log.Crit("unknown storage slot in state: %s", slot)
+			}
+		}
+
+		// Add balances to the total found.
+		switch slotType {
+		case 1:
```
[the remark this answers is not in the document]
mslipper: Would prefer to leave as-is - don't want to refactor too much now.

PR #4611, opened 2023-01-06
fix(data transport layer): Fix typescript type resolution

roninjin10:
Typescript type resolution here is really wierd.  TIL it doesn't match nodes resolution perfectly.   In a git submodule it would resolve to leveldb in parent before it would find a resolution in option.   This blocked me from adding optimism monorepo as submodule in gateway.

Fix: add @types package to data transport layer so it resolves right away in that package.

### DIFF
--- a/packages/data-transport-layer/package.json
+++ b/packages/data-transport-layer/package.json
@@ -58,6 +58,7 @@
   "devDependencies": {
     "@types/cors": "^2.8.9",
     "@types/levelup": "^4.3.0",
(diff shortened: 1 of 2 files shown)


tynes (approved):
This looks good to me, I feel like we had a lot of troubles in the past when it came to types and versions of the node leveldb bindings. Everything is building correctly now. iirc we did some hacky versioning stuff in the past to solve this issue?

PR #4613, opened 2023-01-06
op-chain-ops: export bedrock transition block extradata

tynes:
**Description**

Makes the extradata for the bedrock transition block a public variable. This is useful for exporting the value to be imported in other packages where assertions can be made to ensure that the extradata is correct.

When fetching the `L2OutputOracle`'s configured starting block number, the L2 block with that number can be fetched and then we can assert that it has the magic extradata.

```bash
$ cast call 0x02Ac1923480E7fd234e7F2Fd11596b9F20372374 'startingBlockNumber()(uint256)'
3855343
```

```bash
$ cast block 3855343 --json | jq -r .extraData
0x424544524f434b
```

Now check like so:

```bash
$ cast --to-ascii 0x424544524f434b
BEDROCK
```

### DIFF
--- a/op-chain-ops/genesis/db_migration.go
+++ b/op-chain-ops/genesis/db_migration.go
@@ -21,7 +21,7 @@ import (
 
 var (
 	abiTrue                         = common.Hash{31: 0x01}
-	bedrockTransitionBlockExtraData = []byte("BEDROCK")
+	BedrockTransitionBlockExtraData = []byte("BEDROCK")
 )
 
 type MigrationResult struct {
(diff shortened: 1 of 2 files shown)


op-chain-ops/genesis/db_migration.go
```diff
@@ -21,7 +21,7 @@ import (
 
 var (
 	abiTrue                         = common.Hash{31: 0x01}
-	bedrockTransitionBlockExtraData = []byte("BEDROCK")
+	BedrockTransitionBlockExtraData = []byte("BEDROCK")
```
protolambda, line 24: If we're making it public and mutable maybe it's worth a comment that this must not be larger than 32 bytes (because of post-Merge consensus-layer block encoding)

protolambda (commented):
LGTM. One nit about doc-commenting public vars

tynes:
I pushed a commit that comments on the variable as well as returns an error at runtime if the variable is too large

protolambda (approved):

PR #4615, opened 2023-01-06
Example of spec mismatch

trianglesphere:
This shows the conflict in the specs between the logic laid out in the [batch validity rules](https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#batch-buffering) is in conflict with [rules just below the overview](https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#overview).

This includes example fixes for both the batch submission case and the no batch submission case. If the fix is not applied, the test times out (can observe a livelock in the logs).


trianglesphere:
Closing as we are not going to be including this feature.

PR #4616, opened 2023-01-06
feat: wait for final deposit hh task

tynes:
**Description**

Adds a hh task that can be used as part of the automation
to wait for the final deposit to be ingested by `l2geth`.
It first waits until the dtl shutoff block is set in the
address manager and then it works backwards from the tip
of the sequencer to find the final deposit

To run the script:

```bash
$ npx hardhat wait-for-final-deposit \
    --l2-rpc-url https://mainnet.optimism.io \
    --network mainnet \
    --l1-rpc-url ...
```

Closes ENG-3169


tynes:
One thing that i do worry about here is a consistency issue, going to add something that polls the tip in the background and errors if the the tip ever changes

PR #4597, opened 2023-01-04
fix(ctp): Downgrade `contracts-bedrock` dep to fix CI

clabby:
# Overview

In the latest [version action](https://github.com/ethereum-optimism/optimism/commit/a17b49fafa8b0141ae45c2b979f432f2468fd7ba), `contracts-bedrock` was updated to `v0.11.1`, which broke `contracts-periphery`'s CI. Temporarily downgrades `contracts-periphery`'s `contracts-bedrock` dependency to `v0.11.0` so that it will build.

PR #4617, opened 2023-01-06
chore(ctp): Update ctb dep version

maurelian:
The release action is failing due to a mismatch between the actual contracts-bedrock package version, and the version used in the periphery package. 

This change was made in #4597 to 'fix an error in CI', so hopefully this reversal of the change doesn't cause a new error, or we may find ourselves in a stalemate between CI and the release process. 😳


packages/contracts-bedrock/foundry.toml
```diff
@@ -7,7 +7,6 @@ optimizer_runs = 999999
 remappings = [
   '@openzeppelin/contracts-upgradeable/=node_modules/@openzeppelin/contracts-upgradeable/',
   '@openzeppelin/contracts/=node_modules/@openzeppelin/contracts/',
-  '@eth-optimism/contracts-periphery/=node_modules/@eth-optimism/contracts-periphery/contracts',
```
maurelian, line 10: This was wrong (should remap to `../../@eth-optimism`), but it doesn't appear to be in use so I've just removed it.

tynes: you cannot import remappings outside of the current dir, the base layer `package.json` hoists `contracts-periphery` into the local node modules here, if `contracts-periphery` is no longer a dep to `contracts-bedrock` then we can delete the hoisting in the base `package.json`

maurelian: In the end it seems that: 
1. remappings outside the current dir (as above) do work
2. despite [this change](https://github.com/ethereum-optimism/optimism/pull/4617/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R28) (which I thought would properly apply `nohoist` to the `contracts-bedrock` package), those contracts are still found in the root `node_modules` dir. I have checked that this is the case both locally and by ssh'ing into CI. 

It all seems a bit odd, but given the checks are now passing I'm going to resolve which will hopefully unblock the release process.

packages/contracts-periphery/foundry.toml
```diff
@@ -18,7 +18,7 @@ remappings = [
 	'ds-test/=node_modules/ds-test/src',
   '@openzeppelin/contracts/=node_modules/@openzeppelin/contracts/',
   '@openzeppelin/contracts-upgradeable/=node_modules/@openzeppelin/contracts-upgradeable/',
-  '@eth-optimism/contracts-bedrock/=node_modules/@eth-optimism/contracts-bedrock',
+  '@eth-optimism/contracts-bedrock/=../../node_modules/@eth-optimism/contracts-bedrock',
```
tynes, line 21: I am surprised that this builds? It used to not build like this

tynes:
Curious what is different about weth9, guessing its the metadata hash

tynes (approved):

maurelian:
> Curious what is different about weth9, guessing its the metadata hash

Good question. Do we include metadatahashes in the go bindings?

PR #4620, opened 2023-01-06
op-node: Enforce min L2 timestamp

trianglesphere:
**Description**

This adds a check to the batch queue to verify that supplied batches have a timestamp that is greater than or equal to the L1 origin timestamp. The sequencer should not sequence batches like this, but it was not being enforced in the derivation process.

**Tests**

I added a unit test for this & manually confirmed that the unit test will fail if the fix is not included.

**Metadata**

- Fixes ENG-3156


protolambda (approved):
Nice find

tynes (approved):

PR #4622, opened 2023-01-06
contracts-bedrock: check-l2 checks for genesis magic

tynes:
**Description**

Add additional check for the genesis magic value.
This ensures that the L2OutputOracle's starting
block number is set correctly.


mslipper (approved):

mslipper:
Looks like the devnet is having issues finding the deploy config.

tynes:
> Looks like the devnet is having issues finding the deploy config.

Will push a fix

PR #4627, opened 2023-01-08
2 small proposed changes to bedrock specs overview page

samlaf:
Suggesting these as a PR because I think it will be easier to look at the diffs, rather than submitting an issue or Q&A with a bunch of links. *** THESE ARE ONLY SUGGESTIONS, PLEASE REVIEW ***

Change 1: Who receives L1 transactions from users? This page mentioned it's the rollup node, but this diagram in the [introduction page](https://github.com/ethereum-optimism/optimism/blob/develop/specs/introduction.md#depositing-and-sending-transactions) indicates that it's the EE. I updates the docs here to reflect this fact, but it might be the diagram in the introduction that needs updating... not sure
image

Change 2: changed `engine_newPayloadV1` to `engine_executePayloadV1` to reflect the diagram at the very bottom. Again, not sure whether the written form or the image form is correct, but one of them is necessarily wrong, and needs updating.
image

### DIFF
--- a/specs/overview.md
+++ b/specs/overview.md
@@ -65,7 +65,6 @@ and fault proofs.
 
 - **Rollup Node**:
   - A standalone, stateless binary.
-  - Receives L2 transactions from users.
   - Syncs and verifies rollup data on L1.
   - Applies rollup-specific block production rules to synthesize blocks from L1.
   - Appends blocks to the L2 chain using the Engine API.
@@ -74,6 +73,7 @@ and fault proofs.
 - **Execution Engine (EE)**:
   - A vanilla Geth node with minor modifications to support Optimism.
   - Maintains L2 state.
+  - Receives L2 transactions from users.
   - Sync state to other L2 nodes for fast onboarding.
   - Serves the Engine API to the rollup node.
 - **Batch Submitter**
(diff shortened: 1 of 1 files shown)


protolambda:
> Change 1: Who receives L1 transactions from users? This page mentioned it's the rollup node, but this diagram in the [introduction page](https://github.com/ethereum-optimism/optimism/blob/develop/specs/introduction.md#depositing-and-sending-transactions) indicates that it's the EE. I updates the docs here to reflect this fact, but it might be the diagram in the introduction that needs updating... not sure

The Execution Engine provides the standard JSON RPC that allows users to insert transactions into the mempool. So the EE would be the right thing. We can update the introduction page. Diagram is correct.

> Change 2: changed engine_newPayloadV1 to engine_executePayloadV1 to reflect the diagram at the very bottom. Again, not sure whether the written form or the image form is correct, but one of them is necessarily wrong, and needs updating.

The diagram was created before the Merge, when some Engine API methods still had different names and slightly different semantics. We should update the diagram.

Nice finds, thank you for reviewing the specs. I'll update the diagrams on Monday and then merge this PR combined with the diagram fixes.

PR #4632, opened 2023-01-09
tests(ctb): Add invariant test for `minGasLimit`s in the L1 XDM's `relayMessage` fn

clabby:
# Overview

Adds an actor-based invariant test for the L1 XDM requested by @smartcontracts.

**Invariant**

A call from the `OptimismPortal` to the `L1CrossDomainMessenger`'s `relayMessage` function can never revert if the proper minimum gas limits are supplied.

There are two minimum gas limits we need to worry about:
* **[OUTER]** - Minimum gas required for the call from the `OptimismPortal` to the L1 XDM's `relayMessage` function. This is found by calling the L1 XDM's `baseGas` function with the message payload + the inner min gas limit.
* **[INNER]** - Minimum gas required for the call from the L1 XDM to the `target` contract.

**Tests**

Added `invariants/CrossDomainMessenger.t.sol`

**Metadata**
Closes ENG-3172

### DIFF
--- a/packages/contracts-bedrock/contracts/test/invariants/CrossDomainMessenger.t.sol
+++ b/packages/contracts-bedrock/contracts/test/invariants/CrossDomainMessenger.t.sol
@@ -116,6 +116,7 @@ contract XDM_MinGasLimits is Messenger_Initializer, InvariantTest {
      * gas limits are supplied.
      *
      * There are two minimum gas limits here:
+     *
      * - The outer min gas limit is for the call from the `OptimismPortal` to the
      * `L1CrossDomainMessenger`,  and it can be retrieved by calling the xdm's `baseGas` function
      * with the `message` and inner limit.
--- a/packages/contracts-bedrock/contracts/test/invariants/CrossDomainMessenger.t.sol
+++ b/packages/contracts-bedrock/contracts/test/invariants/CrossDomainMessenger.t.sol
@@ -125,6 +125,6 @@ contract XDM_MinGasLimits is Messenger_Initializer, InvariantTest {
             // it is not in the received messages mapping
             assertFalse(L1Messenger.failedMessages(hash));
         }
-        assertTrue(!actor.outOfGas());
+        assertFalse(actor.outOfGas());
     }
 }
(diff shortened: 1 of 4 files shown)


packages/contracts-bedrock/contracts/test/invariants/CrossDomainMessenger.t.sol
```diff
@@ -0,0 +1,137 @@
...
+     * There are two minimum gas limits here:
+     * - The outer min gas limit is for the call from the `OptimismPortal` to the
+     * `L1CrossDomainMessenger`,  and it can be retrieved by calling the xdm's `baseGas` function
+     * with the `message` and inner limit.
+     *
+     * - The inner min gas limit is for the call from the `L1CrossDomainMessenger` to the target
+     * contract.
+     */
+    function invariant_minGasLimits() public {
+        uint256 length = actor.numHashes();
```
tynes, line 127: What do you think about adding an assertion for `length > 0` here? I don't believe that this will happen and it may make it fragile to config changes (but it doesn't make sense to configure 0 runs?) but otherwise we rely on viewing the output of the test run to verify that this ran at least one time. It would silently do nothing without this assertion

clabby: Unfortunately, we can't (unless we ask the actor to relay in `setUp`, but that could fail as well.). Forge's invariant fuzzer will check the invariant before attempting any call sequences, so the test would not run (hence the weird loop + check for the `reverted` flag).

clabby: Gotcha- yeah, less than optimal. While it is possible, it *should* never perform 0 calls. The default `depth` is 256 and the only target contract is the actor (which has a single external function).

smartcontracts (approved):
This is fantastic, thank you!

clabby:
* develop
  * **PR #4632** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4632" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4632?utm_source=stack-comment).

PR #4633, opened 2023-01-09
fix: don't cancel context before passing to signer

aflock:
**Description**

Since #4555, we can pass a context to the signing function for e.g. signing with a different backend. Right now the batcher's context is being early-cancelled, so fix that. 

**Tests**

Using existing tests

**Invariants**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/op-batcher/batcher/txmgr.go
+++ b/op-batcher/batcher/txmgr.go
@@ -98,8 +98,8 @@ func (t *TransactionManager) CraftTx(ctx context.Context, data []byte) (*types.T
 		return nil, err
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, networkTimeout)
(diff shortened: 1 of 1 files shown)


op-batcher/batcher/txmgr.go
```diff
@@ -121,6 +121,8 @@ func (t *TransactionManager) CraftTx(ctx context.Context, data []byte) (*types.T
 	}
 	rawTx.Gas = gas
 
+	ctx, cancel = context.WithTimeout(ctx, networkTimeout)
```
aflock, line 124: I'm re-using `networkTimeout`, but that will be a little at-odds for non-networked signing setups. Open to suggestions if there's something better.

protolambda (approved):
Nice catch, thank you.

tynes (approved):

PR #4639, opened 2023-01-09
feat(ctb): commit prod Goerli config

smartcontracts:
**Description**
Commits the production Goerli configuration for Bedrock.

**Metadata**

- Fixes ENG-3143

### DIFF
--- a/packages/contracts-bedrock/deploy-config/goerli.json
+++ b/packages/contracts-bedrock/deploy-config/goerli.json
@@ -1,36 +1,45 @@
 {
-  "finalSystemOwner": "DUMMY",
-  "controller": "DUMMY",
+  "numDeployConfirmations": 1,
 
-  "l1StartingBlockTag": "DUMMY",
+  "finalSystemOwner": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "controller": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+
+  "l1StartingBlockTag": "FILL_IN_DAY_OF",
+
   "l1ChainID": 5,
   "l2ChainID": 420,
   "l2BlockTime": 2,
 
-  "maxSequencerDrift": 1200,
+  "maxSequencerDrift": 600,
   "sequencerWindowSize": 3600,
-  "channelTimeout": 120,
+  "channelTimeout": 300,
 
-  "p2pSequencerAddress": "DUMMY",
+  "p2pSequencerAddress": "0x715b7219D986641DF9eFd9C7Ef01218D528e19ec",
   "batchInboxAddress": "0xff00000000000000000000000000000000000420",
-  "batchSenderAddress": "DUMMY",
-  "l2OutputOracleSubmissionInterval": 20,
-  "l2OutputOracleStartingTimestamp": -1,
-  "l2OutputOracleProposer": "DUMMY",
-  "l2OutputOracleChallenger": "DUMMY",
-  "finalizationPeriodSeconds": 2,
+  "batchSenderAddress": "0x7431310e026B69BFC676C0013E12A1A11411EEc9",
 
-  "l2GenesisBlockGasLimit": "0x17D7840",
-  "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
+  "l2OutputOracleSubmissionInterval": 12,
+  "l2OutputOracleStartingBlockNumber": "FILL_IN_DAY_OF",
+  "l2OutputOracleStartingTimestamp": "FILL_IN_DAY_OF",
 
-  "l2CrossDomainMessengerOwner": "DUMMY",
+  "l2OutputOracleProposer": "0x02b1786A85Ec3f71fBbBa46507780dB7cF9014f6",
+  "l2OutputOracleChallenger": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
 
-  "governanceTokenName": "Optimism",
-  "governanceTokenSymbol": "OP",
-  "governanceTokenOwner": "0x038a8825A3C3B0c08d52Cc76E5E361953Cf6Dc76",
+  "finalizationPeriodSeconds": 12,
+
+  "proxyAdminOwner": "0xf80267194936da1E98dB10bcE06F3147D580a62e",
 
   "gasPriceOracleOverhead": 2100,
   "gasPriceOracleScalar": 1000000,
+
+  "governanceTokenSymbol": "OP",
+  "governanceTokenName": "Optimism",
+  "governanceTokenOwner": "0x038a8825A3C3B0c08d52Cc76E5E361953Cf6Dc76",
+
+  "l2GenesisBlockGasLimit": "0x17D7840",
+  "l2GenesisBlockCoinbase": "0x4200000000000000000000000000000000000011",
+  "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
+
   "eip1559Denominator": 50,
   "eip1559Elasticity": 10
 }
--- a/packages/contracts-bedrock/deploy-config/goerli.json
+++ b/packages/contracts-bedrock/deploy-config/goerli.json
@@ -28,6 +28,9 @@
   "finalizationPeriodSeconds": 12,
 
   "proxyAdminOwner": "0xf80267194936da1E98dB10bcE06F3147D580a62e",
+  "baseFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "l1FeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "sequencerFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
 
   "gasPriceOracleOverhead": 2100,
   "gasPriceOracleScalar": 1000000,
--- a/packages/contracts-bedrock/deploy-config/goerli.json
+++ b/packages/contracts-bedrock/deploy-config/goerli.json
@@ -40,7 +40,6 @@
   "governanceTokenOwner": "0x038a8825A3C3B0c08d52Cc76E5E361953Cf6Dc76",
 
   "l2GenesisBlockGasLimit": "0x17D7840",
-  "l2GenesisBlockCoinbase": "0x4200000000000000000000000000000000000011",
   "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
 
   "eip1559Denominator": 50,


mslipper (approved):

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,45 @@
 {
-  "finalSystemOwner": "DUMMY",
-  "controller": "DUMMY",
+  "numDeployConfirmations": 1,
+
+  "finalSystemOwner": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "controller": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
```
protolambda, line 5: These don't match the addresses in the notion table, did we change keys?

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,45 @@
...
-  "l2OutputOracleProposer": "DUMMY",
-  "l2OutputOracleChallenger": "DUMMY",
-  "finalizationPeriodSeconds": 2,
+  "l2OutputOracleStartingBlockNumber": "FILL_IN_DAY_OF",
+  "l2OutputOracleStartingTimestamp": "FILL_IN_DAY_OF",
 
-  "l2GenesisBlockGasLimit": "0x17D7840",
-  "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
+  "l2OutputOracleProposer": "0x02b1786A85Ec3f71fBbBa46507780dB7cF9014f6",
+  "l2OutputOracleChallenger": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
```
protolambda, line 26: This address does not match the one in the notion table either

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,45 @@
...
 
-  "l2CrossDomainMessengerOwner": "DUMMY",
+  "finalizationPeriodSeconds": 10,
```
protolambda, line 28: Last time I checked with @trianglesphere this value was going to be discussed with @tynes @mslipper @smartcontracts . 10 seconds seems extremely low if we want to test the two-step withdrawal monitoring things. Josh/me were thinking of e.g. setting it to 10 min. (And mainnet would ofc. be a week)

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,45 @@
...
-  "channelTimeout": 120,
+  "channelTimeout": 300,
 
-  "p2pSequencerAddress": "DUMMY",
+  "p2pSequencerAddress": "0x715b7219D986641DF9eFd9C7Ef01218D528e19ec",
   "batchInboxAddress": "0xff00000000000000000000000000000000000420",
-  "batchSenderAddress": "DUMMY",
+  "batchSenderAddress": "0x7431310e026B69BFC676C0013E12A1A11411EEc9",
+
   "l2OutputOracleSubmissionInterval": 20,
```
protolambda, line 21: Josh/me suggested 20 min, but it's in number of blocks, so that would be `20*60/12=100` Are we intentionally setting it to 20?

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: Yes - upon further discussion we set this to a multisig to match mainnet.

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: Ditto - set to the multisig.

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: I'm fine raising it. @smartcontracts suggested 60 seconds to allow people at hackathons to easily withdraw. @K-Ho can you chime in?

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: As of now yes - cc @K-Ho

K-Ho: So right now on Optimism Goerli, we submit state batches roughly once every 5 minutes https://goerli.etherscan.io/address/0x9c945aC97Baf48cB784AbBB61399beB71aF7A378. This roughly costs 10 Goerli ETH per month. 
If we choose 20 blocks here, we will replicate the devex of testing withdrawals on Optimism Goerli today, with a little bit more waiting due to two step withdrawals - every time a developer wishes to test sending a withdrawal, they will have to wait 0-5 minutes before proving their withdrawal and then 10 seconds before finalizing their withdrawal.

I'd recommend we parameterize this value to as low of a value as we think the protocol can handle - if we can safely set it to **2 blocks**, that would be a much better devex for testing withdrawals (the average time to try a withdrawal would go from ~3 minutes down to <1 minute) for just a little bit of extra Goerli ETH spend on our end. 

We've had developers complain in the past about it being slow to test if they are doing withdrawals correctly given the time required to test a single end-to-end withdrawal.

packages/contracts-bedrock/deploy-config/goerli.json
K-Ho: I'd strongly recommend 10 seconds - we should be optimizing here for ease of developers testing withdrawals. I assume there's plenty of ways we can test two step withdrawal monitoring even with a 10 sec finalization period, e.g. monitor every withdrawal that has been proven, but not finalized, regardless of whether or not the 10 second finalization period has elapsed.

We've had developers complain in the past about it being slow to test if they are doing withdrawals correctly given the current time required to test a single end-to-end withdrawal, and 10 minutes would be significantly worse than the current time to withdraw on Optimism Goerli (~3 min)

tynes: I agree with @K-Ho - we should be optimizing for developers, ie: it should be possible to do e2e forked network tests on goerli using foundry (once we have the optimism state transition function in forge)

Out 2 step withdrawal infra should be able to handle the 10s withdrawal window, its not like it needs to do anything particularly complex

packages/contracts-bedrock/deploy-config/goerli.json
trianglesphere: FYI I'm pretty sure that this parameter is the number of **L2 Blocks**. 20 L2 blocks -> submit every 40s which is a pretty fast UX. We can drop it down to every 24s or 30s (i.e. ~2 L1 blocks). This will cost us ~100 Goerli ETH per month which seems fine.

packages/contracts-bedrock/deploy-config/goerli.json
smartcontracts: 12s + 12 L2 blocks is my proposal

smartcontracts: = 36 seconds total

smartcontracts: = 3 L1 blocks

smartcontracts:
Confirmed with @K-Ho that the values I used are OK

mslipper:
we need to specify the following values:

```
  "baseFeeVaultRecipient": "0x62790eFcB3a5f3A5D398F95B47930A9Addd83807",
  "l1FeeVaultRecipient": "0x62790eFcB3a5f3A5D398F95B47930A9Addd83807",
  "sequencerFeeVaultRecipient": "0x62790eFcB3a5f3A5D398F95B47930A9Addd83807",
```

protolambda:
@mslipper those were previously configurable, but should now be constants; https://github.com/ethereum-optimism/op-geth/pull/23

mslipper:
Confirmed with @tynes that these fields are required. I created a Goerli multisig for them.

smartcontracts:
> Confirmed with @tynes that these fields are required. I created a Goerli multisig for them.

Can we just use the same multisig that owns the rest of the system?

mslipper:
> Can we just use the same multisig that owns the rest of the system?

It needs to be on L1 - does that multisig exist on both networks?

mslipper:
Ah yeah, the `finalSystemOwner` is an L1 proxy address. We can use that, sure.

protolambda:
> If that's the case then we'll need to remove the validation for these inside op-chain-ops.

@tynes you implemented the fee vault address hardcoding and some of the op-chain-ops checks, can you take a look please?

mslipper:
@protolambda validation is meant to be there, these values can be any L1 address.

mslipper:
Updated to use the `finalSystemOwner`.

trianglesphere (approved):

protolambda (approved):

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,48 @@
 {
-  "finalSystemOwner": "DUMMY",
-  "controller": "DUMMY",
+  "numDeployConfirmations": 1,
+
+  "finalSystemOwner": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
```
tynes, line 4: The `finalSystemOwner` is a Gnosis Safe, see https://goerli.etherscan.io/address/0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f

The owners are:

- [0xc37f6a6c4AB335E20d10F034B90386E2fb70bbF5](https://goerli.etherscan.io/address/0xc37f6a6c4AB335E20d10F034B90386E2fb70bbF5)
- [0xA902A27a7631D502E3Ec17fc5d4c3e0861752c94](https://goerli.etherscan.io/address/0xA902A27a7631D502E3Ec17fc5d4c3e0861752c94)
- [0x9bbFB9919062C29a5eE15aCD93c9D7c3b14d31aa](https://goerli.etherscan.io/address/0x9bbFB9919062C29a5eE15aCD93c9D7c3b14d31aa)
- [0x6964f301EdEBF16C563f97cE969C65ECdB39E918](https://goerli.etherscan.io/address/0x6964f301EdEBF16C563f97cE969C65ECdB39E918)

tynes: Each owner has been funded with 0.5 ETH

tynes: It is a 3 of 4

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,48 @@
...
+  "batchSenderAddress": "0x7431310e026B69BFC676C0013E12A1A11411EEc9",
 
-  "l2GenesisBlockGasLimit": "0x17D7840",
-  "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
+  "l2OutputOracleSubmissionInterval": 12,
+  "l2OutputOracleStartingBlockNumber": "FILL_IN_DAY_OF",
+  "l2OutputOracleStartingTimestamp": "FILL_IN_DAY_OF",
 
-  "l2CrossDomainMessengerOwner": "DUMMY",
+  "l2OutputOracleProposer": "0x02b1786A85Ec3f71fBbBa46507780dB7cF9014f6",
```
tynes, line 25: This is the current state batch submitter address

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,48 @@
...
+  "l2OutputOracleChallenger": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
 
-  "governanceTokenName": "Optimism",
-  "governanceTokenSymbol": "OP",
-  "governanceTokenOwner": "0x038a8825A3C3B0c08d52Cc76E5E361953Cf6Dc76",
+  "finalizationPeriodSeconds": 12,
+
+  "proxyAdminOwner": "0xf80267194936da1E98dB10bcE06F3147D580a62e",
+  "baseFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
```
tynes, line 31: Confirming that the fee recipients are the L1 multisig

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,48 @@
...
+  "l1FeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "sequencerFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
 
   "gasPriceOracleOverhead": 2100,
```
tynes, line 35: We can definitely optimize the overhead as it was previously computed based on the existing batch submission which uses way more gas. Its fine to start with `2100` but we definitely want to calculate a better value in the future

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,48 @@
...
   "gasPriceOracleScalar": 1000000,
+
+  "governanceTokenSymbol": "OP",
+  "governanceTokenName": "Optimism",
+  "governanceTokenOwner": "0x038a8825A3C3B0c08d52Cc76E5E361953Cf6Dc76",
+
+  "l2GenesisBlockGasLimit": "0x17D7840",
+  "l2GenesisBlockCoinbase": "0x4200000000000000000000000000000000000011",
```
tynes, line 43: I thought that I removed `l2GenesisBlockCoinbase` from being used and it just hardcodes this value

tynes: We can remove this value, see https://github.com/ethereum-optimism/optimism/commit/58b958978ca5c83a1a8ac3a4c690abd998f42c2a

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,48 @@
...
+  "l2GenesisBlockBaseFeePerGas": "0x3b9aca00",
```
tynes, line 44: `l2GenesisBlockBaseFeePerGas` in decimal is `1000000000` or `1_000_000_000`

mslipper: 1 gwei, which is correct.

packages/contracts-bedrock/deploy-config/goerli.json
mslipper: Coinbase removed.

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,47 @@
...
-  "p2pSequencerAddress": "DUMMY",
+  "p2pSequencerAddress": "0x715b7219D986641DF9eFd9C7Ef01218D528e19ec",
   "batchInboxAddress": "0xff00000000000000000000000000000000000420",
-  "batchSenderAddress": "DUMMY",
-  "l2OutputOracleSubmissionInterval": 20,
-  "l2OutputOracleStartingTimestamp": -1,
-  "l2OutputOracleProposer": "DUMMY",
-  "l2OutputOracleChallenger": "DUMMY",
-  "finalizationPeriodSeconds": 2,
+  "batchSenderAddress": "0x7431310e026B69BFC676C0013E12A1A11411EEc9",
```
tynes, line 19: The `batchSenderAddress` is the current tx batch submitter address

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,47 @@
...
+  "baseFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "l1FeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "sequencerFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
 
   "gasPriceOracleOverhead": 2100,
   "gasPriceOracleScalar": 1000000,
+
+  "governanceTokenSymbol": "OP",
+  "governanceTokenName": "Optimism",
+  "governanceTokenOwner": "0x038a8825A3C3B0c08d52Cc76E5E361953Cf6Dc76",
```
tynes, line 40: I don't think this owner is correct - its showing that its the address manager. We want it to be the mint manager

https://goerli.etherscan.io/address/0x038a8825A3C3B0c08d52Cc76E5E361953Cf6Dc76

smartcontracts: Wrong network: https://goerli-optimism.etherscan.io/address/0x038a8825a3c3b0c08d52cc76e5e361953cf6dc76#code

tynes: I'm confused, this may be an etherscan bug?

The address in the artifact says the mint manager is there, see https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/deployments/optimism-goerli/MintManager.json

tynes: Confirmed that this is an etherscan bug

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,47 @@
...
-  "l2CrossDomainMessengerOwner": "DUMMY",
+  "l2OutputOracleProposer": "0x02b1786A85Ec3f71fBbBa46507780dB7cF9014f6",
+  "l2OutputOracleChallenger": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
 
-  "governanceTokenName": "Optimism",
-  "governanceTokenSymbol": "OP",
-  "governanceTokenOwner": "0x038a8825A3C3B0c08d52Cc76E5E361953Cf6Dc76",
+  "finalizationPeriodSeconds": 12,
+
+  "proxyAdminOwner": "0xf80267194936da1E98dB10bcE06F3147D580a62e",
```
tynes, line 30: Would be more ideal if the `proxyAdminOwner` is a multisig but this is an L2 account and gnosis is not currently on testnet goerli optimism L2

tynes (approved):

packages/contracts-bedrock/deploy-config/goerli.json
K-Ho: I would like us to send a test transaction from this gnosis safe before the upgrade tomorrow to 100% guarantee that every party has access to their keys.

packages/contracts-bedrock/deploy-config/goerli.json
K-Ho: noting that having a multisig as the controller will make the upgrade slower than our rehearsals, but I like that it's closer to 1:1 with the process we will have on mainnet

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,47 @@
...
+
+  "l1StartingBlockTag": "FILL_IN_DAY_OF",
 
-  "l1StartingBlockTag": "DUMMY",
   "l1ChainID": 5,
   "l2ChainID": 420,
   "l2BlockTime": 2,
 
-  "maxSequencerDrift": 1200,
+  "maxSequencerDrift": 600,
```
K-Ho, line 13: fine for this upgrade, but in the future can we update these values to include units? `maxSequencerDriftSeconds` is way more clear

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -1,36 +1,47 @@
...
   "sequencerWindowSize": 3600,
-  "channelTimeout": 120,
+  "channelTimeout": 300,
```
K-Ho, line 15: in the future would be great to change to`channelTimeoutL1Blocks`

packages/contracts-bedrock/deploy-config/goerli.json
K-Ho, line 30: Where is this address from?

mslipper: It's the legacy address manager owner.

PR #4642, opened 2023-01-09
contracts-bedrock: make erc721 bridge more predeploy like

tynes:
**Description**

Hardcodes the constructor args for the predeploy
based erc721 bridge, since less arguments are needed because we know exactly what addresses the bridge
is going to be at. This follows the same pattern
as the erc20 bridge.


mslipper:
Closing - will require big test refactor.

PR #4643, opened 2023-01-09
op-node: ComputeL2OutputRoot cleanup

tynes:
**Description**

Cleans up the implementation of the `ComputeL2OutputRoot` function. No need for the intermediate `bytes.Buffer`.

cc @protolambda


trianglesphere (approved):
lgtm

PR #4644, opened 2023-01-10
op-node: workaround alchemy eth_getProof by blockhash RPC issue

protolambda:
**Description**

Alchemy does not *yet* support `eth_getProof` calls with blockhash parameter like other RPC providers due to incomplete execution-apis spec (fixed in https://github.com/ethereum/execution-apis/pull/326 )

To work-around it, we check if it's Alchemy & if it's a blockhash argument, then first fetch the block header by hash, and then the proof by number, and verify the retrieved proof against the header to make sure the data is correct.

**Tests**

Alchemy-only functionality, run with `--l1.rpckind=alchemy` to test.

**Metadata**

Fix ENG-3178


protolambda:
Alchemy just let us know they fixed the RPC routing to accept block hashes, this work-around is no longer necessary. Tested it with Alchemy Goerli RPC. Closing this.

PR #4650, opened 2023-01-10
caching getBlockByNumber with 'latest'

zhwrd:
**Description**

Quick and dirty implementation of caching eth_getBlockByNumber ["latest", Any] if numConfirmations is equal to 0

### DIFF
--- a/proxyd/methods.go
+++ b/proxyd/methods.go
@@ -67,12 +67,26 @@ func (e *EthGetBlockByNumberMethodHandler) cacheable(req *RPCReq) (bool, error) 
 	if err != nil {
 		return false, err
 	}
-	return !isBlockDependentParam(blockNum), nil
+	return blockNum != "pending" && (blockNum != "latest" || e.numBlockConfirmations == 0), nil
 }
 
 func (e *EthGetBlockByNumberMethodHandler) GetRPCMethod(ctx context.Context, req *RPCReq) (*RPCRes, error) {
 	if ok, err := e.cacheable(req); !ok || err != nil {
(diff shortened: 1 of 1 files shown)


proxyd/methods.go
```diff
@@ -99,7 +113,7 @@ func (e *EthGetBlockByNumberMethodHandler) PutRPCMethod(ctx context.Context, req
 		if err != nil {
 			return err
 		}
-		if curBlock <= blockNum+uint64(e.numBlockConfirmations) {
+		if curBlock < blockNum+uint64(e.numBlockConfirmations) {
```
zhwrd, line 116: this is a breaking change but only way its going to work... but its more consistent with how eth_blockNumber caching works...

tynes:
Are you trying to get this merged? Unsure why caching the latest tag is useful unless the chain has halted and is expected to be halted

zhwrd:
By far most of the RPC requests that proxyd forwards to mainnet-sequencer.optimism.io are getBlockByNumber [latest]
(nearly 2000 rps) 

Even though the cache is only for a short time, im thinking it will reduce traffic to sequencer quite a bit since the volume is so high.. But it also pre-warms the rest of the cache so effectively we should see very few requests forwarded to the sequencer at all.

Anyways I mainly worked on this to have as an option since you can see from [these graphs](https://mainnet-prod-grafana.optimism.io/d/2oPDOnFVk/sequencer-cpu-usage-debugging?orgId=1&refresh=5s) that sequencer CPU use keeps climbing

PR #4651, opened 2023-01-10
ctb: Add internal devnet

mslipper:
Adds an internal devnet for testing purposes. This network was not migrated, and was instead deployed from a clean slate.


tynes:
I think this needs a rebase for ci to work

PR #4653, opened 2023-01-10
op-node: fix AttributesQueue Reset function

protolambda:
**Description**

When the pipeline is reset with a reorg just after the next attributes fail to be generated then this batch may stick in the queue, causing a potential divergence with nodes that don't have this stale batch left in their pipeline stage.

**Tests**

We already have action testing that fuzzes the RPC errors and shows it eventually syncs, but resets were not captured there yet, as they only trigger on a reorg. Will need to investigate reorg-fuzzing to catch things like this.


tynes (approved):
Reading through the code, this looks right to me

trianglesphere (approved):
lgtm. Test is a known flake.

PR #4654, opened 2023-01-10
op-node,specs: fix framequeue reset, fix channel-bank and frame-queue specs

protolambda:
**Description**

The channel-bank spec was inconsistent with the implementation and design-doc changes. And missing the frame-queue abstraction.

Fix:
- spec clarifies we prune until total_size <= MAX_CHANNEL_BANK_SIZE
- spec matches pruning/reading etc. order of design doc and implementation
- ChannelBank now drops all timed-out-frames before reading more data, not just the first.
- FrameQueue is reset properly now (and Reset signature matches now)
- FrameQueue is specified now
- Move the relevant channel-bank parts from L1 retrieval into the channel bank spec


trianglesphere (approved):
lgtm. Nice finds and spec cleanups.

mslipper:
It's that flaky test again - going to merge this since the flaky test has since been addressed.

PR #4655, opened 2023-01-10
op-e2e: Increase time buffer in missing batch test

mslipper:


protolambda (approved):
I think this test became flaky after the sequence window logic was changed to not output the L2 block into the next epoch pro-actively anymore. This timing change makes a lot of sense to fix that flakiness.

PR #4656, opened 2023-01-10
fix(indexer): Fix flag type

roninjin10:
Flag type is a Bool for a couple of flags.  I think it's just copy paste pasta.


clabby:
Good catch! cc @mslipper who added these flags for clarification- this looks like it was unintentional?

mslipper:
Yeah very good catch. This was unintentional, thanks for the fix!

mslipper (approved):

PR #4658, opened 2023-01-10
test(ctb): Add a fuzz test for large deposit gaps

maurelian:
**Description**

Added a test demonstrating that the resource metering code is tolerant of very large times between deposits, without locking up due to overflows.


tynes (approved):

tynes:
Good thinking, just looks like the gas snapshot needs to be updated

packages/contracts-bedrock/contracts/test/ResourceMetering.t.sol
```diff
@@ -108,4 +108,18 @@ contract ResourceMetering_Test is CommonTest {
         vm.expectRevert("ResourceMetering: cannot buy more gas than available gas limit");
         meter.use(target * elasticity + 1);
     }
+
+    // Demonstrates that the resource metering arithmetic can tolerate very large gaps between
+    // deposits.
+    function testFuzz_meter_largeBlockDiff_succeeds(uint64 _amount, uint256 _blockDiff) external {
+        // This test fails if the following line is commented out.
+        // At 12 seconds per block, this number is effectively unreachable.
+        vm.assume(_blockDiff < 433576281058164217753225238677900874458691);
```
roninjin10, line 117: even after reading documentation I don't understand what vm.assume is doing https://book.getfoundry.sh/cheatcodes/assume

maurelian: It basically constrains the inputs. Any values which contradict this boolean statement will be discarded, IDK for sure but I think that means the function returns early, or else if it fails it doesn't affect the outcome of the test. 

You can try it locally by deleting that line and running: 

```
➜ forge test --mt testFuzz_meter_largeBlockDiff_succeeds -vvvv
```

tynes: pretty sure it tells foundry to not use that fuzz input again

PR #4659, opened 2023-01-10
op-node: Set RPC timeouts

trianglesphere:
**Description**

This is done in BaseClient because all RPC calls in the op-node flow through the base client. I have set it to 10s for the CallContext and 20s for the BatchCallContext. These values are large enough to handle complex computation, but will prevent a slow loris or throttling attack. Smaller timeouts can stil be set further up the chain.

**Tests**

None.

### DIFF
--- a/op-node/client/rpc.go
+++ b/op-node/client/rpc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"time"
 
 	"github.com/ethereum-optimism/optimism/op-service/backoff"
 	"github.com/ethereum/go-ethereum"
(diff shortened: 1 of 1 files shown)


op-node/rollup/driver/state.go
```diff
@@ -327,7 +327,7 @@ func (s *Driver) eventLoop() {
 			s.metrics.SetDerivationIdle(false)
 			s.idleDerivation = false
 			s.log.Debug("Derivation process step", "onto_origin", s.derivation.Origin(), "attempts", stepAttempts)
-			err := s.derivation.Step(context.Background())
+			err := s.derivation.Step(ctx)
```
protolambda, line 330: The cancellation of this `ctx` is synchronous with the function call, and will thus never happen. We could maybe put a timeout on the `Step`, but shouldn't set a step-timeout until the find-sync-start is no longer part of the stepping.

trianglesphere: yea, this is true. Maybe it should be `context.TODO()` to make it clear it's not being used for now

protolambda: `context.TODO()` might trigger some lint. Would just keep it as-is until we refactor the find-sync-start bit to be separate.

protolambda (approved):

tynes (approved):

PR #4660, opened 2023-01-10
fix(bm): tweak package json

smartcontracts:
**Description**
Tweaks the Balance Monitor package json to correctly export dist and src folders.


maurelian (approved):
I see. Thx :)

PR #4665, opened 2023-01-11
indexer: Fix startup errors

mslipper:
- The block locator was being initialized with a zero hash, which caused `Update` return `nil` and crash the process.
- Adds an L2 conf depth parameter, since L2 can now reorg.


indexer/services/l1/service.go
```diff
@@ -260,22 +268,28 @@ func (s *Service) Update(newHeader *types.Header) error {
...
-		finalizedWithdrawals, err := s.portal.GetFinalizedWithdrawalsByBlockRange(s.ctx, startHeight, endHeight)
-		if err != nil {
-			errCh <- err
-			return
-		}
-		finalizedWithdrawalsCh <- finalizedWithdrawals
-	}()
+
+	if s.isBedrock {
+		go func() {
```
mslipper, line 273: cc @clabby - we shouldn't be checking the portal at all when Bedrock isn't enabled.

roninjin10 (approved):

tynes (approved):

PR #4669, opened 2023-01-11
op-node: gossip flood-publish, cli gossip configuration

protolambda:
**Description**

PR diff is large due to p2p-config loading that moved into its own package (to avoid import cycle). No functionality changes to the config loading, aside from new gossip config variables that map to gossipsub configurables (see [gossipsub spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#parameters)):

- D: Configure GossipSub topic stable mesh target count, a.k.a. desired outbound degree, number of peers to gossip to
- Dlo: Configure GossipSub topic stable mesh low watermark, a.k.a. lower bound of outbound degree
- Dhi: Configure GossipSub topic stable mesh high watermark, a.k.a. upper bound of outbound degree, additional peers will not receive gossip
- Dlazy: Configure GossipSub gossip target, a.k.a. target degree for gossip only (not messaging like p2p.gossip.mesh.d, just announcements of IHAVE
- floodpublish: to force locally produced messages (i.e. if sequencing, not other messages) to be published to *every* known connected peer on the topic (meeting a minimal publish-score that honest peers do).

These flags are quite technical and have defaults that work for most regular users, so they're hidden from the CLI user.

Using these flags, one can get better connected on gossip, i.e. mesh with more peers on the blocks topic to serve them.

This p2p cli config reorg is useful since I expect we'll also make other gossip parameters configurable later, like scoring, and optional adoption of episub extension (save bandwidth).

**Tests**

This adds optional advanced configuration, we can test the D/flood-publish params, but that's covered in gossipsub itself already. We can tweak the params on testnets and watch the gossip metrics to compare results.

Fix ENG-3186

### DIFF
--- a/op-node/p2p/cli/load_config.go
+++ b/op-node/p2p/cli/load_config.go
@@ -1,0 +1,301 @@
+package cli
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"strings"
+
+	ds "github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/sync"
+	leveldb "github.com/ipfs/go-ds-leveldb"
+	lconf "github.com/libp2p/go-libp2p/config"
+	"github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/multiformats/go-multiaddr"
+
+	"github.com/ethereum-optimism/optimism/op-node/flags"
+	"github.com/ethereum-optimism/optimism/op-node/p2p"
+
+	"github.com/urfave/cli"
+
+	"github.com/ethereum/go-ethereum/p2p/enode"
+)
+
+func NewConfig(ctx *cli.Context) (*p2p.Config, error) {
+	conf := &p2p.Config{}
+
+	if ctx.GlobalBool(flags.DisableP2P.Name) {
+		conf.DisableP2P = true
+		return conf, nil
+	}
+
+	p, err := loadNetworkPrivKey(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load p2p priv key: %w", err)
+	}
+	conf.Priv = p
+
+	if err := loadListenOpts(conf, ctx); err != nil {
+		return nil, fmt.Errorf("failed to load p2p listen options: %w", err)
+	}
+
+	if err := loadDiscoveryOpts(conf, ctx); err != nil {
+		return nil, fmt.Errorf("failed to load p2p discovery options: %w", err)
+	}
+
+	if err := loadLibp2pOpts(conf, ctx); err != nil {
+		return nil, fmt.Errorf("failed to load p2p options: %w", err)
+	}
+
+	if err := loadGossipOptions(conf, ctx); err != nil {
+		return nil, fmt.Errorf("failed to load p2p gossip options: %w", err)
+	}
+
+	conf.ConnGater = p2p.DefaultConnGater
+	conf.ConnMngr = p2p.DefaultConnManager
+
+	return conf, nil
+}
+
+func validatePort(p uint) (uint16, error) {
+	if p == 0 {
+		return 0, nil
+	}
+	if p >= (1 << 16) {
+		return 0, fmt.Errorf("port out of range: %d", p)
+	}
+	if p < 1024 {
+		return 0, fmt.Errorf("port is reserved for system: %d", p)
+	}
+	return uint16(p), nil
+}
+
+func loadListenOpts(conf *p2p.Config, ctx *cli.Context) error {
+	listenIP := ctx.GlobalString(flags.ListenIP.Name)
+	if listenIP != "" { // optional
+		conf.ListenIP = net.ParseIP(listenIP)
+		if conf.ListenIP == nil {
+			return fmt.Errorf("failed to parse IP %q", listenIP)
+		}
+	}
+	var err error
+	conf.ListenTCPPort, err = validatePort(ctx.GlobalUint(flags.ListenTCPPort.Name))
+	if err != nil {
+		return fmt.Errorf("bad listen TCP port: %w", err)
+	}
+	conf.ListenUDPPort, err = validatePort(ctx.GlobalUint(flags.ListenUDPPort.Name))
+	if err != nil {
+		return fmt.Errorf("bad listen UDP port: %w", err)
+	}
+	return nil
+}
+
+func loadDiscoveryOpts(conf *p2p.Config, ctx *cli.Context) error {
+	if ctx.GlobalBool(flags.NoDiscovery.Name) {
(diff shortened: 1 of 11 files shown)


op-node/p2p/cli/load_config.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	if dbPath == "memory" {
+		dbPath = ""
+	}
+	conf.DiscoveryDB, err = enode.OpenDB(dbPath)
+	if err != nil {
+		return fmt.Errorf("failed to open discovery db: %w", err)
+	}
+
+	conf.Bootnodes = p2p.DefaultBootnodes
```
tynes, line 141: Are there any cases in which users would not want to use the default boot nodes?

tynes: ie whoever runs the default boot nodes gets IPs of every node on the network right?

op-node/p2p/cli/load_config.go
```diff
@@ -0,0 +1,301 @@
...
+
+	return nil
+}
+
+func loadNetworkPrivKey(ctx *cli.Context) (*crypto.Secp256k1PrivateKey, error) {
+	raw := ctx.GlobalString(flags.P2PPrivRaw.Name)
+	if raw != "" {
+		return parsePriv(raw)
+	}
+	keyPath := ctx.GlobalString(flags.P2PPrivPath.Name)
```
tynes, line 248: Default keyPath if its empty string? Other places in the config parsing handle empty string. The defaults could all be handled when defining the flags, but also doesn't prevent somebody from passing the empty string so prob best to check for it here

op-node/p2p/cli/load_config.go
protolambda: Didn't put the default likes this in the PR (just moved the code), but I think it makes sense from a UX perspective. Agreed we could make it fully override-able. Want me to change it to be the default value, and get wiped if you specify another value? @mslipper might rely on current behavior.

op-node/p2p/cli/load_config.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	p, err := crypto.UnmarshalSecp256k1PrivateKey(b)
+	if err != nil {
+		// avoid logging the priv key in the error, but hint at likely input length problem
+		return nil, fmt.Errorf("failed to parse priv key from %d bytes", len(b))
+	}
+	return (p).(*crypto.Secp256k1PrivateKey), nil
+}
+
+func loadGossipOptions(conf *p2p.Config, ctx *cli.Context) error {
```
tynes, line 294: Any sanity checks in here for values that make no sense? Like setting the lo flag higher than the hi flag? Or setting them to 0? Since the no discovery flag should turn it off, that should be used when wanting to configure it off and the values passed here should always be able to work

op-node/p2p/cli/load_config.go
protolambda: This code is not part of the changes either, but agree we could change it. The flag definition defines the default as non-empty. If someone changes it explicitly to be empty, it's likely a configuration error, so returning an error is maybe the right behavior.

op-node/p2p/cli/load_signer.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/urfave/cli"
+
+	"github.com/ethereum-optimism/optimism/op-node/flags"
+	"github.com/ethereum-optimism/optimism/op-node/p2p"
+)
+
+// TODO: implement remote signer setup (config to authenticated endpoint)
```
tynes, line 13: Perhaps we want to implement this in the `op-service` code so then it can be imported into multiple codebases. "Signer type" or something like that, where its the authenticated remote signer (uses `eth_signTransaction` with mTLS auth) and a local signer

op-node/p2p/cli/load_config.go
protolambda: These are checked in `Config.Check()`

op-node/p2p/cli/load_signer.go
```diff
@@ -0,0 +1,33 @@
...
+// and remote signer itself (e.g. a open http client to make signing requests)
+
+// LoadSignerSetup loads a configuration for a Signer to be set up later
+func LoadSignerSetup(ctx *cli.Context) (p2p.SignerSetup, error) {
+	key := ctx.GlobalString(flags.SequencerP2PKeyFlag.Name)
+	if key != "" {
+		// Mnemonics are bad because they leak *all* keys when they leak.
```
tynes, line 20: Mnemonics are easier to back up than hex strings

op-node/p2p/cli/load_signer.go
protolambda: It's an old todo, not part of this PR. At least it's isolated now, should be easy to change once we integrate signers more deeply.

op-node/p2p/cli/load_signer.go
protolambda: Old code / todo, we should just have the signer integration and skip this debate

tynes: Hex strings are bad because you must put them into your paste buffer when you want to use them

op-node/p2p/cli/load_config.go
tynes: Ok i see the sanity checks happen in a `Check()` function

op-node/p2p/cli/load_signer.go
tynes: agreed :)

op-node/p2p/cli/load_signer.go
tynes: Ah got it, i see now that most of this was restructuring

op-node/p2p/cli/load_config.go
tynes: wouldn't be opposed to erroring here if its empty even tho its a diff on top of moving the code

tynes (approved):
Generally looks good to me, a couple small comments

op-node/p2p/cli/load_config.go
protolambda: opened a separate PR for it, let's get this merged first: https://github.com/ethereum-optimism/optimism/pull/4676

mslipper:
In favor of these changes, but need to do a deeper review + understand their implications a bit better before merge. @protolambda OK if we ship this after cutting the prod Goerli releases? Can deploy it to one of our Goerli replicas to test it out before rolling it out to the entire fleet.

protolambda:
@mslipper SGTM! It's an optional thing anyway, and other users don't rely on it (the flags are hidden for a reason)

protolambda:
fixed a typo in the flag clarification commit

mslipper (approved):

PR #4682, opened 2023-01-11
creates indexer docker build cci configs

zhwrd:
**Description**

Creates docker build jobs in circleci for indexer


tynes:
Perhaps its too late, but renaming it to the `op-indexer` would be nice :) def out of scope of this PR

mslipper (approved):

PR #4684, opened 2023-01-12
op-proposer: Cleanup

trianglesphere:
**Description**

This removes the driver / service design from the proposer & puts it all in the L2 Output Submitter. This significantly reduces some of the boilerplate. This also removes some extra config options.

I also created a new config. There is now `Config` and `CLIConfig`. The `Config` is used to initialize the proposer programatically. `CLIConfig` is created by the main function from the CLI parameters.

This also includes a hacky fix to the issue where the proposer would stall if gas estimation failed. I solved this by adding a timeout to the transaction submission.

**Tests**

Current test suite passes. No new tests.

**Metadata**
- Fixes ENG-3189


op-proposer/proposer/l2_output_submitter.go
```diff
@@ -212,75 +220,244 @@ func NewL2OutputSubmitterWithSigner(
...
+	updateGasPrice := func(ctx context.Context) (*types.Transaction, error) {
+		l.log.Info("proposer updating batch tx gas price", "nonce", nonce)
+		return l.UpdateGasPrice(ctx, tx)
+	}
+
+	// Wait until one of our submitted transactions confirms. If no
+	// receipt is received it's likely our gas price was too low.
+	cCtx, cancel := context.WithTimeout(ctx, 100*time.Second)
+	defer cancel()
+	receipt, err := l.txMgr.Send(cCtx, updateGasPrice, l.SendTransaction)
```
trianglesphere, line 421: Properly set a timeout b/c if gas estimation fails inside the txmanager, it hangs forever.

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -212,75 +220,244 @@ func NewL2OutputSubmitterWithSigner(
...
 	}
+	return tx, nil
+}
 
-	return sources.NewRollupClient(client.NewBaseRPCClient(rpcCl)), nil
+// SendTransactionExt sends a transaction through the transaction manager which handles automatic
+// price bumping.
+// It also hardcodes a timeout of 100s.
+func (l *L2OutputSubmitter) SendTransactionExt(ctx context.Context, tx *types.Transaction) error {
+	// Construct the a closure that will update the txn with the current gas prices.
```
protolambda, line 410:
```suggestion
	// Construct the closure that will update the txn with the current gas prices.
```

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -212,75 +220,244 @@ func NewL2OutputSubmitterWithSigner(
...
+	defer l.wg.Done()
+
+	ctx := l.ctx
+
+	ticker := time.NewTicker(l.pollInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			output, shouldPropose, err := l.FetchNextOutputInfo(ctx)
```
protolambda, line 443: Maybe worth creating a context with timeout here to avoid the loop from getting stuck for too long

protolambda (approved):
LGTM. One nit about context usage and a typo

op-proposer/proposer/l2_output_submitter.go
trianglesphere: done

op-proposer/proposer/l2_output_submitter.go
trianglesphere: done

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -212,75 +220,250 @@ func NewL2OutputSubmitterWithSigner(
...
+	defer l.wg.Done()
+
+	ctx := l.ctx
+
+	ticker := time.NewTicker(l.pollInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			cCtx, cancel := context.WithTimeout(ctx, 3*time.Minute)
```
sebastianst, line 443: Magic duration, see above.

I see that this one is more tricky because it runs multiple L1 and rollup client queries. Maybe sth like `L1QueryTimeout+2*L2QueryTimeout`.

sebastianst (commented):
LGTM, only minor nits about dropped errors and magic numbers.

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -212,75 +220,250 @@ func NewL2OutputSubmitterWithSigner(
...
+			output, shouldPropose, err := l.FetchNextOutputInfo(cCtx)
+			if err != nil {
```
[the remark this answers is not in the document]
trianglesphere: In general, I don't think we want to exit the loop on error. I've added logs for them inside the loop, though they also have logs inside the individual functions.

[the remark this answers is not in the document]
sebastianst: Yes, sorry, we don't want to exit the loop.

OK just saw that you log every error in the called functions, like in `FetchNextOutputInfo`. I think that it's the caller's responsibility to log the error, not in the called function, which is why the error is returned back up the stack. This way you could save all the `log.Error` calls in the functions and log at the latest possible place up the stack, in `loop()`. Right now you somehow decide on how to handle errors already in the functions ("log them") but also still return them back to the caller for it to decide again.

But it's just a nit because I think it would be cleaner. No need to address it.

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -212,75 +220,250 @@ func NewL2OutputSubmitterWithSigner(
...
+	// Construct the closure that will update the txn with the current gas prices.
+	nonce := tx.Nonce()
+	updateGasPrice := func(ctx context.Context) (*types.Transaction, error) {
+		l.log.Info("proposer updating batch tx gas price", "nonce", nonce)
+		return l.UpdateGasPrice(ctx, tx)
+	}
+
+	// Wait until one of our submitted transactions confirms. If no
+	// receipt is received it's likely our gas price was too low.
+	cCtx, cancel := context.WithTimeout(ctx, 100*time.Second)
```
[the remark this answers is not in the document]
trianglesphere: Yea, there's a bunch of magic numbers. I've captured this in [this ticket](https://linear.app/optimism/issue/ENG-3253/op-proposer-batcher-fix-contexts-timeouts)

PR #4688, opened 2023-01-12
feat(e2e): Garbage batch test

clabby:
# Overview

Adds a test to check that when a malformed channel output frame is submitted to L1, the L2 safe head is not affected.

Current methods used to malform the frame:
- Stripping the derivation version byte
- Truncating the output frame bytes (removes 4 bytes from the encoded frame)
- Append 4 garbage bytes to the tail end of the encoded frame
- Use an invalid compression writer `zlib` -> `gzip`
- Malform RLP encoding (i.e., set one of the length values to an incorrect number, etc.)

**Metadata**
Closes ENG-2732


clabby:
* develop
  * **PR #4688** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4688" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4688?utm_source=stack-comment).

op-e2e/actions/l2_batcher.go
```diff
@@ -208,3 +226,113 @@ func (s *L2Batcher) ActSubmitAll(t Testing) {
...
+	DIRTY_APPEND,
+	INVALID_COMPRESSION,
+	MALFORM_RLP,
+}
+
+// Configuration for a garbage channel (`ChannelOut` in the `actions` pkg)
+type GarbageChannelCfg struct {
+	useInvalidCompression bool
+	malformRLP            bool
+}
```
clabby, line 255: RFC- where should these live?

op-e2e/actions/garbage_channel_out.go
```diff
@@ -0,0 +1,223 @@
...
+
+// WriterApi is the interface shared between `zlib.Writer` and `gzip.Writer`
+type WriterApi interface {
+	Close() error
+	Flush() error
+	Reset(io.Writer)
+	Write([]byte) (int, error)
+}
+
+// GarbageChannelOut is a modified `derive.ChannelOut` that can be configured to behave differently
```
tynes, line 30: What do you think about adding an interface to `op-node` that `derive.ChannelOut` implements and then import that interface here and make a compile time interface assertion that `GarbageChannelOut` implements

op-e2e/actions/l2_batcher.go
```diff
@@ -208,3 +226,113 @@ func (s *L2Batcher) ActSubmitAll(t Testing) {
 	s.ActL2ChannelClose(t)
 	s.ActL2BatchSubmit(t)
 }
+
+// TODO: Move this to a better location
+type GarbageKind int64
```
tynes, line 231: Should these go in `garbage_channel_out.go`?

op-e2e/actions/garbage_channel_out.go
```diff
@@ -0,0 +1,223 @@
...
+	"io"
+
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+var ErrNotDepositTx = errors.New("first transaction in block is not a deposit tx")
```
tynes, line 19: Aren't these defined in `op-node`?

clabby: Ah yep good shout- can de-duplicate these

op-e2e/actions/garbage_channel_out.go
clabby: I like this idea- done 🤝

op-e2e/actions/garbage_channel_out.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// WriterApi is the interface shared between `zlib.Writer` and `gzip.Writer`
+type WriterApi interface {
+	Close() error
+	Flush() error
+	Reset(io.Writer)
+	Write([]byte) (int, error)
+}
+
```
tynes, line 52: Prob want to add a compile time check that it satisfies the interface, like so: https://medium.com/@matryer/golang-tip-compile-time-checks-to-ensure-your-type-satisfies-an-interface-c167afed3aae

Could also do the same thing for the actual impl in `op-node`

op-e2e/actions/garbage_channel_out.go
```diff
@@ -0,0 +1,246 @@
...
+// GarbageChannelCfg is the configuration for a `GarbageChannelOut`
+type GarbageChannelCfg struct {
+	useInvalidCompression bool
+	malformRLP            bool
+}
+
+// WriterApi is the interface shared between `zlib.Writer` and `gzip.Writer`
+type WriterApi interface {
```
tynes, line 46: Going to defer to @trianglesphere and @protolambda re: naming of interfaces, i know people are opinionated on it and i don't believe we use the `Api` suffix for interfaces currently. More opinionated about it being standard

trianglesphere: yea, I'd avoid putting `API` on the end. We sometimes use `Iface` or `Interface`.
Here I'd use something like `Writer` or `WriteCloser` or `WriteFlushResetCloser` (this is the golang tradition of stuffing everything together: https://pkg.go.dev/io#WriteCloser).  We rely on the package name to separate the different types.

op-node/rollup/derive/channel_out.go
```diff
@@ -17,6 +17,20 @@ import (
 var ErrNotDepositTx = errors.New("first transaction in block is not a deposit tx")
 var ErrTooManyRLPBytes = errors.New("batch would cause RLP bytes to go over limit")
 
+// ChannelOutApi is the interface implemented by ChannelOut
```
trianglesphere: In general we should define interfaces at the point of their usage rather than their definition. So this interface should be in the `op-e2e/actions` package. I'd also remove the `API` from it.

op-e2e/actions/garbage_channel_out.go
clabby: Changed:
* `WriterApi` -> `Writer`
* `ChannelOutApi` -> `ChannelOutIface`

op-node/rollup/derive/channel_out.go
trianglesphere: @clabby not outdated

trianglesphere (approved):
lgtm

PR #4690, opened 2023-01-12
fix(ctb): L2OutputOracle event order bug

smartcontracts:
**Description**
Fixes a small bug where event parameters were being emitted in the wrong order. Adds a test to confirm correct ordering.


tynes (approved):

tynes:
The gas snapshot needs an update

```
Diff in "CrossDomainOwnableThroughPortal_Test::test_depositTransaction_crossDomainOwner_succeeds()": consumed "(gas: 61876)" gas, expected "(gas: 61882)" gas 
```

PR #4694, opened 2023-01-12
proxy: use view modifier with admin/implementation

Zena-Blockchain:
**Description**

fix: add view modifiers to view functions

If a view function does not have a view modifier, the function cannot be viewed in the read tab on the etherscan page.
The implementation address and owner address cannot be checked in the proxy.

**Tests**

Deployed to goerli network to check fixes.

Comparison of the existed (L1ChugSplashProxy) page on optimisem-goreli and deployed (L1ChugSplashProxy) page on optimisem-goreli with modified code applied. I was able to check the implementation address and the owner address on the modified address.

- existed address : https://goerli.etherscan.io/address/0x636Af16bf2f682dD3109e60102b8E1A089FedAa8#readContract

- modified address:
https://goerli.etherscan.io/address/0x03178B19f30f72878D6B59D856e1D90c6DAbc9e2#readContract

 
**Invariants**
 

**Additional context**
 

**Metadata**


packages/contracts-bedrock/contracts/universal/Proxy.sol
```diff
@@ -121,7 +121,7 @@ contract Proxy {
      *
      * @return Owner address.
      */
-    function admin() external proxyCallIfNotAdmin returns (address) {
```
tynes, line 124: This is currently designed in a way where we do not need to worry about function selector clashing. This change will result in the potential for function selector clashing

tynes (changes requested):
I don't think we are going to accept this change. It is too close for us going to production

Also don't commit in a new directory to `deployments` or a new package lock file

maurelian:
> I don't think we are going to accept this change. It is too close for us going to production

Agree. Going to close this, but we do appreciate your effort and suggestions. :)

PR #4702, opened 2023-01-12
feat(ctb): update L2OO interval

smartcontracts:
**Description**
Deploys a new L2OutputOracle implementation with a SUBMISSION_INTERVAL of 120 rather than 12. At a submission interval of 12, the output submitter has difficulty keeping up. While we work on parallelizing the output submitter, we'll keep this at 120.


mslipper (approved):
Build error is fixed on develop. Merging.

PR #4703, opened 2023-01-12
specs: Doc Enhancements

refcell:
**Changelog**

- Fixes missing intradoc links in [`specs/batcher.md`](./specs/batcher.md).
- Fixes specs reference in [`CONTRIBUTING.md`](./CONTRIBUTING.md)


specs/batcher.md
```diff
@@ -4,20 +4,18 @@
 The batch submitter, also referred to as the batcher, is the entity submitting the L2 sequencer data to L1,
 to make it available for verifiers.
 
-[derivation-spec]: ./derivation.md
```
maurelian, line 7: Does this format not work?

refcell: no it didn't render, at least for github

refcell: I think this might actually be a bug with github markdown rendering?

See: https://github.com/ethereum-optimism/optimism/blob/develop/specs/batcher.md

For example, these all render:

[derivation-spec]: ./derivation.md

Explicit: [derivation-spec](derivation-spec)
Explicit2: [derivation-spec](./derivation-spec.md)

Aliased: [derivation-spec][derivation-spec]

Referenced: [derivation-spec]

maurelian: Uh oh, isn't the aliased format used predominantly throughout the specs?

refcell: Ah right - I fixed it to be aliased properly. Believe it was the `-` prefix and relative filepath that borked it.

maurelian: Ah, good catch, but I think it'd be better if the hyphen went into the alias instead of the text itself.

specs/batcher.md
```diff
@@ -4,20 +4,20 @@
...
+[derivation-spec]: derivation.md
```
maurelian, line 7:
```suggestion
[derivation spec]: derivation.md
```

specs/batcher.md
```diff
@@ -4,20 +4,20 @@
...
 
-The format of the data transactions is defined in the [derivation spec]: the data is constructed from L2 blocks
-in the reverse order as it is derived from data into L2 blocks.
+The format of the data transactions is defined in the [derivation-spec]:
```
maurelian, line 9:
```suggestion
The format of the data transactions is defined in the [derivation spec]:
```

maurelian (approved):
I think the format in my suggestions is more readable, but I'll approve and let you decide how to resolve.

PR #4708, opened 2023-01-13
contracts-bedrock: Fix Bridge Event Emission on ETH Receive

refcell:
**Overview**

See https://www.notion.so/oplabs/Legacy-Bridge-Events-Port-981cad28e6fc4b5194162f0089be9228 for more context.

Fixes ENG-3205

Fixes the issue where `ETHDepositInitiated` is not emitted when eth is sent directly to the L1 bridge contract, and is subsequently not picked up by the sequencer.

This pr adds a `receive` override to `L1StandardBridge` that explicitly emits the `ETHDepositInitiated` event. Also changed the `L1StandardBridge` test from expecting a `ETHBridgeInitiated` event to a `ETHDepositInitiated` event.

Additionally, adds a `WithdrawalInitiated` event emission when the `L2StandardBridge` is sent eth directly (the `receive` function is called/dispatched).


packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -39,7 +39,9 @@ contract L1StandardBridge_Receive_Test is Bridge_Initializer {
         assertEq(address(op).balance, 0);
 
         vm.expectEmit(true, true, true, true);
-        emit ETHBridgeInitiated(alice, alice, 100, hex"");
+        emit ETHDepositInitiated(alice, alice, 100, hex"");
+        // Below should NOT be emitted since this is emitted in the StandardBridge receive and not the L1StandardBridge receive.
+        // emit ETHBridgeInitiated(alice, alice, 100, hex"");
```
tynes, line 44: Did you mean to leave this line commented out?

packages/contracts-bedrock/contracts/L1/L1StandardBridge.sol
```diff
@@ -100,6 +100,17 @@ contract L1StandardBridge is StandardBridge, Semver {
...
 
+    /**
+     * @notice Allows EOAs to deposit ETH by sending directly to the bridge.
+     */
+    receive() external payable override onlyEOA {
+        // Notice this emits an `ETHDepositInitiated` event.
+        // Whereas `StandardBridge` contract's `receive` fallback will only call
+        // `_initiateBridgeEth`, which will emit an `ETHBridgeInitiated` event.
+        emit ETHDepositInitiated(msg.sender, msg.sender, msg.value, bytes(""));
+        _initiateBridgeETH(msg.sender, msg.sender, msg.value, RECEIVE_DEFAULT_GAS_LIMIT, bytes(""));
```
tynes, line 111: I'm wondering if emitting it inside of `_initiateBridgeETH` would make this less bug prone in the future, having the least amount of places where it is emitted using shared code paths

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
refcell: Yes, but removed since it's documented in-line.

packages/contracts-bedrock/contracts/L1/L1StandardBridge.sol
maurelian: I agree with @tynes that we want to write it in as few places as possible, but the challenge is that if we put the `ETHDepositInitiated` event into `_initiateBridgeETH()` it would then also be emitted [here](https://github.com/ethereum-optimism/optimism/blob/5ed518a0ea7334d6e6448104a005a094ae4dbbb7/packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol#L172) by the `_initiateWithdrawal()` function. 

Here's one proposal to make this work:
Modify the internal bridging functions (`_initiateBridgeETH()` and `_initiateBridgeERC20()` to take an extra argument (maybe `enum Direction { Deposit, Withdrawal }`) and use that to determine which events to emit.

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -39,7 +39,7 @@ contract L1StandardBridge_Receive_Test is Bridge_Initializer {
         assertEq(address(op).balance, 0);
 
         vm.expectEmit(true, true, true, true);
-        emit ETHBridgeInitiated(alice, alice, 100, hex"");
```
maurelian, line 42: Ah, the thing is that _both_ events should be emitted here.

maurelian: I think this test _might_ still pass if you restore this line.

refcell: good catch!

packages/contracts-bedrock/contracts/L1/L1StandardBridge.sol
```diff
@@ -100,6 +100,21 @@ contract L1StandardBridge is StandardBridge, Semver {
         StandardBridge(_messenger, payable(Predeploys.L2_STANDARD_BRIDGE))
     {}
 
+    /**
+     * @notice Allows EOAs to deposit ETH by sending directly to the bridge.
+     */
+    receive() external payable override onlyEOA {
+        // Notice this emits an `ETHDepositInitiated` event.
+        // Whereas `StandardBridge` contract's `receive` fallback will only call
+        // `_initiateBridgeEth`, which will emit an `ETHBridgeInitiated` event.
```
maurelian, line 109: I think this comment should move into the natspec, and then have a separate comment explaining what the assembly is doing/why it's needed.

```suggestion
    /**
     * @notice Allows EOAs to deposit ETH by sending directly to the bridge.
     *
     * 		   Emits both an `ETHDepositInitiated` and `ETHBridgeInitiated` event.
     */
    receive() external payable override onlyEOA {
    	// Explain b and the assembly...
```

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -38,6 +38,8 @@ contract L1StandardBridge_Receive_Test is Bridge_Initializer {
     function test_receive_succeeds() external {
         assertEq(address(op).balance, 0);
 
+        vm.expectEmit(true, true, true, true);
+        emit ETHDepositInitiated(alice, alice, 100, hex"");
```
tynes, line 42: Could we also be ok with emitting `ETHDepositInitiated` on L2? Then we could emit the event inside of `_initiateBridgeETH` which would reduce diff 

Deposit and withdrawal is all relative so i think it makes sense

cc @maurelian @smartcontracts @clabby

maurelian: It solves the issue easily and minimizes risk so I'm for it. 

All three `ETH` and all three `ERC20` events should go together then I believe.

refcell: I think we need to verify that this won't bork indexing first. Who would be best to ask about that? (we can take this offline)

tynes: I don't think this will impact indexing at all, fwiw there is test coverage in the indexer, integration tests

tynes:
Do you mind squashing this PR into a few clean commits?

packages/contracts-bedrock/contracts/L1/L1StandardBridge.sol
```diff
@@ -100,6 +100,20 @@ contract L1StandardBridge is StandardBridge, Semver {
         StandardBridge(_messenger, payable(Predeploys.L2_STANDARD_BRIDGE))
     {}
 
+    /**
```
tynes, line 103: Can you include a bump to minor semver here as well?

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
tynes: > It solves the issue easily and minimizes risk so I'm for it.
> 
> All three `ETH` and all three `ERC20` events should go together then I believe.
> 
> Does it also remove the need for the assembly block?

Yes it removes the need for the assembly block and keeps the code much more simple. It should not break the indexer. Let me know if you have any more questions @refcell :D

packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol
```diff
@@ -68,6 +68,27 @@ contract L2StandardBridge is StandardBridge, Semver {
...
+            b.offset := 0
+            b.length := 0
+        }
+        _initiateWithdrawal(
+            Predeploys.LEGACY_ERC20_ETH,
+            msg.sender,
+            msg.sender,
+            msg.value,
+            RECEIVE_DEFAULT_GAS_LIMIT,
+            b
```
tynes, line 88: Could we get rid of the assembly by passing through `msg.data`? I know we don't do that elsewhere and I think the data will just end up in an event

refcell: Ah whoops missed this one, yes

maurelian (approved):
Once [this comment is resolved](https://github.com/ethereum-optimism/optimism/pull/4708/files#r1073971099) I think this PR is good to go, but I'll leave it up to @tynes to approve in case he has other changes to request. 

@refcell and I discussed work (for another PR) which will be required to ensure that the legacy and bedrock events are all emitted correctly

packages/contracts-bedrock/tasks/check-l2.ts
```diff
@@ -10,7 +10,7 @@ import { predeploys } from '../src'
 
 // expectedSemver is the semver version of the contracts
 // deployed at bedrock deployment
-const expectedSemver = '1.0.0'
+const expectedSemver = '1.1.0'
```
tynes, line 13: This needs to be changed back to `1.0.0` and then the particular contract that was changed, `1.1.0` needs to be passed in as an override, so the `L2StandardBridge` section

tynes:
https://github.com/ethereum-optimism/optimism/pull/4744 is a better approach

tynes (changes requested):
this is broken

PR #4709, opened 2023-01-13
feat: add view modifiers to view functions

Zena-Blockchain:
Added view modifiers because it was difficult to check the address and owner of the proxy on the etherscan site.

**Description**

**Tests**

**Invariants**

**Additional context**

**Metadata**

### DIFF
--- a/packages/contracts-bedrock/contracts/universal/Proxy.sol
+++ b/packages/contracts-bedrock/contracts/universal/Proxy.sol
@@ -121,7 +121,7 @@ contract Proxy {
      *
      * @return Owner address.
      */
-    function admin() external proxyCallIfNotAdmin returns (address) {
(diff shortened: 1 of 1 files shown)


tynes:
This allows for function selector clashing

maurelian:
@Zena-Blockchain this is a product decision. We'll evaluate.

smartcontracts:
Note also that this modifies a contract that would not be modifiable on mainnet, meaning we would have a diff between the ABI in our contracts and the ABI deployed on mainnet.

tynes (changes requested):
I think that etherscan should have a special field for pulling the eip1967 slots and rendering the values directly instead of us needing to lessen the security model for etherscan

tynes:
Closing this as it is an etherscan issue

PR #4710, opened 2023-01-13
op-node: handle recovered geth restarts

protolambda:
When geth crashes (e.g. if you're debugging a lot and kill it..., or your power shuts down etc.) then geth gets rid of blocks it might have finalized, to get a consistent chain again. But at the same time it fails to rewind the finalized block marker (to which the safe block marker defaults after restarts).

Note that if geth recovers, it seesms to generally go back to the last successful state checkpoint, which could be genesis. So then you enter this with unsafe==bedrock genesis, and unsafe < finalized, causing you to traverse past bedrock genesis, into pre-bedrock data that you can't parse.

So this fixes this by detecting when the head is older than the finalized chain, to restart from there.

### DIFF
--- a/op-node/rollup/sync/start.go
+++ b/op-node/rollup/sync/start.go
@@ -109,6 +109,12 @@ func FindL2Heads(ctx context.Context, cfg *rollup.Config, l1 L1Chain, l2 L2Chain
 		return nil, fmt.Errorf("failed to fetch current L2 forkchoice state: %w", err)
 	}
 
+	// Geth on bad restarts can rewind the head back to last valid state, but fails to rewind the finalization or safe marker.
+	// In this case we must go back to the prior head, to reprocess the pruned finalized/safe data.
+	if result.Unsafe.Number < result.Finalized.Number || result.Unsafe.Number < result.Safe.Number {
+		return &FindHeadsResult{Unsafe: result.Unsafe, Safe: result.Unsafe, Finalized: result.Unsafe}, nil
+	}
+
 	// Remember original unsafe block to determine reorg depth
 	prevUnsafe := result.Unsafe
 


protolambda:
Maybe I should open a bug report upstream instead to handle finalized block label upon chain recovery. This is kind of a workaround

tynes:
> Maybe I should open a bug report upstream instead to handle finalized block label upon chain recovery. This is kind of a workaround

I think this is a good idea but having the code in the `op-node` is important until the fix is in and a hardfork forces everybody to upgrade. Its probably also good in case other ELs also have this same problem...

tynes (approved):

trianglesphere:
I think we see if we can fix it upstream. This will work if we have either a hard reset OR an actual reorg. I'm worried about what happens if we have both at the same time. This is significantly more rare than if we have one or the other, but it could be bad.

op-node/rollup/sync/start.go
```diff
@@ -109,6 +109,12 @@ func FindL2Heads(ctx context.Context, cfg *rollup.Config, l1 L1Chain, l2 L2Chain
 		return nil, fmt.Errorf("failed to fetch current L2 forkchoice state: %w", err)
 	}
 
+	// Geth on bad restarts can rewind the head back to last valid state, but fails to rewind the finalization or safe marker.
```
trianglesphere, line 112: I think we may still need to do the walk back from this point.

protolambda: good point, the if the rewind is short, but still longer than the distance to finalized head, then it'll hit. A regular reorg should (tm) have proper finalized label change though. I'll change this PR to draft and investigate a geth fix.

tynes:
> 

Thoughts on if other ELs have similar problems?

protolambda:
> Thoughts on if other ELs have similar problems?

Perhaps. The finality label is mostly just a label added to geth as part of the Merge, and it seems like older crash recovery code isn't that well adapted to the proof-of-stake change. So the same can apply to other ELs.

PR #4713, opened 2023-01-13
feat(ctb): set up SystemConfig in step 1

smartcontracts:
**Description**
Moves the SystemConfig upgrade/initialization logic into Step 1 of the SystemDictator contract. This means that the Sequencer is correctly able to spin up between steps 2 and 3.


tynes:
This should also update the post check hook right?

tynes:
Implementation looks good, will approve pending CI

PR #4719, opened 2023-01-17
Add support for stopping / starting sequencer from admin API

mdehoog:
**Description**

Adds `admin_stopSequencer` and `admin_startSequencer` commands to `op-node` for stopping and starting the sequencer.

Also adds a `--sequencer.stopped` flag that allows starting the sequencer node in a "stopped" state (not producing blocks).

**Tests**

Added an E2E test.


op-node/rollup/driver/state.go
```diff
@@ -367,6 +375,15 @@ func (s *Driver) eventLoop() {
 			s.derivation.Reset()
 			s.metrics.RecordPipelineReset()
 			close(respCh)
+		case respCh := <-s.startSequencer:
+			s.log.Info("Sequencer has been started")
+			s.driverConfig.SequencerEnabled = true
+			sequencingPlannedOnto = eth.BlockID{}
```
mdehoog, line 381: without this, when restarting the sequencer I was running into this error:
https://github.com/ethereum-optimism/optimism/blob/d8e328ae936c6a5f3987c04cbde7bd94403a96a0/op-node/rollup/derive/batch_queue.go#L160

protolambda: See https://github.com/ethereum-optimism/optimism/pull/4367

This sequencer state should get cleaner and better separated with that, making it easier to handle here.

tynes:
I definitely support this addition, good thinking. Do we want to error on the case that the node isn't explicitly started as a sequencer and then `admin_startSequencer` is called? I lean towards yes because running 2 sequencers at the same time could result in an inconsistent network depending on how the infra is set up

op-node/rollup/driver/state.go
tynes: Are you planning on prioritizing #4367 @protolambda? Can this PR get in before or after?

protolambda:
@tynes the start/stop enables more seamless sequencer rotation, but agreed that simplicity and errors should be prioritized too, since sequencing conflicting L2 blocks is more dangerous than just being offline as sequencer for a while while handing off to a new sequencer (the gap of blocks can be filled).

One option might be to add a block-number argument to the start/stop, so that the old and new sequencers know exactly when they can change their state without conflicts. Or even better; output a block-hash in the stop function, and then pass that argument as the start argument, so the new sequencer can check that where it continues is fully consistent with where the other left off.

tynes:
> One option might be to add a block-number argument to the start/stop, so that the old and new sequencers know exactly when they can change their state without conflicts

In the long run, we want to be able to run multiple nodes (different entities) in sequencer mode but only have them produce blocks when they are the leader.

As part of practical infra setup, we want to be able to run multiple nodes that are virtually all the same sequencer and the infra can fallback to another node easily in case the current leader fails

tynes:
My preference is to error when calling the RPC saying to start sequencing when the CLI arg for being a sequencer was not passed and for infra terms pass the sequencer CLI arg to each node and then call the stop RPC for each except for the leader and then have automation around when the leader dies, call the start command on a different node. Might need a shared mempool for the sequencer nodes for consistency when rolling over

protolambda:
We'd still need previous unsafe blocks to be properly synced between those rotating sequencer nodes. And to do so, we'd need the next sequencer to know when they're missing something from the previous sequencer (with a possible timeout, to not lock up if the old sequencer block is lost).

mdehoog:
> Or even better; output a block-hash in the stop function, and then pass that argument as the start argument, so the new sequencer can check that where it continues is fully consistent with where the other left off.

great idea, will implement this

> My preference is to error when calling the RPC saying to start sequencing when the CLI arg for being a sequencer was not passed and for infra terms pass the sequencer CLI arg to each node and then call the stop RPC for each except for the leader and then have automation around when the leader dies, call the start command on a different node. Might need a shared mempool for the sequencer nodes for consistency when rolling over

@tynes relying on a quick call to stop a "sequencer" follower that should not be sequencing seems a bit dangerous to me. If we want only nodes that have `--sequencer.enabled` to accept an `admin_startSequencer` RPC, we probably want a separate flag like `--sequencer.stopped` or something. For now, is allowing any node to accept `admin_startSequencer` good enough? These nodes have to have been configured with the correct key matter to sequence, so accidentally starting the sequencer on a validator node won't work anyway

tynes:
It feels not ideal to have multiple ways to configure a sequencer - the deployment file will no longer be the source of truth. I feel like if we allow configuration by both RPC and config, we would need some sort of RPC method that returns JSON `{"sequencing": true}` or something like that

> For now, is allowing any node to accept admin_startSequencer good enough?

For now I think this is ok, note that I came up with these thoughts on the spot and would love to jam more on how to best run the infra

>  These nodes have to have been configured with the correct key matter to sequence, so accidentally starting the sequencer on a validator node won't work anyway

The key is for unsafe block signing, but hypothetically the `op-batcher` could submit data from the "wrong" sequencer if 2 are running at the same time

mdehoog:
> It feels not ideal to have multiple ways to configure a sequencer - the deployment file will no longer be the source of truth.
Agreed, this is not ideal. I just pushed a commit that allows starting the sequencer in a stopped state using the `--sequencer.stopped` flag; would be interested in thoughts. Could also potentially use a three-state flag like `--sequencer.mode=[disabled|stopped|enabled]` or something, and deprecated `--sequencer.enabled`.

I also implemented returning errors if start/stop are called when already started/stopped, returning the block hash from the stop RPC, and requiring the block hash on the start RPC.

trianglesphere:
> I also implemented returning errors if start/stop are called when already started/stopped, returning the block hash from the stop RPC, and requiring the block hash on the start RPC.

This makes sense. I've implemented a version of this before. I did it though by having an instant + withBlock (by number, worked well with the consensus algo) versions. Having the block hash + handling it here is important to preserving the unsafe portion of the chain across sequencers.

> sequencer.mode=[disabled|stopped|enabled]

With the tri-state flag is the intention for `disabled` to be the default, then `stopped` would allow it to be turned on, but starts stopped?

trianglesphere:
FYI: We may want to cleanup how the admin API is started/ran. I added it pretty haphazardly a while back. It is the right API to add to. I can follow up and look into it.

protolambda:
> FYI: We may want to cleanup how the admin API is started/ran. I added it pretty haphazardly a while back. It is the right API to add to. I can follow up and look into it.

Maybe we should also reuse the engine-API authentication of geth for the admin API? It has a spec here: https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md

tynes:
> > FYI: We may want to cleanup how the admin API is started/ran. I added it pretty haphazardly a while back. It is the right API to add to. I can follow up and look into it.
> 
> Maybe we should also reuse the engine-API authentication of geth for the admin API? It has a spec here: https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md

Perhaps we should put together a spec for the `admin`  RPC namespace since this discussion is starting to expand outside of the scope of this implementation

op-node/rollup/driver/state.go
protolambda: This PR can get in first. There will be some tweaks to the driver loop to better separate the sequencing code, but the functionality will stay mostly the same. Can't guarantee that for future admin RPC changes though, there are more things left to do in that area.

protolambda:
@Mergifyio update

tynes (approved):

PR #4720, opened 2023-01-17
fix(op-node): Use a reader to parse `SystemConfig` update events

clabby:
# Overview

Updates `op-node`'s `ProcessSystemConfigUpdateLogEvent` function to use a `bytes.Reader` instead of directly indexing the event data.

*Metadata*
Fixes ENG-2975


clabby:
* develop
  * **PR #4720** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4720" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4720?utm_source=stack-comment).

clabby:
> Semgrep found **10** [`read-io-eof`](https://semgrep.dev/r/dgryski.semgrep-go.readeof.read-io-eof) findings:
> 
> * op-node/rollup/derive/system_config.go: [L86-88](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L86-88), [L94-96](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L94-96), [L102-104](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L102-104), [L118-120](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L118-120), [L126-128](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L126-128), [L134-136](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L134-136), [L156-158](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L156-158), [L164-166](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L164-166), [L172-174](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L172-174), [L142-144](https://github.com/ethereum-optimism/optimism/blob/ec73572db8029c28693cbaa83d01512f493013e2/op-node/rollup/derive/system_config.go#L142-144)
> 
> Read() can return n bytes and io.EOF
> 
> Created by [read-io-eof](https://semgrep.dev/r/dgryski.semgrep-go.readeof.read-io-eof?utm_campaign=finding_notification&utm_medium=review_comment&utm_source=github&utm_content=rule).

All errors that could be presented by `Read` are handled- can this be ignored? @protolambda @trianglesphere

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +69,115 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
 	}
 	// indexed 1
 	updateType := ev.Topics[2]
-	// unindexed data
+
+	// Create a reader of the unindexed data
+	reader := bytes.NewReader(ev.Data)
+	// Allocate a placeholder word to read into
+	word := make([]byte, 32)
```
protolambda, line 76: Maybe it's cleaner to define an inline function that reads a word as 32 byte array, instead of repeating the same snippet of reading and converting it?
And the length checks of data should prevent reading errors already, so the inline function can maybe omit the error (although still nice if we can type check or detect programming errors there somehow)

clabby: Good suggestion on the inline func. Updated- let me know if this was what you were looking for here.

op-node/rollup/derive/system_config.go
```diff
@@ -75,33 +78,41 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
 		if len(ev.Data) != 32*3 {
 			return fmt.Errorf("expected 32*3 bytes in batcher hash update, but got %d bytes", len(ev.Data))
 		}
 
-		// Attempt to read the pointer, it should always equal 32.
-		if _, err := reader.Read(word); err != nil {
-			return fmt.Errorf("failed to read pointer in batcher address update")
-		}
+		// Read the pointer, it should always equal 32.
+		readIntoWord(word)
```
protolambda, line 103: Nit: maybe `word := readWord()` works better, to avoid repeatedly using the same word? Can do `if word := readWord(); word == ...`  inline to avoid definitions colliding.

clabby: Sure! I like how this looks more. I originally did it this way so that go wouldn't have to allocate new memory each time a word was read, but still getting familiar with the quirks of the compiler. Would this involve re-allocations?

protolambda: If it returns it should just allocate on the stack if you use small fixed-size arrays and local variables only.  That said, I think readability is more important than performance in this codepath.

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +72,88 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
 	updateType := ev.Topics[2]
-	// unindexed data
+
+	// Create a reader of the unindexed data
+	reader := bytes.NewReader(ev.Data)
+
+	// Helper function to prevent code duplication.
+	readWord := func() []byte {
+		// Allocate a new slice to return
+		b := make([]byte, 32)
```
protolambda, line 82: Would use a `[32]byte` fixed-length return value here as mentioned earlier instead of a slice, but I'll stop nitting, this looks good.

protolambda (approved):

op-node/rollup/derive/system_config.go
protolambda: Regarding Go: the slice->array trick works and is newer, but the most minimal way is probably to define the return value as named array, and read straight into it

protolambda: And in Go it's fine to return by value instead of by pointer reference, especially for small flat values like a 32 byte array. It should stay on the stack that way too, as it can't really escape if not referenced by pointer.

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +72,88 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
 		return nil
 	case SystemConfigUpdateGasLimit:
 		if len(ev.Data) != 32*3 {
 			return fmt.Errorf("expected 32*3 bytes in gas limit update, but got %d bytes", len(ev.Data))
 		}
-		if x := common.BytesToHash(ev.Data[:32]); x != (common.Hash{31: 32}) {
-			return fmt.Errorf("expected offset to point to length location, but got %s", x)
+
+		// Read the pointer, it should always equal 32.
+		if word := readWord(); common.BytesToHash(word[:]) != (common.Hash{31: 32}) {
```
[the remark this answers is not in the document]
clabby: Swapped over to this- thanks!

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +72,86 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
 	switch updateType {
 	case SystemConfigUpdateBatcher:
 		if len(ev.Data) != 32*3 {
 			return fmt.Errorf("expected 32*3 bytes in batcher hash update, but got %d bytes", len(ev.Data))
 		}
-		if x := common.BytesToHash(ev.Data[:32]); x != (common.Hash{31: 32}) {
-			return fmt.Errorf("expected offset to point to length location, but got %s", x)
+
+		// Read the pointer, it should always equal 32.
+		if word := readWord(); common.BytesToHash(word) != (common.Hash{31: 32}) {
```
tynes, line 97: I prefer to use a variable defined at the top of the file for `common.Hash{31: 32}`

op-node/rollup/derive/system_config.go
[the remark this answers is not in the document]
protolambda: You can read into the [32]byte return value by slicing it

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +72,86 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
+
+	// Create a reader of the unindexed data
+	reader := bytes.NewReader(ev.Data)
+
+	// Helper function to prevent code duplication.
+	readWord := func() (b []byte) {
+		if _, err := reader.Read(b); err != nil {
+			// The possible error returned by `Read` is ignored due to the length check of the unindexed data in
+			// all cases of the below switch statement. While we don't panic here, this log should *never* be emitted.
+			logger.Crit("failed to read word from unindexed log data")
```
tynes, line 84: I don't love crashing on malicious input. The L1 multisig secures the abi of the data coming in here, really bad things can happen in the L1 multisig is compromised. Ideally a compromised multisig cannot crash nodes on the network as well as not steal any money from the bridge

tynes: The previous impl did crash on malicious input as well, would it be difficult to handle malicious input at part of this PR

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +72,86 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
+		if word := readWord(); common.BytesToHash(word) != (common.Hash{31: 32}) {
+			return fmt.Errorf("expected length to be 32 bytes, but got %s", word)
 		}
-		if !bytes.Equal(ev.Data[64:64+24], make([]byte, 24)) {
-			return fmt.Errorf("expected zero padding for gaslimit, but got %x", ev.Data)
+
+		// Indexing `word` directly is always safe here, it is guaranteed to be 32 bytes in length.
+		// Check that the gas limit is correctly zero-padded.
+		word := readWord()
+		if !bytes.Equal(word[:24], make([]byte, 24)) {
```
tynes, line 151: Define the 24 0 bytes in a `var` at top of file?

op-node/rollup/derive/system_config.go
protolambda: We have a design doc about safety vs liveness w.r.t. L1 receipts that we read. If one of these contracts is off, something is already severely wrong, and we should likely halt.

op-node/rollup/derive/system_config.go
```diff
@@ -27,6 +28,14 @@ var (
 	ConfigUpdateEventVersion0 = common.Hash{}
 )
 
+// A left-padded uint256 equal to 32.
```
tynes, line 31: slight nit: 

```go
var (
    ...
    ...
)
```

op-node/rollup/derive/system_config.go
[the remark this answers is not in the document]
clabby: Done- appreciate it

op-node/rollup/derive/system_config.go
tynes: `logger.Crit` calls `panic` right? We could halt without crashing the software? Sounds like that is too large of a change for this PR then?

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +81,86 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
 	switch updateType {
 	case SystemConfigUpdateBatcher:
 		if len(ev.Data) != 32*3 {
 			return fmt.Errorf("expected 32*3 bytes in batcher hash update, but got %d bytes", len(ev.Data))
 		}
-		if x := common.BytesToHash(ev.Data[:32]); x != (common.Hash{31: 32}) {
-			return fmt.Errorf("expected offset to point to length location, but got %s", x)
+
+		// Read the pointer, it should always equal 32.
+		if word := readWord(); common.BytesToHash(word[:]) != OneWordUint {
```
protolambda, line 106:
```suggestion
		if word := readWord(); word != OneWordUint {
```

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +81,86 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
+			return fmt.Errorf("expected offset to point to length location, but got %s", word)
 		}
-		if x := common.BytesToHash(ev.Data[32:64]); x != (common.Hash{31: 32}) {
-			return fmt.Errorf("expected length of 1 bytes32, but got %s", x)
+
+		// Read the length, it should also always equal 32.
+		if word := readWord(); common.BytesToHash(word[:]) != OneWordUint {
```
protolambda, line 111:
```suggestion
		if word := readWord(); word != OneWordUint {
```

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +81,86 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
-	case SystemConfigUpdateGasConfig: // left padded uint8
+	case SystemConfigUpdateGasConfig:
 		if len(ev.Data) != 32*4 {
 			return fmt.Errorf("expected 32*4 bytes in GPO params update data, but got %d", len(ev.Data))
 		}
-		if x := common.BytesToHash(ev.Data[:32]); x != (common.Hash{31: 32}) {
-			return fmt.Errorf("expected offset to point to length location, but got %s", x)
+
+		// Read the pointer, it should always equal 32.
+		if word := readWord(); common.BytesToHash(word[:]) != OneWordUint {
```
protolambda, line 129:
```suggestion
		if word := readWord(); word != OneWordUint {
```

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +81,86 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
+			return fmt.Errorf("expected offset to point to length location, but got %s", word)
 		}
-		if x := common.BytesToHash(ev.Data[32:64]); x != (common.Hash{31: 64}) {
-			return fmt.Errorf("expected length of 2 bytes32, but got %s", x)
+
+		// Read the length, it should always equal 64.
+		if word := readWord(); common.BytesToHash(word[:]) != OneWordUint {
```
protolambda, line 134:
```suggestion
		if word := readWord(); word != OneWordUint {
```

Copy-paste bug of `OneWordUint`? It was checking a 64 value before. cc @tynes are you really sure this is more readable? I liked the previous value checks, it's shorter too.

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +81,86 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
 		return nil
 	case SystemConfigUpdateGasLimit:
 		if len(ev.Data) != 32*3 {
 			return fmt.Errorf("expected 32*3 bytes in gas limit update, but got %d bytes", len(ev.Data))
 		}
-		if x := common.BytesToHash(ev.Data[:32]); x != (common.Hash{31: 32}) {
-			return fmt.Errorf("expected offset to point to length location, but got %s", x)
+
+		// Read the pointer, it should always equal 32.
+		if word := readWord(); common.BytesToHash(word[:]) != OneWordUint {
```
protolambda, line 148:
```suggestion
		if word := readWord(); word != OneWordUint {
```

op-node/rollup/derive/system_config.go
clabby: Ah yes- good spot

op-node/rollup/derive/system_config.go
protolambda: The derivation pipeline has a `NewCriticalError(...)` typed error you can return to halt without crashing. But it's a bit difficult because of the function not returning it.

Ideally we check the input length once, and then do the reading safely without unnecessary and verbose error handling.

clabby: Due to the `readWord` reading 32 bytes every time + the length checks in each of the switch cases, this log should be impossible to hit regardless of a malicious or honest input, no? I mainly put it there to protect against any changes in this codepath that don't assert adequate length before reading.

protolambda: Possible solution: track how much we've read from the input in the helper func, return zeroed-data if the reader goes out of bounds, and after reading everything check if we have consumed everything. So then there's only a single error to check

clabby: I'm a fan of this route. @tynes ?

tynes: This sounds good to me! Good thinking @protolambda

tynes: I think a way to do this would be implementing a custom type like `WordReader`, this would let us get rid of the closure and would potentially be useful elsewhere. It could have public methods `Check` for the assertion as well as implement `io.Reader` and only read out `[32]byte` as slices to fit the interface, could also add `ReadHash() common.Hash` for simplicity

op-node/rollup/derive/system_config.go
```diff
@@ -69,50 +83,94 @@ func ProcessSystemConfigUpdateLogEvent(destSysCfg *eth.SystemConfig, ev *types.L
...
+	countReadBytes := 0
+
+	// Helper function to read a word from the log data reader
+	readWord := func() (b [32]byte) {
+		if _, err := reader.Read(b[:]); err != nil {
+			// If there is an error reading the next 32 bytes from the reader, return an empty
+			// 32 byte array. We always check that the number of bytes read (`countReadBytes`)
+			// is equal to the expected amount at the end of each switch case.
+			return b
+		}
```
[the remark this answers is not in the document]
protolambda: bad semgrep bot, this is a bytes reader

op-node/rollup/derive/system_config.go
clabby: Made the changes- sorry for the delay, had to step away for a sec. Should we add a test that asserts `ProcessSystemConfigUpdateLogEvent` properly errors out when given an unexpected # of bytes in the log?

op-node/rollup/derive/system_config.go
```diff
@@ -27,6 +28,19 @@ var (
...
 
+var (
+	// A left-padded uint256 equal to 32.
+	OneWordUint = common.Hash{31: 32}
+	// A left-padded uint256 equal to 64.
+	TwoWordUint = common.Hash{31: 64}
+	// 24 zero bytes (the padding for a uint64 in a 32 byte word)
+	Uint64Padding = make([]byte, 24)
+	// 12 zero bytes (the padding for an Ethereum address in a 32 byte word)
+	AddressPadding = make([]byte, 12)
```
protolambda, line 39: Can consider making these lowercase / package-level to not make them part of the derivation package API.

protolambda (approved):
LGTM, nice work

op-node/rollup/derive/system_config.go
clabby: > Made the changes- sorry for the delay, had to step away for a sec. Should we add a test that asserts `ProcessSystemConfigUpdateLogEvent` properly errors out when given an unexpected # of bytes in the log?

RFC @protolambda (sorry, know you've already approved) ^

There are tests that assert this function's correctness, but given how critical it is, would it be a good idea to add a negative case here? Can also put it up in another PR if that'd be better.

protolambda: These look like good changes to me, and it's not less critical/complex than previously, so I'm fine to merge it. If you see opportunity to implement the test then it's welcome ofc.

protolambda: Mergify will pick it up if you mark these comments as resolved.

PR #4721, opened 2023-01-17
fix(dtl): sync beyond deposit shutoff block

smartcontracts:
**Description**
Fixes a bug in the DTL that caused it to not be able to sync beyond the deposit shutoff block. This would happen because the TO block would be less than the FROM block which would cause certain providers to throw an error.

### DIFF
--- a/.changeset/unlucky-rats-sit.md
+++ b/.changeset/unlucky-rats-sit.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/data-transport-layer': patch
+---
+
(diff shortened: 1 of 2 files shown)


smartcontracts:
I have confirmed that this syncs Goerli to the final block as expected

tynes (approved):

koraykoska:
Why was this not deployed yet? Is there a workaround I can use until the docker image is available?

koraykoska:
@smartcontracts

PR #4539, opened 2022-12-19
op-chain-ops: update script to submit pending withdrawals

tynes:
**Description**

Replaces #4230

Creates a script to send migrated pending withdrawals. It requires the witness scripts generated by `l2geth`.
I didn't find any withdrawals that were broken during the migration.

Closes ENG-2740
Closes ENG-3151

PR #4730, opened 2023-01-17
op-chain-ops: `Withdrawal` type use `hexutil.Bytes`

tynes:
**Description**

Use `hexutil.Bytes` in the `Withdrawal` type so that when writing and reading from JSON, it correctly serializes the data as a 0x prefixed hex string. Without this, it will some base64 looking serialization. Be sure to cast to a byte slice before doing any abi packing, otherwise it will cause errors.

Part of breaking https://github.com/ethereum-optimism/optimism/pull/4539 into smaller pieces


mslipper:
Does this require changes in the witness generation code too?

tynes:
> Does this require changes in the witness generation code too?

No it does not, the witness generation code is in typescript

trianglesphere (approved):
lgtm

PR #4731, opened 2023-01-17
op-chain-ops: add getter for cdm

tynes:
**Description**

Adds a getter to the `LegacyWithdrawal` that converts it into a `CrossDomainMessage`. This simplifies some code when doing migrated legacy withdrawals.

Part of migrating https://github.com/ethereum-optimism/optimism/pull/4539 into smaller pieces


op-chain-ops/crossdomain/legacy_withdrawal.go
```diff
@@ -170,3 +170,16 @@ func (w *LegacyWithdrawal) Value() (*big.Int, error) {
 
 	return value, nil
 }
+
+// CrossDomainMessage turns the LegacyWithdrawal into
+// a CrossDomainMessage
```
protolambda, line 175: Maybe worth a comment about it having a zero gaslimit? Is that even valid?

tynes: I've added a comment here, basically there is no concept of a gas limit with the legacy withdrawals, its also technically a "min gas limit" because the gas limit set on the tx overrides

protolambda (approved):

PR #4736, opened 2023-01-18
Fix `channelTimeout` docs

sebastianst:
Channel timeouts used to be given in seconds, now they're given in number of L1 blocks. This PR fixes two places where the old description was still present.

### DIFF
--- a/packages/contracts-bedrock/src/deploy-config.ts
+++ b/packages/contracts-bedrock/src/deploy-config.ts
@@ -56,7 +56,7 @@ interface RequiredDeployConfig {
   sequencerWindowSize: number
 
   /**
-   * Number of seconds (w.r.t. L1 time) that a frame can be valid when included in L1.
+   * Number of L1 blocks that a frame stays valid when included in L1.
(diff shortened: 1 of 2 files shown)


packages/contracts-bedrock/src/deploy-config.ts
```diff
@@ -56,7 +56,7 @@ interface RequiredDeployConfig {
   sequencerWindowSize: number
 
   /**
-   * Number of seconds (w.r.t. L1 time) that a frame can be valid when included in L1.
+   * Number of L2 blocks that a frame stays valid when included in L1.
```
protolambda, line 59: Shouldn't this be L1 blocks?

sebastianst: oops yes

maurelian:
You might need to login to circleci using your github account in order to get past the SAML error?

protolambda:
@maurelian we tried that

tynes (approved):

PR #4737, opened 2023-01-18
op-node: attributes builder interface

protolambda:
**Description**

Instead of building the attributes with a global function that cannot be substituted, build the attributes by calling an interface method. We can separate the attributes that don't change, and make sequencing easier to test: it's a lot easier to mock attributes than the exact inputs and sources for the creation of the attributes.

And remove the L2 timestamp argument, it can already easily be determined from the L2 parent reference and blocktime.

Also adds a sanity check to the attributes building, to make sure the L1 time is consistent. This can then stay separate from the sequencer codepath.

This is similar to the separation of the calldata source and the L1 retrieval stage: one implements the exact functionality and may be swapped, the other implements the stage behavior.

This PR helps unblock #4367 

**Tests**

No functionality changes, updated tests to use the new interface.


trianglesphere (approved):
lgtm

PR #4740, opened 2023-01-19
op-node: fix pprof missing handler, use op-service pprof util instead

protolambda:
**Description**

@refcell found that the pprof server wasn't properly set up: the pprof handlers were not registered with the http server we created.

We have  a pprof server util in the `op-service`, which does register the handlers properly, so I changed it to use that instead.


op-node/service.go
```diff
@@ -75,10 +76,10 @@ func NewConfig(ctx *cli.Context, log log.Logger) (*node.Config, error) {
 			ListenAddr: ctx.GlobalString(flags.MetricsAddrFlag.Name),
 			ListenPort: ctx.GlobalInt(flags.MetricsPortFlag.Name),
 		},
-		Pprof: node.PprofConfig{
+		Pprof: oppprof.CLIConfig{
 			Enabled:    ctx.GlobalBool(flags.PprofEnabledFlag.Name),
 			ListenAddr: ctx.GlobalString(flags.PprofAddrFlag.Name),
-			ListenPort: ctx.GlobalString(flags.PprofPortFlag.Name),
+			ListenPort: ctx.GlobalInt(flags.PprofPortFlag.Name),
```
protolambda, line 82: It's defined as intflag, previous code was wrong

op-node/cmd/main.go
```diff
@@ -3,16 +3,17 @@ package main
...
 	"os/signal"
+	"strconv"
 	"syscall"
-	"time"
 
 	"github.com/ethereum-optimism/optimism/op-node/cmd/doc"
 
 	"github.com/urfave/cli"
 
+	"github.com/ethereum/go-ethereum/log"
```
trianglesphere, line 15: nit: weird imports (tho also some were old).

trianglesphere (approved):
lgtm.

PR #4743, opened 2023-01-19
feat(op-node): Peer Score Hardening

refcell:
**Description**

Closes CLI-3524
Closes CLI-3245
Closes CLI-3249
Closes CLI-3247

**Changelog**

- [x] Adds peer scoring to the `op-node`.
- [x] Peer scoring metrics.
- [x] Peer scoring CLI Flags


op-node/p2p/gossip.go
```diff
@@ -160,10 +165,21 @@ func NewGossipSub(p2pCtx context.Context, h host.Host, cfg *rollup.Config, gossi
...
 		pubsub.WithEventTracer(&gossipTracer{m: m}),
+		pubsub.WithPeerScoreInspect(BuildPeerScoreInspector(m), peerScoreInspectFrequency),
 	}
 	gossipOpts = append(gossipOpts, gossipConf.ConfigureGossip(&params)...)
 	return pubsub.NewGossipSub(p2pCtx, h, gossipOpts...)
-	// TODO: pubsub.WithPeerScoreInspect(inspect, InspectInterval) to update peerstore scores with gossip scores
+}
+
+// BuildPeerScoreInspector returns a function that is called periodically by the pubsub library to inspect the peer scores.
+// It is passed into the pubsub library as a [pubsub.ExtendedPeerScoreInspectFn] in the [pubsub.WithPeerScoreInspect] option.
```
tynes, line 175: When you put a type in brackets, does it create a hyperlink in the godoc?

tynes: great comment btw

tynes:
Not opposed to the comments being broken out into their own PR

op-node/p2p/gossip.go
refcell: It should! See: https://tip.golang.org/doc/comment
Screenshot 2023-01-19 at 8 31 12 PM

refcell:
> Not opposed to the comments being broken out into their own PR

Will do going forward - keeping for now as i'm still learning.

op-node/p2p/gossip.go
```diff
@@ -140,12 +143,21 @@ func BuildGlobalGossipParams(cfg *rollup.Config) pubsub.GossipSubParams {
...
+// NewGossipSub configures a new pubsub instance with the specified parameters.
+// PubSub uses a GossipSubRouter as it's router under the hood.
+func NewGossipSub(p2pCtx context.Context, h host.Host, g ConnectionGater, cfg *rollup.Config, gossipConf GossipSetupConfigurables, m GossipMetricer) (*pubsub.PubSub, error) {
 	denyList, err := pubsub.NewTimeCachedBlacklist(30 * time.Second)
 	if err != nil {
 		return nil, err
 	}
 	params := BuildGlobalGossipParams(cfg)
+	// TODO: make this configurable behind a cli flag - disabled or default
+	peerScoreParams, err := GetPeerScoreParams("default")
```
protolambda, line 155: Some score params will probably depend on rollup config properties like the block time. Would make sense to add the config as argument.

op-node/metrics/metrics.go
```diff
@@ -278,6 +284,15 @@ func NewMetrics(procName string) *Metrics {
...
 			Help:      "Count of currently connected p2p streams",
 		}),
+		PeerScores: factory.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: ns,
+			Subsystem: "p2p",
+			Name:      "peer_scores",
+			Buckets:   []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10},
+			Help:      "Histogram of p2p peer scores",
+		}, []string{
+			"method",
```
protolambda, line 294:
```suggestion
			"peerid",
```

op-node/p2p/gossip.go
```diff
@@ -424,11 +437,18 @@ func JoinGossip(p2pCtx context.Context, self peer.ID, ps *pubsub.PubSub, log log
 	}
 	go LogTopicEvents(p2pCtx, log.New("topic", "blocks"), blocksTopicEvents)
 
-	// TODO: block topic scoring parameters
+	// TODO: make this configurable behind a cli flag - disabled or default
+	// Set default block topic scoring parameters
```
protolambda, line 441: I'd recommend to name flags based on what they do, rather than on what's default. Then the CLI flag definition is the only place that changes the default behavior.

op-node/p2p/peer_params.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+// TODO: Update these parameters.
+const (
+	slot             = 2 * time.Second
+	epoch            = 12 * time.Second
+	tenEpochs        = 10 * epoch
+	oneHundredEpochs = 100 * epoch
+	decayToZero      = 0.01
+)
```
protolambda, line 19: I think the slot/epoch naming is confusing when we don't use it like L1 and when they have different durations and properties.

Can we instead just use L2 blocks as parameter reference?
And I'm afraid we can't make these all constant; if it really depends on the block in terms of duration, then we need to scale it with the rollup config l2 blocktime parameter.

trianglesphere (commented):
Really nice doc additions in the heartbeat & sources packages, but could you pull them out of this PR into a separate PR. We try to keep PRs as a single unit rather than pulling in a bunch of disparate changes.

op-node/p2p/peer_params.go
```diff
@@ -0,0 +1,114 @@
...
+func NewPeerScoreThresholds() pubsub.PeerScoreThresholds {
+	return pubsub.PeerScoreThresholds{
+		SkipAtomicValidation:        false,
+		GossipThreshold:             -10,
+		PublishThreshold:            -40,
+		GraylistThreshold:           -40,
+		AcceptPXThreshold:           20,
+		OpportunisticGraftThreshold: 0.05,
+	}
+}
```
protolambda, line 114: We should maybe define the scoring scale somewhere before we set thresholds like L1

op-node/p2p/peer_scorer.go
```diff
@@ -0,0 +1,94 @@
...
+	// e.g., score = score - (s.peerConnections[peerID] * ConnectionFactor)
+	// s.connGater.BanAddr(peerID)
+
+	return func(m map[peer.ID]*pubsub.PeerScoreSnapshot) {
+		for id, snap := range m {
+			// Record peer score in the metricer
+			s.metricer.RecordPeerScoring(id, snap.Score)
+
+			// TODO: encorporate the number of peer connections/disconnections into the score
+			// TODO: or should we just affect the score in the OnConnect/OnDisconnect methods?
```
protolambda, line 52: our own peercount shouldn't influence the scores of any of our peers directly

op-node/p2p/peer_scorer.go
```diff
@@ -0,0 +1,94 @@
...
+			// TODO: if we don't have to do this calculation here, we can push score updates to the metricer
+			// TODO: which would leave the scoring to the pubsub lib
+			// peer, err := s.peerStore.Get(id)
+			// if err != nil {
+			// }
+
+			// Check if the peer score is below the threshold
+			// If so, we need to block the peer
+			if snap.Score < PeerScoreThreshold {
+				_ = s.connGater.BlockPeer(id)
```
protolambda, line 62: why mute the error? Can we log it at least?

op-node/p2p/peer_scorer.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+			// Unblock peers whose score has recovered to an acceptable level
+			if (snap.Score > PeerScoreThreshold) && contains(s.connGater.ListBlockedPeers(), id) {
+				_ = s.connGater.UnblockPeer(id)
+			}
+		}
+	}
+}
+
+func contains(peers []peer.ID, id peer.ID) bool {
```
tynes, line 72: This seems like this could be a generic builtin but not sure if `peer.ID` implements `comparable`

tynes:
Going to defer most review here to @protolambda and @trianglesphere as I'm not super familiar with the internals of libp2p and have other things that I need to prioritize

Good job with the documented code @refcell

op-node/p2p/gossip.go
```diff
@@ -39,6 +39,8 @@ const (
 	DefaultMeshDlo   = 6  // topic stable mesh low watermark
 	DefaultMeshDhi   = 12 // topic stable mesh high watermark
 	DefaultMeshDlazy = 6  // gossip target
+	// peerScoreInspectFrequency is the frequency at which peer scores are inspected
+	peerScoreInspectFrequency = 1 * time.Minute
```
ajsutton, line 43: If we only review peer scores once a minute we potentially allow a peer to cause quite a lot of issues before we actually disconnect them. https://linear.app/optimism/issue/CLI-3342/sherlock-035-memory-amplification-with-small-but-invalid-p2p-messages for example would remain an issue since the seen message TTL is about a minute anyway so they still have a full minute of sending garbage before being disconnected.

That said, evaluating peer scores is potentially expensive (it was in the java libp2p, not sure about the go version) so we may need to benchmark this when there are a significant number of known peers before making it too short.

op-node/p2p/gossip.go
```diff
@@ -142,12 +145,19 @@ func BuildGlobalGossipParams(cfg *rollup.Config) pubsub.GossipSubParams {
...
 // PubSub uses a GossipSubRouter as it's router under the hood.
-func NewGossipSub(p2pCtx context.Context, h host.Host, cfg *rollup.Config, gossipConf GossipSetupConfigurables, m GossipMetricer) (*pubsub.PubSub, error) {
+func NewGossipSub(p2pCtx context.Context, h host.Host, g ConnectionGater, cfg *rollup.Config, gossipConf GossipSetupConfigurables, m GossipMetricer, log log.Logger) (*pubsub.PubSub, error) {
 	denyList, err := pubsub.NewTimeCachedBlacklist(30 * time.Second)
 	if err != nil {
 		return nil, err
 	}
 	params := BuildGlobalGossipParams(cfg)
+	// TODO: make this configurable behind a cli flag - disabled or default
+	peerScoreParams, err := GetPeerScoreParams("default", cfg)
```
ajsutton, line 155: Feel free to ignore this, but if peer scoring is disabled, it may be worth not even creating any of the scoring infrastructure. ie when disabled, don't create any score params or thresholds, don't create the scorer and don't add the `pubsub.WithPeerScore` and `pubsub.WithPeerScoreInspect` options at all.

That way disabling peer scoring acts as a feature toggle and if some problem is encountered with peer scoring, it can be completely disabled. Whereas if we just disable it by using a different set of params, any bugs in the scorer would still cause issues and potentially the "disabled" params may wind up not as disabled as we intended etc. Also saves us from needing to test those disabled params.

refcell: Great suggestion, I think for now I'll leave it to be configurable so we can experiment in a devnet environment to gain better intuition for the parameters we need.

op-node/metrics/metrics.go
```diff
@@ -289,6 +293,15 @@ func NewMetrics(procName string) *Metrics {
 			Name:      "stream_count",
 			Help:      "Count of currently connected p2p streams",
 		}),
+		PeerScores: factory.NewHistogramVec(prometheus.HistogramOpts{
```
protolambda, line 296: GaugeVec would work better here

op-node/p2p/peer_scorer_test.go
```diff
@@ -0,0 +1,120 @@
...
+	inspectFn := scorer.SnapshotHook()
+
+	// Mock the snapshot updates
+	// This doesn't return anything
+	testSuite.mockMetricer.On("RecordPeerScoring", peer.ID("peer1"), float64(-101)).Return(nil)
+
+	// Mock a connection gater peer block call
+	// Since the peer score is below the [PeerScoreThreshold] of -100,
+	// the [BlockPeer] method should be called
+	testSuite.mockGater.On("BlockPeer", peer.ID("peer1")).Return(nil)
```
protolambda, line 111: needs something like `testSuite.mockGater.AssertExpectations(t)` to complete mock iirc

op-node/p2p/mocks/ConnectionGater.go
```diff
@@ -0,0 +1,245 @@
...
+	mock.TestingT
+	Cleanup(func())
+}
+
+// NewConnectionGater creates a new instance of ConnectionGater. It also registers a testing interface on the mock and a cleanup function to assert the mocks expectations.
+func NewConnectionGater(t mockConstructorTestingTNewConnectionGater) *ConnectionGater {
+	mock := &ConnectionGater{}
+	mock.Mock.Test(t)
+
+	t.Cleanup(func() { mock.AssertExpectations(t) })
```
protolambda, line 242: nvm the other comment

op-node/p2p/peer_gater.go
```diff
@@ -0,0 +1,54 @@
...
+// Update handles a peer score update and blocks/unblocks the peer if necessary.
+func (s *gater) Update(id peer.ID, score float64) {
+	// Check if the peer score is below the threshold
+	// If so, we need to block the peer
+	if score < PeerScoreThreshold && s.banEnabled {
+		s.log.Warn("peer blocking enabled, blocking peer", "id", id.String(), "score", score)
+		err := s.connGater.BlockPeer(id)
+		s.log.Warn("connection gater failed to block peer", id.String(), "err", err)
+	}
+	// TODO: This will never unblock a peer since the peer will disconnect and the score update hook will not trigger this call.
```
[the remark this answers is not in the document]
refcell: yessir semgrep app sir

op-node/metrics/metrics.go
```diff
@@ -289,6 +293,14 @@ func NewMetrics(procName string) *Metrics {
...
 			Help:      "Count of currently connected p2p streams",
 		}),
+		PeerScores: factory.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: ns,
+			Subsystem: "p2p",
+			Name:      "peer_scores",
+			Help:      "Peer scoring",
+		}, []string{
+			"peerid",
+		}),
```
ajsutton, line 303: I think this potentially opens up a DoS vector from my memory since a malicious actor could connect with thousands of different PeerIDs quite quickly, resulting in each one getting it's own metric (at least for prometheus each label value is effectively a new metric).

https://prometheus.io/docs/practices/naming/#labels has a warning about this.

op-node/p2p/peer_params.go
```diff
@@ -0,0 +1,128 @@
...
+		DecayToZero:            DecayToZero,
+		RetainScore:            oneHundredEpochs,
+	}
+}
+
+// PeerScoreParamsByName is a map of name to function that returns a [pubsub.PeerScoreParams] based on the provided [rollup.Config].
+var PeerScoreParamsByName = map[string](func(blockTime uint64) pubsub.PeerScoreParams){
+	"light": LightPeerScoreParams,
+	"none":  DisabledPeerScoreParams,
+}
```
ajsutton, line 92: The CLI help text said `full` was a valid option as well.

op-e2e/system_test.go
```diff
@@ -646,6 +649,132 @@ func TestSystemMockP2P(t *testing.T) {
 	require.Contains(t, received, receiptVerif.BlockHash)
 }
 
+// TestSystemMockPeerScoring sets up a L1 Geth node, a rollup node, and a L2 geth node and then confirms that
+// the nodes can sync L2 blocks before they are confirmed on L1.
+func TestSystemMockPeerScoring(t *testing.T) {
```
ajsutton, line 654: Ideally we'd also have a test where one node is sending invalid messages and check it gets disconnected.  Might be reasonable to do by having one of the nodes running a sequencer with the wrong key since all the published blocks would then be considered invalid gossip.

ajsutton (commented):
I don't know the implementation details of go libp2p enough to review a lot of this, but left a few comments for the parts I do understand. :)

op-e2e/system_test.go
refcell: Nice! I can add that!

protolambda: +1, can also make one of the receiving verifier nodes here bad, and test that the good verifier nodes continue to receive payloads, while the bad one should stop getting them & be disconnected.

protolambda (approved):
+1 on the invalid-messages testing note by Adrian. We can do so in a separate follow-up PR though. This one is consistent and complete to merge, and already quite large. If you do mark as resolved, do open an issue for the ban testing so we can track it

PR #4744, opened 2023-01-19
fix(ctb): ERC20 bridge events

smartcontracts:
**Description**
Minimal fix for ERC20 bridge events not being emitted correctly when sending directly to bridge.


packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -164,9 +164,9 @@ abstract contract StandardBridge {
     }
 
     /**
-     * @notice Allows EOAs to deposit ETH by sending directly to the bridge.
+     * @notice Allows EOAs to bridge ETH by sending directly to the bridge.
```
tynes, line 167: The one question is whether or not we want to delete the implementation here and just leave the virtual function def

tynes: I lean towards yes here

tynes:
Replaces https://github.com/ethereum-optimism/optimism/pull/4708

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -16,7 +16,6 @@ contract L1StandardBridge_Getter_Test is Bridge_Initializer {
         assert(L1Bridge.OTHER_BRIDGE() == L2Bridge);
         assert(L1Bridge.messenger() == L1Messenger);
         assert(L1Bridge.MESSENGER() == L1Messenger);
-        assertEq(L1Bridge.version(), "1.0.0");
```
refcell, line 19: We should probably just change this to `1.1.0`.

```suggestion
        assertEq(L1Bridge.version(), "1.1.0");
```

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
refcell: +1, make this virtual

refcell (approved):
Minor change if I'm not mistaken.

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
tynes: I deleted this because none of the other contract tests test for the version

maurelian:
* develop
  * **PR #4744** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4744" target="_blank">Graphite</a>  👈
    * **PR #4746** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4746" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4744?utm_source=stack-comment).

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
maurelian: I'm not quite sure what's been agreed to here, but I feel strongly that we should not have an implementation here which is never used and gets overridden everywhere it's inherited. 

Let's just do this please:

`receive() external payable virtual;`

tynes: This is what we were agreeing on

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
tynes: I think we should follow up with a standard pattern for testing the semver

tynes (approved):

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
tynes: ok i added it back, but i still think we should come up with a unified plan for testing semver values

maurelian (approved):
This LGTM and moves us in the right direction. But it doesn't entirely address the issue of always having both the legacy and new events firing together. See the PR stacked on top of this one: #4746.

PR #4746, opened 2023-01-20
Test for both legacy and new events in bridges

maurelian:
Updates the L1 and L2 Standard bridge tests to ensure that both Legacy and Bedrock bridge events are emitted together. 

~~Tests are currently breaking.~~

OK, I was wrong, all these tests just needed some fixing up. Still they are good to have.

PR #4747, opened 2023-01-20
test(ctb): Add bridge tests for the new bridge interface

maurelian:
Adds tests which call the new bridge methods, ie. `bridgeETH`/`bridgeETHTo` and `bridgeERC20`/`bridgeERC20To`.

~~Tests fail because the new interface does not emit the legacy events.~~
I will get the tests working once the approach shown here is agreed to.

**Update:**

Now includes a proposed approach for adding the legacy events to the new interface. 

1. All events are moved into virtual wrapper functions in StandardBridge.sol. ie. 
https://github.com/ethereum-optimism/optimism/blob/4c5ab611844ae053b74f399d22a98a501506cba7/packages/contracts-bedrock/contracts/universal/StandardBridge.sol#L493-L500

2. Those functions are overridden in the child contracts to include the legacy events. ie. 
https://github.com/ethereum-optimism/optimism/blob/4c5ab611844ae053b74f399d22a98a501506cba7/packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol#L233-L243

3. Events are deleted from the child contracts and always emitted (using the wrapper functions) in the internal functions (which is best because that's where the action occurs. ie. 
https://github.com/ethereum-optimism/optimism/blob/4c5ab611844ae053b74f399d22a98a501506cba7/packages/contracts-bedrock/contracts/universal/StandardBridge.sol#L304-L307

This creates a very easy to read diff, and avoids moving the legacy events into the new interface itself.


maurelian:
* develop
  * **PR #4744** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4744" target="_blank">Graphite</a> 
    * **PR #4746** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4746" target="_blank">Graphite</a> 
      * **PR #4747** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4747" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4747?utm_source=stack-comment).

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -463,4 +463,83 @@ abstract contract StandardBridge {
...
+     * @param _amount    Amount of ETH sent.
+     * @param _extraData Extra data sent with the transaction.
+     */
+    function _emitETHBridgeInitiated(
+        address _from,
+        address _to,
+        uint256 _amount,
+        bytes memory _extraData
+    ) internal virtual {
+        emit ETHBridgeFinalized(_from, _to, _amount, _extraData);
```
clabby, line 481:
```suggestion
        emit ETHBridgeInitiated(_from, _to, _amount, _extraData);
```

packages/contracts-bedrock/contracts/L1/L1StandardBridge.sol
```diff
@@ -294,7 +291,62 @@ contract L1StandardBridge is StandardBridge, Semver {
...
+     * @inheritdoc StandardBridge
+     */
+    function _emitETHBridgeInitiated(
+        address _from,
+        address _to,
+        uint256 _amount,
+        bytes memory _extraData
+    ) internal override {
+        emit ETHDepositInitiated(_from, _to, _amount, _extraData);
+        emit ETHBridgeInitiated(_from, _to, _amount, _extraData);
```
clabby, line 307: Instead of overriding and re-declaring the event emissions that already exist in the super function, maybe just call `super()`?

clabby (commented):
This is definitely cleaner than the current method- I'm for it!

packages/contracts-bedrock/contracts/L1/L1StandardBridge.sol
maurelian: yeah... I thought of that, but then you have to go read another function to know what it does.
If `super()` was really complicated I might prefer that, but since it's just a matter of which single line of code or the other, I'd prefer to just make that other code irrelevant.

clabby: Got it- I see your reasoning there, makes sense 😄

tynes (commented):
I like this approach generally, would like it if additional comments are added in the child implementations noting the backwards compatibility

maurelian:
Interesting, looks like a legit failure in the devnet:
https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/13234/workflows/64870f5d-4c55-4e17-9fec-0cb2db3bb798/jobs/423145?invite=true#step-108-154

Will need to investigate.

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -151,7 +161,33 @@ contract L1StandardBridge_DepositETH_Test is Bridge_Initializer {
...
+
+contract L1StandardBridge_DepositETH_Test is PreBridgeETH {
+    // depositETH
+    // - emits ETHDepositInitiated
+    // - emits ETHBridgeInitiated
+    // - calls optimismPortal.depositTransaction
+    // - only EOA
+    // - ETH ends up in the optimismPortal
+    function test_depositETH_succeeds() external {
+        _preBridgeETH(true);
```
tynes, line 175: nit: could use kwargs syntax here to make it obvious that its legacy

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -301,6 +301,9 @@ abstract contract StandardBridge {
         require(_to != address(this), "StandardBridge: cannot send to self");
         require(_to != address(MESSENGER), "StandardBridge: cannot send to messenger");
 
+        // If the _emit function is overri theden it will emit legacy events before emitting the
```
tynes, line 304: nit: typo

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -301,6 +301,9 @@ abstract contract StandardBridge {
...
+        // current Bridge event.
+        _emitETHBridgeFinalized(_from, _to, _amount, _extraData);
         emit ETHBridgeFinalized(_from, _to, _amount, _extraData);
```
tynes, line 307: Hmm this reads kind of strange because its basically the same thing - should both events be inside the function in this case?

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -366,6 +372,9 @@ abstract contract StandardBridge {
             "StandardBridge: bridging ETH must include sufficient ETH value"
         );
 
+        // If the _emit function is overri theden it will emit legacy events before emitting the
+        // current Bridge event.
```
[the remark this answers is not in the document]
maurelian: I wanted to avoid using `super` because then you need to jump from the child contract to the parent to see what it does.

But I'll admit that this also looks weird. I can do that.

packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol
```diff
@@ -187,7 +185,79 @@ contract L2StandardBridge is StandardBridge, Semver {
         } else {
             _initiateBridgeERC20(_l2Token, l1Token, _from, _to, _amount, _minGasLimit, _extraData);
         }
+    }
+
```
tynes, line 189: my one nit here is that this comment doesn't follow the same standards as other comments that follow natspec

tynes (approved):

packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol
maurelian: Would it address you feedback if

maurelian: whoops, didn't mean to hit send on that. :P

I didn't realize that you could add docs on top of using `@inheritdoc`, good call it's much nicer now.

roninjin10 (approved):

PR #4749, opened 2023-01-20
op-chain-ops: Preserve predeploy nonces

mslipper:
Users are unable to use the OptimismMintableERC20Factory contract due to the nonce being reset to zero. This PR updates the predeploy storage wipe function to preserve nonces during the migration. This is fixable on Goerli by incrementing the nonce manually on the factory contract.


protolambda (commented):
Change looks good, but we should probably be testing this as part of the migration checks

op-chain-ops/genesis/setters.go
```diff
@@ -92,7 +92,13 @@ func WipePredeployStorage(db vm.StateDB) error {
 		}
 
 		log.Info("wiping storage", "name", name, "address", *addr)
+
+		// We need to make sure that we preserve nonces.
+		oldNonce := db.GetNonce(*addr)
 		db.CreateAccount(*addr)
```
tynes, line 98: We should also preserve balance? I don't think any of them should have balance but just in case

mslipper: Yeah good call.

protolambda (approved):
LGTM

PR #4755, opened 2023-01-21
op-node: implement EngineControl interface

protolambda:
**Description**

This introduces the `EngineState` and `EngineControl` interfaces, to read and interact with the engine.
The interfaces are minimal & created to avoid external engine state such as forkchoice-state arguments as much as possible.

In the future I would like to split the `EngineQueue` into two parts:
- The actual `EngineQueue` stage with just buffered safe attributes, unsafe payloads, and forkchoice update flag.
- The `Engine`, which will implement the engine-control interface, extended with the finality signal function as optional thing.
  Finality signals are otherwise unrelated to the derivation since the input is not stable, and by doing it this way we can keep the engine updated without conflicts.
  - This will also split the largest stage, which is great for testability and readability.

To keep this PR scope small, I just modified the engine queue for initial `EngineControl` support,
but in the next PRs of this stack I expand on it by adding optional engine metering and cleaning up the sequencer code path by using this.
By using a handle to an `EngineControl`, the sequener code does not have to maintain its own engine state anymore, and we avoid lots of potential inconsistencies,
as well as making the sequencer a lot more testable by not binding it straight to an Engine-API client.

Minor functional change: a payload result that fails to be converted into a block-ref is now handled with a reset instead of temporary error just in case, because the payload data is static so a temporary error doesn't make sense.

**Tests**

No notable functionality changes, same engine-queue codepath is already tested.

Full PR stack:
- #4755
- #4756
- #4757
- #4758


op-node/rollup/derive/engine_queue.go
```diff
@@ -457,21 +486,89 @@ func (eq *EngineQueue) forceNextSafeAttributes(ctx context.Context) error {
 			return NewCriticalError(fmt.Errorf("unknown InsertHeadBlock error type %d: %w", errType, err))
 		}
 	}
+	eq.safeAttributes = eq.safeAttributes[1:]
+	eq.logSyncProgress("processed safe block derived from L1")
+
+	return nil
+}
+
```
tynes, line 494: do you mind adding godocs to these new functions?

tynes: i see that there are godocs on the interface, any implementation specific details that would be useful in the godoc? perhaps the todo

op-node/rollup/derive/engine_queue.go
```diff
@@ -91,6 +120,8 @@ type EngineQueue struct {
 	l1Fetcher L1Fetcher
 }
 
+var _ EngineControl = (*EngineQueue)(nil)
```
tynes, line 123: Maybe something to document about the engine queue is that it should not be used concurrently, had the thought after reading `resetBuildingState`

op-node/rollup/derive/pipeline.go
```diff
@@ -130,6 +132,25 @@ func (dp *DerivationPipeline) UnsafeL2Head() eth.L2BlockRef {
...
+	return dp.eng.CancelPayload(ctx, force)
+}
+
+func (dp *DerivationPipeline) BuildingPayload() (onto eth.L2BlockRef, id eth.PayloadID, safe bool) {
+	return dp.eng.BuildingPayload()
+}
+
+// SetUnsafeHead changes the forkchoice state unsafe head, without changing the engine.
+//
+// deprecated: use the EngineControl interface instead.
```
tynes, line 153: What do you think about logging a deprecation warning here? To make it obvious if this is being called still

op-node/rollup/derive/engine_queue.go
protolambda: That applies to the whole `derive` package. Not sure what the right place for that is. Can add a warning if it helps.

op-node/rollup/derive/engine_queue.go
protolambda: I don't like duplicating docs, and the interface is the first place the user looks & thus the best to document, so I didn't add docs here. No inspiration for implementation-specific notes either. Do you have suggestions for what to document?

op-node/rollup/derive/pipeline.go
protolambda: It's getting removed in the PR 3 in the stack, and don't think we really need to do a lot around deprecation here since it's not used something someone external would depend on. Mostly just putting it here in case that later PR in the stack takes a while to land.

trianglesphere (approved):
lgtm

PR #4756, opened 2023-01-21
op-node: implement metered EngineControl, a wrapper to capture metrics

protolambda:
**Description**

This introduces a minimal wrapper around `EngineControl` that maintains metrics. This way we do not have to worry about maintaining them in any of the engine code-path.
Doing it this way keeps it separate from the derivation (useful because we cannot use `time.Now` in `derive` because of Cannon limitations) and keeps the `Sequencer` simple.

**Tests**

Metrics-only without any significant branching logic, and op-e2e passes through this already.

Note: stacked on #4755 and draft until that is merged.

Full PR stack:
- #4755
- #4756
- #4757
- #4758


op-node/rollup/driver/metered_engine.go
```diff
@@ -0,0 +1,95 @@
...
+	payload, errType, err := m.inner.ConfirmPayload(ctx)
+	if err != nil {
+		m.metrics.RecordSequencingError()
+		return payload, errType, err
+	}
+	now := time.Now()
+	sealTime := now.Sub(sealingStart)
+	buildTime := now.Sub(m.buildingStartTime)
+	m.metrics.RecordSequencerSealingTime(sealTime)
+	m.metrics.RecordSequencerBuildingDiffTime(buildTime - time.Duration(m.cfg.BlockTime)*time.Second)
```
tynes, line 78: `RecordSequencerBuildingDiffTime` is meant to be negative?

protolambda: yes! it's the difference between the block time and building time; building may take longer, or may not use the full time. Since the block time is variable between different networks, we track the difference instead, to keep metrics consistent between different configurations. The expected value should roughly match the sealing time, which we track in a different metric and can be compared. The metrics usage strings explain this (and matt is working on automated doc generation based on that content).

trianglesphere (approved):
lgtm. Nice hook in for the metrics

PR #4757, opened 2023-01-21
op-node: refactor sequencer to use EngineControl, test sequencing

protolambda:
**Description**

This cleans up the sequencer code-path by removing all remaining engine state from the `Driver` and `Sequencer` in favor of using just `EngineControl`,
and exposing a minimal sequencer planning interface to the `Driver` loop.

For origin selection the L1 head is still being passed around, this is addressed in the next PR.

**Tests**

I removed the trail-of-bits test because it's only testing 2 properties around sequencer state, which has now been refactored away.

Instead, we now have a large chaos-monkey like test:
we run the `Sequencer` with a faked `EngineControl` and inputs (origin selection / attributes building) and throw chaos it at it:
- make every type of input randomly and frequently fail
- make the engine control randomly fail to process the instructions of the sequencer
- mess with the sequencer clock

And then at the end of test we assert some chain health metrics,
to make sure we can deal with the above errors and still maintain a reasonably healthy L2 chain.

Removing the old test + writing this new large one also means this PR had to be larger, but I tried to keep the functionality changes limited to just the `Sequencer` and removal of the old sequencer state from the driver.

**Additional context**

Fix ENG-3104

Stacked on #4756 and draft until that is merged.

Full PR stack:
- #4755
- #4756
- #4757
- #4758


trianglesphere (approved):
lgtm. Nice tests

PR #4758, opened 2023-01-21
op-node: fix sequencer time drift edge case, strictly enforce sequencer conf depth

protolambda:
**Description**

L1 Origin selection had related problems in both sequencer and verifier areas. This PR fixes those.

As sequencer it didn't follow the safety-over-liveness priority, and ignored the sequencer conf depth to try and handle sequencer time drift.
This however is dangerous because of the recent verifier "fix" (good fix, but incomplete):
- When the max time drift is exceeded and the next origin cannot be found, then old origin would be used, causing the batch to be dropped.
- But the sequencer still produces blocks after this, which can have later origins & include transactions again, building on top of the empty block with old origin.

This means that as a verifier you end up dropping the empty batch with old origin, causing a halt (of safe blocks) until the sequencer window forces an empty block to be created.
And aside from halting as such, it risks adopting the next L1 origin differently than the sequencer, causing a reorg of the unsafe chain.

Meanwhile we still want deposits to be processed as soon as the sequencer time drift, so we should not give the sequencer too much leeway either,
and force them to adopt the next L1 origin when they theoretically can.

This also means that we need to accept batches with an old origin if the batches hit sequencer time drift edge cases
where it is just not possible to adopt the next L1 origin without breaking the `L2 time >= L1 origin time` invariant.

And while we are at it, we might as well solve the "what if L1 block time is smaller than L2 block time" edge case:
we can ignore the sequencer time drift check if the other checks pass & the batch advances the L1 origin forward.
Advancing the L1 origin is exactly what the sequencer time drift is specified for, so we should not disallow advancing
to a later L1 origin that we still exceed the L2 sequencer time drift of.

In the sequencer we should be strict, and avoid sequencing if we ever doubt about the validity of staying on the current L1 origin:
i.e. if we are over the sequencer time drift, and can't find the next L1 origin, and then wait until we see the L1 origin,
instead of creating a L2 block that could be dropped/reorged out.

With such a sequencer, we can also clean up the conf-depth approach:
the L1 head should not have to leak into the l1 origin selection logic if the logic already has accesses L1 through a limited by-number view.
Like the verifier conf depth, we can re-use the exact same tested code to enforce the conf depth for L1 origin selection.

Changes:
- Fix `L1OriginSelector` to implement the stricter sequencer origin selection as described above
- Update `Driver` to use `L1OriginSelector` with conf-depth util that we also use for verifier conf depth, and simplify the origin selection.
- Fix sequencer time drift section in `CheckBatch` to handle edge cases
- Fix derivation batch filtering spec to define the edge case filtering conditions in the sequencer time drift check case.
- testing, see below

**Tests**

- Updated and extended `driver/origin_selector_test.go` to cover the new strict sequencer behavior.
- Updated and extended `derive/batches_test.go` with cases that cover the (new) different branches you can take.
  - 96% filtering coverage at first. We were missing `uint64(batch.Batch.EpochNum) < epoch.Number`,
   so I fixed the `epoch too old` case to get 100% coverage.

**Invariants**

- Ensure `L2 time >= L1 origin time` at all times
- Ensure max sequencer time drift is respected, but make exception to always preserve L1 Origin advancement and above invariant.

**Additional context**

Note that the sequencer origin selection got stricter (shouldn't produce invalid cases anymore) and the verifier filtering got more relaxed (shouldn't drop the exceptions to the sequencer time drift rule anymore).

Goerli users who don't update are likely risk of halting their view of the chain upon highly irregular L1 chain advancement or very delayed L1 origin inclusion.
This, together with these events already being ultra-rare (especially with the quite high sequencer time drift of 10 minutes),
means that I think we can safely roll this out to Goerli & encourage updating.

Fix ENG-3255

Stacked on #4757 and draft until that is merged.

Full PR stack:
- #4755
- #4756
- #4757
- #4758

### DIFF
--- a/op-e2e/actions/l2_sequencer.go
+++ b/op-e2e/actions/l2_sequencer.go
@@ -18,11 +18,11 @@ type MockL1OriginSelector struct {
 	originOverride eth.L1BlockRef // override which origin gets picked
 }
 
-func (m *MockL1OriginSelector) FindL1Origin(ctx context.Context, l1Head eth.L1BlockRef, l2Head eth.L2BlockRef) (eth.L1BlockRef, error) {
+func (m *MockL1OriginSelector) FindL1Origin(ctx context.Context, l2Head eth.L2BlockRef) (eth.L1BlockRef, error) {
 	if m.originOverride != (eth.L1BlockRef{}) {
 		return m.originOverride, nil
 	}
-	return m.actual.FindL1Origin(ctx, l1Head, l2Head)
+	return m.actual.FindL1Origin(ctx, l2Head)
 }
 
 // L2Sequencer is an actor that functions like a rollup node,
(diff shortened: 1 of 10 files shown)


protolambda:
Closing in favor of #4809 and #4808

PR #4759, opened 2023-01-22
proxyd: Add sender-based rate limiter

mslipper:
This PR adds a sender-based rate limiter to proxyd. The new rate limiter throttles eth_sendRawTransaction requests based on sender in order to prevent abusive behavior on the sequencer.

Integration tests have been added to test this functionality.

It is safe to deploy this without changes to the proxyd config - the new limiter will just be disabled. To enable the limiter, add the following stanza to `proxyd.toml`:

```toml
[sender_rate_limit]
enabled = true
interval = "1s"
limit = 1
```


proxyd/server.go
```diff
@@ -575,6 +596,54 @@ func (s *Server) isUnlimitedUserAgent(origin string) bool {
...
+	// Inflates a types.Transaction object from the transaction's raw bytes.
+	tx := new(types.Transaction)
+	if err := tx.UnmarshalBinary(data); err != nil {
+		log.Debug("could not unmarshal transaction", "err", err, "req_id", GetReqID(ctx))
+		return ErrInvalidParams(err.Error())
+	}
+
+	// Convert the transaction into a Message object so that we can get the
+	// sender. This method performs an ecrecover, which can be expensive.
+	msg, err := tx.AsMessage(types.NewLondonSigner(tx.ChainId()), nil)
```
Inphi, line 628: nit: Can use types.LatestSignerForChainId instead for forwards-compatibility with future tx types

Inphi (approved):

zhwrd (approved):

proxyd/config.go
```diff
@@ -104,21 +104,28 @@ type BatchConfig struct {
 	ErrorMessage string `toml:"error_message"`
 }
 
+type SenderRateLimitConfig struct {
```
tynes, line 107: nit: godoc on this struct so its obvious what the config does

proxyd/server.go
```diff
@@ -409,6 +419,17 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 			continue
 		}
 
+		// Apply a sender-based rate limit if it is enabled. Note that sender-based rate
+		// limits apply regardless of origin or user-agent. As such, they don't use the
+		// isLimited method.
+		if parsedReq.Method == "eth_sendRawTransaction" && s.senderLim != nil {
```
tynes, line 425: Did you mean to add this only to the batch rpc path?

proxyd/server.go
```diff
@@ -135,6 +140,10 @@ func NewServer(
 			return nil, err
 		}
 	}
+	var senderLim FrontendRateLimiter
+	if senderRateLimitConfig.Enabled {
+		senderLim = limiterFactory(time.Duration(senderRateLimitConfig.Interval), senderRateLimitConfig.Limit, "senders")
```
tynes, line 145: We should probably disallow config that will brick accepting transactions, so interval or limit cannot be 0

proxyd/server.go
mslipper: Proxyd makes everything into a batch so this was intended, yeah.

PR #4761, opened 2023-01-23
op-bindings: Check for duplicate artifacts

mslipper:
Checks for duplicate artifacts when instantiating the `hardhat` object. This armors against an issue in the regentool where the regentool was passing in the deployments directory rather than the network directory. This led to artifacts being non-deterministically chosen based on however `fs.WalkDir` iterated over the deployments directory.

Fixes ENG-3054


tynes (approved):

tynes:
Good catch, the directory walking is fragile to passing in the wrong directory

PR #4765, opened 2023-01-23
feat: l1 and l2 source validation

refcell:
**Description**

Fixes ENG-3134

Fixes ENG-3040

On `op-node` initialization, configuration variables are at most loosely checked. This pr introduces validation for L1 and L2 configuration variables that check the L1 and L2 client `op-node/sources` match. (e.g. if an L2 client is provided to the op-node config as an L1 client, this pr makes the `op-node` fail fast on initialization, rather than omit unintuitive errors)


op-node/node/node.go
```diff
@@ -124,6 +124,30 @@ func (n *OpNode) initL1(ctx context.Context, cfg *Config) error {
 		return fmt.Errorf("failed to create L1 source: %w", err)
 	}
 
+	// Validate the L1 Client Chain ID
+	id, err := n.l1Source.L1ChainID(ctx)
+	if err != nil {
+		n.log.Warn("failed to verify L1 RPC chain id", "err", err)
+	} else {
+		if cfg.Rollup.L1ChainID != id {
+			n.log.Warn("incorrect L1 RPC chain id", "expected", cfg.Rollup.L1ChainID, "actual", id)
```
tynes, line 133: Perhaps this should return an error, we don't want syncing inconsistencies

tynes: I think that both of these checks should return errors if they fail

op-node/node/node.go
```diff
@@ -189,6 +213,30 @@ func (n *OpNode) initL2(ctx context.Context, cfg *Config, snapshotLog log.Logger
 		return fmt.Errorf("failed to create Engine client: %w", err)
 	}
 
+	// Validate the L2 Client Chain ID
```
tynes, line 216: What do you think about putting methods on the `cfg.Rollup`?

- `checkL1(backend) err`
- `checkL2(backend) err`

Then the logic can live there instead of here

cc @trianglesphere @protolambda

op-node/node/node.go
tynes: i also think if the rpc request fails, it should error

op-node/node/node.go
trianglesphere: needs a context passed in, but that works & seems pretty nice

op-node/node/node.go
```diff
@@ -124,6 +124,16 @@ func (n *OpNode) initL1(ctx context.Context, cfg *Config) error {
 		return fmt.Errorf("failed to create L1 source: %w", err)
 	}
 
+	// Validate the L1 Client Chain ID
+	if err := cfg.Rollup.CheckL1ChainID(ctx, n.l1Source); err != nil {
+		n.log.Error("failed to verify L1 RPC chain id", "err", err)
```
tynes, line 129: I think this should return the error instead of logging it to prevent the node from starting

op-node/rollup/types.go
```diff
@@ -55,6 +56,64 @@ type Config struct {
...
+	L1BlockRefByNumber(context.Context, uint64) (eth.L1BlockRef, error)
+}
+
+// CheckL1ChainID checks that the configured L1 chain ID matches the client's chain ID.
+func (cfg *Config) CheckL1ChainID(ctx context.Context, client L1Client) error {
+	id, err := client.L1ChainID(ctx)
+	if err != nil {
+		return err
+	}
+	if cfg.L1ChainID != id {
```
tynes, line 70: Hmm this is a `big.Int`, don't you have to use `.Cmp` and check if `nil`?

op-node/node/node.go
```diff
@@ -189,6 +199,16 @@ func (n *OpNode) initL2(ctx context.Context, cfg *Config, snapshotLog log.Logger
 		return fmt.Errorf("failed to create Engine client: %w", err)
 	}
 
+	// Validate the L2 Client Chain ID
+	if err := cfg.Rollup.CheckL2ChainID(ctx, n.l2Source); err != nil {
+		n.log.Error("failed to verify L2 RPC chain id", "err", err)
+	}
+
```
tynes, line 206: Slight preference to combine these into a single function that checks both the chainid and blockhash, what do you think?

op-node/sources/l1_client.go
```diff
@@ -68,6 +69,11 @@ func NewL1Client(client client.RPC, log log.Logger, metrics caching.Metrics, con
 	}, nil
 }
 
+// L1ChainID fetches the chain id for the RPC Client.
+func (s *L1Client) L1ChainID(ctx context.Context) (*big.Int, error) {
```
tynes, line 73: Can we just make this be `ChainID`? instead of `L1ChainID`?

op-node/sources/l2_client.go
```diff
@@ -87,6 +88,12 @@ func NewL2Client(client client.RPC, log log.Logger, metrics caching.Metrics, con
 	}, nil
 }
 
+// L2ChainID fetches the chain id for the RPC Client.
+func (s *L2Client) L2ChainID(ctx context.Context) (*big.Int, error) {
```
tynes, line 92: Same as above, just `ChainID`

tynes: Hmm I do see that the other methods are prefixed with `L2`

op-node/sources/l2_client.go
```diff
@@ -87,6 +87,7 @@ func NewL2Client(client client.RPC, log log.Logger, metrics caching.Metrics, con
 	}, nil
 }
 
+// L2BlockRefByLabel returns the L2 block reference for the given label.
 // L2BlockRefByLabel returns the [eth.L2BlockRef] for the given block label.
```
trianglesphere, line 91: did you intended to mostly duplicate the comment here?

trianglesphere (approved):
one small q about a comment. Rest looks really good.

tynes (approved):

trianglesphere (approved):

PR #4766, opened 2023-01-23
feat: op-node docs

refcell:
**Description**

In-line documentation for the `op-node` split out of #4743


tynes (approved):
This is great, we should strive to have amazing godoc and host them someplace

PR #4775, opened 2023-01-24
go: Remove individual versions for bedrock code

trianglesphere:
**Description**

The following packages had their go.mod file removed:
* op-e2e
* op-batcher
* op-bindings
* op-chain-ops
* op-node
* op-proposer
* op-service
* op-signer

They all had cross dependencies previously and used a go.work file to assume they could be built on the latest code rather than on the versioned code.

This means that all of these packages are versioned with the root go.mod. For packages that still have a mod file, they are versioned independently.

I also had to modify `op-heartbeat` and the `indexer` packages to have them rely on their specified dependencies instead of them using the same commit & the `go.work`

**Metadata**
- Fixes CLI-3301


tynes:
This looks like builds are breaking still

trianglesphere:
> This looks like builds are breaking still

Yea, some of the builds that I've flagged as "not bedrock" still rely on the latest version of bedrock code (op-heartbeat, indexer)

mslipper (approved):
LGTM pending conflict resolution. I'll see if I can resolve them on my own.

We can fix Snyk separately.

PR #4776, opened 2023-01-24
Fix specs discrepancy between the diagram and text

bap2pecs:
^


trianglesphere (changes requested):
Thanks for looking into this, but the spec is correct & the diagram is out of date. If you want to update the diagram, feel free to do so in this PR.

`newPayload` is the correct name. https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#engine_newpayloadv1

bap2pecs:
thanks @trianglesphere. I sent out a new PR to update the diagram: https://github.com/ethereum-optimism/optimism/pull/4781

PR #4777, opened 2023-01-24
feat(op-chain-ops): EOF Crawler CLI tool

clabby:
# Overview

Adds a simple CLI tool to scan a Geth LevelDB for accounts containing the `0xEF` EOF prefix in their runtime bytecode.

**Metadata**
Closes ENG-3031


clabby:
* develop
  * **PR #4777** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4777" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4777?utm_source=stack-comment).

op-chain-ops/eof-crawler/eof_crawler.go
```diff
@@ -0,0 +1,135 @@
...
+			Root:      data.Root[:],
+			CodeHash:  data.CodeHash,
+			SecureKey: it.Key,
+		}
+
+		// Attempt to get the address of the account from the trie
+		addrBytes := st.Get(it.Key)
+		if addrBytes == nil {
+			// Preimage missing! Cannot continue.
+			missingPreimages++
```
trianglesphere, line 96: How many missing pre-images are there? We can actually check the contract code without the preimage, it's just hard to deal with the hash of the address.

clabby: I've only tested this with the [Bedrock datadir for Goerli](https://community.optimism.io/docs/developers/bedrock/public-testnets/#goerli), and there were 0 there. The preimages should always be there given a fully synced node from what I can gather.

Interesting though- How would we check the contract code without the preimage? Looking at geth's state package, it doesn't look possible without the address from what I can see.

clabby: See: [StateDB.Code](https://github.com/ethereum/go-ethereum/blob/59a48e0289b1a7470a8285e665cab12b29117a70/core/state/state_object.go#L455)

op-chain-ops/eof-crawler/eof_crawler.go
```diff
@@ -0,0 +1,135 @@
+package main
```
tynes, line 1: nit: this should expose functions that are imported into `cmd/eof-crawler/main.go` and the main binary should be there

op-chain-ops/eof-crawler/eof_crawler.go
```diff
@@ -0,0 +1,135 @@
...
+	Address   common.Address `json:"address,omitempty"`
+	SecureKey hexutil.Bytes  `json:"key,omitempty"`
+}
+
+// emptyCodeHash is the known hash of an account with no code.
+var emptyCodeHash = crypto.Keccak256(nil)
+
+func main() {
+	// Open an existing Ethereum database
+	db, err := rawdb.NewLevelDBDatabase(os.Args[1], 16, 16, "", true)
```
tynes, line 36: nit: no bounds check on `os.Args`. Could use flags package with a default value? Or just add a bounds check with error returned

op-chain-ops/cmd/eof-crawler/main.go
```diff
@@ -0,0 +1,47 @@
...
+			},
+		},
+		Action: func(ctx *cli.Context) error {
+			dbPath := ctx.String("db-path")
+			if len(dbPath) == 0 {
+				return errors.New("Must specify a db-path")
+			}
+			out := ctx.String("out")
+
+			eof.IndexEOFContracts(dbPath, out)
```
tynes, line 38: This should return an error:

```go
if err := eof.IndexEOFContracts(...); err != nil { return err }
```

op-chain-ops/eof/eof_crawler.go
```diff
@@ -0,0 +1,137 @@
...
+	Address   common.Address `json:"address,omitempty"`
+	SecureKey hexutil.Bytes  `json:"key,omitempty"`
+}
+
+// emptyCodeHash is the known hash of an account with no code.
+var emptyCodeHash = crypto.Keccak256(nil)
+
+// IndexEOFContracts indexes all the EOF contracts in the state trie of the head block
+// for the given db and writes them to a JSON file.
+func IndexEOFContracts(dbPath string, out string) {
```
tynes, line 36: Preference for this to return an error instead of `panic`ing bc then it makes it more portable

op-chain-ops/eof/eof_crawler.go
```diff
@@ -0,0 +1,140 @@
...
+	nonEofContracts := uint64(0)
+	eofContracts := make([]Account, 0)
+
+	it := trie.NewIterator(st.NodeIterator(nil))
+	for it.Next() {
+		// Decode the state account
+		var data types.StateAccount
+		err := rlp.DecodeBytes(it.Value, &data)
+		if err != nil {
+			return errors.New(fmt.Sprintf("Failed to decode state account: %v", err))
```
tynes, line 77: `fmt.Errorf()` over wrapping `errors.New(fmt.Sprintf(...))`

op-chain-ops/eof/eof_crawler.go
```diff
@@ -0,0 +1,140 @@
...
+		addrBytes := st.Get(it.Key)
+		if addrBytes == nil {
+			// Preimage missing! Cannot continue.
+			missingPreimages++
+			continue
+		}
+		addr := common.BytesToAddress(addrBytes)
+
+		// Attempt to get the code of the account from the trie
+		code, err := stateDB.ContractCode(crypto.Keccak256Hash(addrBytes), common.BytesToHash(data.CodeHash))
```
tynes, line 106: `ContractCode` takes a hash?

tynes: Ok yes it does

tynes (approved):

op-chain-ops/eof-crawler/eof_crawler.go
trianglesphere: It's less useful, but would be important to follow up on if we found that there were missing preimages.
So `it.Key == addrHash`. The MPT is just bytes->bytes mapping. The secure MPT (which is what the accounts + storage tries are) is a `keccak(key)->value` mapping.

clabby:
@mergifyio requeue

PR #4778, opened 2023-01-24
feat(ci): add nx to monorepo for incremental cachable builds

roninjin10:
**What is nx**
- Incrementally builds and caches packages.   Similar to bazel.  Extremely powerful monorepo tooling
- nice cli interface that allows anybody to target any package whether it's golang or ts
- optional vs code plugin
- robust ecosystem of nx boilerplates you can install with cli and other tooling
- builds tests and lints golang, solidity, and typescript
- provides a nice remote cache to speed up your local environment
- provides an opt in nx cloud cache to speed up ci and provide more caching when building off of develop
- easy to configure in simple case
- Has extremely robust featureset if you would need it
- Very mature tool has been around for over half a decade
- Gateway repo has been using this with no issues for a while now.

**Implementation**
- nx.json provides a general config that all typescript packages inherit from
- you can also do config.json to granularly configure an individual package

**How nx works**
You have build targets and file groups.   You build a dependency tree by declaring what files packages and targets you depend on.  This is exactly like bazel but lighter weight.  It analyzes this to build a dependency graph.  Then based on files changed it will incrementally build cache and test anything that needs to be rebuilt.

**How much faster does nx make the build**
Depends on which caches you bust.  In general if we want this repo to be fast, we need 3 things
1. Smart building.   Don't rebuild stuff twice.
2. Pnpm.   Build node modules 2x to 3x faster
3. Esbuild.   Build typescript 5x faster

In the case of a commit changing golang, no typescript or solidity needs to be built.  Never having to build the old contracts is a nice win in speed too.

**Why am I doing this pr now**

I actually did most of this months ago but circled back because I was curious how well nx works with golang.

[Video of me demoing nx](https://watch.screencastify.com/v/iRIqAKOP281g8CqUfpZD)


roninjin10:
I put do-not-merge on this because I have had 2 prs in past that have broken the build while passing CI so I don't trust CI

roninjin10:
putting in draft until I can confirm this doesn't in fact break the build

roninjin10:
I think this is Gtg though

packages/contracts/package.json
```diff
@@ -20,8 +20,8 @@
     "standards"
   ],
   "scripts": {
-    "build": "yarn build:contracts && yarn copy:contracts && yarn autogen:artifacts && yarn build:typescript",
-    "build:typescript": "tsc -p ./tsconfig.json",
+    "build": "npx nx run-many --targets=build:contracts,copy:contracts,autogen:artifacts,build:ts",
```
roninjin10, line 23: I should make this us nx more idiomatically like contracts-bedrock does

maurelian:
I've finally found a moment to take a look at this. 

### Pros I see:

1. There is no change required to the developer workflow. All existing `yarn` scripts continue to work. 
2. It is faster, but not crazy faster. Bigger difference on `yarn` than `yarn build`.
```shell
# on willc/nx (rebased on develop) after yarn clean
time yarn # 9.36s
time yarn build #  130.37s

# on develop after yarn clean
time yarn # 37.10s
time yarn build  # 178.94s

```

### Cons I see:

Introduces another config file for everyone to learn and maintain.

packages/contracts-bedrock/package.json
```diff
@@ -15,12 +15,14 @@
   ],
   "scripts": {
     "bindings": "cd ../../op-bindings && make",
-    "build:forge": "forge build",
     "build:with-metadata": "FOUNDRY_PROFILE=echidna yarn build:forge",
     "build:differential": "tsc scripts/differential-testing.ts --outDir dist --moduleResolution node --esModuleInterop",
     "build:fuzz": "go build -o test-case-generator test-case-generator/cmd/fuzz.go",
-    "prebuild": "yarn ts-node scripts/verify-foundry-install.ts",
-    "build": "hardhat compile && yarn autogen:artifacts && yarn build:ts && yarn typechain",
+    "build": "npx nx build:ts && npx nx typechain",
```
maurelian, line 21: Why `npx nx` rather than `yarn nx`?

roninjin10: in general the difference is npx will download the library if you don't already have it installed.  But it first detects if you have it installed already.  So in this case there isn't a difference.   

It can still be nice though.  For example, if a tool uses forge and doesn't need node modules you can still npx do stuff without needing to do a full yarn install.

maurelian: I suppose also that this PR moves `yarn` closer to not much more than a wrapper around `nx`, so using `npx` moves us closer to being able to remove `yarn`, yeah?

Would it eventually work to rewrite this as follows and then call it with `nx build`?
```suggestion
    "build": "nx build:ts && nx typechain",
```

tynes:
I am in favor of this, is there anything we need to do to get this over the line besides testing that all commands work still?

package.json
```diff
@@ -38,10 +38,10 @@
...
   "scripts": {
     "clean": "yarn lerna run clean --parallel",
-    "build": "yarn lerna run build",
-    "test": "yarn lerna run test --parallel",
+    "build": "npx nx run-many --target=build --all --parallel 3",
+    "test": "npx nx run-many --target=build --all --parallel 3",
+    "lint": "npx nx run-many --target=lint --all --parallel 3",
     "test:coverage": "yarn lerna run test:coverage --parallel",
-    "lint": "yarn lerna run lint",
     "lint:ts:check": "yarn lerna run lint:ts:check",
```
tynes, line 45: It looks like there is still some lerna stuff in here, we would also want to replace that as well?

op-proposer/project.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@eth-optimism/op-proposer",
+  "$schema": "../node_modules/nx/schemas/project-schema.json",
+  "projectType": "application",
+  "sourceRoot": "op-proposer",
+  "targets": {
+    "build": {
+      "executor": "@nx-go/nx-go:build",
+      "options": {
+        "outputPath": "dist/op-proposer",
```
trianglesphere, line 10: our makefiles point to the `bin` folder.

op-proposer/project.json
```diff
@@ -0,0 +1,25 @@
...
+        "main": "op-proposer/cmd/main.go"
+      }
+    },
+    "serve": {
```
trianglesphere, line 14: I'm not sure this makes much sense - there is a lot of config required for these services.

trianglesphere (commented):
I'm a little uncertain the benefit for the go code - go already does incremental builds/tests and is very fast at builds even from scratch.

protolambda:
Nx looks great for JS/TS, but I am opposed to the changes to the Go modules:
- We've never had any significant performance issue with building them
- `nx serve` makes no sense when the Go programs require configuration with flags etc. & we have full automation in different places already
- Wrapping Go tools with javascript tools is a big anti-pattern IMO

Can we exclude the bedrock Go modules from this change please?

roninjin10:
Sounds like golang already has incremental builds and when testing them they were really fast anyways from scratch.   What would make sense would be to remove golang as suggested.    

Eventually though, you would want to wrap golang with nx as one giant build target.   Golang still handles the build but nx can still do the simple task of asking "did a golang package change or did a typescript package change"?

package.json
roninjin10: Yea nx and tsup can replace lerna and that would speed up even cold builds from scratch

packages/contracts-bedrock/package.json
roninjin10: Yea using npx makes it less tied to yarn that is true.  I suggest anybody using nx `npm install nx --global` so they never have to type either yarn or tsx in their dev env though.

maurelian:
> I am in favor of this, is there anything we need to do to get this over the line besides testing that all commands work still?

I'd assume that if CI passes, then we're good, and any edge cases can get cleaned up later.

roninjin10:
> I am in favor of this, is there anything we need to do to get this over the line besides testing that all commands work still?


Nothing really.  I think just this:

1. Remove the golang completely
2. Do extra work to make sure we don't break ci build
3. Remove the remote cache key for now.   It's easy to add back now but better to incrementally slowly adopt a cache.  This remote cache can really speed up ci later though especially for golang changes that don't touch typescript as well as typescript changes.   Also would speed up a cold start pulling develop for anybody
4. Consider doing a follow up pr to nuke lerna completely for nx

roninjin10:
- 
- It is faster, but not crazy faster. Bigger difference on yarn than yarn build.
- # on willc/nx (rebased on develop) after yarn clean
- time yarn # 9.36s
- time yarn build #  130.37s
- 
- # on develop after yarn clean
- time yarn # 37.10s
- time yarn build  # 178.94s

Worth noting if we had develop branch pushing to the remote cache your yarn build would have been a few seconds just the time to download the build artifacts

packages/contracts/package.json
roninjin10: These change way less than contracts-bedrock so the lack of granularity is fine

package.json
roninjin10: Would do this in a seperate pr since lerna does some other stuff like publish npm packages that would need to be replaced too

roninjin10: actually we aren't using lerna for that I was mistaken so nuked lerna now

tynes:
A rebase will fix `op-heartbeat tests`

roninjin10:
I think it's ready going to put it in review monday but still being very careful.

The biggest issue I see is the cache gets busted sometimes when it should not be busted.   This causes no real harm though other than the cache not being as smart as it could be and we can tweak that.

roninjin10:
do not merge and ready for review now actually so I have a rebase button to click

roninjin10:
tests still failing

roninjin10:
If we merge this and we also merge https://github.com/ethereum-optimism/optimism/pull/5260

- Any changes to golang code won't have to wait for ts to build ever
- Any changes to js code will be significantly faster to build
- If we turn on the cache even faster
- Issues with falsely caching test results or builds that I have seen in past will go away.  No longer need to manually set check_changed in circle ci in that brittle way

tldr I often hear client team complaining about slow ci and stuff and these 2 prs are the solution

roninjin10:
Also if we turned on remote cache (trivial to do) everytime you pulled develop branch you wouldn't need to rebuild anything it woudl just pull the builds from the cache.

package.json
```diff
@@ -35,14 +35,14 @@
...
-    "lint:check": "yarn lerna run lint:check",
-    "lint:fix": "yarn lerna run lint:fix --parallel",
+    "clean": "npx nx run-many --target=clean",
+    "build": "npx nx run-many --target=build",
+    "test": "npx nx run-many --target=test",
+    "lint": "npx nx run-many --target=lint",
+    "test:coverage": "npx nx run-many --target=test:coverage",
+    "lint:ts:check": "npx nx run-many --target=lint:ts:check",
+    "lint:check": "npx nx run-many --target=lint:check",
+    "lint:fix": "npx nx run-many --target=lint:fix",
```
roninjin10, line 45: these npx are not harming anything but are also unnecessary I can remove if we want
```suggestion
    "lint:fix": "nx run-many --target=lint:fix",
```

roninjin10:
reopening post bedrock

roninjin10:
Not a priority but going to reopen this and try to get it in in next few weeks

tynes:
Now is a good time to get this merged @roninjin10

tynes:
@roninjin10 I've rebased this PR, my one question is why we have the cloud package installed + it configured, its going to be a noop?

tynes:
ok i broke something

roninjin10:
@tynes ideally we should start using it as it will make our builds way way way way way faster. There is a default runner we could use instead.  I wonder if there is a nx cache that we could self host?

roninjin10:
It's never a good idea to turn the cache on right away though.  The cache will be doing it's thing locally so if we are able to use the local one with no issues for 2 weeks or so we can safely turn on the remote cache

roninjin10 (commented):
lgtm

tynes (approved):

PR #4779, opened 2023-01-24
Fix the deposits.md spec

bap2pecs:
I think the description at https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/contracts/L2/L1Block.sol#L42 makes more sense

### DIFF
--- a/specs/deposits.md
+++ b/specs/deposits.md
@@ -221,9 +221,7 @@ The predeploy stores the following values:
   - `timestamp` (`uint64`)
   - `basefee` (`uint256`)
   - `hash` (`bytes32`)
-- `sequenceNumber` (`uint64`): This equals the L2 block number relative to the start of the epoch,
-  i.e. the L2 block distance to the L2 block height that the L1 attributes last changed,
-  and reset to 0 at the start of a new epoch.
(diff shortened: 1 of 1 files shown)


trianglesphere (changes requested):
Hey, I don't think we'll be bringing in this exact change.

The contracts definition is shorter, but does not capture the fact that the sequence number is different for every block in the epoch. It is very explicitly not "number of L2 blocks in the same epoch" but the "number of the specific L2 block inside the epoch".

bap2pecs:
thanks @trianglesphere for the review! But I am still a bit confused about this field.

iiuc, the value in this predeploy contract will be changed in the first transaction in the first L2 block of the epoch. It will remain unchanged until the next epoch.

If this is correct, it represents a different thing as the "sequence number" you mentioned in

> the sequence number is different for every block in the epoch

trianglesphere:
> first transaction in the first L2 block of the epoch

It is changed in the first transaction of every L2 block. Every L2 block has an [L1 attributes tx](https://github.com/ethereum-optimism/optimism/blob/develop/specs/glossary.md#l1-attributes-deposited-transaction)

bap2pecs:
ah I see. thanks for the clarification @trianglesphere. closing out this PR.

PR #4780, opened 2023-01-24
[ENG-2626] op-batcher: Efficient batching

sebastianst:
The current batcher eagerly creates channels each polling interval and sends off the channel. Most of the time, this leads to rather small transactions.

This PR improves the batcher to track how much input data has been put into a channel yet and estimate the compressed output data size. Only if an estimated target channel size has been reached is the channel closed. In particular, a channel might now survive multiple polling intervals.

Tracking channel timeouts and the sequencer window (ENG-2625) will be done in another PR.

Fixes ENG-2626


op-batcher/batcher/channel_builder.go
```diff
@@ -0,0 +1,265 @@
...
+func (c *channelBuilder) HasFrame() bool {
+	return len(c.frames) > 0
+}
+
+func (c *channelBuilder) NumFrames() int {
+	return len(c.frames)
+}
+
+// NextFrame returns the next available frame.
+// HasFrame must be called prio to check if there's a next frame available.
```
refcell, line 246: Nit
```suggestion
// HasFrame must be called prior to check if there's a next frame available.
```

op-batcher/batcher/channel_builder.go
```diff
@@ -0,0 +1,265 @@
...
+
+// AddBlock adds a block to the channel compression pipeline. InputTargetReached
+// should be called aftewards to test whether the target input data amount has
+// been reached. In this case, a new channel must be started.
+//
+// AddBlock returns ErrInputTargetReached if called despite having reached the
+// input data target.
+// It returns ErrTooManyRLPBytes if the channel cannot take more input
+// data. Then, too, a new channel has to be started.
+// It returns
```
refcell, line 113: Comment cut off here

sebastianst: Thanks, inspired me to update and improve the docs regarding full channels.

op-batcher/batcher/channel_builder.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, err
+	}
+
+	return &channelBuilder{
+		cfg: cfg,
+		co:  co,
+	}, nil
+}
+
+func (c *channelBuilder) ID() derive.ChannelID {
```
tynes, line 92: if you could make an interface that any impl of a channel builder could implement, that would be great

op-node/rollup/derive/channel_out.go
```diff
@@ -70,33 +70,32 @@ func (co *ChannelOut) Reset() error {
...
 	if co.rlpLength+buf.Len() > MaxRLPBytesPerChannel {
-		return fmt.Errorf("could not add %d bytes to channel of %d bytes, max is %d. err: %w",
+		return 0, fmt.Errorf("could not add %d bytes to channel of %d bytes, max is %d. err: %w",
 			buf.Len(), co.rlpLength, MaxRLPBytesPerChannel, ErrTooManyRLPBytes)
 	}
 	co.rlpLength += buf.Len()
 
-	_, err = io.Copy(co.compress, &buf)
-	return err
+	written, err := io.Copy(co.compress, &buf)
```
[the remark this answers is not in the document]
sebastianst: It is. I just found it semantically clearer to return the return values of `io.Copy`. I wanted to do `return io.Copy(...)` but couldn't because of missing implicit number type conversions.

op-node/rollup/derive/channel_out.go
```diff
@@ -70,33 +70,32 @@ func (co *ChannelOut) Reset() error {
...
-// if there is a problem adding the block. The only sentinel
-// error that it returns is ErrTooManyRLPBytes. If this error
-// is returned, the channel should be closed and a new one
-// should be made.
-func (co *ChannelOut) AddBlock(block *types.Block) error {
+// AddBlock adds a block to the channel. It returns the RLP encoded byte size
+// and an error if there is a problem adding the block. The only sentinel error
+// that it returns is ErrTooManyRLPBytes. If this error is returned, the channel
+// should be closed and a new one should be made.
+func (co *ChannelOut) AddBlock(block *types.Block) (uint64, error) {
```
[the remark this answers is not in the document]
sebastianst: Thanks, added `ChannelOut.InputBytes() int`!

This allowed me to remove the duplicated tracking of total input bytes. Because of that, now I don't need the returned `rlpbytes` of `ChannelOut.AddBlock` any more. Should I revert returning them? I can still imagine scenarios in the future where it could come in handy to have these returned, and it would be in line with some io functions of the std lib, like `io.Copy`, that return the written amount of data. No strong opinion on my side, but I'd keep it.

op-batcher/batcher/channel_builder.go
sebastianst: Good idea! However, I didn't really think the interface through to the point where I think it would be suitable for an arbitrary channel builder implementation. So what do you think, we leave it at this implementation for now and when the time comes that we want to create a second channel builder implementation, we rethink the interface?

op-node/rollup/derive/channel_out.go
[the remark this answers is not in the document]
trianglesphere: I'm fine either way. If you lean to keeping it, go for it

op-batcher/batcher/driver.go
```diff
@@ -162,8 +166,8 @@ func NewBatchSubmitterWithSigner(cfg Config, addr common.Address, signer SignerF
 		txMgr: NewTransactionManager(l, txManagerConfig, batchInboxAddress, chainID, addr, l1Client, signer(chainID)),
 		done:  make(chan struct{}),
 		log:   l,
-		state: NewChannelManager(l, cfg.ChannelTimeout),
-		// TODO: this context only exists because the event loop doesn't reach done
+		state: NewChannelManager(l, batcherCfg.Channel),
+		// TODO: this context only exists because the even loop doesn't reach done
```
trianglesphere, line 170:
```suggestion
		// TODO: this context only exists because the event loop doesn't reach done
```

sebastianst:
We only get the hive tests to work after we merge this and update the batcher config in the optimism hive repo.

op-batcher/batcher/channel_manager.go
```diff
@@ -153,130 +153,135 @@ func (s *channelManager) pendingChannelIsTimedOut() bool {
...
-		s.datas = s.datas[1:]
-		return r.data, r.id, nil
-	} else {
+	if s.pendingChannel == nil || !s.pendingChannel.HasFrame() {
+		s.log.Trace("no next tx data")
 		return nil, txID{}, io.EOF // TODO: not enough data error instead
 	}
+
+	id, data := s.pendingChannel.NextFrame()
+	if id.frameNumber == 0 {
```
trianglesphere, line 175: The version byte needs to be pre-pendended for every L1 transaction. It's only when we have multiple frames inside a single L1 tx where we don't need to include a version byte for every frames, just at the start. 

https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#batcher-transaction-format

op-batcher/batcher/channel_builder.go
sebastianst: We discussed to do it when we need it :)

op-batcher/batcher/channel_manager.go
sebastianst: Oh sorry, yes I know 🙈 think error

tynes:
Just wondering if some of the todos being added should be addressed now or if we are tracking them

trianglesphere (approved):
lgtm. Really nice work

trianglesphere:
> Just wondering if some of the todos being added should be addressed now or if we are tracking them

I took a look at the todo's. I'm not too worried about them as none of them where that bad & we're still actively working on this codebase.

PR #4783, opened 2023-01-24
op-service: Migrate tx manager package

trianglesphere:
**Description**

This places the txmgr package in the op-service folder instead of the op-proposer folder. This removes the proposer for being a dependency of the batcher. The txmgr code is shared code so it makes more sense to be there.

Also includes release v0.10.11 to properly move the tx manager.

**Metadata**
- Fixes ENG-3190


tynes:
We currently have 2 copies of the `txmgr` package in the monorepo, the other lives here: https://github.com/ethereum-optimism/optimism/tree/develop/bss-core/txmgr

Its a bunch of busywork to delete `bss-core/txmgr`, we will def be able to do it after bedrock mainnet

trianglesphere:
IIRC we only depend on the `op-proproser/service` txmgr in bedrock code. I intend to start modifying this new txmgr & then we'll nuke the old txmgr inside bss core once bedrock is live.

mslipper (approved):

PR #4784, opened 2023-01-24
fix batch-deriv-chain.svg

bap2pecs:
In the right bottom of the digram, the sequence number under block 11 should be 4 instead of 2.


trianglesphere (approved):
nice find. TYVM

PR #4788, opened 2023-01-25
op-batcher: better error handling logic

tynes:
**Description**

Be extra pendantic around expected data being not `nil` to prevent unexpected panics.


trianglesphere (approved):

trianglesphere:
This would happen with an unsynced node or if the node was returning pre-london blocks

PR #4789, opened 2023-01-25
op-node: more config checks

tynes:
**Description**

Check that the L1 contracts are deployed as expected. The accounts passed into the config should have code, otherwise there is a misconfiguration.


refcell (approved):
LGTM

PR #4791, opened 2023-01-26
feat(l2g): properly return NonceTooHigh

smartcontracts:
**Description**
Has the Sequencer properly return NonceTooHigh when the nonce on a transaction is higher than the nonce of the account. NonceTooHigh is not typically in this codepath, but here it makes sense because there is no mempool and the txn is being rejected outright.


tynes (approved):

tynes:
@mergifyio refresh

PR #4800, opened 2023-01-29
op-batcher: Watch timeouts [ENG-2625]

sebastianst:
**Description**

With #4780 , the batcher now fills up a channel until a target amount of input data has been added. So it's now possible that a channel times out, or the sequencer window passes, while a channel is collecting data, if not enough transactions are happening to produce enough data.

This PR adds timeout watching to the batcher.


op-batcher/batcher/driver.go
```diff
@@ -284,11 +288,16 @@ func (l *BatchSubmitter) loop() {
...
-			// Empty the state after loading into it on every iteration.
 		blockLoop:
 			for {
-				// Collect the output frame
-				data, id, err := l.state.TxData(eth.L1BlockRef{})
+				l1tip, err := l.l1Tip(l.ctx)
+				if err != nil {
+					l.log.Error("Failed to query L1 tip")
+					break
+				}
```
sebastianst, line 297: It may be a bit excessive to query the l1 head in every loop iteration. It should be enough to use the l1 head that's queried when the tx receipt is returned.

op-batcher/batcher/driver.go
```diff
@@ -320,3 +327,70 @@ func (l *BatchSubmitter) loop() {
...
+	// query the header.
+	l1ref, err := l.l1BlockRefByReceipt(l.ctx, receipt)
+	if err != nil {
+		// Very unlikely that tx sending worked but then we cannot get the
+		// header. Fall back to latest known L1 time to be on the safe side.
+		l1ref.Time = l.lastKnownL1Time
+		l.log.Warn("Failed to get block ref for successful batcher tx. Setting timestamp to latest know L1 block time.", "block_ref", l1ref)
+	} else {
+		l.lastKnownL1Time = l1ref.Time
+	}
```
sebastianst, line 348: We could simplify this quite a bit if we measure batcher channel timeouts in #blocks instead of seconds, because block numbers are part of tx receipts. It might then also be enough to only pass a `BlockID` to `TxData` instead of a `L1BlockRef`.

sebastianst: done

sebastianst: done

op-batcher/batcher/channel_builder.go
```diff
@@ -107,9 +117,33 @@ func (c *channelBuilder) Blocks() []*types.Block {
...
+}
+
+// TimedOut returns whether the passed block number is after the channel timeout
+// block. If no block timeout is set yet, it returns false.
+func (c *channelBuilder) TimedOut(blockNum uint64) bool {
+	return c.timeout != 0 && blockNum >= c.timeout
+}
+
+// TriggerTimeout checks if the channel is timed out at the given timestamp and
+// in this case sets the channel as full with reason ErrChannelTimedOut.
```
trianglesphere, line 140:
```suggestion
// TriggerTimeout checks if the channel is timed out at the given block number and
// if it is timed out, marks the channel as full with reason ErrChannelTimedOut.
```

trianglesphere: IDK if I have a better name & I'm fine merging without changing it, but `TriggerTimeout` is slightly confusing to me because it doesn't always mark the channel as timed out.

op-node/eth/id.go
```diff
@@ -47,6 +48,15 @@ type L1BlockRef struct {
 	Time       uint64      `json:"timestamp"`
 }
 
+func L1BlockRefFromHeader(block *types.Header) L1BlockRef {
```
trianglesphere, line 51: duplicate code to `block_info.go`

trianglesphere (commented):
I left a couple small comments, and I think after them it looks good to go.

op-batcher/batcher/driver.go
```diff
@@ -273,3 +275,26 @@ func (l *BatchSubmitter) loop() {
...
+// l1Tip gets the current L1 tip as a L1BlockRef. The passed context is assumed
+// to be a lifetime context, so it is internally wrapped with a network timeout.
+func (l *BatchSubmitter) l1Tip(ctx context.Context) (eth.L1BlockRef, error) {
+	tctx, cancel := context.WithTimeout(ctx, networkTimeout)
+	defer cancel()
+	head, err := l.L1Client.HeaderByNumber(tctx, nil)
+	if err != nil {
+		return eth.L1BlockRef{}, fmt.Errorf("getting latest L1 block: %w", err)
+	}
+	return eth.L1BlockRefFromHeader(head), nil
```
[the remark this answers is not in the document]
sebastianst: Thanks, didn't know `InfoToL1BlockRef` yet! A `*types.Header` isn't unfortunately directly usable as a `BlockInfo`. But I added a converter `eth.HeaderBlockInfo` in 6e88feb54 so we can use it as a `BlockInfo`.

op-node/eth/id.go
sebastianst: deleted

op-batcher/batcher/channel_builder.go
sebastianst: Agree, it's a confusing name. Renamed it to `CheckTimeout`.

trianglesphere (approved):
lgtm

PR #4802, opened 2023-01-30
maint(ctb): clean up Hashing tests

smartcontracts:
**Description**
Cleans up tests for Hashing.sol and uses new format.


tynes:
What do you mean by new format? Looks like this just nests function calls into `assertEq` when the variables are only used once so they dont need to be defined outside

smartcontracts:
> What do you mean by new format? Looks like this just nests function calls into `assertEq` when the variables are only used once so they dont need to be defined outside

It's changing the contracts so there's one contract per function being tested.

packages/contracts-bedrock/contracts/test/Hashing.t.sol
```diff
@@ -6,20 +6,33 @@ import { Types } from "../libraries/Types.sol";
...
+            0xf923fb07134d7d287cb52c770cc619e17e82606c21a875c92f4c63b65280a5cc
         );
+    }
+}
 
-        assertEq(sourceHash, 0xf923fb07134d7d287cb52c770cc619e17e82606c21a875c92f4c63b65280a5cc);
+contract Hashing_hashCrossDomainMessage_Test is CommonTest {
+    function setUp() external {
+        _setUp();
     }
```
maurelian, line 31: Nit: it'd be nice if this boilerplate was moved into a contract which is a parent of the other contracts.

Or the `CommonTest._setUp()` could just become a public `setUp()` function.

maurelian (approved):

packages/contracts-bedrock/contracts/test/Hashing.t.sol
tynes: I feel like there was a reason for `CommonTest` to have the prefix underscored set up function, i had trouble getting it to compile awhile ago when i was originally working on it, but i agree that would be best

tynes (approved):

PR #4806, opened 2023-01-31
allow body parser options to be passed in

roninjin10:
Add body parser options so we can configure larger max size

### DIFF
--- a/packages/common-ts/src/base-service/base-service-v2.ts
+++ b/packages/common-ts/src/base-service/base-service-v2.ts
@@ -124,6 +124,7 @@ export abstract class BaseServiceV2<
       metricsSpec: MetricsSpec<TMetrics>
       options?: Partial<TOptions & StandardOptions>
       loop?: boolean
+      bodyParserParams?: bodyParser.OptionsJson
     }
   ) {
     this.loop = params.loop !== undefined ? params.loop : true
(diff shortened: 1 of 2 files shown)


roninjin10:
* develop
  * **PR #4806** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4806" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4806?utm_source=stack-comment).

imranjami (approved):

packages/common-ts/src/base-service/base-service-v2.ts
```diff
@@ -330,6 +331,7 @@ export abstract class BaseServiceV2<
           verify: (req, res, buf, encoding) => {
             ;(req as any).rawBody = buf?.toString(encoding || 'utf8') || ''
           },
+          ...this.params.bodyParserParams,
```
tynes, line 334: Will this cause a problem if the params are undefined? It looks like they are optional

roninjin10: It won't because ...undefined and ...(undefined ?? {} or ...{} or ...null are all the same thing in javascript.  That said being explicit is a good thing
```suggestion
          ...(this.params.bodyParserParams ?? {}),
```

tynes (approved):

PR #4807, opened 2023-01-31
feat(op-node): Add alternative backup sync method via RPC

clabby:
# Overview

Adds an alternative backup sync method via RPC to the `op-node`.

The new CLI flag, `l2.backup-unsafe-sync-rpc`, feeds into the new `sources.L2SyncEndpointSetup`, which lives in the node config. This flag is optional- if it is not present, alternative sync will not be enabled. 

If the flag is set, a `sources.SyncClient` is initialized alongside the driver, which spawns a new event loop that the driver's state loop can communicate back and forth with in order to fetch missing blocks from a backup RPC.

**Tests**

- `op-e2e` - `TestSystemMockAltSync`
- Modified `TestPayloadsQueue` in the `op-node` to account for the `PayloadsQueue` changes.

**Metadata**
- Closes ENG-3218

### DIFF
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -11,7 +11,9 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rpc"
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
@@ -195,6 +197,11 @@ func (s *Driver) eventLoop() {
 		sequencerTimer.Reset(delay)
 	}
 
+	// Create a ticker to check if there is a gap in the engine queue every minute
+	// If there is, we send requests to the backup RPC to retrieve the missing payloads
+	// and add them to the unsafe queue.
+	altSyncTicker := time.NewTicker(60 * time.Second)
+
 	for {
 		// If we are sequencing, and the L1 state is ready, update the trigger for the next sequencer action.
 		// This may adjust at any time based on fork-choice changes or previous errors.
@@ -223,6 +230,61 @@ func (s *Driver) eventLoop() {
 				}
 			}
 			planSequencerAction() // schedule the next sequencer action to keep the sequencing looping
+		// TODO: Should this be lower-priority in the switch case?
+		case <-altSyncTicker.C:
+			size, start, end := s.derivation.GetUnsafeQueueGap()
+
+			// If there is a gap in the queue and a backup sync RPC is configured, attempt to retrieve the missing payloads from the backup RPC
+			if size > 0 && s.config.BackupL2UnsafeSyncRPC != "" {
+				// Dial the backup unsafe sync RPC.
+				client, err := rpc.DialHTTP(s.config.BackupL2UnsafeSyncRPC)
+				if err != nil {
+					s.log.Warn("failed to dial backup unsafe sync RPC", "backup rpc", s.config.BackupL2UnsafeSyncRPC, "err", err)
+					continue
+				}
+
+				// Attempt to fetch the missing payloads from the backup unsafe sync RPC concurrently.
+				// Concurrent requests are safe here due to the engine queue being a priority queue.
+				// TODO: Should enforce a max gap size to prevent spamming the backup RPC or being rate limited.
+				for i := start; i <= end; i++ {
+					go func(blockNumber uint64) {
+						s.log.Info("requesting unsafe payload from backup RPC", "block number", blockNumber, "backup rpc", s.config.BackupL2UnsafeSyncRPC)
+
+						// TODO: Post Shanghai hardfork, the engine API's `PayloadBodiesByRange` method will be much more efficient, but for now,
+						// the `eth_getBlockByNumber` method is more widely available.
+
+						// Fetch the next unsafe block from the backup unsafe sync RPC.
+						var block *types.Block
+						timeoutCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
+						defer cancel()
+						if err = client.CallContext(timeoutCtx, &block, "eth_getBlockByNumber", blockNumber); err != nil {
+							s.log.Warn("failed to retireve unsafe payload from backup RPC", "block number", blockNumber, "backup rpc", s.config.BackupL2UnsafeSyncRPC, "err", err)
+							return
+						}
+
+						// Convert the received block to a `eth.ExecutionPayload`.
+						payload, err := eth.BlockAsPayload(block)
+						if err != nil {
+							s.log.Warn("failed to convert block to execution payload", "block number", blockNumber, "backup rpc", s.config.BackupL2UnsafeSyncRPC, "err", err)
+							return
+						}
+
+						// TODO: Validate the integrity of the payload.
+						// Signature validation is not necessary here since the backup RPC is trusted. (?)
+						if _, ok := payload.CheckBlockHash(); !ok {
+							s.log.Warn("received invalid payload from backup RPC; invalid block hash", "payload", payload.ID(), "backup rpc", s.config.BackupL2UnsafeSyncRPC)
+							return
+						}
+
+						s.log.Info("received unsafe payload from backup RPC", "payload", payload.ID(), "backup rpc", s.config.BackupL2UnsafeSyncRPC)
+
+						// Send the retrieved payload to the `unsafeL2Payloads` channel.
+						s.unsafeL2Payloads <- payload
+
+						s.log.Info("inserted received unsafe payload into priority queue", "payload", payload.ID(), "backup rpc", s.config.BackupL2UnsafeSyncRPC)
+					}(i)
+				}
+			}
 		case payload := <-s.unsafeL2Payloads:
 			s.snapshot("New unsafe payload")
 			s.log.Info("Optimistically queueing unsafe L2 execution payload", "id", payload.ID())
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -232,59 +232,9 @@ func (s *Driver) eventLoop() {
 			planSequencerAction() // schedule the next sequencer action to keep the sequencing looping
 		// TODO: Should this be lower-priority in the switch case?
 		case <-altSyncTicker.C:
-			size, start, end := s.derivation.GetUnsafeQueueGap()
-
-			// If there is a gap in the queue and a backup sync RPC is configured, attempt to retrieve the missing payloads from the backup RPC
-			if size > 0 && s.config.BackupL2UnsafeSyncRPC != "" {
-				// Dial the backup unsafe sync RPC.
-				client, err := rpc.DialHTTP(s.config.BackupL2UnsafeSyncRPC)
-				if err != nil {
-					s.log.Warn("failed to dial backup unsafe sync RPC", "backup rpc", s.config.BackupL2UnsafeSyncRPC, "err", err)
-					continue
-				}
-
-				// Attempt to fetch the missing payloads from the backup unsafe sync RPC concurrently.
-				// Concurrent requests are safe here due to the engine queue being a priority queue.
-				// TODO: Should enforce a max gap size to prevent spamming the backup RPC or being rate limited.
-				for i := start; i <= end; i++ {
-					go func(blockNumber uint64) {
-						s.log.Info("requesting unsafe payload from backup RPC", "block number", blockNumber, "backup rpc", s.config.BackupL2UnsafeSyncRPC)
-
-						// TODO: Post Shanghai hardfork, the engine API's `PayloadBodiesByRange` method will be much more efficient, but for now,
-						// the `eth_getBlockByNumber` method is more widely available.
-
-						// Fetch the next unsafe block from the backup unsafe sync RPC.
-						var block *types.Block
-						timeoutCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
-						defer cancel()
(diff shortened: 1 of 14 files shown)


clabby:
* develop
  * **PR #4807** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4807" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/4807?utm_source=stack-comment).

op-node/rollup/driver/state.go
```diff
@@ -443,3 +456,49 @@ type hashAndErrorChannel struct {
...
+		}
+	}
+}
+
+// fetchUnsafeBlockFromRpc attempts to fetch an unsafe execution payload from the backup unsafe sync RPC.
+// WARNING: This function fails silently (aside from warning logs).
+func (s *Driver) fetchUnsafeBlockFromRpc(ctx context.Context, blockNumber uint64) {
+	s.log.Info("requesting unsafe payload from backup RPC", "block number", blockNumber)
+
+	// TODO: Post Shanghai hardfork, the engine API's `PayloadBodiesByRange` method will be much more efficient, but for now,
```
trianglesphere, line 482: @protolambda does this imply that we'll need to include engine authentication when we do this update?

op-node/rollup/driver/state.go
```diff
@@ -443,3 +456,49 @@ type hashAndErrorChannel struct {
...
+
+// checkForGapInUnsafeQueue checks if there is a gap in the unsafe queue and attempts to retrieve the missing payloads from the backup RPC.
+// WARNING: This function fails silently (aside from warning logs).
+func (s *Driver) checkForGapInUnsafeQueue(ctx context.Context) {
+	start, end := s.derivation.GetUnsafeQueueGap()
+	size := end - start
+
+	// If there is a gap in the queue and a backup sync client is configured, attempt to retrieve the missing payloads from the backup RPC
+	if size > 0 && s.l2SyncCl != nil {
+		// Attempt to fetch the missing payloads from the backup unsafe sync RPC concurrently.
```
trianglesphere, line 468: Everything inside the if statement should be async. I'd also consider removing the `fetchUnsafeBlockFromRpc` from the driver package. Another option for async communication which is more bounded is to send on channels rather than spawning go routines.

clabby: Sounds good- @protolambda recommended that we move the logic that concurrently fetches a range of blocks into a function within the `SyncClient` that accepts the `unsafeL2Payloads` channel and can feed them in as they are received, which should be a cleaner place to put the loop + logic in `fetchUnsafeBlockFromRpc`.

op-node/rollup/driver/state.go
protolambda: Yes, we can make it optional though: e.g. "to use an engine API endpoint for faster sync specify the `--sync.auth` flag" and we'll authenticate the RPC like an engine RPC. For now we can just stick to regular unauthenticated JSON-RPC.

op-node/rollup/driver/state.go
```diff
@@ -443,3 +456,49 @@ type hashAndErrorChannel struct {
...
+}
+
+// fetchUnsafeBlockFromRpc attempts to fetch an unsafe execution payload from the backup unsafe sync RPC.
+// WARNING: This function fails silently (aside from warning logs).
+func (s *Driver) fetchUnsafeBlockFromRpc(ctx context.Context, blockNumber uint64) {
+	s.log.Info("requesting unsafe payload from backup RPC", "block number", blockNumber)
+
+	// TODO: Post Shanghai hardfork, the engine API's `PayloadBodiesByRange` method will be much more efficient, but for now,
+	// the `eth_getBlockByNumber` method is more widely available.
+
```
tynes, line 484: Another option is to use batch rpc calls to prevent the need to open and close a bunch of http connections

clabby: Spoke with @protolambda about this- was decided against due to the batch request failing on partial results, which is not exactly the desired behavior here.

op-node/rollup/driver/state.go
```diff
@@ -195,6 +201,11 @@ func (s *Driver) eventLoop() {
 		sequencerTimer.Reset(delay)
 	}
 
+	// Create a ticker to check if there is a gap in the engine queue every minute
+	// If there is, we send requests to the backup RPC to retrieve the missing payloads
+	// and add them to the unsafe queue.
+	altSyncTicker := time.NewTicker(60 * time.Second)
```
tynes, line 207: I think we want a `defer altSyncTicker.Close()` here to prevent a memory leak

op-node/rollup/driver/state.go
```diff
@@ -442,3 +458,28 @@ type hashAndErrorChannel struct {
 	hash common.Hash
 	err  chan error
 }
+
+// checkForGapInUnsafeQueue checks if there is a gap in the unsafe queue and attempts to retrieve the missing payloads from the backup RPC.
+// WARNING: The sync client's attempt to retrieve the missing payloads is not guaranteed to succeed, and it will fail silently (besides
+// emitting warning logs) if the requests fail.
+func (s *Driver) checkForGapInUnsafeQueue(ctx context.Context) {
```
tynes, line 465: I feel like we should either only call this function if `s.L2SyncCl != nil` or move that check to the top of this function and then return if its `nil`

op-node/sources/sync_client.go
```diff
@@ -0,0 +1,116 @@
...
+		FetchUnsafeBlock: make(chan uint64),
+		done:             make(chan struct{}),
+	}, nil
+}
+
+// Start starts up the state loop.
+// The loop will have been started if err is not nil.
+func (s *SyncClient) Start(unsafeL2Payloads chan *eth.ExecutionPayload) error {
+	if unsafeL2Payloads == nil {
+		return errors.New("unsafeL2Payloads channel must not be nil")
```
tynes, line 58: Perhaps define the error at the top of this file `ErrNoUnsafeL2PayloadChannel` or something like that

op-node/rollup/driver/state.go
```diff
@@ -442,3 +458,28 @@ type hashAndErrorChannel struct {
...
+	start, end := s.derivation.GetUnsafeQueueGap()
+	size := end - start
+
+	// If there is a gap in the queue and a backup sync client is configured, attempt to retrieve the missing payloads from the backup RPC
+	if size > 0 {
```
tynes, line 470: This size check isn't necessary? Because the loop will just never run?  It is nice to gate the log to prevent spam

op-node/sources/sync_client.go
```diff
@@ -0,0 +1,116 @@
...
+
+func NewSyncClient(client client.RPC, log log.Logger, metrics caching.Metrics, config *SyncClientConfig) (*SyncClient, error) {
+	l2Client, err := NewL2Client(client, log, metrics, &config.L2ClientConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	return &SyncClient{
+		L2Client:         l2Client,
+		FetchUnsafeBlock: make(chan uint64),
```
tynes, line 49: This is an unbuffered channel so i think it'll block the driver event loop as its doing work over here, it would be nice to have metrics on how long it takes to fetch each block

op-node/rollup/driver/state.go
```diff
@@ -442,3 +458,28 @@ type hashAndErrorChannel struct {
...
+			s.log.Info("Attempting to fetch missing payloads from backup RPC", "start", start, "end", end, "size", size)
+
+			// Attempt to fetch the missing payloads from the backup unsafe sync RPC concurrently.
+			// Concurrent requests are safe here due to the engine queue being a priority queue.
+			// TODO: Should enforce a max gap size to prevent spamming the backup RPC or being rate limited.
+			for blockNumber := start; blockNumber < end; blockNumber++ {
+				s.L2SyncCl.FetchUnsafeBlock <- blockNumber
+			}
+		} else {
+			s.log.Warn("No backup unsafe sync RPC configured, cannot fetch missing payloads!")
```
tynes, line 482: This log will trigger every 60s when running in normal mode, perhaps we only call this function if the client is not nil and sanity check in here that its not nil to prevent issues with refactoring?

tynes: Actually nevermind, its behind the size check

op-node/rollup/driver/state.go
clabby: Yeah- the size check is purely to gate the logs

op-node/rollup/driver/state.go
clabby: Re-assigned to 15 seconds.

op-node/rollup/driver/state.go
```diff
@@ -223,7 +234,12 @@ func (s *Driver) eventLoop() {
 				}
 			}
 			planSequencerAction() // schedule the next sequencer action to keep the sequencing looping
-		case payload := <-s.unsafeL2Payloads:
+		// TODO: Should this be lower-priority in the switch case?
+		case <-altSyncTicker.C:
+			// Check if there is a gap in the current unsafe payload queue. If there is, attempt to fetch
+			// missing payloads from the backup RPC (if it is configured).
+			s.checkForGapInUnsafeQueue(ctx)
+		case payload := <-s.UnsafeL2Payloads:
```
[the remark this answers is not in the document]
clabby: This is a great question- I don't believe so. There's also the possibility that it takes longer than 15 seconds to fill the gap, meaning that this method could also add duplicate payloads to the `UnsafeL2Payloads` channel (alongside P2P). @protolambda @trianglesphere, what do you think? Mark suggested turning off P2P while backup syncing is active, but that still leaves the possibility that backup sync takes longer than expected & adds duplicate blocks to the `PayloadsQueue`.

op-node/sources/sync_client.go
clabby: Buffered the channel at `128` and enforced a max gap between the unsafe head / payload queue tip to be `128` as well. Does this number seem reasonable @protolambda @trianglesphere? A max gap size of `128` accounts for roughly 4 1/4 minutes worth of L2 blocks.

clabby: Also, should this max gap size (as well as the buffered channel size) be configurable for people that are syncing from a backup RPC that they own and/or know won't rate limit them?

op-node/rollup/driver/state.go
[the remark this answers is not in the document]
trianglesphere: Do we have deduplication somewhere in the existing P2P stack? I guess it might be at the p2p edge. We should make sure that the unsafe payload queue can handle multiple / late unsafe blocks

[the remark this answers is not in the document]
trianglesphere: Switching off p2p sync while we are doing RPC sync seems really tricky & error prone

op-node/sources/sync_client.go
trianglesphere: why does the gap & channel buffer length need to be the same? Are you actually storing blocks inside the channel?

clabby: @trianglesphere The channel stores `uint64`s, which represent block numbers- Correct me if I'm wrong, but for a max gap of 128, the `FetchUnsafeBlock` channel will need to store 128 block numbers at max?

op-node/rollup/driver/state.go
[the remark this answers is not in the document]
clabby: > Do we have deduplication somewhere in the existing P2P stack? I guess it might be at the p2p edge. We should make sure that the unsafe payload queue can handle multiple / late unsafe blocks

Not sure, will look into this today!

op-node/rollup/driver/state.go
```diff
@@ -442,3 +460,29 @@ type hashAndErrorChannel struct {
...
+		if size <= 128 {
+			s.log.Info("Attempting to fetch missing payloads from backup RPC", "start", start, "end", end, "size", size)
+		} else {
+			s.log.Warn("Gap in payload queue tip and unsafe head too large. Skipping fetch from backup RPC", "start", start, "end", end, "size", size, "max", "128")
+		}
+
+		// Attempt to fetch the missing payloads from the backup unsafe sync RPC concurrently.
+		// Concurrent requests are safe here due to the engine queue being a priority queue.
+		for blockNumber := start; blockNumber < end; blockNumber++ {
+			s.L2SyncCl.FetchUnsafeBlock <- blockNumber
```
protolambda, line 485:
```suggestion
select {
case s.L2SyncCl.FetchUnsafeBlock <- blockNumber:
default:
    return // if channel is full, stop trying to insert more work
}
```

op-node/rollup/driver/state.go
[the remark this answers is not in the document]
ajsutton: libp2p deduplicates messages and has been configured with a seenMessageTTL that's long enough for it to remember a gossiped block for the entire period that it's considered valid gossip. However, that mechanism wouldn't (currently) know about any blocks you'll pulled via RPC sync you may get those duplicated.

op-node/sources/sync_client.go
```diff
@@ -0,0 +1,120 @@
...
+	// Signature validation is not necessary here since the backup RPC is trusted.
+	if _, ok := payload.CheckBlockHash(); !ok {
+		s.log.Warn("Received invalid payload from backup RPC; invalid block hash", "payload", payload.ID())
+		return
+	}
+
+	s.log.Info("Received unsafe payload from backup RPC", "payload", payload.ID())
+
+	// Send the retrieved payload to the `unsafeL2Payloads` channel.
+	if err = s.receivePayload(ctx, "ALT_RPC_SYNC", payload); err != nil {
```
tynes, line 114: Perhaps making `"ALT_RPC_SYNC"` a constant with a godoc would be helpful for understanding

protolambda (approved):
nice work

tynes (approved):

clabby:
@mergifyio requeue

PR #4808, opened 2023-01-31
op-node: ensure sequencer selects L1 origin within sequencer time drift and use conf depth util

protolambda:
**Description**

Subset of original #4758

As sequencer it didn't follow the safety-over-liveness priority, and ignored the sequencer conf depth to try and handle sequencer time drift.

This however is dangerous because of the recent verifier "fix" in (good fix, but incomplete):
- When the max time drift is exceeded and the next origin cannot be found, then old origin would be used, causing the batch to be dropped.
- But the sequencer still produces blocks after this, which can have later origins & include transactions again, building on top of the empty block with old origin.

This means that as a verifier you end up dropping the empty batch with old origin, causing a halt (of safe blocks) until the sequencer window forces an empty block to be created. *This already went wrong once due to a L1 RPC outage on Goerli*

And aside from halting as such, it risks adopting the next L1 origin differently than the sequencer, causing a reorg of the unsafe chain.

**Tests**

- Modified `TestOriginSelectorRespectsMaxSeqDrift` into `TestOriginSelectorStrictConfDepth` to enforce conf depth.
- Added `TestOriginSelectorHandlesLateL1Blocks` to cover the case when the L1 block time gap is larger than the sequencer time drift. It preserves the time invariant in favor of the timedrift invariant. **Note: the safe-head progress would be bricked after this (assuming L1 does not reorg out the gap)**, as one of the two invariants breaks. We need some sort of derivation change like in the original PR.

**Invariants**

- Ensure that the sequencer conf depth is maintained
- Ensure that the sequencer never produces a block that exceeds sequencer time drift w.r.t. selected origin.
  - The exception here is if the block *has* to exceed the sequencer time drift to maintain the `L2 timestamp >= L1 origin timestamp` invariant. This can happen if L1 blocks have a huge time gap, larger than the sequencer time drift. This was fixed in #4758 but will now be split into a different PR because it affects derivation. In such case I believe it's best to allow the sequencer to continue, but only until the L1 origin can be introduced without breaking the (L2 timestamp >= L1 origin timestamp` invariant, and disable tx inclusion by sequencer to avoid L2 activity that has to be submitted in such risky L1 situation (large L1 gaps = likely trouble including batch data). 

**Additional context**

See #4758 and #3861 (comments too) for additional context.

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/op-e2e/actions/l2_sequencer.go
+++ b/op-e2e/actions/l2_sequencer.go
@@ -18,11 +18,11 @@ type MockL1OriginSelector struct {
 	originOverride eth.L1BlockRef // override which origin gets picked
 }
 
-func (m *MockL1OriginSelector) FindL1Origin(ctx context.Context, l1Head eth.L1BlockRef, l2Head eth.L2BlockRef) (eth.L1BlockRef, error) {
+func (m *MockL1OriginSelector) FindL1Origin(ctx context.Context, l2Head eth.L2BlockRef) (eth.L1BlockRef, error) {
 	if m.originOverride != (eth.L1BlockRef{}) {
 		return m.originOverride, nil
 	}
-	return m.actual.FindL1Origin(ctx, l1Head, l2Head)
+	return m.actual.FindL1Origin(ctx, l2Head)
 }
 
 // L2Sequencer is an actor that functions like a rollup node,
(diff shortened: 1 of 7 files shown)


trianglesphere (approved):
nice works. Thanks for the fix.

PR #4809, opened 2023-01-31
op-node: handle when L1 time gap is larger than sequencer time drift

protolambda:
**Description**

Subset of original #4758.

Handle rare case when L1 block time gap is larger than the sequencer time drift  to not halt safe head.

**Tests**

See new unit tests to cover new sequencer time drift handling.

**Invariants**

- Ensure `L2 time >= L1 origin time` at all times
- Ensure max sequencer time drift is respected, but make exception to always preserve L1 Origin advancement and above invariant.

**Additional context**

See design doc.

Fix CLI-3354

### DIFF
--- a/op-node/rollup/derive/batches.go
+++ b/op-node/rollup/derive/batches.go
@@ -100,11 +100,31 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
 		return BatchDrop
 	}
 
-	// If we ran out of sequencer time drift, then we drop the batch and produce an empty batch instead,
-	// as the sequencer is not allowed to include anything past this point without moving to the next epoch.
+	// Check if we ran out of sequencer time drift
 	if max := batchOrigin.Time + cfg.MaxSequencerDrift; batch.Batch.Timestamp > max {
-		log.Warn("batch exceeded sequencer time drift, sequencer must adopt new L1 origin to include transactions again", "max_time", max)
-		return BatchDrop
+		if len(batch.Batch.Transactions) == 0 {
+			// If the sequencer is co-operating by producing an empty batch,
+			// then allow the batch if it was the right thing to do to maintain the L2 time >= L1 time invariant.
+			// We only check batches that do not advance the epoch, to ensure epoch advancement regardless of time drift is allowed.
+			if epoch.Number == batchOrigin.Number {
+				if len(l1Blocks) < 2 {
+					log.Info("without the next L1 origin we cannot determine yet if this empty batch that exceeds the time drift is still valid")
+					return BatchUndecided
+				}
+				nextOrigin := l1Blocks[1]
+				if batch.Batch.Timestamp >= nextOrigin.Time { // check if the next L1 origin could have been adopted
+					log.Info("batch exceeded sequencer time drift without adopting next origin, and next L1 origin would have been valid")
+					return BatchDrop
+				} else {
+					log.Info("continuing with empty batch before late L1 block to preserve L2 time invariant")
+				}
+			}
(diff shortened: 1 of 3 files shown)


trianglesphere:
Adding a comment at https://github.com/ethereum-optimism/optimism/blob/b1f28127e54330178d4063bb0b2599331ef320b1/op-node/rollup/derive/batch_queue.go#L236 and in the specs to clarify that we generate empty batches past the sequencer drift until the next L1 block if there is a large gap would be helpful

op-node/rollup/derive/batches.go
```diff
@@ -100,11 +100,31 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
-	// If we ran out of sequencer time drift, then we drop the batch and produce an empty batch instead,
-	// as the sequencer is not allowed to include anything past this point without moving to the next epoch.
+	// Check if we ran out of sequencer time drift
 	if max := batchOrigin.Time + cfg.MaxSequencerDrift; batch.Batch.Timestamp > max {
-		log.Warn("batch exceeded sequencer time drift, sequencer must adopt new L1 origin to include transactions again", "max_time", max)
-		return BatchDrop
+		if len(batch.Batch.Transactions) == 0 {
+			// If the sequencer is co-operating by producing an empty batch,
+			// then allow the batch if it was the right thing to do to maintain the L2 time >= L1 time invariant.
+			// We only check batches that do not advance the epoch, to ensure epoch advancement regardless of time drift is allowed.
```
trianglesphere, line 108: I think allowing for an advance without checking the timestamp might actually handle the case of short L1 block times. Specifically by doing this advance without checking the next origin time upon advancing the origin, we always advance the origin. If we had the situation where we advanced the origin, but then check the block after the advanced origin (index 2 of the L1blocks), we may incorrectly skip the new L1 origin (index 1 of the L1Blocks)

specs/derivation.md
```diff
@@ -598,10 +598,18 @@ Rules, in validation order:
 - `batch.epoch_num > epoch.number+1` -> `drop`: i.e. the L1 origin cannot change by more than one L1 block per L2 block.
 - `batch.epoch_hash != batch_origin.hash` -> `drop`: i.e. a batch must reference a canonical L1 origin,
   to prevent batches from being replayed onto unexpected L1 chains.
-- `batch.timestamp > batch_origin.time + max_sequencer_drift` -> `drop`: i.e. a batch that does not adopt the next L1
-  within time will be dropped, in favor of an empty batch that can advance the L1 origin. This enforces the max L2
-  timestamp rule.
 - `batch.timestamp < batch_origin.time` -> `drop`: enforce the min L2 timestamp rule.
+- `batch.timestamp > batch_origin.time + max_sequencer_drift`: enforce the L2 timestamp drift rule,
```
trianglesphere, line 602: Top part of derivation specs also need to be updated

specs/derivation.md
```diff
@@ -598,10 +598,18 @@ Rules, in validation order:
...
+  but with exceptions to preserve above min L2 timestamp invariant:
+  - `len(batch.transactions) == 0`:
+    - `epoch.number >= batch.epoch_num`:
```
trianglesphere, line 605:
```suggestion
    - `epoch.number == batch.epoch_num`:
```
Then need case for where the epoch is advanced by the batch.

protolambda:
Rebasing this now, to ensure it runs on top of latest derivation changes

protolambda:
will rebase to fix the lint timeout issue

mslipper (approved):
LGTM. Proto walked me through this live, and we reviewed together.

PR #4814, opened 2023-02-01
txmgr: Flatten + simplify

trianglesphere:
**Description**

This simplifies some of the logging + configuration of the transaction manager.

**Tests**

No new, but existing test suite passes after slight modifications to the calling conventions.


tynes:
Looks generally good so far!

tynes (approved):

op-service/txmgr/txmgr.go
```diff
@@ -238,56 +212,31 @@ func (m *SimpleTxManager) Send(
...
-	ctx context.Context,
-	backend ReceiptSource,
-	tx *types.Transaction,
-	queryInterval time.Duration,
-	numConfirmations uint64,
-	sendState *SendState,
-) (*types.Receipt, error) {
-
-	queryTicker := time.NewTicker(queryInterval)
+func (m *SimpleTxManager) waitMined(ctx context.Context, tx *types.Transaction, sendState *SendState) (*types.Receipt, error) {
```
tynes, line 217: This is probably desired to have as part of multiple implementations of the `TxManager` but since we don't have anything like ` MultiTxManager` right now this is fine

trianglesphere: We might need to sync offline about this, but I actually think we probably don't want multiple transaction managers & that waiting for a receipt is actually pretty core to how a transaction manager works.

PR #4815, opened 2023-02-01
:sparkles: Add some experimental attestation station utils

roninjin10:
Just creating a package to use as playground while designing attestation station sdk

PR #4816, opened 2023-02-01
:recycle: feat: make dockerfile more maintainable

roninjin10:
The pr stacked on this one broke the build because dockerfile is very fragile.

To fix make multistage build where first stage collects the package.json files
This makes it so we don't have to maintain this but still get caching benifits.

I copy pasted this pattern from my personal project https://github.com/roninjin10/stax/blob/main/Dockerfile and the gateway repo https://github.com/ethereum-optimism/gateway/blob/develop/Dockerfile


roninjin10:
* develop
  * **PR #4816** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4816" target="_blank">Graphite</a>  👈
    * **PR #4815** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4815" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4816?utm_source=stack-comment).

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1014,6 +1014,7 @@ export class CrossChainMessenger {
    * Queries the OptimismPortal contract's `provenWithdrawals` mapping
    * for a ProvenWithdrawal that matches the passed withdrawalHash
    *
+   * @bedrock
```
roninjin10, line 1017: changed this to force everything to build

ops/docker/Dockerfile.packages
```diff
@@ -33,32 +79,32 @@ COPY --from=foundry /usr/local/bin/cast /usr/local/bin/cast
 # note: this approach can be a bit unhandy to maintain, but it allows
 # us to cache the installation steps
```
roninjin10, line 80:
```suggestion
```

ops/docker/Dockerfile.packages
```diff
@@ -33,32 +79,32 @@ COPY --from=foundry /usr/local/bin/cast /usr/local/bin/cast
 # note: this approach can be a bit unhandy to maintain, but it allows
```
roninjin10, line 79:
```suggestion
```

ops/docker/Dockerfile.packages
```diff
@@ -1,10 +1,56 @@
 # This Dockerfile builds all the dependencies needed by the monorepo, and should
 # be used to build any of the follow-on services
 #
-# ### BASE: Install deps
+
+# ░██████╗████████╗░█████╗░░██████╗░███████╗  ░█████╗░
```
tynes, line 5: I have a preference for removing the ascii art because it makes it more difficult to see what is in the file on a single screen

roninjin10: The giant space between stages is basically a workaround to not having modules in being able to put dockerfiles together

tynes:
Generally in favor of this but going to defer to @zhwrd

ops/docker/Dockerfile.packages
```diff
@@ -1,10 +1,37 @@
...
+# development. The only reason we need a separate stage just for collecting the
+# dependency manifests is that Docker's `COPY` command still does not allow
+# copying based on a glob pattern (see this GitHub issue for more details
+# https://github.com/moby/moby/issues/15858). Being able to copy only manifests
+# into stage 1 (the `COPY --from=manifests` statement) is important to maximize
+# Docker build cache hit rate. `alpine` is chosen as the base image for the
+# first stage because it's the smallest image that have access to the `cp
+# --parents -t` command (by installing the `coreutils` package).
+FROM alpine:3.16 as manifests
+RUN apk add coreutils
```
roninjin10, line 17: When/if we switch to pnpm pnpm has a really nice way of accomplishing this where you install the node modules directly from only lockfile and not package.jsons

nickbalestra (approved):
Thanks for this Will! 

NIT: in the future I would suggest to try to separate such PR into more focused ones, in this case for example:
- 1st PR to submit the fix that is blocking the rest of the stacked work (if possible)
- 2nd PR to add improving of of maintainability of the Dockerfile

They are very tight related though in this case and if nobody has any concern, I am ok with this, so approving it! I'll leave it to you Will if you feel that would still be beneficial to break it apart

zhwrd (approved):

roninjin10:
Nice I somehow missed that this merged

PR #4820, opened 2023-02-01
ctp: remove final hh tests

tynes:
**Description**

This commit gets the foundry tests up to parity for the `AssetReceiver` and `Transactor` and then deletes the hardhat tests. Now this package only has foundry based tests which makes it much easier to standardize longer term with a single way of developing smart contracts in the monorepo.


packages/contracts-periphery/contracts/foundry-tests/AssetReceiver.t.sol
```diff
@@ -17,6 +17,21 @@ contract AssetReceiver_Initializer is Test {
...
+        uint256 amount
+    );
+    event WithdrewERC721(
+        address indexed withdrawer,
+        address indexed recipient,
+        address indexed asset,
+        uint256 id
+    );
+
     function _setUp() public {
```
maurelian, line 35: I think we should use this PLR as an opportunity to start moving away from this pattern.
I.e. if we make the following change, then we can delete the other `setUp` functions (which are just thin wrappers) from the child contracts: 
```suggestion
    function setUp() public {
```

maurelian (approved):
Some nice-to-have suggestions, but LGTM as is.

packages/contracts-periphery/contracts/foundry-tests/AssetReceiver.t.sol
tynes: I have cleaned this up, good call

packages/contracts-periphery/contracts/foundry-tests/AssetReceiver.t.sol
```diff
@@ -52,6 +67,8 @@ contract AssetReceiverTest is AssetReceiver_Initializer {
         // Check that contract balance is 0 initially
         assertEq(address(assetReceiver).balance, 0);
 
+        vm.expectEmit(true, true, true, true);
```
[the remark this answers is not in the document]
tynes: Addressed this, good call

tynes:
I've addressed all comments here, should be good to merge pending passing tests

tynes:
depcheck fail

```
Unused devDependencies
* @types/chai
* @types/mocha
* mocha

Exited with code exit status 255

```

PR #4825, opened 2023-02-02
feat(mon): Add withdrawal monitor to chain-mon

maurelian:
**Description**

Adds a new service (`wd-mon`) to the `chain-mon` package.

**Tests**

I've added no tests, but tested against the Goerli network. The functionality used is very basic.

I would like to add an integration test in a future PR.

### DIFF
--- a/packages/chain-mon/src/wd-mon/service.ts
+++ b/packages/chain-mon/src/wd-mon/service.ts
@@ -1,0 +1,232 @@
+import {
+  BaseServiceV2,
+  StandardOptions,
+  ExpressRouter,
+  Gauge,
+  validators,
+  waitForProvider,
+} from '@eth-optimism/common-ts'
+import { CrossChainMessenger } from '@eth-optimism/sdk'
+import { getChainId, sleep } from '@eth-optimism/core-utils'
+import { Provider } from '@ethersproject/abstract-provider'
+import { Event } from 'ethers'
+import dateformat from 'dateformat'
+
+import { version } from '../../package.json'
+
+type Options = {
+  l1RpcProvider: Provider
+  l2RpcProvider: Provider
+  startBlockNumber: number
+  sleepTimeMs: number
+}
+
+type Metrics = {
+  withdrawalsValidated: Gauge
+  isDetectingForgeries: Gauge
+  nodeConnectionFailures: Gauge
+}
+
+type State = {
+  messenger: CrossChainMessenger
+  highestUncheckedBlockNumber: number
+  finalizationWindow: number
+  forgeryDetected: boolean
+}
+
+export class WithdrawalMonitor extends BaseServiceV2<Options, Metrics, State> {
+  constructor(options?: Partial<Options & StandardOptions>) {
+    super({
+      version,
+      name: 'two-step-monitor',
+      loop: true,
+      options: {
+        loopIntervalMs: 1000,
+        ...options,
+      },
+      optionsSpec: {
+        l1RpcProvider: {
+          validator: validators.provider,
+          desc: 'Provider for interacting with L1',
+        },
+        l2RpcProvider: {
+          validator: validators.provider,
+          desc: 'Provider for interacting with L2',
+        },
+        startBlockNumber: {
+          validator: validators.num,
+          default: -1,
+          desc: 'L1 block number to start checking from',
+          public: true,
+        },
+        sleepTimeMs: {
+          validator: validators.num,
+          default: 15000,
+          desc: 'Time in ms to sleep when waiting for a node',
+          public: true,
+        },
+      },
+      metricsSpec: {
+        withdrawalsValidated: {
+          type: Gauge,
+          desc: 'Latest L1 Block (checked and known)',
+          labels: ['type'],
+        },
+        isDetectingForgeries: {
+          type: Gauge,
+          desc: '0 if state is ok. 1 or more if forged withdrawals are detected.',
+        },
+        nodeConnectionFailures: {
+          type: Gauge,
+          desc: 'Number of times node connection has failed',
+          labels: ['layer', 'section'],
+        },
+      },
+    })
+  }
+
+  async init(): Promise<void> {
+    // Connect to L1.
+    await waitForProvider(this.options.l1RpcProvider, {
+      logger: this.logger,
+      name: 'L1',
+    })
+
+    // Connect to L2.
+    await waitForProvider(this.options.l2RpcProvider, {
+      logger: this.logger,
+      name: 'L2',
+    })
+
+    this.state.messenger = new CrossChainMessenger({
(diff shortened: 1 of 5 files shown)


packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -0,0 +1,231 @@
...
+    try {
+      latestL1BlockNumber = await this.options.l1RpcProvider.getBlockNumber()
+    } catch (err) {
+      this.logger.error(`got error when connecting to node`, {
+        error: err,
+        node: 'l1',
+        section: 'getBlockNumber',
+      })
+      this.metrics.nodeConnectionFailures.inc({
+        layer: 'l1',
```
tynes, line 145: `layer: 'l1'` feels a bit redundant, maybe would be better if its chainid to make it more generic in the case of L3 or L2<>L2

packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    for (const proofEvent of proofEvents) {
+      const exists =
+        await this.state.messenger.contracts.l2.BedrockMessagePasser.sentMessages(
+          proofEvent.args.withdrawalHash
+        )
+      const provenAt = `${dateformat(
+        new Date(
+          (await (
```
tynes, line 196: Nesting `await` is an antipattern

packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    this.logger.info(`checking recent blocks`, {
+      fromBlockNumber: this.state.highestUncheckedBlockNumber,
+      toBlockNumber: latestL1BlockNumber,
+    })
+
+    // Perform the check
+    let proofEvents: Event[]
+    try {
```
tynes, line 166: Double check that this will work, but perhaps `try/catch/finally` and have the `sleep` in the finally so that its impossible to not `sleep` in any branching logic that ends early

maurelian: If the `try` block succeeds, why do we want to sleep?

packages/chain-mon/src/wd-mon/service.ts
maurelian: fixed.

packages/chain-mon/src/wd-mon/service.ts
maurelian: I changed it to [`chainId`](https://github.com/ethereum-optimism/optimism/blob/0515a78417e5e37b4d800c7f2f65995b7ae837e7/packages/chain-mon/src/wd-mon/service.ts#L145). 
There are many other references to L1/L2 though, including the SDK, so I don't think it'd make sense to remove it entirely from this code.

packages/chain-mon/src/wd-mon/service.ts
tynes: I thought we'd always want to sleep so that things are on a polling interval basically, but maybe i misunderstand how the baseservice is architected

packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    this.logger.info(`starting L1 block height`, {
+      startBlockNumber: this.state.highestUncheckedBlockNumber,
+    })
+  }
+
+  // K8s healthcheck
+  async routes(router: ExpressRouter): Promise<void> {
+    router.get('/healthz', async (req, res) => {
```
tynes, line 126: This `router.get` feels like it could live in the common package and be imported into all the monitoring services, but i dont think you need to do that now

packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -0,0 +1,232 @@
...
+    })
+
+    // Perform the check
+    let proofEvents: Event[]
+    try {
+      // The query includes events in the blockNumbers given as the last two arguments
+      proofEvents =
+        await this.state.messenger.contracts.l1.OptimismPortal.queryFilter(
+          this.state.messenger.contracts.l1.OptimismPortal.filters.WithdrawalProven(),
+          this.state.highestUncheckedBlockNumber,
```
tynes, line 171: This can error if the range is too large

packages/chain-mon/src/wd-mon/service.ts
maurelian: Yeah, the main loop time is set here: 

https://github.com/ethereum-optimism/optimism/blob/0515a78417e5e37b4d800c7f2f65995b7ae837e7/packages/chain-mon/src/wd-mon/service.ts#L145

This is just a longer delay.

maurelian: Yeah, the main loop time is set here: 

https://github.com/ethereum-optimism/optimism/blob/0515a78417e5e37b4d800c7f2f65995b7ae837e7/packages/chain-mon/src/wd-mon/service.ts#L44
This is just a longer delay.

packages/chain-mon/src/wd-mon/service.ts
maurelian: Good call, it is copy pasta from another service.

@smartcontracts does it make sense to put into BaseServiceV2? Could be made optional it that's preferrable.

packages/chain-mon/src/wd-mon/service.ts
maurelian: Hmmm... Good call out. Perhaps in the event of an error we try a progressively smaller interval?

What do you suggest?

tynes: I think that if we are running this against our own nodes, this is fine for now. To make this portable, i suggest that we write a utility function that is able to take a `ethers.Contract.queryFilter` function, a filter, a start, an end and an interval and then it basically makes calls the size of the interval and concats them together so it looks like a single call from the pov of the consumer of this util

tynes: To be honest this util should be upstreamed into ethers, if there was an "options" param to `queryFilter`. But for now we would be able to use this util across all TS services and have a config param that states how large the interval is and its infinite if not passed in

maurelian: Agree it's surprising there isn't a standard way to handle this. 

> I think that if we are running this against our own nodes, this is fine for now.

SGTM. 

I believe other comments have all been addressed, is there anything outstanding required for approval?

packages/chain-mon/src/wd-mon/service.ts
mslipper: Let's do this as another PR.

mslipper (approved):

PR #4830, opened 2023-02-02
txmgr: Simplify resubmit logic and scaffolding

trianglesphere:
**Description**

This makes the transaction manager responsible for determining the gas price. It also provides the ability to easily add a check if the gas price bump is correct or not.

This fixes the bug where the proposer fails when it redoes gas estimation. Gas re-estimation is no longer done. This is fine because we have a minimal set of different transactions that use a consistent amount of gas.

**Metadata**
- Fixes CLI-3233

### DIFF
--- a/op-service/txmgr/txmgr.go
+++ b/op-service/txmgr/txmgr.go
@@ -2,14 +2,17 @@ package txmgr
 
 import (
 	"context"
+	"errors"
 	"math/big"
-	"strings"
 	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/txpool"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
+
+	opcrypto "github.com/ethereum-optimism/optimism/op-service/crypto"
 )
 
 // UpdateGasPriceSendTxFunc defines a function signature for publishing a
@@ -40,6 +43,10 @@ type Config struct {
 	// are required to give up on a tx at a particular nonce without receiving
 	// confirmation.
 	SafeAbortNonceTooLowCount uint64
+
+	// Signer is used to sign transactions when the gas price is increased.
+	Signer opcrypto.SignerFn
+	From   common.Address
 }
 
 // TxManager is an interface that allows callers to reliably publish txs,
@@ -50,15 +57,15 @@ type TxManager interface {
 	// until an invocation of sendTx returns (called with differing gas
 	// prices). The method may be canceled using the passed context.
 	//
+	// The initial transaction MUST be signed & ready to submit.
+	//
 	// NOTE: Send should be called by AT MOST one caller at a time.
-	Send(ctx context.Context, updateGasPrice UpdateGasPriceFunc, sendTxn SendTransactionFunc) (*types.Receipt, error)
+	Send(ctx context.Context, tx *types.Transaction) (*types.Receipt, error)
 }
 
-// ReceiptSource is a minimal function signature used to detect the confirmation
-// of published txs.
-//
-// NOTE: This is a subset of bind.DeployBackend.
-type ReceiptSource interface {
+// ETHBackend is the set of methods that the transaction manager uses to resubmit gas & determine
+// when transactions are included on L1.
+type ETHBackend interface {
 	// BlockNumber returns the most recent block number.
 	BlockNumber(ctx context.Context) (uint64, error)
 
@@ -66,6 +73,14 @@ type ReceiptSource interface {
 	// txHash. If lookup does not fail, but the transaction is not found,
 	// nil should be returned for both values.
 	TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error)
+
+	// SendTransaction submits a signed transaction to L1.
+	SendTransaction(ctx context.Context, tx *types.Transaction) error
+
+	// These functions are used to estimate what the basefee & priority fee should be set to.
+	// TODO(CLI-3318): Maybe need a generic interface to support different RPC providers
+	HeaderByNumber(ctx context.Context, number *big.Int) (*types.Header, error)
+	SuggestGasTipCap(ctx context.Context) (*big.Int, error)
 }
 
 // SimpleTxManager is a implementation of TxManager that performs linear fee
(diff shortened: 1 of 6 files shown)


op-service/txmgr/txmgr.go
```diff
@@ -74,12 +87,55 @@ type SimpleTxManager struct {
 	Config // embed the config directly
 	name   string
 
-	backend ReceiptSource
+	backend L1Connection
```
tynes, line 90: Why is this an L1 connection? It could be used to send txs to L2? It really should be a generic backend?

op-service/txmgr/txmgr.go
```diff
@@ -74,12 +87,55 @@ type SimpleTxManager struct {
...
+// error will be returned.
+// We do not re-estimate the amount of gas used because for some stateful transactions (like output proposals) the
+// act of including the transaction renders the repeat of the transaction invalid.
+func (m *SimpleTxManager) IncreaseGasPrice(ctx context.Context, tx *types.Transaction) (*types.Transaction, error) {
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	var gasTipCap, gasFeeCap *big.Int
+
+	if tip, err := m.backend.SuggestGasTipCap(ctx); err != nil {
```
tynes, line 105: Are we sure all clients support this RPC?

op-service/txmgr/txmgr.go
trianglesphere: You're right, it could be a generic "eth backend". I just extended the name because it is no longer just receipts.

op-service/txmgr/txmgr.go
trianglesphere: I'm reasonably sure. I can check RPC provider docs though

trianglesphere: Ok, here's the support matrix. FYI internally we previously used this RPC method

eth_FeeHistory: geth, alchemy, infura
eth_maxPriorityFee (SuggestGasTip): geth, alchemy
quicknode doesn't seem to support either based on my reading of it.

tynes: we shouldn't err in the case when the rpc is not present

op-service/txmgr/txmgr.go
```diff
@@ -196,9 +245,15 @@ func (m *SimpleTxManager) Send(ctx context.Context, updateGasPrice UpdateGasPric
 				continue
 			}
 
-			// Submit and wait for the bumped traction to confirm.
+			// Increase the gas price & submit the new transaction
+			newTx, err := m.IncreaseGasPrice(ctx, tx)
+			if err != nil {
+				m.l.Error("Failed to increase the gas price for the tx", "err", err)
```
tynes, line 251: In the error case we shouldn't try to resubmit the same tx?

op-service/txmgr/txmgr.go
tynes: types are found here: https://github.com/ethereum/go-ethereum/blob/d0a4989a8def7e6bad182d1513e8d4a093c1672d/accounts/abi/bind/backend.go

tynes:
We prob want optional metrics inside of here eventually

op-service/txmgr/txmgr.go
```diff
@@ -196,9 +245,15 @@ func (m *SimpleTxManager) Send(ctx context.Context, updateGasPrice UpdateGasPric
...
+			}
+			// Save the tx so we know it's gas price.
+			tx = newTx
```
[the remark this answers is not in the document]
trianglesphere: Yea it is. I had a local fix that I haven't pushed yet

op-service/txmgr/txmgr.go
trianglesphere: I lean to no. I guess it's technically fine for the global mempool, but I don't want us to spam the same transactions over & over.

op-service/txmgr/txmgr.go
trianglesphere: nice. We actually only need a subset of that so I'll probably rename `L1Connection` rather than import that set.

op-service/txmgr/txmgr.go
trianglesphere: Ideally yes. I just realized that the place we will need to fix this is in `SuggestGasTipCap` or whatever the interface we define is.

op-service/txmgr/txmgr.go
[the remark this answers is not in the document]
trianglesphere: fixed. We `continue` on error to skip this saving.

op-service/txmgr/txmgr.go
trianglesphere: Going to resolve this convo because it is not a regression, but I have created a ticket for it at https://linear.app/optimism/issue/CLI-3318/support-multiple-l1-rpc-providers-with-the-ethconnection-client-in-the

op-service/txmgr/txmgr.go
trianglesphere: hmm, ok, gonna resubmit it

trianglesphere:
> We prob want optional metrics inside of here eventually

Ticketed here: https://linear.app/optimism/issue/CLI-3319/add-metrics-to-the-transaction-manager

trianglesphere:
Hive L1Ops has been failing for a bit. P2P failure seems newer & potentially related to the tx manager changes

op-service/txmgr/txmgr.go
```diff
@@ -74,12 +89,55 @@ type SimpleTxManager struct {
...
+// error will be returned.
+// We do not re-estimate the amount of gas used because for some stateful transactions (like output proposals) the
+// act of including the transaction renders the repeat of the transaction invalid.
+func (m *SimpleTxManager) IncreaseGasPrice(ctx context.Context, tx *types.Transaction) (*types.Transaction, error) {
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	var gasTipCap, gasFeeCap *big.Int
+
+	if tip, err := m.backend.SuggestGasTipCap(ctx); err != nil {
```
tynes, line 107: Not sure if this should merge if the Todo isn't addressed and we handle this RPC not being available. Would be a bad time in prod to need to panic roll over endpoints

tynes: I believe ethclient in geth handles this

op-service/txmgr/txmgr_test.go
```diff
@@ -336,13 +366,11 @@ func TestTxMgrConfirmsMinGasPriceAfterBumping(t *testing.T) {
```
sebastianst, line 366: Does this existing test `TestTxMgrConfirmsMinGasPriceAfterBumping` test the added functionality of the `SimpleTxManager`? If not, we should have an additional test that explicitly covers the branch where at least one gas price increase happens.

sebastianst (dismissed):
generally lgtm; one question about changing the behavior of `Send` implementations (to reduce code duplication) and one about a possibly missing test case.

op-service/txmgr/txmgr.go
```diff
@@ -97,8 +155,12 @@ func NewSimpleTxManager(name string, l log.Logger, cfg Config, backend ReceiptSo
 // invocation of sendTx returns (called with differing gas prices). The method
 // may be canceled using the passed context.
 //
+// The initially supplied transaction must be signed, have gas estimation done, and have a reasonable gas fee.
+// When the transaction is resubmitted the tx manager will re-sign the transaction at a different gas pricing
+// but retain the gas used, the nonce, and the data.
+//
 // NOTE: Send should be called by AT MOST one caller at a time.
-func (m *SimpleTxManager) Send(ctx context.Context, updateGasPrice UpdateGasPriceFunc, sendTx SendTransactionFunc) (*types.Receipt, error) {
+func (m *SimpleTxManager) Send(ctx context.Context, tx *types.Transaction) (*types.Receipt, error) {
```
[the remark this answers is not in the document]
trianglesphere: I would like to do this, but it's harder than you think because we rely on the `ContractBackend` in the `bind` which wants to manage the full lifecycle & relies on having a signer + doing gas estimation.

op-service/txmgr/txmgr_test.go
trianglesphere: Yes, the unit tests fail if you do not increase the gas price in the resubmit loop.

op-service/txmgr/txmgr.go
```diff
@@ -74,12 +89,55 @@ type SimpleTxManager struct {
...
+	} else {
+		gasTipCap = tip
+	}
+
+	if head, err := m.backend.HeaderByNumber(ctx, nil); err != nil {
+		return nil, err
+	} else if head.BaseFee == nil {
+		return nil, errors.New("txmgr does not support pre-london blocks that do not have a basefee")
+	} else {
+		gasFeeCap = CalcGasFeeCap(head.BaseFee, gasTipCap)
```
protolambda, line 120: Worth a comment maybe: `CalcGasFeeCap` ensures the `gasFeeCap` is always covering the `gasTipCap`, and the tx fee params are thus always valid.

protolambda (approved):

op-service/txmgr/txmgr.go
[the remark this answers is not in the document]
sebastianst: We discussed that this is gonna be resolved in a future pr.

PR #4831, opened 2023-02-02
txmgr: Don't submit underpriced replacements

trianglesphere:
**Description**

This prevents the transaction manager from resubmiting transactions that do not have a price increase that is large enough to pass geth's checks. The other option is to increase the gas price bump, but because the gas price has not changed much in the intervening period, it seems better to leave the currently existing transaction in the mempool.

**Tests**

Added simple unit tests.

**Metadata**
- Fixes CLI-2630


op-service/txmgr/txmgr.go
```diff
@@ -118,7 +120,22 @@ func (m *SimpleTxManager) IncreaseGasPrice(ctx context.Context, tx *types.Transa
 		gasFeeCap = CalcGasFeeCap(head.BaseFee, gasTipCap)
 	}
```
tynes, line 122: Thinking of possible options:
- we are below the basefee and need to bump to get above the basefee
- we are above the basefee and there is competition with the tip so we need to bump that
- we are below the basefee and going above the basefee means its not a 15% increase for RBF

I feel like we shouldn't error here and instead just ensure that we bump enough

op-service/txmgr/txmgr.go
```diff
@@ -74,12 +89,70 @@ type SimpleTxManager struct {
...
+	backend L1Connection
 	l       log.Logger
 }
 
+// IncreaseGasPrice takes the previous transaction & potentially clones then signs it with a higher tip.
+// If the basefee + priority fee did not increase by a minimum percent (geth's replacement percent) an
+// error will be returned.
+// We do not re-estimate the amount of gas used because for some stateful transactions (like output proposals) the
+// act of including the transaction renders the repeat of the transaction invalid.
+func (m *SimpleTxManager) IncreaseGasPrice(ctx context.Context, tx *types.Transaction) (*types.Transaction, error) {
```
tynes, line 101: i think this should just bump 15% if its over basefee otherwise bump to basefee

op-service/txmgr/txmgr.go
```diff
@@ -50,22 +58,29 @@ type TxManager interface {
...
 }
 
-// ReceiptSource is a minimal function signature used to detect the confirmation
-// of published txs.
-//
-// NOTE: This is a subset of bind.DeployBackend.
-type ReceiptSource interface {
+// L1Connection is the set of methods that the transaction manager uses to resubmit gas & determine
+// when transactions are included on L1.
+type L1Connection interface {
```
tynes, line 69: this should be able to be a geth backend and we can delete this type

op-service/txmgr/txmgr.go
tynes: This should be added to the `TxManager` interface

op-service/txmgr/txmgr.go
```diff
@@ -196,9 +262,15 @@ func (m *SimpleTxManager) Send(ctx context.Context, updateGasPrice UpdateGasPric
 				continue
```
tynes, line 262: Don't we want to bump the gas price if `IsWaitingForConfirmation`? This feels like we want to wait until a threshold passes then increase the gas price if still above basefee or always increase if below the basefee.

Or we want an interface to allow for different replacement strategies. Prob over optimizing for now

op-service/txmgr/txmgr.go
trianglesphere: Switched to requiring the tip + feecap to both be bumped by a minimum of 15%.

trianglesphere:
txmgr tests failing because it doesn't expect the automatic price bumps. Debugging now.

op-service/txmgr/txmgr.go
```diff
@@ -104,6 +108,10 @@ func (m *SimpleTxManager) IncreaseGasPrice(ctx context.Context, tx *types.Transa
 
 	var gasTipCap, gasFeeCap *big.Int
 
+	// new = old * (100 + priceBump) / 100
+	a := big.NewInt(100 + priceBump)
+	b := big.NewInt(100)
+
```
tynes, line 114: These technically don't need to be allocated on every invocation of this function, they could be `var`s at the top of the file

op-service/txmgr/txmgr.go
```diff
@@ -112,15 +120,30 @@ func (m *SimpleTxManager) IncreaseGasPrice(ctx context.Context, tx *types.Transa
 		gasTipCap = tip
 	}
 
+	// Enforce a min priceBump on the tip. Do this before the feeCap is calculated
+	thresholdTip := new(big.Int).Mul(a, tx.GasTipCap())
+	thresholdTip = thresholdTip.Div(thresholdTip, b)
+	if thresholdTip.Cmp(gasTipCap) > 0 {
+		gasTipCap = thresholdTip
+		m.l.Debug("Override the tip to enforce a price bump")
```
tynes, line 128: Useful to include old and new in this logline? Same as below

op-service/txmgr/txmgr_test.go
```diff
@@ -602,3 +605,82 @@ func TestWaitMinedReturnsReceiptAfterFailure(t *testing.T) {
...
+	newTx, err := mgr.IncreaseGasPrice(ctx, tx)
+	require.NoError(t, err)
+	require.True(t, newTx.GasFeeCap().Cmp(tx.GasFeeCap()) > 0, "new tx fee cap must be larger")
+	require.True(t, newTx.GasTipCap().Cmp(tx.GasTipCap()) > 0, "new tx tip must be larger")
+}
+
+// TestIncreaseGasPriceUseLargeIncreas asserts that if the suggest gas tip
+// returned from L1 is much larger than the required price bump the L1 value
+// is used instead of the price bump
+func TestIncreaseGasPriceUseLargeIncreas(t *testing.T) {
```
tynes, line 651: typo `increase`

tynes:
Implementation looks good to me

op-service/txmgr/txmgr.go
```diff
@@ -112,15 +120,32 @@ func (m *SimpleTxManager) IncreaseGasPrice(ctx context.Context, tx *types.Transa
 		gasTipCap = tip
 	}
 
+	// new = old * (100 + priceBump) / 100
```
protolambda, line 123: The `IncreaseGasPrice` method is called in a loop, with a 1 second interval in testing, and an unchecked but explicitly configured flag/env var in prod. To not burn ourselves when we accidentally set this interval too low, we should try to back-off the gas increases with longer and longer intervals, to avoid rapid increases. Also see other related discussion in Slack.


Raising this now because the minimum gas price increase makes the increases go higher exponentially, it doesn't level off anymore.

protolambda (changes requested):

op-service/txmgr/txmgr.go
trianglesphere: Well the other option is to not bump the gas price at all, but return the previous transaction & resubmit it. My sense is that @tynes wanted to always do the price bump. I guess there's also the difference between the new price is less than the old price & the new price is in between the old price in the threshold - in this case we may want to bump, but if it's below the old price, we probably want to not bump.

trianglesphere: Ok, I added a commit to fix this.

protolambda (approved):
LGTM, nice testing

protolambda:
I think the change that fixes the unbounded exponential case is affecting the `TestIncreaseGasPriceEnforcesMinBump` test

trianglesphere:
> I think the change that fixes the unbounded exponential case is affecting the `TestIncreaseGasPriceEnforcesMinBump` test

Yea, I force pushed a fix which changed the setup gas price values for that test. I needed to increase them to be able to easily put the new values to be in the 15% range.

PR #4833, opened 2023-02-03
contracts-bedrock: add gov token to predeploys contract

tynes:
**Description**

The `GovernanceToken` never had a field in the `Predeploys` contract, this commit adds it.


clabby (approved):
Compliant with [predeploys spec](https://github.com/ethereum-optimism/optimism/blob/develop/specs/predeploys.md#overview), LGTM.

PR #4834, opened 2023-02-03
fix(spec): Withdrawal calls are not forwarded during the proof verification call

ajsutton:
**Description**

The spec incorrectly stated that the call was forwarded at both the proof verification call and during the finalization call. They are actually only forwarded once the finalisation call completes, so remove the mention of forwarding from the proof verification step.


maurelian (approved):
Good catch

PR #4838, opened 2023-02-03
specs: Fix typo & link in withdrawals section

sebastianst:
* missing "to"
* broken link converted to permalink


sebastianst:
Aware of the markdown linter complaining about the long line length. It was over-long before and according to how I understnad the [rule](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md013---line-length), this shouldn't be a problem because there's no whitespace after column 120
> This rule has an exception where there is no whitespace beyond the configured line length. This allows you to still include items such as long URLs without being forced to break them in the middle.

tynes (approved):

tynes:
Good catch

PR #4839, opened 2023-02-03
feat(ci): Add esm exports to sdk

roninjin10:
- Add tsup to build esm modules for packages sdk depends on.  
- Update the package exports to include extension to reduce ambiguity
- In general the old build is unchnaged.   Typescript generates the cjs exports as before we just more explicitly target the build outputs now
- We also build esm and include those imports.   This maximizes compatibility with all build tools
- As a bonus this will fix source maps

We should add this to all packages

image

Failing check fails on develop
image

Other failing checks also fail on develop https://github.com/ethereum-optimism/optimism/pull/4848


roninjin10:
* develop
  * **PR #4839** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4839" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4839?utm_source=stack-comment).

packages/sdk/package.json
```diff
@@ -2,22 +2,24 @@
   "name": "@eth-optimism/sdk",
   "version": "1.10.1",
   "description": "[Optimism] Tools for working with Optimism",
-  "main": "dist/index",
-  "types": "dist/index",
```
roninjin10, line 6: I believe not including .js here was the root cause

tynes:
Perhaps we could merge this after https://github.com/ethereum-optimism/optimism/pull/4778?

packages/sdk/tsup.config.ts
```diff
@@ -0,0 +1,26 @@
...
+  // will create a .js file for commonjs and a .cjs file for esm
+  format: ['esm', 'cjs'],
+  // don't generate .d.ts files.  This is default but being explicit
+  // note this means we need to generate our own .d.ts files
+  // this means we need to run the typechecker as a linter
+  // which is a general best practice anyways.  It also means
+  // we must use our .ts files as "types" in package.json
+  // this is better dx since it means the types take you to
+  // the source code instead of just the .d.ts file in editors
+  noExternal: [/^.*$/],
```
roninjin10, line 20: !!!! This was a hack to test something out!!!!   We must remove this noExternal line or bad things will happen!!!!

roninjin10:
@tynes agreed this one is trying to fix a P0 bug for gateway but I want to look really quick on monday to see if I can do this in a simpler so this pr has less urgency to land

nickbalestra:
> SDK only exports cjs and something about how we formatted the package.json makes some build tools unable to recognize it. 

Trying to understand better, what does 'something' refer to in this line, can you point me to it? 
can we not use the module as a cjs in the gateway, what is blocking us from that, as we have been using it in that way till now I imagine.

My question is that perhaps this PR try to do two different things and we could perhaps takle first the fix that is blocking deployments on the gateway (P0) and then as a follow-up to address P1s like switching to a more modern build system, source maps ecc. 

@tynes what you suggest?

roninjin10:
Contracts needs to not bundle

roninjin10:
(btw not testing locally until ci passes then I'm testing with gateway and doing some extra due dillegence.  So it's expected this will fail a bunch more times

tsc still needs to emit for contracts is the next one

roninjin10:
removed bundling we are now at the point where we are going to be ready to go soon most likely

tsconfig.json
```diff
@@ -12,12 +12,7 @@
...
-    "typeRoots": [
-      "node_modules/@types"
-    ]
+    "typeRoots": ["node_modules/@types"]
   },
-  "exclude": [
-    "node_modules",
-    "dist"
-  ]
+  "exclude": ["node_modules", "dist"]
```
roninjin10, line 17: this file didn't change this is prettier.  Revert before landing

roninjin10:
@nickbalestra @tynes this change is additive not subtractive or changing.   The old build still exists with no changes.   The only difference is we are adding esm outputs

packages/contracts/tsconfig.json
```diff
@@ -4,7 +4,5 @@
     "rootDir": "./src",
     "outDir": "./dist"
   },
-  "include": [
-    "src/**/*"
-  ]
+  "include": ["src/**/*"]
```
roninjin10, line 7: unchanged prettier revert

packages/sdk/tsconfig.json
```diff
@@ -4,7 +4,5 @@
     "rootDir": "./src",
     "outDir": "./dist"
   },
-  "include": [
-    "src/**/*"
-  ]
+  "include": ["src/**/*"]
```
roninjin10, line 7: unchcnaged prettier revert

packages/core-utils/tsconfig.json
```diff
@@ -2,12 +2,8 @@
...
-    "outDir": "./dist"
-  },
-  "compilerOptions": {
+    "outDir": "./dist",
     "typeRoots": ["node_modules/@types"]
   },
-  "include": [
-    "src/**/*"
-  ]
+  "include": ["src/**/*"]
```
roninjin10, line 8: This change is real we had 2 dcompiler options somehow

packages/sdk/package.json
```diff
@@ -2,22 +2,25 @@
...
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.ts",
   "files": [
     "dist/*",
     "src/*"
   ],
   "scripts": {
     "all": "yarn clean && yarn build && yarn test && yarn lint:fix && yarn lint",
-    "build": "tsc -p tsconfig.json",
+    "build": "tsup && yarn tsc",
```
roninjin10, line 14:
```suggestion
    "build": "tsup && yarn tsc -p tsconfig.json",
```

roninjin10:
removing clean is actually a better fix to my contract bundling issue but going to keep both since the change to bundling derisks this pr a tad via making the old buildsystem unchanged instead of mostly unchanged

roninjin10:
`Trying to understand better, what does 'something' refer to in this line, can you point me to it?`

@nickbalestra I can't I would have to debug to figure it out but debugging is unnecessary since just doing everything cannonical best practice way confirmed fixes the issue

roninjin10:
`I am trying to understand what prevent us from using the module as a cjs in the gateway, haven't we been using it in that way till now?.`

Webpack config because create react app bloats webpack handles this issue well.    A more complicated change to gateway would also be able to handle it well.   Core pain we are feeling is the pain of thinking we were deployed when we weren't for many many weeks

packages/sdk/src/utils/merkle-utils.ts
```diff
@@ -37,9 +37,11 @@ export const makeMerkleTreeProof = (
...
   })
 
-  const proof = tree.getProof(bufLeaves[index], index).map((element: any) => {
-    return toHexString(element.data)
-  })
+  const proof = tree
+    .getProof(bufLeaves[index] as any, index)
+    .map((element: any) => {
+      return toHexString(element.data)
+    })
```
roninjin10, line 44: This change (adding an any) should be unnecessary but I'll test that in a follow up pr so it doesn't block this one

packages/sdk/package.json
```diff
@@ -2,22 +2,25 @@
...
     "clean": "rimraf dist/ ./tsconfig.tsbuildinfo",
     "lint": "yarn lint:fix && yarn lint:check",
+    "prelint:check": "yarn typecheck",
```
roninjin10, line 17:
```suggestion
    "prelint:check": "yarn typecheck",
```
Can remove

packages/sdk/package.json
```diff
@@ -2,22 +2,25 @@
...
     "lint:check": "eslint . --max-warnings=0",
     "lint:fix": "yarn lint:check --fix",
     "pre-commit": "lint-staged",
     "test": "hardhat test",
     "test:coverage": "nyc hardhat test && nyc merge .nyc_output coverage.json",
+    "typecheck": "tsc --noEmit",
```
roninjin10, line 23:
```suggestion
    "typecheck": "tsc --noEmit",
```
Can remove

packages/core-utils/tsconfig.json
roninjin10: This is wierd tsc allows this to happen

smartcontracts (commented):
Looks good to me if this is purely additive

roninjin10:
Testing baseline what ci is passing/failing in this pr is here https://github.com/ethereum-optimism/optimism/pull/4848

roninjin10:
Don't close as stale we really need this so source maps work and debugging the SDK can be done without spending 10 minutes putting console logs into a compiled `node_modules/@eth-optimism/sdk/dist/Foo.js` file

roninjin10:
Also need this because a minority but still large chunk of users will need special build tooling to use all our packages.

.changeset/funny-radios-prove.md
```diff
@@ -0,0 +1,8 @@
+---
+'@eth-optimism/contracts-bedrock': minor
+'@eth-optimism/core-utils': minor
+'@eth-optimism/contracts': minor
+'@eth-optimism/sdk': minor
+---
+
+Fix bug with previous canary release and then generate new changeset to force a new canary release to work
```
roninjin10, line 8: Remove this changeset before landing!!!

tynes:
Just wondering if you are planning on continuing this @roninjin10, would like to help get your code merged

trianglesphere (commented):
.

roninjin10:
@tynes this has actually been ready to go in the entire time unless there are changes requested

roninjin10:
Keeping this in draft don't trust this until I manually test everything again plus a tsup config appears to have gone missing rebasing

roninjin10:
going to do this but to every package instead of just the ones gateway uses post bedrock

PR #4840, opened 2023-02-03
op-node: Log on L1 Info deposit mismatch

trianglesphere:
**Description**

Log significantly more information if the invalid transaction is the L1 info transaction.
This is useful for debugging.


op-node/rollup/derive/engine_consolidate.go
```diff
@@ -26,6 +28,27 @@ func AttributesMatchBlock(attrs *eth.PayloadAttributes, parentHash common.Hash,
 	}
 	for i, otx := range attrs.Transactions {
 		if expect := block.Transactions[i]; !bytes.Equal(otx, expect) {
+			if i == 0 {
```
tynes, line 31: This is a pretty unwieldy piece of code, what do you think about putting it into a function so its not all in the hot code path?

op-node/rollup/derive/engine_consolidate.go
```diff
@@ -26,6 +28,27 @@ func AttributesMatchBlock(attrs *eth.PayloadAttributes, parentHash common.Hash,
...
+							"safe_l1_number", safeInfo.Number, "safe_l1_hash", safeInfo.BlockHash,
+							"safe_l1_time", safeInfo.Time, "safe_seq_num", safeInfo.SequenceNumber,
+							"unsafe_l1_number", unsafeInfo.Number, "unsafe_l1_hash", unsafeInfo.BlockHash,
+							"unsafe_l1_time", unsafeInfo.Time, "unsafe_seq_num", unsafeInfo.SequenceNumber)
+					} else {
+						l.Warn("failed to umarshal l1 info", "errSafe", errSafe, "errUnsafe", errUnsafe)
+					}
+				} else {
+					l.Warn("failed to umarshal tx", "errSafe", errSafe, "errUnsafe", errUnsafe)
+				}
```
ajsutton, line 50: nit: Would it be worth extracting this as a separate method? Then the main business logic would be easier to read, but also you could probably avoid the deep nesting here by returning early on errors.

But by all means ignore me, my experience with styling go code involves watching about three YouTube videos. :)

trianglesphere: good structure is good structure regardless of language. I can pull this out into it's own method - that'll help a bunch. We really shouldn't hit this error case, but I've had to recreate this log enough that I finally wanted to merge it.

op-node/rollup/derive/engine_consolidate.go
trianglesphere: done

op-node/rollup/derive/engine_consolidate.go
```diff
@@ -37,3 +43,33 @@ func AttributesMatchBlock(attrs *eth.PayloadAttributes, parentHash common.Hash,
...
+
+// logL1InfoTxns reports the values from the L1 info tx when they differ to aid
+// debugging. This check is the one that has been most frequently triggered.
+func logL1InfoTxns(l log.Logger, l2Number, l2Timestamp uint64, safeTx, unsafeTx hexutil.Bytes) {
+	// First decode into *types.Transaction to get the tx data.
+	var safeTxValue, unsafeTxValue types.Transaction
+	errSafe := (&safeTxValue).UnmarshalBinary(safeTx)
+	errUnsafe := (&unsafeTxValue).UnmarshalBinary(unsafeTx)
+	if errSafe != nil || errUnsafe != nil {
+		l.Error("failed to umarshal tx", "errSafe", errSafe, "errUnsafe", errUnsafe)
```
protolambda, line 55: Needs a return if the transactions failed to decode.

op-node/rollup/derive/engine_consolidate.go
```diff
@@ -37,3 +43,33 @@ func AttributesMatchBlock(attrs *eth.PayloadAttributes, parentHash common.Hash,
...
+	}
+
+	// Then decode the ABI encoded parameters
+	var safeInfo, unsafeInfo L1BlockInfo
+	errSafe = (&safeInfo).UnmarshalBinary(safeTxValue.Data())
+	errUnsafe = (&unsafeInfo).UnmarshalBinary(unsafeTxValue.Data())
+	if errSafe != nil || errUnsafe != nil {
+		l.Error("failed to umarshal l1 info", "errSafe", errSafe, "errUnsafe", errUnsafe)
```
protolambda, line 63: ditto

op-node/rollup/derive/engine_consolidate.go
trianglesphere: nice catch

protolambda (approved):

PR #4844, opened 2023-02-06
add(CrossDomainOwnable3; contracts-bedrock)

dcbuild3r:
**Description**

Adds [`CrossDomainOwnable3.sol`](https://github.com/dcbuild3r/optimism/blob/develop/packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol) and its corresponding tests [`CrossDomainOwnable3.t.sol`](https://github.com/dcbuild3r/optimism/blob/develop/packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol) to the bedrock contracts. These contracts add the notion of local ownership so that a contract can be owned by a local account and not have to use the cross-domain messenger for all `onlyOwner` calls.


dcbuild3r:
@smartcontracts @tynes

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,136 @@
...
+
+contract CrossDomainOwnable3_Test is Messenger_Initializer {
+    XDomainSetter3 setter;
+
+    function setUp() public override {
+        super.setUp();
+        vm.prank(alice);
+        setter = new XDomainSetter3();
+    }
+
```
tynes, line 27: needs a test saying alice is the owner using the owner getter

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,136 @@
...
+    function test_localOnlyOwner_notOwner_reverts() public {
+        vm.prank(bob);
+        vm.expectRevert("Ownable: caller is not the owner");
+        setter.set(1);
+    }
+
+    function test_crossDomainOnlyOwner_notOwner_reverts() public {
+        vm.prank(alice);
```
tynes, line 35: nit: prefer `vm.prank(setter.owner())` to decouple hardcoding knowing alice is the owner.

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,136 @@
...
+        setter.set(1);
+    }
+
+    function test_localOnlyOwner_succeeds() public {
+        vm.prank(alice);
+        setter.set(1);
+        assertEq(setter.value(), 1);
+    }
+
+    event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
```
tynes, line 101: nit: move this to top of contract def

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,136 @@
...
+        vm.prank(bob);
+        vm.expectRevert("Ownable: caller is not the owner");
+        setter.set(1);
+    }
+
+    function test_crossDomainOnlyOwner_notOwner_reverts() public {
+        vm.prank(alice);
+        setter.transferOwnership(alice, false);
+
+        // set the xDomainMsgSender storage slot
```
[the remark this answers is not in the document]
dcbuild3r: I just used this method because it was in `CrossDomainOwnable2.sol`. I'm personally not particularly knowledgeable about the EVM.

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,136 @@
...
+            sender,
+            target,
+            value,
+            minGasLimit,
+            message
+        );
+
+        // It should be a failed message. The revert is caught,
+        // so we cannot expectRevert here.
+        vm.expectEmit(true, true, true, true);
```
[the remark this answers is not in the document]
dcbuild3r: will fix

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    function test_localOnlyOwner_notOwner_reverts() public {
+        vm.prank(bob);
+        vm.expectRevert("Ownable: caller is not the owner");
+        setter.set(1);
+    }
+
+    function test_crossDomainOnlyOwner_notOwner_reverts() public {
+        vm.prank(alice);
+        setter.transferOwnership(alice, false);
```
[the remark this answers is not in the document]
dcbuild3r: I tried implementing this in the tests, but now the contract isn't compiling. Is this annotation `solc` version dependent?

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * @notice Overrides the implementation of the `transferOwnership` function to allow
+     * for local ownership.
+     * @param newOwner The new owner of the contract.
+     * @param _isLocal If false, the contract uses the cross domain _checkOwner function override.
+     * If false it uses the standard Ownable _checkOwner function.
+     */
+    function transferOwnership(address newOwner, bool _isLocal) external {
```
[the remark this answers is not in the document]
dcbuild3r: So `_owner`? The only reason why I did otherwise is because Ownable has `newOwner` and we are making a wrapper, but I can change it to `_owner`

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
dcbuild3r: done, will appear in next commit

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
dcbuild3r: makes sense, will change in all places

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
dcbuild3r: added an `assertEq` in `setUp`

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    function setUp() public override {
+        super.setUp();
+        vm.prank(alice);
+        setter = new XDomainSetter3();
+    }
+
+    function test_localOnlyOwner_notOwner_reverts() public {
+        vm.prank(bob);
+        vm.expectRevert("Ownable: caller is not the owner");
```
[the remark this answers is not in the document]
dcbuild3r: If we want a `CrossDomainOwnable3:` prefix then I'd have to write an extra function. In the name of brevity, I used super.

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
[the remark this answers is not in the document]
dcbuild3r: fixed the error, works now

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,136 @@
...
+        vm.store(address(L2Messenger), key, value);
+
+        vm.prank(address(L2Messenger));
+        vm.expectRevert("CrossDomainOwnable3: caller is not the owner");
+        setter.set(1);
+    }
+
+    function test_crossDomainOnlyOwner_notOwner2_reverts() public {
+        vm.prank(alice);
+        setter.transferOwnership(alice, false);
```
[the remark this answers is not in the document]
dcbuild3r: added an `vm.expectEmit` in all places where `transferOwnership` takes places

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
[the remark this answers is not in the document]
maurelian: Does it need a new function? I think it'd be enough to make this change on line 29 in: 

```diff
-super._checkOwner();
+require(owner() == _msgSender(), "CrossDomainOwnable3: caller is not the owner");

```

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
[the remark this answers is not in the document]
maurelian: IMO if it's consistent with the rest of our tests, we should let this slide and clean it up with the rest when we finally do that spring cleaning.

maurelian:
Thanks for this, let's get it over the line. :)

The only required test that's failing CI is the gas snapshot. You should be able to fix that by running `yarn gas-snapshot` (with the correct foundry version (`foundryup -C 2ff99025abade470a795724c10648c800a41025e`). 

Or just apply this patch to your branch: [0001-chore-Update-gas-snapshot.patch](https://github.com/ethereum-optimism/optimism/files/10702962/0001-chore-Update-gas-snapshot.patch)

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
```diff
@@ -0,0 +1,59 @@
...
+ */
+abstract contract CrossDomainOwnable3 is Ownable {
+    /**
+     * @notice If true, the contract uses the cross domain _checkOwner function override. If false
+     *         it uses the standard Ownable _checkOwner function.
+     */
+    bool internal isLocal = true;
+
+    /**
+     * @notice Overrides the implementation of the `onlyOwner` modifier to check that the unaliased
```
tynes, line 23: Ok so 2 things here
- Definitely prefer to not use `super._checkOwner()` in the `isLocal` case because of the revert message. If you could port the code internal to that and update the revert message to be crossdomainownable3: prefixed?
- Preference for an event in transfer ownership that emits `isLocal`. The set of events should be able to be used to recreate the current state of the contract. Right now its impossible to know anything about `isLocal`. I have a preference for defining a new event in this contract that emits the previous owner, new owner and is local. Not super opinionated on leaving the call is `super.transferOwnership`, we can leave it or replace it with an impl that sets the owner

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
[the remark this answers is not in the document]
dcbuild3r: ah yes, you're right. my bad.

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
dcbuild3r: sounds good, will add the event. Already changed the `super._checkOwner()` for `require(owner() == _msgSender(), "CrossDomainOwnable3: caller is not the owner");` like @maurelian suggested. Thanks for the review and all the help!

dcbuild3r:
every ci test passed besides two tests that seem unrelated to any of the files that this PR touched. So I guess all good @maurelian? I added the gas-snapshot as instructed above.

maurelian:
One more request, can you please squash these down into a single commit?

Any idea what this one is about?

fba0dce24542bb0efdcefd7a323612de99363392

dcbuild3r:
I'm just merging the current state of `optimism/develop` into my fork so that I make sure I'm on the latest commit. How would you want me to squash everything into a single commit? I'm not great at pulling git shenanigans tbh

tynes:
> I'm just merging the current state of `optimism/develop` into my fork so that I make sure I'm on the latest commit. How would you want me to squash everything into a single commit? I'm not great at pulling git shenanigans tbh

You'll want to use `git rebase -i` - there are a lot of docs out there on how to do this

tynes:
From an implementation standpoint,`CrossDomainOwnable3.sol` looks good to me, will look at the tests now

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    function test_localOnlyOwner_succeeds() public {
+        vm.prank(setter.owner());
+        setter.set(1);
+        assertEq(setter.value(), 1);
+    }
+
+    function test_localTransferOwnership_succeeds() public {
+        vm.expectEmit(true, true, true, true);
```
tynes, line 119: One thing that we are not catching here is the `OwnershipTransferred(address,address)` event in the base contract that emits when `super.TransferOwnership` is called. We probably want to have that in test coverage to prevent accidentally removing that and breaking things in the future

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * @notice Overrides the implementation of the `transferOwnership` function to allow
+     * for local ownership.
+     * @param _owner The new owner of the contract.
+     * @param _isLocal If false, the contract uses the cross domain _checkOwner function override.
+     * If false it uses the standard Ownable _checkOwner function.
+     */
+    function transferOwnership(address _owner, bool _isLocal) external {
```
tynes, line 60: We want a test that this reverts when not the owner calls it

tynes: Because of the revert message not being prefixed with `CrossDomainOwnable3`, we may as well pull the impl of `transferOwnership` inside of here and use the proper revert message

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
dcbuild3r: addressed in the next commit!

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
dcbuild3r: will add it!

dcbuild3r: I had to rename `_owner` back to `newOwner` because `_owner` is the internal variable that holds the owner in `Ownable.sol`. I think that's cleaner than having other contract naming conventions takeover. Especially if we move `super.transferOwnership` functionality into the contract.

dcbuild3r: Ok, after some trial and error, I did leave `super.transferOwnership` because it does use `onlyOwner` where we redefine `_checkOwner` to have `CrossDomainOwnable3:` as a prefix for errors. I only added the require statement that the new owner is not the zero address.

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
```diff
@@ -0,0 +1,71 @@
...
+ *         by contracts on L1. Note that this contract is meant to be used with systems that use
+ *         the CrossDomainMessenger system. It will not work if the OptimismPortal is used
+ *         directly.
+ */
+abstract contract CrossDomainOwnable3 is Ownable {
+    /**
+     * @notice If true, the contract uses the cross domain _checkOwner function override. If false
+     *         it uses the standard Ownable _checkOwner function.
+     */
+    bool internal isLocal = true;
```
tynes, line 20: I think that this should be public and we should test it being changed after its changed

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
```diff
@@ -0,0 +1,71 @@
...
+
```
tynes, line 21: Can you add a `@notice` devdoc for this?

PR #4847, opened 2023-02-06
fix(l2g): startup deadlock

smartcontracts:
**Description**
Fixes a deadlock that can occur in l2geth.

### DIFF
--- a/.changeset/calm-spiders-give.md
+++ b/.changeset/calm-spiders-give.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/l2geth': patch
+---
+
+Fixes a deadlock that can occur when l2geth shuts down improperly
--- a/l2geth/miner/worker.go
+++ b/l2geth/miner/worker.go
@@ -636,10 +636,6 @@ func (w *worker) resultLoop() {
 			if block == nil {
 				continue
 			}
-			// Short circuit when receiving duplicate result caused by resubmitting.
-			if w.chain.HasBlock(block.Hash(), block.NumberU64()) {
-				continue
-			}
 			var (
 				sealhash = w.engine.SealHash(block.Header())
 				hash     = block.Hash()


feld:
This fixes my node that wouldn't sync anymore

smartcontracts:
Multiple users have confirmed that this fixes a startup deadlock.

trianglesphere:
@smartcontracts @tynes bump

trianglesphere:
@smartcontracts @tynes bump

Inphi:
Could you provide some context? How does this fix the deadlock?

smartcontracts:
> Could you provide some context? How does this fix the deadlock?

So basically the flow of what's happening here is:
1. Transaction is applied but transaction index is not updated
2. Node recognizes this and tries to reset the transaction index
3. Node already has the block in question and the code deleted in this PR tries to skip applying the block again
4. Since the block never gets applied again there's a deadlock

Inphi:
Oh I see, the short-circuit prevents the new block event from being emitted and the `SyncService.applyTransaction` waits for that event before returning. 
One thing to note is if the block being re-applied isn't the head/latest block, then no new chain event gets emitted. So you'd still have a deadlock. Is this scenario possible, say when applying transactions in a batch?

protolambda:
@smartcontracts can you help answer the question above? Changes LGTM otherwise

smartcontracts:
> Oh I see, the short-circuit prevents the new block event from being emitted and the `SyncService.applyTransaction` waits for that event before returning. One thing to note is if the block being re-applied isn't the head/latest block, then no new chain event gets emitted. So you'd still have a deadlock. Is this scenario possible, say when applying transactions in a batch?

I'm pretty certain that it's always the case that the block/transaction index is only ever off by one, so it's the latest block being re-applied.

Inphi (approved):
LGTM

bimlas:
@smartcontracts  @Inphi 

Is this merged or not? If yes, which release contains it?

PR #4852, opened 2023-02-09
feat(cmn): better config parsing for BSV2

smartcontracts:
**Description**
Updates config parsing for BaseServiceV2 to be more flexible and to allow for non-string input types. Makes it much easier to use BaseServiceV2 programmatically where you may want to supply the actual type instead of a string representation.

### DIFF
--- a/packages/common-ts/src/base-service/config.ts
+++ b/packages/common-ts/src/base-service/config.ts
@@ -1,0 +1,24 @@
+import { Validator } from './validators'
+
+/**
+ * Cleans a configuration object.
+ *
+ * @param config Configuration object to clean.
+ * @param validators Mapping of config keys to validators.
+ * @returns Cleaned configuration object.
+ */
+export const cleanConfig = (
+  config: { [key: string]: any },
+  validators: { [key: string]: Validator<any> }
+) => {
+  const cleaned = {}
+  for (const [key, value] of Object.entries(config)) {
+    const validator = validators[key] || validators[key.toLowerCase()]
+    if (validator) {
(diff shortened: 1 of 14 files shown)


packages/common-ts/src/base-service/config.ts
```diff
@@ -0,0 +1,24 @@
+import { Validator } from './validators'
+
+/**
+ * Cleans a configuration object.
+ *
+ * @param config Configuration object to clean.
+ * @param validators Mapping of config keys to validators.
+ * @returns Cleaned configuration object.
+ */
+export const cleanConfig = (
```
smartcontracts, line 10: envalid was too restrictive, it only allowed string inputs. Created problems for ChugSplash which needs to create a BaseServiceV2 service programmatically and pass in a Provider object directly. Since it wasn't a string, BSV2 complained. This is much simpler and much more flexible.

tynes:
Is this ready for review?

packages/common-ts/src/base-service/validators.ts
```diff
@@ -1,66 +1,91 @@
...
-  host,
-  port,
-  url,
-  json,
-  wallet,
-  provider,
-  jsonRpcProvider,
-  staticJsonRpcProvider,
-  logLevel,
+  str: (input: string) => {
```
roninjin10, line 11: I'm a bit of a maxi but I would highly recomend looking at Zod sometime.   It's a pretty big upgrade over this in ergonomics, typesafety, and much of this pr is doing things that zod just has built in

PR #4853, opened 2023-02-09
Add internal certman package with mTLS certificate reload fixes for k8s volumes

zhwrd:
**Description**

Create internal certificate reloader package that works with k8s secret volume mounted certificates and integrate with op-signer client.


op-service/tls/certman/certman.go
```diff
@@ -0,0 +1,160 @@
...
+
+	certPath := path.Dir(cm.certFile)
+	keyPath := path.Dir(cm.keyFile)
+
+	if err = cm.watcher.Add(certPath); err != nil {
+		return errors.Wrap(err, fmt.Sprintf("certman: can't watch %s", certPath))
+	}
+	if keyPath != certPath {
+		if err = cm.watcher.Add(keyPath); err != nil {
+			return errors.Wrap(err, fmt.Sprintf("certman: can't watch %s", keyPath))
```
mslipper, line 76:
```suggestion
			return fmt.Errorf("certman: can't watch %s: %w", keyPath, err)
```

op-service/tls/certman/certman.go
```diff
@@ -0,0 +1,160 @@
...
+	files := []string{cm.certFile, cm.keyFile}
+	reload := time.Time{}
+
+loop:
+	for {
+		select {
+		case <-cm.watching:
+			cm.log.Info("watching triggered; break loop")
+			break loop
+		case <-ticker.C:
```
mslipper, line 112: Just so I understand how this loop works:

- `fsnotify` triggers events on the ` cm.watcher` chan
- This updates the `reload` time
- The ticker fires, which checks to see if the `reload` time is in the past. If so, it updates the certificate

The goal of doing it like this is to debounce fs events in case the key/cert don't update atomically. Is this correct?

mslipper (approved):

op-service/tls/certman/certman.go
zhwrd: yeah thats correct, since tls.LoadX509KeyPair loads and validates the certificate, need to ensure both key and cert are updated atomically

PR #4854, opened 2023-02-09
Removes broken / unused changeset workflow

zhwrd:
**Description**

We are using https://github.com/apps/changeset-bot so this broken workflow is not necessary


zhwrd:
@tynes hmm ok yeah it's been broken for at least 3-4 months, likely much longer so just assumed it was deprecated but can look into fixing it

tynes:
If its been broken for so long then I'm ok with removing it, it obviously wasn't doing what its supposed to do

tynes (approved):

PR #4855, opened 2023-02-09
contracts-bedrock: add TransferOnion contract

tynes:
**Description**

Adds the `TransferOnion` contract which can be used to batch send tokens. The set of transfers are committed to during contract initialization, and the token owner must approve the `TransferOnion` contract and then anybody can unwrap the commitment to the transfers.

The transfers are committed to with a hash onion. This is a simple vector commitment that must be linearly unwrapped. Each hash commits to the next hash and the very first hash is `bytes32(0)`.

### DIFF
--- a/packages/contracts-bedrock/contracts/periphery/TransferOnion.sol
+++ b/packages/contracts-bedrock/contracts/periphery/TransferOnion.sol
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { ReentrancyGuard } from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/**
+ * @title  TransferOnion
+ * @notice TransferOnion is a hash onion for distributing tokens. The shell commits
+ *         to an ordered list of the token transfers and can be permissionlessly
+ *         unwrapped in order. The SENDER must `approve` this contract as
+ *         `transferFrom` is used to move the token balances.
+ */
+contract TransferOnion is ReentrancyGuard {
+    using SafeERC20 for ERC20;
+
+    /**
+     * @notice Struct representing a layer of the onion.
+     */
+    struct Layer {
+        address recipient;
+        uint256 amount;
+        bytes32 shell;
+    }
+
+    /**
+     * @notice Address of the token to distribute.
+     */
+    ERC20 public immutable TOKEN;
+
+    /**
+     * @notice Address of the account to distribute tokens from.
+     */
+    address public immutable SENDER;
+
+    /**
+     * @notice Current shell hash.
+     */
+    bytes32 public shell;
+
+    /**
+     * @param _token  Address of the token to distribute.
+     * @param _sender Address of the sender to distribute from.
+     * @param _shell  Initial shell of the onion.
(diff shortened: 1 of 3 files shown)


packages/contracts-bedrock/contracts/periphery/TransferOnion.sol
```diff
@@ -0,0 +1,84 @@
...
+        shell = _shell;
+    }
+
+    /**
+     * @notice Peels layers from the onion and distributes tokens.
+     *
+     * @param _layers Array of onion layers to peel.
+     */
+    function peel(Layer[] memory _layers) public nonReentrant {
+        bytes32 tshell = shell;
```
maurelian, line 59: nit: what is the t in `tshell`, temp I assume? 

Can it be written `tShell` or even `tempShell` to be more explicit and visually distinctive?

packages/contracts-bedrock/contracts/periphery/TransferOnion.sol
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * @title  TransferOnion
+ * @notice TransferOnion is a hash onion for distributing tokens. The shell commits
+ *         to an ordered list of the token transfers and can be permissionlessly
+ *         unwrapped in order. The SENDER must `approve` this contract as
+ *         `transferFrom` is used to move the token balances.
+ */
+contract TransferOnion is ReentrancyGuard {
+    using SafeERC20 for ERC20;
```
maurelian, line 16: I assume that this will be used for the `GovernanceToken`? If so it's unnecessary since we know that it will throw on failure.

packages/contracts-bedrock/contracts/periphery/TransferOnion.sol
```diff
@@ -0,0 +1,84 @@
...
+            require(
+                keccak256(abi.encode(layer.recipient, layer.amount, layer.shell)) == tshell,
+                "TokenOnion: what are you doing in my swamp?"
+            );
+
+            // Update the onion layer.
+            tshell = layer.shell;
+
+            // Transfer the tokens.
+            TOKEN.safeTransferFrom(SENDER, layer.recipient, layer.amount);
```
maurelian, line 74: The only revert I'd anticipate is from this contract having insufficient balance, which we can avoid if we're careful. Still it's a little bit scary that this could revert, as it would lock up this function, and any remaining balance, permanently (afaict).

packages/contracts-bedrock/contracts/periphery/TransferOnion.sol
maurelian: Wow, this is expensive too!

Snapshot after removing:

```diff
-TransferOnionTest:test_constructor_succeeds() (gas: 563233)
-TransferOnionTest:test_unwrap_succeeds() (gas: 723651)
+TransferOnionTest:test_constructor_succeeds() (gas: 372457)
+TransferOnionTest:test_unwrap_succeeds() (gas: 531763)
```

packages/contracts-bedrock/contracts/periphery/TransferOnion.sol
tynes: Since this runs on L2, the execution gas i'm less worried about and prefer safety and small calldata. Optimizations would involve tightly packing the calldata because calldata structs are not tightly packed

packages/contracts-bedrock/contracts/periphery/TransferOnion.sol
tynes: Yes sure that is a good callout

maurelian (approved):

PR #4856, opened 2023-02-10
fix(specs): Use alias syntax to address long lines

maurelian:
Fixes the linter error in [CI](https://circleci.com/gh/ethereum-optimism/optimism/440180).


mslipper:
Vitalik’s SSL cert is expired again. The other issue is fixed. Merging.

PR #4857, opened 2023-02-11
ci, ctb: Reduce devnet sequencer drift

mslipper:
The sequencer drift was larger than the sequencing window. This can cause reorgs. To fix I reduced the drift to 10s, which is lower than the 45 second sequencing window specified via the L1 block time and the window size.


trianglesphere (approved):
lgtm. Tests passed which was good. I can keep an eye on the stability to see if it's too low, but it's also clear that if it's too large it causes problems.

PR #4859, opened 2023-02-12
op-node: Monitor static peers in a background process

mslipper:
Currently, connections with static peers are not re-established in the event of disconnection. This causes us to have to restart replicas whenever we restart the sequencer. This PR adds a background process to monitor static peers for connectedness, and automatically reconnect when necessary.


op-node/p2p/host.go
```diff
@@ -44,6 +52,73 @@ func (e *extraHost) ConnectionManager() connmgr.ConnManager {
...
+	tick := time.NewTicker(time.Minute)
+	defer tick.Stop()
+
+	for {
+		select {
+		case <-tick.C:
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+			var wg sync.WaitGroup
+
+			e.log.Debug("polling static peers")
```
protolambda, line 95:
```suggestion
			e.log.Debug("polling static peers", "peers", len(e.staticPeers))
```

protolambda (approved):
LGTM. One nit about adding logging in the loop. It's useful to know how many peers are being checked.

mslipper:
Nice, thanks. Accepted the suggestion.

op-node/p2p/host.go
```diff
@@ -142,26 +217,28 @@ func (conf *Config) Host(log log.Logger, reporter metrics.Reporter) (host.Host,
...
+
+	out := &extraHost{
+		Host:        h,
+		connMgr:     connMngr,
+		log:         log,
+		staticPeers: staticPeers,
+		quitC:       make(chan struct{}),
+	}
+	out.initStaticPeers()
+	if len(conf.StaticPeers) > 0 {
```
trianglesphere, line 238: Can we add static peers via the command line? It seems like we may always want the background process.

protolambda: We can add them via command line on start-up, so it's safe as read-only data in the go routine. We have an RPC to connect/protect dynamically, but not yet a notion of adding static peers to this re-connect loop after startup. Can be added later if really necessary.

mslipper: You mean via RPC? You can add a peer there, but it's stored separately from "static" peers which are just the ones users specify on the CLI at startup.

I think we should do reconnection logic for those peers separately - some of them might be inbound peers, or peers we don't care about reconnecting to.

PR #4863, opened 2023-02-12
op-node: Properly check storage proof

mslipper:
The storage proof was not being properly checked. This PR corrects the check, and adds unit/fuzz tests around the storage proof checks to prevent regressions.

Fixes CLI-3344


op-node/eth/account_proof.go
```diff
@@ -52,7 +52,11 @@ func (res *AccountResult) Verify(stateRoot common.Hash) error {
 		if err != nil {
 			return fmt.Errorf("failed to verify storage value %d with key %s (path %x) in storage trie %s: %w", i, entry.Key, path, res.StorageHash, err)
 		}
-		if !bytes.Equal(val, val) {
+		comparison, err := rlp.EncodeToBytes((*big.Int)(&entry.Value).Bytes())
```
tynes, line 55: I believe there is a `.Big` method on entry.value that would clean up the typecast

mslipper: Nice, this is much cleaner. Fixed.

protolambda (approved):

PR #4864, opened 2023-02-13
fix(op-node): Include chainID in p2p signing hash

ajsutton:
**Description**

Fixes the p2p `SigningHash` function to correctly include the chainID in the resulting hash.

**Tests**

Added tests to confirm that each input to `SigningHash` affects the generated hash.

**Additional context**

*Important*: this is a backwards incompatible change for p2p gossip. Once the sequencer is upgraded with this fix, any node that hasn't been upgraded will reject all gossip because of the difference in how signing hash is calculated. If backwards compatibility is required, we'll have to add support for accepting gossip signed with either the old or new hash and give time for users to upgrade before changing the hash used by the sequencer to sign.

**Metadata**
- Fixes https://linear.app/optimism/issue/CLI-3341/sherlock-067-125-chain-id-is-clobbered-in-p2p-signing-hash

### DIFF
--- a/op-node/p2p/signer.go
+++ b/op-node/p2p/signer.go
@@ -20,6 +20,21 @@ type Signer interface {
 	io.Closer
 }
 
+func LegacySigningHash(domain [32]byte, chainID *big.Int, payloadBytes []byte) (common.Hash, error) {
+	var msgInput [32 + 32 + 32]byte
+	// domain: first 32 bytes
+	copy(msgInput[:32], domain[:])
+	// chain_id: second 32 bytes
+	if chainID.BitLen() > 256 {
+		return common.Hash{}, errors.New("chain_id is too large")
+	}
+	chainID.FillBytes(msgInput[32:64])
+	// payload_hash: third 32 bytes, hash of encoded payload
+	copy(msgInput[32:], crypto.Keccak256(payloadBytes))
+
+	return crypto.Keccak256Hash(msgInput[:]), nil
+}
+
 func SigningHash(domain [32]byte, chainID *big.Int, payloadBytes []byte) (common.Hash, error) {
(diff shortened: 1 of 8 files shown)


op-node/p2p/signer.go
```diff
@@ -30,7 +30,7 @@ func SigningHash(domain [32]byte, chainID *big.Int, payloadBytes []byte) (common
 	}
 	chainID.FillBytes(msgInput[32:64])
 	// payload_hash: third 32 bytes, hash of encoded payload
-	copy(msgInput[32:], crypto.Keccak256(payloadBytes))
```
protolambda, line 33: Fix looks good. But adding this comment to resolve to ensure we can ACK that this is getting merged, since it's a breaking change. As soon as this is deployed to the sequencer, replica nodes will stop following the unsafe chain until they update as well. All nodes will stay in sync with the L1 data though, so I think it's fine.

protolambda (approved):

mslipper:
Change looks good to me, but I'd like to request that we add functionality to support both signing hashes so that replica operators have time to upgrade. Basically let folks know "breaking changes are coming in a week, please upgrade now otherwise your nodes will stop following the unsafe chain." Is that going to be a lot of effort @ajsutton?

mslipper (changes requested):
Requesting changes to prevent auto-merge - see comment above.

ajsutton:
Yep, I figured some backwards compatibility would likely be required but given there was a well written up ticket with the change figured it made sense to get something up as a starting point for the discussion.  Accepting either signature hash looks fairly straight forward, just need to work out how to test it as there doesn't seem to be a lot of existing testing for the signature verification.

mslipper (approved):

mslipper:
Great, thanks Adrian! LGTM.

PR #279, opened 2020-09-23
Set signer.provider to OptimismProvider

K-Ho:
## Description

Previously our `OptimismSigner`'s `provider` was the `Web3Provider` (from metamask) and `optimism` was set to the `OptimismProvider`. This leads to issues where `signer.provider` is used by default. This PR changes `signer.provider` to be the `OptimismProvider` and adds `signer.web3` which is the `Web3Provider` (from metamask)

### Fixes
- Fixes # 

## Contributing Agreement

- [X] I have read and understood the [Optimism Contributing Guide and Code of Conduct](https://github.com/ethereum-optimism/optimism-monorepo/blob/master/.github/CONTRIBUTING.md) and am following those guidelines in this pull request.

PR #4867, opened 2023-02-13
op-node: Fix `Frame` parsing

sebastianst:
**Description**

This PR fixes the binary-unmarshaling of  `Frame`s. The last byte of an encoded frame could be missing and would still have passed as a valid frame.

Function `Frame.UnmarshalBinary` now does proper error checks to ensure all frame data is present in an encoded frame. It also converts `io.EOF` errors that happen intermittently into `io.ErrUnexpectedEOF` errors, which is more idiomatic.

Still in draft to add more tests.

**Tests**

Added:
* Happy marshaling followed by unmarshaling of random frames.
* Failing unmarshaling of frames with an invalid value for last byte `is_last`.
* Failing unmarshaling of frames truncated at various positions.

**Metadata**
- Fixes CLI-3355 (Sherlock `#279`)


sebastianst:
Is it worth making `randomFrame` public in the `testutils` package?

op-node/rollup/derive/frame_test.go
```diff
@@ -23,3 +31,122 @@ func FuzzParseFrames(f *testing.F) {
 		}
 	})
 }
+
+func TestFrameMarshaling(t *testing.T) {
```
tynes, line 35: I think this would be more suitable to fuzz testing. I had a problem in the past where the rng was deterministic when running `go test`

tynes:
> Is it worth making `randomFrame` public in the `testutils` package?

Definitely think this is a good idea to reduce code duplication!

sebastianst:
> > Is it worth making `randomFrame` public in the `testutils` package?
> 
> Definitely think this is a good idea to reduce code duplication!

It would definitely be nice, but looking at the `testutils` package, the scope of its random functions are limited to general Ethereum types, not Optimism types. So if we wanted to make `randomFrame` public, maybe it would be cleaner to move it into a package like `op-node/rollup/derive/test`. (It's a general pattern I've been using in the past: instead of putting many test utils into a single large package, create many `test` sub-packages and put relevant testing tools there. This can help with dependency conflicts in tests down the line.)

Is there any other place in the code where random frames would be used, except for the tests I added here?

op-node/rollup/derive/frame_test.go
```diff
@@ -23,3 +31,218 @@ func FuzzParseFrames(f *testing.F) {
...
+		t.Run(tr.desc, func(t *testing.T) {
+			var opts []frameOpt
+			if !tr.genData {
+				opts = []frameOpt{frameWithDataLen(0)}
+			}
+			frame := randomFrame(rng, opts...)
+			var data bytes.Buffer
+			require.NoError(t, frame.MarshalBinary(&data))
+
+			// truncate last data byte & is_last
```
trianglesphere, line 129:
```suggestion
			// truncate the binary frame per the test spec 
```
nit: no longer just last data byte

trianglesphere (commented):
I love seeing more tests

op-node/rollup/derive/frame_test.go
sebastianst: obsolete comment removed 👍🏻

op-node/rollup/derive/frame_test.go
sebastianst: I also wanted to go with fuzzing first, but Go offers very limited control over how fuzzy data is generated. If you look at `randomFrame`, it covers all possible max values for e.g. the frame number and data length. To the best of my knowledge, I couldn't easily achieve this with go's fuzzing capabilities, so I decided to write the custom random frame generator. Also note that there are already two general fuzz tests at the top of the `frame_test.go` file.

I followed the example in other tests in the monorepo and created a random frame generator `randomFrame` that takes a `rng`, which is seeded with the unix nano time (`time.Now().UnixNano()`) at the start of each test. This should provide sufficiently random seeds and thus sufficiently random generated frames. I'm aware of instances in the past (e.g. [here](https://github.com/golang/go/issues/22037) or [here](https://groups.google.com/g/golang-dev/c/DboHw1eCiSc)) where, on MacOS, repeated calls returned the same value. But if test runs are even milliseconds apart, this isn't a problem. And the latest the CI would run with a new seed every time.

tynes (approved):

PR #4869, opened 2023-02-14
maint(ctb): revamp the rekey task

smartcontracts:
**Description**
The rekey task wasn't being used, so I'm commandeering it for the Getting Started guide. This makes the task more legible for end-users.


mslipper (approved):

tynes:
What do you think about just hardcoding the hardhat keys instead of using this? Then they will be prepopulated with ether in the devnet, also generating keys with this is sus for prod

tynes (approved):

PR #4870, opened 2023-02-14
feat(opn): expose GasUsed in BlockInfo type

smartcontracts:
**Description**
Exposes the GasUsed field in the BlockInfo type. This has no impact on the op-node for Optimism Mainnet, but it will be used in the OP Stack Getting Started guide.


op-node/sources/types.go
```diff
@@ -46,6 +46,7 @@ type HeaderInfo struct {
 	baseFee     *big.Int
 	txHash      common.Hash
 	receiptHash common.Hash
+	gasUsed     uint64
```
trianglesphere, line 49: You need to set this field in `(hdr *rpcHeader) Info`

trianglesphere (changes requested):

op-node/sources/types.go
smartcontracts: Done

mslipper (approved):

tynes (approved):

trianglesphere (approved):

PR #4873, opened 2023-02-14
fix(op-node): Increase gossip seenMessagesTTL

ajsutton:
**Description**

Increases the libp2p gossip `seenMessagesTTL` value so that it prevents a valid block from being replayed before it is invalid according to the application layer validation rules.

Specifically, a block first becomes valid when its timestamp is 5 seconds in the future, and remains valid until its timestamp is 60 seconds in the past. Thus there's a 65 second period where blocks are valid. By setting `seenMessagesTTL` to cover this period, all forms of replaying gossip are prevented as either the libp2p layer will discard the message as a duplicate or it will be rejected based on its timestamp.

While it's possible to ignore these duplicates with application layer validation it's more efficient to discard them at the libp2p layer to avoid the need for any validation to occur (which currently would include signature verification). Using `seenMessagesTTL` is also simpler as it's cache size is based on time so guarantees to cover all messages, whereas adjusting the size of the `blockHeightLRU` in the application layer validation requires establishing an upper maximum on the number of valid blocks which could be received in the time window.

**Metadata**
- Fixes https://linear.app/optimism/issue/CLI-3334/sherlock-177-crafted-p2p-spam-can-render-nodes-permanently-unable-to


tynes:
Adding @protolambda and @trianglesphere for review to this one

protolambda (approved):
LGTM. And this is a non-breaking change in practice (since we don't rely on duplicate block propagation), should be fine to roll out whenever.

PR #4874, opened 2023-02-14
op-chain-ops: Add inject-mints tool

mslipper:
We have some replicas that synced with an old version of l2geth that wasn't adding mints to the witness data. This PR adds a tool called `inject-mints` that will add historical mints data to an existing witness file so that we don't have to resync those replicas.

Fixes CLI-3348


op-chain-ops/cmd/inject-mints/main.go
```diff
@@ -0,0 +1,87 @@
...
+				return fmt.Errorf("error opening witness file: %w", err)
+			}
+
+			log.Info("Reading mint events from DB")
+			headBlock := rawdb.ReadHeadBlock(ldb)
+			headNum := headBlock.NumberU64()
+			seenAddrs := make(map[common.Address]bool)
+			bar := progressbar.Default(int64(headNum))
+			var count uint64
+			progressCb := func(headNum uint64) {
```
tynes, line 64: Does this log automatically or render a progress bar?

mslipper: Renders a pbar.

mslipper: I like having the pbar since it'll give me an ETA. I don't expect to have to run this during the actual migration.

op-chain-ops/cmd/op-migrate/main.go
```diff
@@ -164,10 +164,7 @@ func main() {
 
 			dbCache := ctx.Int("db-cache")
 			dbHandles := ctx.Int("db-handles")
-
-			chaindataPath := filepath.Join(ctx.String("db-path"), "geth", "chaindata")
-			ancientPath := filepath.Join(chaindataPath, "ancient")
-			ldb, err := rawdb.NewLevelDBDatabaseWithFreezer(chaindataPath, dbCache, dbHandles, ancientPath, "", false)
+			ldb, err := db.Open(ctx.String("db-path"), dbCache, dbHandles)
```
tynes, line 167: `db-path` doesn't have `Required: true` set so this can be an empty string. Perhaps we want to check for empty values inside of `db.Open` and return an error? Also could add `Required: true` to cli args in this script, could be a diff PR

tynes (approved):

PR #4876, opened 2023-02-14
fix(ctb): Add creation time check for `SUBMISSION_INTERVAL` > `L2_BLOCK_TIME`

clabby:
# Overview

Adds a creation time assertion that the passed `l2BlockTime` must be > 0 and that the passed `submissionInterval` is > the passed `l2BlockTime`

Also includes a quick monkey-patch for the `contracts-bedrock` tests that were broken in #4775  

Fixes CLI-3366


clabby:
* develop
  * **PR #4876** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4876" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4876?utm_source=stack-comment).

packages/contracts-bedrock/package.json
```diff
@@ -18,7 +18,7 @@
     "build:forge": "forge build",
     "build:with-metadata": "FOUNDRY_PROFILE=echidna yarn build:forge",
     "build:differential": "tsc scripts/differential-testing.ts --outDir dist --moduleResolution node --esModuleInterop",
-    "build:fuzz": "go build -o test-case-generator test-case-generator/cmd/fuzz.go",
+    "build:fuzz": "(cd test-case-generator && go build ./cmd/fuzz.go)",
```
trianglesphere, line 21: this fix lgtm

refcell (approved):
:100:

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -90,6 +90,12 @@ contract L2OutputOracle is Initializable, Semver {
         address _proposer,
         address _challenger
     ) Semver(1, 0, 0) {
```
tynes, line 92: Do you mind adding the same check here?
https://github.com/ethereum-optimism/optimism/blob/3969dc495df37e2ef246aeb02e629f3238e0e7e7/packages/contracts-bedrock/deploy/012-L2OutputOracleImpl.ts#L12

Then the deployment will fail in the typescript instead of deploying a reverting contract

clabby: Sure! Added the check to the deployment script as well. Still think we should retain the check in the contracts in case anyone who forks the stack chooses to go with a different route for deployment, doesn't add much to the deployment cost and prevents the proposer from being bricked.

tynes, line 92: One last nit: i think that this should technically be a minor semver bump

tynes (approved):

egonspace:
`submissionInterval` is in the unit of block count, and `l2BlockTime` is in the unit of seconds. Is it valid to compare these two?

PR #4877, opened 2023-02-14
fix(ctb): add is IERC165 to IOptimismMintableERC20

maurelian:
**Description**

Inherits `IERC165` to `IOptimismMintableERC20`.


tynes:
If you rebase, it should fix the tests

tynes (approved):

PR #4878, opened 2023-02-14
op-chain-ops: more strict checks for config validity

tynes:
**Description**

Adds in additional sanity checks for the config validity to prevent accidental bad configurations of the L2 predeploys. Ensures that specific values are set in the immutables. Note this this makes the immutables config specific to optimism, previously it was implemented in a generic way using a map so that it could be used with other projects that are doing similar things with setting contracts in state with their immutables preset. We could turn it into a struct instead of using a map which could make it easier to use.

Tests were updated as the data in the tests failed the validation.

### DIFF
--- a/op-chain-ops/genesis/config.go
+++ b/op-chain-ops/genesis/config.go
@@ -22,7 +22,10 @@ import (
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
 )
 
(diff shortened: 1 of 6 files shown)


tynes:
Will update this PR with changes to `op-e2e`, looks like its catching bad config there

protolambda (approved):

PR #4879, opened 2023-02-14
contracts-bedrock: add semver to mintable tokens

tynes:
**Description**

Even though the mintable tokens are not upgradable, this commit adds semver to them in case we change the implementations down the road and then update the factories to deploy the new versions. I do think it would be nice to know what version different mintable tokens are on, in case we decide to add functionality to the mintable tokens.

This will make it very easy to know what version the mintable tokens are at, in case newer mintable tokens have different functionality in some way.

Will update this PR with another commit because something with the test case generator is breaking on my machine


tynes:
https://github.com/ethereum-optimism/optimism/pull/4881 fixes breaking tests

maurelian (approved):

PR #4881, opened 2023-02-14
contracts-bedrock: be explicit with mintable erc721 token version

tynes:
**Description**

Previously any solidity version greater than or equal to 0.8.15 was allowed, this updates it such that exactly 0.8.15 is allowed. This matches the style of the other contracts in the system.


tynes:
The legacy integration tests import this contract, needed to update the hardhat config with the version of the compiler that we are using

PR #4882, opened 2023-02-14
contracts-bedrock: CrossDomainOwnable3

tynes:
**Description**

Adds a `CrossDomainOwnable3` contract that allows the
owner to be toggled between an L1 account and an L2 account.
In the L1 version, it is expected that the account goes
through the cross domain messenger.

Co-authored-by: dcbuilder.eth <dcbuilder@protonmail.com>
Co-authored-by: Mark Tyneway <mark.tyneway@gmail.com>

Replaces https://github.com/ethereum-optimism/optimism/pull/4844


dcbuild3r:
🫡

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * @notice Overrides the implementation of the `onlyOwner` modifier to check that the unaliased
+     *         `xDomainMessageSender` is the owner of the contract. This value is set to the caller
+     *         of the L1CrossDomainMessenger.
+     */
+    function _checkOwner() internal view override {
+        if (isLocal) {
+            require(owner() == _msgSender(), "CrossDomainOwnable3: caller is not the owner");
```
maurelian, line 54: We just use `msg.sender` in the other CrossDomainOwnable's. Should be consistent on that IMO.

packages/contracts-bedrock/contracts/test/CrossDomainOwnable3.t.sol
```diff
@@ -0,0 +1,225 @@
...
+
+    function setUp() public override {
+        super.setUp();
+        vm.prank(alice);
+        setter = new XDomainSetter3();
+    }
+
+    function test_constructor_succeeds() public {
+        assertEq(setter.owner(), alice);
+        assertEq(setter.isLocal(), true);
```
maurelian, line 44: Should it be possible to deploy the contract with a non-local initial owner?

dcbuild3r: If we want to we could have it setup so that `_isLocal` is a constructor parameter, but it doesn't make much sense since we'd have to do a cross-domain which is one of the reasons we created `CrossDomainOwnable3.sol` in the first place.

packages/contracts-bedrock/contracts/L2/CrossDomainOwnable3.sol
dcbuild3r: alright. I'll change it.

tynes: Pushed a fix for this @maurelian

maurelian (approved):

PR #4883, opened 2023-02-14
Update op-geth

ajsutton:
**Description**

Update to latest op-geth version
Update the update-op-geth script to account for removal of go.work

**Metadata**
- Part of https://linear.app/optimism/issue/CLI-3281/spike-work-out-differences-between-gettransactionbyhash-response-on


ops/scripts/update-op-geth.py
```diff
@@ -11,34 +10,10 @@
 
 
 def main():
-	for project in ('op-service', 'op-node', 'op-proposer', 'op-batcher', 'op-bindings', 'op-chain-ops', 'op-e2e', 'op-wheel'):
+	for project in ('.', 'op-wheel'):
```
ajsutton, line 13: Note that `indexer/go.mod` also references op-geth but wasn't previously included in the list to update.  Should it be added here?

mslipper: Yes, we should add the indexer here.

mslipper (approved):
LGTM pending indexer fix.

ops/scripts/update-op-geth.py
ajsutton: Done.

PR #4885, opened 2023-02-15
Don't log warnings when legacy block gossip validation fails

ajsutton:
**Description**

Since #4864 nodes on any chain that still had a sequencer using the legacy block gossip hash (all of them currently) would print a `unexpected block author` warning for each gossip message because the signature check using the new block hash algorithm failed.

This updates the logic to check the legacy hash first without logging any warnings, then if that isn't accepted, check with the new hash and log warnings on failure.

**Metadata**
- Part of https://linear.app/optimism/issue/CLI-3341/sherlock-067-125-chain-id-is-clobbered-in-p2p-signing-hash


mslipper (approved):
LGTM

PR #4891, opened 2023-02-15
proxyd: Add global flag to method overrides

mslipper:
This allows us to set global limits on individual RPCs that ignore any origin/user agent exemption.

### DIFF
--- a/proxyd/server.go
+++ b/proxyd/server.go
@@ -39,27 +39,28 @@ const (
 var emptyArrayResponse = json.RawMessage("[]")
 
 type Server struct {
-	backendGroups        map[string]*BackendGroup
-	wsBackendGroup       *BackendGroup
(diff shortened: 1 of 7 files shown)


zhwrd (approved):

proxyd/server.go
```diff
@@ -243,7 +250,9 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
 	}
 
 	isLimited := func(method string) bool {
-		if isUnlimitedOrigin || isUnlimitedUserAgent {
+		isGloballyLimitedMethod := s.isGlobalLimit(method)
+		fmt.Println(method, isGloballyLimitedMethod)
```
Inphi, line 254: leftover debug print.

PR #4898, opened 2023-02-15
bedrock: update op-geth to v1.11.2

protolambda:
**Description**

This updates the monorepo to use the latest `op-geth`, rebased onto the upstream Geth release of today: ~~v1.11.0~~ ~~v1.11.1~~ v1.11.2 (patch release)

See geth-rebase notion doc and `op-geth` `optimism-experimental` branch for geth changes.

See https://github.com/ethereum-optimism/op-geth/tree/optimism-experimental

Notable upstream geth changes:
- configuration for shanghai
- withdrawals in block-building code-path
- removal of legacy receipts (but I preserved support for optimism legacy receipts)
- pebble db was introduced! leveldb is still there, and we'll use that for the foreseeable future. DB-opening functions changed.
- the `core/beacon` types package was refactored and moved to a new `beacon/engine` package

This geth release was long awaited, ~3.5 months! (start nov to mid feb).

This PR depends on the libp2p update PR #4917  and op-wheel bedrock module PR #4918 

**Tests**

No functionality changed. op-e2e and op-node tests pass.

**Metadata**

Fix CLI-3302


protolambda:
Linter is timing out, I think because of dependency changes / Go version changes in there. We may have to update the Go linter in CI.

op-chain-ops/state/memory_db.go
```diff
@@ -312,3 +313,18 @@ func (db *MemoryStateDB) ForEachStorage(addr common.Address, cb func(common.Hash
...
+
+func (db *MemoryStateDB) SetTransientState(addr common.Address, key, value common.Hash) {
+	//TODO implement me
+	panic("implement me")
+}
+
+func (db *MemoryStateDB) Prepare(rules params.Rules, sender, coinbase common.Address, dest *common.Address, precompiles []common.Address, txAccesses types.AccessList) {
+	//TODO implement me
+	panic("implement me")
+}
```
ajsutton, line 330: Should we remove the TODOs here? Looks like there's a number of functions we deliberately don't implement in this version so can just improve the panic messages and leave it like that.

ajsutton (commented):
Looks good to me. If its possible, it might be worth splitting the libp2p upgrade into its own PR and that can go in independent of the Geth upgrade and reduce the changeset size a bit.

op-chain-ops/state/memory_db.go
protolambda: @tynes you introduced this transient state functionality in upstream geth right? Do you think we need to implement these functions?

protolambda:
Work in progress:
- rebasing from v1.11.0 to v1.11.1
- investigating extra-data issue found by Adrian
- updating `fork.yaml` for updated diff description

op-chain-ops/state/memory_db.go
tynes: I did implement this in upstream, I think its fine to leave them unimplemented because they will not be called by us

op-chain-ops/genesis/check.go
```diff
@@ -181,8 +181,16 @@ func PostCheckUntouchables(udb state.Database, currDB *state.StateDB, prevRoot c
 		log.Info("checked code hash", "address", addr, "hash", hash)
 
 		// Ensure that the current/previous roots match
-		prevRoot := prevDB.StorageTrie(addr).Hash()
-		currRoot := currDB.StorageTrie(addr).Hash()
+		prevStorage, err := prevDB.StorageTrie(addr)
```
tynes, line 184: It is possible for the trie (first arg) to be `nil` if the account is not in the trie. I ran into this when working on 4337 stuff. I don't think we should ever hit the `nil` case, but best to check for `nil` to prevent an unexpected panic. If the trie is nil then its `Hash()` is `EmptyStorageRoot`

op-chain-ops/genesis/layer_one.go
```diff
@@ -216,7 +216,10 @@ func BuildL1DeveloperGenesis(config *DeployConfig) (*core.Genesis, error) {
 	}
 
 	for _, dep := range deployments {
-		st := stateDB.StorageTrie(dep.Address)
+		st, err := stateDB.StorageTrie(dep.Address)
```
tynes, line 219: Same as above, `st` can be `nil` but it shouldn't be in this case

op-chain-ops/state/memory_db.go
tynes: Perhaps `panic("Prepare unimplemented")` etc

protolambda:
Updated op-geth to base on top of upstream geth v1.11.2

tynes (approved):

tynes:
Good work!

mslipper (approved):

mslipper:
Merging since Josh is on vacation.

PR #4899, opened 2023-02-16
ci: Add tag service action

mslipper:
Now that we've decoupled Go versioning from binary versioning, we need a way to programmatically tag binary versions in order to trigger Docker builds. This PR adds a GitHub Action that will automatically create and push binary tag versions so that CircleCI can build and push them to GCP.

The action is manually triggered. To use it, navigate to the Actions tab on GitHub and select Tag Service. Then, you can tag a new release by hitting Run Workflow. The menu looks like this:

Screen Shot 2023-02-15 at 8 48 46 PM

The action will automatically detect the latest binary version of a given service, create a new tag with the specified version bump increment, then push the tag to GitHub. Binary tags have the format `<service>/<version>`, so the next op-node binary version would be `op-node/v0.10.15`.

You can see an example service run on [my fork](https://github.com/mslipper/optimism/actions/runs/4190473611).

Limitations:

- It doesn't make GitHub releases for the tags - it only makes the tags themselves.
- It doesn't do any kind of changelog management.
- It doesn't release versions of the `op-*` Go code for import into other Go modules - just binaries.
- It doesn't remove binary builds currently managed by changesets.
- This PR doesn't migrate the indexer and proxyd builds to build in CCI.
- This PR doesn't add the tag to the Go build metadata.

If we like the approach in this PR, I can work to reduce the limitations above in subsequent PRs.

Closes CLI-3416


ops/tag-service/tag-service.py
```diff
@@ -0,0 +1,128 @@
...
+        bumped = bumped.bump_prerelease()
+
+    new_version = 'v' + str(bumped)
+    new_tag = f'{service}/{new_version}'
+
+    log.info(f'Bumped version: {new_version}')
+
+    log.info('Configuring git')
+    # The below env vars are set by GHA.
+    gh_actor = os.environ['GITHUB_ACTOR']
```
tynes, line 100: Will subprocess throw if the command returns non zero error code? Just wondering if we should check to make sure the env vars arent empty strings

tynes: Looks like it returns a returncode on an object, perhaps we want to check that the calls were successful

https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess

mslipper: `check=True` will throw on a nonzero return code.

tynes:
This looks good to me for a first iteration, altho i do think changelogs will be important for other users of our software

ops/tag-service/tag-service.py
tynes: Ok cool so no real need to check for empy env vars since it'll throw

ops/tag-service/tag-service.py
```diff
@@ -0,0 +1,128 @@
...
+        new_tag,
+        '-m',
+        MESSAGE_TEMPLATE.format(service=service, version=new_version)
+    ], check=True)
+
+    log.info('Pushing tag to origin')
+    subprocess.run(['git', 'push', 'origin', new_tag], check=True)
+
+
+def die(msg):
```
tynes, line 122: This is unused?

mslipper: removed

tynes (approved):

.github/workflows/tag-service.yml
```diff
@@ -0,0 +1,57 @@
+name: Tag Service
+
+on:
+  workflow_dispatch:
```
protolambda, line 4: Who is authorized to dispatch this workflow?

ops/tag-service/tag-service.py
```diff
@@ -0,0 +1,123 @@
...
+        'git',
+        'tag',
+        '-a',
+        new_tag,
+        '-m',
+        MESSAGE_TEMPLATE.format(service=service, version=new_version)
+    ], check=True)
+
+    log.info('Pushing tag to origin')
+    subprocess.run(['git', 'push', 'origin', new_tag], check=True)
```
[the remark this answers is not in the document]
mslipper: GHA automatically hides the token.

.github/workflows/tag-service.yml
mslipper: I'll make a new `release-managers` group that can dispatch it.

PR #4901, opened 2023-02-16
Directly test bedrock v0 encoding and hashing against legacy implementations

maurelian:
**Description**

Although it may seem obvious, we were not directly testing that Bedrock's V0 hashing and encoding schemes match those of the legacy system.

**Tests**

Two tests: 
1. `testFuzz_encodeCrossDomainMessage_matchesLegacy_succeeds`
2. `testFuzz_hashCrossDomainMessageV0_matchesLegacy_succeeds`

### DIFF
--- a/packages/contracts-bedrock/contracts/libraries/LegacyCrossDomainUtils.sol
+++ b/packages/contracts-bedrock/contracts/libraries/LegacyCrossDomainUtils.sol
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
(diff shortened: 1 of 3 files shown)


tynes:
Perhaps there is non determinism with weth9 because its pulling in the weth9 from the legacy contracts repo?

tynes:
If this causes build issues by pulling in the dep, I recommend we just copy/paste the function in with a comment including a hyperlink to the git commit that it was copy pasted from, since we are going to delete the contracts package eventually

tynes (approved):

PR #4902, opened 2023-02-16
Make tlsconfig optional for op-signer client

kahuang:
**Description**

Make tls config optional for the op-signer client

**Tests**

Haven't added tests, but did run op-batcher with this config and got it to use the default http client.

**Additional context**

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/op-signer/client/client.go
+++ b/op-signer/client/client.go
@@ -25,35 +25,43 @@ type SignerClient struct {
 }
 
 func NewSignerClient(logger log.Logger, endpoint string, tlsConfig optls.CLIConfig) (*SignerClient, error) {
-	caCert, err := os.ReadFile(tlsConfig.TLSCaCert)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read tls.ca: %w", err)
-	}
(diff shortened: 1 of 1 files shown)


op-signer/client/client.go
```diff
@@ -25,35 +25,41 @@ type SignerClient struct {
 }
 
 func NewSignerClient(logger log.Logger, endpoint string, tlsConfig optls.CLIConfig) (*SignerClient, error) {
-	caCert, err := os.ReadFile(tlsConfig.TLSCaCert)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read tls.ca: %w", err)
-	}
-	caCertPool := x509.NewCertPool()
-	caCertPool.AppendCertsFromPEM(caCert)
+	var httpClient *http.Client
```
tynes, line 28: Could you add a `log.Info` for both cases, just so we know at runtime if tls is being used or not?

kahuang:
@tynes added a log.Info in both cases

zhwrd (approved):

tynes (approved):

PR #4907, opened 2023-02-16
op-batcher: Unit test L2 Reorg Handling

trianglesphere:
**Description**

Adds tests to the channel manager to ensure that it handles L2 reorgs.
I back ported these tests to v0.10.8 of the batcher & confirmed that the
second test properly fails. The first test is a good test to have, and the
second is a regression test against the batcher not handling L2 reorgs properly.

**Metadata**
- Fixes CLI-3326


op-batcher/batcher/driver.go
```diff
@@ -209,6 +215,17 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
 		l.lastStoredBlock = syncStatus.SafeL2.ID()
 	}
 
+	// Ensure that the last stored block is still canonical. Not we need this check in addition to the internal check inside the batch
```
sebastianst, line 218: nit:
```suggestion
	// Ensure that the last stored block is still canonical. Note that we need this check in addition to the internal check inside the batch
```

op-batcher/batcher/driver.go
```diff
@@ -209,6 +215,17 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
+	// queue because if there is a L2 reorg the L2 safe head on the sequence will be stuck until the batch submitter starts submitting again.
+	childCtx, cancel = context.WithTimeout(ctx, networkTimeout)
+	defer cancel()
+	if block, err := l.L2Client.BlockByNumber(childCtx, new(big.Int).SetUint64(l.lastStoredBlock.Number)); err != nil {
+		l.log.Info("Failed to check if the last stored block is canonical", "err", err)
```
sebastianst, line 223: If this query fails, I suggest we `log.Warn` and also return the error directly.
```suggestion
	if block, err := l.L2Client.BlockByNumber(childCtx, new(big.Int).SetUint64(l.lastStoredBlock.Number)); err != nil {
		l.log.Warn("Failed to check if the last stored block is canonical", "err", err, "number", l.lastStoredBlock.Number)
		return eth.BlockID{}, eth.BlockID{}, fmt.Errorf("checking canonical last stored block: %w", err)
```
The next step is querying the blocks on the L2 client, which then probably also fails. Just returning an error here gives the L2 client time to come back online in the next loop tick.

op-batcher/batcher/driver.go
```diff
@@ -209,6 +215,17 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
+	} else if block.Hash() != l.lastStoredBlock.Hash {
+		l.log.Warn("Disccovered that the last stored block is no long canonical (L2 reorg occurred)", "number", block.NumberU64(), "last_hash", l.lastStoredBlock.Hash, "new_hash", block.Hash())
```
sebastianst, line 225: nit
```suggestion
		l.log.Warn("Disccovered that the last stored block is no longer canonical (L2 reorg occurred)", "number", block.NumberU64(), "last_hash", l.lastStoredBlock.Hash, "new_hash", block.Hash())
```

op-batcher/batcher/driver.go
```diff
@@ -148,8 +148,13 @@ func (l *BatchSubmitter) Stop() {
 // 4. Load all new blocks into the local state.
 func (l *BatchSubmitter) loadBlocksIntoState(ctx context.Context) {
 	start, end, err := l.calculateL2BlockRangeToStore(ctx)
-	if err != nil {
-		l.log.Trace("was not able to calculate L2 block range", "err", err)
+	if errors.Is(err, ErrReorg) {
+		l.log.Warn("Found L2 reorg", "err", err)
```
sebastianst, line 152: nit: now also include `err` variable in `log.Warn` in loop below, as is done here

sebastianst (changes requested):

sebastianst:
nevermind, just realized that the root cause of this not working is not found yet.

op-batcher/batcher/channel_manager_test.go
```diff
@@ -0,0 +1,84 @@
...
+		Number:     big.NewInt(1),
+		ParentHash: common.Hash{0xff},
+	}, txs, nil, nil, trie.NewStackTrie(nil))
+
+	err = m.AddL2Block(a)
+	require.Nil(t, err)
+
+	_, _, err = m.TxData(eth.BlockID{})
+	require.Nil(t, err)
+
```
sebastianst, line 81: I suggest we add a second check here to make sure it's drained, like
```suggestion
_, _, err = m.TxData(eth.BlockID{})
	require.ErrorIs(t, err, io.EOF)
```

sebastianst (changes requested):

sebastianst (approved):
lgtm

mslipper (approved):

PR #4909, opened 2023-02-16
contracts-bedrock: L2StandardBridge legacy `withdraw`

tynes:
**Description**

Adds a test for the legacy withdraw functionality on the L2StandardBridge. Previously there were no tests for withdrawing ether via `withdraw`. This adds coverage. Also natspec is added to describe that these functions are not useful for being able to withdraw native L2 tokens. The modern functions must be used to withdraw native L2 tokens, the legacy functions assume that the withdrawal is working on a OptimismMintableERC20 token.

Also regenerate the bindings for the slight change to the bytecode where we save a call in the withdrawal path if we don't need to make the call.

Closes CLI-3333


packages/contracts-bedrock/contracts/L2/L2StandardBridge.sol
```diff
@@ -179,10 +184,10 @@ contract L2StandardBridge is StandardBridge, Semver {
         uint32 _minGasLimit,
         bytes memory _extraData
     ) internal {
-        address l1Token = OptimismMintableERC20(_l2Token).l1Token();
```
smartcontracts, line 182: I don't like the idea of moving this. It confirms that when withdrawing ETH you are referencing the correct token pair.

tynes: I don't understand what you mean here because the branch ensures that as well?

smartcontracts: This specifically guarantees that when using this codepath to withdraw, you specify (OVM_ETH, FakeL1EthTokenAddress) as the token pair and you cannot use (OVM_ETH, RandomToken).

smartcontracts (approved):

PR #4911, opened 2023-02-17
migration: update migrated withdrawal gas limit

tynes:
**Description**

More accurately computes the gas limit for a migrated withdrawal. This is useful to prevent the gaslimit for large transactions from going over the consensus level block gas limit. It is unlikely for a transaction to go over the block gas limit, but this is a preventative fix just in case.

Closes CLI-3336

### DIFF
--- a/op-chain-ops/crossdomain/migrate.go
+++ b/op-chain-ops/crossdomain/migrate.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


op-chain-ops/crossdomain/migrate.go
```diff
@@ -82,8 +83,18 @@ func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger *com
...
+		if b == 0 {
+			dataCost += params.TxDataZeroGas
+		} else {
+			dataCost += params.TxDataNonZeroGasEIP2028
+		}
+	}
+
 	// Set the outer gas limit. This cannot be zero
-	gasLimit := uint64(len(data)*16 + 200_000)
+	gasLimit := dataCost + 200_000
```
smartcontracts, line 97: IMO we should add a cap of ~25m here.

tynes: In the unlikely chance it goes over 25m, then it would create a deposit that cannot be relayed

tynes: Not having that check gives an extra 5 million buffer. Right now we have a loose definition around withdrawals, should there be footguns?

tynes: Ok i see why, will update

tynes:
Addressed the review comment by @smartcontracts, this should be good to merge

maurelian (changes requested):
This all looks about right to me, but I think its worthwhile to update [the tests](https://github.com/ethereum-optimism/optimism/blob/e9045720d0906719f75f60719e5121d69f6feb0e/op-chain-ops/crossdomain/migrate_test.go#L14) to check: 

1. the gas calculation for a string with some mix of 0 and non-zero bytes
2. the gasLimit respects the `25_000_000` boundary

tynes:
@maurelian  There are now tests in both go and js - ready for review again

maurelian (approved):

mslipper (approved):

PR #4912, opened 2023-02-17
contracts-bedrock: prevent deposits with too small gaslimit

tynes:
**Description**

The deposits sent to L2 should have a gas limit that at least covers the intrinsic gas. If the deposits do not pay that much, then it is possible for cause resource usage on L2 without paying for it. This is not a critical bug because the user would still need to pay for L1 inclusion.

Closes CLI-3339


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -371,6 +371,9 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
             );
         }
 
+        // Prevent depositing transactions that have too small of a gas limit.
+        require(_gasLimit >= 21_000, "OptimismPortal: gas limit must cover instrinsic gas cost");
```
smartcontracts, line 375: I think this should really be `21_000 + _data.length` or similar

tynes: I see your point here. I meant to just set a lower bound. It would be more accurate to count the zero and non zero bytes and multiply them by their gas prices and sum the products together. Not sure if all that iteration is worth it, so just adding the data length is a heuristic that underestimates. We could multiply by 16 (way overestimate) or 12 (prob slight overestimate)

tynes: I think that the L1 on chain burn + L1 data cost is sufficient dos protection. It buys simplicity for automation, ie contracts that do deposits dont need to have dynamic gas limit setting

smartcontracts (approved):

PR #4913, opened 2023-02-17
contracts-bedrock: remove crossdomain messenger pausability

tynes:
**Description**

Removes pausability from the messenger and ensures to maintain the storage layout.
This touches a lot of parts of the system. The deploy scripts should get special attention during review.

The storage layout checking script was refactored a bit so that it shows progress when it runs, because it is pretty slow.

**Tests**
Unit tests simply needed to be removed for this feature. The tests covering ownability of the cross domain messengers are removed. The storage layout is preserved, you can look at the layout lock file to see this.

**Additional Context**
This builds on top of https://github.com/ethereum-optimism/optimism/pull/4921 which added pausability to the portal. Now it is expected that the portal can be paused (withdrawals only) and there is no more pausability at the layer of the cross domain messenger. This prevents issues that arise when users try to do withdrawals when the messenger is paused, causing their withdrawals to get stuck.

**Follow Up**
I'd like to unify the spacer contracts in a follow up PR and then add the initializable contract afterwards. This would mean that we only need a single spacer contract and the old initialize storage slot would become a legacy spacer. The `Initializable` contract would follow the spacer contract in the inheritance chain, so that a new storage slot would be used for the new initialized slot. This would allow us to delete the try/catch in the dictator for the call to initializing the portal.


tynes:
This can remove the ownable upgradable from the crossdomain messenger as well

tynes:
Will follow up with notes on how to review this

tynes:
This PR is going to need some changes once https://github.com/ethereum-optimism/optimism/pull/4921 is merged

tynes:
This PR touches the same code as https://github.com/ethereum-optimism/optimism/pull/4919
We might be able to refactor how the spacers work in this PR based if #4919 is merged first

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -28,6 +23,50 @@ contract CrossDomainMessengerLegacySpacer {
...
+     * @notice Spacer for backwards compatibility.
+     */
+    address private spacer_51_0_20;
+
+    /**
+     * @custom:legacy
+     * @custom:spacer __gap
+     * @notice Spacer for backwards compatibility.
+     */
+    uint256[49] private spacer_52_0_1568;
```
clabby, line 53: ibid ^

packages/contracts-bedrock/deploy/021-SystemDictatorSteps-2.ts
```diff
@@ -204,16 +204,15 @@ const deployFn: DeployFunction = async (hre) => {
...
   await doStep({
     isLiveDeployer,
     SystemDictator,
     step: 5,
     message: `
-      Step 5 will initialize all Bedrock contracts but will leave the new L1CrossDomainMessenger
-      paused. After this step is executed, users will be able to deposit and withdraw assets via
-      the OptimismPortal but not via the L1CrossDomainMessenger. The Proposer will also be able to
-      submit L2 outputs to the L2OutputOracle.
+      Step 5 will initialize all Bedrock contracts After this step is executed, the OptimismPortal
```
clabby, line 213: nit
```suggestion
      Step 5 will initialize all Bedrock contracts. After this step is executed, the OptimismPortal
```

packages/contracts-bedrock/deploy/021-SystemDictatorSteps-2.ts
```diff
@@ -319,29 +313,16 @@ const deployFn: DeployFunction = async (hre) => {
...
-    checks: async () => {
-      await assertContractVariable(L1CrossDomainMessenger, 'paused', false)
-    },
-  })
+    await assertContractVariable(OptimismPortal, 'paused', false)
+  }
 
   // At the end we finalize the upgrade.
-  if (await isStep(SystemDictator, 7)) {
+  if (await isStep(SystemDictator, 6)) {
```
clabby, line 321: This if statement and the above can be merged- both rely on the current `SystemDictator` step being `6`.

packages/contracts-bedrock/.storage-layout
```diff
@@ -11,11 +11,11 @@
...
-| __gap                  | uint256[49]              | 52   | 0      | 1568  | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| _paused                | bool                     | 101  | 0      | 1     | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| __gap                  | uint256[49]              | 102  | 0      | 1568  | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| spacer_1_0_1600        | uint256[50]              | 1    | 0      | 1600  | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| spacer_51_0_20         | address                  | 51   | 0      | 20    | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| spacer_52_0_1568       | uint256[49]              | 52   | 0      | 1568  | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| spacer_101_0_1         | bool                     | 101  | 0      | 1     | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| spacer_102_0_1568      | uint256[49]              | 102  | 0      | 1568  | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
```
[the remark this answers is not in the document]
tynes: very astute - so i wanted to do a follow up PR that makes the inheritance chain of this more simple

> I'd like to unify the spacer contracts in a follow up PR and then add the initializable contract afterwards. This would mean that we only need a single spacer contract and the old initialize storage slot would become a legacy spacer. The Initializable contract would follow the spacer contract in the inheritance chain, so that a new storage slot would be used for the new initialized slot. This would allow us to delete the try/catch in the dictator for the call to initializing the portal.

This would make these be spacers. You are right that they currently are not spacers, sorry for the confusion here, it would be more correct if they were labeled as gaps in this PR

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -28,6 +23,50 @@ contract CrossDomainMessengerLegacySpacer {
...
+ * @title CrossDomainMessengerLegacySpacer1
+ * @notice Contract only exists to add a spacer to the CrossDomainMessenger where the
+ *         PausableUpgradable and OwnableUpgradeable variables used to exist. Must be
+ *         the third contract in the inheritance tree of the CrossDomainMessenger.
+ */
+contract CrossDomainMessengerLegacySpacer1 {
+    /**
+     * @custom:legacy
+     * @custom:spacer __gap
+     * @notice Spacer for backwards compatibility.
```
maurelian, line 37: This first gap spacer was pretty confusing to me, apparently it comes from that terrible `ContextUpgradeable` that OwnableUpgradeable` inherits, but it's odd that it comes before Ownable. Could we just add a short comment here to clarify that?

maurelian (approved):

tynes:
@Mergifyio refresh

mslipper (approved):
Go changes LGTM.

PR #4914, opened 2023-02-17
op-chain-ops: Correctly filter withdrawal hashes

mslipper:
Since we're checking every slot in the legacy message passer, we need to pass in the full list of messages to the message passer for verification purposes then filter out messages from senders other than the L2XDM prior to executing the withdrawals migration. This PR fixes the filtering, and adds a check in the withdrawals post-check step to make sure that no potentially fraudulent withdrawals get migrated.

Fixes CLI-3379


op-chain-ops/crossdomain/precheck.go
```diff
@@ -67,7 +67,13 @@ func PreCheckWithdrawals(db *state.StateDB, withdrawals []*LegacyWithdrawal) ([]
 			continue
 		}
 
-		filtered = append(filtered, slotsInp[slot])
+		wd := slotsInp[slot]
+		if wd.Sender == nil || *wd.Sender == predeploys.L2CrossDomainMessengerAddr {
```
smartcontracts, line 71: Should this be `wd.Sender != predeploys.L2CrossDomainMessengerAddr`?

mslipper: Yes. Nice catch. Will unit test these methods to make sure I don't introduce any further issues here.

mslipper:
Converting to draft until I write some unit tests.

smartcontracts (commented):
LGTM but I'll let Mark review

op-chain-ops/crossdomain/precheck.go
```diff
@@ -26,14 +32,15 @@ func PreCheckWithdrawals(db *state.StateDB, withdrawals []*LegacyWithdrawal) ([]
...
 	// Build a mapping of the slots of all messages actually sent in the legacy system.
 	var count int
+	var innerErr error
 	slotsAct := make(map[common.Hash]bool)
 	err := db.ForEachStorage(predeploys.LegacyMessagePasserAddr, func(key, value common.Hash) bool {
 		// When a message is inserted into the LegacyMessagePasser, it is stored with the value
 		// of the ABI encoding of "true". Although there should not be any other storage slots, we
 		// can safely ignore anything that is not "true".
 		if value != abiTrue {
 			// Should not happen!
```
tynes, line 42: Perhaps we should still log here? Since the key/value context is lost?

op-chain-ops/crossdomain/precheck.go
```diff
@@ -26,14 +32,15 @@ func PreCheckWithdrawals(db *state.StateDB, withdrawals []*LegacyWithdrawal) ([]
...
-			log.Error("found unknown slot in LegacyMessagePasser", "key", key.String(), "val", value.String())
+			innerErr = ErrUnknownSlotInMessagePasser
```
tynes, line 43: Any reason to just not return `false` and not use `innerErr`?

tynes: Unless false just stops the iteration and doesn't return an error?

op-chain-ops/crossdomain/precheck.go
```diff
@@ -67,7 +77,13 @@ func PreCheckWithdrawals(db *state.StateDB, withdrawals []*LegacyWithdrawal) ([]
 			continue
 		}
 
-		filtered = append(filtered, slotsInp[slot])
+		wd := slotsInp[slot]
+		if wd.Sender == nil || *wd.Sender != predeploys.L2CrossDomainMessengerAddr {
```
tynes, line 81: Should this be `&&`?

op-chain-ops/crossdomain/precheck_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"testing"
+
+	"github.com/ethereum-optimism/optimism/op-bindings/predeploys"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/trie"
+	"github.com/stretchr/testify/require"
+)
+
```
tynes, line 14: Do you mind adding human readable text to these tests so i can verify against what you say they are supposed to do?

op-chain-ops/crossdomain/precheck.go
mslipper: Correct, `false` just stops iteration and doesn't return an error.

op-chain-ops/crossdomain/precheck.go
mslipper: If we use `&&` this will panic - we want to exit if sender is null or it isn't the `L2XDM`.

op-chain-ops/crossdomain/precheck.go
mslipper: Will wrap the error.

op-chain-ops/crossdomain/precheck_test.go
mslipper: Done.

op-chain-ops/crossdomain/precheck.go
```diff
@@ -53,21 +62,27 @@ func PreCheckWithdrawals(db *state.StateDB, withdrawals []*LegacyWithdrawal) ([]
...
 	for slot := range slotsInp {
 		_, ok := slotsAct[slot]
 		if !ok {
 			log.Info("filtering out unknown input message", "slot", slot.String())
 			continue
 		}
 
-		filtered = append(filtered, slotsInp[slot])
+		wd := slotsInp[slot]
+		if wd.MessageSender != predeploys.L2CrossDomainMessengerAddr {
```
tynes, line 80: Confirming that this is the correct value

op-chain-ops/genesis/check.go
```diff
@@ -498,17 +500,29 @@ func CheckWithdrawalsAfter(db vm.StateDB, data migration.MigrationData, l1CrossD
...
+
+		// If the sender is _not_ the L2XDM, the value should not be migrated.
+		wd := wdsByOldSlot[key]
+		if wd.XDomainSender == predeploys.L2CrossDomainMessengerAddr {
+			// Make sure the value is abiTrue if this withdrawal should be migrated.
+			if migratedValue != abiTrue {
+				innerErr = fmt.Errorf("expected migrated value to be true, but got %s", migratedValue)
+				return false
+			}
+		} else {
```
tynes, line 520: This branch is a good check to add

tynes (approved):

tynes:
Looks good to me

PR #4915, opened 2023-02-17
fix(ctb): initialize L2OO in dictator step 1

smartcontracts:
**Description**
Initializes the L2OutputOracle in step 1 of the SystemDictator. This makes it possible for the proposer to spin up without starting the rest of the system.


mslipper:
🙌 thanks for this.

tynes (approved):

PR #4917, opened 2023-02-17
op-node: update libp2p, fix minor p2p api issues

protolambda:
**Description**

Update libp2p in op-node to latest, and fix minor p2p API issue that I found while updating (some p2p records were not being output correctly in peer-dump response).

Libp2p changed the way the `Host` is initialized to use the Go options pattern, while internally using the `fx.Option` thing of Uber instead of the old reflection hacks. It was a controversial refactor upstream, but still agreed upon as a net improvement, and we do want to keep libp2p updated for any security etc. fixes.

Updating libp2p ensures we can update other dependencies, and use Go 1.20 without warnings. This is required to fix quic/qtls dependency warnings when updating op-geth.

**Tests**

No functionality changes, existing p2p tests still pass with updated host initialization code.

**Metadata**

Fix CLI-3394


op-node/p2p/rpc_server.go
```diff
@@ -134,15 +134,17 @@ func dumpPeer(id peer.ID, nw network.Network, pstore peerstore.Peerstore, connMg
...
+			info.Protocols = append(info.Protocols, string(id))
+		}
 	}
 	// get the first connection direction, if any (will default to unknown when there are no connections)
 	for _, c := range nw.ConnsToPeer(id) {
 		info.Direction = c.Stat().Direction
 		break
 	}
-	if dat, err := pstore.Get(id, "optimismChainID"); err != nil {
+	if dat, err := pstore.Get(id, "optimismChainID"); err == nil {
```
mslipper, line 147: Was this a bug in the old code?

op-node/p2p/rpc_server.go
```diff
@@ -108,19 +108,19 @@ func dumpPeer(id peer.ID, nw network.Network, pstore peerstore.Peerstore, connMg
 			info.NodeID = enode.PubkeyToIDV4((*decredSecp.PublicKey)(typedPub).ToECDSA())
 		}
 	}
-	if dat, err := pstore.Get(id, "ProtocolVersion"); err != nil {
+	if dat, err := pstore.Get(id, "ProtocolVersion"); err == nil {
```
mslipper, line 111: All these equality checks flipped - is that intended?

mslipper (approved):
LGTM pending the one outstanding comment.

op-node/p2p/rpc_server.go
protolambda: yes, they were wrong. If there's no error, then we want to parse the value and output it.

PR #4919, opened 2023-02-17
fix(ctb): Per-message reentrancy guard for `relayMessage`

clabby:
# Overview

Changes the `CrossDomainMessenger`'s `relayMessage` reentrancy guard to be per-message rather than disallowing reentrancy entirely.

**Dislikes about this soln**
~~- The `CrossDomainMessenger` no longer needs to inherit `ReentrancyGuardUpgradeable`, but we would need to alter the storage layouts in order to remove this. Furthers the case for EIP-1967 esq storage slots.~~ EDIT: Added a gap array instead.

**Metadata**
Closes CLI-3388


clabby:
* develop
  * **PR #4919** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4919" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/4919?utm_source=stack-comment).

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -260,13 +299,15 @@ abstract contract CrossDomainMessenger is
...
+    )
+        external
+        payable
+        perMessageNonReentrant(
+            Hashing.hashCrossDomainMessage(_nonce, _sender, _target, _value, _minGasLimit, _message)
+        )
+        whenNotPaused
+    {
         (, uint16 version) = Encoding.decodeVersionedNonce(_nonce);
-        require(
```
tynes, line 265: Why is assertion removed? It should be impossible to prove a >2 message to be honest so it might be safe to remove

clabby: The call to `hashCrossDomainMessage` in the `perMessageNonReentrant` will revert if the version is not `0` or `1`, so it's no longer necessary.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -176,6 +176,45 @@ abstract contract CrossDomainMessenger is
      */
     event FailedRelayedMessage(bytes32 indexed msgHash);
 
+    /**
+     * @notice Modifier for a per-message reentrancy guard.
```
tynes, line 180: Preference to hide this functionality into a library or abstract contract that can be tested independently. Also what about memory safe solidity here?

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -176,6 +176,45 @@ abstract contract CrossDomainMessenger is
...
+     */
+    modifier perMessageNonReentrant(bytes32 _msgHash) {
+        bytes32 _hashMsgHash;
+        assembly {
+            // Re-hash the `_msgHash` with the `0xcafebabe` salt to reduce the possibility
+            // of collisions with existing storage slots.
```
maurelian, line 186: How does this achieve that?

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -176,6 +176,45 @@ abstract contract CrossDomainMessenger is
...
+            }
+            // Trigger the reentrancy lock for `_msgHash`.
+            sstore(_hashMsgHash, 0x01)
+        }
+        _;
+        assembly {
+            // Clear the reentrancy lock for `_msgHash`
+            sstore(_hashMsgHash, 0x00)
+        }
+    }
```
maurelian, line 216: Large assembly blocks should be added within the context of a focussed optimization project.

maurelian (commented):
It's not entirely clear to me that this is the right solution to the problem. Are we sure that we even still need a reentrancy guard here?

What invariants are we trying to preserve?

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
tynes: This is salting the message hash, since its yul and not using a mapping would also essentially salts the key before hashing

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
tynes: I'm personally more ok with yul if the impl is hidden behind a library so that the code itself can be less verbose

clabby: Moved the modifier to a new `ReentrancyGuard` contract.

clabby:
> It's not entirely clear to me that this is the right solution to the problem. Are we sure that we even still need a reentrancy guard here?
> 
> What invariants are we trying to preserve?

The big invariant that we are trying to preserve here is that, if given gas >= the minimum gas limit, a cross domain message will always successfully be relayed if the transaction itself does not revert. Shooting over a DM for more.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -260,13 +262,15 @@ abstract contract CrossDomainMessenger is
         uint256 _value,
         uint256 _minGasLimit,
         bytes calldata _message
-    ) external payable nonReentrant whenNotPaused {
+    )
+        external
+        payable
+        perMessageNonReentrant(
```
smartcontracts, line 268: I think it'd be much simpler if this were just a mapping inside the `CrossDomainMessenger` and didn't require an additional contract to inherit from. Thoughts?

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
smartcontracts: I don't love that this is implicit now. Same comment as above that it'd be much simpler if this were just a mapping like `mapping (bytes32 => bool) public reentrancy`.

smartcontracts:
I would strongly prefer not using assembly here and just stick with a simple mapping.

smartcontracts (dismissed):
Feel strongly enough about not having any assembly here that I'm going to block this for now. Much rather have a simple mapping, especially because we want to reduce the mental overhead down to an absolute minimum for any auditor that's going to be looking at this code before the Bedrock vote.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
smartcontracts: Much rather prefer not having a modifier at all, just put the hash into a mapping and check if the hash is in the mapping already, then clear the mapping afterwards to return the leftover gas.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
clabby: Originally pulled it out at the request of @tynes - Removing the contract in favor of a non-assembly version now.

maurelian:
> The big invariant that we are trying to preserve here is that, if given gas >= the minimum gas limit, a cross domain message will always successfully be relayed if the transaction itself does not revert. Shooting over a DM for more.

We can also satisfy that invariant by removing the reentrancy guard entirely. If we can't do that, then there must be another invariant we need to maintain? Even without assembly code, this adds complexity so we need to ensure that it's justified.

clabby:
> > The big invariant that we are trying to preserve here is that, if given gas >= the minimum gas limit, a cross domain message will always successfully be relayed if the transaction itself does not revert. Shooting over a DM for more.
> 
> We can also satisfy that invariant by removing the reentrancy guard entirely. If we can't do that, then there must be another invariant we need to maintain? Even without assembly code, this adds complexity so we need to ensure that it's justified.

The other very important invariant we need to maintain here is that "cross domain messages should only be able to be successfully relayed once." With the current design, removing the reentrancy guard entirely would allow for relaying a message multiple times before the message hash was assigned to the `successfulMessages` or `failedMessages` mapping. See: [here](https://github.com/ethereum-optimism/optimism/blob/f200fec42fd08fe9a6691ee7d9eb94ad6f25cec0/packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol#L325-L354)

packages/contracts-bedrock/.storage-layout
```diff
@@ -6,25 +6,25 @@
...
-| _initialized       | uint8                    | 0    | 20     | 1     | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| _initializing      | bool                     | 0    | 21     | 1     | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| __gap              | uint256[50]              | 1    | 0      | 1600  | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| _owner             | address                  | 51   | 0      | 20    | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| __gap              | uint256[49]              | 52   | 0      | 1568  | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| _paused            | bool                     | 101  | 0      | 1     | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| __gap              | uint256[49]              | 102  | 0      | 1568  | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
-| _status            | uint256                  | 151  | 0      | 32    | contracts/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
```
tynes, line 19: Do you mind adding a specific spacer for `_status`? Its technically semantically different to have a `__gap[50]` vs a `__gap[49]` + `_status` because the gap indicates that it was never used before. I think status will be set to 1 meaning that we will have dirty storage in a gap which would break an assumption

 It should also make the diff in this file easier to read

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -86,6 +82,18 @@ abstract contract CrossDomainMessenger is
      */
     address public immutable OTHER_MESSENGER;
 
+    /**
+     * @custom:spacer ReentrancyGuardUpgradeable
+     * @notice Spacer for backwards compatibility
+     */
+    uint256 private _status;
```
tynes, line 89: Small nit here, this should follow the spacer syntax and an entry to the storage lockfile should be added. Happy to open a PR to this PR implementing this if unclear how to do this

tynes (commented):
Ideally we can dedupe the tests in the future but fine for now

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
clabby: Done :+1:

tynes (approved):

tynes:
This looks good to me, includes test coverage and is a simple implementation

PR #4921, opened 2023-02-18
feat: portal pausable

tynes:
**Description**

Adds pausability to the `OptimismPortal`
The actor that can pause is the immutable `GUARDIAN` account. This should be set to a multisig or contract that can aribtrate a major bug disclosure.

**Tests**

Includes test coverage of pause + unpause as well as reverts for pause/unpause + deposit, prove and finalize when paused

**Invariants**

- Only the `GUARDIAN` can pause the portal
- When paused, users cannot deposit, prove withdrawals or finalize withdrawals

**Additional context**

- Must be merged alongside https://github.com/ethereum-optimism/optimism/pull/4913
- Follow up PR after this and #4913 are merged to ensure that the dictator/deploy scripts are updated so the system is paused early on
- The `GUARDIAN` is set to the `deployConfig.finalSystemOwner` which is what the owner of the L1xdm was set to


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -130,6 +170,24 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         __ResourceMetering_init();
```
tynes, line 170: For simplicity of the deployment process, add a bool to the initialize function here so that it can be initialized to being paused. In the constructor, call initialize with true so that the impl is always paused, preventing user error

tynes:
For some reason this isn't picking up my latest changes in this view but it is when i go directly to the branch, the latest commit is 0e7f840

tynes:
Going to update the constructor of the portal to accept the paused argument so that it doesn't break the echidna tests, being able to prank in foundry is op

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -110,26 +122,73 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
+    /**
+     * @notice Reverts when paused.
+     */
+    modifier whenNotPaused() {
+        require(paused == false, "OptimismPortal: paused");
+        _;
+    }
+
     /**
      * @custom:semver 1.0.0
```
tynes, line 148: this semver version needs to be updated

tynes:
The codecov report seems a bit nonsensical tbh

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -111,25 +123,74 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
+     * @notice Emitted when the pause is lifted.
+     *
+     * @param account Address of the account triggering the unpause.
+     */
+    event Unpaused(address account);
+
+    /**
+     * @notice Reverts when paused.
+     */
+    modifier whenNotPaused() {
```
clabby, line 142: I like that we're not adding more deps on OZ here, but should we hide this within a utility contract? Or did you implement it within the portal so that the `paused` var is appended to the layout rather than prepended?

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
```diff
@@ -307,7 +316,7 @@ contract SystemDictator is OwnableUpgradeable {
         config.globalConfig.proxyAdmin.upgradeAndCall(
             payable(config.proxyAddressConfig.optimismPortalProxy),
             address(config.implementationAddressConfig.optimismPortalImpl),
-            abi.encodeCall(OptimismPortal.initialize, ())
+            abi.encodeCall(OptimismPortal.initialize, (optimismPortalDynamicConfig))
```
clabby, line 319: Small nit- Generally like to avoid `abi.encodeCall` due to the extra type checking added at runtime. We know that `optimismPortalDynamicConfig` is a bool, so there's no real need.

packages/contracts-bedrock/contracts/test/CommonTest.t.sol
```diff
@@ -208,8 +216,15 @@ contract Messenger_Initializer is L2OutputOracle_Initializer {
...
-        op = new OptimismPortal(oracle, 7 days);
+        op = new OptimismPortal({
+            _l2Oracle: oracle,
+            _guardian: guardian,
+            _finalizationPeriodSeconds: 7 days,
+            _paused: false
+        });
         vm.label(address(op), "OptimismPortal");
+        vm.prank(op.GUARDIAN());
+        op.unpause();
```
clabby, line 227: Portal should already be unpaused, was instantiated w/ `_paused: false` 4 lines above.

packages/contracts-bedrock/deploy/013-OptimismPortalImpl.ts
```diff
@@ -13,12 +13,28 @@ const deployFn: DeployFunction = async (hre) => {
     'L2OutputOracleProxy'
   )
 
+  const finalSystemOwner = hre.deployConfig.finalSystemOwner
+  const finalSystemOwnerCode = await hre.ethers.provider.getCode(
+    finalSystemOwner
+  )
+  if (finalSystemOwnerCode === '0x') {
+    console.log(
+      `WARNING: setting OptimismPortal.GUARDIAN to ${finalSystemOwner} and it has no code`
```
clabby, line 22: Feels like this should hault if done on a live deployment, no?

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
tynes: Oh interesting, I thought that the type checking happened at compile time for `encodeCall`

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: I went back and forth about making a `Pausable` contract and just decided to not do it. Happy to refactor

packages/contracts-bedrock/contracts/test/CommonTest.t.sol
tynes: Ah good catch

packages/contracts-bedrock/deploy/013-OptimismPortalImpl.ts
tynes: Lemme check out the `isLiveDeployer` logic used in the other deployment files

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
clabby: Ah actually you're right about this- mb

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: Slight preference for external since it can be cleanly contained.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -162,7 +223,7 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         uint256 _l2OutputIndex,
         Types.OutputRootProof calldata _outputRootProof,
         bytes[] calldata _withdrawalProof
-    ) external {
+    ) external whenNotPaused {
```
maurelian, line 226: This feels like the conservative approach, but there is an argument to be made that we should allow proofs to continue without delaying finalization.
In case we really want to delay their finalization, we can (at the moment) also use `deleteL2Outputs()` to postpone them.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -361,7 +425,7 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         uint64 _gasLimit,
         bool _isCreation,
         bytes memory _data
-    ) public payable metered(_gasLimit) {
+    ) public payable whenNotPaused metered(_gasLimit) {
```
maurelian, line 428: I have a similar (but stronger) reaction to the comment on `proveWithdrawalTransaction`, ie. I can easily imagine wanting to pause withdrawals but not deposits. 

Not that hard to implement if we can agree on the value of doing so.

packages/contracts-bedrock/contracts/test/CommonTest.t.sol
```diff
@@ -111,6 +112,8 @@ contract L2OutputOracle_Initializer is CommonTest {
     function setUp() public virtual {
         _setUp();
```
maurelian, line 113: While you're in here:
```suggestion
        super.setUp();
```

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -111,25 +123,74 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
+     * @param _guardian                  Address that can pause deposits and withdrawals.
      * @param _finalizationPeriodSeconds Output finalization time in seconds.
+     * @param _paused                    Sets the contract's pausability state.
      */
-    constructor(L2OutputOracle _l2Oracle, uint256 _finalizationPeriodSeconds) Semver(1, 0, 0) {
+    constructor(
+        L2OutputOracle _l2Oracle,
+        address _guardian,
+        uint256 _finalizationPeriodSeconds,
+        bool _paused
```
maurelian, line 159: Very weak preference to keep these related items together:
```suggestion
        L2OutputOracle _l2Oracle,
        uint256 _finalizationPeriodSeconds,
        address _guardian,
        bool _paused
```

Or maybe create a type? 

```
struct PausableConfig {
    address _guardian;
    bool _paused;
}
```

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: The problem here is that it will push the storage layout down and since this is running in prod we can't do that nicely

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: Preference for not a struct so the abi is more clear but get your point on grouping, agreed on that

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: Going to update this to not pause deposits anymore.

I do see your point re: better UX. It feels more simple to me this way and its less tests that need to be written now

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: Going to remove this

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: OK

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: Pardon me for overriding, but just to be safe I'd prefer to resolve once it's done :)

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: We do have [48 slots available](https://github.com/ethereum-optimism/optimism/pull/4921/files/b90a5d5f6e49e8d7e6c2230a7744097eacf41a15#diff-28810b593e6112074b564a947d3e4cc048acacdc3120da28b90081a405df4e41R62) though.

packages/contracts-bedrock/contracts/test/CommonTest.t.sol
tynes: Addressed in this PR https://github.com/ethereum-optimism/optimism/pull/4938

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: I got over this.

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -11,10 +11,83 @@ import { Hashing } from "../libraries/Hashing.sol";
...
+        op.pause();
+
+        assertEq(op.paused(), true);
+    }
+
+    /**
+     * @notice The OptimismPortal reverts when an account that is not the
+     *         GUARDIAN calls `pause()`
+     */
+    function test_pause_onlyOwner_reverts() external {
```
maurelian, line 45: maybe this?

```suggestion
    function test_pause_notGuardian_reverts() external {
```

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -11,10 +11,83 @@ import { Hashing } from "../libraries/Hashing.sol";
...
+        op.unpause();
+
+        assertEq(op.paused(), false);
+    }
+
+    /**
+     * @notice The OptimismPortal reverts when an account that is not
+     *         the GUARDIAN calls `unpause()`
+     */
+    function test_unpause_onlyOwner_reverts() external {
```
maurelian, line 78:
```suggestion
    function test_unpause_notGuardian_reverts() external {
```

maurelian (approved):
Some suggestions but LGTM

tynes:
Fixed naming of tests

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -11,10 +11,83 @@ import { Hashing } from "../libraries/Hashing.sol";
...
+        assertEq(op.paused(), false);
+    }
+
+    /**
+     * @notice The OptimismPortal can be unpaused by the GUARDIAN
+     */
+    function test_unpause_succeeds() external {
+        address guardian = op.GUARDIAN();
+
+        vm.prank(guardian);
```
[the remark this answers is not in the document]
tynes: Slight preference for having close control of caller of all calls, ie the static calls in between don't need to be done by guardian. But not super opinionated, have run into subtle bugs before this way. Not going to push a change because i dont want to trigger ci again

PR #4924, opened 2023-02-21
contracts-bedrock: permit2 enabled bridge

tynes:
**Description**

This is an experimental commit and I do not recommend we merge it until we can get it audited.

This commit integrates usage of `permit2` using its approval flow. It is based on uniswap/permit2 `Permit2Lib.sol` and is a drop in replacement for our existing logic in a backwards compatible way. If `transferFrom` on the token itself doesn't work, it falls back to using `permit2.transferFrom`. This allows for a user to either give an allowance to the bridge or give an allowance to `permit2` and allow `permit2` to interact with the bridge.

Thanks to the uniswap team for feedback on this approach.


packages/contracts-bedrock/contracts/libraries/Permit2Lib.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { console } from "forge-std/console.sol";
```
marktoda, line 6: rm?

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
```diff
@@ -416,7 +418,7 @@ abstract contract StandardBridge {
 
             OptimismMintableERC20(_localToken).burn(_from, _amount);
         } else {
-            IERC20(_localToken).safeTransferFrom(_from, address(this), _amount);
+            IERC20(_localToken).transferFrom2(_from, address(this), _amount);
```
marktoda, line 421: would be nice to optionally be able to pack in a permit signature so users can get 1-tx flow. I.e. can do 

```
IERC20(_localToken).permit2(_from, address(this), _amount, deadline, ...sig)
IERC20(_localToken).transferFrom2(...)
```

This way user can sign permit and just include it with their bridge tx. Would require breaking the external interface or adding a separate entrypoint unless you want to do something wonky like pack it in extraData

marktoda (commented):
some thoughts but stoked to see permit2 integration on the way!

packages/contracts-bedrock/contracts/universal/StandardBridge.sol
tynes: I definitely agree that having a 1 tx flow is where its at. Haven't decided whether the extradata approach or alternative entrypoint is a better approach. The alternate entrypoint requires some refactoring or code duplication, and I didn't want to introduce that into the codebase quite yet.

The extradata is not used for anything yet, was thinking of the possibility of making the first byte of extradata an enum and then abi decoding based on that. Then its extensible into the future. Another idea for the extradata would be including a meta tx to pay out to a relayer that finalizes a withdrawal, but thats even further out into the future

packages/contracts-bedrock/contracts/libraries/Permit2Lib.sol
tynes: yes good catch 😅

PR #4927, opened 2023-02-21
op-chain-ops: better migration handling of legacy withdrawals

tynes:
**Description**

Fixes an issue where permissionless calls of the `LegacyMessagePasser` could cause the migration process to halt due to strict checking of data. It is permissionless to cal the `LegacyMessagePasser`, we want to filter out the other senders and not error on the other senders.


mslipper:
Closing in favor of #4914. Paired with Mark to integrate the change here in that PR.

PR #4930, opened 2023-02-21
op-node: handle sequencer errors with derivation-levels, reset when L1 origins are inconsistent

protolambda:
**Description**

On Goerli we have hit an interesting sequencer edge-case:
- The sequencer conf-depth is lower than the verifier conf-depth
- The L1 reorgs with a reorg-depth in-between these
- The verifier only ever sees the good chain
- The sequencer builds an unsafe L2 block with a L1 origin that doesn't match the canonical chain
- The verifier code-path drops the batch because of the bad L1 origin (epoch) value
- The verifier code-path patiently waits for the sequencer to provide an alternative L2 batch with good origin, because the sequencing window has not elapsed yet.
- The sequencer code-path doesn't undo the unsafe L2 block with bad L1 origin, nor does the verifier code-path

This situation locks up the chain until the sequencing window elapses. To prevent this, we need the sequencer to build a good L2 block, without having to restart it manually to get it back onto an unsafe L2 head with good L1 origin (the work-around we've executed two times now).

Changes:
- Removed unused `idleDerivation`; the sequencing and derivation codepaths are already independent, and handle conflicts through through the shared engine-control
- Added error-handling cases to the `RunNextSequencerAction` to handle when starting/sealing fails due to crit/reset/temp/other error.
- Changed the L1-origin consistency check in the sequencer to a reset-worthy error
- Make `RunNextSequencerAction` bubble up the error if it's critical, it's up to the driver to handle that (in this case log + shutdown driver, since we prefer safety over liveness).

**Tests**

- Added an action-test to run through this edge-case as regression test.
- Updated the sequencer chaos test to simulate these type of errors that require a `Reset()` call to recover from.

**Invariants**

Ensure that the selected L1 origin of every new sequenced L2 block is consistent (i.e. equal or parent-hash of new block matches current) with the previous L1 origin, and execute a reset to make things consistent otherwise.

**Additional context**

Follow-up to Goerli sequencing issue.

**Metadata**

Fix CLI-3404

### DIFF
--- a/op-e2e/actions/l2_sequencer.go
+++ b/op-e2e/actions/l2_sequencer.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/metrics"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
 	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
 	"github.com/ethereum-optimism/optimism/op-node/rollup/driver"
@@ -47,7 +48,7 @@ func NewL2Sequencer(t Testing, log log.Logger, l1 derive.L1Fetcher, eng L2API, c
 	}
 	return &L2Sequencer{
 		L2Verifier:              *ver,
-		sequencer:               driver.NewSequencer(log, cfg, ver.derivation, attrBuilder, l1OriginSelector),
+		sequencer:               driver.NewSequencer(log, cfg, ver.derivation, attrBuilder, l1OriginSelector, metrics.NoopMetrics),
 		mockL1OriginSelector:    l1OriginSelector,
 		failL2GossipUnsafeBlock: nil,
 	}
@@ -121,7 +122,7 @@ func (s *L2Sequencer) ActBuildToL1Head(t Testing) {
 // ActBuildToL1HeadUnsafe builds empty blocks until (incl.) the L1 head becomes the L1 origin of the L2 head
 func (s *L2Sequencer) ActBuildToL1HeadUnsafe(t Testing) {
 	for s.derivation.UnsafeL2Head().L1Origin.Number < s.l1State.L1Head().Number {
-		// Note: the
+		// Note: the derivation pipeline does not run, we are just sequencing a block on top of the existing L2 chain.
 		s.ActL2StartBlock(t)
 		s.ActL2EndBlock(t)
 	}
@@ -144,6 +145,7 @@ func (s *L2Sequencer) ActBuildToL1HeadExcl(t Testing) {
 // ActBuildToL1HeadExclUnsafe builds empty blocks until (excl.) the L1 head becomes the L1 origin of the L2 head, without safe-head progression.
 func (s *L2Sequencer) ActBuildToL1HeadExclUnsafe(t Testing) {
 	for {
+		// Note: the derivation pipeline does not run, we are just sequencing a block on top of the existing L2 chain.
 		nextOrigin, err := s.mockL1OriginSelector.FindL1Origin(t.Ctx(), s.derivation.UnsafeL2Head())
 		require.NoError(t, err)
 		if nextOrigin.Number >= s.l1State.L1Head().Number {
--- a/op-e2e/actions/l2_sequencer.go
+++ b/op-e2e/actions/l2_sequencer.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"context"
+	"errors"
 
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/stretchr/testify/require"
@@ -54,6 +55,10 @@ func NewL2Sequencer(t Testing, log log.Logger, l1 derive.L1Fetcher, eng L2API, c
 
 // ActL2StartBlock starts building of a new L2 block on top of the head
 func (s *L2Sequencer) ActL2StartBlock(t Testing) {
+	s.ActL2StartBlockCheckErr(t, nil)
+}
+
+func (s *L2Sequencer) ActL2StartBlockCheckErr(t Testing, checkErr error) {
 	if !s.l2PipelineIdle {
 		t.InvalidAction("cannot start L2 build when derivation is not idle")
 		return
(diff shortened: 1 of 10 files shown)


op-e2e/actions/l2_sequencer.go
```diff
@@ -103,7 +118,16 @@ func (s *L2Sequencer) ActBuildToL1Head(t Testing) {
 	}
 }
 
-// ActBuildToL1HeadExcl builds empty blocks until (excl.) the L1 head becomes the L2 origin
+// ActBuildToL1HeadUnsafe builds empty blocks until (incl.) the L1 head becomes the L1 origin of the L2 head
+func (s *L2Sequencer) ActBuildToL1HeadUnsafe(t Testing) {
+	for s.derivation.UnsafeL2Head().L1Origin.Number < s.l1State.L1Head().Number {
+		// Note: the
```
tynes, line 124: looks line an unfinished comment here

op-node/rollup/derive/pipeline.go
```diff
@@ -25,6 +25,11 @@ type L1Fetcher interface {
 	L1TransactionFetcher
 }
```
tynes, line 27: nit: godocs

op-node/rollup/driver/metered_engine.go
```diff
@@ -31,9 +31,9 @@ type MeteredEngine struct {
 }
 
 // MeteredEngine implements derive.EngineControl
```
tynes, line 33: nit: doc out of date

op-node/rollup/driver/sequencer.go
```diff
@@ -159,28 +160,54 @@ func (d *Sequencer) BuildingOnto() eth.L2BlockRef {
 // RunNextSequencerAction starts new block building work, or seals existing work,
 // and is best timed by first awaiting the delay returned by PlanNextSequencerAction.
 // If a new block is successfully sealed, it will be returned for publishing, nil otherwise.
-func (d *Sequencer) RunNextSequencerAction(ctx context.Context) *eth.ExecutionPayload {
+// Only critical errors are bubbled up, other errors are handled internally.
+func (d *Sequencer) RunNextSequencerAction(ctx context.Context) (*eth.ExecutionPayload, error) {
 	if _, buildingID, _ := d.engine.BuildingPayload(); buildingID != (eth.PayloadID{}) {
 		payload, err := d.CompleteBuildingBlock(ctx)
 		if err != nil {
```
tynes, line 167: This function is very branchy, perhaps a switch statement will make it more readable. Its also nearly the same logic at the top and the bottom, perhaps some code could be deduplicated?
- adding time to `nextAction` could be turned into an internal function
- the loglines are the same, maybe a function with a callback where the logs happen internal to the function and the callback executes the actions that differ between the top branches? (`buildingID != (eth.PayloadId{})`?
- perhaps a top level `var` for null payload id to do comparisons against

Just suggestions if you like them

op-node/rollup/driver/state.go
```diff
@@ -212,7 +209,11 @@ func (s *Driver) eventLoop() {
 
 		select {
 		case <-sequencerCh:
-			payload := s.sequencer.RunNextSequencerAction(ctx)
+			payload, err := s.sequencer.RunNextSequencerAction(ctx)
+			if err != nil {
+				s.log.Error("Sequencer critical error", "err", err)
```
tynes, line 214: A metric here would be nice

op-node/rollup/driver/sequencer.go
tynes: Also metrics on any types of errors could be useful here

op-e2e/actions/l2_sequencer_test.go
```diff
@@ -100,3 +102,72 @@ func TestL2Sequencer_SequencerDrift(gt *testing.T) {
...
 	require.True(t, engine.l2ForceEmpty, "engine should not be allowed to include anything after sequencer drift is surpassed")
 }
+
+// TestL2Sequencer_SequencerOnlyReorg regression-tests a Goerli halt where the sequencer
+// would build an unsafe L2 block with a L1 origin that then gets reorged out,
+// while the verifier-codepath only ever sees the valid post-reorg L1 chain.
+func TestL2Sequencer_SequencerOnlyReorg(gt *testing.T) {
+	t := NewDefaultTesting(gt)
+	p := &e2eutils.TestParams{
+		MaxSequencerDrift:   20, // larger than L1 block time we simulate in this test (12)
```
tynes, line 112: Is there any way to make an assertion on this in case the L1 block time gets changed in the future?

op-node/rollup/driver/sequencer.go
protolambda: Unfortunately the error checking is using error-targets, not concrete values, so I can't use a switch statement.
And the log-lines are different; the messages indicate which branch we hit.
And I prefer an inline null payload-id check, since it's immutable and a default value.
Good point on metrics, will add those.

op-e2e/actions/l2_sequencer_test.go
protolambda: Will switch to default test params, this isn't strictly necessary, and the defaults should stay maintained as they are used in a lot of the tests.

op-node/rollup/driver/sequencer.go
```diff
@@ -159,28 +169,57 @@ func (d *Sequencer) BuildingOnto() eth.L2BlockRef {
...
-			d.nextAction = d.timeNow().Add(time.Second)
-			if buildingID != (eth.PayloadID{}) { // don't keep stale block building jobs around, try to cancel them
-				d.CancelBuildingBlock(ctx)
+			if errors.Is(err, derive.ErrCritical) {
+				return nil, err // bubble up critical errors.
+			} else if errors.Is(err, derive.ErrReset) {
+				d.log.Error("sequencer failed to seal new block, requiring derivation reset", "err", err)
+				d.metrics.RecordSequencerReset()
+				d.nextAction = d.timeNow().Add(time.Second * time.Duration(d.config.BlockTime)) // hold off from sequencing for a full block
+				if buildingID != (eth.PayloadID{}) {                                            // cancel what we were doing
```
tynes, line 183: Do you need to call `engine.Reset` in this reset block?

tynes: Shouldn't the top check catch this condition? So it doesn't need to be checked here again? Would save some branching

op-node/rollup/driver/sequencer.go
```diff
@@ -159,28 +169,57 @@ func (d *Sequencer) BuildingOnto() eth.L2BlockRef {
 // RunNextSequencerAction starts new block building work, or seals existing work,
 // and is best timed by first awaiting the delay returned by PlanNextSequencerAction.
 // If a new block is successfully sealed, it will be returned for publishing, nil otherwise.
-func (d *Sequencer) RunNextSequencerAction(ctx context.Context) *eth.ExecutionPayload {
+// Only critical errors are bubbled up, other errors are handled internally.
+func (d *Sequencer) RunNextSequencerAction(ctx context.Context) (*eth.ExecutionPayload, error) {
```
tynes, line 173: Would love some comments on expected invariants of this function. Its a really important function in the state machine

protolambda:
@tynes ready for review again

tynes (approved):

PR #4931, opened 2023-02-21
Add e2e tests for Regolith op-geth changes

ajsutton:
**Description**

Adds e2e tests for the op-geth changes from Regolith.

Also introduces a e2e test util `devnet` to avoid needing so much boilerplate code in each e2e test to create the L1 and L2 chains and go through the block creation process.

**Tests**

* Starting regolith at genesis
* Starting regolith after genesis
* Correct gasUsed reported for deposit tx
* Correct ContractAddress for deposit tx that create contracts
* Unused gas from deposit tx can be used by other transactions
* User is not paid ETH for unused deposit tx gas

**Metadata**
- Part of https://linear.app/optimism/issue/CLI-3384/implement-deposits-hf-regolith-op-geth-changes

### DIFF
--- a/op-chain-ops/genesis/config.go
+++ b/op-chain-ops/genesis/config.go
@@ -69,6 +69,9 @@ type DeployConfig struct {
 	L2GenesisBlockParentHash    common.Hash    `json:"l2GenesisBlockParentHash"`
 	L2GenesisBlockBaseFeePerGas *hexutil.Big   `json:"l2GenesisBlockBaseFeePerGas"`
 
+	// Seconds after genesis block that Regolith hard fork activates. 0 to activate at genesis. Nil to disable regolith
+	L2GenesisRegolithTimeOffset *uint64 `json:"l2GenesisRegolithTimeOffset"`
+
 	// Owner of the ProxyAdmin predeploy
 	ProxyAdminOwner common.Address `json:"proxyAdminOwner"`
 	// Owner of the system on L1
--- a/op-chain-ops/genesis/config.go
+++ b/op-chain-ops/genesis/config.go
@@ -70,7 +70,7 @@ type DeployConfig struct {
 	L2GenesisBlockBaseFeePerGas *hexutil.Big   `json:"l2GenesisBlockBaseFeePerGas"`
 
 	// Seconds after genesis block that Regolith hard fork activates. 0 to activate at genesis. Nil to disable regolith
-	L2GenesisRegolithTimeOffset *hexutil.Uint64 `json:"l2GenesisRegolithTimeOffset"`
+	L2GenesisRegolithTimeOffset *hexutil.Uint64 `json:"l2GenesisRegolithTimeOffset,omitempty"`
 
 	// Owner of the ProxyAdmin predeploy
 	ProxyAdminOwner common.Address `json:"proxyAdminOwner"`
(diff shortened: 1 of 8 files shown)


ajsutton:
Note: this is expected to fail horribly currently as the op-geth changes haven't been pulled through. It needs https://github.com/ethereum-optimism/op-geth/pull/45 

Not sure this is following the design ethos of the existing e2e setup. I started by following the pattern of other tests in `op_geth_test.go` but found that required a huge amount of boiler plate code to setup the nodes and go through the various engineAPI calls to import a block which distracted from the logic of the actual test.  So I've pulled out that boiler plate into `devnet.go` which works great for these tests and could likely be adapted a little to be useable by the other `op_geth_test.go` tests.  It has a different feel to things like the `l2_sequencer.go` actions though.

Not sure if this kind of test should be driving via the sequencer more or if `devnet` mostly just needs a bunch of renaming/tweaking to look more like the existing actions. The other option is that these tests should just be added via hive instead (doable but we don't seem to have made them reliable yet).

op-chain-ops/genesis/config.go
```diff
@@ -69,6 +69,9 @@ type DeployConfig struct {
 	L2GenesisBlockParentHash    common.Hash    `json:"l2GenesisBlockParentHash"`
 	L2GenesisBlockBaseFeePerGas *hexutil.Big   `json:"l2GenesisBlockBaseFeePerGas"`
 
+	// Seconds after genesis block that Regolith hard fork activates. 0 to activate at genesis. Nil to disable regolith
+	L2GenesisRegolithTimeOffset *uint64 `json:"l2GenesisRegolithTimeOffset"`
```
tynes, line 73: This should be a `hexutil.Uint64` so that its standard with the rest of the json serialization/deserialization

op-chain-ops/genesis/genesis.go
```diff
@@ -95,6 +96,17 @@ func NewL2Genesis(config *DeployConfig, block *types.Block) (*core.Genesis, erro
 	}, nil
 }
 
+func regolithTime(config *DeployConfig, block *types.Block) *uint64 {
```
tynes, line 99: What do you think about making this a function on the `DeployConfig` itself?

```go
func (cfg *DeployConfig) RegolithTime(block *types.Block) *uint64 {}
```

Might make it more portable + we could easily add a unit test this way in the file that parses deploy config json

op-e2e/devnet.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/log"
+	gn "github.com/ethereum/go-ethereum/node"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/stretchr/testify/require"
+)
+
```
tynes, line 29: Would super appreciate any godocs in this file to make this easy for a future teammate to use

op-chain-ops/genesis/genesis.go
ajsutton: Yeah makes sense to me. I'm not 100% sure if we'll ultimately want to specify RegolithTime as an offset from genesis but it's certainly very helpful in tests currently so seems like a good starting point until we know otherwise.

op-e2e/devnet.go
ajsutton: Yep makes sense, I suspect there'll likely to be a bunch of changes through there to better match the naming of other actions but worth adding godocs before considering it actually ready.

op-node/rollup/derive/l1_block_info.go
```diff
@@ -145,11 +145,17 @@ func L1InfoDeposit(seqNumber uint64, block eth.BlockInfo, sysCfg eth.SystemConfi
...
 
 // L1InfoDepositBytes returns a serialized L1-info attributes transaction.
-func L1InfoDepositBytes(seqNumber uint64, l1Info eth.BlockInfo, sysCfg eth.SystemConfig) ([]byte, error) {
+func L1InfoDepositBytes(seqNumber uint64, l1Info eth.BlockInfo, sysCfg eth.SystemConfig, isRegolith bool) ([]byte, error) {
 	dep, err := L1InfoDeposit(seqNumber, l1Info, sysCfg)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create L1 info tx: %w", err)
 	}
+
+	// TODO: Passing a isRegolith bool here is nasty, but is a quick way to keep e2e tests passing before op-node is updated properly
```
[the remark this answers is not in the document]
ajsutton: Love this bot, but in this case the TODO should be resolved before merging.

ajsutton:
Closing in favour of #4962

PR #4932, opened 2023-02-22
contracts-bedrock: fix standard bridge token pair checking

tynes:
**Description**

This ensures that the legacy interface is only used for legacy tokens and that the modern interface can be used on the OptimismMintableERC20 token.

Note that this is not the gas efficient way to implement this, but its ok for now as we are focused on minimal diff changes.

In the future we could refactor this code to ensure that the ERC165Checker calls a minimal amount of times.


packages/contracts-bedrock/contracts/test/StandardBridge.t.sol
```diff
@@ -0,0 +1,134 @@
...
+    ERC20 internal erc20;
+    LegacyMintable internal legacy;
+
+    function setUp() external {
+        bridge = new StandardBridgeTester({
+            _messenger: payable(address(0)),
+            _otherBridge: payable(address(0))
+        });
+
+        mintable = new OptimismMintableERC20({
```
maurelian, line 84: One thing here is that [OptimismMintableERC20](https://github.com/maurelian/optimism/blob/3261f54860bb91bcd0edc3992c1f2f0a8261142e/packages/contracts-bedrock/contracts/universal/OptimismMintableERC20.sol#L17) inherits from the Legacy token as well, which I think is one of major reasons we missed this compatibility issue, and I worry we're still not testing that properly. 

Could we find a way to make sure that all the StandardBridge tests are run on both tokens?

maurelian: Is there any reason we can't just remove `ILegacyMintableERC20` from `OptimismMintableERC20`?

tynes: > Could we find a way to make sure that all the StandardBridge tests are run on both tokens?

I'd prefer to scope this in another task as its a refactor of our tests if thats cool. We need to dedupe the L1 and L2 tests while we are at it. I'd like this PR to be focused on unit tests around the tested functions

tynes: > Is there any reason we can't just remove ILegacyMintableERC20 from OptimismMintableERC20?

This may cause a headache for offchain tooling, they would need to do the interface checking themselves offchain. I'm not super opposed to it to be honest tho, perhaps we should think about the impacts of the decision before making it

maurelian: > I'd prefer to scope this in another task as its a refactor of our tests if thats cool. 

I do think that in order to be confident that these issues are fully resolved we need to fully test tokens which are exclusively `IOptimismMintableERC20` and `ILegacyMintableERC20`. It's fine with me if that happens in another PR/ticket.

tynes:
@maurelian I've added tests for legacy tokens and included the impl of the legacy token since this has been waiting in review for awhile, I figured I'd just push the commit here

maurelian (approved):

PR #4934, opened 2023-02-22
fix(spec): attributes deposited source hash calculation

ncitron:
**Description**

It seems there is an inconsistency in the `sourceHash` calculation for the attributes deposited transactions between the spec and the actual implementation. Looking at the [implementation code](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/deposit_source.go#L36) the inner keccak should surround both `l1BlockHash` and `seqNumber` rather than just `l1BlockHash`.


maurelian (approved):
LGTM, thanks!

PR #4935, opened 2023-02-22
contracts-bedrock: cleanup tests

tynes:
**Description**

Move some event declarations into higher level contracts and delete a setUp function that was doing nothing but calling super.setUp.

This allows us to more easily reuse the events across different tests.


packages/contracts-bedrock/.gas-snapshot
```diff
@@ -158,7 +158,7 @@ L2OutputOracleTest:test_constructor_badTimestamp_reverts() (gas: 70767)
 L2OutputOracleTest:test_constructor_l2BlockTimeZero_reverts() (gas: 45786)
 L2OutputOracleTest:test_constructor_succeeds() (gas: 33695)
 L2OutputOracleTest:test_deleteL2Outputs_afterLatest_reverts() (gas: 211855)
-L2OutputOracleTest:test_deleteL2Outputs_ifNotChallenger_reverts() (gas: 18883)
```
maurelian, line 161: LOL, I'm so curious what changed. 
Almost tempted to check-in the `ir` output so we can see the results of such things as they happen.

maurelian (approved):

PR #4938, opened 2023-02-22
contracts-bedrock: fix setup chain in tests

tynes:
**Description**
Fixes the foundry tests with the way that `setUp` is used, now its only used when necessary and removes `_setUp`


packages/contracts-bedrock/contracts/test/Semver.t.sol
```diff
@@ -19,7 +19,8 @@ contract Semver_Test is CommonTest {
     /**
      * @notice Deploy a Semver contract
      */
-    function setUp() external {
+    function setUp() public virtual override {
+        super.setUp();
```
maurelian, line 23: Seems we don't _need_ this, is it going to slow things down at all?

maurelian (approved):
1 comment, but LGTM

packages/contracts-bedrock/contracts/test/Semver.t.sol
tynes: There are definitely some that we don't need, happy to follow up with additional cleanup in the future

tynes:
Ok @maurelian I fixed your issue and regenerated the snapshot

Overall these tests need better organization + cleanup

tynes:
Perhaps we can slowly knock out cleaning up individual files

maurelian:
Yeah, that's definitely the move. 

I would like to write issues so that the community can help out. This is good to go tho

PR #4940, opened 2023-02-22
contracts-bedrock: require dynamic config set at correct time

tynes:
The initialization of the L2OutputOracle must happen after the dynamic config is set.

**Description**

Fixes a bug introduced in https://github.com/ethereum-optimism/optimism/commit/654ea9598100dd5823645ad3b70d5e13afcf910a

The L2OO would be initialized with null data

This is going to need changes to the deploy scripts to pass CI


tynes:
Closing this, will reimplement later

PR #4949, opened 2023-02-23
fix(ctb): Add correct prefix to errors

maurelian:
**Description**

Fixes some error messages that were inconsistent with our style guide.


tynes:
Can you bump minor semver for these?

tynes:
Will need to change `check-l2` hh task and regenerate bindings as well

maurelian:
> Can you bump minor semver for these?

Just to be sure: minor, not patch?

tynes (approved):

PR #4950, opened 2023-02-23
op-chain-ops: Fix misleading comment

maurelian:
**Description**

Fixes a misleading comment.

### DIFF
--- a/op-chain-ops/genesis/db_migration.go
+++ b/op-chain-ops/genesis/db_migration.go
@@ -192,7 +192,7 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 	}
 
 	// Finally we migrate the balances held inside the LegacyERC20ETH contract into the state trie.
-	// Note that we do NOT delete the balances from the LegacyERC20ETH contract.
+	// We also delete the balances from the LegacyERC20ETH contract.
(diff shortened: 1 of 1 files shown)


op-chain-ops/genesis/db_migration.go
```diff
@@ -192,7 +192,7 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
 	}
 
 	// Finally we migrate the balances held inside the LegacyERC20ETH contract into the state trie.
-	// Note that we do NOT delete the balances from the LegacyERC20ETH contract.
+	// We also delete the balances from the LegacyERC20ETH contract.
```
maurelian, line 195: Reference for this comment change: 

https://github.com/ethereum-optimism/optimism/blob/a6a14b706d76c91f5cfe0188a2dcb014bc86a936/op-chain-ops/ether/migrate.go#L64-L66

tynes: The contract will return you ETH balance when you call `balanceOf` tho, that was a modification we made

tynes (approved):

op-chain-ops/genesis/db_migration.go
maurelian: right but that's purely in the logic (bytecode) not storage afaiui.

PR #4955, opened 2023-02-23
fix(ctb): Use multiplication of divisors for more precision

maurelian:
In ResourceMetering.sol, the modifier metered() involves 2 consecutive divisions in one of its arithmetic operations prior to having the result assigned to baseFeeDelta.

The double truncation could lead to the calculated value making an additional step lesser than its supposed value.


tynes (approved):

tynes:
Looks like it needs bindings + gas snapshot

PR #4957, opened 2023-02-23
op-e2e: Introduce op-geth e2e test util

ajsutton:
**Description**

Introduce `l2_geth` test util to reduce the boiler plate code required for op-geth tests. Actual test coverage is unchanged currently but this will make it much easier to add additional tests for op-geth specific behaviour.

### DIFF
--- a/op-e2e/op_geth.go
+++ b/op-e2e/op_geth.go
@@ -1,0 +1,203 @@
+package op_e2e
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math/big"
+	"reflect"
+	"testing"
+
+	"github.com/ethereum-optimism/optimism/op-chain-ops/genesis"
+	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils"
+	"github.com/ethereum-optimism/optimism/op-node/client"
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
+	"github.com/ethereum-optimism/optimism/op-node/sources"
+	"github.com/ethereum-optimism/optimism/op-node/testlog"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/log"
+	gn "github.com/ethereum/go-ethereum/node"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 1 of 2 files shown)


op-e2e/l2_geth.go
```diff
@@ -0,0 +1,203 @@
...
+var (
+	// ErrForkChoiceUpdatedNotValid is returned when a forkChoiceUpdated returns a status other than Valid
+	ErrForkChoiceUpdatedNotValid = errors.New("forkChoiceUpdated status was not valid")
+	// ErrNewPayloadNotValid is returned when a newPayload call returns a status other than Valid, indicating the new block is invalid
+	ErrNewPayloadNotValid = errors.New("newPayload status was not valid")
+)
+
+// L2Geth is an actor that functions as a l2 op-geth node
+// It provides useful functions for advancing and querying the chain
+type L2Geth struct {
```
protolambda, line 38: Can we call this OpGeth instead? The legacy pre-bedrock geth fork is also called l2geth

ajsutton: ah good point. Yep, will rename.

ajsutton:
@protolambda if you have a chance to give this a final review I'll follow up with the actual bedrock tests in a separate PR.

op-e2e/op_geth.go
```diff
@@ -0,0 +1,203 @@
...
+	if res.PayloadID == nil {
+		return nil, errors.New("forkChoiceUpdated returned nil PayloadID")
+	}
+	return res, nil
+}
+
+// CreatePayloadAttributes creates a valid PayloadAttributes containing a L1Info deposit transaction followed by the supplied transactions.
+func (d *OpGeth) CreatePayloadAttributes(txs ...*types.Transaction) (*eth.PayloadAttributes, error) {
+	timestamp := d.L2Head.Timestamp + 2
+	l1Info, err := derive.L1InfoDepositBytes(d.sequenceNum, d.L1Head, d.SystemConfig)
```
protolambda, line 182: With https://github.com/ethereum-optimism/optimism/pull/4961 this will need an additional arg to determine if it's for regolith. This testing runs without the rollup-node, so there's no `rollup.Config`, but you can pass in true/false based on the `timestamp` here, or some other test flag.

protolambda (approved):
LGTM, but the L1-info deposit thing will need an extra arg. Can we merge that PR first, and then make the new test-util regolith-ready from the start?

op-e2e/op_geth.go
ajsutton: Yep happy to merge the other PR first.  My regolith draft of this went with passing a boolean there too so easy to add in. It will actually use the Geth ChainConfig to decide if it's regolith or not.

ajsutton: #4961 is merged and I've updated this with the extra param.  It is just false for now as we don't have the updated Geth with regolith config yet. Will update to use the chain config when we start writing regolith e2e tests.

PR #4961, opened 2023-02-23
op-node: regolith fork systemtx change

protolambda:
**Description**

Implement the Regolith `is-system-tx == false` change, and set the l1-info-tx gas to 1M gas.

**Tests**

- Extended attributes builder tests to cover different L1/L2 blocktime activation contexts
- Extended L1 info deposit tests to cover regolith option

This only tests the derivation-pipeline change, which is a relatively small part of Regolith. Once the op-geth changes land and get pulled in as op-node dependency we can implement full e2e-tests (and update existing tests to also run regolith more).

**Metadata**

Fix CLI-3462


ajsutton (approved):
Nice.

op-node/rollup/derive/l1_block_info.go
```diff
@@ -115,7 +119,7 @@ func L1InfoDepositTxData(data []byte) (L1BlockInfo, error) {
 
 // L1InfoDeposit creates a L1 Info deposit transaction based on the L1 block,
 // and the L2 block-height difference with the start of the epoch.
-func L1InfoDeposit(seqNumber uint64, block eth.BlockInfo, sysCfg eth.SystemConfig) (*types.DepositTx, error) {
+func L1InfoDeposit(seqNumber uint64, block eth.BlockInfo, sysCfg eth.SystemConfig, regolith bool) (*types.DepositTx, error) {
```
mslipper, line 122: Perhaps it'd be better to pass the config in directly, in case we need this method to support additional HFs in the future? I've seen this done in Geth in various places.

This comment isn't blocking - if you prefer the way it's done here, I'm happy to defer to your judgement.

mslipper (approved):
One non-blocking question.

op-node/rollup/derive/l1_block_info.go
tynes: Agree on this, less refactoring in the future

tynes: Looks like it would add some overhead to the tests

protolambda: Adding a full config for a single boolean does add a lot of overhead, which I'd like to avoid. This code-path is meant to be simple, it's up to the struct-based utils like the AttributesBuilder to handle the configuration.

PR #4962, opened 2023-02-23
op-e2e: Add e2e tests for bedrock behaviour that will change in regolith

ajsutton:
**Description**

Adds tests to confirm that the behaviour that will change in regolith is unchanged if regolith isn't scheduled.

**Additional context**

Builds on #4957


op-e2e/op_geth_test.go
```diff
@@ -70,3 +71,108 @@ func TestInvalidDepositInFCU(t *testing.T) {
...
+	defer cancel()
+	opGeth, err := NewOpGeth(t, ctx, &cfg)
+	require.NoError(t, err)
+	defer opGeth.Close()
+
+	aliceAddr := cfg.Secrets.Addresses().Alice
+
+	// Deposit TX with a higher gas limit than required
+	depositTx := types.NewTx(&types.DepositTx{
+		SourceHash:          opGeth.L1Head.Hash(),
```
tynes, line 106: This technically isn't the correct way to compute the source hash. It doesn't impact the test but it also is a little confusing

ajsutton: ah interesting.  Previously this was just left nil with a TODO.  I could add a helper function to `OpGeth` to generate this correctly but it would wind up duplicating some code from the `derive` pipeline. So given the SourceHash isn't actually used I'll go back to just omitting `SourceHash`. If it becomes important in the future `OpGeth` could be updated to automatically set it to something reasonable by default which keeps the test code minimal.

tynes (approved):

PR #4963, opened 2023-02-24
op-geth: Update to latest op-geth and add e2e tests

ajsutton:
**Description**

Updates op-geth to bring in the Regolith consensus changes for the execution layer.

Updates the update-op-geth.py script to account for the new geth version and op-wheel being in the main go.mod now.

**Tests**

Adds additional e2e tests for the new Regolith behaviour and updates the bedrock e2e tests to have versions with regolith not scheduled and just not yet active.

**Metadata**
Completes CLI-3475


op-chain-ops/genesis/genesis.go
```diff
@@ -54,6 +54,7 @@ func NewL2Genesis(config *DeployConfig, block *types.Block) (*core.Genesis, erro
 		TerminalTotalDifficulty:       big.NewInt(0),
 		TerminalTotalDifficultyPassed: true,
 		BedrockBlock:                  new(big.Int).SetUint64(uint64(config.L2GenesisBlockNumber)),
+		RegolithTime:                  config.RegolithTime(block.Time()),
```
mslipper, line 57: Will also need to set this in `op-chain-ops/genesis/db_migration.go` as part of the hardfork configs starting on line 283. I'm tracking this for myself in CLI-3384, but should be as easy as a one-line change in this PR.

indexer/go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/ethereum-optimism/optimism/indexer
 
 go 1.17
 
-replace github.com/ethereum/go-ethereum v1.10.26 => github.com/ethereum-optimism/op-geth v0.0.0-20230214215134-401b7fd3309b
+replace (
+	github.com/ethereum/go-ethereum v1.10.26 => github.com/ethereum-optimism/op-geth v0.0.0-20230214215134-401b7fd3309b
```
sebastianst, line 6: In the `require` list below, the version is still at `v1.10.26`. Didn't you update it to `v1.11.2` on purpose?

op-chain-ops/genesis/config.go
```diff
@@ -284,6 +287,17 @@ func (d *DeployConfig) InitDeveloperDeployedAddresses() error {
 	return nil
 }
 
+func (d *DeployConfig) RegolithTime(genesisTime uint64) *uint64 {
+	if d.L2GenesisRegolithTimeOffset == nil {
+		return nil
+	}
+	v := uint64(0)
+	if *d.L2GenesisRegolithTimeOffset > 0 {
+		v = genesisTime + uint64(*d.L2GenesisRegolithTimeOffset)
```
sebastianst, line 296: nit
```suggestion
	if offset := *d.L2GenesisRegolithTimeOffset; offset > 0 {
		v = genesisTime + uint64(offset)
```

op-e2e/op_geth_test.go
```diff
@@ -71,104 +74,474 @@ func TestInvalidDepositInFCU(t *testing.T) {
...
+		regolithTime     hexutil.Uint64
+		activateRegolith func(ctx context.Context, opGeth *OpGeth)
+	}{
+		{name: "ActivateAtGenesis", regolithTime: 0, activateRegolith: func(ctx context.Context, opGeth *OpGeth) {}},
+		{name: "ActivateAfterGenesis", regolithTime: 2, activateRegolith: func(ctx context.Context, opGeth *OpGeth) {
+			_, err := opGeth.AddL2Block(ctx)
+			require.NoError(t, err)
+		}},
+	}
+	for _, test := range tests {
```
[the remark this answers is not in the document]
sebastianst: Curious if semgrep supports this. Anyone knows?

indexer/go.mod
protolambda: We can change it to `v1.11.2` here in the `replace`, but then it also needs to change to `v1.11.2` in the `require`.

sebastianst: I made the comment here because Github doesn't allow me to make comments to non-changed lines ^^ I was referring to the `require` line. We could then remove the redundant replace, it currently mentions two go-ethereum versions.

ajsutton: I think we should leave updating the indexer to a separate PR. Indexer wasn't updated to use the new optimism branch with the latest Geth version.  Just switching op-geth to the new version doesn't work as it also lists explicit versions of op-node, op-service, op-e2e and op-bindings which would need to be updated (and then I don't make sense of the various error messages I'm getting and it feels like something too big to tack onto this PR).

sebastianst: Thanks for clarifying.

op-e2e/op_geth_test.go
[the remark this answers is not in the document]
ajsutton: Added to both table tests in the file. Definitely makes sense.

op-chain-ops/genesis/genesis.go
ajsutton: Yep makes sense. Done.

op-e2e/op_geth_test.go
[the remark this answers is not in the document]
trianglesphere: I'm surprised semgrep/native linters aren't catching this, they usually find loop capture bugs.

ajsutton:
Fixes https://linear.app/optimism/issue/CLI-3399/add-regolith-fork-time-for-mainnet-to-op-geth

indexer/go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/ethereum-optimism/optimism/indexer
...
+	github.com/ethereum/go-ethereum v1.11.2 => github.com/ethereum-optimism/op-geth v1.11.2-aea0402.0.20230227230209-0705cf1b7df9
```
protolambda, line 7: We shouldn't have to specify this replace of geth v1.11.2 if we still use geth v1.10.26 as dependency

protolambda: I'll open a PR to update it, doesn't have to block this

protolambda (approved):
LGTM, we can include the very latest op-geth log related commit (88e99a84819787108bd5fd674d73f1839e4f10bd) later, since it doesn't change functionality for testing

mslipper (approved):

PR #4964, opened 2023-02-24
op-node: fix stale payload-ID in sequencer log

protolambda:
**Description**

Small logging fix: It was logging a zeroed payload-ID instead of the new payload-ID when starting block-building.
This also adds the L2 parent block as log information, to get a better idea of what block is actually being built.


ajsutton (approved):
Yep, makes sense. The parent block info is definitely useful.

protolambda:
Blocked by devnet CI flakiness, waiting for the flakiness fix of: https://github.com/ethereum-optimism/optimism/pull/4965

PR #4965, opened 2023-02-24
ci-devnet: increase devnet params

protolambda:
**Description**

- sequencer time drift to 5 min of L1 blocks
- seq window to 10 min
- channel timeout to 6 min of L1 blocks

- The setup takes 4+ min in the case of the deployed-contracts devnet version
- The batcher cannot handle a sequencer that is catching up to the wallclock: it submits a small batch and waits for confirmation, instead of waiting for more data to submit while knowing the sequencer is catching up to the wallclock time after the start with 4 min lag).

Combined the two issues above caused the old sequencer-window to be hit, causing a reorg, and thus causing smart-contract interaction tests to fail at random places.

The 5 min seq time drift should be plenty, but if it's hit, it can still heal, and has a better chance of doing so with empty L2 blocks. The 10 min seq window should prevent unnecessary reorgs from happening.

The harthat version of the deploy config still uses the old params, since the L1 time is the default 15 seconds instead of the 3 seconds like devnetL1, which effectively makes the sequender-window longer in terms of time span, and thus allows the batcher to submit the batches in time.


mslipper (approved):

mslipper:
Screen Shot 2023-02-23 at 6 38 24 PM

We love to see it. Thanks Proto!

PR #4966, opened 2023-02-24
op-chain-ops: Do not manually copy balance in WipePredeployStorage

mslipper:
`db.CreateAccount` copies over the balance, but wipes nonces. The previous version of this code manually copied the balance after calling `db.CreateAccount`, which effectively doubled the predeploy's balance. None of the L2 predeploys carried balances on Goerli, however to prevent ETH from being created and destroyed during the migration this PR removes the manual balance migration and adds a test to ensure that nonces are copied and state is properly wiped.

Fixes CLI-3443


op-chain-ops/genesis/setters_test.go
```diff
@@ -0,0 +1,50 @@
...
+	require.NoError(t, WipePredeployStorage(stateDB))
+
+	for _, addr := range predeploys.Predeploys {
+		a := *addr
+		if FrozenStoragePredeploys[a] {
+			require.Equal(t, storeVal, stateDB.GetState(a, storeVal))
+		} else {
+			require.Equal(t, common.Hash{}, stateDB.GetState(a, storeVal))
+		}
+		require.Equal(t, big.NewInt(99), stateDB.GetBalance(a))
```
tynes, line 47: Wait maybe i'm reading this wrong, but shouldn't this be checking that the balance is 0? The balance isn't wiped?

mslipper: We're checking that the balance is `99`, which is what we set every predeploy's balance to in the first for loop above. We need to make sure that the balance and nonce are properly carried over.

mslipper: The balance is not wiped by `CreateAccount`, it gets carried over - this is what led to the Sherlock doubling issue.

tynes (approved):

PR #4976, opened 2023-02-24
fix(specs): Corrections to messenger workflow and iface

maurelian:
**Description**

Fixes to messenger specs


specs/messengers.md
```diff
@@ -27,20 +27,35 @@ upgrade.
...
-    function sendMessage(address _target, bytes memory _message, uint32 _minGasLimit) payable external;
-    function successfulMessages(bytes32) view external returns (bool);
-    function xDomainMessageSender() view external returns (address);
+    event SentMessageExtension1(address indexed sender, uint256 value);
+
+    function MESSAGE_VERSION() external view returns (uint16);
+    function MIN_GAS_CALLDATA_OVERHEAD() external view returns (uint64);
+    function MIN_GAS_CONSTANT_OVERHEAD() external view returns (uint64);
+    function MIN_GAS_DYNAMIC_OVERHEAD_DENOMINATOR() external view returns (uint64);
+    function MIN_GAS_DYNAMIC_OVERHEAD_NUMERATOR() external view returns (uint64);
```
maurelian, line 43: Not sure if these should be included tbh.

tynes: I'd like to autogen the interfaces because they fall out of date easily and include links to them here. We don't need to do this now

tynes (approved):

specs/messengers.md
maurelian: Totally. Until we get that in place, I think we should just have a placeholder like "TODO: Add interface here", and then fix that up during a polish before deploying.

PR #4977, opened 2023-02-25
specs: define Regolith upgrade changes to deposit transactions

protolambda:
**Description**

This defines the Regolith upgrade changes to deposits, excluding the exact activation condition of the hardfork.

Changes:
- Clarify the transaction API field additions when clarifying the transaction-type field additions.
- The EVM processing part defines the behavior in terms of differences with standard EVM behavior, to make differences clear and avoid inaccuracies (we do not want to re-specify the whole EVM)
- Define the post-processing with receipt `depositNonce` addition with Regolith.
- Define the gas-used changes of Regolith
- Define the extended Deposit receipt type
- Define the changes to L1-info deposit derivation

This also corrects some inaccuracies that remain from previous pre-Goerli deposit changes:
- The deposit-tx type follows the standard typing pattern. End summer last year I believe we removed the sub-version thing, and the spec still mentioned it.
- The `isSystemTx` boolean was referenced in different parts of the spec, but not in the main deposit transaction definition.
- The contract-deployment case was mentioned, but the processing section was only handling EVM calls.

**Additional context**

See design-doc for additional context. The Regolith hardfork basically addresses inaccuracies in the meta-data of the deposit processing (gas usage in receipt, system-tx metering change, nonce data).

**Metadata**

Fix CLI-3512


specs/deposits.md
```diff
@@ -129,27 +138,47 @@ deposit contract][deposit-contract].
...
-the gas pool on L2. Gas usage exactly matches other transaction types (including intrinsic gas).
-If a deposit runs out of gas or has some other failure, the mint will succeed and the nonce of the
-account will be increased, but no other state transition will occur.
-
-If `isSystemTransaction` in the deposit is set to `true`, the gas used by the deposit is unmetered.
-It must not be subtracted from the gas pool and the `usedGas` field of the receipt must be set to 0.
+The deposit transaction is processed exactly like a type-3 (EIP-1559) transaction, with the exception of:
+
+- No fee fields are verified: the deposit does not have any, as it pays for gas on L1.
+- Not `nonce` field is verified: the deposit does not have any, it's uniquely identified by its `sourceHash`.
```
ajsutton, line 149:
```suggestion
- No `nonce` field is verified: the deposit does not have any, it's uniquely identified by its `sourceHash`.
```

specs/deposits.md
```diff
@@ -129,27 +138,47 @@ deposit contract][deposit-contract].
...
+
+Note that this includes contract-deployment behavior like with regular transactions,
+and gas metering is the same (with the exception of fee related changes above), including metering of intrinsic gas.
+
+Any non-EVM state-transition error emitted by the EVM execution is processed in a special way:
+
+- It is transformed into an EVM-error:
+  i.e. the deposit will always be included, but its receipt will indicate a failure,
+  if it runs into a non-EVM a state-transition error, e.g. failure to transfer the specified
+  `value` amount of ETH due to insufficient account-balance.
```
ajsutton, line 170:
```suggestion
  i.e. the deposit will always be included, but its receipt will indicate a failure
  if it runs into a non-EVM state-transition error, e.g. failure to transfer the specified
  `value` amount of ETH due to insufficient account-balance.
```

ajsutton (approved):
Looks good, just a couple of minor typos.

The only other change in op-geth not mentioned here is that `IsSystemTx: true` is forbidden from Regolith. The only source of that though would be the op-node and it's spec has been updated to say it sets IsSystemTx to false so we probably don't need to mention it.

specs/deposits.md
```diff
@@ -52,8 +53,9 @@ transaction types:
    for the rationale).
 3. They buy their L2 gas on L1 and, as such, the L2 gas is not refundable.
 
-We define a new [EIP-2718] compatible transaction type with the prefix `0x7E`, and then a versioned
-byte sequence. The first version has `0x00` as the version byte and then as the following fields
+We define a new [EIP-2718] compatible transaction type with the prefix `0x7E`.
```
tynes, line 56: perhaps add `to represent a deposit transaction` because its a bit of a jump going from this sentence to the next

protolambda:
Got a 502 bad-gateway on yarn task of integration tests. Will re-run from failed when CI is done

trianglesphere (approved):
lgtm. Really nice docs on this.

PR #4981, opened 2023-02-25
ci: Combine release jobs, add release approval

mslipper:
Combines the Docker build/publish jobs into one, and makes them all depend on a `hold` job that asks for approval. The `hold` step allows someone from the `release-managers` security group to approve the build before it is released. Without this step, tags created by automation tools like GitHub Actions will not be pushed.

I also added automatic release tags to each Docker build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -228,6 +228,7 @@ jobs:
             docker build \
             $(echo -ne $DOCKER_TAGS | tr '\n' ' ') \
             -f <<parameters.docker_file>> \
+            --build-arg VERSION=$CIRCLE_TAG \ 
             <<parameters.docker_context>>
       - run:
           name: Configure Docker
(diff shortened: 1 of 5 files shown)


wecreateproject:
.

.circleci/config.yml
```diff
@@ -199,17 +213,36 @@ jobs:
...
+    machine:
+      image: ubuntu-2204:2022.07.1
+      resource_class: xlarge
     steps:
       - gcp-cli/install
       - gcp-oidc-authenticate
       - checkout
       - run:
-          name: Tag
+          name: Build
```
tynes, line 224: I'm fine with this being inlined but also may be better to have more complex logic in a script and pass in args

tynes (approved):
Generally looks good to me altho complex enough to need to see it run and work to be sure

.circleci/config.yml
mslipper: I can clean this up in separate PRs - need to see it run against develop first.

mslipper:
I need to merge this to develop to actually test it - the OIDC permissions require a full GCP account to be set up. I'll run a test now, and push bugfixes as necessary.

PR #4990, opened 2023-02-26
op-batcher: Add channel duration limiting

sebastianst:
**Description**

Adds a new functionality to the batcher's channel builder to have a limited channel duration. This helps in making sure that a channel is not kept open for too long during times of low L2 tx volume, or to eagerly create batches in test environments.

A limited channel duration is enabled by using the new command line flag `--max-channel-duration` or environment variable `OP_BATCHER_MAX_CHANNEL_DURATION`. If set to `0`, it is disabled, which is the default.

**Tests**

All batcher parameters in e2e and devnet tests have been updated. They are now set to
```go
		MaxChannelDuration:        1,
		MaxL1TxSize:               120_000,
		TargetL1TxSize:            100_000,
		TargetNumFrames:           1,
		ApproxComprRatio:          0.4,
		SubSafetyMargin:           4,
```
in e2e tests, reflecting more realistic values for target & max tx size and the approximate compression ration. Channel duration limiting is enabled with a low value of  `MaxChannelDuration = 1`, making sure that batches are eagerly created.

**Additional context**

Channel duration limiting can be used to balance UX with gas savings. Without it, and proper values for max and target tx size, the batcher will keep channels open for a long time during low L2 tx volume to make the channel as full as possible (while still making sure not to get too close to the channel timeout or end of a sequencing window).

**Metadata**
- Fixes CLI-3468


op-batcher/batcher/config.go
```diff
@@ -47,6 +47,13 @@ type CLIConfig struct {
 	// RollupRpc is the HTTP provider URL for the L2 rollup node.
 	RollupRpc string
 
+	// MaxChannelDuration is the maximum duration (in #L1-blocks) to keep the
+	// channel open. This allows control over how long a channel is kept open
+	// during times of low transaction volume.
```
trianglesphere, line 52: I think this is the best place to explain the behavior that there will be a first frame of a channel every `MaxChannelDuration + TxConfDepth` L1 blocks

trianglesphere (approved):
Changes lgtm. The only thing I'd like added is a comment explaining the interaction between the MaxChannelDuration & the TxConfDepth

sebastianst:
OK suddenly all `TestMixedWithdrawalValidity` tests fail and I don't see a failing assertion. Investigating...

trianglesphere (approved):
lgtm

PR #4992, opened 2023-02-27
op-e2e: Update TestCrossLayerUser test to check deposits and tx work with and without regolith

ajsutton:
**Description**

Updates `TestCrossLayerUser` to run with regolith not scheduled, not yet active, active from genesis or activated after genesis. This provides an effective e2e test for the op-node Regolith changes and that op-node and op-geth work together correctly before, during and after the Regolith fork.

The details of the op-geth changes are better covered by op_geth_test in terms of asserting actual gas used etc.

**Additional context**

Extracted the actual test as a separate method, mostly to make the diff smaller but it also saves the whole test being indented a couple of levels.  Happy to inline it again if that's the preferred style.


op-e2e/actions/user_test.go
```diff
@@ -12,17 +13,42 @@ import (
...
+	futureTime := hexutil.Uint64(20)
+	farFutureTime := hexutil.Uint64(2000)
+	tests := []regolithScheduledTest{
+		{name: "NoRegolith", regolithTime: nil, activateRegolith: false},
+		{name: "NotYetRegolith", regolithTime: &farFutureTime, activateRegolith: false},
+		{name: "RegolithAtGenesis", regolithTime: &zeroTime, activateRegolith: true},
+		{name: "RegolithAfterGenesis", regolithTime: &futureTime, activateRegolith: true},
+	}
+	for _, test := range tests {
+		testCfg := test // Use a fixed reference as the tests run in parallel
```
sebastianst, line 41: They don't run in parallel yet, but you could call `t.Parallel()` below inside `t.Run` to make them.

ajsutton: Turns out they do run in parallel - the call is hidden down in some of the setup calls. It took me a while to work out...

trianglesphere (approved):

PR #4993, opened 2023-02-27
contracts: fix typo in TeleportrDisburser.md

eltociear:
**Description**
availabe -> available


roninjin10 (approved):

tynes:
I believe this is autogenerated based on the natspec, should update there, but i think those contracts have been removed, so if they have been removed we should delete this file

smartcontracts:
> I believe this is autogenerated based on the natspec, should update there, but i think those contracts have been removed, so if they have been removed we should delete this file

Correct, the contracts were removed. I think this PR is good to go but let's also re-run the doc generation to remove any unused files.

PR #5000, opened 2023-02-27
fix(ops): Fix wd-mon name in Dockerfile.packages

maurelian:
**Description**

This PR adds wd-mon to the release process. 

It replicates the treatment of `drippie-mon` since both services are part of the same package and are identical in the way that they are deployed.

My goal is that this PR will create a new repository for the `wd-mon` image on [docker hub](https://hub.docker.com/u/ethereumoptimism). Please review this PR for whether or not this will satisfy that requirement.


trianglesphere:
adding @zhwrd. I'm not super familiar with the github actions release processes.

mslipper (approved):

PR #5005, opened 2023-02-27
feat: OP Stack documentation

smartcontracts:
**Description**
Adds OP Stack documentation to the monorepo.


smartcontracts:
> Does this necessarily need to live in the monorepo? Its just going to bloat the repo size a bunch

Yes. We need to establish a culture where people update docs when the code gets updated and if the docs are in another repo it never happens. We're going to be replacing vuepress with something lighter so the bloat will decrease significantly.

mslipper (approved):

tynes:
> > Does this necessarily need to live in the monorepo? Its just going to bloat the repo size a bunch
> 
> Yes. We need to establish a culture where people update docs when the code gets updated and if the docs are in another repo it never happens. We're going to be replacing vuepress with something lighter so the bloat will decrease significantly.

fwiw i dont think the location of the docs will change a cultural issue, something much better for this would be adding a reminder in the pr template to update docs when necessary. this change alone will be very unlikely to change the cultural issue

PR #5006, opened 2023-02-27
feat(ctb): Do not allow the `CHALLENGER` to delete finalized outputs

clabby:
# Overview

Adds a check to `L2OutputOracle.deleteOutputs` that prevents any finalized outputs from being deleted by the `CHALLENGER`.

Also moves the `FINALIZATION_PERIOD_SECONDS` immutable to the `L2OutputOracle`

**Metadata**
Fixes CLI-3446


clabby:
* develop
  * **PR #5006** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5006" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5006?utm_source=stack-comment).

tynes:
This is going to make the version of the sdk in on the tip of develop not compatible with the version necessary to interact with goerli, we are going to have to be careful doing a release with timing of the upgraded contracts on goerli

packages/contracts-bedrock/contracts/test/CommonTest.t.sol
```diff
@@ -134,7 +134,8 @@ contract L2OutputOracle_Initializer is CommonTest {
             startingBlockNumber,
             startingTimestamp,
             proposer,
-            owner
+            owner,
```
tynes, line 137: slight nit: move towards named function args syntax, no need to do it in this PR if you don't want to but we should be trying to slowly move that direction when time permits :)

packages/contracts-bedrock/contracts/test/L2OutputOracle.t.sol
```diff
@@ -411,6 +414,20 @@ contract L2OutputOracleTest is L2OutputOracle_Initializer {
         vm.expectRevert("L2OutputOracle: cannot delete outputs after the latest output index");
         oracle.deleteL2Outputs(latestOutputIndex - 2);
     }
+
+    function test_deleteL2Outputs_finalized_reverts() external {
+        test_proposeL2Output_proposeAnotherOutput_succeeds();
+
+        // Warp past the finalization period + 1 second
+        vm.warp(block.timestamp + 7 days + 1);
```
tynes, line 422: Can we decouple hardcoded params by calling the finalization period getter here?

clabby: Yeah- will do after retro :100:

tynes:
This release of the sdk should work with goerli: https://github.com/ethereum-optimism/optimism/releases/tag/%40eth-optimism%2Fsdk%401.10.4

.changeset/plenty-kangaroos-march.md
```diff
@@ -0,0 +1,6 @@
+---
+'@eth-optimism/contracts-bedrock': minor
+'@eth-optimism/sdk': minor
+---
```
tynes, line 4: An easy way to prevent this from breaking goerli is a major release

tynes (approved):

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -48,11 +48,6 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
      */
     uint256 internal constant FINALIZE_GAS_BUFFER = 20_000;
 
-    /**
-     * @notice Minimum time (in seconds) that must elapse before a withdrawal can be finalized.
-     */
-    uint256 public immutable FINALIZATION_PERIOD_SECONDS;
```
smartcontracts, line 54: Technically this breaks the API but I'm OK with that because we're not on mainnet yet.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -149,18 +144,15 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
      * @param _guardian                  Address that can pause deposits and withdrawals.
-     * @param _finalizationPeriodSeconds Output finalization time in seconds.
      * @param _paused                    Sets the contract's pausability state.
      */
     constructor(
         L2OutputOracle _l2Oracle,
-        uint256 _finalizationPeriodSeconds,
         address _guardian,
         bool _paused
     ) Semver(1, 1, 0) {
```
smartcontracts, line 153: Should at least get a minor version bump

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -88,7 +93,8 @@ contract L2OutputOracle is Initializable, Semver {
         uint256 _startingBlockNumber,
         uint256 _startingTimestamp,
         address _proposer,
-        address _challenger
+        address _challenger,
+        uint256 _finalizationPeriodSeconds
     ) Semver(1, 1, 0) {
```
smartcontracts, line 98: Needs version bump

.changeset/orange-seas-tie.md
```diff
@@ -0,0 +1,6 @@
+---
+'@eth-optimism/sdk': major
+'@eth-optimism/contracts-bedrock': minor
```
roninjin10, line 3: nit: this is definitely a major bump here too

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -34,6 +34,11 @@ contract L2OutputOracle is Initializable, Semver {
      */
     address public immutable PROPOSER;
 
+    /**
+     * @notice Minimum time (in seconds) that must elapse before a withdrawal can be finalized.
+     */
+    uint256 public immutable FINALIZATION_PERIOD_SECONDS;
```
roninjin10, line 40: Is my understanding correct that this file pretty much the only change we made and every other change in this pr is just a cascading fix based on this change?

roninjin10 (approved):
image

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
clabby: Yeah- the only new code is the check for whether or not the first output that is deleted is finalized or not. Everything else is just moving `FINALIZATION_PERIOD_SECONDS` from the `OptimismPortal` -> `L2OutputOracle` so that it is in the proper spot for future changes. Since the `L2OutputOracle` needed to know about the finalization period, we just did it in this change.

.changeset/orange-seas-tie.md
clabby: This would be the first major release of the `contracts-bedrock` package. There's precedence for API breaking releases that did not warrant a major bump of the package in the past. With this being such a small change + the contracts being pre-prod, this doesn't feel like the right PR to do this in imo. cc @tynes

tynes: This is a major release of the sdk and a minor release of the contracts which should be fine?

tynes: We should do the major release of the contracts with the code freeze

clabby: Yeah, that is compliant with the [README's Versioning section](https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock#versioning).

roninjin10:
Just a nit, but the way changesets are supposed to work is it's a flag that says these things
- this change should not be released until the next major release
- this change is a breaking change

So you might have 10 prs that trigger major but when you actually release only 1 major bump happens.   Since you plan on doing a major release, the way to use changesets idiomatically is to go ahead and label every major breaking change as a major change.   When you publish only 1 major bump will happen.

That said, I get why you don't do this.   The way we publish our changesets is also poorly done and we would have to change that to make the system work.

PR #5008, opened 2023-02-28
contracts-bedrock: ResourceMetering more readable

tynes:
**Description**

Makes the arithmetic in the `ResourceMetering` contract more readable by using explicit argument syntax when calling the `Arithmetic` library functions. This syntax is not used throughout the rest of the codebase, although it should be. It was determined that it would be too large of a change to use this syntax throughout the codebase now, but it is something that we want to do in the future. This adds the syntax to a few key places in `ResourceMetering` which makes it much easier to read and will therefore increase auditability and reviewability of this code.


clabby (approved):

maurelian (approved):

maurelian:
> It was determined that it would be too large of a change to use this syntax throughout the codebase now, but it is something that we want to do in the future

I missed that conversation. 

Looking at this PR, I think it's a very safe change to make, with no affect on the compiler output. It's also easy to constrain to the implementation.

I'd be open to us or an external contributor making these changes one file at a time starting with the most crucial contracts (Bridges, xDMs, Portal, Oracle).

PR #5010, opened 2023-02-28
op-node/chaincfg: Schedule Regolith for Optimism Goerli at 1679079600

ajsutton:
**Description**

Schedules Regolith on Optimism Goerli at 1679079600 which is March 17, 2023 @ 7:00:00 pm UTC / March 17, 2023 12:00 MST

**Additional context**

Also updates op-geth to include https://github.com/ethereum-optimism/op-geth/pull/57 which is the op-geth side of the configuration changes.

**Metadata**
- Fixes https://linear.app/optimism/issue/CLI-3396/add-regolith-fork-time-for-gorli-to-op-geth


ajsutton:
@K-Ho for the record this is the other half of scheduling Regolith on Goerli.

op-node/chaincfg/chains.go
```diff
@@ -39,6 +39,7 @@ var Beta1 = rollup.Config{
 	L1SystemConfigAddress:  common.HexToAddress("0x686df068eaa71af78dadc1c427e35600e0fadac5"),
 }
 
+var goerliRegolithTime = uint64(1679079600)
```
tynes, line 42: since we are going to need to do this multiple times, perhaps define a function

```
func u64Ptr(uint64) *uint64
```

ajsutton: Yep, good idea.  Done.

PR #5017, opened 2023-02-28
fix(ctb): `finalizeWithdrawalTransaction` gas limit dos (v2)

clabby:
# Overview

V2 of #4954, see that PR for the full explanation of the changes. This one implements @smartcontracts' proposed method of breaking out the `callWithMinGas` function into `SafeCall` and testing it independently, which is a good bit cleaner.

Will be followed up with a PR to use the new `callWithMinGas` function in the `CrossDomainMessenger`.

**Metadata**
Fixes CLI-3387
Fixes CLI-3386


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -371,26 +366,16 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
-            _tx.target,
-            gasleft() - FINALIZE_GAS_BUFFER,
-            _tx.value,
-            _tx.data
-        );
+        // call to run out of gas via a returndata bomb. We also want to maintain the property
+        // that the amount of gas supplied to the call to the target contract is at least the gas
+        // limit specified by the user. If there is not enough gas in the callframe to accomplish
+        // this, `callWithMinGas` will revert.
+        bool success = SafeCall.callWithMinGas(_tx.target, _tx.gasLimit, _tx.value, _tx.data);
```
maurelian, line 378: So now we make the call with at least `_tx.gasLimit`, but do we have enough left to handle the gas clean up, event emission and estimation logic?

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
```diff
@@ -34,4 +34,70 @@ library SafeCall {
...
+        uint256 _value,
+        bytes memory _calldata
+    ) internal returns (bool) {
+        bool _success;
+        assembly {
+            // Assertion: gasleft() >= ((_minGas + 200) * 64) / 63
+            //
+            // Because EIP-150 ensures that, at max, 64/64ths of the remaining gas in the call
+            // frame may be passed to a subcontext, we need to ensure that the gas will not be
+            // truncated to hold the invariant. In addition, exactly 51 gas is consumed between
```
maurelian, line 60: not clear what 'the invariant' is referring to here.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: Hmmm... I guess that doesn't matter because if not then the whole thing OOGs and reverts... 
... right?

clabby: Exactly- the `callWithMinGas` function doesn't care about this, if there's not enough gas left, the function will revert anyways.

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
clabby: Referring to the invariants laid out in the invariant documentation for this function- glad to explicitly state in the comments if you'd like :smile:

clabby:
Moving back into draft- docs could use some rewording + tests should preferably account for fluctuating dynamic gas in the external call.

clabby:
* develop
  * **PR #5017** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5017" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5017?utm_source=stack-comment).

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -363,26 +358,16 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
-        bool success = SafeCall.call(
-            _tx.target,
-            gasleft() - FINALIZE_GAS_BUFFER,
-            _tx.value,
-            _tx.data
-        );
+        // call to run out of gas via a returndata bomb. We also want to maintain the property
+        // that the amount of gas supplied to the call to the target contract is at least the gas
+        // limit specified by the user. If there is not enough gas in the callframe to accomplish
+        // this, `callWithMinGas` will revert.
```
maurelian, line 369: This is very nitty, but I'd suggest a small rewrite to the above: 

        // Trigger the call to the target contract. We use a custom low level method
        // SafeCall.callWithMinGas to ensure two key properties
        //   1. Target contracts cannot force this call to run out of gas by returning a very large
        //      amount of data (and this is OK because we don't care about the returndata here). 
        //   2. The amount of gas provided to the call to the target contract is at least the gas
        //      limit specified by the user. If there is not enough gas in the callframe to accomplish
        //      this, `callWithMinGas` will revert.
        // Additionally, if there is not enough gas remaining to complete the execution after the 
        // call returns, this function will revert.

roninjin10 (approved):

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
```diff
@@ -34,4 +34,71 @@ library SafeCall {
...
+        address _target,
+        uint256 _minGas,
+        uint256 _value,
+        bytes memory _calldata
+    ) internal returns (bool) {
+        bool _success;
+        assembly {
+            // Assertion: gasleft() >= ((_minGas + 200) * 64) / 63
+            //
+            // Because EIP-150 ensures that, at max, 64/64ths of the remaining gas in the call
```
maurelian, line 58:
```suggestion
            // Because EIP-150 ensures that a maximum of 63/64ths of the remaining gas in the call
```

maurelian:
Applied the `do-not-merge` label because I'd like this to get 2 code owner reviews.

clabby:
> Applied the `do-not-merge` label because I'd like this to get 2 code owner reviews.

Agree- Will poke Kelvin or Mark about this one if I come across them today 😄

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
```diff
@@ -39,4 +39,117 @@ contract SafeCall_call_Test is CommonTest {
...
+        address from,
+        address to,
+        uint64 minGas,
+        uint64 value,
+        bytes memory data
+    ) external {
+        vm.assume(from.balance == 0);
+        vm.assume(to.balance == 0);
+        // no precompiles
+        vm.assume(uint160(to) > 10);
```
tynes, line 53: There is a helper in forge-std for no precompiles
https://github.com/foundry-rs/forge-std/blob/c2236853aadb8e2d9909bbecdc490099519b70a4/src/StdCheats.sol#L200

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
```diff
@@ -39,4 +39,117 @@ contract SafeCall_call_Test is CommonTest {
...
+    }
+
+    function test_callWithMinGas_noLeakageLow_succeeds() external {
+        SimpleSafeCaller caller = new SimpleSafeCaller();
+
+        for (uint64 i = 5000; i < 50000; i++) {
+            uint256 snapshot = vm.snapshot();
+
+            // 26,069 is the exact amount of gas required to make the safe call
+            // successfully.
```
tynes, line 87: I don't love this magic number here that is susceptible to changes in the gas schedule and compiler version, altho i'm not really sure how much better we can do

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
```diff
@@ -39,4 +39,117 @@ contract SafeCall_call_Test is CommonTest {
...
+        assertEq(from.balance, 0, "from balance is 0");
+        vm.deal(from, value);
+        assertEq(from.balance, value, "from balance not dealt");
+
+        // Bound minGas to [0, l1_block_gas_limit]
+        minGas = uint64(bound(minGas, 0, 30_000_000));
+
+        vm.expectCallMinGas(to, value, minGas, data);
+        vm.prank(from);
+        bool success = SafeCall.callWithMinGas(to, minGas, value, data);
```
tynes, line 73: I feel like nested calls would be a good way to test this because then eip150 comes into play more

tynes (approved):
This is definitely the best solution yet

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
maurelian: I wonder how many of these `vm.assume()` calls are boiler plate that can be moved into a single helper function (something for another PR).

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
```diff
@@ -39,4 +39,117 @@ contract SafeCall_call_Test is CommonTest {
...
+
+    function test_callWithMinGas_noLeakageLow_succeeds() external {
+        SimpleSafeCaller caller = new SimpleSafeCaller();
+
+        for (uint64 i = 5000; i < 50000; i++) {
+            uint256 snapshot = vm.snapshot();
+
+            // 26,069 is the exact amount of gas required to make the safe call
+            // successfully.
+            if (i < 26069) {
```
maurelian, line 88: Nit:
```suggestion
            if (i < 26_069) {
```

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
maurelian: @clabby I think this number is actually important enough to put into Constants.sol

maurelian: hmmm, good question. I need to think about this a bit more...

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
clabby: Nested calls are tested fairly thoroughly in the invariant tests as well as `test_callWithMinGas_noLeakageLow_succeeds` & `test_callWithMinGas_noLeakageHigh_succeeds`

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
maurelian: OK, I think it's fine as is.

packages/contracts-bedrock/contracts/test/invariants/SafeCall.t.sol
```diff
@@ -0,0 +1,123 @@
...
+    Vm internal vm;
+    uint256 public numFailed;
+
+    constructor(Vm _vm) {
+        vm = _vm;
+    }
+
+    function performSafeCallMinGas(uint64 gas, uint64 minGas) external {
+        // Bound the minimum gas amount to [2500, type(uint48).max]
+        minGas = uint64(bound(minGas, 2500, type(uint48).max));
```
maurelian, line 80: where does 2500 come from?

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
```diff
@@ -34,4 +34,71 @@ library SafeCall {
...
+        assembly {
+            // Assertion: gasleft() >= ((_minGas + 200) * 64) / 63
+            //
+            // Because EIP-150 ensures that, a maximum of 63/64ths of the remaining gas in the call
+            // frame may be passed to a subcontext, we need to ensure that the gas will not be
+            // truncated to hold this function's invariant: "If a call is performed by
+            // `callWithMinGas`, it must receive at least the specified minimum gas limit." In
+            // addition, exactly 51 gas is consumed between the below `GAS` opcode and the `CALL`
+            // opcode, so it is factored in with some extra room for error.
+            if lt(gas(), div(shl(0x06, add(_minGas, 0xC8)), 0x3F)) {
```
maurelian, line 64: Can we use `mul` instead of `shl` and decimal instead of hex numbers just so that the logic matches the formula given above?

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
```diff
@@ -14,29 +14,149 @@ contract SafeCall_call_Test is CommonTest {
...
+            }
+
+            assertTrue(vm.revertTo(snapshot));
+        }
+    }
+
+    function test_callWithMinGas_noLeakageHigh_succeeds() external {
+        SimpleSafeCaller caller = new SimpleSafeCaller();
+
+        for (uint256 i = 15_200_000; i < 15_300_000; i++) {
```
maurelian, line 114: I like what you did above better:

```suggestion
        for (uint64 i = 15_200_000; i < 15_300_000; i++) {
```

Then can remove the extra cast in `makeSafeCall` below.

packages/contracts-bedrock/contracts/test/invariants/SafeCall.t.sol
clabby: These numbers are rough- it's a value that is enough to execute the call. Glad to get more exact if you'd like!

packages/contracts-bedrock/contracts/test/invariants/SafeCall.t.sol
```diff
@@ -0,0 +1,123 @@
...
+    function invariant_callWithMinGas_neverForwardsMinGas_reverts() public {
+        assertEq(actor.numSuccessful(), 0, "no successful calls allowed");
+    }
+
+    function performSafeCallMinGas(uint64 minGas) external {
+        SafeCall.callWithMinGas(address(0), minGas, 0, hex"");
+    }
+}
+
+contract SafeCaller_Succeeds_Actor is StdUtils {
```
maurelian, line 70: this contract is nearly identical to the next. Making it more DRY would be good for maintainablility.

maurelian (approved):

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
maurelian: Do nested calls really make a difference? It's always just the same 63/64ths rule for each sub-call.

clabby: @maurelian You're right, but Foundry tests have a default gas limit of `9223372036854775807`. This means for this test specifically, it will *always* have enough gas to satisfy the min gas limit. In the other tests, we limit the amount of gas in the callframe that invokes `callWithMinGas` to a realistic amount.

PR #5018, opened 2023-02-28
Multi architecture docker images

trianglesphere:
**Description**

This builds the `op-node`, `op-batcher`, and `op-proposer` for both `linux/amd64` and `linux/arm64` docker images.

Other docker images can specify different architectures, but unless the dockerfile is updated to use go's cross compile, it does a full emulation during the build which is very slow.

**Tests**

Tested that this build works in CI.

**Metadata**
- Fixes CLI-3531


op-batcher/Dockerfile
```diff
@@ -17,7 +17,11 @@ COPY ./.git /app/.git
 
 WORKDIR /app/op-batcher
 
-RUN make op-batcher VERSION="$VERSION"
+RUN go mod download
+
+ARG TARGETOS TARGETARCH
```
tynes, line 22: Can these have overridable default values?

trianglesphere: These values are always populated by docker.

protolambda (commented):
LGTM, but will defer final approval of docker build changes to dev-infra. cc @mslipper

mslipper (approved):

PR #5034, opened 2023-03-02
fix(atst): return correct typescript types for bytes

roninjin10:
- was returning a string when we should return `0x${string}`
- Makes it so users have to cast it when using it with wagmi

### DIFF
--- a/.changeset/nasty-chairs-agree.md
+++ b/.changeset/nasty-chairs-agree.md
@@ -2,4 +2,4 @@
 '@eth-optimism/atst': minor
 ---
 
-Fix types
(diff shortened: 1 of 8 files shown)


.changeset/nasty-chairs-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'@eth-optimism/atst': minor
+---
+
+Fix types
```
roninjin10, line 5:
```suggestion
Fix string type that should be `0x${string}`
```

jakim929 (approved):

PR #5035, opened 2023-03-02
fix batcher push back frame when tx failed

keroro520:
**Description**

When Batcher fails to submit a frame, it should add the failed frame back to `channelBuilder.frames`. Failure frames have the lowest index, so they should be put at the beginning, not the end, of `channelBuilder.frames`. Batcher will then retry sending the failed frame.

**Tests**

If there are any tests I need to add, please let me know.

### DIFF
--- a/op-batcher/batcher/channel_builder.go
+++ b/op-batcher/batcher/channel_builder.go
@@ -308,13 +308,13 @@ func (c *channelBuilder) NextFrame() (txID, []byte) {
 	return f.id, f.data
 }
 
-// PushFrame adds the frame back to the internal frames queue. Panics if not of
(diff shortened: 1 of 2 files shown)


trianglesphere:
Hi, I don't believe that we need this behavior. Frames do not have to be submitted on L1 in order. They are re-assembled even if out of order in the derivation process.

keroro520:
> Hi, I don't believe that we need this behavior. Frames do not have to be submitted on L1 in order. They are re-assembled even if out of order in the derivation process.

Yes, you are right. The derivation process indeed re-assembles frames into a complete channel. I am going to close this PR.

PR #5047, opened 2023-03-03
op-node/rollup/derive: Check origin of L2 unsafe head match new L1 origin

ajsutton:
**Description**

Each time we reach a new L1 origin in the EngineQueue stage, check it indicates the L2 block is non-canonical and reset if so.

**Tests**

Add tests that drive the engine queue stage and advance the L1 origin to cover each case being checked.

**Metadata**
- Fixes https://linear.app/optimism/issue/CLI-3405/check-l1-origin-of-unsafe-head-when-we-traverse-in-engine-queue-stage

### DIFF
--- a/op-node/rollup/derive/engine_queue.go
+++ b/op-node/rollup/derive/engine_queue.go
@@ -224,7 +224,12 @@ func (eq *EngineQueue) Step(ctx context.Context) error {
 	}
 	outOfData := false
 	if len(eq.safeAttributes) == 0 {
-		eq.origin = eq.prev.Origin()
+		newOrigin := eq.prev.Origin()
+		// Check if the L2 unsafe head origin is consistent with the new origin
+		if err := eq.verifyNewL1Origin(ctx, newOrigin); err != nil {
+			return err
+		}
+		eq.origin = newOrigin
 		eq.postProcessSafeL2() // make sure we track the last L2 safe head for every new L1 block
 		if next, err := eq.prev.NextAttributes(ctx, eq.safeHead); err == io.EOF {
 			outOfData = true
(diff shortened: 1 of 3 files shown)


ajsutton:
`TestL2Sequencer_SequencerOnlyReorg` is failing and it looks like it's deliberately trying to trigger this situation (has a todo for this ticket and everything). It can be updated to pass, though I'm not sure it will still be testing what it originally set out to do - probably a useful e2e test to cover this change though, checking the verifier resets earlier than it had.

ajsutton:
I've pushed an attempt to update the test - I'm not too familiar with the details of how the actor code works so that's worth a detailed review.

op-node/rollup/derive/engine_queue.go
```diff
@@ -224,7 +224,12 @@ func (eq *EngineQueue) Step(ctx context.Context) error {
 	}
 	outOfData := false
 	if len(eq.safeAttributes) == 0 {
-		eq.origin = eq.prev.Origin()
+		newOrigin := eq.prev.Origin()
+		// Check if the L2 safe head origin is consistent with the new origin
+		if err := eq.verifyNewL1Origin(ctx, newOrigin); err != nil {
```
trianglesphere, line 229: This seems odd that when we'll create a new safe head we verify local state against the unsafe head. It seems like it mostly works (two parts: 1st is I think we could get stuck in a spot where we have no safe attributes, but do have an invalid L2 chain, 2nd is accurately checking the canonical nature of the L1 origin).

In general, it seems like we should just be checking if the L1 origin of an unsafe block is canonical or not (& maybe we need to do this on a loop outside of inserting the unsafe blocks).

ajsutton: I'm definitely keen to get @protolambda 's view on it, but from what I can see:
* if we have no safe attributes and an invalid L2 chain, this new verify method will trigger a reset and as part of doing that reset we'll go through `FindL2Heads` which filters the possible L2 unsafe heads and only considers those with a canonical L1 origin. So the reset process will ensure that we get back to having a valid L2 chain
* Checking just the unsafe head when the L1 origin changes makes sense to me because if the unsafe head is built on a canonical L1 origin then the whole L2 chain is too. If it's not we could potentially iterate back to find the latest suitable head but that's what reset is doing. And whether the L2 chain origin is canonical or not can only change when the L1 chain we're tracking changes (either by reorg or advancing) which is the L1 origin advancing. So I think this has the same result as checking each unsafe block but with fewer checks needed, at the cost of doing a full reset if any part of the unsafe chain is no longer built on canonical L1.

op-node/rollup/derive/engine_queue.go
```diff
@@ -224,7 +224,12 @@ func (eq *EngineQueue) Step(ctx context.Context) error {
 	}
 	outOfData := false
 	if len(eq.safeAttributes) == 0 {
-		eq.origin = eq.prev.Origin()
+		newOrigin := eq.prev.Origin()
+		// Check if the L2 safe head origin is consistent with the new origin
```
trianglesphere, line 228:
```suggestion
		// Check if the L2 unsafe head origin is consistent with the new origin
```

op-node/rollup/derive/engine_queue.go
trianglesphere: Ok, I think I was actually confused about this.
If there is no next attributes, then we verify the L1 origin of the L2 *unsafe* head. This is a pretty good place to do this check. The other potential place is when accepting unsafe payloads, but this code path gets hit more often.

op-node/rollup/derive/engine_queue.go
```diff
@@ -245,6 +250,35 @@ func (eq *EngineQueue) Step(ctx context.Context) error {
 	}
 }
 
+func (eq *EngineQueue) verifyNewL1Origin(ctx context.Context, newOrigin eth.L1BlockRef) error {
```
trianglesphere, line 253: nit: godoc

trianglesphere (commented):
I looked a bit closer at the fix & it makes sense to me now.
I just need to review the tests.

trianglesphere (commented):
tests lgtm

trianglesphere (approved):
lgtm

PR #5052, opened 2023-03-04
[op-signer] Populate from address in TransactionArgs

mdehoog:
**Description**

The `op-signer`'s `TransactionArgs` struct that is passed to the `eth_signTransaction` RPC method contains a `From` field, but that field is not populated. This PR fixes that.

**Tests**

No tests added. Tested the functionality in our development nodes.

**Additional context**

We are switching our batcher / proposer to use the op-signer functionality. Rather than hard coding the address on both sides, this is a useful fix so we can share the same signer between multiple callers.


op-service/crypto/signature.go
```diff
@@ -56,10 +57,10 @@ func SignerFactoryFromConfig(l log.Logger, privateKey, mnemonic, hdPath string,
 		fromAddress = common.HexToAddress(signerConfig.Address)
 		signer = func(chainID *big.Int) SignerFn {
 			return func(ctx context.Context, address common.Address, tx *types.Transaction) (*types.Transaction, error) {
-				if address.String() != signerConfig.Address {
+				if !bytes.Equal(address[:], fromAddress[:]) {
```
mdehoog, line 60: this fixes an error:
```
err="attempting to sign for 0x1DEEd254795b3D86b2f980b2fEcecbc8910EA84A, expected 0x1deed254795b3d86b2f980b2fececbc8910ea84a: "
```

mslipper (approved):

op-service/crypto/signature.go
mslipper: nice catch, thanks!

PR #5057, opened 2023-03-06
op-node: Set timeouts on http servers

ajsutton:
**Description**

Sets reasonable timeouts on ppc, metrics and stateviz servers.

Used the default timeouts from Geth RPC since they seem reasonable and are battle-tested.

**Metadata**
- Fixes https://linear.app/optimism/issue/CLI-3345/sherlock-261-http-services-dont-set-timeouts

### DIFF
--- a/op-node/metrics/metrics.go
+++ b/op-node/metrics/metrics.go
@@ -10,7 +10,7 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/ethereum-optimism/optimism/op-node/server"
+	ophttp "github.com/ethereum-optimism/optimism/op-node/http"
 	"github.com/ethereum-optimism/optimism/op-service/metrics"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
@@ -528,7 +528,7 @@ func (m *Metrics) RecordSequencerSealingTime(duration time.Duration) {
 // The server will be closed when the passed-in context is cancelled.
 func (m *Metrics) Serve(ctx context.Context, hostname string, port int) error {
 	addr := net.JoinHostPort(hostname, strconv.Itoa(port))
-	server := server.NewHttpServer(promhttp.InstrumentMetricHandler(
+	server := ophttp.NewHttpServer(promhttp.InstrumentMetricHandler(
 		m.registry, promhttp.HandlerFor(m.registry, promhttp.HandlerOpts{}),
 	))
 	server.Addr = addr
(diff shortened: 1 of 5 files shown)


op-node/server/http.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"net/http"
+
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+// Use default timeouts from Geth as battle tested default values
+var timeouts = rpc.DefaultHTTPTimeouts
+
+func NewHttpServer(handler http.Handler) *http.Server {
```
ajsutton, line 12: I like idea of having a single method that gives a default http.Server instance, including the timeouts. It avoids duplicating code in a bunch of places and if there are other settings that should be applied everywhere in the future we now have a single point for it.

However, I dislike `server` as a module name but `http` conflicts with `net.http` far too much and is worth avoiding given we accept a `http.Handler` as a param. Better suggestions welcome. :)

op-node/metrics/metrics.go
```diff
@@ -528,12 +528,10 @@ func (m *Metrics) RecordSequencerSealingTime(duration time.Duration) {
 // The server will be closed when the passed-in context is cancelled.
 func (m *Metrics) Serve(ctx context.Context, hostname string, port int) error {
 	addr := net.JoinHostPort(hostname, strconv.Itoa(port))
-	server := &http.Server{
-		Addr: addr,
-		Handler: promhttp.InstrumentMetricHandler(
-			m.registry, promhttp.HandlerFor(m.registry, promhttp.HandlerOpts{}),
-		),
-	}
+	server := server.NewHttpServer(promhttp.InstrumentMetricHandler(
```
tynes, line 531: There is variable shadowing here, with `server` being both a variable and a package name. Perhaps we either import the package as `op_server` or call the variable here something else

op-node/server/http.go
tynes: There is a `crypto` package which conflicts with geth's crypto package and we usually import it as `op_crypto` so I'm fine with an `http` package that we import as `op_http`

ajsutton: Ok that works. Renamed back to `http` with the import named `ophttp` (we seem to import crypto as `opcrypto` and avoid snake case).

op-node/metrics/metrics.go
ajsutton: Fixed by renaming module.

mslipper (approved):

PR #5058, opened 2023-03-06
op-chain-ops: Migration fixes

mslipper:
- Simplify import structure
- Add proper checks for invalids torage slots
- Add unit testing
- Remove need to run the TypeScript witness-generation steps


tynes:
This should probably also delete the typescript package for parsing the witness files so that there is a single way to do things?

op-chain-ops/crossdomain/types.go
```diff
@@ -30,3 +33,35 @@ type WithdrawalMessage interface {
 	Hash() (common.Hash, error)
 	StorageSlot() (common.Hash, error)
 }
+
+// InvalidMessage represents a message to the L1 message passer that
+// cannot be decoded as a withdrawal. They are defined as a separate
+// type in order to completely disambiguate them from any other
+// message.
+type InvalidMessage SentMessage
+
```
tynes, line 42: Confirming that this serialization is correct based on the following code: https://github.com/ethereum-optimism/optimism/blob/6f1d8d9ab13d9cb1ef68360f74f82139152ac05d/packages/contracts/contracts/L2/predeploys/OVM_L2ToL1MessagePasser.sol#LL33C54-L33C54

tynes (approved):

mslipper:
> This should probably also delete the typescript package for parsing the witness files so that there is a single way to do things?

Yes. Will do this separately because it will require messing with changesets.

PR #5062, opened 2023-03-06
devnet: Enable Regolith from genesis when deploying the local devnet

ajsutton:
**Description**

op-e2e: Add test to reproduce eth_getLogs failure

**Tests**

**Invariants**

**Additional context**

**Metadata**
- Fixes https://linear.app/optimism/issue/CLI-3555/error-from-eth-getlogs-when-regolith-enabled
- Fixes https://linear.app/optimism/issue/CLI-3561/eth-getlogs-fails-on-pre-bedrock-blocks

### DIFF
--- a/ops-bedrock/Dockerfile.l2
+++ b/ops-bedrock/Dockerfile.l2
@@ -1,4 +1,4 @@
-FROM ethereumoptimism/op-geth:optimism
+FROM us-docker.pkg.dev/oplabs-tools-artifacts/images/op-geth:optimism
 
 RUN apk add --no-cache jq
 
--- a/go.mod
+++ b/go.mod
@@ -191,6 +191,6 @@ require (
 	nhooyr.io/websocket v1.8.7 // indirect
 )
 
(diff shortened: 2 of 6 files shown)


s7v7nislands:
maybe also changed in getting-started.json?

ajsutton:
> maybe also changed in getting-started.json?

Yeah I think that makes sense though probably worth doing in a separate PR just for better visibility.

mslipper (approved):

trianglesphere (approved):

roninjin10 (approved):

ops-bedrock/Dockerfile.l2
```diff
@@ -1,4 +1,4 @@
-FROM ethereumoptimism/op-geth:optimism
+FROM us-docker.pkg.dev/oplabs-tools-artifacts/images/op-geth:optimism
```
tynes, line 1: Is this change to the dockerfile something that we will be planning on keeping?

trianglesphere: yea, we're trying to migrate to `us-docker.pkg.dev` from dockerhub

PR #5064, opened 2023-03-07
contracts-bedrock: prevent overflows in ResourceMetering + prevent grief

tynes:
**Description**

When multiplying a uint64 and uint128 and assigning to a uint256, the solidity checked math applies to the types of the numbers being multiplied. This means that solidity catches an overflow when the result of larger than a uint128 even though its assigned to a uint256.
This PR uses unchecked math to prevent this kind of overflow instead of casting the values to uint256 because it is safe to use unchecked math here. The following inequality always holds true, which shows that it is safe:

``` solidity
type(uint64).max * type(uint128).max < type(uint256).max
```

Update:
The above unchecked logic was removed in favor of lowering the max deposit base fee.  The values are simply type casted to `uint256` when doing the math to prevent overflows instead of doing unchecked math.

Test coverage is added for max resource consumption with a max deposit base fee as well as some test coverage for invariants that must be held for the system to work properly, such as the max base fee being larger than the min base fee and the initial base fee being in between the min and the max.

The invariant must be held that the user is able to request the max amount of resources when the base fee is at its max value. If the max base fee is too large, then it will result in more L1 gas being consumed than the L1 block gas limit. Test coverage is added to ensure that the block gas limit can compute a deposit when the max base fee is set.

Update 2:
The max resource amount has been updated to 20 million with the same target resources, meaning that the base fee will ratchet up much more quickly. This keeps the gas prices the same for when we are around the target and they will increase much more aggressively when there is high usage. This will make it much more expensive to continuously frontrun a deposit and buy the max resources, causing it to revert.

The CSV file generated can be found in google sheets [here](https://docs.google.com/spreadsheets/d/1-ON9USnaQVzAN87lcFw-pkizYXP4MOy9MIG6aiKz5rE/edit#gid=61306804)

It is a known issue that the deposit base fee can rise to a point where it is impossible to buy more resources on L2. It should be an invariant that you can buy the max resources no matter what the deposit base fee is, but we cannot address that before going to mainnet. We need to eventually add an alternative deposit path where the L2 gas can be purchased by sending ether instead of with a gas burn. That is too large of a project to take on before mainnet, but that is how the invariant can be held.

The following data was taken from the csv file

|prevBaseFee|prevBoughtGas|prevBlockNumDiff|l1BaseFee|requestedGas|gasConsumed|ethPrice|usdCost|success|
|-|-|-|-|-|-|-|-|-|														
|50000000000|0|0|50000000000|20000000|20000318|$1600|$1600|success
|50000000000|0|0|1000000000|100000|5000370|$1600|$8|success

The amount of gas burnt is dynamic based on the current L1 base fee to hold user fees constant. You can see that it costs $1600 when the eth price is $1600 to buy the max resource limit when the deposit base fee is 50 gwei. You can see when the L1 base fee is 1 gwei and the user requests 100000 gas, it results in 5000370 gas being used. This is a 50:1 ratio.


packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -134,7 +134,14 @@ abstract contract ResourceMetering is Initializable {
...
         // Determine the amount of ETH to be paid.
-        uint256 resourceCost = _amount * params.prevBaseFee;
+        // Safety: _amount is a uint64
+        //         params.prevBaseFee is a uint128
+        //         resourceCost is a uint256
+        //         type(uint64).max * type(uint128).max < type(uint256).max
+        uint256 resourceCost;
+        unchecked {
+            resourceCost = _amount * params.prevBaseFee;
+        }
```
sebastianst, line 144: Still curious, why did it then throw an overflow error? Is it because `_amount` is a `uint64` and `params.prevBaseFee` is a `uint128` and the solc performs the calculation `_amount * params.prevBaseFee` first in `uint128` and only then converts into the target type `uint256`? Because in that case we'd need to do
```suggestion
        uint256 resourceCost = uint256(_amount) * uint256(params.prevBaseFee);
```
instead. But I don't know the solc that well, so no idea :)

sebastianst: Do we have tests that would fail in case the overflow calculation would indeed only happen in `uint128`?

tynes: Its possible to do the casting, but there is no real need to do so because it will result in checked math which will bloat the bytecode and cause it to use slightly more gas. This is why I opted into using an unchecked block, since we have a guarantee that there will not be an overflow

Re: tests, I do have another commit locally on top of this commit that I am working on to thoroughly test to ensure that all possible values can be represented, I broke this change out into its own commit so that it would be easier to review on its own. It is reproducible in chisel. I'll see if i can make a unit test for this specifically without a large diff

tynes:
I've added an additional commit that includes a table test for testing the resource params at a bunch of different values and then it writes the output to csv. It includes the USD cost to buy up deposit gas

packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -54,11 +54,16 @@ abstract contract ResourceMetering is Initializable {
 
     /**
      * @notice Maximum base fee value, cannot go higher than this.
+     *         This value must be small enough to allow a user to request the
+     *         MAX_RESOURCE_LIMIT when the base fee is at its max value.
+     *         Setting this value too large can result in more gas being
+     *         consumed than the L1 block gas limit.
      */
-    int256 public constant MAXIMUM_BASE_FEE = int256(uint256(type(uint128).max));
+    int256 public constant MAXIMUM_BASE_FEE = int256(uint256((type(uint32).max / 7) * 6));
```
maurelian, line 62: should mul before div here I think

maurelian: How does the math work here again? 7 and 6 kind of feel like magic numbers.

packages/contracts-bedrock/foundry.toml
```diff
@@ -19,6 +19,9 @@ ffi = true
 fuzz_runs = 16
 no_match_contract = 'EchidnaFuzz'
 
+fs_permissions = [
+  { 'access'='read-write', 'path'='./' },
```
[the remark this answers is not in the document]
tynes: Good catch

packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -50,17 +51,21 @@ abstract contract ResourceMetering is Initializable {
...
+    int256 public constant MINIMUM_BASE_FEE = 1 gwei;
 
     /**
      * @notice Maximum base fee value, cannot go higher than this.
+     *         It is possible for the MAXIMUM_BASE_FEE to raise to a value
+     *         that is so large it will consume the entire gas limit of
+     *         an L1 block.
      */
-    int256 public constant MAXIMUM_BASE_FEE = int256(uint256(type(uint128).max));
+    int256 public constant MAXIMUM_BASE_FEE = int256(uint256((type(uint128).max)));
```
smartcontracts, line 62: Double parens seems unnecessary

packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -84,6 +89,18 @@ abstract contract ResourceMetering is Initializable {
         // Run the underlying function.
         _;
 
+        // Run the metering function.
+        _metered(_amount, initialGas);
+    }
+
+    /**
+
```
smartcontracts, line 97: Extra whitespace

smartcontracts (approved):
lgtm minus comments

maurelian (approved):

clabby (approved):
LGTM

PR #5065, opened 2023-03-07
op-node: when confirm payload failed should cancel it

s7v7nislands:
**Description**

When confirm payload failed should cancel it

**Additional context**

I think we should refactor the code, and split the derivation and sequencer build block, BuildingPayload() in some cases will be a conflict between these two modes.


protolambda (approved):
LGTM

mslipper:
@mergifyio dequeue

PR #5066, opened 2023-03-07
op-node: remove unused function

s7v7nislands:
**Description**

Remove unused function LastL2Time()


trianglesphere (approved):
nice find. This looks like it was dead the moment it was introduced.

PR #5069, opened 2023-03-07
batch_decoder: Reassemble Channels

trianglesphere:
**Description**

This pulls all frames into a channel structure. It uses the same code as derivation
to determine if a channel is valid. It then decodes the batches from the channel.

**Metadata**
- Fixes CLI-3559


protolambda (approved):
Awesome cli util

PR #5073, opened 2023-03-07
feat(atst): Updated SDK docs

qbzzt:


tynes:
Can you squash this into a single commit?

packages/atst/docs/sdk.md
```diff
@@ -26,246 +41,248 @@ yarn
...
-await writeAttestation(preparedTx)
-```
+For react hooks we recomend using the [wagmi cli](https://wagmi.sh/cli/getting-started) with the [etherscan plugin](https://wagmi.sh/cli/plugins/etherscan) and [react plugin](https://wagmi.sh/cli/plugins/react) to automatically generate react hooks around the attestation station.
 
-### getEvents
+Use `createKey` and `createValue` to convert your raw keys and values into bytes that can be used in the attestation station contract calls.
+Use `parseString`, `parseBool`, `parseAddress` and `parseNumber` to convert values returned by attestation station to their correct data type.
 
-To getEvents use getEvents with a provider and any filters to filter the event
+For convenience we also export the hooks here:
```
roninjin10, line 277:
```suggestion
For convenience we also [export the hooks](../src/react.ts) here:
```

roninjin10 (approved):
Yea should squash the commits

PR #5074, opened 2023-03-07
op-node/derivation: JSON Encoding for Frames

trianglesphere:
**Description**

JSON Encode & Decode frames.

### DIFF
--- a/op-node/rollup/derive/params.go
+++ b/op-node/rollup/derive/params.go
@@ -1,7 +1,7 @@
 package derive
 
 import (
-	"encoding/json"
+	"encoding/hex"
 	"errors"
 	"fmt"
 )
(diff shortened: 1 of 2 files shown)


op-node/rollup/derive/params.go
```diff
@@ -49,3 +51,19 @@ func (id ChannelID) TerminalString() string {
...
+	h, err := hex.DecodeString(s)
+	if err != nil {
+		return err
+	}
+	if len(h) != ChannelIDLength {
+		return errors.New("invalid length")
+	}
+	copy(id[:], h)
+	return nil
+}
```
protolambda, line 69: Implementing MarshalText and UnmarshalText (see https://pkg.go.dev/encoding#TextMarshaler) may be simpler, since it avoids the json-quoting work. And the JSON golang code will handle it as a string then.

trianglesphere: I implemented this & it worked like a charm.

protolambda (approved):

PR #5075, opened 2023-03-08
batch_decoder: Force close channels

trianglesphere:
**Description**

Adds a sub-command to create force close transactions.

**Metadata**
- Fixes CLI-3564


protolambda (approved):

op-node/rollup/derive/channel_out.go
```diff
@@ -213,3 +213,58 @@ func BlockToBatch(block *types.Block) (*BatchData, error) {
...
+	}
+
+	var out bytes.Buffer
+	out.WriteByte(DerivationVersion0)
+
+	if !closed {
+		f := Frame{
+			ID:          id,
+			FrameNumber: 0,
+			Data:        nil,
```
tynes, line 246: the type uses `[]byte` instead of `hextil.Bytes` , resulting in base64 encoded data produced by this tool, i think we should consider using gencodec for creating json marshalling code

trianglesphere: IMO it's not a huge problem. The data is a chunk of a compressed stream. It's meaningless until reassembled and decompressed.

PR #5076, opened 2023-03-08
contracts-bedrock: make ResourceMetering params private

tynes:
**Description**

The `ResourceMetering` params should be private so that other contracts that inherit it do not set the params by accident. This is important to ensure safety of the system.

The echida tests needed to be updated as well as places that the resource params were set. A backwards compatible getter was added to ensure that no tooling will break.


tynes:
I'm ok with closing this if people think its too big of a change @maurelian

maurelian:
yeah, tbh I kind of think we can not do this and minimize the diff.

PR #5078, opened 2023-03-08
feat(ctb): Print tenderly simulation URLs during deployment

maurelian:
**Description**

Facilitates deployments on Goerli and Mainnet by printing simulation page URLs.

Manually tested.

### DIFF
--- a/packages/contracts-bedrock/deploy/021-SystemDictatorSteps-2.ts
+++ b/packages/contracts-bedrock/deploy/021-SystemDictatorSteps-2.ts
@@ -13,6 +13,7 @@ import {
   jsonifyTransaction,
   isStep,
   doStep,
+  getTenderlySimulationLink,
 } from '../src/deploy-utils'
 
 const deployFn: DeployFunction = async (hre) => {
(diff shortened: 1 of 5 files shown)


packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -314,20 +316,24 @@ export const isStep = async (
...
+  hre: HardhatRuntimeEnvironment,
+  opts: {
+    isLiveDeployer?: boolean
+    SystemDictator: ethers.Contract
+    step: number
+    message: string
+    checks: () => Promise<void>
+  }
+): Promise<void> => {
   if (!(await isStep(opts.SystemDictator, opts.step))) {
```
tynes, line 337: You have a provider with the system dictator contract, can use that to query the chainid instead of passing thru the hre through multiple layers in many places

packages/contracts-bedrock/deploy/021-SystemDictatorSteps-2.ts
```diff
@@ -123,7 +124,7 @@ const deployFn: DeployFunction = async (hre) => {
   })
 
   // Step 4 transfers ownership of the AddressManager and L1StandardBridge to the ProxyAdmin.
-  await doStep({
+  await doStep(hre, {
```
tynes, line 127: You should be able to remove passing through `hre` now?

maurelian: Good catch fixed.

tynes:
Looks good to me pending rebase

maurelian:
rebased

tynes (approved):

PR #5082, opened 2023-03-08
contracts-bedrock: remove deployment ownable check l1xdm

tynes:
**Description**

The new `L1CrossDomainMessenger` is no longer ownable so we no longer need to transfer the ownership to the dictator as part of the system upgrade. The dictator does not need ownership of the old contract because it doesn't call `pause` on the `L1CrossDomainMessenger` anymore as part of the on chain upgrade. Instead, the pausable functionality is in the `OptimismPortal` which is set with a constructor arg and its pausability can be toggled using functions on the portal.


mslipper:
Fixes CLI-3850

maurelian (approved):

PR #5083, opened 2023-03-08
op-batcher: Refactor frame & tx data handling

sebastianst:
**Description**

Introduces two separate types for frame data and tx data to make a clear
distinction between the two. Also prepares the batcher for future changes
when it will support sending multiple frames per tx.

**Tests**

Added a regression test to validate that requeued transactions have the same data (and don't prepend 0 version bytes) and that transaction data can be parsed correctly.

I added a random L2 block generator. Because of dependency cycles, I couldn't put it into package `op-node/testutils` since it uses a function from the `derive` package. Hence added the random generator to a new package `op-node/rollup/derive/test`, following the pattern to create testing code dependent on a package in a `.../test` sub-package.

**Additional context**

Cleaner solution of #5050

### DIFF
--- a/op-batcher/batcher/tx_data.go
+++ b/op-batcher/batcher/tx_data.go
@@ -1,0 +1,54 @@
+package batcher
+
+import (
+	"fmt"
+
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
+)
+
+// txData represents the data for a single transaction.
+//
+// Note: The batcher currently sends exactly one frame per transaction. This
+// might change in the future to allow for multiple frames from possibly
+// different channels.
+type txData struct {
+	frame frameData
+}
+
+// ID returns the id for this transaction data. It can be used as a map key.
+func (td *txData) ID() txID {
+	return td.frame.id
+}
+
+// Bytes returns the transaction data. It's a version byte (0) followed by the
+// concatenated frames for this transaction.
+func (td *txData) Bytes() []byte {
+	return append([]byte{derive.DerivationVersion0}, td.frame.data...)
+}
+
+// Frame returns the single frame of this tx data.
+//
+// Note: when the batcher is changed to possibly send multiple frames per tx,
+// this should be changed to a func Frames() []frameData.
+func (td *txData) Frame() frameData {
+	return td.frame
+}
+
+// txID is an opaque identifier for a transaction.
+// It's internal fields should not be inspected after creation & are subject to change.
+// This ID must be trivially comparable & work as a map key.
+//
+// Note: transactions currently only hold a single frame, so it can be
+// identified by the frame. This needs to be changed once the batcher is changed
+// to send multiple frames per tx.
+type txID = frameID
+
+func (id txID) String() string {
+	return fmt.Sprintf("%s:%d", id.chID.String(), id.frameNumber)
+}
+
+// TerminalString implements log.TerminalStringer, formatting a string for console
+// output during logging.
+func (id txID) TerminalString() string {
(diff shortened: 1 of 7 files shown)


op-batcher/batcher/tx_data.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// ID returns the id for this transaction data. It can be used as a map key.
+func (td *txData) ID() txID {
+	return td.frame.id
+}
+
+// Bytes returns the transaction data.
+func (td *txData) Bytes() []byte {
+	return append([]byte{0}, td.frame.data...)
```
s7v7nislands, line 18: should we add the explanation of []byte{0}?

sebastianst: good idea, added some more commentary to `txData` and its methods!

op-batcher/batcher/tx_data.go
```diff
@@ -0,0 +1,52 @@
...
+
+// ID returns the id for this transaction data. It can be used as a map key.
+func (td *txData) ID() txID {
+	return td.frame.id
+}
+
+// Bytes returns the transaction data. It's a version byte (0) followed by the
+// concatenated frames for this transaction.
+func (td *txData) Bytes() []byte {
+	return append([]byte{0}, td.frame.data...)
```
trianglesphere, line 24: do we want to use `derive.DerivationVersion0` or even a local constant?

op-node/rollup/derive/test/random.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
+	"github.com/ethereum-optimism/optimism/op-node/testutils"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/trie"
+)
+
+// RandomL2Block returns a random block whose first transaction is a random
+// Deposit transaction.
```
trianglesphere, line 14:
```suggestion
// L1 Info Deposit transaction.
```

op-node/rollup/derive/test/random.go
```diff
@@ -0,0 +1,23 @@
...
+func RandomL2Block(rng *rand.Rand, txCount int) (*types.Block, []*types.Receipt) {
+	l1Block := types.NewBlock(testutils.RandomHeader(rng),
+		nil, nil, nil, trie.NewStackTrie(nil))
+	l1InfoTx, err := derive.L1InfoDeposit(0, l1Block, eth.SystemConfig{}, false)
```
trianglesphere, line 18: I think this works. It's probably a little tricky to move the `L1InfoDeposit` to a base package

op-batcher/batcher/tx_data.go
sebastianst: Yes we should, changed! I don't like magic numbers myself 🪄

op-node/rollup/derive/test/random.go
sebastianst: Yes there's a 🐰 🕳️ of moving some types & functions into different packages as to not have to import `rollup/derive` everywhere. But that seemed too much to handle in this PR 😅

op-node/rollup/derive/test/random.go
sebastianst: Added. Also randomized the `regolith` bool.

trianglesphere (approved):
lgtm

PR #5085, opened 2023-03-08
op-node/service: Deduplicate logger + Better defaults

trianglesphere:
**Description**

This PR does the following

- Add `logfmt` as a supported log format. It just needed to be exposed to the command line.
- Remove the op-node's log config & use op-service's version instead.
- Change the default logger to be use `logfmt` instead of `terminal with color`
- Change the behavior of `text` to mean `terminal` if it is a terminal or `logfmt` if there is no terminal.
  - The default logger is still `text`.
- Added a log formatted version of the startup banner. It will still pretty-print if running in terminal mode.
  - To do this I had the move where the startup banner was logged to earlier on in the startup process.

**Rationale**

One of the tradeoffs that we need to make while logging is between having enough information to effectively debug problems & not providing information overload to end users or emitting too many logs which takes up disk space.

I have kept terminal logging the same for users that are directly running in a terminal, but for users that are running in docker or another non-terminal environment I switched the default to `logfmt`. 

The `logfmt` format can easily be parsed by logging infrastructure while it remains more human readable than JSON. It also includes full hashes instead of the abridged terminal hash which is critical for being able to debug complex issues. One example is by abridging transaction hashes in the batcher I am not able to easily determine which transactions got included on chain and which did not. It is important that this is the default for external users because it means that when we get logs they will have enough information to be actionable. If home users are running in docker, the logs remain human readable if they need to look through them.

Log Format Example log line
```
t=2023-03-08T18:25:38+0000 lvl=info msg="Rollup Config"            l2_chain_id=901 l2_network="unknown L2" l1_chain_id=900 l1_network="unknown L1" l2_start_time=1,678,299,758 l2_block_hash=0x645df1804fa827dc9eae1d661501a827e128ebae1060bef7b11bb0506cf28686 l2_block_number=0 l1_block_hash=0x25348ccc26d94ae5ac0e2e2ac5c10270d4b21cbdab8fb8f937f73c507e2a08dd l1_block_number=0 regolith_time="(not configured)"
```

**Tests**

Tested manually.


op-service/log/cli.go
```diff
@@ -28,13 +28,13 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
-			Usage:  "Format the log output. Supported formats: 'text', 'json'",
+			Usage:  "Format the log output. Supported formats: 'text', 'terminal', 'logfmt', 'json', 'json-pretty',",
 			Value:  "text",
 			EnvVar: opservice.PrefixEnvVar(envPrefix, "LOG_FORMAT"),
 		},
 		cli.BoolFlag{
 			Name:   ColorFlagName,
-			Usage:  "Color the log output",
+			Usage:  "Color the log output if in terminal mode",
 			EnvVar: opservice.PrefixEnvVar(envPrefix, "LOG_COLOR"),
```
ajsutton, line 38: Previously logs would default to using colour when outputting to the terminal but now it requires specifically specifying `--log.color`.

ajsutton (commented):
I like the new non-terminal output - definitely helpful to have it be parsable and more detailed when not going to a terminal.

op-service/log/cli.go
tynes: I think that is fine, we would only really want colors if running locally right?

tynes (approved):

op-service/log/cli.go
ajsutton: I think if we're outputting to a local terminal we should default to using color, otherwise the color functionality will generally not be used.  Definitely agree if it's outputting to logs or anything non-terminal that defaulting to no color makes sense.

trianglesphere: I believe that it should still use colors by default (even without the flag) when running in a terminal

trianglesphere: oh, I see. This was a bug in the op-service version, but not in the op-node version

trianglesphere: Fixed in 3f0d2bd6111acbb32e3a2806b4b2c1dc28cac3df

op-node/cmd/main.go
```diff
@@ -104,6 +101,13 @@ func RollupNodeMain(ctx *cli.Context) error {
 		return err
 	}
 
+	// Only pretty-print the banner if it is a terminal log. Other log it as key-value pairs.
+	if logCfg.Format == "terminal" {
+		log.Info("rollup config:\n" + cfg.Rollup.Description(chaincfg.L2ChainIDToNetworkName))
+	} else {
+		cfg.Rollup.LogDescription(log, chaincfg.L2ChainIDToNetworkName)
+	}
+
```
protolambda, line 110: To avoid duplicate code I'm fine with just always calling `LogDescription` instead

trianglesphere: Is this something that you want to do?

mslipper (approved):

PR #5088, opened 2023-03-09
fix(fd): use correct fn for getTotalElements

smartcontracts:
**Description**
Fault detector was using L2OutputOracle.latestOutputIndex() to query the total number of outputs submitted, which was off by one. Correct function is L2OutputOracle.nextOutputIndex(). Caused crashes of the fault detector on startup if no outputs had been submitted to the L2OO yet.

### DIFF
--- a/packages/fault-detector/src/service.ts
+++ b/packages/fault-detector/src/service.ts
@@ -124,7 +124,7 @@ export class FaultDetector extends BaseServiceV2<Options, Metrics, State> {
       this.state.oo = {
         contract: oo,
         filter: oo.filters.OutputProposed(),
-        getTotalElements: async () => oo.latestOutputIndex(),
+        getTotalElements: async () => oo.nextOutputIndex(),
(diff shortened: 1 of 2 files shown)


packages/fault-detector/src/service.ts
```diff
@@ -124,7 +124,7 @@ export class FaultDetector extends BaseServiceV2<Options, Metrics, State> {
       this.state.oo = {
         contract: oo,
         filter: oo.filters.OutputProposed(),
-        getTotalElements: async () => oo.latestOutputIndex(),
+        getTotalElements: async () => oo.nextOutputIndex(),
```
smartcontracts, line 127: Note the off-by-one that would cause an underflow and an error on system startup: https://github.com/ethereum-optimism/optimism/blob/576a8fcbfe0239df5691392af0613cd43ba09608/packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol#L305-L307

tynes (approved):

PR #5098, opened 2023-03-09
CI: Split contracts bedrock

trianglesphere:
**Description**

Splits the contracts test into several smaller components.


.circleci/config.yml
```diff
@@ -66,9 +66,38 @@ jobs:
...
       - check-changed:
           patterns: op-bindings,op-chain-ops,packages/
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - yarn-packages-v2-{{ checksum "yarn.lock" }}
       - run:
           name: Install dependencies
-          command: yarn && git diff --exit-code
+          command: yarn install --imutable
```
maurelian, line 75: I don't think that `--imutable` does anything. We're still on an old version of [yarn](https://github.com/ethereum-optimism/optimism/blob/35d4a37c2b9a59e9bf16fa4c5e22283f2348c3d2/.yarnrc#L1), and I think ugprading is non-trivial.

.circleci/config.yml
```diff
@@ -66,9 +66,38 @@ jobs:
...
+      - save_cache:
+          name: Save Yarn Package Cache
+          key: yarn-packages-v2-{{ checksum "yarn.lock" }}
```
maurelian, line 78: I assume this might also be an issue with the yarn version we use.

trianglesphere: why would the cache be an issue?

maurelian: TBH IDK what this is actually doing, it's just the `v2` that's triggering me.

.circleci/config.yml
maurelian: yarn v2 might be less bad, not sure.

PR #5105, opened 2023-03-10
op-batcher: enable graceful shutdown, closing current channel

BrianBland:
**Description**

This change is intended to enable graceful shutdowns for the `BatchSubmitter` when the `Stop` method is called.

In order to shut down safely, this first closes any current pending channel, then publishes any remaining frames on the channel back to the L1. Once the channel has been completely closed and submitted, the submitter loop should exit without loading any new L2 blocks into its local state.

One caveat is that a repeated failure to submit frames back to the L1 may cause this loop to hang even when a shutdown is desired, as this prioritizes channel completion over shutdown speed. This can be worked around by passing a channel into the `Stop` method with a reasonable deadline, as is now the case in at least one e2e test.

**Tests**

Added tests for various channel manager `Close()` behaviors, including "close before first use", "close with no pending channel", "close with pending channel", and "close when all transactions are failed"

**Invariants**

**Additional context**

**Metadata**
- Fixes CLI-3244

### DIFF
--- a/op-batcher/batcher/channel_builder.go
+++ b/op-batcher/batcher/channel_builder.go
@@ -18,6 +18,7 @@ var (
 	ErrMaxDurationReached    = errors.New("max channel duration reached")
 	ErrChannelTimeoutClose   = errors.New("close to channel timeout")
 	ErrSeqWindowClose        = errors.New("close to sequencer window timeout")
+	ErrTerminated            = errors.New("channel terminated")
 )
 
 type ChannelFullError struct {
@@ -188,7 +189,7 @@ func (c *channelBuilder) Reset() error {
 }
 
 // AddBlock adds a block to the channel compression pipeline. IsFull should be
-// called aftewards to test whether the channel is full. If full, a new channel
+// called afterwards to test whether the channel is full. If full, a new channel
 // must be started.
 //
 // AddBlock returns a ChannelFullError if called even though the channel is
@@ -314,9 +315,10 @@ func (c *channelBuilder) IsFull() bool {
 //     would have been exceeded by the latest AddBlock call,
 //   - ErrMaxFrameIndex if the maximum number of frames has been generated
 //     (uint16),
-//   - ErrMaxDurationReached if the max channel duration got reached.
-//   - ErrChannelTimeoutClose if the consensus channel timeout got too close.
-//   - ErrSeqWindowClose if the end of the sequencer window got too close.
+//   - ErrMaxDurationReached if the max channel duration got reached,
+//   - ErrChannelTimeoutClose if the consensus channel timeout got too close,
+//   - ErrSeqWindowClose if the end of the sequencer window got too close,
+//   - ErrTerminated if the channel was explicitly terminated.
 func (c *channelBuilder) FullErr() error {
 	return c.fullErr
 }
@@ -400,6 +402,16 @@ func (c *channelBuilder) outputFrame() error {
 	c.frames = append(c.frames, frame)
 	c.outputBytes += len(frame.data)
 	return err // possibly io.EOF (last frame)
+}
+
+// Close immediately marks the channel as full with an ErrTerminated
+// if the channel is not already full. This ensures that no additional
+// frames will be added to the channel.
+func (c *channelBuilder) Close() error {
+	if !c.IsFull() {
+		c.setFullErr(ErrTerminated)
+	}
+	return c.FullErr()
 }
 
 // HasFrame returns whether there's any available frame. If true, it can be
--- a/op-batcher/batcher/channel_builder.go
+++ b/op-batcher/batcher/channel_builder.go
@@ -308,8 +308,8 @@ func (c *channelBuilder) IsFull() bool {
 // FullErr returns the reason why the channel is full. If not full yet, it
 // returns nil.
 //
-// It returns a ChannelFullError wrapping one of six possible reasons for the
-// channel being full:
+// It returns a ChannelFullError wrapping one of the following possible reasons
+// for the channel being full:
 //   - ErrInputTargetReached if the target amount of input data has been reached,
 //   - derive.MaxRLPBytesPerChannel if the general maximum amount of input data
 //     would have been exceeded by the latest AddBlock call,
@@ -405,12 +405,11 @@ func (c *channelBuilder) outputFrame() error {
 }
 
 // Close immediately marks the channel as full with an ErrTerminated
-// if the channel is not already full, then outputs any remaining frames.
-func (c *channelBuilder) Close() error {
+// if the channel is not already full.
+func (c *channelBuilder) Close() {
 	if !c.IsFull() {
 		c.setFullErr(ErrTerminated)
 	}
-	return c.closeAndOutputAllFrames()
 }
 
 // HasFrame returns whether there's any available frame. If true, it can be
--- a/op-batcher/batcher/channel_builder.go
+++ b/op-batcher/batcher/channel_builder.go
@@ -405,13 +405,12 @@ func (c *channelBuilder) outputFrame() error {
 }
 
 // Close immediately marks the channel as full with an ErrTerminated
-// if the channel is not already full. This ensures that no additional
-// frames will be added to the channel.
+// if the channel is not already full, then outputs any remaining frames.
 func (c *channelBuilder) Close() error {
 	if !c.IsFull() {
 		c.setFullErr(ErrTerminated)
 	}
-	return c.FullErr()
+	return c.closeAndOutputAllFrames()
 }
 
 // HasFrame returns whether there's any available frame. If true, it can be
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -360,9 +360,9 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
 // creation of any new channels.
 // This ensures that no new frames will be produced, but there may be any number
 // of pending frames produced before this call which should still be published.
-func (s *channelManager) Close() error {
+func (s *channelManager) Close() {
 	if s.closed {
-		return nil
+		return
 	}
 
 	s.closed = true
@@ -373,9 +373,9 @@ func (s *channelManager) Close() error {
 	}
 
 	if s.pendingChannel == nil {
-		return nil
+		return
 	}
 
 	s.pendingChannel.Close()
-	return nil
+	return
 }
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -377,5 +377,4 @@ func (s *channelManager) Close() {
 	}
 
 	s.pendingChannel.Close()
-	return
 }
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -81,12 +81,9 @@ func (s *channelManager) TxFailed(id txID) {
 	}
 
 	s.metr.RecordBatchTxFailed()
-	// If this channel has no submitted transactions, put the pending blocks back into the
-	// local saved blocks and reset this state so it can try to build a new channel.
-	if len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
-		s.log.Info("Channel has no submitted transactions", "chID", s.pendingChannel.ID())
-		s.blocks = append(s.pendingChannel.Blocks(), s.blocks...)
-		s.clearPendingChannel()
+	if s.closed && len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
+		s.log.Info("Channel has no submitted transactions, clearing for shutdown", "chID", s.pendingChannel.ID())
+		s.Clear()
 	}
 }
 
@@ -194,11 +191,6 @@ func (s *channelManager) TxData(l1Head eth.BlockID) (txData, error) {
 		return s.nextTxData()
 	}
 
-	// Avoid producing new frames if the channel has been explicitly closed.
-	if s.closed {
-		return txData{}, io.EOF
-	}
-
 	// No pending frame, so we have to add new blocks to the channel
 
 	// If we have no saved blocks, we will not be able to create valid frames
@@ -210,8 +202,11 @@ func (s *channelManager) TxData(l1Head eth.BlockID) (txData, error) {
 		return txData{}, err
 	}
 
-	if err := s.processBlocks(); err != nil {
-		return txData{}, err
+	// Avoid processing blocks if the channel manager has been explicitly closed.
+	if !s.closed {
+		if err := s.processBlocks(); err != nil {
+			return txData{}, err
+		}
 	}
 
 	// Register current L1 head only after all pending blocks have been
@@ -365,9 +360,21 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
 // This ensures that no new frames will be produced, but there may be any number
 // of pending frames produced before this call which should still be published.
 func (s *channelManager) Close() error {
+	if s.closed {
+		return nil
+	}
+
 	s.closed = true
+
+	// Any pending state can be proactively cleared if there are no submitted transactions
+	if len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
+		s.Clear()
+	}
+
 	if s.pendingChannel == nil {
 		return nil
 	}
-	return s.pendingChannel.Close()
+
+	s.pendingChannel.Close()
+	return nil
 }
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -344,3 +344,13 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
 		SequenceNumber: l1info.SequenceNumber,
 	}
 }
+
+// CloseCurrentChannel closes the current pending channel, if one exists.
+// This ensures that no new frames will be produced, but there still may be any
+// number of pending frames produced before this call.
+func (s *channelManager) CloseCurrentChannel() error {
+	if s.pendingChannel == nil {
+		return nil
+	}
+	return s.pendingChannel.Close()
+}
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -358,8 +358,9 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
 
 // Close closes the current pending channel, if one exists, and prevents the
 // creation of any new channels.
-// This ensures that no new frames will be produced, but there may be any number
-// of pending frames produced before this call which should still be published.
+// This ensures that no new blocks will be added to the channel, but there may be any number
+// of frames still produced by calling `OutputFrames()`, which flushes the compression buffer.
+// These frames still need to be published.
 func (s *channelManager) Close() {
 	if s.closed {
 		return
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -63,6 +63,7 @@ func (s *channelManager) Clear() {
 	s.log.Trace("clearing channel manager state")
 	s.blocks = s.blocks[:0]
 	s.tip = common.Hash{}
+	s.closed = false
 	s.clearPendingChannel()
 }
 
@@ -83,7 +84,7 @@ func (s *channelManager) TxFailed(id txID) {
 	s.metr.RecordBatchTxFailed()
 	if s.closed && len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
 		s.log.Info("Channel has no submitted transactions, clearing for shutdown", "chID", s.pendingChannel.ID())
-		s.Clear()
+		s.clearPendingChannel()
 	}
 }
 
@@ -368,7 +369,7 @@ func (s *channelManager) Close() error {
 
 	// Any pending state can be proactively cleared if there are no submitted transactions
 	if len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
-		s.Clear()
+		s.clearPendingChannel()
 	}
 
 	if s.pendingChannel == nil {
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -41,6 +41,9 @@ type channelManager struct {
 	pendingTransactions map[txID]txData
 	// Set of confirmed txID -> inclusion block. For determining if the channel is timed out
 	confirmedTransactions map[txID]eth.BlockID
+
+	// if set to true, prevents production of any new channel frames
+	closed bool
 }
 
 func NewChannelManager(log log.Logger, metr metrics.Metricer, cfg ChannelConfig) *channelManager {
@@ -78,6 +81,13 @@ func (s *channelManager) TxFailed(id txID) {
 	}
 
 	s.metr.RecordBatchTxFailed()
+	// If this channel has no submitted transactions, put the pending blocks back into the
+	// local saved blocks and reset this state so it can try to build a new channel.
+	if len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
+		s.log.Info("Channel has no submitted transactions", "chID", s.pendingChannel.ID())
+		s.blocks = append(s.pendingChannel.Blocks(), s.blocks...)
+		s.clearPendingChannel()
+	}
 }
 
 // TxConfirmed marks a transaction as confirmed on L1. Unfortunately even if all frames in
@@ -182,6 +192,11 @@ func (s *channelManager) TxData(l1Head eth.BlockID) (txData, error) {
 	// Short circuit if there is a pending frame.
 	if dataPending {
 		return s.nextTxData()
+	}
+
+	// Avoid producing new frames if the channel has been explicitly closed.
+	if s.closed {
+		return txData{}, io.EOF
 	}
 
 	// No pending frame, so we have to add new blocks to the channel
@@ -345,10 +360,12 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
 	}
 }
 
-// CloseCurrentChannel closes the current pending channel, if one exists.
-// This ensures that no new frames will be produced, but there still may be any
-// number of pending frames produced before this call.
-func (s *channelManager) CloseCurrentChannel() error {
+// Close closes the current pending channel, if one exists, and prevents the
+// creation of any new channels.
+// This ensures that no new frames will be produced, but there may be any number
+// of pending frames produced before this call which should still be published.
+func (s *channelManager) Close() error {
+	s.closed = true
 	if s.pendingChannel == nil {
 		return nil
 	}
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -187,8 +187,8 @@ func (s *channelManager) TxData(l1Head eth.BlockID) (txData, error) {
 	dataPending := s.pendingChannel != nil && s.pendingChannel.HasFrame()
 	s.log.Debug("Requested tx data", "l1Head", l1Head, "data_pending", dataPending, "blocks_pending", len(s.blocks))
 
-	// Short circuit if there is a pending frame.
-	if dataPending {
+	// Short circuit if there is a pending frame or the channel manager is closed.
+	if dataPending || s.closed {
 		return s.nextTxData()
 	}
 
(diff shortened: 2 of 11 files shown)


op-batcher/batcher/driver.go
```diff
@@ -289,18 +289,17 @@ func (l *BatchSubmitter) loop() {
...
-				// hack to exit this loop. Proper fix is to do request another send tx or parallel tx sending
-				// from the channel manager rather than sending the channel in a loop. This stalls b/c if the
-				// context is cancelled while sending, it will never fully clear the pending txns.
+				// Attempt to gracefully terminate the current channel, ensuring that no new frames will be
+				// produced. Any remaining frames must still be published to the L1 to prevent stalling.
 				select {
-				case <-l.ctx.Done():
-					break blockLoop
+				case <-l.done:
+					l.state.CloseCurrentChannel()
```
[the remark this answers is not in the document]
trianglesphere: right, I see. I think though we don't hit the second `case <-l.done` because I believe a channel close only notifies once.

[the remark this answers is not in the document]
trianglesphere: better logging version: https://go.dev/play/p/PZA9qclOcLD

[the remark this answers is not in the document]
BrianBland: Once a channel is closed, it should become nonblocking, always returning the zero value for its defined type. I think we'll need both `case <-l.done` calls in order to both close the current channel in the inner loop as well as finally exit from the outer loop. This also should allow us to exit from the outer loop without even starting another inner loop if we happen to trigger a shutdown while we're outside of the inner loop, for example while processing L2 block state.

I think this should illustrate the desired behavior: https://go.dev/play/p/FviWCDBJBtL

Now that I'm writing out this explanation, I'm starting to feel like I should extract the inner loop into its own function for clarity 😅

BrianBland:
> Thanks for implementing this. I'd love to see some tests around this. Unit tests would be ideal, but we also have some frameworks which can handle more complex system setups.

Agreed about the need for testing. I'll see what I can put together here.

op-batcher/batcher/driver.go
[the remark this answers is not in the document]
trianglesphere: I see now. A new function for the inner loop is totally fine.

BrianBland:
I believe that this _should_ be technically tested by some of the e2e tests, such as `TestMigration`: https://github.com/ethereum-optimism/optimism/blob/68c357c5921abb0c519d6a404a6864d53d52f098/op-e2e/migration_test.go#L121

This is proving to be fairly difficult to unit test due to the dependency on the `L1Client` and `L2Client` `*ethclient.Client` fields. Is there a preferred framework we can use to simulate these blockchain dependencies?

op-batcher/batcher/channel_manager.go
```diff
@@ -174,6 +185,11 @@ func (s *channelManager) TxData(l1Head eth.BlockID) (txData, error) {
 		return s.nextTxData()
 	}
 
+	// Avoid producing new frames if the channel has been explicitly closed.
```
trianglesphere, line 188:
```suggestion
	// Avoid producing new frames if the channel manager has been explicitly closed.
```

op-batcher/batcher/channel_builder.go
```diff
@@ -337,6 +338,15 @@ func (c *channelBuilder) outputFrame() error {
 	return err // possibly io.EOF (last frame)
 }
 
+// Close immediately marks the channel as full with an ErrTerminated
+// if the channel is not already full, then outputs any remaining frames.
+func (c *channelBuilder) Close() error {
+	if !c.IsFull() {
+		c.setFullErr(ErrTerminated)
+	}
+	return c.closeAndOutputAllFrames()
```
trianglesphere, line 347: I think you might want to move `c.closeAndOutputAllFrames()` to inside the not full if statement. It will return an error here b/c the channel shouldn't be closed twice.

op-batcher/batcher/channel_manager.go
```diff
@@ -71,6 +74,14 @@ func (s *channelManager) TxFailed(id txID) {
 	} else {
 		s.log.Warn("unknown transaction marked as failed", "id", id)
 	}
+
+	// If this channel has no submitted transactions, put the pending blocks back into the
+	// local saved blocks and reset this state so it can try to build a new channel.
```
trianglesphere, line 79: What's the purpose of removing the pending channel instead of adding the frames back to it? I think this would cause batches to be slow if the transaction failed

op-batcher/batcher/driver.go
```diff
@@ -264,49 +262,54 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
+			l.publishStateToL1(ctx)
+		case <-l.done:
+			return
+		}
+	}
+}
+
+// publishStateToL1 loops through the block data loaded into `state` and
+// submits the associated data to the L1 in the form of channel frames.
+func (l *BatchSubmitter) publishStateToL1(ctx context.Context) {
```
trianglesphere, line 283: nice cleanup here.

op-batcher/batcher/channel_manager.go
```diff
@@ -71,6 +74,14 @@ func (s *channelManager) TxFailed(id txID) {
...
+	if len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
+		s.log.Info("Channel has no submitted transactions", "chID", s.pendingChannel.ID())
+		s.blocks = append(s.pendingChannel.Blocks(), s.blocks...)
+		s.clearPendingChannel()
+	}
```
sebastianst, line 84: We could even explicitly speed up shutdown in this case like
```suggestion
	if s.closed && len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
		s.log.Info("Channel has no submitted transactions, clearing for shutdown", "chID", s.pendingChannel.ID())
		s.Clear()
	}
```
wdyt @trianglesphere ?

op-batcher/batcher/channel_manager_test.go
```diff
@@ -106,24 +108,193 @@ func TestChannelManager_TxResend(t *testing.T) {
...
+		Difficulty: common.Big0,
+		Number:     big.NewInt(100),
+	}, nil, nil, nil, trie.NewStackTrie(nil))
+	l1InfoTx, err := derive.L1InfoDeposit(0, lBlock, eth.SystemConfig{}, false)
+	require.NoError(t, err)
+	txs := []*types.Transaction{types.NewTx(l1InfoTx)}
+
+	a := types.NewBlock(&types.Header{
+		Number: big.NewInt(0),
+	}, txs, nil, nil, trie.NewStackTrie(nil))
```
sebastianst, line 185: We'll soon merge https://github.com/ethereum-optimism/optimism/pull/5167 which contains a function `newMiniL2Block` removing this duplicated code.

op-batcher/batcher/driver.go
```diff
@@ -264,49 +262,54 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
-				case <-l.ctx.Done():
-					break blockLoop
-				default:
-				}
-			}
+		// Collect next transaction data
+		txdata, err := l.state.TxData(l1tip.ID())
+		if err == io.EOF {
+			l.log.Trace("no transaction data available")
+			break // local for loop
```
sebastianst, line 295:
```suggestion
			break
```

op-batcher/batcher/driver.go
```diff
@@ -264,49 +262,54 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
+		}
 
+		// Attempt to gracefully terminate the current channel, ensuring that no new frames will be
+		// produced. Any remaining frames must still be published to the L1 to prevent stalling.
+		select {
 		case <-l.done:
-			return
+			l.state.Close()
+		default:
 		}
```
sebastianst, line 313: I suggest we move this to the top of the function. Better react quicker to a shutdown request.

op-batcher/batcher/driver.go
```diff
@@ -264,49 +262,54 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
 	ticker := time.NewTicker(l.PollInterval)
 	defer ticker.Stop()
 	for {
 		select {
 		case <-ticker.C:
 			l.loadBlocksIntoState(l.ctx)
+			l.publishStateToL1(ctx)
+		case <-l.done:
+			return
+		}
```
sebastianst, line 277: Because of random `select` channel selection, we need to give priority to `l.done`. Also, don't we still want to publish the state once more after closing `l.done`? Otherwise it might happen that a stop request comes in and we just return here instead of publishing remaining frames.
```suggestion
		select {
		case <-ticker.C:
			select {
		        case <-l.done:
				l.publishStateToL1(ctx)
				return
			default:
			}
			
			l.loadBlocksIntoState(l.ctx)
			l.publishStateToL1(ctx)
		case <-l.done:
			l.publishStateToL1(ctx)
			return
		}
```

op-batcher/batcher/driver.go
```diff
@@ -264,49 +262,54 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
 func (l *BatchSubmitter) loop() {
 	defer l.wg.Done()
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
sebastianst, line 266: I think we can just replace this with
```suggestion
	ctx := context.TODO()
```
(or `context.Background()`). `TODO` because we might want to add a second force-shutdown context to the batch submitter's interal state. The current `l.ctx` now acts like a orderly-shutdown context because it now only cancels block loading.

op-batcher/batcher/driver.go
```diff
@@ -264,49 +262,54 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
+
+// publishStateToL1 loops through the block data loaded into `state` and
+// submits the associated data to the L1 in the form of channel frames.
+func (l *BatchSubmitter) publishStateToL1(ctx context.Context) {
+	for {
+		l1tip, err := l.l1Tip(ctx)
+		if err != nil {
+			l.log.Error("Failed to query L1 tip", "error", err)
+			return
+		}
```
sebastianst, line 289:
```suggestion
		l1ctx, cancel := context.WithTimeout(ctx, networkTimeout)
		l1tip, err := l.l1Tip(ctx)
		cancel()
		if err != nil {
			l.log.Error("Failed to query L1 tip", "error", err)
			return
		}
```
Or a broken l1 client connection might block very long here.

I think @trianglesphere is implementing more elaborate internal timeout handling in the transaction manager below, so should be fine below to just pass the `ctx` there.

sebastianst (changes requested):
Thanks for adding this! 

We should think about a way to not only test the `channelManager` but also the `BatchSubmitter` itself, because there's a lot going on now with different contexts, closing `done` channels etc.

op-batcher/batcher/driver.go
sebastianst: I now realize that this is probably not possible currently, because once the manager is closed, it doesn't return tx data.

op-batcher/batcher/driver.go
trianglesphere: Should we instead move the close check from the bottom of the `publishStateToL1` function to the top? I think that'd take care of closing it at the right time. With that being true (& synchronous tx submission), I don't think we'd actually get a case where we would close a pending channel.

op-batcher/batcher/channel_manager.go
BrianBland: This was added to avoid the scenario in which a repeatedly-failing transaction results in an infinite loop during shutdown.

For example, if the batcher does not have enough ETH to cover L1 batch submission, then each call to `TxFailed` will clear the `pendingTransactions` list but keep the failing frame. On the next loop, we will call `TxData` which will detect and return this re-enqueued frame, which could repeat indefinitely.

By clearing the pending channel, this can detect and recover from the scenario where a channel has no submitted frames, and can treat this the same as if no channel was ever constructed in the first place.

It might make the most sense to go with @sebastianst's suggestion below and only applying this behavior when the channel manager is already closed: https://github.com/ethereum-optimism/optimism/pull/5105/files/ba5b58c79604c0821ef28553e6ed01263ae3d544#r1142082400

op-batcher/batcher/channel_builder.go
BrianBland: The goal was to make a call to `ChannelBuilder.Close` emit any remaining frames and prevent any future frames from being outputted.

Another option would be to have `ChannelBuilder.Close` only call `setFullErr` (if unset) and leave the responsibility of outputting remaining frames to the standard `OutputFrames` method. If we take this approach, then we will probably want to cease calling `ChannelManager.processBlocks()` once the manager has been closed in order to ensure a timely shutdown. This alternative approach would probably help us to avoid double-closing the channel builder, making this potentially a safer route.

op-batcher/batcher/channel_manager_test.go
```diff
@@ -106,24 +108,193 @@ func TestChannelManager_TxResend(t *testing.T) {
...
-	txdata0bytes := txdata0.Bytes()
-	data0 := make([]byte, len(txdata0bytes))
+
+	// confirm one frame to keep the channel open
+	m.TxConfirmed(txdata0.ID(), eth.BlockID{})
+
+	txdata1, err := m.TxData(eth.BlockID{})
+	require.NoError(err)
+	txdata1bytes := txdata1.Bytes()
+	data1 := make([]byte, len(txdata1bytes))
```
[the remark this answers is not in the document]
BrianBland: This change was necessary because I altered the behavior of `TxFailed` to reset the channel entirely if the channel had no pending or confirmed transactions. A channel reset would result in a new channel ID, and thus different transaction data. I can probably revert these test changes if we only clear the channel in `TxFailed` if the `ChannelManager` is already closed.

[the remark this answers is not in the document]
sebastianst: I think that makes sense, only clearing if closed.

op-batcher/batcher/driver.go
BrianBland: Interesting point about the random `select` ordering. There's a few assumptions in the current implementation:
1. `l.ctx` is cancelled when `l.done` is readable (closed). This is because `Stop` calls `l.cancel()` before `close(l.done)`
2. A cancelled `l.ctx` context will effectively prevent `l.loadBlocksIntoState(l.ctx)` from having any meaningful side effects.
3. `l.publishStateToL1(ctx)` will also have no effect without any loaded blocks.

This still could result in at least one unnecessary cycle, potentially multiple if these methods take longer than expected or the `ticker` speed is sufficiently fast. I can reorder this code to prioritize the `done` channel :+1:

op-batcher/batcher/driver.go
BrianBland: It should be fine to move this to the top of the function, as closing an open/pending channel will still cause all of the pending frames to be emitted.

op-batcher/batcher/channel_manager.go
BrianBland: This actually identifies an interesting ordering issue:
If we mark the channel manager's only transaction as failed and then close it, we'll still have this pending frame. As a result, we'll still attempt to submit this one more time. The previous implementation would not be subject to this ordering problem, but would be more prone to unnecessary channel clearing.

BrianBland: I realized that we can also handle this other case inside of `ChannelManager.Close()` in order to avoid unnecessary channel erasure as well as extra retries.

op-batcher/batcher/driver.go
BrianBland: This now supports passing in a context to the `Stop()` method, which can be used to short-circuit any long blocking client calls, courtesy of @mdehoog

op-batcher/batcher/driver.go
```diff
@@ -292,46 +302,59 @@ func (l *BatchSubmitter) loop() {
...
-				// from the channel manager rather than sending the channel in a loop. This stalls b/c if the
-				// context is cancelled while sending, it will never fully clear the pending txns.
-				select {
-				case <-l.ctx.Done():
-					break blockLoop
-				default:
-				}
+			// prioritize the `done` condition over the ticker, even though select ordering is randomized
+			select {
+			case <-l.done:
```
trianglesphere, line 307: @sebastianst  I don't think we need this. Even if `l.done` would be hit here, we would hit it when we enter `l.publishStateToL1`. We should also cancel the `l.loadCtx` early on in the shutdown process.

BrianBland: Given that `l.cancelLoad()` is called immediately prior to `close(l.done)`, we can be sure that `l.loadBlocksIntoState(l.loadCtx)` will run with a cancelled context whenever `l.done` is readable. We can probably safely revert this code back to the previous simpler state without the explicit `done` prioritization

op-batcher/batcher/channel_manager.go
```diff
@@ -344,3 +355,26 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
 		SequenceNumber: l1info.SequenceNumber,
 	}
 }
+
+// Close closes the current pending channel, if one exists, and prevents the
+// creation of any new channels.
+// This ensures that no new frames will be produced, but there may be any number
+// of pending frames produced before this call which should still be published.
```
sebastianst, line 362:
```suggestion
// This ensures that no new blocks will be added to the channel, but there may be any number
// of frames still produced by calling `OutputFrames()`, which flushes the compression buffer.
// These frames still need to be published.
```

Technically, we might still produce frames from the batches in the compression pipeline. But we don't add any more blocks to the channel builder.

op-batcher/batcher/channel_manager.go
```diff
@@ -195,8 +203,11 @@ func (s *channelManager) TxData(l1Head eth.BlockID) (txData, error) {
 		return txData{}, err
 	}
 
-	if err := s.processBlocks(); err != nil {
-		return txData{}, err
+	// Avoid processing blocks if the channel manager has been explicitly closed.
+	if !s.closed {
+		if err := s.processBlocks(); err != nil {
```
sebastianst, line 208: Note: I think this makes sense, but `s.processBlocks()` would return anyways without adding any blocks because the channel is marked as full.

op-batcher/batcher/channel_manager.go
```diff
@@ -60,6 +63,7 @@ func (s *channelManager) Clear() {
 	s.log.Trace("clearing channel manager state")
 	s.blocks = s.blocks[:0]
 	s.tip = common.Hash{}
+	s.closed = false
```
sebastianst, line 66: If I see correctly, it currently wouldn't happen that we clear the channel manager after closing it, because we don't load L2 blocks any more, which is where the `Clear()` might be called. 

But if we ever get into this situation, I think we should actually keep the `channelManager` closed forever once it got closed. The batcher doesn't revert from shutting down.

```suggestion
// We deliberately let `s.closed` stay at its current value, because once a channelManger got closed, it should stay closed.

File: op-batcher/batcher/driver.go [hunk 18]
```diff
@@ -292,46 +302,59 @@ func (l *BatchSubmitter) loop() {
...
+	for {
+		// Attempt to gracefully terminate the current channel, ensuring that no new frames will be
+		// produced. Any remaining frames must still be published to the L1 to prevent stalling.
+		select {
+		case <-ctx.Done():
+			l.state.Close()
 		case <-l.done:
+			l.state.Close()
+		default:
+		}
```
sebastianst, line 334: See other comment, we can probably just remove the `done` channel and do
```suggestion
		select {
		case <-ctx.Done():
			l.state.Close()
		case <-l.shutdownCtx.Done():
			l.state.Close()
		default:
		}
```

sebastianst (changes requested):

op-batcher/batcher/channel_manager.go
BrianBland: This is actually currently necessary in order to handle the case where the `pendingChannel` is `nil`. More likely we should avoid creating a new `pendingChannel` (via `ensurePendingChannel`) whenever this channel manager is already closed.

op-batcher/batcher/channel_manager.go
BrianBland: We're intending to stop and potentially restart the batcher via the [Admin API](https://github.com/ethereum-optimism/optimism/blob/develop/op-batcher/rpc/api.go), and resetting this `closed` flag would be necessary to support a restart.

sebastianst: I see, didn't think about restarting a stopped batcher in the same program instance yet. I assumed that when it is stopped, the whole application would be restarted.

I think quite a few more changes would need to be made to support restarting the batcher using the same `BatchSubmitter` instance.  E.g. the lifetime context `shutdownCtx` gets closed etc. Since the batcher is supposed to be stateless anyways, it is probably easier to just always restart the whole application. Or at least create a new `BatchSubmitter` instance instead of reusing it.

But anyways, I'm fine with also resetting `c.closed` to false here, since it doesn't have any affect currently (as far as I can tell). We can rack our brains about how to best restart the batch submitter via rpc when the time comes.

op-batcher/batcher/channel_manager.go
sebastianst: You're right, I agree that it probably makes sense to not even call `ensurePendingChannel` if `s.closed`. To avoid moving more and more stuff into the `if !s.closed {...}` block, we could extend above early exit to
```
	// Short circuit if there is a pending frame or we're shutting down
	if dataPending || s.closed {
		return s.nextTxData()
	}
```

For that to work, we would need to call `s.outputFrames()` at the end of `channelManager.Close()`, to call it one last time when shutting down. (Did we have that in an earlier iteration? 😅 )

op-batcher/batcher/channel_manager.go
```diff
@@ -344,3 +355,27 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
...
+	// Any pending state can be proactively cleared if there are no submitted transactions
+	if len(s.confirmedTransactions) == 0 && len(s.pendingTransactions) == 0 {
+		s.clearPendingChannel()
+	}
+
+	if s.pendingChannel == nil {
+		return
+	}
+
+	s.pendingChannel.Close()
```
sebastianst, line 380: see previous comment, we might consider adding 
```
return s.outputFrames()
```
here (would then need to add error return to `Close`) and then early-exit in `TxData` when closed.

op-e2e/setup.go
```diff
@@ -214,12 +214,12 @@ type System struct {
 	Mocknet           mocknet.Mocknet
 }
 
-func (sys *System) Close() {
+func (sys *System) Close(ctx context.Context) {
 	if sys.L2OutputSubmitter != nil {
 		sys.L2OutputSubmitter.Stop()
 	}
 	if sys.BatchSubmitter != nil {
```
sebastianst, line 221:
```suggestion
func (sys *System) Close() {
	if sys.L2OutputSubmitter != nil {
		sys.L2OutputSubmitter.Stop()
	}
	
	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()
	if sys.BatchSubmitter != nil {
```

Instead of adding a `ctx` parameter to `Close`, we could also consider to just always close the batcher with some default context (like you do in the `TestMissingBatchE2E` test). You pass `context.Background()` to all other tests anyways. 

But I'm not that familiar with the e2e test, would like to hear @trianglesphere opinion here.

op-e2e/migration_test.go
```diff
@@ -352,7 +352,7 @@ func TestMigration(t *testing.T) {
 	}, lgr.New("module", "batcher"), batchermetrics.NoopMetrics)
 	require.NoError(t, err)
 	t.Cleanup(func() {
-		batcher.StopIfRunning()
+		batcher.StopIfRunning(context.Background())
```
sebastianst, line 355: maybe timeout context?

sebastianst (changes requested):
Looks great now 🌞

See comments, I agree with your remark that `ensurePendingChannel` shouldn't even be called when the channel manager is closed.

op-batcher/batcher/driver.go
```diff
@@ -159,11 +158,11 @@ func (l *BatchSubmitter) Start() error {
...
 }
 
-func (l *BatchSubmitter) StopIfRunning() {
-	_ = l.Stop()
+func (l *BatchSubmitter) StopIfRunning(ctx context.Context) {
+	_ = l.Stop(ctx)
 }
 
-func (l *BatchSubmitter) Stop() error {
+func (l *BatchSubmitter) Stop(ctx context.Context) error {
```
mdehoog, line 165: The `ctx` parameter in is no longer used anywhere. The original purpose of passing this in was so that the passed in context could force cancel this call if needed, and had a small go routine that would cancel the "kill" context if the parameter was canceled: https://github.com/BrianBland/optimism/pull/1/commits/6cce718f7afdf5423bda6f2cc515c6d08d75d505#diff-c734d1296b2fd691221b92df3edf09c7533c507a74c2316117745c75c3ad5776R168-R175

did we decide that was a bad idea?

op-batcher/batcher/channel_manager.go
mdehoog: @sebastianst Base has a stop/start batcher flow in our runbook, which we use every now-and-then. The `shutdownCtx` and `killCtx` gets recreated on `start` so it should work fine. I agree that it's relatively easy to restart the entire process, but for us it's been easier to call an RPC to reset the state, because our environment is pretty locked down for security reasons.

sebastianst: Sounds good. We could also consider changing the rpc `Start` implementation to throw away a shut-down `BatchSubmitter` instance (i.e., if stopped before) and re-create a new one. This way, the restart would still happen inside the same process but we wouldn't have the headache of making sure that a `BatchSubmitter` instance can safely be reused after shutdown. It would probably come with its own problems though.

op-batcher/batcher/driver.go
sebastianst: If we want to add an (optional) timeout parameter to the stop rpc, it could still make sense to have this context. Then the rpc could use the provided timeout parameter to create a `context.WithTimeout`.

BrianBland: Let me add that back - I might have accidentally lost that to rebases/merge conflicts 😅

op-e2e/setup.go
BrianBland: This definitely sounds reasonable enough to me for the purpose of e2e tests, but happy to change course if there's a good reason to not take this approach.

op-batcher/batcher/channel_manager.go
BrianBland: Haha I think that's where we started: https://github.com/ethereum-optimism/optimism/pull/5105/commits/a9318e6fc89a221a02b1d0760133cd545e108242#r1142729550

We chose to take this approach in order to avoid double-closing the channel builder, but I'll see if that can be avoided.

op-batcher/batcher/driver.go
```diff
@@ -174,9 +173,18 @@ func (l *BatchSubmitter) Stop() error {
...
+	wrapped, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go func() {
+		<-wrapped.Done()
+		cancelKill()
+	}()
+
+	l.cancelShutdownCtx()
 	l.wg.Wait()
+	l.cancelKillCtx()
```
sebastianst, line 187:
```suggestion
	l.cancelShutdownCtx()
	defer l.cancelKillCtx()
	l.wg.Wait()

```
I think we fixed that in an earlier iteration already.

sebastianst (approved):
Looks good now!

It seems one fix got reverted, see comment.

op-batcher/batcher/driver.go
BrianBland: This was intentionally re-introduced in order to allow us to use the passed-in context as a potential "kill" signal on shutdown: https://github.com/ethereum-optimism/optimism/pull/5105#discussion_r1149618650

sebastianst: I see, sorry. Can't we then just do?:
```suggestion
	// go routine will call cancelKill() if the passed in ctx is ever Done
	go func() {
		<-ctx.Done()
		l.cancelKillCtx()
	}()

	l.cancelShutdownCtx()
	l.wg.Wait()
	l.cancelKillCtx()
```
What's the purpose of `wrapped`?

BrianBland: Without `wrapped` and, more-importantly, the `cancel` signal, it could be possible to have the spawned goroutine run forever if the passed in `ctx` is never "done", like for example in the case where `ctx` is `context.Background()`

BrianBland:
@mergifyio requeue

op-batcher/batcher/driver.go
sebastianst: Persisting a comment from private DM here: could also have done the following to avoid creating another context:
```suggestion
	go func() {
		select: {
		case <-ctx.Done():
			l.cancelKillCtx()
		// make sure goroutine returns eventually
		case <-l.killCtx:
			return
		}
	}()

	l.cancelShutdownCtx()
	l.wg.Wait()
	l.cancelKillCtx()
```

Issue #5048, opened 2023-03-03
Migration have error expected storage slots


hungzinza:
**Describe the bug**
My test-node was stopped at height 4,061,223
Network: goerli
I run bedrock-init to upgrade it.
I enable flag: `no-check` when run command 
  ```
./bin/op-migrate \
    --l1-rpc-url $OP_NODE__RPC_ENDPOINT \
    --ovm-addresses $WITNESS_OUT_PATH/data/ovm-addresses.json \
    --evm-addresses $WITNESS_OUT_PATH/data/evm-addresses.json \
    --ovm-messages $WITNESS_OUT_PATH/data/ovm-messages.json \
    --evm-messages $WITNESS_OUT_PATH/data/evm-messages.json \
    --ovm-allowances $WITNESS_OUT_PATH/data/ovm-allowances.json \
    --network $ETH_NETWORK \
    --no-check \
    --deploy-config ../packages/contracts-bedrock/deploy-config/$ETH_NETWORK.json \
    --rollup-config-out ../packages/contracts-bedrock/deploy-config/$ETH_NETWORK.json \
    --hardhat-deployments ../packages/contracts-bedrock/deployments,../packages/contracts/deployments,../packages/contracts-periphery/deployments \
    --db-path $LEGACY_GETH_COPY_DIR
```
I have an error:
`expected 6 storage slots for L2CrossDomainMessenger but got 4`

![Screenshot from 2023-03-03 11-53-07](https://user-images.githubusercontent.com/96384818/222634626-88cd2ee3-287a-42d2-8883-f2e8791f725e.png)

**System Specs:**
I use latest version of this repo: https://github.com/smartcontracts/simple-optimism-node

**Additional context**

NatLukcrypto:
interesting

hungzinza:
Hi @NatLukcrypto, Could you help me check it? I'm stuck at this error for a long time

tynes:
This is a known bug based on changes to the cross domain messenger, fixes are incoming

hungzinza:
@tyne
I'm waiting for it. Thank you

PR #5106, opened 2023-03-10
op-chain-ops: Improve OVM_ETH migration, remove invalid check slots

mslipper:
When we ran the migration rehearsal, the OVM_ETH migration took approximately an hour and a half. This was because the storage slots in the OVM_ETH contract were iterated twice: once to check them, and another time to actually migrate them. To reduce the amount of time taken, this PR updates the OVM_ETH migration to check and migrate in one step. This does not cause database corruption on failure, since the trie changes are buffered in memory and aren't committed until all the checks pass. Total migration time should be reduced by approximately 40 minutes.

Separately, I removed the invalid check slots referring to the EIP-1967 slots on the L2XDM. These slots no longer exist, and were causing the post-check steps to fail.

Closes CLI-3590
Closes CLI-3576
Closes https://github.com/ethereum-optimism/optimism/issues/5048


mslipper:
Marking as draft while I benchmark.

op-chain-ops/genesis/check.go
```diff
@@ -59,9 +59,6 @@ var (
 			common.Hash{31: 0x97}: common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000001"),
 			// Slot 0xcc (204) is xDomainMsgSender
 			common.Hash{31: 0xcc}: common.HexToHash("0x000000000000000000000000000000000000000000000000000000000000dead"),
-			// EIP-1967 storage slots
-			AdminSlot:          common.HexToHash("0x0000000000000000000000004200000000000000000000000000000000000018"),
```
tynes, line 63: The messenger is still behind a proxy - its 51 and 151 that no longer exist

See:
- https://github.com/ethereum-optimism/optimism/blob/4071f64e5dd582beaa28c393b20757193361a4ca/packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol#L77
- https://github.com/ethereum-optimism/optimism/blob/4071f64e5dd582beaa28c393b20757193361a4ca/packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol#L45

Balochistan-has (commented):
https://github.com/ethereum-optimism/optimism/pull/5106/files#diff-c08feda74e0f1bce60c0ed2b207a0d29895c7a326d6172b2c8a7a26f5cc30114

op-chain-ops/genesis/db_migration.go
```diff
@@ -192,10 +182,15 @@ func MigrateDB(ldb ethdb.Database, config *DeployConfig, l1Block *types.Block, m
...
+	// We also need to verify that we have all of the storage slots for the LegacyERC20ETH contract
+	// that we expect to have. An error will be thrown if there are any missing storage slots.
+	// Unlike with withdrawals, we do not need to filter out extra addresses because their balances
+	// would necessarily be zero and therefore not affect the migration.
+	//
+	// Once verified, we migrate the balances held inside the LegacyERC20ETH contract into the state trie.
 	// We also delete the balances from the LegacyERC20ETH contract.
 	log.Info("Starting to migrate ERC20 ETH")
-	err = ether.MigrateLegacyETH(db, addrs, int(config.L1ChainID), noCheck)
+	err = ether.MigrateLegacyETH(db, migrationData.Addresses(), migrationData.OvmAllowances, int(config.L1ChainID), noCheck, commit)
```
tynes, line 193: I want to make sure there is no diff passing in `migrationData.Addresses()` compared to `addrs` which is returned from `PreCheckBalances`

https://github.com/ethereum-optimism/optimism/blob/35d4a37c2b9a59e9bf16fa4c5e22283f2348c3d2/op-chain-ops/ether/precheck.go#L18

tynes: It looks good based on the new logic

mslipper: The OVM ETH migration only migrates what's actually in state. It doesn't create any new state like the withdrawals migration does, so we don't need to filter the input like we do with withdrawals.

tynes (approved):

PR #5107, opened 2023-03-10
[op-node: derivation pipeline] Don't allow a channel to be reopened

mdehoog:
**Description**

Base had a safe-block halt for a number of hours last night. After downloading the L1 data and simulating the derivation pipeline, I found a number of duplicate `frame_number>0 IsLast=true` frames.

Currently the derivation pipeline will open a new channel with the same ID for channels that have completed (timed out or read) if a new frame is submitted for that channel. I believe according to the [spec](https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#loading-frames), this is a bug:
```
 - Duplicate closes (new frame is_last == 1, but the channel has already seen a closing frame) are dropped.
 - If a frame is closing (is_last == 1) any existing higher-numbered frames are removed from the channel.
```

**Tests**

I added a new regression test that reproduces the issue.

**Invariants**

This may change the chain derivation and may require a hard fork.

**Additional context**

Safe block derivation before the fix:
Screenshot 2023-03-09 at 10 16 37 PM

After:
Screenshot 2023-03-09 at 11 50 21 PM

(there was still a halt with the fix, this is due to a dangling unclosed channel from restarting the batcher as an attempt to fix the initial chain halt)

### DIFF
--- a/op-node/rollup/derive/channel_bank.go
+++ b/op-node/rollup/derive/channel_bank.go
@@ -58,6 +58,9 @@ func (p *ProcessedChannels) Seen(id ChannelID) bool {
 }
 
 func (p *ProcessedChannels) Add(id ChannelID) {
+	if p.Seen(id) {
+		return
+	}
 	p.m[id] = true
 	p.a = append(p.a, id)
 	if len(p.a) > MaxProcessedChannelsSize {
--- a/op-node/rollup/derive/channel_bank.go
+++ b/op-node/rollup/derive/channel_bank.go
@@ -32,24 +32,51 @@ type ChannelBank struct {
 	log log.Logger
 	cfg *rollup.Config
 
-	channels     map[ChannelID]*Channel // channels by ID
-	channelQueue []ChannelID            // channels in FIFO order
+	channels          map[ChannelID]*Channel // channels by ID
+	channelQueue      []ChannelID            // channels in FIFO order
+	processedChannels ProcessedChannels      // processed channels by ID
 
-	prev    NextFrameProvider
-	fetcher L1Fetcher
+	prev NextFrameProvider
+}
+
+// ProcessedChannels tracks channels that have been processed (either fully-read or timed-out).
+// This is to ensure if a new frame arrives for a completed channel, we don't create a new
+// channel for that frame.
+type ProcessedChannels struct {
+	m map[ChannelID]bool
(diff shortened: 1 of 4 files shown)


op-node/rollup/derive/channel_bank.go
```diff
@@ -32,24 +32,51 @@ type ChannelBank struct {
...
+}
+
+func (p *ProcessedChannels) Seen(id ChannelID) bool {
+	return p.m[id]
+}
+
+func (p *ProcessedChannels) Add(id ChannelID) {
+	p.m[id] = true
+	p.a = append(p.a, id)
+	if len(p.a) > MaxProcessedChannelsSize {
```
mdehoog, line 63: this could also be time-based if we want to make spec changes to allow re-use of a channelID within a certain time period

sebastianst:
Summary from internal discussion: this surfaced the issue of the implementation currently not fully following the spec in that frames with duplicate channel IDs are not properly discarded (but consistently so, so no risk of diverging nodes). However, this particular solution would lead to inconsistent behavior among op-nodes because it only adds a runtime cache. If a node (re)starts, it still doesn't know the channels IDs that have been processed before. So this could actually lead to diverging forks among them.

We're currently evaluating whether and how to change the spec in this regard. Until then we should put work on this PR on pause.

mdehoog:
Makes sense! Will close for now. Happy to open another PR once the spec solidifies.

ajsutton:
@mdehoog Just to provide an update here to be sure you see it. I came up with three potential options here (the three PRs referenced above). The two modifications to the derivation spec that would avoid these safe head stalls are consensus breaking changes so to be safe should really be done as a hard fork. Technically we could apply the new rules retroactively at least on op-goerli but there's always a risk that some channels will be added to L1 while some nodes are rolling out and cause a chain split (plus while I've analysed the existing batches from op-goerli, I haven't looked at other chains).

So the current plan is to update the spec to match the current implementation (#5182) to remove ambiguity about how things are actually working.  Then in the next hard fork we should be able to move to processing the first ready frame which avoids these stalls even if the batcher does leave a channel unclosed (#5196 but rewritten to be done at a hard fork boundary).  In the meantime there is also work being done on the batcher to avoid it leaving this unclosed channels when it restarts.

Let me know if you have any concerns.

PR #5112, opened 2023-03-11
contracts-bedrock: prevent too low of gas config

tynes:
**Description**

The chain liveliness depends on the L2 gas limit being larger than the gas consumed by a deposit. If a deposit consumes more gas than what is possible in an L2 block, then the chain will halt and be unable to produce blocks.

This was previously handled by having a hardcoded constant in the system config contract. This constant must be kept in sync with the value in the optimism portal. It is also possible to deploy a new optimism portal such that it has a different resource limit. To remove the need to always having to keep these values in sync, a handle to the portal is given to the system config contract. The system config contract will call the portal to get the value dynamically.

This ensures that we do not need to remember what the value is and update the contracts in tandem.


tynes:
If we opt into this solution, then I will need to modify the L1 deploy process such that the correct optimism portal contract exists at the correct location before the deployment of the system config contract. Will require a refactor of how the L1 deployment works so that contracts are deployed + initialized + moved to the account one at a time, in dependency order.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -117,7 +119,8 @@ contract SystemConfig is OwnableUpgradeable, Semver {
         uint64 _gasLimit,
         address _unsafeBlockSigner
     ) public initializer {
-        require(_gasLimit >= MINIMUM_GAS_LIMIT, "SystemConfig: gas limit too low");
+        uint256 maxResourceLimit = uint256(ResourceMetering(PORTAL).MAX_RESOURCE_LIMIT());
+        require(_gasLimit >= maxResourceLimit, "SystemConfig: gas limit too low");
```
trianglesphere, line 123: I'm a big fan of these changes. We may want to make sure to include the 1 Million gas used for the L1 Info Tx in this as well.

tynes: Good catch

packages/contracts-bedrock/deploy/017-SystemConfigImpl.ts
```diff
@@ -18,6 +22,7 @@ const deployFn: DeployFunction = async (hre) => {
       batcherHash,
       hre.deployConfig.l2GenesisBlockGasLimit,
       hre.deployConfig.p2pSequencerAddress,
+      Artifact__OptimismPortalProxy.address,
```
protolambda, line 25: Maybe it's worth adding a check to the l2-check script, to verify it's configured correctly.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +214,15 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+    /**
+     * @notice Returns the minimum gas limit allowed by the system. If the L2
+     *         gas limit is set to a value smaller than this, then it is
+     *         possible for a block to be produced that uses more gas than what
+     *         is allowed on L2, resulting in a liveliness failure.
+     */
+    function _minimumGasLimit() internal returns (uint256) {
+        uint256 maxResourceLimit = uint256(ResourceMetering(PORTAL).MAX_RESOURCE_LIMIT());
+        return maxResourceLimit + SYSTEM_TRANSACTION_MAX_GAS;
```
protolambda, line 226: Instead of reading from the constant from the portal contract, we can also define it here, and read the system config from the resource-metering (or really from the optimism portal) contract. This prevents a circular dependency later on if we end up making the system-config a dependency of the portal contract to read other system configurables.

protolambda (approved):
LGTM. Just a few comments about future contract dependency / checks. The PR description makes sense, but I'm not sure if future additional configurable values change this.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
tynes: Thats a good point, the call overhead wont matter because its in the gas burn path and the gas used in computing the eip1559 params counts towards the gas burning

tynes: After thinking about this a bit, i agree that its a good idea to centralize as many config params in a single place for governance purposes, then do we move the eip 1559 params there too? Due to needing to make a lot of decisions, i lean towards not making any drastic changes as part of this pr

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +217,15 @@ contract SystemConfig is OwnableUpgradeable, Semver {
             sstore(slot, _unsafeBlockSigner)
         }
     }
+
+    /**
+     * @notice Returns the minimum gas limit allowed by the system. If the L2
+     *         gas limit is set to a value smaller than this, then it is
+     *         possible for a block to be produced that uses more gas than what
+     *         is allowed on L2, resulting in a liveliness failure.
```
maurelian, line 225: nit: I think 'liveness' is more common terminology

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +217,15 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     */
+    function _minimumGasLimit() internal view returns (uint256) {
```
maurelian, line 227: Should we expose this as a public method? That would align with the vision of the SystemConfig as the central location for all system info.

tynes:
I've move the min gas check outside of the constructor and added a comment that it should happen offchain. I added a check to the deployment code. It would require a large refactor of how we do deployments/initializations to be able to allow a contract to make a call in the constructor.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
tynes: No super opinionated, can make it into an external function

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +217,15 @@ contract SystemConfig is OwnableUpgradeable, Semver {
             sstore(slot, _unsafeBlockSigner)
         }
     }
+
+    /**
+     * @notice Returns the minimum gas limit allowed by the system. If the L2
+     *         gas limit is set to a value smaller than this, then it is
+     *         possible for a block to be produced that uses more gas than what
+     *         is allowed on L2, resulting in a liveness failure.
```
maurelian, line 225: Is that what would happen, or would deposit blocks be rejected or invalid? I honestly don't know.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +217,15 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     */
+    function minimumGasLimit() public view returns (uint256) {
+        uint256 maxResourceLimit = uint256(ResourceMetering(PORTAL).MAX_RESOURCE_LIMIT());
```
maurelian, line 228: I think this needs a comment to explain what's going on, i.e. the relationship between the minimum gas limit here, and the maximum resource limit in the portal.

maurelian (commented):
Code looks solid, just looking for some small tweaks to comments then will approve.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -117,7 +128,6 @@ contract SystemConfig is OwnableUpgradeable, Semver {
         uint64 _gasLimit,
         address _unsafeBlockSigner
     ) public initializer {
-        require(_gasLimit >= MINIMUM_GAS_LIMIT, "SystemConfig: gas limit too low");
```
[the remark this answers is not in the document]
maurelian: Or maybe add a check against the deploy-config itself?

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -79,23 +83,30 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
         uint256 _overhead,
         uint256 _scalar,
         bytes32 _batcherHash,
         uint64 _gasLimit,
-        address _unsafeBlockSigner
-    ) Semver(1, 0, 1) {
+        address _unsafeBlockSigner,
+        address _portal
+    ) Semver(1, 1, 0) {
+        PORTAL = _portal;
```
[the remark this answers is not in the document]
tynes: I'm confused as to what you mean here?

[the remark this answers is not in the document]
maurelian: I should have said "comment only". 
I mean that the diff would be a lot easier to read if solidity supported comma dangles. 
Resolving.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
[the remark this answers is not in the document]
tynes: There is a diff to the deploy flow that does this but it still needs to hardcode numbers because the number exists in the contract as a constant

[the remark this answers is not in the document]
maurelian: Sorry I missed that, looking again, I think the diff in the deploy flow should be sufficient here. 
Will let you resolve ths.

op-chain-ops/genesis/layer_one.go
```diff
@@ -229,7 +230,14 @@ func BuildL1DeveloperGenesis(config *DeployConfig) (*core.Genesis, error) {
 		}
 
 		memDB.CreateAccount(depAddr)
-		memDB.SetCode(depAddr, dep.Bytecode)
+
+		bytecode := dep.Bytecode
+		// Ensure that the immutable values in the system config are set correctly
+		if dep.Name == "SystemConfig" {
```
tynes, line 236: This code can be removed now

packages/contracts-bedrock/deploy/017-SystemConfigImpl.ts
```diff
@@ -1,13 +1,36 @@
...
+
+  const [OptimismPortal] = await getContractsFromArtifacts(hre, [
+    {
+      name: 'OptimismPortal',
+      signerOrProvider: deployer,
+    },
+  ])
+
+  const MAX_RESOURCE_LIMIT = await OptimismPortal.MAX_RESOURCE_LIMIT()
+  const minGasLimit = MAX_RESOURCE_LIMIT.add(1_000_000)
```
tynes, line 29: i hate this magic number

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
tynes: According to @trianglesphere it would cause the chain to halt. I don't believe we have test coverage of this case

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +221,18 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     * @notice Returns the minimum gas limit allowed by the system. If the L2
+     *         gas limit is set to a value smaller than this, then it is
+     *         possible for a block to be produced that uses more gas than what
+     *         is allowed on L2, resulting in a liveness failure. The MAX_RESOURCE_LIMIT
+     *         represents the amount of gas that can be consumed by deposits and
+     *         the SYSTEM_TRANSACTION_MAX_GAS represents the maximum amount of
+     *         gas the system transaction can consume in a block.
+     */
+    function minimumGasLimit() public view returns (uint256) {
+        uint256 maxResourceLimit = uint256(ResourceMetering(PORTAL).MAX_RESOURCE_LIMIT());
```
smartcontracts, line 235: I find this relationship pretty ugly. If anything, `ResourceMetering` should be pulling the max resource limit from `SystemConfig` rather than the other way around. That way you don't need a minimum gas limit like this. You just derive the maximum available deposit gas from the maximum resource gas as defined in this contract.

tynes: Considered this design but decided against it balancing amount of diff

clabby: Also seems like a cleaner layout to me, but this feels like a change that should be done after we ship Bedrock.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
clabby: How difficult would it be to add a test for this case in `op-e2e`? Is this something we could do prior to mainnet launch, or okay to put in the backlog in your opinion?

PR #5114, opened 2023-03-11
feat(ctb): Optimize differential testing

clabby:
# Overview

Optimizes the differential testing in `contracts-bedrock` by switching to a Go implementation of `differential-testing.ts`.

## Figures
`contracts-bedrock`'s "test and generate coverage" task takes 2 mins, 51 seconds in this branch.

Top: Go
Bottom: TS
Duration of `Hashing.sol` diff tests decreased by `97.6%`
Screenshot from 2023-03-10 22-06-47

## TODO
**Diff test methods**
- [x] `decodeVersionedNonce`
- [x] `encodeCrossDomainMessage`
- [x] `hashCrossDomainMessage`
- [x] `hashDepositTransaction`
- [x] `encodeDepositTransaction`
- [x] `hashWithdrawal`
- [x] `hashOutputRootProof`
- [x] `getProveWithdrawalTransactionInputs`

**Misc**
- [x] Clean up binary
    - [x] Use bindings for ABI encoding `WithdrawalTransaction` + `OutputRootProof` types (?) 
- [x] Ditch old TS script

**Metadata**
Closes CLI-3601


clabby:
* develop
  * **PR #5114** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5114" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5114?utm_source=stack-comment).

packages/contracts-bedrock/scripts/differential-testing/differential-testing.go
```diff
@@ -0,0 +1,352 @@
...
+	withdrawalSlot, _ = abi.NewType("tuple", "ArraySlotHash", []abi.ArgumentMarshaling{
+		{Name: "withdrawalHash", Type: "bytes32"},
+		{Name: "zeroPadding", Type: "bytes32"},
+	})
+	withdrawalSlotArgs = abi.Arguments{
+		{Name: "slotHash", Type: withdrawalSlot},
+	}
+
+	// Withdrawal transaction tuple (uint256, address, address, uint256, uint256, bytes)
+	withdrawalTransaction, _ = abi.NewType("tuple", "WithdrawalTransaction", []abi.ArgumentMarshaling{
```
tynes, line 56: Ideally the types generated from the bindings are used when possible

packages/contracts-bedrock/scripts/differential-testing/utils.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/ethereum-optimism/optimism/op-chain-ops/crossdomain"
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+var UnknownNonceVersion = errors.New("Unknown nonce version")
+
+var L2ToL1MessagePasser = common.HexToAddress("0x4200000000000000000000000000000000000016")
```
tynes, line 17: You could import this from the predeploys?

packages/contracts-bedrock/scripts/differential-testing/utils.go
```diff
@@ -0,0 +1,155 @@
...
+
+var EmptyTrieHash = common.HexToHash("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
```
tynes, line 19: This could be imported from go-ethereum

packages/contracts-bedrock/scripts/differential-testing/utils.go
```diff
@@ -0,0 +1,155 @@
...
+	} else {
+		return nil, UnknownNonceVersion
+	}
+
+	return encoded, err
+}
+
+// hashWithdrawal hashes a withdrawal transaction.
+func hashWithdrawal(nonce *big.Int, sender common.Address, target common.Address, value *big.Int, gasLimit *big.Int, data []byte) (common.Hash, error) {
+	// Pack withdrawal
```
tynes, line 58: This could be useful here: https://github.com/ethereum-optimism/optimism/blob/35d4a37c2b9a59e9bf16fa4c5e22283f2348c3d2/op-chain-ops/crossdomain/withdrawal.go#L122

packages/contracts-bedrock/scripts/differential-testing/utils.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return common.Hash{}, err
+	}
+
+	// Hash packed withdrawal (we ignore the pointer)
+	return crypto.Keccak256Hash(packed[32:]), nil
+}
+
+// hashOutputRootProof hashes an output root proof.
+func hashOutputRootProof(version common.Hash, stateRoot common.Hash, messagePasserStorageRoot common.Hash, latestBlockHash common.Hash) (common.Hash, error) {
```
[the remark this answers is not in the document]
clabby: There is, but looks like it passed the params in the wrong order. Fixed it and using it now :+1:

packages/contracts-bedrock/scripts/differential-testing/differential-testing.go
```diff
@@ -0,0 +1,325 @@
...
+
+		// Hash encoded cross domain message
+		hash := crypto.Keccak256Hash(encoded)
+
+		// Pack hash
+		packed, err := fixedBytesArgs.Pack(&hash)
+		checkErr(err, "Error encoding output")
+
+		fmt.Print(hexutil.Encode(packed))
+		break
```
tynes, line 141: breaks aren't necessary here because go has a fallthrough statement, break is default behavior

op-node/rollup/output_root.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // ComputeL2OutputRoot computes the L2 output root
-func ComputeL2OutputRoot(l2OutputRootVersion eth.Bytes32, blockHash common.Hash, blockRoot common.Hash, storageRoot common.Hash) eth.Bytes32 {
+func ComputeL2OutputRoot(l2OutputRootVersion eth.Bytes32, blockRoot common.Hash, storageRoot common.Hash, blockHash common.Hash) eth.Bytes32 {
```
trianglesphere, line 11: Did you check every usage of this function? The type are all the same so this has a large risk of being misued. This also now matches the usage in `HashOutputRootProof`, but that'd cause a mismatch.

trianglesphere (commented):
Go script seems fine to me. I wonder what the overhead of the TS was - maybe startup, I'm not 100% sure.

I'm pretty worried about changing the order of the arguments in `ComputeL2OutputRoot`. I'd be a little more comfortable if we had unit tests + test vectors there.

op-node/rollup/output_root.go
clabby: The only usage my LSP could find was within `op-node/node/api.go` on L117, and I updated that :+1: The `HashOutputRootProof` function was passing them in the order that they are concatenated for hashing, but the params for `ComputeL2OutputRoot` were in the wrong order for some reason.

clabby:
> Go script seems fine to me. I wonder what the overhead of the TS was - maybe startup, I'm not 100% sure.
> 
> I'm pretty worried about changing the order of the arguments in `ComputeL2OutputRoot`. I'd be a little more comfortable if we had unit tests + test vectors there.

Re: The TS overhead- Ran a flamegraph, looked to be primarily ethers' keccak256 as well as the `SecureTrie` datastructure that we were using. RLP encoding wasn't nearly as fast as it is in Go, but wasn't as bad as the others.

tynes (approved):

packages/contracts-bedrock/package.json
```diff
@@ -17,7 +17,7 @@
     "bindings": "cd ../../op-bindings && make",
     "build:forge": "forge build",
     "build:with-metadata": "FOUNDRY_PROFILE=echidna yarn build:forge",
```
tynes, line 19: We can prob get rid of some deps now, the ethereum-js stuff that was only used in diff-testing.ts

clabby: Sure- want to do that in this PR or a separate one?

op-node/rollup/output_root.go
trianglesphere: "wrong order" is in the eye of the beholder. If you properly supplied the function it'd hash them in the correct order.

I looked at the specs & `HashOutputRootProof` was wrong - it assumed that `ComputeL2OutputRoot` would take parameters in order that they'd be hashed. It's definitely a tricky API & probably should be changed, but that was not inherently incorrect.

We also don't appear to use `HashOutputRootProof` - I'm not sure if it was every used.

I'm fine making this change because I'm reasonably confident we found all usages of `ComputeL2OutputRoot`, but we've made a change to the external API where if someone updated it wouldn't throw a compiler error & instead by subtly wrong.

clabby: > "wrong order" is in the eye of the beholder. If you properly supplied the function it'd hash them in the correct order.

Great point- by "wrong order," I meant that the order was "wrong" if you assume that the correct order of the parameters is the order in which they'd be concatenated for hashing (which in my head is the most intuitive).

`HashOutputRootProof` was unused prior to this, but there is a TODO comment to use it in the withdrawal migration script: https://github.com/ethereum-optimism/optimism/blob/develop/op-chain-ops/cmd/withdrawals/main.go#L907-L914 (cc @tynes)

packages/contracts-bedrock/package.json
tynes: Either is fine! Prob could just add to this one

op-node/rollup/output_root.go
trianglesphere: we should fix that in there. Separately, I didn't realize that we had tooling to do withdrawals in op-chain-ops.

packages/contracts-bedrock/package.json
clabby: Done :+1: 

Removed `@ethereumjs/trie` & `@ethereumjs/util`

tynes (approved):

tynes:
Need a review from @trianglesphere or @protolambda on the op-node change

protolambda:
Passing this to @trianglesphere who already commented on the argument order of the compute-output-root function. Using it correctly is essential for the output to be correct & withdrawals to work. Tests should catch this, but if we're changing this we might break some external usages, especially since the typing is the same. It might even be worth changing to a single struct arg, so we can explicitly name the inputs.

tynes (approved):

PR #5116, opened 2023-03-12
fix(spec): use correct to field for contract creation deposits

ncitron:
**Description**

The spec mistakenly defines the `to` field of user deposited transactions to be equal to zero when `isCreated` is `true`. While the emitted log must be enforced to be zero, the actual transaction produced by the derivation process must have `to` set to `null`.

I use the term `null` as that is what Ethereum RPCs will use, but would you like me to instead use `nil` to match up more nicely with go conventions and the geth implementation?


protolambda (approved):
Nice catch. In case of contract-creation it's indeed an empty string in RLP / null in RPC, like L1. An older version of bedrock used the zero-address, but it changed to be more equivalent to L1.

tynes (approved):

Issue #4793, opened 2023-01-26
Malformed batch error on local node, infinite loop


irfanullahatozi:
**Describe the bug**
I am try run a local node of optimism from the community instruction page i followed the exact same steps but when i am run it its gives me malformed batch error on batch submitter docker container.

**To Reproduce**
1.  go to https://community.optimism.io/docs/developers/build/dev-node/
2. follow the instruction in the page after build below error faced
 See error

**Expected behavior**
process is stuck in an infinite loop, and didn't open localhost:8545
**Screenshots**
error screen shot
![image](https://user-images.githubusercontent.com/108583652/214840154-8cedfae7-ba2b-4fa4-ac44-815ae44cf554.png)
batch container logs screen shot
![image](https://user-images.githubusercontent.com/108583652/214840386-191d4380-1ada-4c91-b664-6ad73385609d.png)

**System Specs:**
 - OS: Ubuntu 22.04

**Additional context**
Make me clear if i am doing in a wrong way or its a bug

Issue #4205, opened 2022-12-05
Malformed batch error if there are no sequencer transactions


drinkius:
**Describe the bug**
We are working of creating an Optimism fork and currently trying to deploy the unchanged contracts from the master branch, commit 2b3b09be. When the first bridge transactions are appearing on L2 (although they are failing, which might contribute to the issue) - when the Batch Submitter service tries to create a batch out of those blocks created on l2geth - we are getting the **ErrMalformedBatch** error.

The culprit seems to be hidden in commit [fe680568](https://github.com/ethereum-optimism/optimism/commit/fe6805684da30b6b8d223e9eeff772fff267bb7a) - here there is a condition marked: [_There must be transactions if there are contexts_](https://github.com/ethereum-optimism/optimism/blob/fe6805684da30b6b8d223e9eeff772fff267bb7a/go/batch-submitter/drivers/sequencer/encoding.go#L185) 
If the batch doesn't contain any Sequencer transactions (only the bridge transactions from L1) - then the **AppendSequencerBatchParams** object doesn't contain any Txs since they are added [only for the Sequenced transactions](https://github.com/ethereum-optimism/optimism/blob/fe6805684da30b6b8d223e9eeff772fff267bb7a/go/batch-submitter/drivers/sequencer/batch.go#L140)

Maybe we are missing something in the setup here, or maybe the service has already been updated for Bedrock release which we currently are ignoring and trying to deploy the existing Optimism release

**To Reproduce**
1. Deploy new version of the contracts from commit 2b3b09be
2. Run the L2 from scratch

**Expected behavior**
Batches exclusively with transactions originating from the L1 are written by the Sequencer using Batch Submitter service

**System Specs:**
 - OS: Ubuntu 20.04
 - Package Version (or commit hash): 2b3b09be

smartcontracts:
We're currently all hands on deck getting Bedrock out the door, so work on fixes to the legacy codebase may be slow right now. I'll get this triaged for you.

tynes:
You should consider using the new system over the legacy system, its a lot better

drinkius:
Thanks, when should we expect the Bedrock release then? If you think it's not worth investing time in the legacy system then we'd better wait with the development ourselves. Is it worth trying to run Bedrock testnet for ourselves if it's even a possibility?

tynes:
[...]

PR #5120, opened 2023-03-13
batch-submitter: allow deposit only batches

tynes:
**Description**

Fixes a long standing bug that allows the batch submitter to send batches that only contain deposit transactions. Includes test coverage of the batch creation.

The on chain code requires that there must be at least one context. A context contains information about the queue transactions that should be included. It makes no sense to error when there is a context but no sequencer txs because the context must be submitted to the chain.

Fixes https://github.com/ethereum-optimism/optimism/issues/4793
Fixes https://github.com/ethereum-optimism/optimism/issues/4205
Fixes CLI-3582


maurelian (approved):
I did a quick review of the `appendSequencerBatch()` method and I did not see any possible issues that could occur if there are no sequencer transactions in batch.

PR #5122, opened 2023-03-13
feat(ctb): Optionally print cast commands during migration

maurelian:
**Description**

Allows adding an env var, `CAST_COMMANDS` which when set will print the cast command that should be run to cause the awaited action.

### DIFF
--- a/packages/contracts-bedrock/src/deploy-utils.ts
+++ b/packages/contracts-bedrock/src/deploy-utils.ts
@@ -395,3 +395,15 @@ export const getTenderlySimulationLink = async (
     }).toString()}`
   }
 }
+
+/**
+ * Returns a cast commmand for submitting a given transaction.
+ *
+ * @param tx Ethers transaction object.
+ * @returns the cast command
+ */
+export const getCastCommand = (tx: ethers.PopulatedTransaction): string => {
+  if (process.env.CAST_COMMANDS) {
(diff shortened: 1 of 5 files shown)


packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -396,3 +396,18 @@ export const getTenderlySimulationLink = async (
...
+ * Returns a cast commmand for submitting a given transaction.
+ *
+ * @param tx Ethers transaction object.
+ * @returns the cast command
+ */
+export const getCastCommand = (
+  tx: ethers.PopulatedTransaction
+): string => {
+  if (process.env.CAST_COMMANDS) {
+    return `cast send ${tx.to} ${tx.data} --from ${tx.from} --value ${tx.value}`
```
tynes, line 410: This doesn't work the last time that i tried it - it doesn't accept raw calldata, it expects an interface + args and tries to abi encode. There is a flag on another command that could be used but needs to be added to send

tynes: https://github.com/foundry-rs/foundry/blob/a1f41be934a850251b03c40cbc25e9953ffe0d10/cli/src/cmd/cast/call.rs#L28

`--data`

tynes: dapptools does support this style fwiw

tynes (approved):

tynes:
Looks like this needs a lint then lgtm

PR #5123, opened 2023-03-13
feat(op-batcher): Channel Builder Unit Tests

refcell:
**Description**

Fixes CLI-3610
Fixes CLI-3617

Adds extensive unit testing to the `op-batcher` Channel Builder.


op-batcher/batcher/channel_config_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"testing"
+
+	"github.com/ethereum-optimism/optimism/op-batcher/batcher"
+	"github.com/stretchr/testify/require"
+)
+
+// TestInputThreshold tests the [ChannelConfig.InputThreshold] function.
+func TestInputThreshold(t *testing.T) {
```
trianglesphere, line 11: This test would be a good candidate for table tests.
We only care about `TargetFrameSize, TargetNumFrames, & ApproxComprRatio` here & we have a known output (for each input set) so it's easy to test different combos like `TargetNumFrames > 1`

op-batcher/batcher/channel_builder_test.go
```diff
@@ -0,0 +1,326 @@
...
+
+	// We should be able to increment to the next frame
+	constructedTx, constructedBytes := cb.NextFrame()
+	testSuite.Equal(expectedTx, constructedTx)
+	testSuite.Equal(expectedBytes, constructedBytes)
+	testSuite.Equal(0, cb.NumFrames())
+
+	// The next call should panic since the length of frames is 0
+	defer func() { _ = recover() }()
+	cb.NextFrame()
```
trianglesphere, line 70: There are function is `stretch/testify` to check for panics: https://pkg.go.dev/github.com/stretchr/testify/require?utm_source=godoc#Panics

op-batcher/batcher/channel_builder_test.go
trianglesphere, line 1: One set of tests that would be nice is to make sure we return every full error defined in https://github.com/ethereum-optimism/optimism/blob/80377f723e0fa12e780a74686bf61d1683ee4ff7/op-batcher/batcher/channel_builder.go#L368-L372.

Right now I think it's only `ErrInputTargetReached`

op-batcher/batcher/channel_builder_test.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/trie"
+
+	"github.com/stretchr/testify/suite"
+)
+
+// ChannelBuilderTestSuite encapsulates testing on the ChannelBuilder.
+type ChannelBuilderTestSuite struct {
+	suite.Suite
```
trianglesphere, line 19: The only thing we've standardized on in testing is using `require` instead of `assert`. IMO using `suite` is overkill here, but I also have a pretty strong preference to using stock tooling.

refcell: Agree, may as well keep it consistent. Reverted.

op-batcher/batcher/channel_builder_test.go
sebastianst: Already tracked for a future PR (see Linear) :)

refcell: Had scoped in a separate issue [here](https://linear.app/optimism/issue/CLI-3617/full-channel-coverage-tests), but working on it in this pr now since both you and seb made a note of it :)

trianglesphere: separate PR is fine, just as long as it is being tracked.

trianglesphere, line 1: All the fuzz tests need to be added to a makefile / fuzz runner.
Go is currently only able to run a single fuzz test at a time. If we keep expanding fuzz tests, we may need to create a fuzzing cluster & move it to a long running process.

op-batcher/batcher/channel_builder_test.go
```diff
@@ -0,0 +1,607 @@
...
+	// will exceed [derive.MaxRLPBytesPerChannel]
+	err := cb.AddBlock(block)
+	return err
+}
+
+// FuzzDurationTimeoutZeroMaxChannelDuration ensures that when whenever the MaxChannelDuration
+// is set to 0, the channel builder cannot have a duration timeout.
+func FuzzDurationTimeoutZeroMaxChannelDuration(f *testing.F) {
+	for i := range [10]int{} {
+		f.Add(uint64(i))
```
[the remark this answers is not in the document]
refcell: It should iterate over and seed with 0 to 9, just using it for sanity checking for when fuzzing is not explicitly provided.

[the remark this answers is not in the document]
refcell: Shouldn't hurt in the fuzzing case since it just provides deterministic seeds, and for non-fuzzing makes sure to run the test with non-zero values

op-batcher/batcher/channel_builder_test.go
```diff
@@ -0,0 +1,607 @@
...
+		// Check the timeout
+		cb.timeout = 0
+		cb.updateSwTimeout(&derive.BatchData{
+			BatchV1: derive.BatchV1{
+				EpochNum: rollup.Epoch(epochNum),
+			},
+		})
+		calculatedTimeout := epochNum + seqWindowSize - subSafetyMargin
+		cb.checkTimeout(calculatedTimeout)
+		if cb.timeout != 0 {
```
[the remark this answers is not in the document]
refcell: Yes so when the `calculatedTimeout` is fuzzed as `0`, the previous call to the channel builder `updateSwTimeout` function will update the timeout to be 0, then the `checkTimeout` won't set the full error since the timeout is 0.

op-batcher/batcher/channel_builder_test.go
refcell: Will add to the makefile for now, but let's explore this further offline and open a new ticket for building out a proper fuzz runner?

Happy to take it on

op-batcher/batcher/channel_builder_test.go
```diff
@@ -0,0 +1,607 @@
...
+
+// TestOutputFramesMaxFrameIndex tests the [channelBuilder.OutputFrames]
+// function errors when the max frame index is reached.
+func TestOutputFramesMaxFrameIndex(t *testing.T) {
+	channelConfig := defaultTestChannelConfig
+
+	// Lower the max frame size so that we can test
+	channelConfig.MaxFrameSize = 1
+	channelConfig.TargetNumFrames = math.MaxInt
+	channelConfig.TargetFrameSize = 1 // math.MaxUint64
```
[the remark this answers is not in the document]
trianglesphere: I was specifically referring to "// math.MaxUint64"

op-batcher/batcher/channel_config_test.go
```diff
@@ -33,6 +33,16 @@ func TestInputThreshold(t *testing.T) {
 				require.Equal(t, uint64(2), output)
 			},
 		},
+		{
+			input: testInput{
+				TargetFrameSize:  1,
```
trianglesphere, line 38: target frames size still == 1

op-batcher/batcher/channel_builder_test.go
```diff
@@ -272,7 +291,7 @@ func FuzzSeqWindowCloseZeroTimeout(f *testing.F) {
 		calculatedTimeout := epochNum + seqWindowSize - subSafetyMargin
 		cb.checkTimeout(calculatedTimeout)
 		if cb.timeout != 0 {
-			require.ErrorIsf(t, cb.FullErr(), ErrSeqWindowClose, "Sequence window close should be reached")
+			require.ErrorIs(t, cb.FullErr(), ErrSeqWindowClose)
```
trianglesphere, line 294: you can still have the string on `ErrorIs`

trianglesphere (commented):
I guess I wasn't as clear as I should have been with `ErrorIsf` vs `ErrorIs`. You can directly replace `ErrorIsf` with `ErrorIs` & it will work the same. You can generally (but not always) do the inverse swap because `ErrorIsf` just takes a string as the first varArg while `ErrorIs` takes a generic argument. My intention was not to remove the formatting, but to switch over to `ErrorIs`.

Also, several of these comments got marked as resolved without actually being resolved.

op-batcher/batcher/channel_builder_test.go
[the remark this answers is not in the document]
refcell: oh nope, not intentional

op-batcher/batcher/channel_builder_test.go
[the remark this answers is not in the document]
refcell: I don't have good heuristics for what interesting test cases here are. I think those can be added in the future/another pr.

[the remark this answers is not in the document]
refcell: Want to be scoping prs more or else we'll just have creep

PR #5128, opened 2023-03-14
ops-bedrock: Fix docker-compose file for MacOS

BrianBland:
**Description**

Use the "bind" mode when mounting files (not directories) from the host OS into the ops-bedrock containers. This fixes an issue with docker for Mac where the default volume mount always creates a directory on the container side, even when the path points to a file on the host.

**Tests**

**Invariants**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/ops-bedrock/docker-compose.yml
+++ b/ops-bedrock/docker-compose.yml
@@ -20,8 +20,14 @@ services:
       - "7060:6060"
     volumes:
       - "l1_data:/db"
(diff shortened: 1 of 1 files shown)


tynes:
Can somebody running macos confirm that this works?

trianglesphere (commented):
How did this problem pop up? I've been running this on mac & haven't had issues with the devnet before.

PR #5141, opened 2023-03-14
op-service/txmgr: Better handle price bumps

trianglesphere:
**Description**

This doesn't fully solve the problem that we need to bump both the FeeCap and FeeTip by 10% for geth to accept it, but it handles it slightly better than it did.

**Tests**

Added unit tests + a test that is skipped until we more completely solve the issue.
Fully fixing it requires a larger set of changes.


op-service/txmgr/txmgr.go
```diff
@@ -154,11 +154,22 @@ func (m *SimpleTxManager) IncreaseGasPrice(ctx context.Context, tx *types.Transa
...
+			gasFeeCap = tx.GasFeeCap()
+			reusedFeeCap = true
+		} else {
+			m.l.Debug("Overriding the fee cap to enforce a price bump becasue we increased the tip", "previous", tx.GasFeeCap(), "suggested", gasFeeCap, "new", thresholdFeeCap)
+			gasFeeCap = thresholdFeeCap
+		}
 	} else if thresholdFeeCap.Cmp(gasFeeCap) > 0 {
-		m.l.Debug("Overriding the fee cap to enforce a price bump", "previous", tx.GasFeeCap(), "suggested", gasFeeCap, "new", thresholdFeeCap)
+		if reusedTip {
+			// TODO (CLI-3620): Increase the basefee then recompute the feecap
```
tynes, line 167: any reason to not include this fix in this PR?

trianglesphere: It's a lot of code & would be easier if https://linear.app/optimism/issue/CLI-3548/move-transaction-creation-into-the-txmgr was implemented

protolambda (approved):

tynes (approved):

PR #5142, opened 2023-03-14
op-batcher/proposer: Increase sendTx timeout

trianglesphere:
**Description**

This will hopefully help with some of the send issues we've been seeing.
I'd like to wait a release cycle or two to include this so I can track down
other issues, but long term I think this is the right thing to do.

### DIFF
--- a/op-batcher/batcher/txmgr.go
+++ b/op-batcher/batcher/txmgr.go
@@ -55,7 +55,7 @@ func (t *TransactionManager) SendTransaction(ctx context.Context, data []byte) (
 		return nil, fmt.Errorf("failed to create tx: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


refcell:
Why not just put these behind config variables with default values?

refcell:
Can clamp the values in a runtime validation as well to require code changes for signification changes

trianglesphere:
> Why not just put these behind config variables with default values?

Worth doing in the future once we have better tracking on where txns are (in mempool or if they never hit the mempool). For now we need these changes to be rolled out to increase stability.

refcell (approved):
LGTM

sebastianst (approved):
Also am in favor of putting this timeout into the TxMgrConfig soon, but should help as a quick fix.

PR #5143, opened 2023-03-14
op-node: handle sequencer during catch-up, handle when engine state invalidates attributes

protolambda:
**Description**

Previous race-condition (code of a week ago):
- verifier code queues up payload attributes to force-insert, while syncing batches from L1 eagerly
- verifier code fails to confirm the started block building work
- sequencer tries to build alternative block, accesses the engine, which then first completes the still open work
- verifier code doesn't dequeue the payload attributes, and tries to re-apply them after sequencer finishes
- it's invalid to build a block on top of itself

This was found through logs provided by Conduit, thank you for bug-reporting!

Recently there was a PR to force-cancel building jobs unconditionally: #5065 
This PR fixes the race-condition problem by clearing the state, but introduced other problems:
- temporary errors result in force-cancellation, reducing the chances we can recover from timeouts and such
- the engine-queue part is now responsible for not leaving the engine in a bad state. I want to isolate the engine-queue and engine parts more (it's the biggest part of the derivation, we should split it up in smaller parts), and this counteracts that.

Instead, to really be safe, we can invalidate the payload attributes when the safe-head doesn't match anymore.
We still force-cancel when the error is not temporary.
This can also help avoid other building issues, by ensuring the attributes are coupled strictly to the parent block-ref they were generated for.

The sequencer code also shouldn't have been running in the first place while the engine was busy syncing from L1.
We don't have the notion of "idle derivation" anymore for quite a while now, because derivation interrupts sequencing building time a lot, and shouldn't interfere in the happy case. To avoid conflicts between derivation and sequencing, while being more graceful to the sequencing, we now:
- avoid sequencing when the engine is building a safe block; no point in building an unsafe block instead. And backoff, since more safe blocks are likely to follow
- avoid sequencing when the engine is inconsistent; i.e. when the derivation pipeline is resetting, temporarily disable sequencing altogether.

Sequencing code should still run in other cases, e.g. to produce the L2 chain after downtime, since there might not be any batch data etc.

If it hasn't fully traversed to the tip of the L1 chain, we may also still want to disable L2 sequencing until it is closer to the tip. But this is out of scope of the race condition, and now tracked in CLI-3627.

**Tests**

- engine queue test to reproduce the race-condition, and test that we handle the temporary error + drop the stale payload attributes
- e2e and sequencing code paths cover that the sequencing codepaths still work

**Invariants**

Ensures we never end up building the payload attributes on top of a safe-head that already includes them (and thus avoiding the `parent timestamp == new timestamp` like geth error as reported by Conduit)

**Additional context**

**Metadata**

Fix CLI-3597 (devnet RCA)
Fix CLI-3567 (sequencer RCA)
Fix CLI-3624 (bugfix)

### DIFF
--- a/op-node/rollup/derive/engine_queue.go
+++ b/op-node/rollup/derive/engine_queue.go
@@ -430,6 +430,9 @@ func (eq *EngineQueue) tryNextUnsafePayload(ctx context.Context) error {
 }
 
 func (eq *EngineQueue) tryNextSafeAttributes(ctx context.Context) error {
+	if eq.safeAttributes == nil {  // sanity check the attributes are there
+		return nil
+	}
 	// validate the safe attributes before processing them. The engine may have completed processing them through other means.
 	if eq.safeHead != eq.safeAttributesParent {
 		if eq.safeHead.ParentHash != eq.safeAttributesParent.Hash {
--- a/op-node/rollup/derive/engine_queue.go
+++ b/op-node/rollup/derive/engine_queue.go
@@ -104,8 +104,10 @@ type EngineQueue struct {
 
 	finalizedL1 eth.L1BlockRef
 
-	safeAttributes *eth.PayloadAttributes
-	unsafePayloads PayloadsQueue // queue of unsafe payloads, ordered by ascending block number, may have gaps
+	// The queued-up attributes
+	safeAttributesParent eth.L2BlockRef
+	safeAttributes       *eth.PayloadAttributes
+	unsafePayloads       PayloadsQueue // queue of unsafe payloads, ordered by ascending block number, may have gaps
 
 	// Tracks which L2 blocks where last derived from which L1 block. At most finalityLookback large.
 	finalityData []FinalityData
@@ -225,6 +227,7 @@ func (eq *EngineQueue) Step(ctx context.Context) error {
 		return err
 	} else {
 		eq.safeAttributes = next
+		eq.safeAttributesParent = eq.safeHead
 		eq.log.Debug("Adding next safe attributes", "safe_head", eq.safeHead, "next", eq.safeAttributes)
 		return NotEnoughData
 	}
@@ -427,6 +430,17 @@ func (eq *EngineQueue) tryNextUnsafePayload(ctx context.Context) error {
 }
 
 func (eq *EngineQueue) tryNextSafeAttributes(ctx context.Context) error {
+	// validate the safe attributes before processing them. The engine may have completed processing them through other means.
+	if eq.safeHead != eq.safeAttributesParent {
+		if eq.safeHead.ParentHash != eq.safeAttributesParent.Hash {
+			return NewResetError(fmt.Errorf("safe head changed to %s with parent %s, conflicting with queued safe attributes on top of %s",
+				eq.safeHead, eq.safeHead.ParentID(), eq.safeAttributesParent))
+		}
+		eq.log.Warn("queued safe attributes are stale, safe-head progressed",
+			"safe_head", eq.safeHead, "safe_head_parent", eq.safeHead.ParentID(), "attributes_parent", eq.safeAttributesParent)
+		eq.safeAttributes = nil
+		return nil
+	}
 	if eq.safeHead.Number < eq.unsafeHead.Number {
 		return eq.consolidateNextSafeAttributes(ctx)
 	} else if eq.safeHead.Number == eq.unsafeHead.Number {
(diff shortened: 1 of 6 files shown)


op-node/rollup/derive/engine_queue.go
```diff
@@ -426,6 +429,17 @@ func (eq *EngineQueue) tryNextUnsafePayload(ctx context.Context) error {
 }
 
 func (eq *EngineQueue) tryNextSafeAttributes(ctx context.Context) error {
+	// validate the safe attributes before processing them. The engine may have completed processing them through other means.
```
tynes, line 432: What do you think of having a `nil` check on `eq.safeAttributes` just in case?

op-node/rollup/derive/pipeline.go
```diff
@@ -106,6 +106,12 @@ func NewDerivationPipeline(log log.Logger, cfg *rollup.Config, l1Fetcher L1Fetch
 	}
 }
 
+// EngineReady returns true if the engine is ready to be used.
+// When it's being reset its state is inconsistent, and should not be used externally.
+func (dp *DerivationPipeline) EngineReady() bool {
+	return dp.resetting > 0
```
s7v7nislands, line 112: `return dp.resetting >= len(dp.stages)` ?

protolambda: Only the engine stage status matters here, at position 0, the other stages are free to be in reset-state while the engine is used.

tynes:
if you rebase, it should fix the contracts-bedrocks build issues

op-node/rollup/derive/engine_queue.go
protolambda: done

op-node/rollup/derive/engine_queue_test.go
```diff
@@ -837,3 +844,188 @@ func TestVerifyNewL1Origin(t *testing.T) {
...
+		Transactions:          nil,
+		NoTxPool:              false,
+		GasLimit:              &gasLimit,
+	}
+
+	prev := &fakeAttributesQueue{origin: refA, attrs: attrs}
+	eq := NewEngineQueue(logger, cfg, eng, metrics, prev, l1F)
+	require.ErrorIs(t, eq.Reset(context.Background(), eth.L1BlockRef{}, eth.SystemConfig{}), io.EOF)
+
+	t.Log(refB0)
```
tynes, line 934: Did you mean to leave these logs in there? Also log vs logf

op-node/rollup/driver/sequencer.go
```diff
@@ -29,6 +29,10 @@ type SequencerMetrics interface {
 	RecordSequencerReset()
 }
 
+// Sequencing produces unsafe blocks only, and should not interrupt the L2 block building of safe blocks,
+// e.g. when catching up with an L1 chain with existing batch data.
+const safeBuildInterruptBackoff = 5 * time.Second
```
trianglesphere, line 34: I'm a little worried that this backoff is greater than the L2 block time.

op-node/rollup/derive/engine_queue.go
```diff
@@ -427,6 +430,20 @@ func (eq *EngineQueue) tryNextUnsafePayload(ctx context.Context) error {
 }
 
 func (eq *EngineQueue) tryNextSafeAttributes(ctx context.Context) error {
+	if eq.safeAttributes == nil { // sanity check the attributes are there
+		return nil
+	}
+	// validate the safe attributes before processing them. The engine may have completed processing them through other means.
+	if eq.safeHead != eq.safeAttributesParent {
+		if eq.safeHead.ParentHash != eq.safeAttributesParent.Hash {
```
trianglesphere, line 438: won't we always hit this case & force a reset if `eq.safeHead != eq.safeAttributesParent` is true? My understanding of the code is that if the safeHead has been updated separately from this, we want to cancel this safeAttributes, but we also take this chance to check for a L2 reorg.

I'm also slightly worried about cancelling a safeAttributes here.

op-node/rollup/driver/sequencer.go
tynes: should the backoff be the blocktime?

op-node/rollup/derive/engine_queue.go
protolambda: See PR description, we should have some sanity check that the attributes we have are consistent with the state that we generated them for. If that state isn't consistent anymore, and we didn't reset, it means the engine went ahead and continued to process something that was attempted last. And it'll allow us to separate the engine and the derivation more.

op-node/rollup/driver/sequencer.go
protolambda: If there is a safe block being built, it means it's temporarily busy with constructing from L1 data, and we shouldn't try to sequence at the same time. This is more like an error backoff than a request to sync soon, and should mostly just be higher than the expected block building time to not stumble over the same block building job repeatedly. Setting it to the blocktime works too. Will change it, but I think at best it starts sequencing a tiny bit faster after it completes syncing a block from L1.

trianglesphere (approved):
Seems fine to me. This code is definitely getting diffuse & hard to reason about.

protolambda:
@mergifyio requeu

protolambda:
@mergifyio requeue

protolambda:
Rebased, let's try again

PR #5144, opened 2023-03-14
op-node: support shanghai L1 blockhash input verification

protolambda:
**Description**

The Shapella (Shanghai + Capella) L1 network upgrade introduced a breaking change in the L1 block header format the op-node input-validation was not accounting for yet.
This updates the L1 RPC decoding to pull the withdrawals-root from the json, pass it to the geth header type, and verify the block-hash in a Shanghai-compatible way.

Without this fix, the `--l1.trustrpc` flag is required, or the L1 RPC responses will not be accepted due to the unrecognized input data.

Implementers note: the inclusion of the withdrawals-root is not verified against the L1 time, as that would require L1 chain configuration. Instead it includes it if the data is there, and not if it is not there. It affects the block-hash, so verification is still safe even with a bad L1 RPC, as it can be distinguished from the canonical L1 chain.

**Tests**

Minor test framework change: Update the action-test L1-miner actor to include Shapella data into test L1 blocks.

Tests:
- run through a L1 chain that forks into Shanghai, with a verifier and sequencer. The actors use the standard client bindings and derivation code. Both pre and post Shanghai L1 headers will be traversed through by the verifier, and the L2 chain has L1 origins pointing to pre and post Shanghai L1 blocks.
- unit tests for pre and post-shanghai block hash verification of JSON data

**Invariants**

Accept post-Shanghai L1 block headers as input.

**Metadata**

Fix CLI-3628


op-node/sources/types.go
```diff
@@ -115,6 +118,9 @@ type rpcHeader struct {
 	// BaseFee was added by EIP-1559 and is ignored in legacy headers.
 	BaseFee *hexutil.Big `json:"baseFeePerGas" rlp:"optional"`
 
+	// WithdrawalsRoot was added by EIP-4895 and is ignored in legacy headers.
+	WithdrawalsRoot *common.Hash `json:"withdrawalsRoot" rlp:"optional"`
```
trianglesphere, line 122: Do we every RLP encode or decode this struct? I know we already have the rlp tags, but I don't think we need them.

trianglesphere (changes requested):
rpcHeader fix lgtm, but instead of making an action test can we make unit tests which de-serializes the json of a pre & post shapella block & then verifies the block hash?
This is a very simple deterministic calculation that is well suited to unit testing.

tynes:
If shanghai is turned on at genesis in op-geth, do we need to make any changes there?

trianglesphere (approved):
lgtm

tynes (approved):

protolambda:
rebased on develop

PR #5149, opened 2023-03-15
op-node: fix channel logging

protolambda:
Some important logs were being dropped, make sure to log to the logger, not the global one without handler.


trianglesphere (approved):
Nice catch. I've done quick audits for things like this before, but missed this instance.
We haven't deployed it yet, but in https://github.com/ethereum-optimism/optimism/pull/5086 I set the root logger to be what was configured so if we make this mistake again, it should cause fewer problems.

tynes (approved):

PR #5150, opened 2023-03-15
op-batcher: close channel RPC endpoint

protolambda:
Util to request batcher to close any channel, with frame-number of closing-frame as argument, to avoid having to introspect stateful channel data within the op-batcher


sebastianst:
My take on handling the channel close request in the main loop: https://github.com/ethereum-optimism/optimism/tree/seb/close-channel-rpc

PR #5151, opened 2023-03-15
op-chain-ops: Add parallel migration

mslipper:
After some research, I discovered that we can iterate over Geth's storage in parallel as long as we don't share the state database. This PR updates the OVM ETH check script to parallelize state iteration across 64 workers. The parallelization works by partitioning storage keyspace based on the number of workers. To optimize further, I also moved the state balance checking into the main iteration loop to avoid unnecessary iteration. Note that state mutation must be done serially. Overall, this should safe about 40 minutes during the migration.

This PR was tested by running it against a mainnet data directory. The entire process took approximately one hour. As part of this testing I discovered an invalid check in `check.go`, which was verifying the wrong storage slot in the withdrawals check function. This has been fixed.I confirmed with Mark that the updated `MessageSender` value is the correct one to be checking for. The filtering code was correct from the beginning.

Builds on #5147.

### DIFF
--- a/op-chain-ops/ether/precheck.go
+++ b/op-chain-ops/ether/precheck.go
@@ -1,9 +1,12 @@
 package ether
 
 import (
-	"errors"
 	"fmt"
 	"math/big"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/trie"
 
 	"github.com/ethereum-optimism/optimism/op-chain-ops/crossdomain"
 	"github.com/ethereum-optimism/optimism/op-chain-ops/util"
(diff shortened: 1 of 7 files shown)


op-chain-ops/ether/precheck.go
```diff
@@ -1,142 +1,301 @@
...
-		slotsAct[key] = value
-		count++
-		return true
-	})
-	if err != nil {
-		return nil, fmt.Errorf("cannot iterate over LegacyERC20ETHAddr: %w", err)
-	}
+		// Create a new storage trie. Each trie returned by db.StorageTrie
+		// is a copy, so this is safe for concurrent use.
+		st, err := db.StorageTrie(predeploys.LegacyERC20ETHAddr)
```
tynes, line 110: `st` can technically be `nil` but it never should be in our case, but we might as well check for `nil` for thoroughness

op-chain-ops/ether/precheck.go
```diff
@@ -1,142 +1,301 @@
...
+				log.Crit("mal-formed data in state: %v", err)
+			}
+
+			// We can safely ignore specific slots (totalSupply, name, symbol).
+			if ignoredSlots[key] {
+				continue
+			}
+
+			// No accounts should have a balance in state. If they do, bail.
+			addr := slotsAddrs[key]
```
tynes, line 157: Should we check that `addr != common.Address{}`?

tynes: prob can use `if !ok`

tynes (approved):

PR #5152, opened 2023-03-15
feat(op-batcher): Batcher Config Validation

refcell:
**Description**

Fixes CLI-3618
Fixes CLI-3623

The constraint in the `FramePublished` function can underflow if the op-batcher's `ChannelTimeout` and `SubSafetyMargin` are misconfigured where `SubSafetyMargin < ChannelTimeout`.

This pr adds validation in the `op-batcher` initialization that returns an error if `SubSafetyMargin < ChannelTimeout`.

Additionally, if `MaxFrameSize` is set to 0, the `op-batcher` will enter an infinite loop in the Channel Builder's `outputReadyFrames` method since we have:

```go
for c.co.ReadyBytes() >= int(c.cfg.MaxFrameSize) {
    ...
}
```


tynes (approved):

refcell:
@mergifyio requeue

PR #5155, opened 2023-03-15
op-e2e: test high-demand large L2 blocks case

protolambda:
**Description**

Test to ensure the channel functionality works as expected in case of high-demand for lots of L2 calldata. This ensures we can handle batch-derivation when L2 blocks grow larger than L1 blocks and batcher data-tx target size.

This also includes some minor changes to the action test actors, to modify tx-options of the batcher, catch the full-channel error, and be more explicit about tx-inclusion (to not rely on pending-block for nonce values and such).


tynes (approved):

trianglesphere (approved):
may need to rebase to fix CI

protolambda:
rebased

PR #5167, opened 2023-03-15
Batcher Metrics

sebastianst:
**Description**

This PR adds metrics to the batcher. The following metrics are recorded:
* channel 
  * opened
  * closed
  * fully submitted
  * timed out
* L2 blocks loaded into queue
* L2 blocks added to channel
* batch transaction
  * submitted
  * success
  * failed

Some metrics are implemented as simple events (like tx success), other contain many data dimensions, like `RecordL2BlocksAdded`.

Also added info logging of different stats when channels are closed. This has also long been overdue.

**Open Tasks**

Some open tasks were not done yet as to limit this PR's size.

- CLI-3642 - Some reusable components were taken from the `op-node` metrics, namely `RefMetrics` and `EventMetrics`, and copied into package `op-service/metrics`. The duplicated code should be removed in a follow-up by making the `op-node` use these.
- CLI-3641 - Record channel IDs and transaction IDs similar to how block hashes are recorded, to allow for a visual analysis of changing channels/transactions.
- CLI-3640 - Record reasons for channel closing. A dependency cycle has to be resolved for that.

**Metadata**
- Fixes CLI-3240

### DIFF
--- a/op-batcher/batcher/batch_submitter.go
+++ b/op-batcher/batcher/batch_submitter.go
@@ -12,9 +12,9 @@ import (
 	gethrpc "github.com/ethereum/go-ethereum/rpc"
 	"github.com/urfave/cli"
 
+	"github.com/ethereum-optimism/optimism/op-batcher/metrics"
 	"github.com/ethereum-optimism/optimism/op-batcher/rpc"
 	oplog "github.com/ethereum-optimism/optimism/op-service/log"
-	opmetrics "github.com/ethereum-optimism/optimism/op-service/metrics"
 	oppprof "github.com/ethereum-optimism/optimism/op-service/pprof"
 	oprpc "github.com/ethereum-optimism/optimism/op-service/rpc"
 )
@@ -36,9 +36,10 @@ func Main(version string, cliCtx *cli.Context) error {
 	}
 
 	l := oplog.NewLogger(cfg.LogConfig)
+	m := metrics.NewMetrics("default")
 	l.Info("Initializing Batch Submitter")
 
-	batchSubmitter, err := NewBatchSubmitterFromCLIConfig(cfg, l)
+	batchSubmitter, err := NewBatchSubmitterFromCLIConfig(cfg, l, m)
 	if err != nil {
 		l.Error("Unable to create Batch Submitter", "error", err)
 		return err
@@ -64,16 +65,15 @@ func Main(version string, cliCtx *cli.Context) error {
 		}()
 	}
 
-	registry := opmetrics.NewRegistry()
 	metricsCfg := cfg.MetricsConfig
 	if metricsCfg.Enabled {
 		l.Info("starting metrics server", "addr", metricsCfg.ListenAddr, "port", metricsCfg.ListenPort)
 		go func() {
-			if err := opmetrics.ListenAndServe(ctx, registry, metricsCfg.ListenAddr, metricsCfg.ListenPort); err != nil {
+			if err := m.Serve(ctx, metricsCfg.ListenAddr, metricsCfg.ListenPort); err != nil {
 				l.Error("error starting metrics server", err)
 			}
 		}()
-		opmetrics.LaunchBalanceMetrics(ctx, l, registry, "", batchSubmitter.L1Client, batchSubmitter.From)
+		m.StartBalanceMetrics(ctx, l, batchSubmitter.L1Client, batchSubmitter.From)
 	}
 
 	rpcCfg := cfg.RPCConfig
@@ -94,6 +94,9 @@ func Main(version string, cliCtx *cli.Context) error {
 		cancel()
 		return fmt.Errorf("error starting RPC server: %w", err)
 	}
+
+	m.RecordInfo(version)
+	m.RecordUp()
 
 	interruptChannel := make(chan os.Signal, 1)
 	signal.Notify(interruptChannel, []os.Signal{
--- a/op-batcher/batcher/channel_builder.go
+++ b/op-batcher/batcher/channel_builder.go
@@ -136,6 +136,8 @@ type channelBuilder struct {
 	blocks []*types.Block
 	// frames data queue, to be send as txs
 	frames []frameData
+	// total amount of output data of all frames created yet
+	outputBytes int
 }
 
 // newChannelBuilder creates a new channel builder or returns an error if the
@@ -156,9 +158,19 @@ func (c *channelBuilder) ID() derive.ChannelID {
 	return c.co.ID()
 }
 
-// InputBytes returns to total amount of input bytes added to the channel.
+// InputBytes returns the total amount of input bytes added to the channel.
 func (c *channelBuilder) InputBytes() int {
 	return c.co.InputBytes()
+}
+
+// ReadyBytes returns the amount of bytes ready in the compression pipeline to
+// output into a frame.
+func (c *channelBuilder) ReadyBytes() int {
+	return c.co.ReadyBytes()
+}
+
+func (c *channelBuilder) OutputBytes() int {
+	return c.outputBytes
 }
 
 // Blocks returns a backup list of all blocks that were added to the channel. It
@@ -381,10 +393,11 @@ func (c *channelBuilder) outputFrame() error {
 	}
 
 	frame := frameData{
-		id:   txID{chID: c.co.ID(), frameNumber: fn},
+		id:   frameID{chID: c.co.ID(), frameNumber: fn},
 		data: buf.Bytes(),
 	}
 	c.frames = append(c.frames, frame)
+	c.outputBytes += len(frame.data)
 	return err // possibly io.EOF (last frame)
 }
 
--- a/op-batcher/batcher/channel_builder.go
+++ b/op-batcher/batcher/channel_builder.go
@@ -185,21 +185,21 @@ func (c *channelBuilder) Reset() error {
 // already full. See description of FullErr for details.
 //
 // Call OutputFrames() afterwards to create frames.
-func (c *channelBuilder) AddBlock(block *types.Block) error {
+func (c *channelBuilder) AddBlock(block *types.Block) (derive.L1BlockInfo, error) {
 	if c.IsFull() {
-		return c.FullErr()
+		return derive.L1BlockInfo{}, c.FullErr()
 	}
 
-	batch, err := derive.BlockToBatch(block)
+	batch, l1info, err := derive.BlockToBatch(block)
 	if err != nil {
-		return fmt.Errorf("converting block to batch: %w", err)
+		return l1info, fmt.Errorf("converting block to batch: %w", err)
 	}
 
 	if _, err = c.co.AddBatch(batch); errors.Is(err, derive.ErrTooManyRLPBytes) {
 		c.setFullErr(err)
-		return c.FullErr()
+		return l1info, c.FullErr()
 	} else if err != nil {
-		return fmt.Errorf("adding block to channel out: %w", err)
+		return l1info, fmt.Errorf("adding block to channel out: %w", err)
 	}
 	c.blocks = append(c.blocks, block)
 	c.updateSwTimeout(batch)
@@ -209,7 +209,7 @@ func (c *channelBuilder) AddBlock(block *types.Block) error {
 		// Adding this block still worked, so don't return error, just mark as full
 	}
 
-	return nil
+	return l1info, nil
 }
 
 // Timeout management
--- a/op-batcher/batcher/channel_builder.go
+++ b/op-batcher/batcher/channel_builder.go
@@ -196,6 +196,9 @@ func (c *channelBuilder) Reset() error {
 // AddBlock returns a ChannelFullError if called even though the channel is
 // already full. See description of FullErr for details.
 //
+// AddBlock also returns the L1BlockInfo that got extracted from the block's
+// first transaction for subsequent use by the caller.
+//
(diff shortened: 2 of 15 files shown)


op-node/rollup/derive/l2block_util.go
```diff
@@ -0,0 +1,64 @@
...
+	Hash() common.Hash
+	ParentHash() common.Hash
+	NumberU64() uint64
+	Time() uint64
+	Transactions() types.Transactions
+}
+
+// PayloadToBlockRef extracts the essential L2BlockRef information from an L2
+// block ref source, falling back to genesis information if necessary.
+func L2BlockToBlockRef(block L2BlockRefSource, genesis *rollup.Genesis) (eth.L2BlockRef, error) {
```
trianglesphere, line 28: I assume this is copied from `PayloadToBlockRef`. Fine for now, but we are starting to get a proliferation of types.

op-batcher/batcher/batch_submitter.go
```diff
@@ -36,9 +36,10 @@ func Main(version string, cliCtx *cli.Context) error {
 	}
 
 	l := oplog.NewLogger(cfg.LogConfig)
+	m := metrics.NewMetrics("default")
```
trianglesphere, line 39: When would we change "default" to something else?

op-batcher/batcher/channel_manager.go
```diff
@@ -273,3 +333,14 @@ func (s *channelManager) AddL2Block(block *types.Block) error {
 
 	return nil
 }
+
+func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo) eth.L2BlockRef {
```
trianglesphere, line 337: Shouldn't this be pulling in from `l2block_util.go`?

trianglesphere (commented):
this is exciting

op-node/rollup/derive/l2block_util.go
sebastianst: It is. I actually also implemented a converter for an `ExecutionPayload` into a `L2BlockRefSource` and then this could be reused, but then discarded that as too fancy.

op-batcher/batcher/batch_submitter.go
sebastianst: I just copied that from all other metrics impls. Wondering myself 😅

op-batcher/batcher/channel_manager.go
sebastianst: We could but here we have the situation that we don't need and want to extract the `l1info` again, so made this local helper. The extraction is extra work which I saved here. But there's also an argument to favor simplicity.

op-node/rollup/derive/l2block_util.go
sebastianst: The reason I added this is to avoid a marshalling roundtrip. A block could also first be converted into an exec payload. But I found that inefficient.

op-batcher/batcher/batch_submitter.go
trianglesphere: Ok sounds good.

trianglesphere:
@sebastianst needs a go lint on `l2block_utils.go`

trianglesphere:
Also, all TODO's need linear tickets to pass semgrep.

tynes:
This is looking good

op-batcher/batcher/channel_builder.go
```diff
@@ -147,11 +149,21 @@ func (c *channelBuilder) ID() derive.ChannelID {
...
 
+// ReadyBytes returns the amount of bytes ready in the compression pipeline to
+// output into a frame.
+func (c *channelBuilder) ReadyBytes() int {
+	return c.co.ReadyBytes()
+}
+
+func (c *channelBuilder) OutputBytes() int {
+	return c.outputBytes
+}
```
refcell, line 165: Super nice to have this - wanted to add in my batcher prs but figured it was out of scope.

op-batcher/batcher/channel_builder.go
```diff
@@ -175,22 +187,25 @@ func (c *channelBuilder) Reset() error {
...
 	if c.IsFull() {
-		return c.FullErr()
+		return derive.L1BlockInfo{}, c.FullErr()
 	}
 
-	batch, err := derive.BlockToBatch(block)
+	batch, l1info, err := derive.BlockToBatch(block)
 	if err != nil {
-		return fmt.Errorf("converting block to batch: %w", err)
+		return l1info, fmt.Errorf("converting block to batch: %w", err)
```
refcell, line 201: Not necessarily in the scope of this pr, but defining a custom error type might help tracking errors in metrics and provide a cleaner way for comparison and externally interacting with the batcher.

op-batcher/batcher/channel_builder_test.go
```diff
@@ -490,7 +504,7 @@ func TestBuilderAddBlock(t *testing.T) {
 	require.NoError(t, err)
 
 	// Add a nonsense block to the channel builder
-	err = addNonsenseBlock(cb)
+	err = addMiniBlock(cb)
 	require.NoError(t, err)
```
refcell, line 508: Can inline all of these to be more idiomatic

```suggestion
	require.NoError(t, addMiniBlock(cb))
```

op-batcher/batcher/channel_builder_test.go
```diff
@@ -618,3 +632,75 @@ func TestFramePublished(t *testing.T) {
...
+		l += blockBatchRlpSize(t, block)
+
+		_, err := cb.AddBlock(block)
+		require.NoError(err)
+		require.Equal(cb.InputBytes(), l)
+	}
+}
+
+func TestChannelBuilder_OutputBytes(t *testing.T) {
+	require := require.New(t)
```
refcell, line 655: You shouldn't need to do this, `require` is already imported from testify.

op-batcher/batcher/channel_builder_test.go
```diff
@@ -618,3 +632,75 @@ func TestFramePublished(t *testing.T) {
 	// Now the timeout will be 1000
 	require.Equal(t, uint64(1000), cb.timeout)
 }
+
+func TestChannelBuilder_InputBytes(t *testing.T) {
+	require := require.New(t)
```
refcell, line 637: ibid

op-batcher/batcher/channel_builder_test.go
```diff
@@ -618,3 +632,75 @@ func TestFramePublished(t *testing.T) {
...
+		f := cb.NextFrame()
+		frames = append(frames, f)
+		flen += len(f.data)
+	}
+
+	require.Equal(cb.OutputBytes(), flen)
+}
+
+func defaultChannelBuilderSetup(t *testing.T) (*channelBuilder, ChannelConfig) {
+	t.Helper()
```
refcell, line 692: Personally think it doesn't make sense to pass testing into helpers at all. Just bubble up the error and allow each test to handle respectively, but this is ok

op-batcher/batcher/driver.go
```diff
@@ -316,6 +334,14 @@ func (l *BatchSubmitter) loop() {
...
 }
 
+func (l *BatchSubmitter) recordL1Tip(l1tip eth.L1BlockRef) {
+	if l.lastL1Tip == l1tip {
+		return
+	}
+	l.lastL1Tip = l1tip
+	l.metr.RecordLatestL1Block(l1tip)
+}
+
```
refcell, line 344: Nice! I don't see this being used yet in the pr - you'll want to change the driver's block loop to use this.

op-batcher/batcher/driver.go
```diff
@@ -203,24 +211,34 @@ func (l *BatchSubmitter) loadBlocksIntoState(ctx context.Context) {
...
 	block, err := l.L2Client.BlockByNumber(ctx, new(big.Int).SetUint64(blockNumber))
-	cancel()
 	if err != nil {
-		return eth.BlockID{}, err
+		return nil, fmt.Errorf("getting L2 block: %w", err)
 	}
+
 	if err := l.state.AddL2Block(block); err != nil {
-		return eth.BlockID{}, err
+		return nil, fmt.Errorf("adding L2 block to state: %w", err)
```
refcell, line 237: Use defined errors?

op-batcher/batcher/channel_manager_test.go
```diff
@@ -101,29 +101,17 @@ func TestChannelManagerReturnsErrReorg(t *testing.T) {
...
-	x := types.NewBlock(&types.Header{
-		Number:     big.NewInt(1),
-		ParentHash: common.Hash{0xff},
-	}, txs, nil, nil, trie.NewStackTrie(nil))
+	a := newMiniL2Block(0)
+	x := newMiniL2BlockWithNumberParent(0, big.NewInt(1), common.Hash{0xff})
 
-	err = m.AddL2Block(a)
+	err := m.AddL2Block(a)
 	require.NoError(t, err)
```
refcell, line 115: Can just inline

```suggestion
	require.NoError(t, m.AddL2Block(a))
```

op-batcher/batcher/channel_manager.go
```diff
@@ -194,8 +204,8 @@ func (s *channelManager) TxData(l1Head eth.BlockID) (txData, error) {
 	// all pending blocks be included in this channel for submission.
 	s.registerL1Block(l1Head)
 
-	if err := s.pendingChannel.OutputFrames(); err != nil {
-		return txData{}, fmt.Errorf("creating frames with channel builder: %w", err)
+	if err := s.outputFrames(); err != nil {
+		return txData{}, err
```
refcell, line 208: 🙏

refcell (changes requested):
Few minor nits

op-batcher/batcher/channel_builder.go
sebastianst: Custom Go error types should be used sparingly and only make sense if there’s a real chance that the caller receiving this error might react differently than to other errors. Otherwise they just clutter the package’s API. If you look e.g. at the standard library's packages, they only each provide a handful of custom errors, if at all, and only for those where there's a real chance the calling code might react to it. Good example is `io.EOF`, which is an "expected" error that the caller often reacts differently to. In this sense, error types are usually just used for "expected" errors that still represent a healthy program flow.

This example here is a good example of an error that does _not_ represent a healthy program flow. If the channel builder receives invalid block data, something is seriously wrong with the RPC client delivering blocks. I don't see a realistic scenario right now where the calling code would react to this differently. 

The default should be to just return `fmt.Errorf()` simple errors and only introduce a custom error type once _needed_. Otherwise it's premature optimization and clutters the package API.

op-batcher/batcher/channel_builder_test.go
sebastianst: I agree and also do it this way. This particular diff here is not in a test I wrote but someone else. I just renamed the function `addNonsenseBlock` to `addMiniBlock` in this commit and that's where this diff comes from. I didn't want to further blow up this PR by refactoring other tests. So I'll leave it at this. Next time this test's content is touched, the author should fix it.

op-batcher/batcher/channel_builder_test.go
sebastianst: See [docs](https://pkg.go.dev/github.com/stretchr/testify/require#New), this creates an assertion object `require` (that shadows the `require` package) bound to the `t *testing.T` so that assertions in this test don't need to always specify `t`. E.g. after this, we can write `require.NoError(SomeFunc())` instead of `require.NoError(t, SomeFunc())`.

See also the [`testify's README`](https://github.com/stretchr/testify#assert-package) section about `assert`, they recommend this pattern of shadowing the package name with `.New(t)` if multiple assertions are made in a test.

I recommend reading the testify go docs and readme, they have many hidden gems.

op-batcher/batcher/channel_builder_test.go
sebastianst: see above answer.

op-batcher/batcher/channel_builder_test.go
sebastianst: It's a common pattern, which is why there's `t.Helper()` so that helper functions are not part of a stack trace when an error happens. If there's a common way of handling an error, you save some repetition in each calling test by just doing the right assertions in the helper function.

op-batcher/batcher/driver.go
sebastianst: Thanks for making me aware! I forgot to call it in the `blockLoop`, which was the point of creating this method. I was already wondering why the latest l1 block didn't show up in the metrics...

op-batcher/batcher/driver.go
sebastianst: See above answer, avoid custom error types unless there's a specific reason.

op-batcher/batcher/channel_manager_test.go
sebastianst: See above response, not my test.

op-batcher/batcher/channel_builder.go
sebastianst: Yes :) Added it in this PR because I needed it for metrics and the evaluation of the compression ratio.

sebastianst:
Output at `http://localhost:7301/` after letting the batcher running for some time: [batcher-metrics-prometheus.raw.txt](https://github.com/ethereum-optimism/optimism/files/11018160/batcher-metrics-prometheus.raw.txt)

PR #5168, opened 2023-03-16
feat(op-batcher): Full E2E Reorg Test

refcell:
**Description**

Fixes CLI-3631

This is also intended to check CLI-3632

Adds a full e2e test to the `op-batcher` in an effort to test worst-case reorgs.


op-batcher/batcher/batch_submitter.go
```diff
@@ -73,7 +73,11 @@ func Main(version string, cliCtx *cli.Context) error {
 				l.Error("error starting metrics server", err)
 			}
 		}()
-		opmetrics.LaunchBalanceMetrics(ctx, l, registry, "", batchSubmitter.L1Client, batchSubmitter.From)
+		l1Client, err := dialEthClientWithTimeout(ctx, cfg.L1EthRpc)
```
trianglesphere, line 76: Why do you need to dial this separately?

op-batcher/batcher/txmgr.go
```diff
@@ -19,25 +18,48 @@ import (
 
 const networkTimeout = 2 * time.Second // How long a single network request can take. TODO: put in a config somewhere
 
+// ExternalTxManager is a minimal interface for the [TransactionManager]
+// that is a subset of [txmgr.SimpleTxManager] functionality
+//
+//go:generate mockery --name ExternalTxManager --output ./mocks
```
trianglesphere, line 24: This seems like an interesting tool, but I'm going to look into it a bit more before I feel comfortable adding it. At a high level I'm not sure if mocks (& asserting that they are called) is the right framework. We don't care how they are interacted with, but instead care a lot about the values that they return.

op-batcher/batcher/batch_submitter.go
```diff
@@ -73,7 +73,11 @@ func Main(version string, cliCtx *cli.Context) error {
...
+		if err != nil {
+			l.Error("Failed to connect to L1 provider to enable ")
```
ajsutton, line 78: This feels like an unfinished sentence (at least the trailing space seems weird)

op-batcher/batcher/config.go
```diff
@@ -1,29 +1,61 @@
...
 	oppprof "github.com/ethereum-optimism/optimism/op-service/pprof"
 	"github.com/ethereum-optimism/optimism/op-service/txmgr"
 	opsigner "github.com/ethereum-optimism/optimism/op-signer/client"
 )
 
+// L1ConfigProvider is a minimal interface that allows the batch submitter to query
+// the L1 chain, and is coalesced into a [txmgr.ETHBackend] for the [TransactionManager].
+//
+//go:generate mockery --name L1ConfigProvider --output ./mocks
+type L1ConfigProvider interface {
```
ajsutton, line 28: `*ConfigProvider` is an unexpected name for these interfaces. Is it really config we're getting here? Something like `L1Chain` or `L1DataProvider` seems more accurate to me.  It is referenced from the batcher's `Config` but I'd still name it based on what role it serves rather than how it's passed in.

op-batcher/batcher/driver_test.go
```diff
@@ -0,0 +1,252 @@
...
+func constructDefaultBatchSubmitter(l log.Logger, mockTxMgr ExternalTxManager, l1Client L1ConfigProvider, l2Client L2ConfigProvider, rollupNode RollupNodeConfigProvider) (*BatchSubmitter, error) {
+	resubmissionTimeout, err := time.ParseDuration("30s")
+	if err != nil {
+		return nil, err
+	}
+
+	pollInterval, err := time.ParseDuration("5s")
+	if err != nil {
+		return nil, err
+	}
```
ajsutton, line 35:
```suggestion
	resubmissionTimeout := 30 * time.Second
	pollInterval := 5 * time.Second
```

op-batcher/batcher/driver_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+	return &b, nil
+}
+
+// TestDriverLoadBlocksIntoState ensures that the [BatchSubmitter] can load blocks into the state.
+func TestDriverLoadBlocksIntoState(t *testing.T) {
```
ajsutton, line 94: hmm, this is quite a long test to try and read through and the majority of it is setting up mocks which makes it pretty hard to follow and quite dependent on the internal implementation and how it interacts with the mocks.

Given that testing the batcher well depends on being able to control the L1 and L2 chains to setup the different scenarios, I wonder if we should take some time to build suitable stubs or some other kind of library code to make it easier for tests to express what those chains should do.

Maybe we could jump on a call early next week and play around with some options together? I'm not so familiar with this bit of the codebase so not exactly sure what the best options would be but it seems like a good opportunity to improve our testing tools to make this kind of test easier in general.

refcell: Yea well so this was intended to be as minimal as possible right. Can take offline but introducing a library is a lot of work given that we have an ambitious roadmap to execute on. This will probably be hard enough to get merged, I'd rather spend time and cycles on the most important parts of our roadmap than libraries personally.

refcell: If we had written the codebase to use proper minimal interfaces, this would have been easy to implement from within `op-e2e`, which is actually where I started writing this pr. But a lot of constructors and internals/unexported funcs and fields take concrete structs and struct pointers - which is impossible to mock in a clean way from an external package like `op-e2e`.

op-batcher/batcher/driver_test.go
refcell: lmao, 10x dev for real

op-batcher/batcher/batch_submitter.go
refcell: HAHAHA you caught me 😅

op-batcher/batcher/batch_submitter.go
refcell: Very unfortunately because `opmetrics` takes the explicit struct pointer `*ethclient.Client` which I cannot re-use the `batchSubmitter`'s L1Client interface for

op-batcher/batcher/config.go
refcell: Renamed to data providers - that sounds a lot cleaner 🙏

op-batcher/batcher/txmgr.go
refcell: Agree - I'm open to changing this pr up **a lot**. It's very much a PoC as opposed to a mergable - though we will need to test once we have a fix for flushing the batcher on reorgs.

op-batcher/batcher/driver_test.go
ajsutton: Yeah, it's definitely hard to test a lot of this stuff, and coming along afterwards to try and add tests is especially hard. The catch though is if we keep saying it's too hard to write good tests and would take too much time to build the testing infrastructure we need or refactor the code to be more testable, the more untestable code we add and the more time we spend chasing our tails fixing regressions that weren't caught by tests and/or the more time we spend trying to maintain brittle, complicated tests.  Testing well and taking time to refactor is actually what makes you able to go faster.

That said, we may only wind up taking a small step to improving things with this particular test and it winds up looking much like this rather than getting something really clean, but its important that each change we make at least makes things a little better.

ajsutton: Had a play with this today to get a better feel for what's possible. I think your refactoring in this PR actually make it a lot easier to get the batcher into the e2e framework and that already has the control we need to be able to create L1 and L2 reorgs.

So I came up with https://github.com/ethereum-optimism/optimism/commit/6814bfae89dffdb84d131f7bc10eb801768de0d8 (the `aj/batcher-e2e` branch) which is still fairly messy but I think is a reasonable start.  It's running the batcher entirely async but using the synchronous miner and sequencer so it's a bit of a weird mix of system and actions test at the moment - that's probably ok, just need to move the file to somewhere sensible and make it clear what we're actually trying to test.

I did need to make a couple of small changes to the batcher itself to make things more deterministic. Instead of calling `Start()` and having it run entirely async, I exposed a `Step()` method which is the code that runs on each tick so now the test can essentially just control when those ticks occur.

In terms of asserting the state of the batcher, we may need to add some accessor functions to `BatchSubmitter` which would be reasonable, or we may be able to hook into the batcher metrics. Using metrics like that is always nice because then the metrics are tested and you know you have enough information being exposed to properly reason about the state of the system - the test is basically doing what a person would do when investigating an issue.

refcell: Nice! Yes that's the goal - make it much easier to test from within a completely external package like op-e2e.

sebastianst:
Converting to draft until we have decided on a strategy on how to improve batcher e2e testing. We learned a lot from this PR already 💡

refcell:
Closing and placing on the backlog.

PR #5170, opened 2023-03-16
specs: Defines an explicit expiry time for channels

ajsutton:
**Description**

Previously it was expected that seen channels were tracked indefinitely so later frames from that channel could be ignored. Setting an explicit timeframe for expiry provides an upper bound on the length of time that such channels need to be tracked.

**Invariants**

For a channel to be valid:
* All frames of a channel must be included within `CHANNEL_TIMEOUT` blocks of the first frame
* No channel expires before it would have timed out or been completed

**Additional context**

Putting this up as draft just for visibility. There needs to be further investigation done on the risks of this change causing a chain split compared to both the current spec and the actual implementation.

Already known is that the current implementation would differ from both the existing and updated spec, in the case where a channel is validly submitted and then, at any point after the frames for that channel is read or times out and within the sequence window of the first block it had a frame in,  the same channel ID is reused for a second, valid, channel. The existing implementation would have pruned the first channel when it timed out or was read and then considered the second channel valid even though it reused the ID. With this change implemented it would consider that second channel invalid because of the channel ID reuse.

Not that even with this change, safe head will still be unable to advance if there is a channel left incomplete until the channel timeout is reached. It does however prevent the case noted in https://github.com/ethereum-optimism/optimism/pull/5107 where an invalid extra frame was included the last frame of a channel within the sequencing window.

The sequencing window was used for the expiry timeline on the assumption that the node would already need to scan blocks at least that far before it's safe head at startup in order to guarantee it can find the data for the next batch on-chain.

**Metadata**

### DIFF
--- a/specs/derivation.md
+++ b/specs/derivation.md
@@ -517,23 +517,30 @@ New frames for timed-out channels are dropped instead of buffered.
 
 The channel-bank can only output data from the first opened channel.
 
-Upon reading, first all timed-out channels are dropped.
+Upon reading, first all expired channels are removed from the known complete channels.
+
+A channel is expired if:
+`current_l1_block.number > channel.open_l1_block.number + max(SEQUENCING_WINDOW_SIZE, CHANNEL_TIMEOUT)`
+
+After expiring known channels, all timed-out channels are moved to the set of known complete channels.
 
 After pruning timed-out channels, the first remaining channel, if any, is read if it is ready:
 
 - The channel must be closed
 - The channel must have a contiguous sequence of frames until the closing frame
 
+After being read, the channel is moved to the set of known complete channels.
+
 If no channel is ready, the next frame is read and ingested into the channel bank.
 
 #### Loading frames
 
-When a channel ID referenced by a frame is not already present in the Channel Bank,
+When a channel ID referenced by a frame is not already present in the Channel Bank, or in the known complete channels,
 a new channel is opened, tagged with the current L1 block, and appended to the channel-queue.
 
 Frame insertion conditions:
 
-- New frames matching existing timed-out channels are dropped.
+- New frames matching known complete channels are dropped.
 - Duplicate frames (by frame number) are dropped.
 - Duplicate closes (new frame `is_last == 1`, but the channel has already seen a closing frame) are dropped.
 


specs/derivation.md
```diff
@@ -517,23 +517,29 @@ New frames for timed-out channels are dropped instead of buffered.
 
 The channel-bank can only output data from the first opened channel.
 
-Upon reading, first all timed-out channels are dropped.
+Upon reading, first all expired channels are removed from the known complete channels.
+
+A channel is expired if: `current_l1_block.number > channel.open_l1_block.number + max(SEQUENCING_WINDOW_SIZE, CHANNEL_TIMEOUT)`
+
```
ajsutton, line 523: I would expect that `SEQUENCING_WINDOW_SIZE` is always larger than `CHANNEL_TIMEOUT` but I haven't seen anywhere that explicitly requires it, hence the max.

ajsutton:
A more radical alternative (that definitely needs more thinking through and is more likely to require a hard fork to activate) would be to change the order that channels are read.  Currently the ordering of channels is decided based on the order of their first frames. As a result, if a channel is opened but not closed, the safe head stalls until the channel timeout is reached even though valid channel data may have been available.

If instead we processed channels based on the order that they were completed (ie ready to read), any unclosed channels would just remain pending and derivation would proceed with the channels that were ready.  We could then safely prune channels as soon as they timeout or are read (as per the current code) as while an extra trailing frame would still open a new channel, it wouldn't block processing.  It would also be a more general solution, preventing safe head from stalling on any incomplete channels.

The derivation pipeline already has to reorder batches that are received out of order from channels (waiting until the `batch.timestamp == nextTimestamp` so the derivation pipeline will still wait for valid batches. However if there are multiple valid batches submitted in different, valid channels that overlap on L1, it's possible a different chain would be selected because of the different channel ordering.

Also note that the channel validity rules already require all frames from the channel to have been read before any batches from the channel are processed (because all batches must be discarded if the channel is not completed before its timeout).

specs/derivation.md
```diff
@@ -517,23 +517,30 @@ New frames for timed-out channels are dropped instead of buffered.
 
 The channel-bank can only output data from the first opened channel.
 
-Upon reading, first all timed-out channels are dropped.
+Upon reading, first all expired channels are removed from the known complete channels.
+
+A channel is expired if:
+`current_l1_block.number > channel.open_l1_block.number + max(SEQUENCING_WINDOW_SIZE, CHANNEL_TIMEOUT)`
```
trianglesphere, line 523: min?

trianglesphere (commented):
Is the known complete for future changes?

specs/derivation.md
ajsutton: No we want the max of the two. Min is essentially what the op-node is doing now - treating channels as unknown as soon as they timeout.  We want to extend the period that the channel is known so extra frames that are received for those channels can be ignored, rather than have them start a new channel that we have to wait for it to timeout.

ajsutton:
> Is the known complete for future changes?

@trianglesphere I'm not sure what you're asking here.

specs/derivation.md
ajsutton: Note that the channel will still timeout after `CHANNEL_TIMEOUT` - from then until the end of the sequencing window any frames for that channel will be immediately dropped. After the sequencing window frames with that channel ID will be treated as a new channel again.

ajsutton:
Closing this. It has similar risks when done as a soft fork to ordering channels by the last frame received (https://github.com/ethereum-optimism/optimism/pull/5196) but doesn't solve as many stall cases.

PR #5171, opened 2023-03-16
Fix backwards log in op-node

### LINKED ISSUE #4892: Backwards log in op-node
Log message here is backwards, the config should be treated as the source of truth, not the node you're connected to. I'd also like if the log could be a little more clear about which chain ID is coming from the config and which is coming from the L2 node. Just looking at log there's no way to tell unless you know the structure of the log already. I had to dig through the source to figure this out.

https://github.com/ethereum-optimism/optimism/blob/1e780a398d000bea75954a62d9a67eefa5c652ab/op-node/rollup/types.go#L151

fgimenez:
**Description**

Fixes the order of checked chain ids in error message and aims to make clear where each of the compared ids comes from. 

**Metadata**
- Fixes #4892

### DIFF
--- a/op-node/rollup/types.go
+++ b/op-node/rollup/types.go
@@ -157,7 +157,7 @@ func (cfg *Config) CheckL2ChainID(ctx context.Context, client L2Client) error {
 		return err
 	}
 	if cfg.L2ChainID.Cmp(id) != 0 {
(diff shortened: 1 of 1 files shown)


smartcontracts:
Thank you!

PR #5176, opened 2023-03-16
fix(ci): reintroduce changesets patch

smartcontracts:
**Description**
Reintroduces a patch to the changesets cli that was required for private packages to be released and tagged properly.


tynes:
hmm how did this get removed?

smartcontracts:
> hmm how did this get removed?

Changesets was upgraded to a newer version with a PR in Jan. A second follow-up PR then removed the patch (presumably it was breaking things?). Definitely a process issue here.

PR #5179, opened 2023-03-16
op-node: generalize and improve alt-sync

protolambda:
**Description**

This PR iterates on top of the RPC-sync implementation to prepare for P2P-sync functionality as second form of alt-sync method, by encapsulating the RPC sync and updating its interface to make p2p sync fit in as well. Also fixes an off-by-one bug in the range to sync.

Changes:

- Add trust-RPC flag to `L2SyncEndpointConfig`. Previously it was hardcoded to true, yet also called blockhash verification. We can just allow the user to set it like they do with the L1 RPC.
- Change the sync-config check to be checked during the other config checks
- Clarify `GetUnsafeQueueGap` `start` is incl., `end` is excl., and fix off by one issue: we don't want to include the first block of the queue, but we do want to include the expected block number if queue is empty.
- Move RPC sync init code to its own function, separating it more from the l2 init
- Don't `Close` the RPC sync by reaching into attributes of the driver, but instead maintain the sync as separate lifetime thing in the node to close; there will be other sync methods.
- Put RPC sync requests behind an `AltSync` interface: the node itself implements it, and then routes it to the active & preferred sync methods that each also implement it, to allow more to be added.
  - The interface requests a full range of blocks, to avoid splitting the range into tinier tasks upfront, to allow the sync-method implementation to do batch / divide as desired.
  - The RPC `SyncClient` now consumes these range requests and does the fetching, instead of getting individual block numbers. Once new range information arrives, we now also drop previously scheduled information that may be stale.
- The driver now triggers sync sooner, when it's not changing the unsafe head as frequently as expected (thus not requesting syncing while it's deriving or has gossip coming in as expected). Also don't trigger sync when resetting the engine state (since we can't determine the range of L2 blocks accurately until the engine state is consistent again)

**Tests**

- Change RPC op-e2e test to disable batcher completely. When the sync runs around the same time as it syncs the data from L2, not all the blocks may be synced through the RPC. By disabling the batcher, we ensure it's really syncing through the RPC, and avoid potential flakiness.
- Change the same RPC test to use stringified block IDs, for more readable debug messages of what is and is not included in the received/published arrays
- Rename `L2SyncRPCConfig` to `PreparedL2SyncEndpoint` to match the other config utils of L1 and L2 endpoints that insert a prepared RPC client directly.

**Metadata**

Fix CLI-3643


op-node/rollup/derive/engine_queue.go
```diff
@@ -682,7 +682,8 @@ func (eq *EngineQueue) Reset(ctx context.Context, _ eth.L1BlockRef, _ eth.System
 	return io.EOF
 }
 
-// GetUnsafeQueueGap retrieves the current [start, end] range of the gap between the tip of the unsafe priority queue and the unsafe head.
+// GetUnsafeQueueGap retrieves the current [start, end) range (incl. start, excl. end)
```
ajsutton, line 685: the `incl. start, excl. end` description here is messing with my head. My understanding is that we're returning the start and end of the range of blocks we're missing and we'd need to request all of those blocks, including the start and end.

op-node/rollup/driver/driver.go
```diff
@@ -82,8 +81,18 @@ type Network interface {
 	PublishL2Payload(ctx context.Context, payload *eth.ExecutionPayload) error
 }
 
+type AltSync interface {
+	// RequestL2Range informs the sync source that the given range of L2 blocks is missing,
```
ajsutton, line 85: We probably should be clear about whether this range is inclusive or not as well.

op-node/rollup/driver/state.go
```diff
@@ -200,11 +199,12 @@ func (s *Driver) eventLoop() {
 		sequencerTimer.Reset(delay)
 	}
 
-	// Create a ticker to check if there is a gap in the engine queue every 15 seconds
-	// If there is, we send requests to the backup RPC to retrieve the missing payloads
-	// and add them to the unsafe queue.
-	altSyncTicker := time.NewTicker(15 * time.Second)
+	// Create a ticker to check if there is a gap in the engine queue, whenever
+	// If there is, we send requests to sync source to retrieve the missing payloads.
```
ajsutton, line 203:
```suggestion
	// Create a ticker to check if there is a gap in the engine queue. Whenever
	// there is, we send requests to sync source to retrieve the missing payloads.
```

op-node/rollup/driver/state.go
```diff
@@ -462,35 +471,29 @@ type hashAndErrorChannel struct {
...
 	genesisTimestamp := s.config.Genesis.L2Time
+	if wallClock < genesisTimestamp {
+		s.log.Debug("nothing to sync, did not reach genesis L2 time yet", "genesis", genesisTimestamp)
+		return nil
+	}
 	wallClockGenesisDiff := wallClock - genesisTimestamp
-	expectedL2Block := wallClockGenesisDiff / s.config.BlockTime
+	// Note: round down, we should not request blocks into the future.
+	blocksSinceGenesis := wallClockGenesisDiff / s.config.BlockTime
+	expectedL2Block := s.config.Genesis.L2.Number + blocksSinceGenesis
```
ajsutton, line 490: Just being paranoid here:
1. The genesis block number is the point at which bedrock kicked in right? Or was there a consistent block time before bedrock as well?
2. Do we have appropriate throttling to handle the case where the sequencer is offline and so unsafe head stalls for a period? We don't want to spam our sync source for requests for blocks that don't exist yet.

op-node/sources/sync_client.go
```diff
@@ -51,41 +65,87 @@ func NewSyncClient(receiver receivePayload, client client.RPC, log log.Logger, m
...
+				r.start += 1 // continue with next block
+			}
+			// Reschedule
+			if r.start < r.end {
+				select {
+				case s.requests <- r:
+				default:
+					// drop syncing job if we are too busy with sync jobs already.
+				}
+			}
```
ajsutton, line 148: Does this mean the request is rescheduled at the end of the channel? So we'd break the range to request up into blocks of 10, then request the first block from each, then the second from each etc. That works, but seems rather inefficient.  Any reason we don't just have a for loop in here that requests all the blocks in the batch one at a time?

ajsutton (commented):
Looks like a good improvement. Just a couple of questions that I could be pretty easily persuaded to not worry about. :)

op-node/rollup/driver/state.go
protolambda: > The genesis block number is the point at which bedrock kicked in right? 

Yes. It's a bit of a weird name, "bedrock anchor point" is maybe more accurate. But for most op-stack chains, it'll be the same as genesis anyway.

protolambda: > Do we have appropriate throttling to handle the case where the sequencer is offline and so unsafe head stalls for a period?

op-node/sources/sync_client.go
protolambda: The requests in the channel are meant to be a small range, see p2p sync PR which uses a similar mechanism but chops it up like that. Here we currently only really have one item in the channel (we drain on new information), don't do the splitting yet, and don't sync from multiple RPC sources (yet, would like to change the sync-client to a list of sync clients and have some parallel fetching + redundancy that way).

op-node/rollup/derive/engine_queue.go
protolambda: including the start, excluding the end. We could also redefine it as `start, count` if that helps, that's what L1 uses: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#beaconblocksbyrange

op-node/rollup/driver/driver.go
```diff
@@ -82,8 +81,18 @@ type Network interface {
...
+	// and should be retrieved from any available alternative syncing source.
```
protolambda, line 86:
```suggestion
	// and should be retrieved from any available alternative syncing source.
	// The start of the range is inclusive, the end is exclusive.
```

op-node/rollup/derive/engine_queue.go
ajsutton: oh right, yes. Hence the need for the + 1. I just tied myself in knots for some reason...

op-node/rollup/driver/state.go
ajsutton: Looking at this again, at most we'd be scheduling a batch of requests every time the ticker goes off (2 * blockTime) which is not unreasonable.

Maybe worth logging a ticket to investigate further - it could be improved by detecting when a block we request is not found and bailing out rather than still sending a request for the whole range. It's not a new problem being introduced by this PR anyway.

op-node/sources/sync_client.go
ajsutton: Up on line 105 we're splitting the single incoming request into batches of 10 before we push them into the `s.requests` channel:

```go
	for i := start; i < end; i += maxRangePerWorker {
 		r := syncRequest{start: i, end: i + maxRangePerWorker}
```

Unless I'm really misunderstanding the flow here...

op-node/flags/flags.go
```diff
@@ -226,6 +233,7 @@ var optionalFlags = append([]cli.Flag{
 	HeartbeatMonikerFlag,
 	HeartbeatURLFlag,
 	BackupL2UnsafeSyncRPC,
+	BackupL2UnsafeSyncRPCTrustRPC,
```
trianglesphere, line 236: This is conflicting with some flag changes. Will need a merge/rebase.

op-node/rollup/driver/state.go
```diff
@@ -200,11 +199,12 @@ func (s *Driver) eventLoop() {
 		sequencerTimer.Reset(delay)
 	}
 
-	// Create a ticker to check if there is a gap in the engine queue every 15 seconds
-	// If there is, we send requests to the backup RPC to retrieve the missing payloads
-	// and add them to the unsafe queue.
-	altSyncTicker := time.NewTicker(15 * time.Second)
+	// Create a ticker to check if there is a gap in the engine queue. Whenever
+	// there is, we send requests to sync source to retrieve the missing payloads.
+	syncCheckInterval := time.Duration(s.config.BlockTime) * time.Second * 2
```
trianglesphere, line 204: Why are we casting blocktime to a duration. Blocktime is in seconds, duration is in NS (https://pkg.go.dev/time#Duration)

protolambda: We're multiplying it to `time.Second`. We need to cast to duration to get it in the right type for the multiplication. I can also cast `time.Second` to uint64 first and then cast the result of multiplication, if you prefer.

trianglesphere: Sorry, I had a brain fart looking at the cast. That's correct.

op-node/rollup/driver/state.go
protolambda: Tracked in CLI-3683 now

op-node/service.go
```diff
@@ -56,6 +56,9 @@ func NewConfig(ctx *cli.Context, log log.Logger) (*node.Config, error) {
 	}
 
 	l2SyncEndpoint := NewL2SyncEndpointConfig(ctx)
+	if l2SyncEndpoint.L2NodeAddr == "" { // optional config
+		l2SyncEndpoint = nil
```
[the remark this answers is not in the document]
protolambda: Changed it to handle it differently, and tested locally on goerli now; `Setup()` now returns nil if it's unset, and so the configuration is always present, but during init it can decide if the sync endpoint is set in the op node.

protolambda:
Rebased and addressed PR review. Ready for final review

op-node/node/node.go
```diff
@@ -197,29 +201,28 @@ func (n *OpNode) initL2(ctx context.Context, cfg *Config, snapshotLog log.Logger
...
+	if cfg.L2Sync == nil { // RPC as alternative sync method is completely optional
+		return nil
+	}
+	rpcSyncClient, trustRPC, err := cfg.L2Sync.Setup(ctx, n.log)
+	if err != nil {
+		return fmt.Errorf("failed to setup L2 execution-engine RPC client for backup sync: %w", err)
 	}
 
-	n.l2Driver = driver.NewDriver(&cfg.Driver, &cfg.Rollup, n.l2Source, n.l1Source, syncClient, n, n.log, snapshotLog, n.metrics)
+	// The sync client's RPC is always trusted
```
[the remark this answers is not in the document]
ajsutton: This got resolved but the comment is still there. :)

op-node/rollup/driver/driver.go
ajsutton: This change didn't seem to actually get applied either.

op-node/rollup/driver/state.go
ajsutton: This didn't get changed either.  Either something went wrong with rebasing or my GitHub is broken...

op-node/sources/sync_client.go
ajsutton: This is still requesting the first block of each range in turn rather than requesting block sequentially.  Did we deliberately decide to stick with this or did a change get lost here too?

protolambda: Pro: reduce range by 1 item allows us to remove the request and update fully, avoiding consecutive requests for the same L2 block number to be in the channel (although not ruling out that an in-flight request isn't duplicate)

Con: the batches result in out of order requests if the channel is filled with multiple requests.

Adding a small cache to handle duplicate requests, and doing 1-by-1 probably works better, at least while we don't support by-range RPC requests. Will change it.

op-node/node/node.go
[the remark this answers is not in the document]
protolambda: Had it batched but forgot to commit it, my bad

op-node/rollup/driver/state.go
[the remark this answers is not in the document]
clabby: Was thinking about how this bug got merged into `develop` some more this morning- thankfully you caught it, but do you think it would be worth while to add variance to the l2 genesis in the e2e tests to catch bugs like this going forward? Are there any other areas off of the top of your head that would benefit from making the l2 genesis block non-zero in the tests?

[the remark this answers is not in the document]
protolambda: No other areas that I know of, most if not all other logic is relative based on head blocks rather than genesis, e.g. the sequencer planning uses the head.
I'll open a ticket for testing non-zero genesis better, but don't think it's as urgent since only alt-sync depends on it AFAIK.

op-node/sources/sync_client.go
protolambda: Opted to change it back to by-block-number request scheduling, but now with re-attempts + exponential backoff on the request. If it continues to fail, it will reschedule at the end of the requests queue. And once there's a new range, it'll still drop all previously scheduled work and schedule the new range.

ajsutton (approved):

protolambda:
Hmm, found an issue with the sync-changes: the payload queue is removing duplicates based on block number, not block hash: so on reorgs it is more likely to drop the new canonical blocks.

trianglesphere (commented):
Need to fix lint + tests, but after that should be good to go.

PR #5180, opened 2023-03-16
op-node: P2P req-resp alt sync method support

protolambda:
**Description**

Experimental feature: P2P Req-resp based L2 block syncing, to ensure the tip of the L2 chain is more reliably synced when data is slow to confirm in L1 or L2 gossip is missed.

This PR implements parallel block-fetching, a small quarantine for the results, and a promotion-process to get verify things by parent-hash against blocks with established trust (e.g. parent blocks of requested sync targets, or parent blocks of previously promoted blocks). This is effectively a reverse-header-sync implementation that can balances the work between peers.

Changes:
- Change request-range interface to use block refs, or zeroed end block ref for open range.
- Implements `p2p/sync.go`, including:
  - a sync client that implements the alt-sync interface to fetch blocks by range
  - a sync server that serves the req-resp protocol
  - rate-limiting across server and client
- Defines this sync req-resp protocol in the p2p-interface specs.
- Adds a feature-flag to enable it.

This functionality is off by default, since production-use relies on peer-scoring / banning to make sure the new p2p surface cannot be misused (aside from the rate-limits that are already there), and test-running this with a feature-flag on testnets will help harden it too.

**Tests**

- Op-e2e test that syncs a new verifier node that joins a bit later, while L1 batch-submission is disabled and gossip is missed by this node
- Server<>Client P2P Sync unit test
- Sync test with multiple peers and a temporary gap in available payloads

**Metadata**

Fix CLI-3644


protolambda:
@sebastianst no it does not. That issue also covers other LRU caches that we already have, which are still using v1. For the new things I introduce here I figured I should adopt v2 from the start. I added more details to the issue to point to the v1 usages.

op-node/p2p/sync.go
```diff
@@ -0,0 +1,756 @@
...
+		}
+		pr := peerRequest{num: num, complete: new(atomic.Bool)}
+
+		log.Debug("Scheduling P2P block request", "num", num)
+		// schedule number
+		select {
+		case s.peerRequests <- pr:
+			s.inFlight[num] = pr.complete
+		case <-ctx.Done():
+			log.Info("did not schedule full P2P sync range", "current", num, "err", ctx.Err())
```
trianglesphere, line 359: return here?

op-node/p2p/sync.go
```diff
@@ -0,0 +1,756 @@
...
+				}
+			} else {
+				log.Debug("completed p2p sync request", "num", pr.num)
+			}
+			took := time.Since(start)
+			// TODO(CLI-3732): update scores: depending on the speed of the result,
+			//  increase the p2p-sync part of the peer score
+			//  (don't allow the score to grow indefinitely only based on this factor though)
+
+			if s.metrics != nil {
```
trianglesphere, line 488: Should always run this code & use a noop metrics endpoint.

op-node/p2p/sync.go
```diff
@@ -0,0 +1,756 @@
...
+	log log.Logger
+
+	cfg *rollup.Config
+
+	metrics SyncClientMetrics
+
+	newStreamFn     newStreamFn
+	payloadByNumber protocol.ID
+
+	sync.Mutex
```
trianglesphere, line 181: what does the lock protect? Just the peers map?

op-node/p2p/sync.go
```diff
@@ -0,0 +1,756 @@
...
+			resultCode = 2
+		} else {
+			resultCode = 3
+		}
+		// try to write error code, so the other peer can understand the reason for failure.
+		_, _ = stream.Write([]byte{resultCode})
+	} else {
+		log.Debug("successfully served sync response", "req", req)
+	}
+	if srv.metrics != nil {
```
trianglesphere, line 669: ditto should not be checking for nil metrics.

op-node/p2p/sync.go
```diff
@@ -0,0 +1,756 @@
...
+
+	requests     chan rangeRequest
+	peerRequests chan peerRequest
+
+	results chan syncResult
+
+	resCtx    context.Context
+	resCancel context.CancelFunc
+
+	receivePayload receivePayloadFn
```
trianglesphere, line 208: Is my understanding that this is for when we have receivedPayloads?

op-node/p2p/sync.go
```diff
@@ -0,0 +1,756 @@
...
+		// TODO(CLI-3732): downscore peer for having provided us a bad block that never turned out to be canonical
+	} else {
+		s.log.Debug("evicting trusted payload from quarantine", "id", value.payload.ID(), "peer", value.peer)
+	}
+}
+
+func (s *SyncClient) tryPromote(h common.Hash) {
+	parentRes, ok := s.quarantine.Get(h)
+	if ok {
+		// Simply reschedule the result, to get it (and possibly its parents) out of quarantine without recursion
```
trianglesphere, line 383:
```suggestion
		// Simply reschedule the result, to get it (and possibly its parents) out of quarantine without recursion
		// s.results is buffered, but skip the promotion if the channel is full as it would cause a deadlock
```

It just took me a sec to figure out the control flow here.

op-node/p2p/sync.go
protolambda: nice catch

op-node/p2p/sync.go
protolambda: renamed it to `peersLock`, it's really just the `peers` map, it was covering more in earlier iteration.

op-node/p2p/sync.go
protolambda: yes indeed

op-node/p2p/node.go
```diff
@@ -146,6 +184,8 @@ func (n *NodeP2P) Close() error {
 		if err := n.host.Close(); err != nil {
 			result = multierror.Append(result, fmt.Errorf("failed to close p2p host cleanly: %w", err))
 		}
+		// TODO close sync loop
```
trianglesphere, line 187: I'm surprised semgrep doesn't seem to be flagging this. Does it need to be dealt with?

trianglesphere (approved):
lgtm. Nice work

PR #5181, opened 2023-03-17
Make MaxReorgSeqWindows pluggable from rollupConfig. Use the original…

kahuang:
**Description**

Making MaxReorgSeqWindows pluggable from rollup.Config so that new container images don't have to be built to change this parameter at runtime.
**Tests**

No tests, but happy to add if you let me know where.
**Invariants**

Default this change does nothing (as it uses the previous value), but allows the MaxReorgSeqWindows to be pluggable
**Additional context**

Because this parameter is currently not configurable, it means if we need to change it, we need to build a new container image, versus modifying a parameter in the rollup config. This change makes it easier to use the same container image version, and just change the parameter via the rollup config.
**Metadata**

### DIFF
--- a/.changeset/tricky-donkeys-lick.md
+++ b/.changeset/tricky-donkeys-lick.md
@@ -1,0 +1,9 @@
+---
+'@eth-optimism/chain-mon': patch
+'@eth-optimism/data-transport-layer': patch
+'@eth-optimism/fault-detector': patch
+'@eth-optimism/message-relayer': patch
+'@eth-optimism/replica-healthcheck': patch
+---
+
+Empty patch release to re-release packages that failed to be released by a bug in the release process.


op-node/rollup/sync/start.go
```diff
@@ -156,8 +156,14 @@ func FindL2Heads(ctx context.Context, cfg *rollup.Config, l1 L1Chain, l2 L2Chain
 		if (n.Number == result.Finalized.Number) && (n.Hash != result.Finalized.Hash) {
 			return nil, fmt.Errorf("%w: finalized %s, got: %s", ReorgFinalizedErr, result.Finalized, n)
 		}
+
+		// If the config didn't set MaxReorgSeqWindows, use the default.
+		maxReorgSeqWindows := uint64(MaxReorgSeqWindows)
```
tynes, line 161: I think we should renamve `MaxReorgSeqWindows` to `DefaultMaxReorgSeqWindows` and during config parsing, set the value on the config struct to the default if its 0 or missing in the config file

tynes: Then we can default to using the config struct in this codepath no matter what

kahuang: Yes can do. Is there a single path for this? Just the json load from file method? I was worried I wouldn't cover all of them (if there are multiple)

op-node/rollup/types.go
```diff
@@ -62,6 +62,9 @@ type Config struct {
 	MaxSequencerDrift uint64 `json:"max_sequencer_drift"`
 	// Number of epochs (L1 blocks) per sequencing window, including the epoch L1 origin block itself
 	SeqWindowSize uint64 `json:"seq_window_size"`
+	// Number of sequence windows back that we can re-org. If a re-org needs to occur that's larger than this
+	// parameter, then the re-org will not happen, which can halt the chain.
+	MaxReorgSeqWindows uint64 `json:"max_reorg_seq_windows"`
```
trianglesphere, line 67: TBH I think we should make it a hidden flag rather than in the rollup config. We plan to bake in rollup configs so users can do stuff like `--network=goerli`.

Something like https://github.com/ethereum-optimism/optimism/blob/2351ed462355cd516ddb8176d1ffdae607d1b7ed/op-node/flags/p2p_flags.go#L131-L138

PR #5182, opened 2023-03-17
specs: Update spec channel handling to match the current implementation

ajsutton:
**Description**

Updates the derivation spec to match the current implementation's channel handling rules - particularly that channel IDs are not considered known forever.

 - Dropping frames due to timed out channels or duplication is only done for channels that have not yet been pruned from the channel bank. 
 - Timed out channels are only pruned from the channel-bank if they are prior to a readable channel (ie not all timed out channels are necessarily pruned)
 - Explicitly state that channels are removed from the channel-bank when they are read.

Related to https://github.com/ethereum-optimism/optimism/pull/5107 This won't fix the stall but does bring the spec and actual implementation into agreement.

**Metadata**


s7v7nislands:
hi, how to access the https://linear.app/optimism/issue/CLI-3600/review-spec-duplicate-channel-rules ? it seems not public.

ajsutton:
Yeah it's not public but that ticket actually just points back to #5107 which is where this actually all started.

sebastianst (commented):
LGTM, just 2 small nits regarding consistency.

We could also add a sentence after the "frame insertion conditions" like 
"In particular, a channel's ID can be reused after the channel got removed from the channel bank."

ajsutton:
The bedrock-markdown failure is because eip4844.com is no longer available.

sebastianst:
> The bedrock-markdown failure is because eip4844.com is no longer available.

Interesting, just checked the URL and it works. Rerunning workflow.

sebastianst (approved):

ajsutton:
> Interesting, just checked the URL and it works. Rerunning workflow.

Someone must have gotten around to renewing the domain name. It was a domain placeholder page when I tried it. :)

sebastianst:
@snario fixed it 🙏🏻

PR #5188, opened 2023-03-17
op-node: make find-sync-start more resilient against errors by improving cache usage

protolambda:
**Description**

This improves find-sync-start by improving L1 and L2 rpc caches to include full ranges of block-ref info, and adjusting the code to hit L1 block ref cache when possible.

So when it does hit an error, it'll recover more quickly using the cached data, avoiding a cascade of additional already executed RPC requests following a rate limit hit.
See the notion doc regarding problem & other long term ideas of improving this.

**Tests**

No change in behavior other than utilizing the L1 cache when possible. Existing find-sync start and op-e2e tests cover this new code.

**Metadata**

Fix CLI-3650

### DIFF
--- a/op-node/sources/l1_client.go
+++ b/op-node/sources/l1_client.go
@@ -24,6 +24,7 @@ type L1ClientConfig struct {
 func L1ClientDefaultConfig(config *rollup.Config, trustRPC bool, kind RPCProviderKind) *L1ClientConfig {
 	// Cache 3/2 worth of sequencing window of receipts and txs
 	span := int(config.SeqWindowSize) * 3 / 2
+	fullSpan := span
 	if span > 1000 { // sanity cap. If a large sequencing window is configured, do not make the cache too large
 		span = 1000
 	}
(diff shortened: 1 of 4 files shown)


op-node/sources/l1_client.go
```diff
@@ -24,6 +24,7 @@ type L1ClientConfig struct {
 func L1ClientDefaultConfig(config *rollup.Config, trustRPC bool, kind RPCProviderKind) *L1ClientConfig {
 	// Cache 3/2 worth of sequencing window of receipts and txs
 	span := int(config.SeqWindowSize) * 3 / 2
+	fullSpan := span
 	if span > 1000 { // sanity cap. If a large sequencing window is configured, do not make the cache too large
```
trianglesphere, line 28: what is the memory usage for the other components - this might be worth expanding for everything or limiting the span to be much larger.

trianglesphere (approved):
Sync start fix LGTM

tynes (approved):

op-node/sources/l1_client.go
protolambda: Opened CLI-3680 to track, I think it's a separate thing that we can improve, but it is not as urgent.

PR #5189, opened 2023-03-17
op-proposer: add l2 block height metrics

shrimalmadhur:
**Description**

- Adding a new metric to `op-proposer` which emits the latest proposed block height 

**Tests**

**Invariants**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


op-service/metrics/block.go
```diff
@@ -0,0 +1,17 @@
+package metrics
+
+import (
+	"math/big"
+
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
```
0x00101010, line 7: remove extra blank line

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -113,10 +114,11 @@ func Main(version string, cliCtx *cli.Context) error {
...
 type L2OutputSubmitter struct {
-	txMgr txmgr.TxManager
-	wg    sync.WaitGroup
-	done  chan struct{}
-	log   log.Logger
+	txMgr           txmgr.TxManager
+	wg              sync.WaitGroup
+	done            chan struct{}
+	log             log.Logger
+	metricsRegistry *prometheus.Registry
```
0x00101010, line 121: `metricsResigry` => `mr`

op-service/metrics/block.go
```diff
@@ -0,0 +1,17 @@
+package metrics
```
0x00101010, line 1: nit: `block.go` => `proposer.go`, this feels like a more proposer related concept

op-service/metrics/proposer.go
trianglesphere, line 1: Hey, thanks for implementing this. We've also started on metrics for the batcher, and there we are following the pattern of using `op-service` for generic metrics factories/flags/running the server & then putting the metrics inside the op-proposer codebase.

See https://github.com/ethereum-optimism/optimism/pull/5167/files

shrimalmadhur: oh I see. that makes sense. does that means we will have a separate metrics library under `op-proposer` dedicated to that and not rely on `op-service` ?

trianglesphere: The metrics library in `op-proposer` still depends on `op-service` for things like starting the metrics server / metrics flags / documentation / creating the registry.

In `op-proposer` it will have typed functions to record events in it's own metrics package/library.

shrimalmadhur:
> Awesome to see this coming! As @trianglesphere remarked, please follow the pattern in #5167 and create a `op-proposer/metrics` package, with a `noop` implementation, etc.
> 
> Also please reuse the `RefMetrics` from `op-service/metrics`, like in the batcher metrics.

Thanks @sebastianst and @trianglesphere for the guidance. I will make those changes similar to other packages and send out another revision. :)

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -375,6 +380,10 @@ func (l *L2OutputSubmitter) SendTransaction(ctx context.Context, tx *types.Trans
 		return err
 	}
 
+	if l.metricsEnabled {
+		// Emit the proposed block Number
```
shrimalmadhur, line 384: @trianglesphere @sebastianst I am a bit stuck in how we propose the L2 block here. I am trying to use ref_metrics way of emitting L2 block - https://github.com/ethereum-optimism/optimism/pull/5189/files#diff-0ad98c4d91561ec2bf2386d9d3a5d7e766ba866d749163568e04e6e22bcc2f32R98 but I don't have access to l2ref the way we have in batcher [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-batcher/batcher/driver.go#L218) 

Is there a way to get it or should I introduce a custom method just to emit the block number?

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -113,10 +113,12 @@ func Main(version string, cliCtx *cli.Context) error {
...
 type L2OutputSubmitter struct {
-	txMgr txmgr.TxManager
-	wg    sync.WaitGroup
-	done  chan struct{}
-	log   log.Logger
+	txMgr          txmgr.TxManager
+	wg             sync.WaitGroup
+	done           chan struct{}
+	log            log.Logger
+	metricsEnabled bool
```
sebastianst, line 120: you probably don't need this if you pass in the noop metrics.

op-proposer/proposer/l2_output_submitter.go
sebastianst: The main loop that calls `SendTransaction` has the `L2BlockRef` in `output`. Could either pass the `L2BlockRef` here or you call the `RecordL2BlocksProposed` from the main loop after successfully sending.

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -375,6 +380,10 @@ func (l *L2OutputSubmitter) SendTransaction(ctx context.Context, tx *types.Trans
 		return err
 	}
 
+	if l.metricsEnabled {
```
sebastianst, line 383: not needed with noop metrics

shrimalmadhur: oh I see. that makes sense.

op-proposer/proposer/l2_output_submitter.go
shrimalmadhur: oh ya. I thought of using `rollupClient` to fetch it but I think the 2nd suggestion makes more sense to include it from the caller itself. I will update.

shrimalmadhur: removed this as metrics is being called from caller

shrimalmadhur:
the `main` and `bedrock-markdown` are failing. But I don't see any PR related concern. 🤔

shrimalmadhur:
> the `main` and `bedrock-markdown` are failing. But I don't see any PR related concern. 🤔

Hey @sebastianst @trianglesphere, how do I get these two failing tests to pass? want to make sure I make this PR merge ready.

trianglesphere:
bedrock-markdown test is failing because we do a link check & there is a dead link. That is why main is failing. We can ignore those tests. The rest of the tests which actually tests these changes are passing.

op-proposer/proposer/l2_output_submitter.go
```diff
@@ -13,6 +13,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/ethereum-optimism/optimism/op-proposer/metrics"
```
trianglesphere, line 16: import should be with the rest of the local imports

trianglesphere (approved):
I've confirmed that these work in the local devnet. Really nice work with this.

sebastianst (approved):
Looks great now ✨ just that one import should be reordered.

op-proposer/proposer/l2_output_submitter.go
shrimalmadhur: just fixed this! thanks for catching this.

shrimalmadhur:
> Looks great now ✨ just that one import should be reordered.

perfect. just pushed both changes.

PR #5191, opened 2023-03-17
feat(op-service): Move Tx Creation to the txmgr

refcell:
**Description**

Fixes CLI-3548

Moves transaction creation to the `txmgr` in the `op-service`. Rational is that there are two primary places where txs are sent - the `op-proposer` and `op-batcher`. In each, we currently have to role our own transaction construction since the `op-service` doesn't support creation. This is a generalized solution to tx creation in the `txmgr`.


op-service/txmgr/txmgr.go
```diff
@@ -101,6 +114,86 @@ type SimpleTxManager struct {
...
+		ChainID:   candidate.ChainID,
+		Nonce:     nonce,
+		To:        &candidate.Recipient,
+		GasTipCap: gasTipCap,
+		GasFeeCap: gasFeeCap,
+		Data:      candidate.TxData,
+	}
+	m.l.Info("creating tx", "to", rawTx.To, "from", candidate.From)
+
+	gas, err := core.IntrinsicGas(rawTx.Data, nil, false, true, true, false)
```
tynes, line 186: This isn't a safe way to do gas estimation unless there is no code at `to`

op-service/txmgr/txmgr.go
```diff
@@ -101,6 +114,86 @@ type SimpleTxManager struct {
...
+// It queries L1 for the current fee market conditions as well as for the nonce.
+// NOTE: This method SHOULD NOT publish the resulting transaction.
+func (m *SimpleTxManager) CraftTx(ctx context.Context, candidate TxCandidate) (*types.Transaction, error) {
+	gasTipCap, gasFeeCap, err := m.calcGasTipAndFeeCap(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	childCtx, cancel := context.WithTimeout(ctx, m.Config.NetworkTimeout)
+	nonce, err := m.backend.NonceAt(childCtx, candidate.From, nil)
```
tynes, line 170: passing `nil` as the blocktag, does that use pending or latest?

refcell: nonce from the latest known block i believe

op-batcher/batcher/config.go
```diff
@@ -31,6 +31,15 @@ type Config struct {
 	From         common.Address
 
 	TxManagerConfig txmgr.Config
+	NetworkTimeout  time.Duration
+
+	// TxManagerTimeout is the maximum amount of time
+	// the driver should wait for the [txmgr] to send a transaction.
+	TxManagerTimeout time.Duration
```
trianglesphere, line 38: This is a good thing to parameterize, but a little out of scope for this PR which should be focused on where the transaction is created. Also given that we have the op-propser which uses the txmanager with the same flags, we will want to create a common set of flags inside the txmanager.

op-service/txmgr/txmgr.go
```diff
@@ -69,6 +80,9 @@ type TxManager interface {
 	//
 	// NOTE: Send should be called by AT MOST one caller at a time.
 	Send(ctx context.Context, tx *types.Transaction) (*types.Receipt, error)
+
+	// CraftTx is used to craft a transaction using a [TxCandidate].
+	CraftTx(ctx context.Context, candidate TxCandidate) (*types.Transaction, error)
```
trianglesphere, line 85: I guess we need `CraftTx` because we're implementing the proposer changes separately, but the goal is that `Send` takes the `TxCandidate` instead of the `Tx`

op-service/txmgr/txmgr.go
```diff
@@ -101,6 +121,102 @@ type SimpleTxManager struct {
 	l       log.Logger
 }
 
+// TxCandidate is a transaction candidate that can be submitted to ask the
+// [TxManager] to construct a transaction with gas price bounds.
+type TxCandidate struct {
+	// TxData is the transaction data to be used in the constructed tx.
+	TxData []byte
+	// Recipient is the recipient (or `to`) of the constructed tx.
+	Recipient common.Address
```
trianglesphere, line 130: `to` is well known + standard

op-service/txmgr/txmgr.go
```diff
@@ -101,6 +121,102 @@ type SimpleTxManager struct {
...
+	// GasLimit is the gas limit to be used in the constructed tx.
+	GasLimit uint64
+	// From is the sender (or `from`) of the constructed tx.
+	From common.Address
+	/// ChainID is the chain ID to be used in the constructed tx.
+	ChainID *big.Int
```
trianglesphere, line 136: Should probably been in the `TxManager` constructor (or fetched from L1 by the txmgr) rather than in the `TxCandidate`

trianglesphere (commented):
The file `op-batcher/batcher/txmgr.go` should also be removed.

op-batcher/batcher/config.go
refcell: I can change this in the proposer but it seems like _that_ would be out of the scope of this pr. lmk

op-service/txmgr/txmgr.go
```diff
@@ -196,8 +325,15 @@ func NewSimpleTxManager(name string, l log.Logger, cfg Config, backend ETHBacken
 	if cfg.NumConfirmations == 0 {
 		panic("txmgr: NumConfirmations cannot be zero")
 	}
+	if cfg.NetworkTimeout == 0 {
+		cfg.NetworkTimeout = 2 * time.Second
+	}
+
+	// On error, ignore. We will try to get the chainID again when used.
+	chainID, _ := backend.ChainID(context.Background())
```
sebastianst, line 333: Why do we do that? Shouldn't we just let `NewSimpleTxManager` return an `error` here? If we cannot query while setting up the txmgr, something is wrong.

op-service/txmgr/txmgr.go
```diff
@@ -196,8 +325,15 @@ func NewSimpleTxManager(name string, l log.Logger, cfg Config, backend ETHBacken
 	if cfg.NumConfirmations == 0 {
 		panic("txmgr: NumConfirmations cannot be zero")
```
sebastianst, line 326: maybe return `error` instead?

op-service/txmgr/txmgr.go
```diff
@@ -89,18 +103,133 @@ type ETHBackend interface {
...
+func (m *SimpleTxManager) CraftTx(ctx context.Context, candidate TxCandidate) (*types.Transaction, error) {
+	gasTipCap, gasFeeCap, err := m.calcGasTipAndFeeCap(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Fetch the sender's nonce from the latest known block (nil `blockNumber`)
+	childCtx, cancel := context.WithTimeout(ctx, m.Config.NetworkTimeout)
+	nonce, err := m.backend.NonceAt(childCtx, candidate.From, nil)
+	cancel()
```
sebastianst, line 182: We're not in a loop, so you can safely just do the usual `defer cancel()`. 
```suggestion
	defer cancel()
	nonce, err := m.backend.NonceAt(childCtx, candidate.From, nil)
```
Better practice because if `NonceAt` were to panic, a `defer` would still free up resources.

sebastianst (changes requested):
Looking good, just some minor comments.

Would also like to see `CraftTx` be done inside `Send` and removed from the interface, but not sure how much this would blow up the pr.

op-service/txmgr/txmgr.go
trianglesphere: I'm fine leaving this as is for now. We have lots of config things to cleanup

op-batcher/batcher/driver_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum-optimism/optimism/op-node/testlog"
+	"github.com/ethereum-optimism/optimism/op-service/txmgr"
+	"github.com/ethereum-optimism/optimism/op-service/txmgr/mocks"
+)
+
+// TestSendTransaction tests that the driver can send a transaction
+// through the txmgr.
+func TestSendTransaction(t *testing.T) {
```
trianglesphere, line 22: I'm not sure what this test gain us. We check the ability of "can send" with types + in e2e tests. There are two other types of bug which would cause a failure to send: A misconfigured transaction manger which this test will not catch & not actually calling sendTransaction in the batcher. This does ensure that we call `sendTransaction`, but it's a very obvious function that needs to be called with many other layers of testing.

op-service/txmgr/txmgr.go
sebastianst: if it were the only remark in the ctor, yes, but there's also the chain id fetching, which should return an error here in the ctor instead of checking for a set chain id in every send call.

op-service/txmgr/txmgr.go
refcell: Got it - op-proposer changes (along with the config variables you noted above are in the linear ticket https://linear.app/optimism/issue/CLI-3685/featop-proposer-migrate-the-proposer-to-txmgr-tx-creation so we track). CLI-3685 (the op-proposer changes) will be the next pr I focus on tomorrow after this.

op-batcher/batcher/config.go
```diff
@@ -31,6 +31,15 @@ type Config struct {
...
+
+	// OfflineGasEstimation specifies whether the batcher should calculate
+	// gas estimations offline using the [core.IntrinsicGas] function.
+	OfflineGasEstimation bool
```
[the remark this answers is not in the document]
refcell: Removed.

op-batcher/batcher/driver_test.go
refcell: IMO - since I touched an external function, it'd be strange not to have any tests. Although it _may_ be tested elsewhere, there should be a _local_ test - and although yes it's just testing the happy path, that's the bare minimum.

op-service/txmgr/txmgr.go
refcell: Well this changes the interface of the `NewSimpleTxManager` which I _really_ would like to not touch.

op-service/txmgr/txmgr.go
refcell: Soooo change the `NewSimpleTxManager` method to return an error??

sebastianst: It's only called in three places (1 test, batcher setup and proposer setup), so this wouldn't be too bad and would match the pattern we have in many ctors where we do some networking setup. Then we would also need to pass a `ctx`, which the call sites have though. 

But I looked at it again and have an even cleaner suggestion: add the `ChainID` to the `txmgr.Config`. It has multiple advantages
* Call sites know the id already or can query the id.
* Can remove the `chainID` from the `SimpleTxManager` state.
* Can remove `ChainID` from `ETHBackend`.
* Signature of `NewSimpleTxManager` doesn't need to be changed.

It's cleaner because the txmgr shouldn't need to query it itself and it can just require it as a config parameter.

In any case, setting up the chain id _should not_ be part of `CraftTx`. It belongs in setup code. It's a workaround to call it there just to not do other changes.

And as a general remark, it's totally ok in my opinion to change a function signature if that leads to a cleaner solution.

sebastianst (approved):

trianglesphere (approved):

PR #5193, opened 2023-03-17
specs: add permissionless L2OutputOracle

tynes:
**Description**

Small refactor of the specs to begin mention of the permissionless output proposals. Updates the specs around the `L2OutputOracle` to differentiate between v1 and v2

Closes CLI-3481

### DIFF
--- a/specs/proposals.md
+++ b/specs/proposals.md
@@ -10,6 +10,8 @@
 **Table of Contents**
 
 - [Proposing L2 Output Commitments](#proposing-l2-output-commitments)
+  - [L2OutputOracle v1.0.0](#l2outputoracle-v100)
+  - [L2OutputOracle v2.0.0](#l2outputoracle-v200)
 - [L2 Output Commitment Construction](#l2-output-commitment-construction)
 - [L2 Output Oracle Smart Contract](#l2-output-oracle-smart-contract)
   - [Configuration](#configuration)
@@ -21,8 +23,8 @@
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 After processing one or more blocks the outputs will need to be synchronized with L1 for trustless execution of
-L2-to-L1 messaging, such as withdrawals. Outputs are hashed in a tree-structured form which minimizes the cost of
-proving any piece of data captured by the outputs.
+L2-to-L1 messaging, such as withdrawals.
+The output proposals act as the bridge's view into the L2 state.
 Proposers submit the output roots to L1 and can be contested with a fault proof,
 with a bond at stake if the proof is wrong.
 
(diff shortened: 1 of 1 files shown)


specs/proposals.md
```diff
@@ -35,24 +37,40 @@ are part of later specification milestones.
...
+### L2OutputOracle v1.0.0
 
-The proposer may also delete multiple output roots by calling the `deleteL2Outputs()` function and specifying the
+The submission of output proposals is permissioned to a single account. It is
+expected that this account continues to submit output proposals over time to
+ensure that user withdrawals do not halt.
+
+The [L2 output proposer](../op-proposer) is expected to submit output roots on a
+deterministic interval based on the configured `SUBMISSION_INTERVAL` in the
+`L2OutputOracle`. The larger the `SUBMISSION_INTERVAL`, the less often L1
```
ajsutton, line 55: With this change we lost any mention of:
*  the output hash being tree structured (probably ok - it wasn't specified in any detail here anyway)
* any mention of proposals being delayed if there's no new finalised output
* the specific requirement of the timestamp block number having to correspond to the `startingBlockNumber` plus the next multiple of the `SUBMISSION_NITERVAL` value. It's somewhat implied but less clear.

The original spec wasn't really precise about how things actually worked either so I'm not against this change, just wanted to check that we're happy for it to be a little less specific about the current setup.

ajsutton (commented):
Looks good generally, just checking what level of detail we want to keep.

specs/proposals.md
tynes: The output hash being tree structured is wrong? Its just a flat hash over the values, which is why I removed that. The proposer can be configured to submit only safe values or unsafe values which is implementation specific, but that could be mentioned. We no longer pass in the timestamp, but we could specify what must be passed in

ajsutton (approved):
LGTM. Thanks for clarifying.

tynes:
I've added in some additional information to this PR

specs/proposals.md
```diff
@@ -35,24 +37,62 @@ are part of later specification milestones.
...
+
+The honest `op-proposer` algorithm assumes a connection to the `L2OutputOracle` contract to know
+the L2 block number that corresponds to the next output proposal that must be submitted. It also
+assumes a connection to an `op-node` to be able to query the `optimism_syncStatus` RPC endpoint.
+
+```python
+import time
+
+while True:
+    next_checkpoint_block = L2OutputOracle.nextBlockNumber()
```
tynes, line 70: I don't love mixing snake and camelcase but did so because it was a contract interaction, wondering if i should just make everything camel case

specs/proposals.md
```diff
@@ -35,24 +37,62 @@ are part of later specification milestones.
...
+interval based on the configured `SUBMISSION_INTERVAL` in the `L2OutputOracle`. The larger
+the `SUBMISSION_INTERVAL`, the less often L1 transactions need to be sent to the `L2OutputOracle`
+contract, but L2 users will need to wait a bit longer for an output root to be included in L1 that
+includes their intention to withdrawal from the system.
+
+The honest `op-proposer` algorithm assumes a connection to the `L2OutputOracle` contract to know
+the L2 block number that corresponds to the next output proposal that must be submitted. It also
+assumes a connection to an `op-node` to be able to query the `optimism_syncStatus` RPC endpoint.
+
+```python
```
trianglesphere, line 66: I guess we have already referenced the `optimism_outputAtBlock` API, but IMO the pseudocode is an implementation detail (same with the RPC mention)

tynes: I'm not sure exactly what you mean here, are you suggesting we delete the pseudocode? I do think that the pseudocode is useful personally as its the definition of the honest proposer. Describing it in code is much more precise than using human language, I'd consider things on top of this definition to be implementation details

refcell: Generally, I think it'd be easier to maintain these specs without the specific pseudocode implementation and rpc reference. But, I think it's acceptable here since it is specifically pinned to a version of the `L2OutputOracle` (v1.0.0).

specs/proposals.md
```diff
@@ -35,24 +37,62 @@ are part of later specification milestones.
...
+
+The submission of output proposals is permissionless and there is no interval at which output
+proposals must be submitted at. It is expected that users will "just in time" propose an output
+proposal to facilitate their own withdrawal. A bond must be placed with an output proposal to
+disincentivize the proposal of malicious outputs. If it can be proven that the output is malicious,
+either via fault proof or by an attestation proof, then the bond can be slashed and used as a
+payment to the users who paid for gas to remove the malicious output.
+
+The `op-proposer` can still be used to submit output proposals. A naive implementation of the
+`op-proposer` will still submit output proposals on an interval. An more ideal implementation
```
refcell, line 91:
```suggestion
`op-proposer` will still submit output proposals on an interval. A more ideal implementation
```

refcell (changes requested):
Small nits

tynes:
@refcell I merged in your suggestions, ty

specs/proposals.md
tynes: I did specifically pin the algo to v1 for this purpose

refcell (approved):
Look good to me.

specs/proposals.md
refcell: Are we good to resolve this and merge?

specs/proposals.md
```diff
@@ -35,24 +37,62 @@ are part of later specification milestones.
...
+disincentivize the proposal of malicious outputs. If it can be proven that the output is malicious,
+either via fault proof or by an attestation proof, then the bond can be slashed and used as a
+payment to the users who paid for gas to remove the malicious output.
+
+The `op-proposer` can still be used to submit output proposals. A naive implementation of the
+`op-proposer` will still submit output proposals on an interval. A more ideal implementation
+will use heuristics such as time of last submission or number of pending withdrawals that have
+yet to be included in an output proposal.
+
+Version `v2.0.0` includes breaking changes to the `L2OutputOracle` ABI.
```
maurelian, line 95: This section is pretty vague, but seems 'good enough' for now.

maurelian (approved):

specs/proposals.md
```diff
@@ -163,7 +203,7 @@ will not match that of the block with that number and the call will revert.
...
 ### Constants
 
-| Name                  | Value | Unit    |
-| --------------------- | ----- | ------- |
-| `SUBMISSION_INTERVAL` | TBD   | blocks  |
-| `L2_BLOCK_TIME`       | `2`   | seconds |
+| Name                  | Value   | Unit    |
+| --------------------- | ------- | ------- |
+| `SUBMISSION_INTERVAL` | `120`   | blocks  |
+| `L2_BLOCK_TIME`       | `2`     | seconds |
```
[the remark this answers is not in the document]
tynes: I do think its useful to have "recommended values", the eth2 spec does this

specs/proposals.md
tynes: You are right, will follow up with additions as we work on it

PR #5195, opened 2023-03-18
op-proposer: log L2OutputOracle version + address

tynes:
**Description**

Logs the L2OutputOracle version and address after building a client to call it. This is generally useful information to have and future implementations of the `op-proposer` will need to function differently depending on the version to keep the software compatible with both permissioned and permissionless output proposal submission.


ajsutton (approved):
LGTM

PR #5196, opened 2023-03-19
specs: Update derivation spec to process the first ready frame

ajsutton:
**Description**
Updates the derivation spec channel handling so that it will skip over incomplete and not timed out channels to find a frame that is ready to read. As frames are ingested one at a time and only when there are no channels ready to read, this results in  channels being read in the order that they become ready, instead of the order of their first frame. In turn, this prevents unclosed frames from delaying the progression of the safe head until they time out if valid data is submitted in another channel.

Other changes to make the spec match the current implementation:
* Dropping frames due to timed out channels or duplication is only done for channels that have not yet been pruned from the channel bank.
* Timed out channels are only pruned from the channel-bank if they are prior to a readable channel.
* Explicitly state that channels are removed from the channel-bank when they are read.

**Additional context**

This is technically a breaking consensus change. Suppose there are two valid channels A and B containing attributes for the same L2 block numbers where the first frame of A is included on L1 before the first frame of B, but the last frame of A is included on L1 _after_ the last frame of B. In this case the old spec would have processed A and ignored B, but the new code will process B and ignore A.

An analysis of channels submitted for Optimism Goerli confirmed there are no complete channels that would be processed in a different order because of this change.

**Metadata**
- See also #5107

### DIFF
--- a/specs/derivation.md
+++ b/specs/derivation.md
@@ -515,14 +515,15 @@ New frames for timed-out channels are dropped instead of buffered.
 
 #### Reading
 
-The channel-bank can only output data from the first opened channel.
+Upon reading, process each channel in the order they were added to the channel bank. For each frame:
 
-Upon reading, first all timed-out channels are dropped.
+- if the channel is timed out, remove it from the channel-bank
+- otherwise, if the channel is ready, then read it, remove it from the channel bank and stop processing further channels
 
-After pruning timed-out channels, the first remaining channel, if any, is read if it is ready:
+A channel is ready if:
 
-- The channel must be closed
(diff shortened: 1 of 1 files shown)


tynes:
Not sure how i feel about a soft consensus critical change, we would need to get everybody to upgrade asap, and i think the only way to trigger a chain split would be through the batcher acting strangely

ajsutton:
So given the timeline to the bedrock release and that this only causes issues when the batcher has an issue and doesn't close the channel, I think the plan will be to do this as a hard fork later. Will just update the spec to match the current implementation for initial release.

So will close this PR as it will need to be written differently to be specified as a hard fork.

PR #5197, opened 2023-03-19
feat(ctb): Add negative test for token compliance in the `L2ERC721Bridge`

clabby:
# Overview

Adds a quick test to the `L2ERC721Bridge` that asserts that the `finalizeBridgeERC721` function should revert if the local token does not claim to be compliant with the `IOptimismMintableERC721` interface.

**Tests**

* Added: `test_finalizeBridgeERC721_interfaceNotCompliant_reverts`

**Invariants**

* If the `_localToken` passed to `finalizeBridgeERC721` does not claim to be compliant with the `IOptimismMintableERC721` interface, the function should revert.

**Additional context**

Part of the token compatibility retro from Sherlock.


packages/contracts-bedrock/contracts/test/L2ERC721Bridge.t.sol
```diff
@@ -349,3 +376,30 @@ contract L2ERC721Bridge_Test is Messenger_Initializer {
         );
     }
 }
+
+/// @dev A non-compliant ERC721 token that does not implement the full ERC721 interface.
```
tynes, line 380: this kind of devdoc uses less lines but we have standardized on the other type  `/* ... */`

tynes (approved):
Looks good to me besides the one comment

PR #5198, opened 2023-03-20
op-chain-ops: Check and mutate DB in parallel during OVM_ETH migration

mslipper:
Updates the precheck script to check and mutate the DB during the OVM_ETH migration. This saves a significant amount of time during the migration, since we avoid performing a full storage iteration over the OVM_ETH state followed by a single-threaded iteration over the addresses to migrate. Now, both run in parallel which makes the runtime of the OVM_ETH migration bounded by how quickly we can update the trie.

Also adds some logging in the final check step for withdrawals.

Fixes CLI-3662
Fixes CLI-3676


ajsutton (commented):
Looks ok to me but will defer to @tynes for approval as I'm not familiar with the legacy content we're migrating and am still fairly new to go concurrency.

mslipper:
Resolved nits. Also ran this end-to-end on a migration disk and discovered that I forgot to set the total supply to zero - this has been fixed with the latest push.

op-chain-ops/ether/migrate.go
```diff
@@ -27,73 +45,337 @@ var (
...
 
-	// Fin.
 	return nil
 }
+
+// PartitionKeyspace divides the key space into partitions by dividing the maximum keyspace
+// by count then multiplying by i. This will leave some slots left over, which we handle below. It
+// returns the start and end keys for the partition as a common.Hash. Note that the returned range
+// of keys is inclusive, i.e., [start, end] NOT [start, end).
+func PartitionKeyspace(i int, count int) (common.Hash, common.Hash) {
```
tynes, line 361: `i` should never be negative but I don't think thats possible based on how this function is called

op-chain-ops/ether/migrate.go
```diff
@@ -27,73 +45,337 @@ var (
...
+	if i > count-1 {
+		panic("i must be less than count - 1")
+	}
+
+	// Divide the key space into partitions by dividing the key space by the number
+	// of jobs. This will leave some slots left over, which we handle below.
```
tynes, line 367: Slightly worried about precision loss with this division, since its not a decimal i believe it'll round down like python `//`

op-chain-ops/ether/migrate.go
```diff
@@ -27,73 +45,337 @@ var (
...
+			}
+
+			// Skip if the value is empty.
+			rawValue := it.Value
+			if len(rawValue) == 0 {
+				continue
+			}
+
+			// Get the preimage.
+			key := common.BytesToHash(st.GetKey(it.Key))
```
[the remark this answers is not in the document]
mslipper: Checked.

op-chain-ops/ether/migrate.go
mslipper: Yes, I want it to round down. The precision loss is handled by the if check - any lost precision is made up for in the last partition.

op-chain-ops/ether/migrate.go
mslipper: Added check.

op-chain-ops/ether/migrate.go
```diff
@@ -27,73 +45,348 @@ var (
...
+	// Add the old SequencerEntrypoint because someone sent it ETH a long time ago and it has a
+	// balance but none of our instrumentation could easily find it. Special case.
+	entrySK := CalcOVMETHStorageKey(sequencerEntrypointAddr)
+	slotsAddrs[entrySK] = sequencerEntrypointAddr
+	slotsInp[entrySK] = BalanceSlot
+
+	// WaitGroup to wait on each iteration job to finish.
+	var wg sync.WaitGroup
+	// Channel to receive storage slot keys and values from each iteration job.
+	outCh := make(chan accountData)
```
trianglesphere, line 109: An unbuffered channel is fine here b/c we aren't able to deadlock in this code, but I'd lean to a buffered channel here b/c we don't need the synchronization. IDK if it will have any performance impact though.

mslipper: We do need the synchronization. There's nothing explicitly draining `outCh`, so if we make this channel buffered we'll have to drain it after waiting for the all the workers to exit. Keeping this channel unbuffered guarantees that the account data channel is empty before the `wg.Wait()` unblocks.

trianglesphere: Oh I see, the collector is it's own go-routine. It doesn't seem like it need to be, but making it buffered does require a fair bit more changes.

mslipper: Yeah. LMK if you feel strongly about any modifications here, otherwise I'll leave as-is.

PR #5199, opened 2023-03-20
op-chain-ops: Improve OVM_ETH post check

mslipper:
Improves OVM_ETH post-checks by sampling balances and making sure that they appear in state after the migration. Previously, the OVM_ETH post-check only verified that the total supply was set to zero.

Will test this on a fresh L2 migration.

Depends on #5198


mslipper:
Converting to draft while the test runs.

mslipper:
Migration completed successfully.

op-chain-ops/genesis/check.go
```diff
@@ -365,14 +376,94 @@ func PostCheckPredeployStorage(db vm.StateDB, finalSystemOwner common.Address, p
 }
 
 // PostCheckLegacyETH checks that the legacy eth migration was successful.
-// It currently only checks that the total supply was set to 0.
-func PostCheckLegacyETH(db vm.StateDB) error {
+// It checks that the total supply was set to 0, and randomly samples storage
```
tynes, line 379: Curious how much time this saves in practice. Also wondering if the random sampling part should happen after the various checks that return early anyways

mslipper: Single-threaded iteration takes about 40 minutes. We can parallelize and read the whole state, but that will make the migration take longer.

I can do the state read as part of a separate PR - will require a refactor of the OVM_ETH migration itself to utilize a common lib for parallel state traversal.

mslipper: This already runs after the various checks that return early with the exception of the withdrawals migration.

tynes (approved):

PR #5204, opened 2023-03-20
fix(op-node): Peer Scoring Metrics

refcell:
**Description**

Fixes CLI-3646

Adds a quick-fix to the peer scoring p2p flags in `op-node` and adds back the `peer_id` so we can track peer scores in testing environments.


op-node/metrics/metrics.go
```diff
@@ -302,6 +302,7 @@ func NewMetrics(procName string) *Metrics {
 			// No label names here since peer ids would open a service attack vector.
 			// Each peer id would be a separate metric, flooding prometheus.
 			// See: https://prometheus.io/docs/practices/naming/#labels
```
tynes, line 304: Wait does this go against the comment?

refcell: Yes.

op-node/flags/p2p_flags.go
```diff
@@ -276,6 +276,9 @@ var p2pFlags = []cli.Flag{
 	NoDiscovery,
 	P2PPrivPath,
 	P2PPrivRaw,
+	PeerScoring,
```
trianglesphere, line 279: Any other flags we missed?

op-node/metrics/metrics.go
refcell: I don't know.

trianglesphere: Why do we need to "actually track how peer scores are calculated on a per-peer basis". Is this something where we don't have insight into the system. Do we need this to make sure that the feature is working, or will we also need this on an ongoing basis to track the performance of the peer scores feature.

refcell: This is the only metrics for peer scoring atm. So I don't see how else we would glean insight into peer scores.

Again, i'm not sure how to track this - I haven't touched prometheus metrics much so any direction here would be greatly appreciated.

tynes: there should be an RPC api that exposes metrics on a peer to peer basis? i believe we can set up in grafana curl calls to build a chart

trianglesphere: > how else we would glean insight into peer scores

We just need to figure out metrics that don't include the peer ID. The limit with grafana is that if we include externally controllable data as a label, we expose ourselves to a DOS attack. Can do a distribution of scores. Can do events when we want to ban peers (& then use logs to figure out which peer we banned). Can do events when we raise/lower scores. I'm sure there are more that I haven't thought of.

op-node/flags/p2p_flags.go
refcell: Nope!

op-node/metrics/metrics.go
```diff
@@ -491,8 +500,25 @@ func (m *Metrics) RecordGossipEvent(evType int32) {
...
+	}
+}
+
 func (m *Metrics) RecordPeerScoring(peerID peer.ID, score float64) {
-	m.PeerScores.WithLabelValues(peerID.String()).Set(score)
+	prevScore, ok := m.PeerStore[peerID]
+	if ok {
+		m.PeerScores.WithLabelValues(getLabelForScore(prevScore)).Dec()
+	}
+	m.PeerScores.WithLabelValues(getLabelForScore(score)).Inc()
```
sebastianst, line 520: Could be improved not to update if score stays the same. Otherwise there would be a decrease followed by an increase right after.

op-node/metrics/metrics.go
```diff
@@ -491,8 +500,25 @@ func (m *Metrics) RecordGossipEvent(evType int32) {
...
+	m.PeerStore[peerID] = score
```
ajsutton, line 521: Discussed offline with @refcell and @protolambda - there's a memory leak here because previous scores aren't removed when libp2p eventually forgets a peer ID.

Simplest solution is likely to do the grouping in `peer_scorer.go` by updating the `SnapshotHook` to something like:
```go
func (s *ScoreBands) Record(score float64) {
	if score <= -40 {
		s.Banned = s.Banned + 1
	} else if score <= 0 {
		s.Low = s.Low + 1
	} ... etc
}

// SnapshotHook returns a function that is called periodically by the pubsub library to inspect the peer scores.
// It is passed into the pubsub library as a [pubsub.ExtendedPeerScoreInspectFn] in the [pubsub.WithPeerScoreInspect] option.
// The returned [pubsub.ExtendedPeerScoreInspectFn] is called with a mapping of peer IDs to peer score snapshots.
func (s *scorer) SnapshotHook() pubsub.ExtendedPeerScoreInspectFn {
	return func(m map[peer.ID]*pubsub.PeerScoreSnapshot) {
		scoreBands := &ScoreBands{}
		for id, snap := range m {
			scoreBands.Record(snap.Score)

			// Update with the peer gater
			s.gater.Update(id, snap.Score)
		}
		s.metricer.RecordPeerScoring(scoreBands)
	}
}
```

The `ScoreBands` struct maybe should live in metrics module - not sure...

sebastianst (changes requested):

ajsutton:
The `Bucketable` interface is a really nice little abstraction.  Good thinking.

op-node/metrics/metrics.go
```diff
@@ -116,6 +120,9 @@ type Metrics struct {
 
 	TransactionsSequencedTotal prometheus.Counter
 
+	// Store a map of peer to their score so we can build a score histogram
+	PeerStore map[peer.ID]float64
+
```
sebastianst, line 125: Is this still used? Please make sure to clean up all unused artifacts from the first version of the implementation.

op-node/metrics/metrics.go
```diff
@@ -36,6 +36,10 @@ const (
 	BatchMethod = "<batch>"
 )
 
+type Bucketable interface {
+	PlaceIntoBucket(*prometheus.GaugeVec)
+}
+
```
sebastianst, line 42: See other comment, this looks fancy but imo removes the abstraction that the `metrics` package should provide, an abstraction around prometheus.

op-node/p2p/peer_scorer.go
```diff
@@ -43,18 +44,50 @@ func NewScorer(peerGater PeerGater, peerStore Peerstore, metricer GossipMetricer
...
+type ScoreBands struct {
+	Graylist int
+	NoGossip int
+	NoPX     int
+	Friend   int
+}
+
+// Record records a peer score in the appropriate band.
+func (s *ScoreBands) Record(score float64) {
+	thresholds := NewPeerScoreThresholds()
```
sebastianst, line 58: where is `NewPeerScoreThresholds()`?

op-node/p2p/peer_scorer.go
```diff
@@ -43,18 +44,50 @@ func NewScorer(peerGater PeerGater, peerStore Peerstore, metricer GossipMetricer
...
+	if score <= thresholds.GraylistThreshold {
+		s.Graylist = s.Graylist + 1
+	} else if score <= thresholds.GossipThreshold {
+		s.NoGossip = s.NoGossip + 1
+	} else if score <= thresholds.AcceptPXThreshold {
+		s.NoPX = s.NoPX + 1
+	} else {
+		s.Friend = s.Friend + 1
+	}
+}
```
sebastianst, line 68: I'd prefer to have a non-hard-coded scores band config. E.g. we could have a config parameter `OP_NODE_METRICS_P2P_SCORE_BANDS` of the format `<threshold>:<label>;...` like `-40:graylist;0:no_gossip;...` and then read this out during setup of the metrics. Should require that thresholds are ordered.

sebastianst (changes requested):
please make sure to clean up old implementation artifacts

op-node/metrics/metrics.go
```diff
@@ -287,21 +294,27 @@ func NewMetrics(procName string) *Metrics {
...
+			"banned",
+			"low",
+			"medium",
+			"high",
+		}),
+		StreamCount: factory.NewGauge(prometheus.GaugeOpts{
+			Namespace: ns,
+			Subsystem: "p2p",
+			Name:      "stream_count",
+			Help:      "Count of currently connected p2p streams",
```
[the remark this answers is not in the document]
refcell: If you look at this diff, this was previously present, so yes, I'd assume so.

op-node/p2p/peer_scorer.go
```diff
@@ -43,18 +44,50 @@ func NewScorer(peerGater PeerGater, peerStore Peerstore, metricer GossipMetricer
...
+
+// PlaceIntoBucket satisfies the [metrics.Bucketable] interface.
+func (s *ScoreBands) PlaceIntoBucket(gaugeVec *prometheus.GaugeVec) {
+	gaugeVec.WithLabelValues("graylist").Set(float64(s.Graylist))
+	gaugeVec.WithLabelValues("no_gossip").Set(float64(s.NoGossip))
+	gaugeVec.WithLabelValues("no_px").Set(float64(s.NoPX))
+	gaugeVec.WithLabelValues("friend").Set(float64(s.Friend))
+}
```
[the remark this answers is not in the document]
refcell: That's fair, I was trying to interface things as much as possible to prevent having to shove `ScoreBands` into Metrics because that's an ugly solution. But, ok switching to that if this is too clunky.

op-node/p2p/peer_scorer.go
refcell: in `peer_params.go`

op-node/p2p/peer_scorer.go
sebastianst: OK I'm fine with not doing it in this PR but please create a ticket for it in Linear. Right now, with these static bands, if we look at the metrics and want to adjust thresholds, we have to go through the full cycle of opening a PR here, create a release and get it deployed. I would like to avoid these cycles by just making them configurable in the first place. Shouldn't be much work. Just parsing a config string and working on a `map[int]int` and `[]int` of thresholds instead of the static struct.

refcell: Will just impl

op-node/flags/p2p_flags.go
```diff
@@ -34,6 +34,15 @@ var (
...
 		EnvVar:   p2pEnv("PEER_SCORING"),
 	}
+	PeerScoreBands = cli.StringFlag{
+		Name: "p2p.score.bands",
+		Usage: "Sets the peer score bands used primarily for peer score metrics. " +
+			"Should be provided in following format: <threshold>:<label>;<threshold>:<label>;..." +
+			"For example: -40:graylist;-20:restricted;0:nopx;20:friend;",
+		Required: false,
+		Value:    "none",
+		EnvVar:   p2pEnv("PEER_SCORING"),
```
protolambda, line 44:
```suggestion
		EnvVar:   p2pEnv("SCORE_BANDS"),
```

protolambda (approved):

protolambda (approved):

op-node/p2p/peer_scorer.go
```diff
@@ -1,16 +1,92 @@
...
+// Bucket returns the appropriate band for a given score.
+func (s *bandScoreThresholds) Bucket(score float64) string {
+	for band, threshold := range s.bands {
+		if score >= threshold {
+			return band
+		}
+	}
+	// If there is no band threshold lower than the score,
+	// the peer must be placed in the lowest bucket.
+	return s.lowestBand
```
sebastianst, line 89: This doesn't work unfortunately because map iteration in Go is in random order. It's a contentious "feature" of Go.

In the constructor/`Parse` you have to record the name-threshold pairs in a slice to get an order and then iterate over that slice so that you return the highest threshold that is >= the score.

Please also add a test for this, it would have failed here.

 The constructor should also validate that the thresholds are ordered. This should also be tested.

op-node/p2p/band_scorer_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// TestBandScorer_ParseDefault tests the [BandScorer.Parse] function
+// on the default band scores cli flag value.
+func TestBandScorer_ParseDefault(t *testing.T) {
+	defaultScoringBands := "-40:graylist;-20:restricted;0:nopx;20:friend;"
+
+	// Create a new band scorer.
+	bandScorer := NewBandScorer()
```
sebastianst, line 15: See other comment, please also test `Bucket` and add an additional test of `Parse` that parsing fails if thresholds are not ordered from low to high.

op-node/p2p/peer_scorer.go
```diff
@@ -1,16 +1,92 @@
...
+//
+//go:generate mockery --name BandScorer --output mocks/
+type BandScorer interface {
+	Parse(str string) error
+	Bucket(score float64) string
+	Reset()
+}
+
+// NewBandScorer constructs a new [BandScorer] instance.
+func NewBandScorer() BandScorer {
```
sebastianst, line 43: It is good practice in Go to _return_ concrete implementations while _consuming_ interfaces, unless there's a specific reason to return an interface. There are many write ups on that, like [this one](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a).

Here, it probably makes sense to make the `bandScoreThresholds` type public and return it. Then you also don't need to convert the interface to the concrete type in the tests.

op-node/p2p/peer_scorer.go
```diff
@@ -1,16 +1,92 @@
...
+	return &bandScoreThresholds{
+		bands: make(map[string]float64),
+	}
+}
+
+// Reset wipes the internal state of the [BandScorer].
+func (s *bandScoreThresholds) Reset() {
+	s.bands = make(map[string]float64)
```
sebastianst, line 51: It is more efficient in Go to reuse maps, especially here because they retain the same cardinality and keys. You might want to iterate here and set all values to 0. Or you're lazy and [`Clear`](https://pkg.go.dev/golang.org/x/exp/maps#Clear) it.

op-node/p2p/peer_scorer.go
```diff
@@ -1,16 +1,92 @@
...
+	log                 log.Logger
+	gater               PeerGater
+	bandScoreThresholds BandScorer
+}
+
+// bandScoreThresholds holds the thresholds for classifying peers
+// into different score bands.
+type bandScoreThresholds struct {
+	bands      map[string]float64
+	lowestBand string
```
sebastianst, line 25: can be lowest value in slice

sebastianst (changes requested):

op-node/p2p/band_scorer_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+// TestBandScorer_BucketCorrectly tests the [BandScorer.Bucket] function
+// on a variety of scores.
+func TestBandScorer_BucketCorrectly(t *testing.T) {
+	// Create a new band scorer.
+	bandScorer := NewBandScorer()
+	require.NoError(t, bandScorer.Parse("-40:graylist;-20:restricted;0:nopx;20:friend;"))
+
+	// Let's validate that the [BandScorer] sorts the bands correctly.
+	require.Equal(t, bandScorer.bands, []scorePair{
```
trianglesphere, line 33: No need to test the parsing again.

op-node/p2p/band_scorer_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+// TestBandScorer_BucketInverted tests the [BandScorer.Bucket] function
+// on a variety of scores, in descending order.
+func TestBandScorer_BucketInverted(t *testing.T) {
+	// Create a new band scorer.
+	bandScorer := NewBandScorer()
+	require.NoError(t, bandScorer.Parse("20:friend;0:nopx;-20:restricted;-40:graylist;"))
+
+	// Let's validate that the [BandScorer] sorts the bands correctly.
+	require.Equal(t, bandScorer.bands, []scorePair{
```
trianglesphere, line 60: Ditto no need to test parsing here. If this is a different parse test, move all parsing testing to a single spot.

op-node/p2p/peer_scorer.go
```diff
@@ -48,13 +137,20 @@ func NewScorer(peerGater PeerGater, peerStore Peerstore, metricer GossipMetricer
 // The returned [pubsub.ExtendedPeerScoreInspectFn] is called with a mapping of peer IDs to peer score snapshots.
 func (s *scorer) SnapshotHook() pubsub.ExtendedPeerScoreInspectFn {
 	return func(m map[peer.ID]*pubsub.PeerScoreSnapshot) {
+		// Reset the score bands
+		s.bandScoreThresholds.Reset()
```
trianglesphere, line 141: Why do we do the reset here?

op-node/p2p/peer_scorer_test.go
```diff
@@ -92,17 +101,23 @@ func (testSuite *PeerScorerTestSuite) TestSnapshotHookBlockPeer() {
...
 		testSuite.logger,
 	)
 	inspectFn := scorer.SnapshotHook()
 
-	// Mock the snapshot updates
-	// This doesn't return anything
-	testSuite.mockMetricer.On("RecordPeerScoring", peer.ID("peer1"), float64(-101)).Return(nil)
+	// Mock the band scorer calls
+	testSuite.mockBandScorer.On("Reset").Return(nil)
+	testSuite.mockBandScorer.On("Bucket", float64(-101)).Return("graylist")
```
trianglesphere, line 111: If you used the actual `bandScorer` here would you get the same results?

op-node/p2p/peer_scorer.go
```diff
@@ -1,16 +1,104 @@
...
+// BandScorer is an interface for placing peer scores
+// into various bands.
+//
+// Implementations are expected to construct internals using the
+// [Parse] function and then expose the [Bucket] function for
+// downstream [BandScorer] consumers.
+//
+//go:generate mockery --name BandScorer --output mocks/
+type BandScorer interface {
+	Parse(str string) error
```
trianglesphere, line 42: When do we call `Parse` after construction? It seems like `parsing` should be part of the `new` step & then we only retain the `Bucket()` function in this interface.

trianglesphere (commented):
I really like the score bands & stream_count metrics, but I'm a little confused on how the `BandScorer` is being used.

ajsutton:
hmm, running this locally with `--p2p.scoring.peers=light --p2p.ban.peers` I always get:
```
op_node_default_p2p_peer_scores{band=""} 1
```

and my very malicious peer that's spamming invalid gossip isn't being disconnected.
Even when I specify `--p2p.score.bands="-40:graylist;-20:restricted;0:nopx;20:friend;"` as well I still don't get any named bands in metrics.

op-node/p2p/peer_scores_test.go
```diff
@@ -118,8 +121,10 @@ func (testSuite *PeerScoresTestSuite) TestNegativeScores() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	testSuite.mockMetricer.On("RecordPeerScoring", mock.Anything, float64(0)).Return(nil)
-	testSuite.mockMetricer.On("RecordPeerScoring", mock.Anything, float64(-1000)).Return(nil)
+	testSuite.mockBandScorer.On("Reset").Return(nil)
+	testSuite.mockBandScorer.On("Bucket", float64(0)).Return("nopx")
+	testSuite.mockBandScorer.On("Bucket", float64(-1000)).Return("graylist")
+	testSuite.mockMetricer.On("SetPeerScores", mock.Anything).Return(nil)
```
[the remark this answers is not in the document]
refcell: Yea, it's still testing the correct assertions, so I think it's fair to leave this for now to get this pr over the line - starting to lose momentum here..

op-node/flags/p2p_flags.go
```diff
@@ -34,6 +34,15 @@ var (
 		Value:    "none",
 		EnvVar:   p2pEnv("PEER_SCORING"),
 	}
+	PeerScoreBands = cli.StringFlag{
```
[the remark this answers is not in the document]
refcell: Good catch 🙏

op-node/p2p/peer_scorer_test.go
refcell: Yes - Will remove the mock to trim down the pr :)

op-node/p2p/peer_scorer.go
refcell: Done

op-node/p2p/peer_scorer.go
```diff
@@ -1,16 +1,99 @@
...
+
+// BandScorer is an interface for placing peer scores
+// into various bands.
+//
+// Implementations are expected to construct internals using the
+// [Parse] function and then expose the [Bucket] function for
+// downstream [BandScorer] consumers.
+type BandScorer interface {
+	Bucket(score float64) string
+	Reset()
```
sebastianst, line 42: remove

op-node/p2p/peer_scorer.go
```diff
@@ -1,16 +1,99 @@
...
+		return s.bands[i].threshold < s.bands[j].threshold
+	})
+
+	return s, nil
+}
+
+// Reset wipes the internal state of the [BandScorer].
+func (s *BandScoreThresholds) Reset() {
+	s.bands = s.bands[:0]
+}
```
sebastianst, line 82: remove, we don't use & need

op-node/p2p/peer_scorer.go
```diff
@@ -1,16 +1,99 @@
...
+	bands []scorePair
+}
+
+// BandScorer is an interface for placing peer scores
+// into various bands.
+//
+// Implementations are expected to construct internals using the
+// [Parse] function and then expose the [Bucket] function for
+// downstream [BandScorer] consumers.
+type BandScorer interface {
```
sebastianst, line 40: Is there now still a reason for having the interface? It's not mocked any more in tests and I don't see a use case for injecting a different logic here right now.

sebastianst (changes requested):
Looks good to me now, excited to see this in action soon 🤩 Just some minor suggestions for further simplification.

op-node/p2p/peer_scorer.go
refcell: alrighty...

sebastianst: yes, if there's a reason for an inteface

trianglesphere: The other half to `accept interfaces, return structs` is that the interface should (usually) be defined at the point of usage, not in the code which defines the struct (unless it's something like `io.Read` where there are a lot of different implementations in the same package)

op-node/p2p/prepared.go
```diff
@@ -68,6 +68,10 @@ func (p *Prepared) PeerScoringParams() *pubsub.PeerScoreParams {
 	return nil
 }
 
+func (p *Prepared) PeerBandScorer() *BandScoreThresholds {
+	return nil
```
trianglesphere, line 72: Does this cause problems in the e2e tests? I don't think we check for the scorer being nil. I think we'd need metrics on in the e2e tests to cause problems here.

trianglesphere (commented):
ops changes lgtm. Only one q about the prepared p2p stuff

trianglesphere (approved):

PR #5205, opened 2023-03-20
contracts: add a method for building legacy bindings

tynes:
**Description**

This will be useful for porting hh tasks to go because hh runs out of memory. This creates a new bindings package called `legacy`. There are no changes to the legacy contracts, so we do not need to worry about checking for changes.


Inphi (approved):
LGTM!

maurelian (approved):

PR #5206, opened 2023-03-20
feat(ctb): Add constructor to init the SD implementation

maurelian:
**Description**

Adds a constructor to initialize the Migration System Dictator. 

**Metadata**
- Fixes #CLI-3450


packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
clabby, line 1: Deploy scripts need to be updated to account for the implementation already being initialized: https://github.com/ethereum-optimism/optimism/blob/98fbe9d223e908bd9d1209a83d491a8f0225a4b4/packages/contracts-bedrock/deploy/019-SystemDictatorInit.ts#L117

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
```diff
@@ -155,6 +155,33 @@ contract SystemDictator is OwnableUpgradeable {
         currentStep++;
     }
 
+    /**
+     * @notice Constructor required to ensure that the implementation of the SystemDictator is
+     *         initialized upon deployment.
+     */
+    constructor() {
```
clabby, line 162: Not sure if I'm missing something, but doesn't this just make the contract unusable? Wouldn't we want to pass a `DeployConfig` into the constructor and initialize all of the values upon deployment?

tynes: This enforces that it must be used from behind a proxy

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
maurelian: That line is just generating the data to submit. 
The full context:

https://github.com/ethereum-optimism/optimism/blob/98fbe9d223e908bd9d1209a83d491a8f0225a4b4/packages/contracts-bedrock/deploy/019-SystemDictatorInit.ts#L115-L118

shows that this is making a call to the Proxy, which is uninit'd because the constructor doesn't run in its context.

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
```diff
@@ -155,6 +155,33 @@ contract SystemDictator is OwnableUpgradeable {
...
+        // Using this shorter variable as an alias for address(0) just prevents us from having to
```
tynes, line 163: when your style guide forces you to leave comments explaining things

tynes (approved):

tynes:
I do like how the impl is initialized with null values, I feel like we should be doing similar things with our other contracts to ensure they cannot be used in scams or by accident. Would require refactors of the tests for any that are not used behind proxies, since it enforces a proxy must be used

PR #5207, opened 2023-03-20
feat(ctb): Add larger phases containing migration steps to MSD

maurelian:
Adds `phase1()` and `phase2()` methods to the MSD which consolidates the various `stepN()` methods into two groups. 

This will significantly reduce the number of multisig transactions which need to be executed.

I tested this primarily by running `npx hardhat deploy` scripts locally.

Converted to draft. I'd still like feedback on this, but we should be careful about deciding whether or not to make the change.

### DIFF
--- a/packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
+++ b/packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
@@ -411,26 +411,9 @@ contract SystemDictator is OwnableUpgradeable {
     }
 
     /**
-     * @notice Calls the first 2 steps of the migration process.
-     */
-    function phase1() external onlyOwner {
-        step1();
-        step2();
-    }
-
-    /**
-     * @notice Calls the remaining steps of the migration process.
-     */
-    function phase2() external onlyOwner {
-        step3();
-        step4();
-        step5();
-    }
-
-    /**
      * @notice Tranfers admin ownership to the final owner.
      */
-    function finalize() external onlyOwner {
+    function finalize() public onlyOwner {
         // Transfer ownership of the ProxyAdmin to the final owner.
         config.globalConfig.proxyAdmin.transferOwnership(config.globalConfig.finalOwner);
 
@@ -455,6 +438,24 @@ contract SystemDictator is OwnableUpgradeable {
 
         // Mark the deployment as finalized.
         finalized = true;
+    }
+
+    /**
+     * @notice Calls the first 2 steps of the migration process.
+     */
+    function phase1() external onlyOwner {
+        step1();
+        step2();
+    }
+
+    /**
+     * @notice Calls the remaining steps of the migration process.
+     */
+    function phase2() external onlyOwner {
+        step3();
+        step4();
+        step5();
+        finalize();
     }
 
     /**
--- a/packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
+++ b/packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
@@ -215,7 +215,7 @@ contract SystemDictator is OwnableUpgradeable {
     /**
      * @notice Configures the ProxyAdmin contract.
      */
-    function step1() external onlyOwner step(1) {
+    function step1() public onlyOwner step(1) {
         // Set the AddressManager in the ProxyAdmin.
         config.globalConfig.proxyAdmin.setAddressManager(config.globalConfig.addressManager);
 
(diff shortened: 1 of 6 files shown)


packages/contracts-bedrock/deploy-config/hardhat.json
```diff
@@ -16,12 +17,22 @@
...
-  "baseFeeVaultRecipient": "0xBcd4042DE499D14e55001CcbB24a551F3b954096",
+  "l2GenesisBlockGasLimit": "0x17D7840",
+
+  "baseFeeVaultRecipient": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
+  "l1FeeVaultRecipient": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
+  "sequencerFeeVaultRecipient": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
+
+  "governanceTokenName": "Optimism",
+  "governanceTokenSymbol": "OP",
+  "governanceTokenOwner": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
```
maurelian, line 34: These values are required to satisfy the [deployConfig](https://github.com/ethereum-optimism/optimism/blob/e9fe9c2c59cb63167ea8ec5ab9657166520a68c1/packages/contracts-bedrock/src/deploy-config.ts#L116-L151) type. 

Although I think some of them should have been made optional, I didn't want to expand the scope of this PR to do that.

tynes:
I am not opposed to merging this. Is there any additional work that you need to do to move it from draft to ready for review?

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
```diff
@@ -440,6 +440,24 @@ contract SystemDictator is OwnableUpgradeable {
         finalized = true;
     }
 
+    /**
```
tynes, line 443: Just wondering if we want to have the steps be public or internal. I have flip flopped on this decision multiple times, I don't feel like there is an obvious answer so I'm fine with leaving as is

packages/contracts-bedrock/deploy/021-SystemDictatorSteps-2.ts
```diff
@@ -221,17 +166,59 @@ const deployFn: DeployFunction = async (hre) => {
...
     isLiveDeployer,
     SystemDictator,
-    step: 5,
+    phase: 2,
     message: `
+      Phase 2 consists of the following steps:
+
+      Step 3 will clear out some legacy state from the AddressManager. Once you execute this step,
+      you WILL NOT BE ABLE TO RESTART THE SYSTEM using exit1(). You should confirm that the L2
+      system is entirely operational before executing this step.
```
tynes, line 178: The problem here is that the L2OO isn't initialized, so "make sure the L2 system is entirely operational" isn't quite possible. https://github.com/ethereum-optimism/optimism/pull/5253 is an attempt to fix this, but it needs changes to work correctly

packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -373,6 +394,61 @@ export const doStep = async (opts: {
...
+ * @param opts.checks Checks to perform after executing the step.
+ */
+export const doPhase = async (opts: {
+  isLiveDeployer?: boolean
+  SystemDictator: ethers.Contract
+  phase: number
+  message: string
+  checks: () => Promise<void>
+}): Promise<void> => {
+  if (!(await isStartOfPhase(opts.SystemDictator, opts.phase))) {
```
tynes, line 414: nit: this code is much more difficult to read than it needs to be, nesting `await` in expressions is an antipattern

```js
const isStart = await isStartOfPhase(...)
if (!isStart) { ... }
```

packages/contracts-bedrock/deploy/021-SystemDictatorSteps-2.ts
maurelian: You're right. This also came up on that PR. 

I'm not sure we actually want to change the step numbers, but it does feel to me like the `updateDynamicConfig()` call is best treated as if it was `phase2` in 3 (rather than 2) step process.

packages/contracts-bedrock/src/deploy-utils.ts
maurelian: Good call out. It followed the same pattern used above in `isStep()`, but I'll fix both for consistency.

maurelian:
Closing, in favor of a set of smaller PRs.

PR #5211, opened 2023-03-21
op-chain-ops: Fix data race in error handling

mslipper:
The error channel in the OVM_ETH migration is buffered. In rare cases, errors were written to this channel in such a way that they were not processed by the collector goroutine. This meant that errors were not being caught by the `lastErr != nil` check, and were instead triggering the total supply check below.

This error reliably reproduces when running the new TestMigrateBalancesRandomMissing test. This test generates a random state, and randomly removes an address/allowance from the witness data. It runs this 100 times per test. We didn't catch this with the other random test because it doesn't exercise the error handling codepath.


ajsutton (approved):

mslipper:
Linter error is a flake, will re-run once other tests pass.

PR #5213, opened 2023-03-21
op-chain-ops: Extract concurrent state iterator into util

mslipper:
Pulls the concurrent state iterator into a re-usable library. Additional tests have been added to assert that the iterator touches every key in state at least once. This will allow us to perform a complete check of the OVM_ETH migration as the last step of the migration.


tynes:
This makes it way cleaner

PR #5215, opened 2023-03-21
contracts-bedrock: cleanup slither.db.json

s7v7nislands:
**Description**

cleanup after merge https://github.com/ethereum-optimism/optimism/pull/5201

**Additional context**

Is the slither used anymore?


tynes:
Slither runs in CI but it's currently pretty noisy

tynes:
How exactly did you generate this diff? I notice that its removing a contract that was renamed, `DepositFeed` -> `OptimismPortal`

s7v7nislands:
I just deleted the elements about DepositFeed.
I run Slither locally, it seems to have many warnings.

maurelian:
We do need to reactivate slither. TBH we should probably just blow out the DB and start again with a minimal ruleset.

maurelian:
@s7v7nislands thanks for the contribution, it's much appreciated! I'm just going to go ahead and close this issue for now as we should focus on getting slither running properly again rather than patches.

If you're interested, a rough spec for doing so would be: 
1. delete the db.json file
2. run slither in the bedrock package with only the highest severity issues
3. update the slither config to match that

Something like that. Now that the bedrock contracts are effectively frozen, I'd like to avoid peppering them with slither-disable comments.

s7v7nislands:
> @s7v7nislands thanks for the contribution, it's much appreciated! I'm just going to go ahead and close this issue for now as we should focus on getting slither running properly again rather than patches.
> 
> If you're interested, a rough spec for doing so would be:
> 
> 1. delete the db.json file
> 2. run slither in the bedrock package with only the highest severity issues
> 3. update the slither config to match that
> 
> Something like that. Now that the bedrock contracts are effectively frozen, I'd like to avoid peppering them with slither-disable comments.

I will give it a try.

PR #5226, opened 2023-03-22
op-node/p2p: Remove support for legacy gossip signing hashes

ajsutton:
**Description**

Remove support for legacy gossip signing hashes. The chain ID must now be included correctly in gossip signing hashes to be accepted.

~Leaving as draft for a while longer. We have upgraded the op-goerli sequencer to sign using the new signature hashes but should ensure we have a stable release out with that change and give time for other partners to upgrade.~

Sequencer is now updated and a stable release is out that publishes with the new signature hashes so this is ready to go.

**Metadata**
- Fixes https://linear.app/optimism/issue/CLI-3615/remove-support-for-old-gossip-signing-hash


op-node/p2p/gossip_test.go
```diff
@@ -66,43 +50,37 @@ func TestVerifyBlockSignature(t *testing.T) {
...
-
-		t.Run("WrongSigner "+test.name, func(t *testing.T) {
-			runCfg := &testutils.MockRuntimeConfig{P2PSeqAddress: common.HexToAddress("0x1234")}
-			signer := &PreparedSigner{Signer: test.newSigner(secrets.SequencerP2P)}
-			sig, err := signer.Sign(context.Background(), SigningDomainBlocksV1, cfg.L2ChainID, msg)
-			require.NoError(t, err)
-			result := verifyBlockSignature(logger, cfg, runCfg, peerId, sig[:65], msg)
-			require.Equal(t, pubsub.ValidationReject, result)
-		})
+	t.Run("Valid", func(t *testing.T) {
```
ajsutton, line 53: I kind of like this use of sub-tests to reuse the common init code and group the various signature verification tests together, but given it's not a table test anymore I'm not sure it's idiomatic.  Happy to change it to run separate tests if that's the usual approach.

sebastianst: It's always a good pattern to use `t.Run` to structure sub-tests. Doesn't only need to be used in a table test.

protolambda:
Does the PR description still apply? Didn't regolith include the signing hash verification code to handle both? If the sequencer is using the new signing code I think we're ready to merge this.

ajsutton:
> Does the PR description still apply? Didn't regolith include the signing hash verification code to handle both? If the sequencer is using the new signing code I think we're ready to merge this.

Oops, that's my bad - I forgot I'd written that in the description.  Yes I think this is good to merge now.

trianglesphere (approved):
lgtm

PR #5230, opened 2023-03-22
op-node: implement RPC poll interval and rate-limit options and provide L1 RPC CLI options

protolambda:
**Description**

- Refactor RPC endpoint config setup functions to return the sources client config. This way we don't have to pass around an increasing number of options, and we can make the existing options configurable via flags/env, like the max-batch-size for L1 now.
- Implement basic rate-limiter RPC wrapper, so users can optionally configure a sanity-limit for RPC usage.
- Fix the default HTTP-subscription-fallback poll interval: it defaults to 12 seconds now, and is configurable to be set lower.
- Refactor client.NewRPC to extend with options more easily.

**Tests**

Updated op-e2e to use the new RPC option pattern and specify a poll-time based on L1 block time.

Not sure how best to test the rate-limit functionality, but it's off by default and just a single call to x-std-lib rate limiter.

**TODOs**

Fix CLI-3694


op-node/node/client.go
```diff
@@ -135,16 +141,48 @@ type L1EndpointConfig struct {
 	// L1RPCKind identifies the RPC provider kind that serves the RPC,
 	// to inform the optimal usage of the RPC for transaction receipts fetching.
 	L1RPCKind sources.RPCProviderKind
+
+	// RateLimit specifies a self-imposed rate-limit on L1 requests. 0 is no rate-limit.
```
tynes, line 145: L1 requests in the L2 config?

op-node/node/client.go
```diff
@@ -135,16 +141,48 @@ type L1EndpointConfig struct {
...
+	// It is recommended to use websockets or IPC for efficient following of the changing block.
+	// Setting this to 0 disables polling.
+	HttpPollInterval time.Duration
 }
 
 var _ L1EndpointSetup = (*L1EndpointConfig)(nil)
 
-func (cfg *L1EndpointConfig) Setup(ctx context.Context, log log.Logger) (cl client.RPC, trust bool, kind sources.RPCProviderKind, err error) {
-	l1Node, err := client.NewRPC(ctx, log, cfg.L1NodeAddr)
+func (cfg *L1EndpointConfig) Check() error {
```
tynes, line 160: Does it make sense to have a difference between L1 rpc config and L2 rpc config? Also does it make sense to have a difference between an L1 rpc client and an L2 rpc client? Feels like a single, robust client could be implemented and then naming of struct members could differentiate which network the client is dialed to

op-node/node/client.go
protolambda: It's part of the `L1EndpointConfig`, I think the github diff folding makes it hard to read. It's not part of the l2 engine or l2 sync configs.

op-node/node/client.go
protolambda: The client code is already largely shared, it's a tree structure like this:
- eth-client -> L1-client
- eth-client -> L2-client
- L2-client -> engine client

Since each client adds additional more domain-specific caches / methods. The bulk of it is shared in the eth-client though.

The L1 configuration is a bit more separate, but since the L2 engine (JWT authenticated) and L1 (non-local, more important to optimize with rate-limits and polling etc. adjustments) are quite different, I don't think we should combine them here.

op-e2e/setup.go
```diff
@@ -413,9 +413,12 @@ func (cfg SystemConfig) Start(_opts ...SystemConfigOption) (*System, error) {
...
 		rollupCfg.L1 = &rollupNode.L1EndpointConfig{
-			L1NodeAddr: l1EndpointConfig,
-			L1TrustRPC: false,
-			L1RPCKind:  sources.RPCKindBasic,
+			L1NodeAddr:       l1EndpointConfig,
+			L1TrustRPC:       false,
+			L1RPCKind:        sources.RPCKindBasic,
+			RateLimit:        0,
+			BatchSize:        20,
+			HttpPollInterval: time.Duration(cfg.DeployConfig.L1BlockTime) * time.Second,
```
trianglesphere, line 421: Might be worth lowering this for the tests.

trianglesphere (approved):
Really nice changes. I'm not sure why the test is flaky.

op-e2e/setup.go
protolambda: It's passing locally with this time, but good point, will reduce it for faster testing time, and hopefully it improves the flake.

PR #5231, opened 2023-03-22
op-node: reset back to optimal receipt fetching methods

protolambda:
**Description**

Reset it back after a while to not fallback permanently to sub-optimal RPC methods for receipt fetching.

**Tests**

Updated test to see if method preference can be sticky and reset back if configured to do so.

**TODOs**

Fix CLI-3697


op-node/sources/receipts_test.go
```diff
@@ -226,6 +232,12 @@ func TestEthClient_FetchReceipts(t *testing.T) {
 			providerKind: RPCKindAlchemy,
 			setup:        fallbackCase(30, AlchemyGetTransactionReceipts),
 		},
+		{
+			name:         "alchemy sticky",
+			providerKind: RPCKindAlchemy,
+			staticMethod: true,
+			setup:        fallbackCase(30, AlchemyGetTransactionReceipts, AlchemyGetTransactionReceipts),
```
trianglesphere, line 239: How is does this verify undoing the fallback?

trianglesphere (commented):
Fallback seems to work. Do we know if we've hit this before?
Also, on the testing side, how does the test verify that we undo the fallback?

op-node/sources/receipts_test.go
protolambda: Compare it against the test case below, where it changes from AlchemyGetTransactionReceipts to EthGetBlockReceipts when the bool is not true.

trianglesphere (approved):
lgtm

PR #5232, opened 2023-03-22
op-node: harden eth RPC block/header retrieval by checking parameters

protolambda:
**Description**

With type-safe arguments, make sure that every block / header call result is verified to match the requested hash or number parameter. This is something the upstream geth ethclient does not do, but it seemed worth hardening, just in case of caching bugs in L1 RPC providers etc.

**Tests**

Added test cases for by-number and by-hash checking of the result.

**Metadata**

Fix CLI-3698

**TODOs**


op-node/eth/label.go
```diff
@@ -17,3 +17,9 @@ const (
 	// - L2: Derived chain tip from finalized L1 data
 	Finalized = "finalized"
 )
+
```
tynes, line 20: compile time interface check for `rpcBlockID`?

protolambda: It's a bit verbose for a private interface to export it just for a compile time check, it'd be a circular dependency (or have to move the interface), and the sources package will already complain if a block label cannot be used as `rpcBlockID`.

op-node/sources/eth_client_test.go
```diff
@@ -140,3 +140,40 @@ func TestEthClient_InfoByNumber(t *testing.T) {
...
+	ctx := context.Background()
+	m.On("CallContext", ctx, new(*rpcHeader),
+		"eth_getBlockByNumber", []any{n.String(), false}).Run(func(args mock.Arguments) {
+		*args[1].(**rpcHeader) = &rhdr2
+	}).Return([]error{nil})
+	s, err := NewL1Client(m, nil, nil, L1ClientDefaultConfig(&rollup.Config{SeqWindowSize: 10}, true, RPCKindBasic))
+	require.NoError(t, err)
+	_, err = s.InfoByNumber(ctx, uint64(n))
+	require.Error(t, err, "cannot accept the wrong block")
+	m.Mock.AssertExpectations(t)
```
trianglesphere, line 159: Do we really care about if a set of functions was called? It seems we primarily care about properly verifying the returned value.

trianglesphere (approved):
Change LGTM, though have we had issues with this before>

op-node/sources/eth_client_test.go
protolambda: Checking the mock was called doesn't hurt. Agreed it's mostly about verifying the error though.

PR #5233, opened 2023-03-22
contracts-bedrock: modularize deposit resource config

tynes:
**Description**

Modularize the deposit resource config so that the source of truth is the `SystemConfig` contract. Now when users do a deposit, the `OptimismPortal` will make a call to the `SystemConfig` to get the params instead of reading them from constant variables. This will not impact the amount of gas users spend when depositing because it tracks the gas usage of the computations required to determine how much gas will be burnt and will burn the difference between the gas used to compute how much gas needs to be burnt and the actual amount of gas to be burnt.

A future PR may have the `SystemConfig` emit an event when the resource config is updated, this would allow the `op-node` to track the config for the deposit gas market pricing. Right now it would be a no-op, but in the future we could move the computations that currently happen on L1 to make the 1559 curve into the `op-node` so that the expensive calculations happen offchain and then allow the user to pay for the gas on L2.

Closes CLI-3704

**TODOs**


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -135,19 +141,22 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
+     * @param _config                    Address of the SystemConfig contract.
      */
     constructor(
         L2OutputOracle _l2Oracle,
         address _guardian,
-        bool _paused
-    ) Semver(1, 2, 0) {
+        bool _paused,
+        SystemConfig _config
+    ) Semver(1, 3, 0) {
```
maurelian, line 156: Nit, semver was bumped [here](https://github.com/ethereum-optimism/optimism/pull/5006), which hasn't been deployed to prod yet. Do we need this bump?

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -197,7 +224,7 @@ contract SystemConfig is OwnableUpgradeable, Semver {
 
     /**
      * @notice Low level setter for the unsafe block signer address. This function exists to
-     *         deduplicate code around storing the unsafeBlockSigner address in storage.
+     *         deduplicate code arou,nd storing the unsafeBlockSigner address in storage.
```
maurelian, line 227: nit typo

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -197,6 +206,30 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         // Intentionally empty.
     }
 
+    // TODO: I don't like having so many return args but I also don't like
+    // wrapping them into a struct
```
maurelian, line 210: Agree neither are nice, but I think I prefer the struct...

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -197,6 +206,30 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
+    function resourceConfig() public view override returns (SystemConfig.ResourceConfig memory) {
```
maurelian, line 211: I'm a bit unclear why this function needs to go into the Portal when it is only used in the ResourceMetering code?

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
clabby: +1 - the struct is more readable imo. The tuple is a bit more efficient, but copying it back into memory wipes that efficiency away anyways.

packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -168,14 +133,21 @@ abstract contract ResourceMetering is Initializable {
         }
     }
 
+    /**
+     * @notice
+     */
+    function resourceConfig() public virtual returns (SystemConfig.ResourceConfig memory);
```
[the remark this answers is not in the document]
clabby: I think it's here mainly due to the `ResourceMetering` contract needing access to the `SystemConfig` that exists as a state var within a contract that inherits it. Public state variables can only override functions w/ external visibility, which would mean that we'd need to perform an extra call whenever we wanted to access the `SystemConfig`'s `resourceConfig` function from within `ResourceMetering` if we went that route w/o any other modifications.

Agree that it's a weird addition to the portal, though. Maybe instead of having the `SystemConfig` live in the portal, we place it in `ResourceMetering`? I think eventually we'll need it for other things in the portal, but for now, that might be a way to clean this up a bit.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: I believe we should specify bumping semver on each change because we other projects may deploy at any time. Kind of annoying, otherwise we should be very specific about releases of the contracts as a system

maurelian: OK, that makes sense.

packages/contracts-bedrock/deploy/017-SystemConfigImpl.ts
```diff
@@ -18,6 +23,14 @@ const deployFn: DeployFunction = async (hre) => {
...
       hre.deployConfig.l2GenesisBlockGasLimit,
       hre.deployConfig.p2pSequencerAddress,
+      {
+        maxResourceLimit: 20_000_000,
+        elasticityMultiplier: 10,
+        baseFeeMaxChangeDenominator: 8,
+        systemTxMaxGas: 1_000_000,
+        minimumBaseFee: ethers.utils.parseUnits('1', 'gwei'),
+        maximumBaseFee: uint128Max,
+      },
```
maurelian, line 33: This feels like hidden config. Shouldn't these go in a `deploy-config/{{NETWORK}}.json` file?

specs/guaranteed-gas-market.md
maurelian, line 1: comment only: thanks for remembering the specs.

packages/contracts-bedrock/deploy/017-SystemConfigImpl.ts
maurelian: Your reasoning make sense to me. Putting it into a short comment here would be good.

refcell (approved):
Nice; lovingly grated, tastefully made

specs/guaranteed-gas-market.md
tynes: ty

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -65,10 +95,16 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
 
     /**
-     * @notice L2 gas limit.
+     * @notice L2 block gas limit.
      */
     uint64 public gasLimit;
 
+    /**
+     * @notice The configuration for the deposit fee market. Used by the OptimismPortal
+     *         to meter the cost of buying L2 gas on L1.
```
smartcontracts, line 104: Worth adding a comment here as to why this is internal. I know it's explained in the external function but nice to have here too.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -79,52 +115,66 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
         uint256 _scalar,
         bytes32 _batcherHash,
         uint64 _gasLimit,
-        address _unsafeBlockSigner
-    ) Semver(1, 0, 1) {
-        initialize(_owner, _overhead, _scalar, _batcherHash, _gasLimit, _unsafeBlockSigner);
+        address _unsafeBlockSigner,
+        ResourceConfig memory _config
+    ) Semver(1, 1, 0) {
+        initialize({
```
smartcontracts, line 137: Fine, I'll let it slide for this one :wink:

op-e2e/e2eutils/setup.go
```diff
@@ -79,7 +79,7 @@ func MakeDeployParams(t require.TestingT, tp *TestParams) *DeployParams {
 		L1GenesisBlockNonce:         0,
 		CliqueSignerAddress:         common.Address{}, // proof of stake, no clique
 		L1GenesisBlockTimestamp:     hexutil.Uint64(time.Now().Unix()),
-		L1GenesisBlockGasLimit:      15_000_000,
+		L1GenesisBlockGasLimit:      30_000_000,
```
maurelian, line 82: Why does this change?

op-e2e/setup.go
```diff
@@ -85,7 +85,7 @@ func DefaultSystemConfig(t *testing.T) SystemConfig {
 		L1GenesisBlockBaseFeePerGas: uint642big(7),
 
 		L2GenesisBlockNonce:         0,
-		L2GenesisBlockGasLimit:      8_000_000,
+		L2GenesisBlockGasLimit:      30_000_000,
```
maurelian, line 88: Why does this change? Should it be 20M like in the system config?

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +257,69 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     *         - base fee change denominator must be greater than 0
+     *         - max resource limit plus system tx gas must be less than or
+     *           equal to the L2 gas limit
+     *         - elasticity multiplier must be greater than 0
+     *         - no precision loss when computing target resource limit
+     *
+     * @param _config The new resource config
+     */
+    function _setResourceConfig(ResourceConfig memory _config) internal {
+        require(
```
smartcontracts, line 293: Might be slightly cleaner if the invariants are explained inline with each of the require statements, rather than as the top-level function comment.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +257,69 @@ contract SystemConfig is OwnableUpgradeable, Semver {
             sstore(slot, _unsafeBlockSigner)
         }
     }
+
+    /**
+     * @notice A getter for the resource config. Ensures that the struct is
+     *         returned instead of a tuple.
```
smartcontracts, line 263: Missing return value comment

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +257,69 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     */
+    function resourceConfig() external view returns (ResourceConfig memory) {
```
smartcontracts, line 265: Minor nitpick is to put the view functions together near the top. Not super opinionated on that.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +257,69 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     * @notice An internal setter for the resource config. Ensures that the
+     *         config is sane before storing it. Holds the following invariants:
+     *         - min base fee must be less than or equal to max base fee
+     *         - base fee change denominator must be greater than 0
+     *         - max resource limit plus system tx gas must be less than or
+     *           equal to the L2 gas limit
+     *         - elasticity multiplier must be greater than 0
+     *         - no precision loss when computing target resource limit
+     *
+     * @param _config The new resource config
```
smartcontracts, line 290:
```suggestion
     * @param _config The new resource config.
```

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
```diff
@@ -160,6 +161,15 @@ contract SystemDictator is OwnableUpgradeable {
      *         initialized upon deployment.
      */
     constructor() {
+        SystemConfig.ResourceConfig memory rcfg = SystemConfig.ResourceConfig({
```
maurelian, line 164: Should these be zero values, to avoid giving the impression that they are actually impactful?

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +257,69 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     */
+    function _setResourceConfig(ResourceConfig memory _config) internal {
+        require(
+            _config.minimumBaseFee <= _config.maximumBaseFee,
+            "SystemConfig: min base fee must be less than max base"
+        );
+        require(_config.baseFeeMaxChangeDenominator > 0, "SystemConfig: denominator cannot be 0");
+        require(
```
smartcontracts, line 298: Worth documenting that the `gasLimit` may need to be increased before you can increase the `maxResourceLimit` or the `systemTxMaxGas`.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -29,6 +29,36 @@ contract SystemConfig is OwnableUpgradeable, Semver {
         UNSAFE_BLOCK_SIGNER
     }
 
+    /**
+     * @notice Represents the configuration for the EIP-1559 based curve for
```
smartcontracts, line 33: Nit, 100-char line length preferred for the codebase.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -29,6 +29,36 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     *                                            transaction.
+     * @custom:field maximumBaseFee               The max deposit base fee, it is clamped to this
+     *                                            value.
+     */
+    struct ResourceConfig {
+        uint32 maxResourceLimit;
+        uint8 elasticityMultiplier;
+        uint8 baseFeeMaxChangeDenominator;
+        uint32 minimumBaseFee;
+        uint32 systemTxMaxGas;
```
smartcontracts, line 58: We need to document that this parameter needs to be increased if you are adding new system txns in the stack or if the gas given to a system txn changes.

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -207,4 +257,69 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+        _resourceConfig = _config;
+    }
+
+    /**
+     * @notice Returns the minimum L2 gas limit that can be safely set for the system to
+     *         operate. The L2 gas limit must be larger than or equal to the amount of
+     *         gas that is allocated for deposits per block plus the amount of gas that
+     *         is allocated for the system transaction.
+     *         This function is used to determine if changes to parameters are safe.
```
smartcontracts, line 320: Needs return comment

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -197,6 +206,14 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         // Intentionally empty.
     }
 
+    /**
+     * @notice Getter for the resource config. Used internally by the ResourceMetering
+     *         contract. The SystemConfig is the source of truth for the resource config.
```
smartcontracts, line 211: Needs return comment

packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -5,6 +5,7 @@ import { Initializable } from "@openzeppelin/contracts/proxy/utils/Initializable
 import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
 import { Burn } from "../libraries/Burn.sol";
 import { Arithmetic } from "../libraries/Arithmetic.sol";
+import { SystemConfig } from "../L1/SystemConfig.sol";
```
smartcontracts, line 8: `ResourceConfig` struct should be defined within `ResourceMetering`, not `SystemConfig`. Better relationship between the two contracts.

packages/contracts-bedrock/contracts/L1/ResourceMetering.sol
```diff
@@ -168,14 +134,20 @@ abstract contract ResourceMetering is Initializable {
...
+
     /**
      * @notice Sets initial resource parameter values. This function must either be called by the
      *         initializer function of an upgradeable child contract.
      */
     // solhint-disable-next-line func-name-mixedcase
     function __ResourceMetering_init() internal onlyInitializing {
         params = ResourceParams({
-            prevBaseFee: INITIAL_BASE_FEE,
+            prevBaseFee: 1 gwei,
```
smartcontracts, line 150: Hmmm, need to rethink this a little in the case that 1 gwei is > than the max or < than the min.

tynes: The problem is that you cannot access other contracts in the constructor due to the way that the deploy system works, but even if its out of the range of the max or min it should be clamped in on the first run

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
tynes: :P <3

packages/contracts-bedrock/contracts/echidna/FuzzOptimismPortal.sol
```diff
@@ -3,16 +3,37 @@ pragma solidity 0.8.15;
...
 
     constructor() {
+        SystemConfig.ResourceConfig memory config = SystemConfig.ResourceConfig({
+            maxResourceLimit: 20_000_000,
+            elasticityMultiplier: 10,
+            baseFeeMaxChangeDenominator: 8,
+            minimumBaseFee: 1 gwei,
+            systemTxMaxGas: 1_000_000,
+            maximumBaseFee: type(uint128).max
+        });
```
maurelian, line 20: Do we still have a constants.sol (and a constants.ts) file that this could be the duplicated in?

op-e2e/setup.go
tynes: the invariant must hold that `deposit gas limit + system tx gas limit < L2 block gas limit`

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
tynes: Yes I agree, altho we do not do that for all of our contracts, would prefer to place null values in the impls constructor's call to `initialize` but would enforce that all contracts are tested behind a proxy which tbh kind of a good thing

op-e2e/e2eutils/setup.go
tynes: To make it like L1

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
tynes: ideally in the long term, changes to this can be used to pick how much gas is supplied to the system tx

smartcontracts:
Marking do-not-merge while review happens

tynes:
seems like the majority of comments are around documentation

op-e2e/setup.go
maurelian: OK, I'm just trying to understand; it could be as low as 21 million, but we're just bumping it up to 30  because it's cleaner?

(please resolve if that's correct)

maurelian (approved):
Approving on the basis that merging will proceed once comments have been resolved.

op-e2e/setup.go
tynes: Yup exactly! Figured we should just do what mainnet does

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
tynes: added docs here about it

packages/contracts-bedrock/contracts/echidna/FuzzOptimismPortal.sol
tynes: moved to constants, good call

packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
tynes: I agree with you but we cannot use all 0 values because of certain invariants that must hold with regards to the config values. Slight preference to leave as the default values so we know that the config will not cause a revert, basically the reverts exist to prevent dummy configs from being passed in.

clabby (approved):
LGTM once CI is green

PR #5234, opened 2023-03-22
op-node: Fix logging flags with incorrect prefix

trianglesphere:
**Description**

During the migration to using shared logging flags in PR #5085 the env var flags for the logging options got messed up for just the op-node. They had an extra underscore between the `OP_NODE` prefix and the `LOG_FLAG` value.

**TODOs**


op-node/flags/flags.go
```diff
@@ -14,10 +14,10 @@ import (
 
 // Flags
 
-const envVarPrefix = "OP_NODE_"
+const envVarPrefix = "OP_NODE"
 
 func prefixEnvVar(name string) string {
-	return envVarPrefix + name
+	return envVarPrefix + "_" + name
 }
```
sebastianst, line 21: We could instead use `op-service.PrefixEnvVar` here, like for the batcher and proposer.

ajsutton (commented):
It took me quite a bit of digging to work out why it had to be this way since this looks very much like a no-op. Would be great to add a test to cover this.  You may come up with something better but I got to:

```go
func TestCorrectEnvVarPrefix(t *testing.T) {
	for _, flag := range Flags {
		values := reflect.ValueOf(flag)
		envVarValue := values.FieldByName("EnvVar")
		if envVarValue == (reflect.Value{}) {
			t.Errorf("Failed to find EnvVar for flag %v", flag.GetName())
			continue
		}
		envVar := envVarValue.String()
		if envVar[:len("OP_NODE_")] != "OP_NODE_" {
			t.Errorf("Flag %v env var (%v) does not start with OP_NODE_", flag.GetName(), envVar)
		}
		if strings.Contains(envVar, "__") {
			t.Errorf("Flag %v env var (%v) has duplicate underscores", flag.GetName(), envVar)
		}
	}
}
```

op-node/flags/flags.go
trianglesphere: We could. That's a bigger change though. The batcher/proposer also don't have a local `prefixEnvVar` function.

trianglesphere:
@ajsutton I found this after noticing that the flag (as an env var) wasn't working on the internal devnet. I had tested this locally with the flag value. EnvVars are way harder to test because it's not an error to have undefined env vars.

ajsutton:
@trianglesphere 
> @ajsutton I found this after noticing that the flag (as an env var) wasn't working on the internal devnet. I had tested this locally with the flag value. EnvVars are way harder to test because it's not an error to have undefined env vars.

We could just skip things with undefined env vars - though currently that test passes with your fix. I've come across a few bugs with our CLI flags so would be good to start improving testing around them even if it's not perfect to start with.

op-node/flags/flags.go
sebastianst: Just thought now that you touch this, might as well delete the code duplication. Up to you.

trianglesphere: TBH this is minor enough + something we don't really want to change that I don't mind having it duplicated.

protolambda (approved):
LGTM

PR #5240, opened 2023-03-23
fix(op-node): Redundant Peer Gater Logs

refcell:
**Description**

@ajsutton reported redundant logging whereby already blocked peers were logged as being blocked on every peer gater `Update` called. To fix this, this pr simply pulls up the `ListBlockedPeers` call to check if the peer is already blocked and only emits the warning log if the peer is **not** already blocked.

**Metadata**

Fixes CLI-3700


op-node/p2p/peer_gater.go
```diff
@@ -40,15 +40,18 @@ func NewPeerGater(connGater ConnectionGater, log log.Logger, banEnabled bool) Pe
 func (s *gater) Update(id peer.ID, score float64) {
 	// Check if the peer score is below the threshold
 	// If so, we need to block the peer
+	isAlreadyBlocked := slices.Contains(s.connGater.ListBlockedPeers(), id)
 	if score < PeerScoreThreshold && s.banEnabled {
-		s.log.Warn("peer blocking enabled, blocking peer", "id", id.String(), "score", score)
+		if !isAlreadyBlocked {
```
ajsutton, line 45: We should be able to skip re-blocking the peer entirely if they're already blocked rather than just skipping the log line.

Also the `Contains(...ListBlockedPeers())` is a pretty expensive check. Inside `connGater` there's actually a nice map keyed by the blocked peer ID, but `ListBlockedPeers` copies all the blocked peers to a list and then we're doing an `O(n)` search of those. If I spam with a lot of unique peer IDs I can potentially burn a lot of CPU here.

We really just want to have a `connGater.IsBlocked(peerId)` method exposed that uses the internal map for fast look up and returns a boolean.

refcell: Don't have access to the Connection Gater's internal `blockedPeers` mapping, so I set up a sidecar `blockedMap` inside `gater` that is update alongside connection gating.

Let me know what you think of this solution 🙏

ajsutton (approved):
LGTM. Shame the libp2p doesn't provide a nice way to tell if a peer is blocked or not.

PR #5241, opened 2023-03-23
fix(op-node): Peer Score Metrics - Band Resets

refcell:
**Description**

As noted by @ajsutton, the peer score metrics bands don't reset to 0. This pr introduces a fix to the `SnapshotHook` on the peer scorer that will zero out all bands first so all bands with concrete, current peer counts are provided to the metricer's `SetPeerScores` function.

**Tests**

Updated the `TestScorer_SnapshotHook` test in the `peer_scorer_test.go` file so we mock a peer changing score and the band updating correctly when the `SnapshotHook` function is called.

**Metadata**

Fixes CLI-3699


ajsutton (approved):
LGTM.

PR #5242, opened 2023-03-23
feat(mon): create balance-mon service

smartcontracts:
**Description**
Creates a new balance-mon service inside of chain-mon and removes the legacy balance-monitor that relied on Forta. New balance-mon is much simpler and achieves the same results that the legacy package was achieving.


.github/workflows/release.yml
maurelian, line 1: comment only: could GH have made this diff any less readable?

packages/chain-mon/.env.example
```diff
@@ -1,3 +1,13 @@
+###############################################################################
+#                               ↓ balance-mon ↓                               #
+###############################################################################
+
+# RPC pointing to network to monitor balances on
+BALANCE_MON__RPC=
+
+# Comma-separated list of addresses to monitor
+BALANCE_MON__ADDRESSES=
```
maurelian, line 9: I'm not sure the best way to do this in the `BaseServiceV2` arch, but I think it'd be preferable if we could put names on the addresses like with the previous balance monitor. That would make it much easier to configure dashboards and alerting on multiple network.

https://github.com/ethereum-optimism/optimism/blob/ef5c1f7737cc272a1eb8d880de3bfe1563da9f95/packages/balance-monitor/.env.example#L6-L12

smartcontracts: Easiest options here are comma separated or JSON array, either works

smartcontracts: JSON array is slightly easier to read imo

smartcontracts: Easy to configure:

```
BALANCE_MON__ACCOUNTS='[
  {
    "address": "0x68108902De3A5031197a6eB3b74b3b033e8E8e4d",
    "nickname": "wallet1"
  }
]'
```

maurelian (approved):

maurelian:
Wondering if ArgoCD will pick up the deletion of the `balance-monitor` and deactivate it or something? 
I'd want to have this service deployed before we shut that down.

smartcontracts:
I'm planning to close this PR and replace it with two distinct PRs, one to add the new service and one to remove the old service.

PR #5246, opened 2023-03-23
txgmr,op-batcher,op-proposer: Migrate transaction creation to the txmgr

trianglesphere:
**Description**

This consolidates the `CraftTx` and `Send` functions in the transaction manger. The transaction manager is now fully responsible for creating & sending transactions.

I did have to reimplement some of the behavior inside the action testing because it is unable to handle blocking the actual proposer.

**Tests**

Added a test to ensure that the manual packing is the same the expected packing from the contracts. Note that this code is also tested by the devnet & op-e2e which exercise the full withdrawals codepath.

**Metadata**

- Fixes CLI-3685

**TODOs**


op-proposer/proposer/abi_test.go
```diff
@@ -0,0 +1,87 @@
...
+		Version:    [32]byte{},
+		OutputRoot: [32]byte{0xff},
+		BlockRef: eth.L2BlockRef{
+			Number: 10,
+		},
+		WithdrawalStorageRoot: [32]byte{0xdd},
+		StateRoot:             [32]byte{0xee},
+		Status: &eth.SyncStatus{
+			CurrentL1: eth.L1BlockRef{
+				Hash:   [32]byte{0xcc},
```
sebastianst, line 66: In general when testing encodings, I prefer to have all fields filled with random data, to avoid overlooking encoding errors where "zeros equal other zeros". You could use `testutil.RandomHash` here.

sebastianst (changes requested):
Looks very good, just one suggestion about improving the abi encoding test.

protolambda (approved):
LGTM, good refactor

sebastianst (approved):
Awesome ✨

PR #5249, opened 2023-03-24
contracts-bedrock: emit systemconfig event

tynes:
**Description**

Adds an event to the system config contract for when the deposit eip1559 config is changed. The derivation pipeline can be aware of this event but it is not consensus critical. Adding this event now makes the contract more standard, as setting values in it will always emit an event. The derivation pipeline will halt on unknown events, so it is updated with code that parses the event as a no-op. Also add some go test code for the new event.

This event must not be triggered until all nodes on the network upgrade. This means that we cannot alter the resource config until we are confident that the network has fully upgraded, or else it will cause a chain split.

This event is useful for the future to allow for the 1559 curve to be computed in the `op-node` instead of on chain. This would greatly simplify the L1 deposit code path if this change was pursued.

**TODOs**


tynes:
Don't merge this, going to remove the new event but leave in the test

PR #5253, opened 2023-03-24
contracts-bedrock: move L2oo init to step1 of deploy

tynes:
**Description**

This allows the full sequencer to start up after `step1`. The dynamic config check needs to be moved to ensure that the config is set up correctly.

This PR should get close review from @maurelian and @smartcontracts before merging to ensure that it is safe.

Closes CLI-3435
Closes CLI-3670

### DIFF
--- a/packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
+++ b/packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
@@ -254,6 +254,22 @@ contract SystemDictator is OwnableUpgradeable {
                 )
             )
         );
+
+        // Dynamic config must be set before we can initialize the L2OutputOracle.
+        require(dynamicConfigSet, "SystemDictator: dynamic oracle config is not yet initialized");
+
+        // Upgrade and initialize the L2OutputOracle.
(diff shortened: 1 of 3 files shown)


packages/contracts-bedrock/contracts/deployment/SystemDictator.sol
```diff
@@ -254,6 +254,22 @@ contract SystemDictator is OwnableUpgradeable {
...
+            payable(config.proxyAddressConfig.l2OutputOracleProxy),
+            address(config.implementationAddressConfig.l2OutputOracleImpl),
+            abi.encodeCall(
+                L2OutputOracle.initialize,
+                (
+                    l2OutputOracleDynamicConfig.l2OutputOracleStartingBlockNumber,
+                    l2OutputOracleDynamicConfig.l2OutputOracleStartingTimestamp
+                )
+            )
+        );
```
maurelian, line 272: Actually, I don't think we can do this until after step2() when we set the DTL shut off block.

Perhaps we could make it something we do before `step3()`.

maurelian:
I think we can close this. Will handle it in another PR on top of some more recent MSD changes.

PR #5258, opened 2023-03-24
txmgr: Initialize signer and more in the tx manager proper

trianglesphere:
**Description**

This simplifies how the transaction manager is initialized.

Note: I am connecting to the L1 RPC twice instead of re-using a shared client. Right now we don't do anything where we might want this (like global rate limiting / retries), but we may need to change this in the future. I am relying on re-using a RPC flag for the transaction manager, but it works for now.

Some of the setup is a little tricky because we initialize the batcher/proposer from code partway through the initialization process rather than always from `main`.

This also moves some of the functions around inside the tx manager (but does not change their behavior).

Flag Changes
- `l2-output-hd-path` and `sequencer-hd-path` -> `hd-path`. **Breaking Change**

**TODOs**

- [ ] Manually verify that the flags are the same.


op-service/txmgr/txmgr.go
```diff
@@ -111,14 +72,36 @@ type ETHBackend interface {
...
 
 	backend ETHBackend
 	l       log.Logger
 }
 
+// NewSimpleTxManager initializes a new SimpleTxManager with the passed Config.
+func NewSimpleTxManager(name string, l log.Logger, cfg Config) *SimpleTxManager {
+	if cfg.NumConfirmations == 0 {
+		panic("txmgr: NumConfirmations cannot be zero")
+	}
```
sebastianst, line 87: Maybe also add this check already to `CLIConfig.Check` so it's caught earlier.

op-service/txmgr/txmgr.go
```diff
@@ -111,14 +72,36 @@ type ETHBackend interface {
...
+	if cfg.NetworkTimeout == 0 {
+		cfg.NetworkTimeout = 2 * time.Second
+	}
```
sebastianst, line 90: Maybe move into new method `Config.PopulateDefaults()`.

op-service/txmgr/cli.go
```diff
@@ -0,0 +1,175 @@
...
+		ResubmissionTimeout:       ctx.GlobalDuration(ResubmissionTimeoutFlagName),
+	}
+}
+
+func NewConfig(cfg CLIConfig, l log.Logger) (Config, error) {
+	if err := cfg.Check(); err != nil {
+		return Config{}, err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
sebastianst, line 108:
```suggestion
	networkTimeout := 2 * time.Second

	ctx, cancel := context.WithTimeout(context.Background(), networkTimeout)
```
I'd set once and then reuse (3 times). 

When it is added to the `CLIConfig` in a follow-up, can then just be read from there.

op-service/txmgr/cli.go
```diff
@@ -0,0 +1,175 @@
...
+	defer cancel()
+	l1, err := ethclient.DialContext(ctx, cfg.L1RPCURL)
+	if err != nil {
+		return Config{}, err
+	}
+
+	ctx, cancel = context.WithTimeout(context.Background(), 2*time.Second)
```
sebastianst, line 115:
```suggestion
	ctx, cancel = context.WithTimeout(context.Background(), networkTimeout)
```

op-service/txmgr/cli.go
```diff
@@ -0,0 +1,175 @@
...
+	signerFactory, from, err := opcrypto.SignerFactoryFromConfig(l, cfg.PrivateKey, cfg.Mnemonic, cfg.SequencerHDPath, cfg.SignerCLIConfig)
+	if err != nil {
+		return Config{}, err
+	}
+
+	return Config{
+		Backend:                   l1,
+		ResubmissionTimeout:       cfg.ResubmissionTimeout,
+		ChainID:                   chainID,
+		NetworkTimeout:            2 * time.Second,
```
sebastianst, line 131:
```suggestion
		NetworkTimeout:            networkTimeout,
```

op-service/txmgr/txmgr.go
```diff
@@ -111,14 +72,36 @@ type ETHBackend interface {
...
+
+	return &SimpleTxManager{
+		chainID: cfg.ChainID,
+		name:    name,
+		cfg:     cfg,
+		backend: cfg.Backend,
+		l:       l.New("service", name),
```
sebastianst, line 97: So this shouldn't be something like `"tx-mgr"`?

sebastianst (changes requested):
Love the new encapsulation, how L1 write interaction is now invisible to the proposer & batcher. And another net negative diff 🎉 

Shouldn't we also update the tx manager env flags in `ops-bedrock/docker-compose.yml`? And possibly other places as well, not sure.

A good follow up would be to also add (optional) flags for the `NetworkTimeout` and `ReceiptQueryInterval` to the tx mgr cli config.

op-service/txmgr/txmgr.go
trianglesphere: Like `txmgr-for`?. This is to indicate which service is using the txmgr. We don't really need this when running the full system because we can sort logs by component. We kinda use this in op-e2e, but use a `role` to differentiate the different services: https://github.com/ethereum-optimism/optimism/blob/18b5f8ac35dd5a53e98a393c0ab1448c5b0d6267/op-e2e/setup.go#L148-L153

op-service/txmgr/txmgr.go
trianglesphere: done

op-service/txmgr/txmgr.go
sebastianst: Ah sorry, I didn't get that it's the key name 😅

sebastianst (approved):

PR #5260, opened 2023-03-24
feat(monorepo): `yarn` -> `pnpm` migration

clabby:
# Overview

> **Note**
> Very broken, WIP.

Migrates the monorepo from `yarn` -> `pnpm`.


clabby:
* develop
  * **PR #5260** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5260" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5260?utm_source=stack-comment).

packages/hardhat-deploy-config/README.md
```diff
@@ -5,5 +5,5 @@
 ## Installation
 
 ```
-yarn add @eth-optimism/hardhat-deploy-config
+pnpm add @eth-optimism/hardhat-deploy-config
```
roninjin10, line 8:
```suggestion
pnpm i @eth-optimism/hardhat-deploy-config
```

roninjin10:
If we are using resolutions or hoist in our package.json those no longer are doing anything and you can delete.

You can delete the workspaces key from package.json

Here is dangerously hoist docs if you need it https://pnpm.io/npmrc#shamefully-hoist (though hoisting stuff that we previously said no hoist will break things)

.circleci/config.yml
```diff
@@ -67,15 +67,15 @@ jobs:
       - check-changed:
           patterns: op-bindings,op-chain-ops,packages/
       - restore_cache:
-          name: Restore Yarn Package Cache
+          name: Restore PNPM Package Cache
           keys:
-            - yarn-packages-v2-{{ checksum "yarn.lock" }}
+            - pnpm-packages-v2-{{ checksum "pnpm-lock.yaml" }}
```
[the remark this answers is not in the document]
clabby: Should be installed through the ci-builder though, no?

.npmrc
```diff
@@ -0,0 +1,14 @@
...
+hoist-pattern[]=!@eth-optimism/contracts-bedrock/@rari-capital/solmate
+hoist-pattern[]=!@openzeppelin
+hoist-pattern[]=!@eth-optimism/contracts-periphery/ds-test
+hoist-pattern[]=!@eth-optimism/contracts-periphery/forge-std
+hoist-pattern[]=!@eth-optimism/contracts-periphery/@rari-capital/solmate
+hoist-pattern[]=!forta-agent
+
+# run prefoo scripts
+# npm and yarn support this as a default pnpm defaults to false
+enable-pre-post-scripts=true
```
roninjin10, line 14: We should remove this behavior [in favor of nx](https://github.com/ethereum-optimism/optimism/pull/4778) controlling the order that things need to be built

.npmrc
```diff
@@ -0,0 +1,14 @@
+hoist-pattern[]=!**/typechain/*
+hoist-pattern[]=!**/@typechain/*
+hoist-pattern[]=!@eth-optimism/contracts-bedrock/ds-test
+hoist-pattern[]=!@eth-optimism/contracts-bedrock/forge-std
+hoist-pattern[]=!@eth-optimism/contracts-bedrock/@rari-capital/solmate
+hoist-pattern[]=!@openzeppelin
+hoist-pattern[]=!@eth-optimism/contracts-periphery/ds-test
+hoist-pattern[]=!@eth-optimism/contracts-periphery/forge-std
+hoist-pattern[]=!@eth-optimism/contracts-periphery/@rari-capital/solmate
+hoist-pattern[]=!forta-agent
```
roninjin10, line 10:
```suggestion
``` 
Haven't tested yet but these are all unnecessary.  By default, pnpm never hoists.   This makes a lot of issues I've seen @maurelian and @smartcontracts battle with from time to time go away.

CONTRIBUTING.md
```diff
@@ -43,7 +43,7 @@ You *must* include a `changeset` file in your PR when making a change that would
 Adding a `changeset` file is easy:
 
 1. Navigate to the root of the monorepo.
-2. Run `yarn changeset`. You'll be prompted to select packages to include in the changeset. Use the arrow keys to move the cursor up and down, hit the `spacebar` to select a package, and hit `enter` to confirm your selection. Select *all* packages that require a new release as a result of your PR.
+2. Run `pnpm changeset`. You'll be prompted to select packages to include in the changeset. Use the arrow keys to move the cursor up and down, hit the `spacebar` to select a package, and hit `enter` to confirm your selection. Select *all* packages that require a new release as a result of your PR.
```
roninjin10, line 46: Unrelated to this pr but this is too much detail.  Once you run pnpm changeset you can figure out the rest because the prompt guides you along

integration-tests/package.json
```diff
@@ -4,8 +4,8 @@
   "version": "0.5.22",
   "description": "[Optimism] Integration tests",
   "scripts": {
-    "lint": "yarn lint:fix && yarn lint:check",
-    "lint:fix": "yarn lint:check --fix",
+    "lint": "pnpm lint:fix && pnpm lint:check",
```
roninjin10, line 7: Unrelated to this pr but this is not a normal way of doing things.   CI should be using lint:check and this has no reason to be running the linter twice

```suggestion
    "lint": "pnpm lint:fix",
```

Don't commit this as this is out of scope for pr

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
 RUN apk add coreutils
 
 WORKDIR /tmp
-COPY yarn.lock .nvmrc package.json ./src/
+COPY pnpm-lock.yaml pnpm-workspace.yaml .nvmrc package.json ./src/
```
roninjin10, line 20: @clabby I added pnpm-workspace.yaml here

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
...
   # .nvmrc
   cp .nvmrc ../manifests/
 
 FROM ethereumoptimism/foundry:latest as foundry
-FROM node:16-alpine3.14 as base
+# bullseye-slim is debian based
+# we use it rather than alpien because it's not much
+# bigger and alpine is often missing packages for node applications
+# alpine is not officially supported by node.js
+FROM node:16.16.0-bullseye-slim as base
```
roninjin10, line 38: This should be in it's own PR and reviewed by zach and any security folks.   This is a pretty big change to use debian instead of alpine.

From my experience and also gathering the wisdom of the crowd building js docker images, because node.js doesn't support alpine officially and doesn't recomend using it, and because I and many others have struggled in past mantaining alpine builds, it's not worth using alpine. 

Negative tradeoff

- bullseye-slim is less minimal than alpine so slightly bigger surface area for security vulnerabilities.   It's an extremely common and battle tested image though
- bullseye-slim is bigger.  It's not significantly bigger though and that is offset a bit by doing a multistage build

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
...
 
 # Base: install deps
-RUN apk --no-cache add curl \
+RUN apt-get update && apt-get install -y \
+  curl \
   jq \
   python3 \
   ca-certificates \
   git \
+  g++ \
```
roninjin10, line 47: This is needed to compile c++ when running pnpm install

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
...
   make \
   gcc \
+  pkg-config \
```
roninjin10, line 50:
```suggestion
```
TODO I accidentally added this twice

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
...
   musl-dev \
-  linux-headers \
   bash \
-  build-base \
-  gcompat
```
roninjin10, line 46: These 3 deps I deleted are needed on the ultra minimal alpine image but bullseye-slim has support for this functionality already.    These are all needed to create compatibility between alpine and c++ compiler on alpine but not debian

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
...
+  # the following 4 deps are needed for node-hid
+  # which is a deep sub dependency of ethers to install
+  # correctly
+  pkg-config \
+  libusb-1.0-0-dev \
+  libudev-dev \
+  --no-install-recommends
```
roninjin10, line 59: a very problematic dependency in my experience is node-hid which is a deep sub dependency of ethers.    Here these installations simply allow some postinstall scripts to run correctly.  yarn would allow these to fail silently

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
...
 
-ENV GLIBC_KEY=https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub
-ENV GLIBC_KEY_FILE=/etc/apk/keys/sgerrand.rsa.pub
-ENV GLIBC_RELEASE=https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.35-r0/glibc-2.35-r0.apk
-
-RUN wget -q -O ${GLIBC_KEY_FILE} ${GLIBC_KEY} \
-  && wget -O glibc.apk ${GLIBC_RELEASE} \
-  && apk add glibc.apk --force
```
roninjin10, line 54: We don't need to do this anymore because it's debian and not alpine and this works by default

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
...
+RUN npm install pnpm --global
```
roninjin10, line 61: @clabby this is why the docker build wasn't working

ops/docker/Dockerfile.packages
```diff
@@ -17,68 +17,75 @@ FROM alpine:3.16 as manifests
...
 
 # Copy manifest files into the image in
-# preparation for `yarn install`.
+# preparation for `pnpm install`.
 COPY integration-tests/package.json ./integration-tests/package.json
 COPY --from=manifests /tmp/manifests  ./
 COPY *.json ./
 
-RUN yarn install --frozen-lockfile && yarn cache clean
+RUN pnpm install
```
roninjin10, line 74:
```suggestion
RUN pnpm install --frozen-lockfile
```
TODO I messed this up this should still frozen lockfile

ops/docker/ci-builder/Dockerfile
```diff
@@ -49,14 +49,19 @@ RUN apt-get update && \
...
   go install gotest.tools/gotestsum@latest && \
   curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.48.0 && \
   curl -fLSs https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh | bash && \
   chmod +x /usr/local/bin/check-changed
 
+RUN echo "downloading pnpm" && \
+  curl -o /pnpm-install.sh -L https://get.pnpm.io/install.sh && \
+  chmod +x /pnpm-install.sh && \
+  /pnpm-install.sh && \
+  rm /pnpm-install.sh
```
roninjin10, line 63: Could just `npm i -g pnpm` instead

ops/docker/hardhat/Dockerfile
```diff
@@ -4,11 +4,17 @@ FROM node:16-alpine
...
 # install deps
 WORKDIR /hardhat
-RUN yarn install && yarn cache clean
+
+RUN npm install pnpm --global
+# Note this is not using the pnpm lockfile but rather installing from scratch
+# this is not deterministic and the versions being installed here are not guaranteed to match
+# the versions in the lockfile.
+# be aware of this when debugging issues
+RUN pnpm install
```
roninjin10, line 17: Note we don't use frozen lockfile because there is no lockfile

package.json
```diff
@@ -3,54 +3,25 @@
...
-      "**/@openzeppelin/*",
-      "@eth-optimism/contracts-periphery/ds-test",
-      "@eth-optimism/contracts-periphery/forge-std",
-      "@eth-optimism/contracts-periphery/@rari-capital/solmate",
-      "forta-agent"
-    ]
-  },
   "private": true,
+  "engines": {
+    "pnpm": ">=7"
```
roninjin10, line 8: would be a good idea to also add

"node": ">16",

but that is out of scope for this pr

roninjin10: this will make it so if user uses npm or yarn, everything will work fine but they will get a warning in console that pnpm is the correct engine to use

package.json
```diff
@@ -3,54 +3,25 @@
...
+    "clean": "lerna run clean --parallel",
+    "build": "lerna run build",
+    "test": "lerna run test --parallel",
+    "test:coverage": "lerna run test:coverage --parallel",
+    "lint:ts:check": "lerna run lint:ts:check",
+    "lint:check": "lerna run lint:check",
+    "lint:fix": "lerna run lint:fix --parallel",
+    "lint:specs:fix": "markdownlint-cli2-fix \"./specs/**/*.md\"",
+    "lint:specs:check": "markdownlint-cli2  \"./specs/**/*.md\"",
+    "lint:specs:toc": "doctoc '--title=**Table of Contents**' ./specs",
```
roninjin10, line 20: Could nuke all this for pnpm

`pnpm concurrently run clean --parallel`

I wouldn't bother though because we really should use nx.  There is a pr that is pretty much ready to merge now https://github.com/ethereum-optimism/optimism/pull/4778

package.json
```diff
@@ -76,7 +47,7 @@
     "eslint-plugin-react": "^7.24.0",
     "eslint-plugin-unicorn": "^42.0.0",
     "husky": "^6.0.0",
-    "lerna": "^4.0.0",
+    "lerna": "^6.0.0",
```
roninjin10, line 50: Upgrade lerna to version that supports pnpm

package.json
```diff
@@ -3,54 +3,25 @@
...
+    "lint:fix": "lerna run lint:fix --parallel",
+    "lint:specs:fix": "markdownlint-cli2-fix \"./specs/**/*.md\"",
+    "lint:specs:check": "markdownlint-cli2  \"./specs/**/*.md\"",
+    "lint:specs:toc": "doctoc '--title=**Table of Contents**' ./specs",
     "postinstall": "patch-package",
-    "ready": "yarn lint && yarn test",
+    "ready": "pnpm lint && pnpm test",
     "prepare": "husky install",
-    "release": "yarn build && yarn changeset publish",
-    "update:yarn": "yarn set version 1.x"
```
roninjin10, line 53: This script was used to update yarn version in .yarn/releases

packages/atst/package.json
```diff
@@ -26,13 +26,13 @@
...
+    "build": "tsup",
     "generate": "wagmi generate",
-    "lint": "yarn lint:fix && yarn lint:check",
+    "lint": "pnpm lint:fix && pnpm lint:check",
     "lint:check": "eslint . --max-warnings=0",
-    "lint:fix": "yarn lint:check --fix",
+    "lint:fix": "pnpm lint:check --fix",
     "test": "vitest",
-    "typecheck": "yarn tsc --noEmit"
+    "typecheck": "tsc --noEmit"
```
roninjin10, line 35: in pnpm (and yarn and npm) you don't need to type yarn or pnpm first to run a node_module script

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ orbs:
   gcp-cli: circleci/gcp-cli@3.0.1
 commands:
   gcp-oidc-authenticate:
-    description: "Authenticate with GCP using a CircleCI OIDC token."
+    description: 'Authenticate with GCP using a CircleCI OIDC token.'
```
roninjin10, line 8: TODO my prettier/linter changed all these automatically

ops/docker/Dockerfile.packages
roninjin10: Could likely get it to work on alpine but it would take a long time to figure out or a hack (simple hack is not running postinstall scripts at all which would work but is a hack and could cause hard to debug issues in future)

roninjin10: All node.js images should use a debian image by defaut.   If we needed an image as tiny and minimal as alpine, there are better alternatives such as copying into [gcr.io/distroless/nodejs18](https://github.com/roninjin10/stax/blob/main/Dockerfile#L242)

.npmrc
maurelian: I didn't even know this file existed.

roninjin10: It didn't exist until this pr.   What I meant was the behavior pnpm has of never hoisting by default is what fixes a lot of pain you and Kelvin went through.  Anytime a type stopped working randomly or some dependency changed out from under you part of what caused that con fusing behavior was how yarn/npm hoists.

.npmrc
```diff
@@ -0,0 +1,4 @@
+
```
roninjin10, line 1:
```suggestion
```

.npmrc
roninjin10: My nx pr removes all the prebuild pretest prelint ect. in favor of nx

packages/contracts-bedrock/package.json
```diff
@@ -67,7 +67,7 @@
     "@foundry-rs/easy-foundryup": "^0.1.3",
     "@nomiclabs/hardhat-ethers": "^2.0.0",
     "@nomiclabs/hardhat-waffle": "^2.0.0",
-    "@rari-capital/solmate": "https://github.com/rari-capital/solmate.git#8f9b23f8838670afda0fd8983f2c41e8037ae6bc",
+    "@rari-capital/solmate": "https://github.com/transmissions11/solmate.git#8f9b23f8838670afda0fd8983f2c41e8037ae6bc",
```
roninjin10, line 70: Here I changed rari-capital to the repo that rari-capital redirects to if you go to https://github.com/rari-capital/solmate.git#8f9b23f8838670afda0fd8983f2c41e8037ae6bc .  This was causing an issue with pnpm correctly cloning submodules within solmate

.npmrc
roninjin10: Not shamefully hoisting here means we need to test anything that doesn't have test coverage

s7v7nislands:
@clabby any plan for this PR? Can I make a new PR to fix `lerna` too old, the node-gyp module can't compile on python 3.11?

trianglesphere:
> Can I make a new PR to fix `lerna` too old, the node-gyp module can't compile on python 3.11?

Go for it. I think we likely want to do this migration, but it is not a top priority right now. This will also take a little bit of time b/c we don't want to break devs if/when we do this migration.

PR #5261, opened 2023-03-25
fix(devnet): Updated the L2 address to 9545

qbzzt:
That's the correct value, not 8546 (which is, IIRC, websocket for L1).


ajsutton (approved):
LGTM. Confirmed l2 geth port 8545 is exposed to the host as port 9545 so that's the port it should be accessed on.

PR #5265, opened 2023-03-27
feat(contracts-periphery): create OptimistInviter contract

jakim929:
**context & discussion**
https://www.notion.so/oplabs/Invite-to-Mint-Optimist-NFT-c2bf38e31d264de387268a7896760dd7?pvs=4#08e356a3654c44e5a1e1a58448453e22

OptimistInviter will be used by OptimistAllowlist (to be added) to add invite functionality to the Optimist NFT. The OptimistInviter contract issues attestations to the AttestationStation, and includes a commit & reveal scheme to allow invite recipients to claim the invite to any wallet.

Our multi-sig (INVITE_GRANTER) will have the ability to give people invites that they can distribute.

- Added OptimistInviter contract
- Added tests


packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+     *
+     * @param issuer Address that issued the signature.
+     * @param claimer Address that claimed the invite.
+     */
+    event InviteClaimed(address indexed issuer, address indexed claimer);
+
+    /**
+     * @notice Version identifier, used for upgrades.
+     */
+    uint8 public constant VERSION = 1;
```
roninjin10, line 33: this feels bad that we have this and we also have semver.  There must be a better way

roninjin10 (commented):
Nice only skimmed it can give it a real review later

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
maurelian: Agree. James brought this up with me as well. For now we can perhaps just sidestep the matter by using `initializer` (without a version) instead of `reinitalizer`.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * @notice Allows anyone to commit a received signature along with the address to claim to.
+     *         This is necessary to prevent front-running when the invitee is claiming the invite.
+     *
+     * @param _commitment A hash of the claimer and signature concatenated.
+                          keccak256(abi.encode(_claimer, _signature))
+     */
+    function commitInvite(bytes32 _commitment) public {
```
maurelian, line 135: Is this committing to an invite or a claim?

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+            }
+        }
+    }
+
+    /**
+     * @notice Allows anyone to commit a received signature along with the address to claim to.
+     *         This is necessary to prevent front-running when the invitee is claiming the invite.
+     *
+     * @param _commitment A hash of the claimer and signature concatenated.
+                          keccak256(abi.encode(_claimer, _signature))
```
maurelian, line 133:
```suggestion
    *                     keccak256(abi.encode(_claimer, _signature))
```

maurelian: Surprised we have no validation of natspec formatting.

packages/contracts-periphery/contracts/foundry-tests/OptimistInviter.t.sol
```diff
@@ -0,0 +1,689 @@
...
+
+        carolERC1271Wallet = new TestERC1271Wallet(carol);
+
+        // Give alice and bob and sally some ETH
+        vm.deal(alice_inviteGranter, 1 ether);
+        vm.deal(bob, 1 ether);
+        vm.deal(sally, 1 ether);
+        vm.deal(eve, 1 ether);
+
+        vm.label(alice_inviteGranter, "alice_inviteGranter");
```
tynes, line 60: there is a helper function in forge-std `makeAddr` that will create an address and call `vm.label`, could be helpful to begin migrating to that

packages/contracts-periphery/contracts/foundry-tests/OptimistInviter.t.sol
```diff
@@ -0,0 +1,689 @@
...
+
+        _initializeContracts();
+    }
+
+    /**
+     * @notice Instantiates an AttestationStation, and an OptimistInviter.
+     */
+    function _initializeContracts() internal {
+        attestationStation = new AttestationStation();
+        vm.expectEmit(false, false, false, false);
```
tynes, line 78: nit: should always pass in `true` for all args as well as a 5th arg with the expected account emitting the event. Not sure if they merged in the overload that takes only the account, this variant defaults to all trues

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+
+        uint256 length = _accounts.length;
+
+        for (uint256 i; i < length; ) {
+            // The granted invites are stored as an attestation from this contract on the
+            // AttestationStation contract. Number of invites is stored as a encoded uint256 in the
+            // data field of the attetation.
+            ATTESTATION_STATION.attest(
+                _accounts[i],
+                bytes32("optimist.can-invite"),
```
tynes, line 118: `bytes32("optimist.can-invite")` should be a constant

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+            "OptimistInviter: only invite granter can grant invites"
+        );
+
+        uint256 length = _accounts.length;
+
+        for (uint256 i; i < length; ) {
+            // The granted invites are stored as an attestation from this contract on the
+            // AttestationStation contract. Number of invites is stored as a encoded uint256 in the
+            // data field of the attetation.
+            ATTESTATION_STATION.attest(
```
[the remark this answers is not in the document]
tynes: Agree on this, then we make a single call instead of `n` calls

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+                _accounts[i],
+                bytes32("optimist.can-invite"),
+                abi.encode(_inviteCount)
+            );
+
+            unchecked {
+                i++;
```
tynes, line 123: technically `++i` saves more gas if you are trying to optimize

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+        ATTESTATION_STATION = _attestationStation;
+    }
+
+    /**
+     * @notice Initializes the OptimistInviter contract, setting the EIP712 context.
+     *
+     * @param _name Contract name
+     */
+    function initialize(string memory _name) public reinitializer(VERSION) {
+        __EIP712_init(_name, version());
```
tynes, line 94: Not sure if we should tie the semver version to the eip712 version. Trying to think of the impact of this

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: It's supposed to be the invite recipient (who received the signature), committing to the hash of their own address and the signature they received.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+            ),
+            "OptimistInviter: invalid signature"
+        );
+
+        // The issuer includes a pseudorandom nonce in the signature to prevent replay attacks.
+        // This checks that the nonce has not been used for this issuer before. The nonces are
+        // scoped to the issuer address, so the same nonce can be used by different issuers without
+        // clashing.
+        require(
+            !usedNonces[_claimableInvite.issuer][_claimableInvite.nonce],
```
tynes, line 188: slit style nit: i find `== false` easier to read than `! ...`
the less cog overhead for reviewers when writing smart contracts, the better

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+        // The invite issuer is included in the data of the attestation.
+        ATTESTATION_STATION.attest(
+            _claimer,
+            bytes32("optimist.can-mint-from-invite"),
+            abi.encode(_claimableInvite.issuer)
+        );
+
+        // Reduce the issuer's invite count by 1 by re-attesting the optimist.can-invite attestation
+        // with the new count.
+        count--;
```
tynes, line 219: This  could technically be wrapped in an unchecked block because of the count check above

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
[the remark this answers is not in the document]
jakim929: This uses slightly more gas than the current approach. Was wondering why is making a single call preferable?

```
function setInviteCounts(address[] calldata _accounts, uint256 _inviteCount) public {
        // Only invite granter can grant invites
        require(
            msg.sender == INVITE_GRANTER,
            "OptimistInviter: only invite granter can grant invites"
        );

        uint256 length = _accounts.length;

        AttestationStation.AttestationData[]
            memory attestations = new AttestationStation.AttestationData[](length);

        for (uint256 i; i < length; ) {
            attestations[i] = AttestationStation.AttestationData({
                about: _accounts[i],
                key: bytes32("optimist.can-invite"),
                val: abi.encode(_inviteCount)
            });

            unchecked {
                ++i;
            }
        }

        // The granted invites are stored as an attestation from this contract on the
        // AttestationStation contract. Number of invites is stored as a encoded uint256 in the
        // data field of the attestation.
        ATTESTATION_STATION.attest(attestations);
    }
```

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+                )
+            )
+        );
+
+        // Uses SignatureChecker, which supports both regular ECDSA signatures from EOAs as well as
+        // ERC-1271 signatures from contract wallets or multi-sigs. This means that if the issuer
+        // wants to revoke a signature, they can use a smart contract wallet to issue the signature,
+        // then invalidate the signature after issuing it.
+        require(
+            SignatureCheckerUpgradeable.isValidSignatureNow(
```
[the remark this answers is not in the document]
jakim929: It doesn't have any state it's using. I'll just use the SignatureChecker instead to avoid confusion.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,228 @@
...
+                _claimableInvite.issuer,
+                digest,
+                _signature
+            ),
+            "OptimistInviter: invalid signature"
+        );
+
+        // The issuer includes a pseudorandom nonce in the signature to prevent replay attacks.
```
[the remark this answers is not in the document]
jakim929: Yep! I'll update it to make it clearer.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: Will fix!

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: Updated to use `initializer` instead `reinitializer` and removed the VERSION constant.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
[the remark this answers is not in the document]
maurelian: Looks like the cost is due to memory expansion while re-encoding. 
I think you could do the original encoding off-chain and just change the arguments to accept an array of attestations.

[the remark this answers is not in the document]
roninjin10: +1 makes ux of using this contract worse but likely worth just dealing with raw bytes

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,234 @@
...
+        __EIP712_init(_name, version());
+    }
+
+    /**
+     * @notice Allows invite granter to set the number of invites an address has.
+     *
+     * @param _accounts    An array of accounts to update the invite counts of.
+     * @param _inviteCount Number of invites to set to.
+     */
+    function setInviteCounts(address[] calldata _accounts, uint256 _inviteCount) public {
```
roninjin10, line 110: I would just name this attest and same api as attestation station

packages/contracts-periphery/contracts/foundry-tests/OptimistInviter.t.sol
```diff
@@ -0,0 +1,691 @@
...
+        _commitInviteAs(sally, signature);
+
+        vm.prank(ted);
+        optimistInviter.commitInvite(keccak256(abi.encode(sally, signature)));
+
+        vm.expectEmit(true, true, true, true, address(attestationStation));
+        emit AttestationCreated(
+            address(optimistInviter),
+            sally,
+            bytes32("optimist.can-mint-from-invite"),
```
tynes, line 441: would prefer to make a call to the contract to get the bytes32 topic values rather than hardcoding them because it makes it more portable

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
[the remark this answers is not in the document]
tynes: I believe you could prevent unecessary allocation of memory by copying the free memory pointer at the top of the loop body and then resetting the free memory pointer at the bottom of the loop body to the prev value. Its a bit of a hack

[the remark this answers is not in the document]
tynes: to test gas usage correctly, you need to put the functions in different contracts so that the gas used in function dispatch is controlled for. if there are multiple functions in the same contract the amount of gas to dispatch to them may be different

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: Added descriptions overall, at the top of the contract as well as on this function.

packages/contracts-periphery/contracts/foundry-tests/OptimistInviter.t.sol
```diff
@@ -0,0 +1,689 @@
+//SPDX-License-Identifier: MIT
```
roninjin10, line 1: clean tests nice job!

packages/contracts-periphery/contracts/testing/helpers/TestERC1271Wallet.sol
```diff
@@ -0,0 +1,28 @@
...
+import { IERC1271 } from "@openzeppelin/contracts/interfaces/IERC1271.sol";
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+import { ECDSA } from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+// solhint-disable max-line-length
+/**
+ * Simple ERC1271 wallet that can be used to test the ERC1271 signature checker.
+ * https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/mocks/ERC1271WalletMock.sol
+ */
+
```
roninjin10, line 13:
```suggestion
```

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
roninjin10: Steel man for special implementation:
- no need to write js code turning invites to attestations
- No worry about a junky Attestation causing issues later with say an indexer
- Better etherscan ui
- Self documenting easier to understand how to use contract

Steel man for polymorphism
- all tooling like atst sdk just plugs right in
- Less lines of solidity code
- Slightly less gas
- Simpler to understand just following a simple proxy design pattern
- maximally flexible.   You can issue different invite counts per account for example

I’m partial for the second because in general everything is much easier the further up the stack you go if things have polymorphism with their interfaces and it is more flexible

roninjin10 (commented):
looks great so far

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: While I’m aligned for the need for a more general AttestationStation proxy contract, I’m leaning towards keeping the current implementation, and keep the scope of this contract as minimal as possible.
My main point of concern is just that I don’t think we should make this contract a full proxy for the multi-sig, that allows the owner to make any arbitrary attestations through this contract.
This is specifically a contract that is meant to handle issuing invites, and the goal wasn’t to make a general purpose ATST proxy our multi-sig. I feel that having an attest function that allows any type of attestation to be made will hurt readability of the contract.
Having this contract only be concerned with issuing a very limited set of attestations, “optimist.can-invite” and “optimist.can-mint-from-invite” makes it clear how this contract is meant to be used, and what this contract should or shouldn’t be able to do.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: Discussed with Mark:

1. Every time we upgrade the contract, we should call initialize again to bump the EIP712 version (by calling __EIP712_init again). The EIP712 domain separator hash is stored in storage by EIP712Upgradeable.
2. Upgrading the version will cause all signatures that haven't been claimed yet to be invalid - we should be aware of this when upgrading the version, and communicate this on product side if necessary.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,234 @@
...
+     *         bytes32("optimist.can-invite")
+     */
+    bytes32 public constant CAN_INVITE_ATTESTATION_KEY =
+        0x6f7074696d6973742e63616e2d696e7669746500000000000000000000000000;
+
+    /**
+     * @notice Attestation key allowing the attested account to mint.
+     *         bytes32("optimist.can-mint-from-invite")
+     */
+    bytes32 public constant CAN_MINT_FROM_INVITE_ATTESTATION_KEY =
```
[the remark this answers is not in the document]
jakim929: I have another PR coming up for the OptimistNFT, so I'll just make sure to include it then!

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,274 @@
...
+     *         The hash, keccak256(abi.encode(_claimer, _signature)), should have been already
+     *         committed using commitInvite. Before issuing the "optimist.can-mint-from-invite"
+     *         attestation, this function checks that
+     *           1) the hash corresponding to the _claimer and the _signature was committed
+     *           2) the _signature is signed correctly by the issuer
+     *           3) the _signature hasn't already been used to claim an invite before
+     *           4) the _signature issuer has not used up all of their invites
+     *         This function doesn't require that the _claimer is calling this function.
+     *
+     * @param _claimer Address that will be granted the invite.
```
smartcontracts, line 191: Nit: param whitespace is off

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,274 @@
...
+        usedNonces[_claimableInvite.issuer][_claimableInvite.nonce] = true;
+
+        // Check the AttestationStation contract to see how many invites the issuer has left.
+        bytes memory attestation = ATTESTATION_STATION.attestations(
+            address(this),
+            _claimableInvite.issuer,
+            CAN_INVITE_ATTESTATION_KEY
+        );
+        // Failing this check means that the issuer was never granted any invites to begin with.
+        require(attestation.length > 0, "OptimistInviter: issuer has no invites");
```
smartcontracts, line 245: Worth differentiating the error messages to make things easier to debug if necessary

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,274 @@
...
+        uint256 length = _accounts.length;
+
+        AttestationStation.AttestationData[]
+            memory attestations = new AttestationStation.AttestationData[](length);
+
+        for (uint256 i; i < length; ) {
+            // The granted invites are stored as an attestation from this contract on the
+            // AttestationStation contract. Number of invites is stored as a encoded uint256 in the
+            // data field of the attestation.
+            attestations[i] = AttestationStation.AttestationData({
```
smartcontracts, line 146: Unsure how I feel about using the ATST to store invite counts. No real benefit over storing internally. If anything, could be worth giving a single attestation representing the fact that the user *did* have invites but not bothering to use it to store state.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,274 @@
...
+        require(attestation.length > 0, "OptimistInviter: issuer has no invites");
+
+        uint256 count = abi.decode(attestation, (uint256));
+
+        // Failing this check means that the issuer has used up all of their existing invites.
+        require(count > 0, "OptimistInviter: issuer has no invites");
+
+        // Create the attestation that the claimer can mint from the issuer's invite.
+        // The invite issuer is included in the data of the attestation.
+        ATTESTATION_STATION.attest(
```
smartcontracts, line 254: Although it's technically fine here, I would reduce the count first and then do the external call out of habit because you never know when you'll accidentally introduce a reentrancy bug like this. If the ATST did an external call during this `attest` it could be used to reenter and do multiple transactions without reducing the invite count.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * @notice EIP712 typehash for the ClaimableInvite type.
+     */
+    bytes32 public constant CLAIMABLE_INVITE_TYPEHASH =
+        keccak256("ClaimableInvite(address issuer,bytes32 nonce)");
+
+    /**
+     * @notice Attestation key for granting invites.
+     */
+    bytes32 public constant CAN_INVITE_ATTESTATION_KEY = bytes32("optimist.can-invite");
```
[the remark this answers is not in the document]
jakim929: good point- For this contract it's a quick change.

But we already issued some attestations using non-hashed keys for the mint attestation for the [OptimistNFT contract](https://github.com/ethereum-optimism/optimism/blob/91f81d9976a1c0af27fb40a9d4f56c3315791808/packages/contracts-periphery/contracts/universal/op-nft/Optimist.sol#L114-L119). Prob too late to change, because we've already issued them, but going forward the policy should be to use hashed keys whenever possible.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: Since there's no explicit benefit of tying the semver to the EIP712 version, and separating them gives us more flexibility in upgrading the contract, I changed the EIP712 version to be stored separately so it can be updated independently when there's a change to the signature scheme.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: Updated the contract to store a separate mapping of invite counts. The contract still issues a single attestation to show that the user was allowed to invite others, but the contract no longer depends on the data field to handle the invite count.

jakim929:
I fixed up the requested changes - would appreciate a final pass / approval!

roninjin10 (approved):
lgtm

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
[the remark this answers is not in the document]
roninjin10: The atst sdk and our documentation both hash if over 32 and don't hash if under 32.   If you make it a best practice to always hash please update both sdk and docs to represent that cc @qbzzt

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
roninjin10: Nice catch!

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
roninjin10: nice catch

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,273 @@
...
+        __EIP712_init(_name, EIP712_VERSION);
+    }
+
+    /**
+     * @notice Allows invite granter to set the number of invites an address has.
+     *
+     * @param _accounts    An array of accounts to update the invite counts of.
+     * @param _inviteCount Number of invites to set to.
+     */
+    function setInviteCounts(address[] calldata _accounts, uint256 _inviteCount) public {
```
smartcontracts, line 142: Do we generally expect every account to have the same number of invites? If so, this is fine. If not, we could consider making this an array of structs of (account, inviteCount). Just depends on how we expect to use it.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,273 @@
...
+
+    /**
+     * @notice Initializes this contract, setting the EIP712 context.
+     *
+     *         Only update the EIP712_VERSION when there is a change to the signature scheme.
+     *         After the EIP712 version is changed, any signatures issued off-chain but not
+     *         claimed yet will no longer be accepted by the claimInvite function. Please make
+     *         sure to notify the issuers that they must re-issue their invite signatures.
+     *
+     * @param _name Contract name
```
smartcontracts, line 130: Nit: missing period
```suggestion
     * @param _name Contract name.
```

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,273 @@
...
+ *              received signature keccak256(abi.encode(addressToReceiveTo, receivedSignature))
+ *              using the commitInvite function
+ *           6) claimer reveals the plaintext ClaimableInvite and the signature using the
+ *              claimInvite function, receiving the "optimist.can-mint-from-invite" attestation
+ */
+contract OptimistInviter is Semver, EIP712Upgradeable {
+    /**
+     * @notice Emitted when an invite is claimed.
+     *
+     * @param issuer Address that issued the signature.
```
smartcontracts, line 41: Nit: param spacing

```suggestion
     * @param issuer  Address that issued the signature.
```

packages/contracts-periphery/contracts/testing/helpers/OptimistInviterHelper.sol
```diff
@@ -0,0 +1,108 @@
...
+
+import { ECDSA } from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+import { OptimistInviter } from "../../universal/op-nft/OptimistInviter.sol";
+
+/**
+ * Simple helper contract that helps with testing flow and signature for OptimistInviter contract.
+ * Made this a separate contract instead of including in OptimistInviter.t.sol for reusability.
+ */
+contract OptimistInviterHelper {
```
smartcontracts, line 12: Worth adding comments to the functions/variables here

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,273 @@
...
+     *         front-running when the invitee is claiming the invite. Without a commit and reveal
+     *         scheme, anyone who is watching the mempool can take the signature being submitted
+     *         and front run the transaction to claim the invite to their own address.
+     *
+     *
+     * @param _commitment A hash of the claimer and signature concatenated.
+     *                    keccak256(abi.encode(_claimer, _signature))
+     */
+    function commitInvite(bytes32 _commitment) public {
+        commitments[_commitment] = true;
```
smartcontracts, line 189: Instead of setting to `true` here we need to set to `block.timestamp` and we need to enforce a minimum amount of time between commit and reveal (ENS uses 1 minute). Otherwise someone can frontrun by committing and revealing quickly once they spot the reveal transaction.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: Yeah, the main we expect to use it is to issue everyone 3 invites.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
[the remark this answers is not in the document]
tynes: i think we should only hash if len > 32, this is how many things in ethereum consensus work, like for example leaves in merkle trees that are in the block header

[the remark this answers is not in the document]
tynes: solidity also hashes things that are abitrary len when their size is greater than 31 to determine storage slots

packages/contracts-periphery/contracts/testing/helpers/OptimistInviterHelper.sol
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * @notice Keeps track of current nonce to generate new nonces for each invite.
+     */
+    uint256 public currentNonce;
+
+    constructor(OptimistInviter _optimistInviter, string memory _name) {
+        optimistInviter = _optimistInviter;
+        name = _name;
+        currentNonce = 0;
```
tynes, line 45: no need to set this to 0, its already 0

jakim929: fixed!

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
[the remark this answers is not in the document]
jakim929: Got it - since this is a common accepted pattern, will stick to the policy of only hashing if a value is len > 32.

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
jakim929: a

packages/contracts-periphery/contracts/universal/op-nft/OptimistInviter.sol
```diff
@@ -0,0 +1,301 @@
...
+     *         transaction succeeds. With this, frontrunning a commitment requires that an attacker
+     *         be able to prevent the honest claimer's claimInvite transaction from being included
+     *         for this long.
+     */
+    uint256 public constant MIN_COMMITMENT_PERIOD = 60;
+
+    /**
+     * @notice Struct that represents a claimable invite that will be signed by the issuer.
+     *
+     * @custom:field issuer   Address that issued the signature. Reason this is explicitly included,
```
smartcontracts, line 92: Nit, spacing between field name and description is off

PR #5270, opened 2023-03-27
specs: attestation dispute game

tynes:
**Description**

First pass at defining the attestation dispute game contract. This should be filled in with more detail as we progress on it

Closes CLI-3479
Closes CLI-3480


refcell (approved):
Small consideration, but as we discussed offline, can work on improvements iteratively as long as we are correct

PR #5271, opened 2023-03-27
indexer: update to latest bedrock dependencies

protolambda:
Update indexer to use the latest Bedrock modules. It's not part of the bedrock monorepo root Go module yet, and thus the need to update the deps.
Also fixes a duplicate `replace` in `go.mod`


trianglesphere (approved):

tynes:
confirming this builds on my machine

tynes (approved):

PR #5276, opened 2023-03-28
Generate JSON deploy config from typescript

maurelian:
**Description**

1. Adds a `mainnet-rehearsal` network so that we are not continuously modifying the `mainnet` config. 
2. Adds an hh task for generating a JSON config from a TS file. This allows us to document our reasoning for various config values in the file. 

Feedback request:

Should this hh task have an entry in the package.json?

**TODOs**


tynes:
The problem with using `mainnet-rehearsal` as the network means we will need to copy the `mainnet` network's deployments in `contracts` package into a network `mainnet-rehearsal` and create a network in that package as well

packages/contracts-bedrock/tasks/generate-deploy-config.ts
```diff
@@ -0,0 +1,28 @@
+import fs from 'fs'
+import { task } from 'hardhat/config'
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
+
+task(
+  'generate-deploy-config',
+  'generates a json config file for the current network'
+).setAction(async ({}, hre: HardhatRuntimeEnvironment) => {
+  try {
+    const base = `${hre.config.paths.deployConfig}/${hre.network.name}`
```
tynes, line 10: nit: `path.join`

packages/contracts-bedrock/tasks/generate-deploy-config.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (fs.existsSync(`${base}.ts`)) {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const config = require(`${base}.ts`).default
+      console.log('config:', config)
```
tynes, line 14: perhaps no logging of the full config here? especially since its shared with a string in the same log

maurelian: Whoops, debug code left in.

packages/contracts-bedrock/tasks/generate-deploy-config.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { task } from 'hardhat/config'
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
+
+task(
+  'generate-deploy-config',
+  'generates a json config file for the current network'
+).setAction(async ({}, hre: HardhatRuntimeEnvironment) => {
+  try {
+    const base = path.join(hre.config.paths.deployConfig, hre.network.name)
```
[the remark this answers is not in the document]
maurelian: oh, good call out semgrep, but in this case the config values are provided by the config file, which is subject to PR review.

maurelian:
> The problem with using `mainnet-rehearsal` as the network means we will need to copy the `mainnet` network's deployments in `contracts` package into a network `mainnet-rehearsal` and create a network in that package as well

oh, that's ugly. I will revert to using `mainnet`, but use the new ability to add comments to the config, to clarify.

tynes:
Good to merge after linting errors are fixed

PR #5277, opened 2023-03-28
Use OptimismPortal.FINALIZATION_PERIOD_SECONDS if L2OutputOracle.version is '1.0.0'.

zchn:
**Description**

@eth-optimism/sdk starting at 2.0.0 is assuming the `FINALIZATION_PERIOD_SECOND` is moved to L2OutputOracle, which is not the case for L2OutputOracle.version === '1.0.0'. 

Version 1.0.0 of the L2OutputOracle is still deployed at Goerli, and all SDK consumers depending on the `FINALIZATION_PERIOD_SECOND` are broken right now (e.g. the fault-detector). This PR makes the sdk work with both 1.0.0 and the latest version of smart contracts.

**Tests**

No tests added. Manually ran `yarn start` in fault-detector to confirm this fixed the breakage.

Happy add a test if necessary.

**Invariants**

N/A since this PR does not change critical path.


packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1003,9 +1003,17 @@ export class CrossChainMessenger {
    * @returns Current challenge period in seconds.
    */
```
tynes, line 1004: Slight preference to get the version of the oracle once and assign it to the class and then use that version because we will have more version dependent logic in the future

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1003,9 +1003,17 @@ export class CrossChainMessenger {
...
-      : await this.contracts.l1.StateCommitmentChain.FRAUD_PROOF_WINDOW()
+    if (!this.bedrock) {
+      return (await this.contracts.l1.StateCommitmentChain.FRAUD_PROOF_WINDOW()).toNumber()
+    }
+
+    const oracle_version = await this.contracts.l1.L2OutputOracle.version()
+    const challengePeriod = oracle_version === '1.0.0'
+      ? BigNumber.from(await this.contracts.l1.OptimismPortal.provider.call({
+        to: this.contracts.l1.OptimismPortal.address,
+        data: '0xf4daa291' // FINALIZATION_PERIOD_SECOND
```
smartcontracts, line 1014:
```suggestion
        data: '0xf4daa291' // FINALIZATION_PERIOD_SECONDS
```

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1003,9 +1003,17 @@ export class CrossChainMessenger {
...
    */
   public async getChallengePeriodSeconds(): Promise<number> {
-    const challengePeriod = this.bedrock
-      ? await this.contracts.l1.L2OutputOracle.FINALIZATION_PERIOD_SECONDS()
-      : await this.contracts.l1.StateCommitmentChain.FRAUD_PROOF_WINDOW()
+    if (!this.bedrock) {
+      return (await this.contracts.l1.StateCommitmentChain.FRAUD_PROOF_WINDOW()).toNumber()
+    }
+
+    const oracle_version = await this.contracts.l1.L2OutputOracle.version()
```
smartcontracts, line 1010: Should be `oracleVersion`

packages/sdk/src/cross-chain-messenger.ts
smartcontracts: Worth explaining in a comment why this is using a `call` rather than a function. I understand it but someone reading the code might not.

packages/sdk/src/cross-chain-messenger.ts
roninjin10: Doing it in the constructor would be bad because some uses of the cross chain messenger initialize cross chain messenger ephemerally in a function instead of passing around a global object.   We would want to lazily cache it.

```typescript
public async getOracleVersion(): Promise<number> {
  if (!this.oracleVersion) this.oracleVersion = await this.contract.l1.L2OutputOracle.version() 
  return this.oracleVersion
}
```

I don't think we should do this though because we'd want to think through the implications of this for long running processes.   Very quickly we would want the cache to expire and the engineering effort / surface area for bugs becomes too big for such a tiny optimization

roninjin10 (approved):

packages/sdk/src/cross-chain-messenger.ts
roninjin10:
```suggestion
    const oracleVersion = await this.contracts.l1.L2OutputOracle.version()
```

packages/sdk/src/cross-chain-messenger.ts
roninjin10: Actually thinking about it more I wouldn't be opposed to doing ^

roninjin10: I actually do think that's a good idea but even if we do do that I'd recommend separating that into it's own pr and keeping this pr focused

PR #5279, opened 2023-03-28
contracts-bedrock: delete dead SystemConfig code

tynes:
**Description**

The old constant for the min gas limit was left behind in the PR that made the min gas limit calculation safe. This commit deletes the constant and updates the tests to use the new method.


refcell (approved):
Gas savings, we love to see it

tynes:
Linting fix is here https://github.com/ethereum-optimism/optimism/pull/5281

tynes:
Rebuilt the bindings, this should fix the bindings build task

PR #5286, opened 2023-03-29
txmgr: Restructure internals and add TxNotInMempoolTimeout

trianglesphere:
**Description**

This restructures a lot of the internal of transaction sending. It adds a new flag and features where it aborts sending the transaction if the transaction is never seen on the network after a default of 2 minutes (configurable via txmgr flag).

The transaction manager's `sendState` has been modified to track successful publishes and the time of its creation. If it goes for `TxNotInMempoolTimeout` time without seeing a successful transaction publish (i.e. it believes the tx is not in the mempool) then it will abort.

The goal of this change is to ensure that the transaction manager does not get stuck in state where the transaction will not get confirmed while not accidentally recording transactions as failed when they in fact succeeded. 

**TODOs**


op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+	log.Info("publishing transaction")
+
+	cCtx, cancel := context.WithTimeout(ctx, m.cfg.NetworkTimeout)
+	defer cancel()
+	err := m.backend.SendTransaction(cCtx, tx)
+	// ProcessSendError will log the error at the correct verbosity
+	sendState.ProcessSendError(err, log)
+
+	// Poll for the transaction to be ready & then send the result to receiptChan
+	receipt, err := m.waitMined(ctx, tx, sendState)
```
ajsutton, line 258: Previously if `SendTransaction` returned an error, we wouldn't try to wait for the tx whereas now we carry on checking for the receipt regardless of a send error.  Is that deliberate?

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+			// Increase the gas price & submit the new transaction
+			tx = m.increaseGasPrice(ctx, tx)
 			wg.Add(1)
 			go sendTxAsync(tx)
 
-		// The passed context has been canceled, i.e. in the event of a
-		// shutdown.
 		case <-ctx.Done():
+			m.l.Info("Send transaction context cancelled", "err", ctx.Err())
 			return nil, ctx.Err()
```
ajsutton, line 235: This is also logging and returning an error. Might be better to just have:
```suggestion
			return nil, fmt.Errorf("send transaction context cancelled: %w", ctx.Err())
```

ajsutton (changes requested):
Generally looks good - left a few clarifying comments. Main concern for me is having to sleep in tests - that's a very common source of intermittency and leads to slower tests.

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
 		case receipt := <-receiptChan:
+			m.l.Info("Found transaction receipt", "hash", receipt.TxHash)
 			return receipt, nil
 		}
 	}
 }
 
-// waitMined implements the core functionality of WaitMined, with the option to
-// pass in a SendState to record whether or not the transaction is mined.
+// publishAndWaitForTx publishes the transaction to the transaction pool and thens waits for it with [waitMined].
```
sebastianst, line 244:
```suggestion
// publishAndWaitForTx publishes the transaction to the transaction pool and then waits for it with [waitMined].
```

op-service/txmgr/txmgr.go
sebastianst, line 258: If  sending the tx failed, we'll record it with `sendState.ProcessSendError(err, log)` but why do we then always wait for it to be mined? Aren't there at least some errors where we could safely abort here, before calling `waitMined`?

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+	cCtx, cancel := context.WithTimeout(ctx, m.cfg.NetworkTimeout)
+	defer cancel()
+	err := m.backend.SendTransaction(cCtx, tx)
+	// ProcessSendError will log the error at the correct verbosity
+	sendState.ProcessSendError(err, log)
+
+	// Poll for the transaction to be ready & then send the result to receiptChan
+	receipt, err := m.waitMined(ctx, tx, sendState)
+	if err != nil {
+		return
```
sebastianst, line 260: Maybe log error, at least debug level?

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to fetch the suggested gas tip cap: %w", err)
-	} else if tip == nil {
-		return nil, nil, errors.New("the suggested tip was nil")
+	receipt, err := m.backend.TransactionReceipt(ctx, txHash)
+	if errors.Is(err, ethereum.NotFound) {
+		if sendState != nil {
+			sendState.TxNotMined(txHash)
+		}
+		m.l.Trace("Transaction yet mined", "hash", txHash)
```
sebastianst, line 295:
```suggestion
		m.l.Trace("Transaction not yet mined", "hash", txHash)
```

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+func (m *SimpleTxManager) queryReceipt(ctx context.Context, txHash common.Hash, sendState *SendState) *types.Receipt {
+	ctx, cancel := context.WithTimeout(ctx, m.cfg.NetworkTimeout)
 	defer cancel()
-	tip, err := m.backend.SuggestGasTipCap(cCtx)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to fetch the suggested gas tip cap: %w", err)
-	} else if tip == nil {
-		return nil, nil, errors.New("the suggested tip was nil")
+	receipt, err := m.backend.TransactionReceipt(ctx, txHash)
+	if errors.Is(err, ethereum.NotFound) {
```
sebastianst, line 291: Does the backend always guarantee to translate rpc errors into correct Go error types?
I saw you using error string comparison in `ProcessSendError`
```
	} else if strings.Contains(err.Error(), core.ErrNonceTooLow.Error()) {
```
so maybe either here you also have to do strings comparison, or the other one can be changed to an `errors.Is` check?

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+		if sendState != nil {
```
sebastianst, line 292: Why is there a `nil` `sendState` case? for tests?

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+			sendState.TxNotMined(txHash)
+		}
+		m.l.Trace("Transaction yet mined", "hash", txHash)
+		return nil
+	} else if err != nil {
+		m.l.Info("Receipt retrievel failed", "hash", txHash, "err", err)
```
sebastianst, line 298:
```suggestion
		m.l.Info("Receipt retrieval failed", "hash", txHash, "err", err)
```

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+			}
 		}
 	}
 }
 
-// suggestGasPriceCaps suggests what the new tip & new basefee should be based on the current L1 conditions
-func (m *SimpleTxManager) suggestGasPriceCaps(ctx context.Context) (*big.Int, *big.Int, error) {
-	cCtx, cancel := context.WithTimeout(ctx, m.cfg.NetworkTimeout)
+// queryReceipt queries for the receipt and returns the receipt if it has passed the confirmation depth
+func (m *SimpleTxManager) queryReceipt(ctx context.Context, txHash common.Hash, sendState *SendState) *types.Receipt {
```
sebastianst, line 287: Why don't you return an `error` as well and handle (=log) the error at the calling site?

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +215,124 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+	// underflows.
+	if txHeight+m.cfg.NumConfirmations <= tipHeight+1 {
+		m.l.Info("Transaction confirmed", "hash", txHash)
+		return receipt
+	}
+
+	// Safe to subtract since we know the LHS above is greater.
+	confsRemaining := (txHeight + m.cfg.NumConfirmations) - (tipHeight + 1)
+	m.l.Debug("Transaction not yet confirmed", "hash", txHash, "confsRemaining", confsRemaining)
+	return nil
```
sebastianst, line 335: So every `waitMined` loop iteration queries the receipt from the backend anew? Is that easier than waiting for `confsRemaining` blocks in `waitMined` and only then query the receipt again for confirmation?

sebastianst (changes requested):
I'm wondering, do we even need the async tx sending in `send` with `sendTxAsync` via a goroutine and then waiting for the receipt on a channel?

It looks to me like the whole transaction sending loop could also just be a normal synchronous call to `publishAndWaitForTx`, which would simply return `*types.Receipt, error`. But maybe I'm just oversimplifying the algorithm here.

op-service/txmgr/txmgr.go
sebastianst: Just realize that Adrian has probably asked the same question :D

op-service/txmgr/txmgr.go
trianglesphere: You described the behavior correctly. It is probably easier to implement (and already exists). The receipt query interval is relatively slow by default (30s).

op-service/txmgr/txmgr.go
trianglesphere: The txpool errors have been cast correctly. I just copied historical code. I could try switching this over to an `err.Is` and see if that works.

op-service/txmgr/txmgr.go
trianglesphere: Because the logic is pretty complex. It used to all be inside `waitMined` (the minus diff just above), but it felt nicer to pull it into this function.

op-service/txmgr/txmgr.go
trianglesphere: Rest of the Q is the same.

op-service/txmgr/txmgr.go
sebastianst: would be a good cleanup, now that you're refactoring this

op-service/txmgr/txmgr.go
sebastianst: ok sounds reasonable. should we maybe change this interval to 12s?

op-service/txmgr/send_state.go
```diff
@@ -1,59 +1,75 @@
...
-	}
-
+func (s *SendState) ProcessSendError(err error, lgr log.Logger) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	// Record this nonce too low observation.
-	s.nonceTooLowCount++
+	// Record the type of error & log appropriately
+	if err == nil {
```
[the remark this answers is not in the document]
trianglesphere: It's a switch in go, but I converted it.

op-service/txmgr/send_state_test.go
```diff
@@ -160,3 +172,20 @@ func TestSendStateIsNotWaitingForConfirmationAfterTxUnmined(t *testing.T) {
 	sendState.TxNotMined(testHash)
 	require.False(t, sendState.IsWaitingForConfirmation())
 }
+
+// TestSendStateTimeoutAbort ensure that this will abort if it passes the tx pool timeout
+// when no successful transactions have been recorded
+func TestSendStateTimeoutAbort(t *testing.T) {
+	sendState := newSendStateWithTimeout(10 * time.Millisecond)
+	time.Sleep(20 * time.Millisecond)
```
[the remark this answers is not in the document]
trianglesphere: I mean we could inject the timeout time & force the end user to do the calculation.
I'm not really worried about these tests flaking, though it is true that the sleep isn't great.

op-service/txmgr/txmgr.go
trianglesphere: Confirmed that we need to do the strings check here.

trianglesphere: We actually need to do a string check for every one.

op-service/txmgr/txmgr.go
```diff
@@ -134,7 +132,7 @@ func (m *SimpleTxManager) Send(ctx context.Context, candidate TxCandidate) (*typ
 	tx, err := m.craftTx(ctx, candidate)
 	if err != nil {
 		m.l.Error("Failed to create the transaction", "err", err)
-		return nil, err
+		return nil, fmt.Errorf("failed to create the tx: %w", err)
```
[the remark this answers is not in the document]
trianglesphere: I went with returning rather than logging here.

op-service/txmgr/txmgr.go
trianglesphere: yea. Will do when I modify config defaults.

trianglesphere:
> I'm wondering, do we even need the async tx sending in `send` with `sendTxAsync` via a goroutine and then waiting for the receipt on a channel?
> 
> It looks to me like the whole transaction sending loop could also just be a normal synchronous call to `publishAndWaitForTx`, which would simply return `*types.Receipt, error`. But maybe I'm just oversimplifying the algorithm here.

It's because when we update the gas price, we are waiting for multiple different transaction hashes. Say a tx gets included in L1, but we then published another transaction with a higher gas price before we realize. If we only waited for the second TX we would miss that the first TX we sent got included on L1.

We could make it synchronous by polling every transaction that has been sent

op-service/txmgr/send_state_test.go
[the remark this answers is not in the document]
ajsutton: Passing in the deadline doesn't really help though because we still need to access the current time to check if we've passed that deadline. It's pretty standard to introduce some abstraction for the system clock to make it testable - otherwise these little sleeps in tests just keep growing over time and slowing everything down.
If it was a big change I could understand but it's such an easy win to pass in `time.Now` and use it as an abstract function within `SendState` that I'm really really keen to do it right and avoid the sleep entirely.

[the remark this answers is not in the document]
trianglesphere: Ok, I faked the clock. It's not quite as bad as I though, though I still don't love having to inject it.

ajsutton (approved):

ajsutton:
Actually I'm going to dismiss my review and just leave approval to Seb since he has better context. I really just meant to remove my request changes.

sebastianst:
> > I'm wondering, do we even need the async tx sending in `send` with `sendTxAsync` via a goroutine and then waiting for the receipt on a channel?
> > It looks to me like the whole transaction sending loop could also just be a normal synchronous call to `publishAndWaitForTx`, which would simply return `*types.Receipt, error`. But maybe I'm just oversimplifying the algorithm here.
> 
> It's because when we update the gas price, we are waiting for multiple different transaction hashes. Say a tx gets included in L1, but we then published another transaction with a higher gas price before we realize. If we only waited for the second TX we would miss that the first TX we sent got included on L1.
> 
> We could make it synchronous by polling every transaction that has been sent

Ah thanks, that makes sense. If we're running into problems with the current async architecture, I'd vote for this refactor. But fine for now since it works.

op-service/txmgr/txmgr.go
```diff
@@ -272,124 +216,137 @@ func (m *SimpleTxManager) send(ctx context.Context, tx *types.Transaction) (*typ
...
+			log.Warn("transaction is underpriced", "err", err)
 		default:
-			if sendState != nil {
-				sendState.TxNotMined(txHash)
-			}
-			m.l.Trace("Transaction not yet mined", "hash", txHash)
+			log.Error("unable to publish transaction", "err", err)
 		}
+		return
+	}
```
sebastianst, line 272: Are we guaranteed that every `m.backend` implementation that we will ever use never wraps any of these errors? If not, we should use `strings.Contains(err.Error(), foo.Error())` or `strings.HasSuffix` because an exact string match would not match for wrapped errors.

op-service/txmgr/txmgr.go
sebastianst: did you keep an `errors.Is` check, instead of a string comparison, in this line deliberately? asking since you changed all the other checks to string checks.

I'd also extract all these checks into a little helper
```go
func errorsStrIs(err, sub error) {
    return strings.Contains(err.Error(), sub.Error())
}
```

sebastianst (commented):
Looks great, just two open questions regarding error strings checking

op-service/txmgr/txmgr.go
trianglesphere: added

sebastianst (approved):

mslipper (approved):

trianglesphere:
@mergifyio refresh

PR #5292, opened 2023-03-29
contracts-bedrock: upgrade script for post sherlock

tynes:
**Description**

Includes a complete upgrade script that can be ran for upgrading any op-stack based chain that was deployed before the sherlock audit fixes were all included.

The script is ran with the following command:

```bash
forge script \
    scripts/upgrades/PostSherlock.s.sol \
    --rpc-url $ETH_RPC_URL \
    --private-key $PRIVATE_KEY \
    --sig 'run(address,address)' \
    $SAFE_ADDRESS \
    $PROXY_ADMIN_ADDRESS
```

Signers should run the script and be sure to add any required flags so that hardware wallets can be used. The script will submit a single transaction that approves the upgrade and then checks to see if there have been enough approvals and then triggers the upgrade if there have been enough approvals.

This comes with test coverage but requires a goerli RPC url. To run the tests, use the following command:

```bash
forge test --contracts scripts/upgrades --rpc-url $ETH_RPC_URL
```


packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
```diff
@@ -0,0 +1,382 @@
...
+        bytes32 hash = safe.getTransactionHash({
+            to: address(multicall),
+            value: 0,
+            data: data,
+            operation: Enum.Operation.DelegateCall,
+            safeTxGas: 0,
+            baseGas: 0,
+            gasPrice: 0,
+            gasToken: address(0),
+            refundReceiver: address(0),
```
refcell, line 145: Is there a downside to specifying the `msg.sender` here rather than `address(0)`?

packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
```diff
@@ -0,0 +1,382 @@
...
+
+    /**
+     * @notice Builds the signatures by tightly packing them together.
+     *         Ensures that they are sorted.
+     */
+    function buildSignatures() internal view returns (bytes memory) {
+        address[] memory addrs = new address[](approvals.length);
+        for (uint256 i; i < approvals.length; i++) {
+            addrs[i] = approvals[i];
+        }
```
refcell, line 259: Why even make the `approvals` storage bound if we just end up clearing and pushing new one's for each call of `_run`. Would it be possible to remove this and pass in the approvals as a memory address array to the `buildSignatures()` function so we don't need to do this?

refcell (commented):
Comments - Nice work!

clabby (commented):
LGTM- Double checked the versions and addresses for the goerli implementations / proxies and checks out. Agree w/ @refcell that `approvals` should be in memory, but given that it's a script I don't mind if we keep it as is. Deferring to @maurelian for explicit approval.

packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
tynes: You can't make a memory array without knowing it's size up front right? So to do this, we'd need to have to filter empty elements because the size of the array is dynamic as is

packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
tynes: I copied the arguments used in transactions that we have sent, `address(0)` generally represents default values and i believe it'll use `tx.origin` when its `address(0)`

packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
clabby: You don't necessarily need to filter empty elements (can just manually adjust the length after writing the elements), but it'd be uglier than the current method.

tynes:
Something is wrong with the cache in CI, getting disk corruption errors. Am currently running the integration test build locally to debug

PR #5296, opened 2023-03-30
ctb: Update slither config

maurelian:
**Description**

- Excludes all findings except for high severity.
- Also moves all slither config to the slither.config.json file.
- I ran slither triage to re-check for all slither detectors and updated the slither db with findings to ignore.
- The slither job is very slow in CI, so I've moved it to its own job and increased the resource class size.
- CI will now fail on _high severity_ slither findings.


tynes:
The slither db has absolute paths to files on your filesystem which is kind of strange, meaning its not deterministic between running it on different machines. What is the slither db useful for? To make it deterministic, we would want to run it in a docker image

maurelian:
> What is the slither db useful for? 

The db is created when you run slither using `--triage-mode`. 
In triage mode, slither will ask about each finding to determine if it should be shown for the next run. The results of this process are saved in `slither.db.json`.

> The slither db has absolute paths to files on your filesystem which is kind of strange, meaning its not deterministic between running it on different machines. 
> To make it deterministic, we would want to run it in a docker image

I agree, that is weird and in my experience slither doesn't depend on those values anyways. I think it's fine because triage mode should only ever be run locally.
Rather than forcing people to use docker, I've added a new yarn command to run triage mode, and updated the script so that it will remove these absolute paths from the db. I've done this manually in the past, and slither doesn't seem to depend on having the `filename_absolute` property in the db anyways.

trianglesphere:
CI is failing.

maurelian:
looks like `yarn slither` is going to time out, not sure why, will need to investigate later.

packages/contracts-bedrock/scripts/slither.sh
```diff
@@ -2,10 +2,24 @@
...
-slither . --foundry-out-directory artifacts
-
-mv $TEMP/test contracts/test
+  # For whatever reason the slither db contains a filename_absolute property which includes the full
+  # local path to source code on the machine where it was generated. This property does not
+  # seem to be required for slither to run, so we remove it.
+  DB=slither.db.json
+  TEMP_DB=temp-slither.db.json
+  mv $DB $TEMP_DB
+  jq 'walk(if type == "object" then del(.filename_absolute) else . end)' $TEMP_DB >$DB
```
tynes, line 21: nit: no space after `>`

tynes (approved):

trianglesphere (approved):
lgtm. Was waiting for CI to go green.

PR #5312, opened 2023-03-31
contracts-bedrock: add l1 deploy tag

tynes:
**Description**

Prepares for adding in L2 deployment scripts by adding a `l1` tag to the deploy scripts that should run on L1. Also add the tag flag to the deployment that happens in CI. The L2 contract deployments will be added in another commit. It is not a problem if the L2 contracts are deployed to L1 in a testing environment but using tags to save time and gas should be done. Also lints a few lines of code where awaits are nested, nested awaits are difficult to read.


maurelian (approved):
This is good for me, though I am slightly confused about why it is necessary. All of these scripts are specific to L1.

PR #5321, opened 2023-03-31
fix: unrevert + attempt to fix build

tynes:
**Description**

Attempt to fix the build and un revert the removal of the solidity script. This script is necessary for 1) other networks to upgrade and 2) be the foundation for our next set of upgrades. We cannot be code frozen until all contracts are upgraded, so we need this script. If we cannot use the deps then I will vendor the functionality we need.


tynes:
Will follow up with a fix. I think its as simple as vendoring `LibSort` from `solady` and then vendoring the `GnosisSafe` interface

maurelian (commented):
LGTM as soon as the build is passing

tynes:
I have removed the deps and now the used code is vendor'd, its not a lot of code so we should be able to manage the vendoring

tynes:
This should build with no issues now, upgrading the L2 contracts will require this merged

PR #5323, opened 2023-03-31
specs: Bond Manager Interface 🎩 

refcell:
# Overview

Implements the specifications for the Bond Manager as detailed in CLI-3486.

**Metadata**

- Closes CLI-3486


specs/bond-manager.md
```diff
@@ -0,0 +1,100 @@
...
+  - [Simple Bond](#simple-bond)
+  - [Variable Bond](#variable-bond)
+- [Types of Bond Managers](#types-of-bond-managers)
+  - [OracleBondManager](#oraclebondmanager)
+  - [AttestationBondManager](#attestationbondmanager)
+  - [FaultBondManager](#faultbondmanager)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
```
tynes, line 18: Could we talk about the bond manager in a more generic sense first, without connection to the `L2OutputOracle`? Starting off with what a bond is (value) that acts as economic security and it can be claimed if there is proof of malicious intent? Something like that

specs/bond-manager.md
```diff
@@ -0,0 +1,100 @@
...
+
+In the `L2OutputOracle`, a permissionless proposer can post an output proposal, which
+becomes finalized after a pre-determined interval detailed in the
+[proposals doc](./proposals.md). When proposing an output, the proposer must
+post a bond that is used to disincentivize spam. When the output is finalized,
+the bond may be claimed by the proposer. But in the case that the output is invalid,
+the bond *should* be seized by a set of challenge agents or a
+[dispute game](./dispute-game.md).
+
+### AttestationBondManager
```
tynes, line 84: I think we can focus on having a single type of bond manager for now and say that a new type of bond manager will be introduced in the future to enable the challenge response nature of the fault proof dispute game

specs/bond-manager.md
```diff
@@ -0,0 +1,100 @@
...
+
+At its core, the bond manager is straightforward - it escrows or holds ether and returns
+it when finalized or seized. But the uncertainty of introducing bonds lies in the
+bond _sizing_, i.e. how much should a bond be? Sizing bonds correctly entails a variety of
+tradeoffs. Price them too high, and not enough proposers will post outputs. Price them too low, and
+challengers won't be incentivized to delete outputs if the gas cost of doing so outways the bond itself.
+
+Below, we outline two different approaches to sizing bonds and the tradeoffs of each.
+
+### Simple Bond
```
tynes, line 46: I think we should try to describe the type of bond independent of the thing that it is contesting (the L2 Output Oracle). We want to say that there are some invariants that must be held for this to be incentive compatible:
- known cost of progressing the game
- known upper bound on cost of completing the game

The size of the bond can be computed up front when the dispute game is created. We define a linear function that is something like: `number of signers * amount of gas used to progress game (add signature) + security overhead`. We could also multiply in the security overhead but made it simple to not deal with decimals

specs/bond-manager.md
```diff
@@ -0,0 +1,100 @@
...
+
+With this approach, the Simple Bond is fixed to `1 ether`. By working backwards, we
+can establish that the cost of challenging an output proposal must not
+exceed `1 ether`. As such, this implies a base fee of `10,000` for a challenge costing
+`100,000` gas with a `1 gwei priority fee`. This leaves challenger agents with a significant
+buffer between the historical highest gas price of roughly `236` in 2020, and the base gas fee of `10,000`.
+
+### Variable Bond
+
+Better bond heuristics can be used to establish a bond price that accounts for
```
tynes, line 62: Definitely would be cool to include more detail from https://www.notion.so/oplabs/Simple-Secure-Bonding-Protocol-26bb82a74a464708a9885c9115529756 but we aren't going to be implementing that in the short term so its ok to not have it be super detailed

specs/bond-manager.md
refcell: Agreed 👍

specs/bond-manager.md
refcell: I'm not sure how much more context I can add here. The notion doc is fairly minimal and I like to not put unfinished ideation/decisions in the specs.

specs/bond-manager.md
```diff
@@ -0,0 +1,111 @@
...
+The _Simple Bond_ is a very conservative approach to bond management, establishing a **fixed** bond
+size. The idea behind simple bond pricing is to establish the worst case gas cost for
+the next step in the dispute game.
+
+With this approach, the size of the bond is computed up-front when a dispute game is created.
+For example, in an attestation dispute game, this bond size can be computed using the following
+simple linear function:
+```
+size = number of signers * gas used to progress game + security overhead
+```
```
tynes, line 61: real quick, thinking we may want to alter the order of operations

```
size = number of signers * (gas used to progress game + security overhead)
```

So there is a security overhead attached with each move in the game instead of a single value added as the overhead, this would mean the security overhead is equally as effective as the number of signers increases

specs/bond-manager.md
```diff
@@ -0,0 +1,111 @@
...
+
+Notice that since the bond size is linearly proportional to the number of signers, the economic
+security a given bond size provides decreases as the number of signers increases.
+
+Working backwards, if we assume the number of signers to be `5` and the gas used to progress
+the game is `100,000` gas, then the bond size should cover `500,000` gas. This means that a bond
+of `1 ether` would cover the cost of progressing the game for `5` signers as long as the gas price
+(base fee) does not exceed `2,000 gwei`. This leaves a significant buffer between the highest
```
tynes, line 69: does not exceed ... for the entire finalization window. It would be prohibitively expensive to hold the basefee this high etc

tynes:
Looks like specs are failing linting

```
yarn lint:specs:toc && git diff --exit-code ./specs
```

specs/bond-manager.md
```diff
@@ -0,0 +1,112 @@
...
+
+Concretely, outputs will be permissionlessly proposed to the `L2OutputOracle` contract.
+When calling the propose function, the ether value is sent as the bond. This bond is
+then held by a bond manager contract. The bond manager contract is responsible for
+both the [dispute-games](./dispute-game.md) and the `L2OutputOracle` (further detailed
+in [proposals](./proposals.md)).
+
+The bond manager will need to handle bond logic for a variety of different
+[dispute-games](./dispute-game-interface.md). In the simplest "attestation" dispute game,
+bonds will not be required since the attestors are permissioned. But in more complex games,
```
clabby, line 34:
```suggestion
bonds will not be required since the attestors are a permissioned set of trusted entities. But in more complex games,
```

clabby: Not necessary, but might be a good flag

specs/bond-manager.md
```diff
@@ -0,0 +1,112 @@
...
+such as the fault dispute game, challengers and defenders perform a series of
+alternating onchain transactions requiring bonds at each step.
+
+## The Bond Problem
+
+At its core, the bond manager is straightforward - it escrows or holds ether and returns
+it at maturity or seized. But the uncertainty of introducing bonds lies in the
```
clabby, line 41: At the moment, the other specs do not call for the bond manager returning the bond once the proposal reaches maturity (i.e. a `push`). The bond can be pulled from the bond manager and pushed to the entities that participate in providing attestor signatures to the attestation game upon resolution, though.

This opens up a bit of an interesting design decision where a caller of `OptimismPortal.finalizeWithdrawalTransaction` returns the bond on a matured output proposal (if it has not already been returned), but I think it makes more sense to go with @smartcontracts' initial idea from CDMX where the BondManager silo's liquid (withdrawable) funds from locked (bonded) funds.

specs/bond-manager.md
```diff
@@ -0,0 +1,112 @@
...
+  /// @notice Post the bond of the bond owner.
+  function postBond(address bondOwner) external;
+
+  /// @notice Seize the bond of the bond owner.
+  /// @dev `nonce` is required since the bond owner may have multiple outstanding bonds.
+  function seizeBond(address bondOwner, uint256 nonce) external;
+
+  /// @notice Claim the bond of the bond owner.
+  /// @dev `nonce` is required since the bond owner may have multiple outstanding bonds.
+  function claimBond(address bondOwner, uint256 nonce) external;
```
tynes, line 97: how do you know what nonce to use?

specs/bond-manager.md
```diff
@@ -0,0 +1,112 @@
...
+actors posting this bond are responsible for funding this contract.
+
+## Contract Interface
+
+Below is a minimal interface for the bond manager contract.
+
+```solidity
+interface BondManager {
+  /// @notice Post the bond of the bond owner.
+  function postBond(address bondOwner) external;
```
tynes, line 89: this needs to be payable?

specs/bond-manager.md
```diff
@@ -0,0 +1,112 @@
...
+The _Simple Bond_ is a very conservative approach to bond management, establishing a **fixed** bond
+size. The idea behind simple bond pricing is to establish the worst case gas cost for
+the next step in the dispute game.
+
+With this approach, the size of the bond is computed up-front when a dispute game is created.
+For example, in an attestation dispute game, this bond size can be computed using the following
+simple linear function:
+
+```md
+size = number of signers * (gas used to progress game + security overhead)
```
clabby, line 61: Each call to `challenge` should cost a constant amount of gas except for the last one. The final entity to `challenge` will implicitly also call `resolve`, which makes this slightly inaccurate.

I think it should be more like:
```
bond_size = (signer_threshold * challenge_gas) + resolution_gas(signer_threshold) + security overhead

File: specs/bond-manager.md [hunk 12]
```diff
@@ -0,0 +1,112 @@
...
+separate oracle contract, `GasPriceFluctuationTracker`, that tracks gas fluctuations
+within a pre-determined bounds. This replaces the ideal solution of tracking
+challenge costs over all L1 blocks, but provides a reasonable bounds. The initial
+actors posting this bond are responsible for funding this contract.
+
+## Contract Interface
+
+Below is a minimal interface for the bond manager contract.
+
+```solidity
```
tynes, line 86: our standard is seaport style comments

specs/bond-manager.md
clabby: the `resolution_gas` function in the above eq will need to be determined upon implementation, but it should still scale linearly wrt the `signer_threshold`

specs/bond-manager.md
```diff
@@ -0,0 +1,112 @@
...
+interface BondManager {
```
clabby, line 87: I think we need a bit more information to "tag" bonds here. The bond owner is an important piece, but the proposer of an output is not currently tracked in the `OutputProposal` struct: https://github.com/ethereum-optimism/optimism/blob/9c4eeaaa102efe5ed5e4e87d5b3ece9e2d7ca85c/packages/contracts-bedrock/contracts/libraries/Types.sol#L9-L22

Maybe rather than keying bonds by an address, we key them by a unique identifier?

clabby: i.e., when the `AttestationDisputeGame` resolves, it needs to be able to find the bond associated with the output that it's about to delete. The only information available is the old output root, the l1 timestamp that it was proposed at, and the l2blocknumber that the old output root commits to.

specs/bond-manager.md
refcell: Ah yes, I didn't mean to intend a push here, but will make it explicit since you interpreted it as such :)

specs/bond-manager.md
refcell: I think this was meant to be a rough approximation anyways since `gas used to progress game` is very general :P

Let's discuss offline since this is not decided.

specs/bond-manager.md
refcell: Re-worked it a bit :D

specs/bond-manager.md
```diff
@@ -0,0 +1,146 @@
...
+In the context of permissionless output proposals, bonds are a value that must
+be attached to an output proposal. In this case, the bond will be paid in ether.
+
+By requiring a bond to be posted with an output proposal, spam and invalid outputs
+are disincentivized. It's important to note that it is directly disincentivized
+because if the output is invalid, it will be seized by a set of challenge agents
+when the output is deleted. Thus, the bond acts as a form of economic security.
+
+Concretely, outputs will be permissionlessly proposed to the `L2OutputOracle` contract.
+When calling the propose function, the ether value is sent as the bond. This bond is
```
tynes, line 27: We should decouple mention of a particular function, i think a better way to frame it is "submitting an output proposal"

specs/bond-manager.md
```diff
@@ -0,0 +1,146 @@
...
+actors posting this bond are responsible for funding this contract.
+
+## Contract Interface
+
+Below is a minimal interface for the bond manager contract.
+
+```solidity
+/**
+ * @title IBondManager
+ * @notice The IBondManager is an interface for a contract that handles bond management.
```
tynes, line 90: I don't think we need the word bond in any of the functions because its the bond manager. This is a personal preference but `seize` feels cleaner to me than `seizeBond`. Same for each of the other function names

specs/bond-manager.md
```diff
@@ -0,0 +1,146 @@
...
+the `l2BlockNumber` associated with the output proposal since there will only ever be one
+outstanding output proposal for a given `l2BlockNumber`.
+This also avoids the issue where an output proposer can have multiple bonds if bonds were
+instead tied to the address of the output proposer.
+
+## Bond Manager Implementation
+
+Initially, the bond manager will only be used by the `L2OutputOracle` contract
+for output proposals in the attestation [dispute game](./dispute-game.md). Since
+the attestation dispute game has a permissioned set of attestors, there are no
```
tynes, line 139: This is a bit unclear because PoPs will have a bond but for posting the output. The spec for the attestation dispute game should mention that new bonds are not posted for each step. This could mention that dispute games could either have their bond split up so that a new bond is posted to cover the next step (more capital efficient) or the bond being posted in full up front

specs/bond-manager.md
refcell: 100% good catch

specs/bond-manager.md
refcell: Tried to change it up a little bit - let me know if this is more along the lines of what you're thinking. I would note that i think it's ok to explicitly put this differentiation here so that it's clear the bond management for dispute games is different from that of the `L2OutputOracle` contract's outputs.

If we don't state this explicitly it's probably easy to confuse where bonds are posted and why.

specs/bond-manager.md
```diff
@@ -0,0 +1,146 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+In the context of permissionless output proposals, bonds are a value that must
+be attached to an output proposal. In this case, the bond will be paid in ether.
+
+By requiring a bond to be posted with an output proposal, spam and invalid outputs
+are disincentivized. It's important to note that it is directly disincentivized
```
tynes, line 22: > It's important to note that it is directly disincentivized
because if the output is invalid, it will be seized by a set of challenge agents
when the output is deleted.

Its not clear what `it` is in this context, this sentence can be made more clear

refcell: Cleaned it up a bit :)

clabby (approved):

tynes (approved):

PR #5303, opened 2023-03-30
feat(contracts-periphery): create OptimistAllowlist contract

jakim929:
This contract will serve as the source of truth for whether someone can mint an Optimist NFT. In the next PR, will update Optimist.sol to use OptimistAllowlist to check mint eligibility.

Right now, it checks 3 signals
- attestations from our multisig (ALLOWLIST_ATTESTOR) - we've already issued ~70 of these
- attestations from the Coinbase Wallet Quest 
- attestations from the OptimistInviter contract (for people who claimed invites)

PR #5267, opened 2023-03-27
update Optimist.sol contract to use allowlist contract

jakim929:

PR #5324, opened 2023-03-31
feat(contracts-periphery): update Optimist contract to use OptimistAllowlist

jakim929:
Updates Optimist NFT contract to use OptimistAllowlist as source of truth for whether someone can mint or not.

Added tests check the new ways that allow you to mint
- Valid attestation from Coinbase Quest Attestor
- Valid attestation from the OptimistInviter

Added the https://github.com/mds1/multicall/blob/master/src/Multicall3.sol contract as a test helper to test that multicall works correctly to batch actions when claiming the NFT.
- The reveal step (calling optimistInviter.claimInvite) and minting step (calling optimist.mint), will be batched by the frontend to improve UX.


jakim929:
* develop
  * **PR #5265** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5265" target="_blank">Graphite</a> 
    * **PR #5305** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5305" target="_blank">Graphite</a> 
      * **PR #5324** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5324" target="_blank">Graphite</a>  👈
    * **PR #5303** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5303" target="_blank">Graphite</a> 
    * **PR #5266** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5266" target="_blank">Graphite</a> 
      * **PR #5267** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5267" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5324?utm_source=stack-comment).

packages/contracts-periphery/contracts/testing/helpers/Multicall3.sol
```diff
@@ -0,0 +1,261 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+// https://github.com/mds1/multicall/blob/master/src/Multicall3.sol
```
jakim929, line 4: Copied over contract from https://github.com/mds1/multicall/blob/master/src/Multicall3.sol

packages/contracts-periphery/contracts/universal/op-nft/Optimist.sol
```diff
@@ -6,40 +6,54 @@ import {
...
+     * @param _baseURIAttestor    Address of the baseURI attestor.
      * @param _attestationStation Address of the AttestationStation contract.
+     * @param _optimistAllowlist  Address of the OptimistAllowlist contract
      */
     constructor(
         string memory _name,
         string memory _symbol,
-        address _attestor,
-        AttestationStation _attestationStation
-    ) Semver(1, 0, 0) {
```
roninjin10, line 40: SInce you deleted ATTESTOR this is a major change

packages/contracts-periphery/contracts/testing/helpers/Multicall3.sol
roninjin10: Can you install this with forge or npm?

roninjin10 (changes requested):

packages/contracts-periphery/contracts/testing/helpers/Multicall3.sol
jakim929: I tried with forge install, but because forge uses git submodules, it gets added to the repo top-level .gitmodules, which isn't ideal. Do you have a suggest way of installing submodules scoped to the `contracts-periphery` package?

tynes: It should be fine if its in the top level `gitmodules` - I believe the build always happens with the monorepo, you can still have the path its installed to being in `packages/contracts-periphery`. Will need to change CI to ensure the submodules are installed

packages/contracts-periphery/contracts/foundry-tests/Optimist.t.sol
```diff
@@ -284,54 +488,142 @@ contract OptimistTest is Optimist_Initializer {
...
+            inviterPrivateKey,
+            optimistInviterHelper.getDigest(claimableInvite)
+        );
+        bytes memory signature = abi.encodePacked(r, s, v);
+
+        bytes32 hashedCommit = keccak256(abi.encode(bob, signature));
+
+        // commit the invite
+        vm.prank(bob);
+        optimistInviter.commitInvite(hashedCommit);
```
tynes, line 597: Signature generation isn't deterministic so including the sig in the commit hash is slightly sus

jakim929: Discussed with Mark:

- The secp256k1 signature generation is non-deterministic, so if the issuer attempts to create the same signature twice using the same payload, they may end up with two different signatures
- This shouldn't be a big problem in the intended use for this contract, as it's a very unlikely scenario where the issuer attempts to create the same signature twice. 

Signature issuer side
- They issue a signature, then it gets stored in localStorage
- They create a url with the signature in the params that they send to the invitee

Recipient side
- They receive the link through chat or messenger
- They go to the link, and the signature gets stored in their localStorage
- They commit the signature
- They reveal with the signature in their local storage

If the recipient’s browser cache is lost, they need to find the link again in order to continue with the process.

At worst, the if the recipient commits an invite, then loses the link, and clears the browser cache, they need to ask the issuer to send them the link again. If the issuer has also lost the local copy of the created signature, it is difficult for the issuer to generate the same signature. In this case, it’s recommended that the issuer just generates a new signature/link, and the recipient goes through the commit/reveal process again.

packages/contracts-periphery/.gitmodules
jakim929, line 1: Added because calling forge install in /contracts-periphery throws error if there's no local .gitmodules file.

.gitmodules
```diff
@@ -1,3 +1,8 @@
 [submodule "tests"]
 	path = l2geth/tests/testdata
 	url = https://github.com/ethereum/tests
+[submodule "packages/contracts-periphery/lib/multicall"]
+	path = packages/contracts-periphery/lib/multicall
+	url = https://github.com/mds1/multicall
+[submodule "lib/multicall"]
+	branch = v3.1.0
```
jakim929, line 8: result of `forge install mds1/multicall`

packages/contracts-periphery/contracts/testing/helpers/Multicall3.sol
jakim929: Updated the PR to install the repo as a submodule using `forge install`. This required using `deployCode` since the solidity versions aren't compatible (Multicall3.sol uses 0.8.12).

packages/contracts-periphery/contracts/universal/op-nft/Optimist.sol
```diff
@@ -6,40 +6,54 @@ import {
...
-        ATTESTOR = _attestor;
+        address _baseURIAttestor,
+        AttestationStation _attestationStation,
+        OptimistAllowlist _optimistAllowlist
+    ) Semver(2, 0, 0) {
```
smartcontracts, line 53: Semver bump is mismatched with the semver in the comment on this function

smartcontracts (commented):
One small comment

packages/contracts-periphery/contracts/universal/op-nft/Optimist.sol
jakim929: fixed!

smartcontracts (approved):

packages/contracts-periphery/contracts/foundry-tests/Optimist.t.sol
```diff
@@ -1,161 +1,413 @@
...
+    // Helps with EIP-712 signature generation
+    OptimistInviterHelper optimistInviterHelper;
+
+    // To test multicall for claiming and minting in one call
+    IMulticall3 multicall3;
+
+    address internal carol_baseURIAttestor;
+    address internal alice_allowlistAttestor;
+    address internal eve_inviteGranter;
+    address internal ted_coinbaseAttestor;
```
roninjin10, line 60: nit would have gone with `brian_coinbaseAttestor`

packages/contracts-periphery/contracts/universal/op-nft/Optimist.sol
```diff
@@ -105,17 +119,15 @@ contract Optimist is ERC721BurnableUpgradeable, Semver {
     }
 
     /**
-     * @notice Checks whether a given address is allowed to mint the Optimist NFT yet. Since the
-     *         Optimist NFT will also be used as part of the Citizens House, mints are currently
-     *         restricted. Eventually anyone will be able to mint.
+     * @notice Checks OptimistAllowlist to determine whether a given address is allowed to mint
+     *         the Optimist NFT yet. Since the Optimist NFT will also be used as part of the
+     *         Citizens House, mints are currently restricted. Eventually anyone will be able
+     *         to mint.
```
roninjin10, line 125:
```suggestion
     *         the Optimist NFT. Since the Optimist NFT will also be used as part of the
     *         Citizens House, mints are currently restricted. Eventually anyone will be able
     *         to mint.
```
nit

packages/contracts-periphery/contracts/universal/op-nft/Optimist.sol
```diff
@@ -6,40 +6,54 @@ import {
...
 contract Optimist is ERC721BurnableUpgradeable, Semver {
+    /**
+     * @notice Attestation key used by the attestor to attest the baseURI.
+     */
+    bytes32 public constant BASE_URI_ATTESTATION_KEY = bytes32("optimist.base-uri");
+
+    /**
+     * @notice Attestor who attests to baseURI.
+     */
+    address public immutable BASE_URI_ATTESTOR;
```
roninjin10, line 27: Unrelated to pr: In general I think using the attestation station for the base uri is unnecessary complexity we should remove

packages/contracts-periphery/contracts/foundry-tests/Optimist.t.sol
```diff
@@ -284,54 +506,142 @@ contract OptimistTest is Optimist_Initializer {
...
         assertEq(optimist.supportsInterface(iface721), true);
     }
+
+    /**
+     * @notice Checking that multi-call using the invite & claim flow works correctly, since the
+     *         frontend will be making multicalls to improve UX. The OptimistInviter.claimInvite
+     *         and Optimist.mint will be batched
+     */
+    function test_multicall_batchingClaimAndMint_succeeds() external {
+        uint256 inviterPrivateKey = 0xbeefbeef;
```
roninjin10, line 588: image

roninjin10 (approved):

packages/contracts-periphery/contracts/universal/op-nft/Optimist.sol
jakim929: Yeah, I didn't have context on this decision. Was this in order to showcase a use for the AttestationStation?

roninjin10: Yea I think we were just attempting to use attestations anywhere it could be used even if it didn't make sense

roninjin10: Pretty sure @smartcontracts lightly recommended removing it too but we didn't just because it would require some offchain code in @gateway/backend to change too

PR #5330, opened 2023-04-03
contracts-bedrock: fix semver tag

tynes:
**Description**

The docstring for `L1FeeVault` was not bumped
properly. This commit bumps the commit message.
We should have automation around this kind of thing to prevent it from happening in the future.


clabby (approved):
Good catch- LGTM

PR #5334, opened 2023-04-03
ci: Update op-e2e

trianglesphere:
**Description**

This fixes go based CI in several ways
1. Set test.parallel flag so it matches the actual amount of CPUs request
2. Add more calls to t.Parallel in op-e2e
3. Reduce the amount of blocks in TestBigL2Txs to reduce the test time.

CI Time for op-e2e goes from 16 min -> 4:30 min.

**Metadata**

- Fixes CLI-3762

### DIFF
--- a/op-e2e/actions/l2_batcher_test.go
+++ b/op-e2e/actions/l2_batcher_test.go
@@ -384,7 +384,7 @@ func TestExtendedTimeWithoutL1Batches(gt *testing.T) {
 }
 
 // TestBigL2Txs tests a high-throughput case with constrained batcher:
-//   - Fill 100 L2 blocks to near max-capacity, with txs of 120 KB each
+//   - Fill 40 L2 blocks to near max-capacity, with txs of 120 KB each
 //   - Buffer the L2 blocks into channels together as much as possible, submit data-txs only when necessary
 //     (just before crossing the max RLP channel size)
 //   - Limit the data-tx size to 40 KB, to force data to be split across multiple datat-txs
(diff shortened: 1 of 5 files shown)


op-e2e/actions/l2_batcher_test.go
```diff
@@ -428,7 +428,7 @@ func TestBigL2Txs(gt *testing.T) {
 	}
 
 	// build many L2 blocks filled to the brim with large txs of random data
-	for i := 0; i < 100; i++ {
+	for i := 0; i < 40; i++ {
```
trianglesphere, line 431: @protolambda I reduced the amount of blocks here b/c this test was slow + resource intensive

protolambda (approved):
amazing

ajsutton (approved):
LGTM. Nice work.  I've been running all the e2e tests locally a fair bit recently and IntelliJ runs them in parallel which has been quite reliable. I'd be surprised if there isn't some intermittency that turns up so we'll need to keep an eye out and fix it when it does, but it shouldn't be frequent from what I've seen running locally.

trianglesphere:
> LGTM. Nice work. I've been running all the e2e tests locally a fair bit recently and IntelliJ runs them in parallel which has been quite reliable. I'd be surprised if there isn't some intermittency that turns up so we'll need to keep an eye out and fix it when it does, but it shouldn't be frequent from what I've seen running locally.

So far with the changes I've been making it's been more reliable this time (though I will watch our for flakes). I think the big improvement is specifying the parallelism. We may have been hitting this CircleCI issue where it doesn't appear to be setting the `proc/cpu_count` value correctly: https://circleci.com/docs/configuration-reference/#resourceclass

mslipper (approved):

zhwrd (approved):

PR #5341, opened 2023-04-04
op-node: fix p2p sync client wait group problem

protolambda:
**Description**

This fixes a panic that may happen on shutdown  with a race condition. The sync client is not on by default, and it's only during shutdown, so the severity is relatively low.

What happens is that if a peer (re)connects while it's shutting down, the wg task counter increases, while at that point it's supposed to only go down. (resource context is cancelled, and Done() calls are bringing it back to 0)

So to fix the problem, we have to avoid adding any peers after we started the shutdown. I.e. use the `peersLock` to not just lock adding/removal of peers, but also a `closing` boolean, so that subsequent peers are not added after it's starting to shut down. That way we can safely `Wait()`.

Thanks @tynes for reporting the panic.

```
panic: sync: WaitGroup is reused before previous Wait has returned [recovered]
	panic: sync: WaitGroup is reused before previous Wait has returned
```

**TODOs**

Fix CLI-3764

### DIFF
--- a/op-node/p2p/sync.go
+++ b/op-node/p2p/sync.go
@@ -202,11 +202,18 @@ type SyncClient struct {
 
 	results chan syncResult
 
+	receivePayload receivePayloadFn
+
+	// resource context: all peers and mainLoop tasks inherit this, and start shutting down once resCancel() is called.
 	resCtx    context.Context
 	resCancel context.CancelFunc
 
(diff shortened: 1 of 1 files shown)


trianglesphere (approved):
lgtm. Thanks for the quick fix.

PR #5344, opened 2023-04-04
discord: update discord invite links

protolambda:
Update discord invite links.

CI was breaking because some of the docs, like the readme, are being checked for uptime, to make sure it's not a dead link. The `discord-gateway.optimism.io` site is no longer used, so those links need to be updated. I also changed the `https://discord.optimism.io` in the op-stack docs, since that one still redirects to the now disabled `discord-gateway`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ If you want to build Optimism, check out the [Protocol Specs](./specs/).
 
 ## Community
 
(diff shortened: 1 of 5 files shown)


zhwrd (approved):

jakim929 (approved):
lgtm!

PR #5345, opened 2023-04-04
op-batcher: Shutdown pprof and metrics after main loop returns

sebastianst:
**Description**

The batcher's main function currently stops pprof and metrics immediately when receiving any stop signal, like SIGINT. This PR swaps the order and let's it first wait for the main loop to return and only then cancel pprof and metrics.

The error from stopping the http server is now also logged.
 
**TODOs**

- [x] Author or reviewer has added an entry to the [current release notes draft][RND], if appropriate. [not appropriate]


trianglesphere (commented):
Seems like it should work.

trianglesphere (approved):

PR #5349, opened 2023-04-05
op-program: Implement oracle backed database

ajsutton:
**Description**

Adds an implementation of `ethdb.KeyValueStore` sufficient to be used for block processing.

Also adds separate oracle method to access code by hash to support the way Geth now adds a prefix to the hash when storing new code values in the database.

Removed the context from oracle methods and stored one as a property in the fetcher implementation. This isn't the typical golang pattern but it fits better here as we'll need to support a fully synchronous, local execution mode where having a context doesn't make sense and a context isn't passed into the `ethdb.KeyValueStore` methods.

**Tests**

Tested the database adapter methods directly for low lever behaviour, as well as a backing for an op-geth statedb to ensure the database behaviour is compatible with the real code using it.

**Metadata**

- Part of https://linear.app/optimism/issue/CLI-3741/l2-oracle-backed-execution-engine

### DIFF
--- a/op-program/l2/db.go
+++ b/op-program/l2/db.go
@@ -1,0 +1,91 @@
+package l2
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/ethdb/memorydb"
+)
+
+var codePrefixedKeyLength = common.HashLength + len(rawdb.CodePrefix)
+
+var (
+	ErrInvalidKeyLength = errors.New("pre-images must be identified by 32-byte hash keys")
+)
+
+type OracleKeyValueStore struct {
+	db     ethdb.KeyValueStore
+	oracle StateOracle
+}
+
+func NewOracleBackedDB(oracle StateOracle) *OracleKeyValueStore {
+	return &OracleKeyValueStore{
+		db:     memorydb.New(),
+		oracle: oracle,
+	}
+}
+
+func (o *OracleKeyValueStore) Get(key []byte) ([]byte, error) {
+	has, err := o.db.Has(key)
+	if err != nil {
+		return nil, fmt.Errorf("checking in-memory db: %w", err)
+	}
+	if has {
+		return o.db.Get(key)
+	}
+
+	if len(key) == codePrefixedKeyLength && bytes.HasPrefix(key, rawdb.CodePrefix) {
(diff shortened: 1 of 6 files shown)


op-program/l2/oracle.go
```diff
@@ -0,0 +1,8 @@
+package l2
+
+import "github.com/ethereum/go-ethereum/common"
+
+type StateOracle interface {
+	NodeByHash(nodeHash common.Hash) ([]byte, error)
+	CodeByHash(codeHash common.Hash) ([]byte, error)
+}
```
protolambda, line 8: The interface looks good to me, but note that in the implementation we may switch to mixing the hash with type data, to avoid collisions with other types of hashes (and potentially other types of oracle data like the bootstrap inputs).
A go-doc to clarify the usage (especially the expected caller, and error conditions) would be nice.

op-program/l2/db.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, fmt.Errorf("checking in-memory db: %w", err)
+	}
+	if has {
+		return o.db.Get(key)
+	}
+	source := o.oracle.NodeByHash
+	if bytes.HasPrefix(key, rawdb.CodePrefix) {
+		key = key[len(rawdb.CodePrefix):]
+		source = o.oracle.CodeByHash
+	}
```
protolambda, line 43: This looks unsafe: the node hashes are random, and so the code-prefix may overlap with an actual MPT node. We should check the prefix only if the key length is larger than 32 bytes I think.

protolambda (changes requested):
LGTM, but the prefix check looks unsafe, so I'll mark this as request-changes for now. If it's actually safe this way then a comment would help for later readers.

op-program/l2/oracle.go
ajsutton: I _think_ it will make sense to keep this API using `common.Hash` and the implementation of it that backs off to the real oracle will be responsible for mixing the type data into the hash. That way the details of the oracle access are encapsulated here and callers don't need to care about any of the details of how the oracle is actually called (including how the type is mixed into the hash, memory mapping vs sys call vs local etc). Not hard to change if we need to though.

And yeah will add godoc.

op-program/l2/db.go
ajsutton: Good catch - it's not safe.  Will fix and add tests specifically for that.

op-program/l2/oracle.go
ajsutton: Done. I'm more and more leaning towards remove the error return value from this and just return `nil` to indicate that data is unavailable rather than an error but don't want to get tied up with retry logic in the fetchers until I've got something working end to end.

protolambda (approved):

PR #5350, opened 2023-04-05
op-program: Extract block processing logic from engine api

ajsutton:
**Description**

Make the block processing logic easier to follow and reusable by extracting it from the engine api. Besides its current usage to create blocks in the engine api it will be used to import blocks in the oracle based engine backend. Only real difference between creating and importing blocks is that when importing you commit the changes to the DB and a couple of extra verifications that the output roots all match.

**Tests**

Mostly covered by existing tests since this is just rearranging existing functionality but added a couple around verifying the gas limit wasn't changed too much.

**Metadata**

- Part of https://linear.app/optimism/issue/CLI-3741/l2-oracle-backed-execution-engine

### DIFF
--- a/op-program/l2/engineapi/block_processor.go
+++ b/op-program/l2/engineapi/block_processor.go
@@ -1,0 +1,118 @@
+package engineapi
+
+import (
+	"errors"
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/consensus"
+	"github.com/ethereum/go-ethereum/consensus/misc"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+var (
+	ErrExceedsGasLimit = errors.New("tx gas exceeds block gas limit")
+	ErrUsesTooMuchGas  = errors.New("action takes too much gas")
+)
+
+type BlockDataProvider interface {
+	StateAt(root common.Hash) (*state.StateDB, error)
+	GetHeader(common.Hash, uint64) *types.Header
+	Engine() consensus.Engine
+	GetVMConfig() *vm.Config
+	Config() *params.ChainConfig
+	consensus.ChainHeaderReader
+}
+
+type BlockProcessor struct {
+	header       *types.Header
+	state        *state.StateDB
+	receipts     types.Receipts
+	transactions types.Transactions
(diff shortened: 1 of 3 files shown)


op-program/l2/engineapi/l2_engine_api.go
```diff
@@ -102,7 +92,7 @@ func computePayloadId(headBlockHash common.Hash, params *eth.PayloadAttributes)
 }
 
 func (ea *L2EngineAPI) RemainingBlockGas() uint64 {
-	return ea.l2GasPool.Gas()
+	return ea.blockProcessor.gasPool.Gas()
```
protolambda, line 95: Maybe return 0 if the `blockProcessor` is nil?

protolambda (approved):
Nice refactor. One nit about `RemainingBlockGas()`

op-program/l2/engineapi/block_processor.go
```diff
@@ -0,0 +1,123 @@
...
+	b.transactions = append(b.transactions, tx)
+	return nil
+}
+
+func (b *BlockProcessor) Assemble() (*types.Block, error) {
+	b.header.GasUsed = b.header.GasLimit - b.gasPool.Gas()
+	block, err := b.dataProvider.Engine().FinalizeAndAssemble(b.dataProvider, b.header, b.state, b.transactions, nil, b.receipts, nil)
+	if err != nil {
+		return nil, err
+	}
```
protolambda, line 110: Previously it just ran the `intermediateRoot` + `NewBlock` combination. I think the PoS implementation of `FinalizeAndAssemble` will do the same, but maybe worth documenting if there's actually a difference.

ajsutton: Right now there's no differnce, but it will also apply withdrawals if Shanghai is active. While we don't activate Shanghai currently the aim here is to reuse as much of the geth block building logic as we can. This is the standard method used in geth to complete block building once transactions have been applied so it should be updated for anything required in future forks as well.

I've also removed the updating of `header.GasUsed` from this assemble which isn't required. The `GasUsed` value is updated in `ApplyTransaction` as it goes along which I think makes it a bit clearer as well since then we're clearly just delegating to the geth code to finish off the block for us.

op-program/l2/engineapi/l2_engine_api.go
ajsutton: Yep makes sense.

op-program/l2/engineapi/block_processor.go
protolambda: Hmm, I suppose it's 0 by default, if `ApplyTransaction` is never called? So yes, can remove the part in sealing then.

protolambda (approved):

op-program/l2/engineapi/block_processor.go
ajsutton: Yes, starts at 0 and then `ApplyTransaction` increments it as it goes.

PR #5353, opened 2023-04-05
Migration script clean up

maurelian:

PR #5352, opened 2023-04-05
fix(ctb): Update hardhat.json to fix hh deploy

maurelian:
Adds a `phase1()` methods to the MSD which consolidates the various stepN() methods into two groups.

This is the first PR in a series that will significantly reduce the number of multisig transactions which need to be executed.

I tested this primarily by running the deploy scripts locally.

```
npx hardhat deploy --tags setup,phase1,phase2
```

PR #5354, opened 2023-04-05
feat(ctb): Add doOwnershipTransfer to deploy utils

maurelian:
1. Creates a reusable function `doOwnershipTransfer` which deduplicates the logic for transferring ownership of the 4 system contracts to the MSD.
2. Combines the `awaitCondition` call for each of those transfers into a single call at the end. This provides better support for a `multicall` based transfer, by printing out all the info required to create a batch of transactions at once. It also still supports a 'one at a time' transfer approach, as it will check all contracts before progressing.

### DIFF
--- a/packages/contracts-bedrock/deploy/020-SystemDictatorSteps-1.ts
+++ b/packages/contracts-bedrock/deploy/020-SystemDictatorSteps-1.ts
@@ -11,10 +11,8 @@ import {
   assertContractVariable,
   getContractsFromArtifacts,
   getDeploymentAddress,
+  doOwnershipTransfer,
   doPhase,
-  printJsonTransaction,
-  printTenderlySimulationLink,
-  printCastCommand,
 } from '../src/deploy-utils'
 
 const uint128Max = ethers.BigNumber.from('0xffffffffffffffffffffffffffffffff')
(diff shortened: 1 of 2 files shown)


maurelian:
* develop
  * **PR #5352** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5352" target="_blank">Graphite</a> 
    * **PR #5353** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5353" target="_blank">Graphite</a> 
      * **PR #5354** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5354" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5354?utm_source=stack-comment).

packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -322,6 +322,39 @@ export const printJsonTransaction = (tx: ethers.PopulatedTransaction): void => {
...
+ *
+ * @param opts Options for executing the step.
+ * @param opts.isLiveDeployer True if the deployer is live.
+ * @param opts.proxy proxy contract.
+ * @param opts.dictator dictator contract.
+ */
+export const doOwnershipTransfer = async (opts: {
+  isLiveDeployer?: boolean
+  proxy: ethers.Contract
+  name: string
```
maurelian, line 336: I don't like passing the name as an arg, I'm sure there's a better way, but good enough for now IMO.

packages/contracts-bedrock/deploy/020-SystemDictatorSteps-1.ts
```diff
@@ -73,10 +71,13 @@ const deployFn: DeployFunction = async (hre) => {
...
-    await ProxyAdmin.transferOwnership(SystemDictator.address)
-  } else {
-    console.log(`Proxy admin already owned by MSD`)
+    await doOwnershipTransfer({
+      isLiveDeployer,
+      proxy: ProxyAdmin,
+      name: 'ProxyAdmin',
+      transferFunc: 'transferOwnership',
+      dictator: SystemDictator,
+    })
```
maurelian, line 80: I recognize that the proxy admin is always owned by the `deployer` key as it is configured that way in the deployment scripts, so this doesn't use all the functionality of the new function, but I still think its cleaner.

tynes (approved):
This is a great addition, good work

PR #5356, opened 2023-04-05
contracts-bedrock: fix L2OutputOracle comment

tynes:
**Description**

The comment mistakenly said that the `pause` functionality would pause both deposits and withdrawals, while it only has the ability to pause withdrawals.

The `PostSherlock` deploy script is also updated. We should be able to make the script generic in the future so that it doesn't need to be upgraded over time.


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -153,7 +152,7 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         address _guardian,
         bool _paused,
         SystemConfig _config
-    ) Semver(1, 3, 0) {
+    ) Semver(1, 3, 1) {
```
maurelian, line 155: Do we update semvers on comments too?

maurelian (approved):

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: Yup
https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock#versioning

maurelian:
Pushed the commit from #5383 that should unblock this.

PR #5264, opened 2023-03-27
specs: DisputeGame interface + behavior

tynes:
**Description**

WIP dispute game interface specs based on work by @clabby 

Closes CLI-3479
Closes CLI-3478

PR #5360, opened 2023-04-06
feat(specs): Challenge Agent specification

clabby:
# Overview

Adds specification for an implementation of a challenge agent for the OP Stack.

> **Note**
> Merge after #5264, references and contains links to both of those specifications which is causing CI to fail.

**Metadata**

Closes CLI-3480


tynes:
There is a lowercase `l2` in the bottom right most box in the diagram. Also the split for the agent going between making the game and playing the game is a bit confusing, perhaps there is a way to have making the game be more obvious, like the step where its "check if the game exists, decision, make the game" then circle back around and "check if the game exists, decision, play the game" or something like that

specs/challenger.md
```diff
@@ -0,0 +1,88 @@
...
+  of Optimism, this is the Optimism Mainnet.
+- **rollup-node** - In the context of this document, the rollup node describes the
+  [rollup-node specification](./rollup-node.md). In the context of Optimism, this is the implementation
+  of the [rollup-node specification](./rollup-node.md), the `op-node`.
+
+## Event and Response Lifecycle
+
+The Challenger agent is expected to be able to listen for and respond to several different events
+on the data availability layer. These events and responses differ depending on the type of dispute game being
+played. For more information on the separate types of dispute games, see the
```
refcell, line 41: Verbiage nit

```suggestion
on the data availability layer. The events are parameterized depending on the
type of dispute game being played. Challenger agents respond in different ways
according to the different types of games. For specifications of game types, see the
```

specs/challenger.md
```diff
@@ -0,0 +1,88 @@
...
+[Dispute Game Interfaces specification](./dispute-game-interfaces.md) and
+[Dispute Game specification](./dispute-game.md).
+
+### `GameType.FAULT`
```
refcell, line 45: Would be nice if the order of the game types matched the ordering of the enum for consistency 🤠

refcell (changes requested):
Nice just minor suggestions!

specs/challenger.md
clabby: Totally- originally had it like this, but I figured that `VALIDITY` is far enough down the line that it could go at the bottom of the doc.

specs/challenger.md
```diff
@@ -0,0 +1,89 @@
...
+  - [`GameType.FAULT`](#gametypefault)
+  - [`GameType.ATTESTATION`](#gametypeattestation)
+  - [`GameType.VALIDITY`](#gametypevalidity)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Description
+
+The Challenger is an off-chain agent that listens for faulty claims made about the state of
+the L2 on the data availability layer. It is responsible for challenging these incorrect claims
```
tynes, line 19: Hmm i'm not sure if this is accurate. Technically the claims power the bridge which is the settlement layer

specs/challenger.md
refcell: Sure!

refcell (approved):
LGTM 👍

tynes (approved):

PR #5362, opened 2023-04-06
op-batcher: Sanity check block for non-empty transactions

Inphi:


tynes:
This is technically a hard fork

Inphi:
How so? Without the check, op-batcher will panic if there aren't any transactions.

ajsutton:
Given we access the first transaction in the block immediately after this to check its a deposit transaction, I'd expect batches with no transactions to already be rejected, just with a much less clear message.

Would be good if we can write a test for this though - running it without this extra check should reproduce the panic and with it would then return an error.

ajsutton (approved):
LGTM but will need a review from proto or Josh anyway. Given this is in the the derivation code does it trigger a reset of the code freeze period?

Inphi:
Technically (or not) it shouldn't. Because the `BlockToBatch` function is only used by the batcher. I'm unsure if the batcher is also subject to the code freeze.

protolambda (approved):

PR #5363, opened 2023-04-06
op-program: Integrate existing L1Client implementation to fetch L1 data

ajsutton:
**Description**

Adds CLI and config options to specify the L1 client info and creates a `L1Client` implementation to retrieve the data required by the derivation pipeline.

**Tests**

Added tests for each new CLI option and config options.

**Additional context**

The `L1Client` implementation isn't what will ultimately be used - we'll need to build an alternate `L1Fetcher` implementation that works with the pre-image oracle but it's a quick win to let us setup the required CLI args and will be enough to support running the actual derivation pipeline. That lets us get something working end to end asap.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3742/integrate-existing-l1fetcher


protolambda (approved):

protolambda:
rebased it to resolve merge conflicts. Old work is preserved in `aj/l1-fetcher-only-before`. Dropped one test case change, since I don't think we should let the L2 genesis-path be exposed as deep into the program, but rather load the genesis and provide the loaded genesis wherever possible as argument.

PR #5366, opened 2023-04-06
ci: make slither not fail CI

tynes:
**Description**

slither is halting CI. It thinks that yul is an antipattern. It is blocking PRs from being merged because of a `if (boolean == false)` check which is completely sane, sure you can do `if (!boolean)` but that is less readable. This commit prevents slither from halting CI. We can figure out how much we want to use slither in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -393,7 +393,7 @@ jobs:
           name: slither
           command: |
             slither --version
-            yarn slither
+            yarn slither || exit 0
           working_directory: packages/contracts-bedrock
 
   contracts-bedrock-validate-spaces:


tynes:
> Slither is passing on develop. Is there a specific PR which is causing it to fail?

There are a bunch of PRs, see https://github.com/ethereum-optimism/optimism/pull/5356
The reason it is failing incredibly lame and not helpful

clabby:
> > Slither is passing on develop. Is there a specific PR which is causing it to fail?
> 
> There are a bunch of PRs, see #5356 The reason it is failing incredibly lame and not helpful

This does look annoying, feels like something a linter should do rather than Slither. The suggestion does look like a valid lint, tho.

To break it down:
* `!boolValue` compiles to 2 bytes of asm:
```
DUP<N>  // [bool]
ISZERO  // [not_bool]
```
* `boolValue == false` compiles to 4 bytes of asm:
```
DUP<N>     // [bool]
PUSH1 0x00 // [false, bool]
EQ         // [bool == false]
```

The optimizer picks this up in the non-IR pipeline most of the time, though.

maurelian (approved):

trianglesphere (approved):
Approving b/c @maurelian approved.

maurelian:
I had a second thought, will update shortly

refcell:
This was fixed in #5356 by excluding `boolean_equality` in slither's config file at [slither.config.json#L2](https://github.com/ethereum-optimism/optimism/blob/bf8b130c94a1c7d676f411068eb170aea3d40a9f/packages/contracts-bedrock/slither.config.json#L2), correct @tynes?

PR #5368, opened 2023-04-06
op-wheel: add --allow-gaps flag to engine block building commands

protolambda:
Implement `--allow-gaps` block building option to use with the `op-wheel engine auto` and `op-wheel engine block` commands. This allows it to behave like skipped slots on the L1 beacon chain, where the timestamp can make larger jumps in slots, while the block height still increments by 1.

cc @maurelian for L1 part of migration testing.


maurelian (approved):
LGTM

PR #5369, opened 2023-04-06
op-e2e: fix p2p gossip flakiness

protolambda:
Small test setup fixes:
- Disable batcher to see L2 gossip effect in isolation. Adjusting the block time doesn't work well when it runs in a laggy environment.
- Increase timeouts.
- Only start sequencing after the verifiers are running, to give the verifiers a proper chance of receiving the first few payloads, instead of publishing those while the verifiers might not be fully up and running yet.

### DIFF
--- a/op-e2e/system_test.go
+++ b/op-e2e/system_test.go
@@ -589,9 +589,10 @@ func TestSystemMockP2P(t *testing.T) {
 	}
 
 	cfg := DefaultSystemConfig(t)
-	// slow down L1 blocks so we can see the L2 blocks arrive well before the L1 blocks do.
-	// Keep the seq window small so the L2 chain is started quick
-	cfg.DeployConfig.L1BlockTime = 10
+	// Disable batcher, so we don't sync from L1
+	cfg.DisableBatcher = true
+	// disable at the start, so we don't miss any gossiped blocks.
+	cfg.Nodes["sequencer"].Driver.SequencerStopped = true
 
 	// connect the nodes
 	cfg.P2PTopology = map[string][]string{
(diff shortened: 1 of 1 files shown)


op-e2e/system_test.go
```diff
@@ -634,11 +632,11 @@ func TestSystemMockP2P(t *testing.T) {
 	require.Nil(t, err, "Sending L2 tx to sequencer")
 
 	// Wait for tx to be mined on the L2 sequencer chain
-	receiptSeq, err := waitForTransaction(tx.Hash(), l2Seq, 6*time.Duration(sys.RollupConfig.BlockTime)*time.Second)
+	receiptSeq, err := waitForTransaction(tx.Hash(), l2Seq, 10*time.Duration(sys.RollupConfig.BlockTime)*time.Second)
 	require.Nil(t, err, "Waiting for L2 tx on sequencer")
 
 	// Wait until the block it was first included in shows up in the safe chain on the verifier
-	receiptVerif, err := waitForTransaction(tx.Hash(), l2Verif, 6*time.Duration(sys.RollupConfig.BlockTime)*time.Second)
+	receiptVerif, err := waitForTransaction(tx.Hash(), l2Verif, 10*time.Duration(sys.RollupConfig.BlockTime)*time.Second)
```
trianglesphere, line 639: Timeout may not have been increase enough. I tried increasing them to 1hr (effectively no timeout) in https://github.com/ethereum-optimism/optimism/pull/5373

protolambda:
@trianglesphere I found that the verifiers were missing the first few payloads because the sequencer was too quick to publish them, before the verifiers were running. I changed it now to start sequencing only later into the test.

PR #5373, opened 2023-04-06
op-e2e: Remove timeouts in TestSystemMockP2P

trianglesphere:
**Description**

This will hopefully make this test a little less flakey.


protolambda:
See other PR, I don't think it's the timeout that makes it less flaky, it's primarily the race condition on receiving the first block or not, depending on startup speed of verifier and sequencer. Increasing the timeout without disabling the batcher just makes it fall back to L1.

PR #5374, opened 2023-04-06
op-e2e: 1s L2 block time

trianglesphere:
**Description**

This goes from 2s to 1s L2 block times. This is more realistic to have multiple L2 blocks per L1 blocks and it also shaves another minute off of CI. ty to @sebastianst for suggesting this.


protolambda (changes requested):
Once the L1 time == L2 time, and the L2 chain falls behind, catching up with L1 becomes quite difficult, as there's no way to reduce the sequencer time drift anymore. Changing it to 1s may result in long-term problems, and might also result in flakiness (if temporary failures are less likely to fully recover during the execution of a test). 

In circle CI we only seem to gain 1 min in the op-e2e tests, out of ~4.5 min, while the two devnet jobs and integration testing is still a full 15 min each or so.

Can you add a description of why you think the change to 1 second is safe? I would prefer to close this PR and focus on the devnet together to improve CI time.

protolambda (approved):

PR #5376, opened 2023-04-06
op-node: Log when reading channels

trianglesphere:
**Description**

Logs when reading from a channel to match the log when it creates the channel.

### DIFF
--- a/op-node/rollup/derive/channel_bank.go
+++ b/op-node/rollup/derive/channel_bank.go
@@ -124,6 +124,7 @@ func (cb *ChannelBank) Read() (data []byte, err error) {
 	if !ch.IsReady() {
 		return nil, io.EOF
 	}
+	cb.log.Info("Reading channel", "channel", first, "frames", len(ch.inputs))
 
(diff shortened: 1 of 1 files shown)


op-node/rollup/derive/channel_bank.go
```diff
@@ -124,6 +124,7 @@ func (cb *ChannelBank) Read() (data []byte, err error) {
 	if !ch.IsReady() {
 		return nil, io.EOF
 	}
+	cb.log.Info("Reading channel", "channel", first, "frames", len(ch.inputs))
```
protolambda, line 127: Should we change these logs to debug? IMHO it's a bit too verbose for the regular user.

protolambda (approved):
LGTM. Consider changing the log level though

op-node/rollup/derive/channel_bank.go
trianglesphere: The open log is info. Adding individual frames is debug.

PR #5385, opened 2023-04-07
op-node: Add channel_input_bytes metric

Inphi:
Adds gauge metric for the amount of compressed bytes read from the channel just prior to batch decoding.

This is useful when paired with the batcher `output_bytes` to get a better sense on the data rates throughout the system.

### DIFF
--- a/op-node/metrics/metrics.go
+++ b/op-node/metrics/metrics.go
@@ -64,6 +64,7 @@ type Metricer interface {
 	RecordSequencerBuildingDiffTime(duration time.Duration)
 	RecordSequencerSealingTime(duration time.Duration)
 	Document() []metrics.DocumentedMetric
+	RecordChannelInputBytes(num int)
 	// P2P Metrics
 	SetPeerScores(scores map[string]float64)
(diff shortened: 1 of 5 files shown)


sebastianst (approved):
lgtm

op-node/metrics/metrics.go
```diff
@@ -331,6 +334,12 @@ func NewMetrics(procName string) *Metrics {
 			"direction",
 		}),
 
+		ChannelInputBytes: factory.NewGauge(prometheus.GaugeOpts{
```
protolambda, line 337: With a gauge you'll only find the latest scraped number in the metrics. A data tx alone may contain more than one channel.
To get accurate metrics we should use a histogram, or a counter that only ever increases (and graph the rate of increase)

protolambda (changes requested):
With gauge + set metric functionality we'll miss data, and get inaccurate rate metrics

op-node/metrics/metrics.go
Inphi: You're right. Counter makes better sense here. Will fix.

protolambda (approved):

PR #5391, opened 2023-04-07
feat(docs/op-stack): Add op-proposer

qbzzt:
In addition:
- Change `op-geth` to run in archive mode
- Usability improvements

Closing: https://linear.app/optimism/issue/DEVRL-837/add-op-proposer-to-op-stack-docs


docs/op-stack/src/docs/build/getting-started.md
```diff
@@ -420,46 +451,63 @@ Once you have multiple nodes, it makes sense to use these command line parameter
...
+```bash
+cd ~/optimism/op-proposer
+
+./bin/op-proposer \
+    --poll-interval 12s \
+    --rpc.port 8560 \
+    --rollup-rpc http://localhost:8547 \
+    --l2oo-address $L2OO_ADDR \
+    --private-key $PROPOSER_KEY \
+    --allow-non-finalized \
```
smartcontracts, line 501: Including this by default might be a bit of a footgun

smartcontracts: Even with the comment below

smartcontracts (commented):
Some small comments. In the future would be great to get these broken up into smaller PRs that have a tightly scoped purpose -- in this case I think the environment variable work could be separated from the addition of the op-proposer content. Not a big deal but will be very useful going forward to speed up review overall.

docs/op-stack/src/docs/build/getting-started.md
qbzzt: You mean `--allow-non-finalized`? I worry that if I don't we'll see people unhappy that `op-proposer` is so slow.

qbzzt: OK, I'll remove it.

smartcontracts (approved):

PR #5392, opened 2023-04-08
op-program: split into client and host packages

protolambda:
**Description**

Split the program into client and host, so we can isolate the client functionality (pure state-transition, onchain) better from the host (CLI and VM modes).

**Tests**

Just file moves and minor import path changes, no functionality changes.


ajsutton (approved):
LGTM. It won't be a clean split until we support loading the config via the oracle but these modules make sense.

PR #5393, opened 2023-04-08
op-program: pre-image oracle (work in progress)

protolambda:
**Description**

Based on top of #5392 

This implements a pre-image oracle interface, and implements the `Oracle` interfaces in l1 and l2 client packages as `PreimageOracle` that interacts with the preimage oracle to get the data.

This also includes "hinting": to indicate to the host what data is of interest for the next pre-image fetching. This can be a no-op by the host in the onchain environment, or when running in a read-only mode from a KV store of known preimages.

**Tests**

Work in progress.

### DIFF
--- a/op-program/client/l2/oracle.go
+++ b/op-program/client/l2/oracle.go
@@ -1,8 +1,13 @@
 package l2
 
 import (
+	"fmt"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rlp"
+
+	"github.com/ethereum-optimism/optimism/op-program/preimage"
 )
 
 // StateOracle defines the high-level API used to retrieve L2 state data pre-images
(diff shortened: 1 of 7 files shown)


op-program/client/l2/oracle.go
```diff
@@ -26,3 +31,41 @@ type Oracle interface {
...
+func NewPreimageOracle(raw preimage.Oracle, hint preimage.Hinter) *PreimageOracle {
+	return &PreimageOracle{
+		oracle: raw,
+		hint:   hint,
+	}
+}
+
+func (p *PreimageOracle) BlockByHash(blockHash common.Hash) *types.Block {
+	p.hint.Hint(L2BlockHint(blockHash))
+	blockRlp := p.oracle.Get(preimage.Keccak256Key(blockHash))
```
ajsutton, line 53: One important thing to note here - `blockHash` is actually the hash of the block _header_ not the full block (as is always the case in Ethereum). So we'll have to retrieve the block header RLP then get the transactions based on the TxRoot and then reassemble the block. We can hard code empty uncles and withdrawals.

protolambda:
Closing in favor of the new improved pre-image oracle work in #5449

PR #5397, opened 2023-04-09
op-node: l1 and l2 block content metrics

protolambda:
**Description**

Metrics of the L1 and L2 blocks that the op-node processes, to analyze usage and data throughput.

Note: the tx-decoding for the L2 block metrics may reduce performance slightly. I'm considering moving these metrics behind a feature-flag.


protolambda:
Closing in favor of potential op-exporter project

PR #5398, opened 2023-04-10
[op-batcher] Add support for multiple batcher transactions per L1 block

mdehoog:
**Description**

Adds support for sending multiple transactions from the batcher at once. A new flag has been added: `--max-pending-tx`, which controls the number of pending transactions sent to the `txmgr`.

The `txmgr` was refactored to allow concurrent calls to `Send`. Some internal nonce tracking code was added that manually increments the nonce for each tx (after initially pulling it from an L1 node). If any of the transaction sends experiences an error, the entire set of pending txs is canceled, and the cached nonce is cleared.

Note: currently the batcher only supports a single pending channel, which means that only when channels have multiple frames will multiple txs be submitted at once. Multiple pending channels will come in a later PR.

**Tests**

Add tests for txmgr nonce management + reset, as well as tests for the new `TxQueue`. 

**Additional context**

At times we are experiencing more L2 state than the batcher can keep up with, and our safe chain is falling behind our unsafe chain. This PR will allow a greater throughput of batcher transactions by allowing multiple txs per L1 block.

Here's an [example block](https://goerli.etherscan.io/txs?block=8856605&p=1) that has 10 transactions from our batcher ([0x53394266fc80e5d4e6d25b3d0b7ca243859b7b09](https://goerli.etherscan.io/address/0x53394266fc80e5d4e6d25b3d0b7ca243859b7b09)).

**TODOs**


tynes:
I feel like the better way to implement this is in the `TxManager`, having it support concurrent transaction submission. Then any service that uses the `TxManager` has this ability. It also feels a bit weird to cap the amount for max transactions per block. Does that actually mean "max pending transactions"? I didn't go super in depth into the implementation here to be honest

mdehoog:
@tynes totally agree 👍 will refactor

tynes:
> @tynes totally agree +1 will refactor

I think @trianglesphere has been thinking about this

trianglesphere (commented):
I've been thinking about this as well & was going to start work on this as well this week. Some of the refactors we've made to the batcher make it slightly harder to parallelize, but it should still be relatively easy to do async tx sending & async tx confirmation. There might be some slight difficulty with shutdown, but only `driver.go` should have to be modified in the batcher.

op-batcher/batcher/config.go
```diff
@@ -76,6 +77,9 @@ type CLIConfig struct {
 	// and creating a new batch.
 	PollInterval time.Duration
 
+	// MaxTxPerBlock is the maximum number of batcher transactions to submit in single L1 block.
+	MaxTxPerBlock uint64
```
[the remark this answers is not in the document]
mdehoog: good suggestion, done

op-service/txmgr/txmgr.go
```diff
@@ -126,8 +137,52 @@ type TxCandidate struct {
...
+	receipt, err := m.send(ctx, candidate)
+	if err != nil {
+		m.reset()
+	}
+	return receipt, err
+}
+
+// reset the transaction manager. All currently pending transactions will receive
+// a ResetErr error. This is called if any pending send returns an error.
+func (m *SimpleTxManager) reset() {
```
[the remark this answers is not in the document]
mdehoog: yeah, reset is called whenever any of the transactions experiences a non-recoverable error, i.e. when `sendState.ShouldAbortImmediately()` or if the context is canceled. It is called here: https://github.com/ethereum-optimism/optimism/blob/aa7350864d12264faac8754b56822095018b468c/op-service/txmgr/txmgr.go#L146

op-batcher/batcher/job_runner.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Wait waits on all running jobs to stop.
+func (s *JobRunner) Wait() {
+	s.wg.Wait()
+}
+
+// Run will wait until the number of running jobs is below the max concurrency,
+// and then run the next job. The JobFactory should return `nil` if the next
+// job does not exist. Returns the error returned from the JobFactory (if any).
+func (s *JobRunner) Run(factory JobFactory) error {
```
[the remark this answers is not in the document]
mdehoog: The intention of the runner not taking a context is that the jobs themselves should manage their own context... this runner blocks until jobs are canceled. Similar to a wait group, `Wait()` does not take a context.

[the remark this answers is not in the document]
mdehoog: If understand your question, you are asking why this is a "function that, when called, creates the job function", and why can't it just take the job to be run?

We could totally do that. Couple of reasons the implementation is like this:
 - it means the job creation can return an error, which is returned by `Run()`
 - it allows job setup (and the potential for a job to not be available) to happen in the factory

In the batcher's case, when calling `factory()`, it runs the `publishStateToL1Job` setup code, i.e. calling the `l1Tip`,  and calling `state.TxData` to see if there's any transaction data available. If not, the factory will return `nil` indicating that there's no job to run. Otherwise it returns a job `func()` that is run below in the go routine, which performs the transaction send.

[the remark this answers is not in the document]
trianglesphere: I guess `Run` taking a context makes more sense if you have take `JobFn(ctx)` instead of the factory.

op-batcher/batcher/driver.go
```diff
@@ -284,87 +285,129 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
+		// keep publishing state until either:
+		//  - we've drained all pending data (EOF)
+		//  - an error occurs
+		//  - we get killed by the context
+		for {
+			select {
+			case <-l.killCtx.Done():
+				return
+			default:
+				err := queue.Send(l.killCtx, l.publishStateToL1Factory(), receiptsCh)
```
trianglesphere, line 333: Can you provide `l.publishStateToL1Job` instead of going through the factory?

op-service/txmgr/queue.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+type TxFactory[T any] func(ctx context.Context) (*TxCandidate, T, error)
+
+type Queue[T any] struct {
+	txMgr          TxManager
+	maxPending     uint64
+	pendingChanged func(uint64)
+	pending        uint64
+	cond           *sync.Cond
```
trianglesphere, line 22: I think that we could use a semaphore (`go/x/` [impl](https://pkg.go.dev/golang.org/x/sync/semaphore)) or a channel based version to simplify the internals here.

trianglesphere (commented):
Nice. I really like the `Queue` interface & how it's used. Inside the queue, I think that code is correct, but I don't think we need the full complexity of `sync.Cond` there.

op-service/txmgr/queue.go
```diff
@@ -0,0 +1,103 @@
+package txmgr
+
+import (
+	"context"
+	"github.com/ethereum/go-ethereum/core/types"
```
trianglesphere, line 5: lint is failing b/c file is not go-imported.

op-batcher/batcher/driver.go
mdehoog: the reason for doing it through the factory is that `publishStateToL1Job` is called lazily only if there's room in the pending pool for the tx. If we called it directly, we would have a bunch of local state and nowhere to put it, including an `l1Tip` that may be out-of-date when the tx queue allows the tx later

op-service/txmgr/queue.go
mdehoog: finally thought i found a good use for `sync.Cond` 😄  but `Semaphore` makes much more sense... thanks

op-batcher/batcher/job_runner.go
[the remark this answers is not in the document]
mdehoog: the Queue refactor made `ctx` a parameter (and it's required by the semaphore anyway)

op-service/txmgr/txmgr.go
trianglesphere, line 153: I just had another idea here that requires a little more thinking about the API. We could use an `ErrGroup` to manage the resets (because if you use `group.WithContext` any failure will propagate to the other transactions which should cancel them. Then you'll need to figure out how to reset the nonce (not sure if it should be done internally or externally).

There's another open question about if the `ErrGroup` is inside the `SimpleTxManager` or inside `TxQueue` (errgroups also have a semaphore like functionality). If the errgroup is inside the `TxQueue` there's a last question of how to handle the reset (internal vs external vs the queue taking the simpleTxMgr or taking an interface)

op-service/txmgr/txmgr.go
```diff
@@ -84,6 +87,13 @@ type SimpleTxManager struct {
 	backend ETHBackend
 	l       log.Logger
 	metr    metrics.TxMetricer
+
+	nonce     *uint64
+	lock      sync.RWMutex
+	wg        sync.WaitGroup
+	resetC    chan struct{}
+	resetWG   sync.WaitGroup
+	resetting atomic.Bool
```
ajsutton, line 96: There's a fair few fields being added here to control various aspects of concurrency.  It would probably be helpful to add some comments to explain what they're actually used for.

It certainly is a lot of concurrency management all tied to the way reset works. If the `ErrGroup` idea @trianglesphere mentioned below could simplify things that would definitely be a win. The current code works, it's just hard to maintain code that has multiple overlapping concurrency controls like this because the interactions can be hard to reason about.

mdehoog: thanks, agree 👍  I removed all of the concurrency management from `txmgr` except for the nonce + lock

op-service/txmgr/queue.go
```diff
@@ -27,34 +27,37 @@ type Queue[T any] struct {
...
+		maxPending = math.MaxInt
 	}
-	var s *semaphore.Weighted
+	group, cxt := errgroup.WithContext(context.Background())
 	if maxPending > 0 {
-		// only create a semaphore for limited-size queues
-		s = semaphore.NewWeighted(int64(maxPending))
+		group.SetLimit(int(maxPending))
+	} else {
+		group.SetLimit(-1)
```
trianglesphere: No need to set the limit if there is none - it defaults to no limit.

op-batcher/batcher/driver.go
```diff
@@ -284,86 +287,130 @@ func (l *BatchSubmitter) calculateL2BlockRangeToStore(ctx context.Context) (eth.
...
 	for {
 		select {
-		case <-ticker.C:
+		case <-loadTicker.C:
 			l.loadBlocksIntoState(l.shutdownCtx)
-			l.publishStateToL1(l.killCtx)
+		case <-publishTicker.C:
+			_, _ = queue.TrySend(l.killCtx, l.publishStateToL1Factory(), receiptsCh)
+		case r := <-receiptsCh:
+			l.handleReceipt(r)
```
sebastianst, line 305: Is it necessary to decouple tx sending from block loading? Was it added to load the L1Tip only at the time of tx creation? I guess that's also the reason why there's lazy-generation of TxCandidates with a factory?

Could we instead, similarly to before, iterate over `l.state.TxData` and then call `queue.Send`? I think it's also fine to block if the queue is full, we wouldn't gain anything from loading L2 blocks earlier anyways. If that's possible, it could simplify a few things, like not needing to work with an injected factory or `TrySend`.

op-service/txmgr/queue.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return err
+}
+
+// mergeWithGroupContext creates a new Context that is canceled if either the given context is
+// Done, or the group context is canceled. The returned CancelFunc should be called once finished.
+//
+// If the group context doesn't exist or has already been canceled, a new one is created after
+// waiting for existing group threads to complete.
+func (q *Queue[T]) mergeWithGroupContext(ctx context.Context) (context.Context, context.CancelFunc) {
```
sebastianst, line 128: I have an intuition that we can get rid of `mergeWithGroupContext` if we add a life-time context to the `Queue` struct (would use `l.killCtx` when we initialize it in the batcher) and then derive the errgroup using [`WithContext`](https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext). So we basically flip the order of contexts around and derive the errgroup from that one context always. Since we always pass the `killCtx` to `Send`/`TrySend` anyways, this would have the same end result but without `mergeWithGroupContext`.

mdehoog: Good call out, refactored 👍

op-service/txmgr/queue.go
```diff
@@ -0,0 +1,154 @@
...
+	go func() {
+		defer cancel()
+		select {
+		case <-groupContext.Done():
+		case <-cl:
+		}
+	}()
+	return ctx, func() {
+		close(cl)
+	}
```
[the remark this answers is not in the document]
mdehoog: You're right, not sure why I thought cancel could only be called once

op-batcher/batcher/driver.go
mdehoog: The original reason was to ensure that L1Tip was up-to-date, but because this is only done for channel duration timeouts, if the channel manager returns a tx to send it is either because the channel already timed out or was full, so it's fine for the L1 tip to be a little behind in this case.

I refactored to remove the factory and just use `queue.Send`.

op-batcher/batcher/channel_manager.go
```diff
@@ -199,6 +199,11 @@ func (s *channelManager) TxData(l1Head eth.BlockID) (txData, error) {
 		return txData{}, io.EOF
 	}
 
+	// we have blocks, but we cannot add them to the channel right now
+	if s.pendingChannel != nil && s.pendingChannel.IsFull() {
+		return txData{}, io.EOF
+	}
```
mdehoog, line 205: this is now required because we call `TxData` when the `pendingChannel` is closed, which will return an "already closed" error when calling `outputFrames` below

op-batcher/batcher/job_runner.go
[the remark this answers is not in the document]
mdehoog: Removed the factory, because the `l1Tip` being up-to-date doesn't particularly matter if the txData is returned (because the channel was full or the max channel duration was hit if that is the case)

op-batcher/batcher/driver.go
```diff
@@ -286,84 +287,114 @@ func (l *BatchSubmitter) loop() {
...
 
-		l1tip, err := l.l1Tip(ctx)
-		if err != nil {
-			l.log.Error("Failed to query L1 tip", "error", err)
+	for {
+		select {
+		case r := <-receiptsCh:
+			l.handleReceipt(r)
+		case <-txDone:
 			return
```
sebastianst, line 341: I think there's a possible race here after the last tx has been sent. Unlikely, but `txDone` might be closed very close to when the last receipt arrives here. Then the last receipt might not get processed.

And, is there receipt handling in `publishStateToL1` only to cover the case where the shutdown ctx is closed? The main loop also processes receipts.

sebastianst (changes requested):
Generally looks good now!

op-batcher/batcher/driver.go
```diff
@@ -286,84 +287,114 @@ func (l *BatchSubmitter) loop() {
 
 	ticker := time.NewTicker(l.PollInterval)
 	defer ticker.Stop()
+
+	receiptsCh := make(chan txmgr.TxReceipt[txData])
```
nolanxyg, line 291: hi, is it better to use buffered channel?

sebastianst: I also thought about that, but I think it's not necessary here. It changes the behavior of `Queue.Wait`. Right now, it is guaranteed to only return after the last receipt has been read from the channel, because it's non-buffered. If we add a buffer here, `Queue.Wait` would return earlier. It depends on what behavior we want `Wait` to have. And if we want buffered behavior instead, the buffer needs to be large enough, or behavior might be unpredicted.

op-batcher/batcher/driver.go
mdehoog: When `drain=true`, because it's an unbuffered channel, the `queue.wait` will only return once all the receipts have been processed.

When `drain=false`, it doesn't particularly matter if we miss handling a tx in the channel here, because we also handle receipts in the main for loop in the `loop` function. That is the reason for having the receipt handling in both places.

We need to handle receipts here, because otherwise this method can deadlock if the max pending txs is not large enough to handle all of the txs.

An alternative is to remove the receipt handling here, and block this method from returning until all txs are complete, but this current implementation is more compatible with the future state of multiple pending channels.

sebastianst: Thanks, makes sense! I also realized that I answered my question myself in the comment to @nolanxyg below 😄

sebastianst (approved):
LGTM now! Just a nit about a redundant check in tests.

op-service/txmgr/queue_test.go
```diff
@@ -0,0 +1,241 @@
...
+				}()
+			}
+			// wait for the queue to drain (all txs complete or failed)
+			queue.Wait()
+			duration := time.Since(start)
+			// expect the execution time within a certain window
+			now := time.Now()
+			require.WithinDuration(t, now.Add(test.total), now.Add(duration), 500*time.Millisecond, "unexpected queue transaction timing")
+			require.Greater(t, duration, test.total, "test was faster than expected")
+			require.Less(t, duration, test.total+500*time.Millisecond, "test was slower than expected")
```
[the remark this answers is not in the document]
mdehoog: whoops sorry, i forgot to remove this

PR #5404, opened 2023-04-10
op-batcher: Add Channel Bytes Counters

Inphi:
**Description**

Adds a counters for bytes input and output to/from a channel. Note that this is different from the existing gauge metrics that tracks the current number of bytes written to and read from a channel. The total Counter lets us more accurate compute bitrate of outputs.

**TODOs**


op-batcher/metrics/metrics.go
```diff
@@ -220,6 +228,10 @@ func (m *Metrics) RecordChannelClosed(id derive.ChannelID, numPendingBlocks int,
 	m.ChannelInputBytes.WithLabelValues(StageClosed).Set(float64(inputBytes))
 	m.ChannelOutputBytes.Set(float64(outputComprBytes))
 
+	outputBytesDelta := outputComprBytes - m.lastChannelOutputBytes
+	m.lastChannelOutputBytes = outputComprBytes
```
trianglesphere, line 232: `outputComprBytes` comes from `s.pendingChannel.OutputBytes()` and it should only be called once per channel so I don't think we need to subtract the previous count.

Inphi: yup. I'll fix that.
I'm also wondering if we can just use the existing gauge to compute the output data rate. I'm playing with promql's `deriv` function to see if could work instead.

trianglesphere: IMO it's better to have a counter & then compute the diff/rate to figure out what the guage would have read. I think we should add this counter.

op-batcher/metrics/metrics.go
```diff
@@ -144,6 +146,16 @@ func NewMetrics(procName string) *Metrics {
 			Help:      "Compression ratios of closed channel.",
 			Buckets:   append([]float64{0.1, 0.2}, prometheus.LinearBuckets(0.3, 0.05, 14)...),
 		}),
+		ChannelInputBytesTotal: factory.NewCounter(prometheus.CounterOpts{
+			Namespace: ns,
+			Name:      "input_bytes_total",
+			Help:      "Total number of bytes to a channel.",
+		}),
+		ChannelOutputBytesTotal: factory.NewCounter(prometheus.CounterOpts{
```
sebastianst, line 154: I propose we make these `CounterVec`s with 1 dimension `"stage"` and then below record at `StageClosed`. Then when we later want to record metrics when channels are submitted, or L2 blocks are loaded (only input bytes), we don't need to change the counter vecs again.

trianglesphere: TBH I'm not a huge fan of this. My reading of how prometheus [suggests to use labels](https://prometheus.io/docs/practices/naming/) is that the sum of `output_bytes_total` across the different stages isn't really meaningful. I think it makes more sense to have different counters than partition by label.

Inphi: I think @trianglesphere is right. Not all stages are comparable, including channel outputs and inputs since only one of them is compressed. The only thing vectors give us is to be able to automatically add new metrics to a grafana panel witohut creating a new query. But in this case, it's not used for the primary usecase for vectors/labels; aggregating and segmenting metrics.

sebastianst: There might be a misunderstanding here. I never meant to merge these two `Counter`s into one `CounterVec`, because that would indeed not make sense. Mixing input and output bytes in a single counter vec and labelling them "input" and "output" would be a bad idea, because these are not comparable values, it wouldn't be "meaningful" to put them both into the same `CounterVec`.

What I meant was to change both into a `CounterVec` with a single dimension `"stage"` and then add label `StageClosed` where there are currently recorded. The stage here refers to the stages "loaded" (into blocks queue), "added" (to open channel), "closed" (channel) and "submitted" (channel fully included on L1) because that's the stages that data flows through the batcher.

This has turned into a debate on principles that I suggest we continue in private. So happy to just merge this PR the way it is and we can adapt later when we have agreed on a common usage of vector metrics.

PR #5406, opened 2023-04-10
op-node/doc: Document op-batcher & op-proposer metrics

trianglesphere:
**Description**

This extends the metrics doc tool to include op-batcher & op-proposer metrics.

**Metadata**

- Fixes CLI-3795


op-node/Dockerfile
```diff
@@ -6,6 +6,8 @@ RUN apk add --no-cache make gcc musl-dev linux-headers git jq bash
 
 # build op-node with the shared go.mod & go.sum files
 COPY ./op-node /app/op-node
+COPY ./op-batcher /app/op-batcher
+COPY ./op-proposer /app/op-proposer
```
protolambda, line 10: I do not believe these should be dependencies of the op-node, since the op-batcher and op-proposer already rely on the op-node. The op-batcher and op-proposer codebases are not meant to leak into user land, even if it's only metrics.

protolambda (changes requested):
Can you add it as sub-command to the batcher and proposer CLI commands instead? That way we keep the codebases nicely separated.

trianglesphere:
> Can you add it as sub-command to the batcher and proposer CLI commands instead?

Done

tynes (approved):

PR #5408, opened 2023-04-10
op-e2e: Increase TestMockP2P timeout

trianglesphere:


op-e2e/system_test.go
```diff
@@ -640,11 +640,11 @@ func TestSystemMockP2P(t *testing.T) {
 	require.Nil(t, err, "Sending L2 tx to sequencer")
 
 	// Wait for tx to be mined on the L2 sequencer chain
-	receiptSeq, err := waitForTransaction(tx.Hash(), l2Seq, 10*time.Duration(sys.RollupConfig.BlockTime)*time.Second)
+	receiptSeq, err := waitForTransaction(tx.Hash(), l2Seq, time.Hour)
```
ajsutton, line 643: We probably shouldn't need to wait an hour for the tx to be included. High timeouts are good but circleci will likely timeout before we hit this limit and we then wouldn't get any error message etc which makes debugging harder.

Maybe something more like 5 minutes? Assuming we have something like a 2s block time that's still a whole lot of blocks.

trianglesphere: TBH I was trying to disable the timeout entirely. If this test hangs go will kill it after 20m

trianglesphere: Timeout is set here: https://github.com/ethereum-optimism/optimism/blob/dca2cb6351ac2b6df623c5825daacab6f88a5ae0/.circleci/config.yml#L657

Maybe we should go back to the golang default of 10m because the tests are now parallel. We only need the 20m if we run sequentially.

ajsutton: Hmm, I think it's nice if the test fails rather than times out (whether that's go or circleci) because you usually get better failure messages that way which can point you to what the test was waiting on etc. I'm not against this if we're seeing the timeout hit incorrectly or even just want to see if it helps, just seems odd to me that the tx can be delayed so significantly but that we'd still expect it to eventually be included.

ajsutton (approved):

op-e2e/system_test.go
trianglesphere: Set to 5m to be under the default go timeout.

PR #5412, opened 2023-04-11
specs: initial fault proof program/VM specs

protolambda:
**Description**

Initial specifications of pre-image oracle protocol, fault-proof program, and fault-proof VM, and a brief section about the interactive dispute game for context.

Includes a diagram of the op-program host/client architecture discussed with Adrian, updated to better separate the op-program pre-image fetching, and the part that is common between the op-program test Host and the fault proof VM responsibilities.

The pre-image oracle protocol is likely to change, but this after a few iterations and implementation prototypes I'm quite confident in this approeach:
- prefixed hashes to separate domains, handle updates and keep the common keccak256 case simple and readable.
- prefixed hash for the input-pre-image type, used during bootstrapping, local to the dispute
- a draft of what a key-type for permissionless application-layer pre-image types could look like
- the hinting abstraction works really well, it separates the program pre-fetching and the error handling.

**Metadata**

Fix CLI-3796


specs/fault-proof.md
```diff
@@ -0,0 +1,412 @@
...
+while keeping the rest of the key "readable" (matching the original hash).
+
+#### Type `3`: Global generic key
+
+Reserved. This scheme allows for unlimited application-layer pre-image types without fault-proof VM redeployments.
+
+This is a generic version of a global key store: `key = 0x03 ++ keccak256(x, sender)[1:]`, where:
+
+- `key` is a `bytes32`, which can be a hash of an arbitrary-length type of cryptographically secure commitment.
+- `sender` is a `bytes32` identifying the pre-image inserter address (left-padded to 32 bytes)
```
ajsutton, line 109: What's `x` in this case? I think maybe it's actually being described as `key`.

specs/fault-proof.md
```diff
@@ -0,0 +1,412 @@
...
+
+Users may standardize around upgradeable external pre-image contracts,
+in case the implementation of the verification of the pre-image is expected to change.
+
+The store update function is `update(key bytes32, offset uint64, span uint8, value bytes32)`:
+
+- Only part of the pre-image, starting at `offset`, and up to (incl.) 32 bytes `span` can be inserted at a time.
+- Pre-images may have an undefined length (e.g. a stream), we only need to know how many bytes of `value` are usable.
+- The key and offset will be hashed together to uniquely store the `value` and `span`, for later pre-image serving.
+
```
ajsutton, line 125: Do we need to implement this specific approach to breaking down data here or can we let the global contract accept any size data and programs can choose to break it up using their own key scheme as needed?

specs/fault-proof.md
```diff
@@ -0,0 +1,412 @@
...
+The `<length prefix>` here may be arbitrarily high:
+the client can stop reading at any time if the required part of the pre-image has been read.
+
+After the client writes new `<request>` bytes, the server should be prepared to respond with
+the pre-image starting from `offset == 0` upon `read` calls.
+
+The server may limit `read` results artificially to only a small amount of bytes at a time,
+even though the full pre-image is ready: this is expected regular IO protocol,
+and the client will just have to continue to read the small results at a timme,
```
ajsutton, line 214:
```suggestion
and the client will just have to continue to read the small results at a time,
```

specs/fault-proof.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### `l1-transactions <blockhash>`
+
+Requests the host to prepare the list of transactions of the L1 block with `<blockhash>`:
+prepare the RLP pre-images of each of them, including transactions-list MPT nodes.
+
+#### `l1-receipts <blockhash>`
+
+Requests the host to prepare the list of receipts of the L1 block with `<blockhash>`:
+prepare the RLP pre-images of each of them, including transactions-list MPT nodes.
```
ajsutton, line 352:
```suggestion
prepare the RLP pre-images of each of them, including receipts-list MPT nodes.
```

specs/fault-proof.md
```diff
@@ -0,0 +1,412 @@
...
+
+## Fault Proof Interactive Dispute Game
+
+[Interactive Dispute Game]: #Fault-Proof-Interactive-Dispute-Game
+
+The interactive dispute game allows actors to resolve a dispute with an onchain challenge-response game
+that bisects the execution trace of the VM, bounded with a base-case that proves a VM trace step.
+
+The game is multi-player: different non-aligned actors may participate when bonded,
+
```
ajsutton, line 405:
```suggestion
The game is multi-player: different non-aligned actors may participate when bonded.

```

ajsutton (commented):
Hangs together well overall. Mostly just nits.

specs/fault-proof.md
```diff
@@ -0,0 +1,412 @@
...
+<hint bytes> := byte sequence
+<end> := 0 byte
+```
+
+The `<end>` trailing zero byte allows the server to block the program
+(since the communication is blocking) until the hint is processed.
+
+### Pre-image communication
+
+Pre-images are communicated with a minimal wire-protocol over a blocking two-way stream:
```
trianglesphere, line 195: Why "wire-protocol over a blocking two-way stream"? A blocking syscall seems to describe this better. More specifically this seems to pull from the unix syscall lineage with the following API

```
// Header
int open(char *prefixed_pre_image_hash)
int read(int fd, char *buf)


// Usage
fd = open(prefixed pre-image hash)
buf = malloc(...)
while(true) {
    c = read(fd, &buf);
    if (c == 0) break;
}

```

specs/fault-proof.md
protolambda: Mixed some things up here, will fix.

specs/fault-proof.md
protolambda: The VM may read the pre-image up to 32 bytes at a time, but may read less if it chooses to. The pre-image store would just always serve the full available 32 bytes at that offset, and it's up to the VM to mask out the part it didn't want to read. This simplifies the pre-image store contract to just "32 bytes at any offset in the pre-image", with a special case for padding at the end if the pre-image isn't a multiple of 32 bytes.

specs/fault-proof.md
```diff
@@ -0,0 +1,412 @@
...
+When the pre-image oracle serves a request, and the request cannot be served from an existing collection of pre-images
+(e.g. a local pre-image cache) then the VM can execute the hint to retrieve the missing pre-image(s).
+It is the responsiblity of the program to provide sufficient hinting for every pre-image request.
+Some hints may have to be repeated: the VM only has to execute the last hint when handling a missing pre-image.
+
+Note that hints may produce multiple pre-images:
+e.g. a hint for an ethereum block with transaction list may prepare pre-images for the header,
+each of the transactions, and the intermediate merkle-nodes that form the transactions-list Merkle Patricia Trie.
+
+Hinting is implemented with a minimal wire-protocol over a blocking one-way stream:
```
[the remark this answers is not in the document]
protolambda: The neat thing about the hinting concept is that we move the error handling from the program to the VM (which can pass it back to some pre-image fetching tool that comes separately with the program).

This makes it so that the program code-path that runs **onchain is the exact same as the one that runs offchain** during pre-image fetching, avoiding any surprises. The only difference would be the compiler target. But no differences in program logic.

And outside of the program the pre-image fetching can be deferred, cached, retried, etc. all without complicating the program logic or design.

specs/fault-proof.md
protolambda: Yes, it's meant to be implemented with the syscall interface. But I didn't want to enshrine linux syscalls in this spec, although I do think it works very well: we get to avoid magic memory-mapping, and can unify onchain and offchain pre-image oracle code-paths. Since offchain, even outside of a VM emulator, you can just attach a file-descriptor to the program process to serve it the pre-images.
I'll add the blocking syscall as example of this behavior outside of the pre-image oracle thing.

trianglesphere: I mean we should define our own types & condition rather than exactly copying linux here, but I think we should define the write/read/pipe syscalls (though maybe not now) because that will be how the program interacts with the world (and then you need to layer the wire format on top of the syscalls)

specs/fault-proof.md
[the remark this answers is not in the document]
trianglesphere: So does this mean that the offchain program is going to have a separate process & then do the read/writes? I still think that you need to intercept the syscalls somewhere & that might be different (though that is a smaller diff than the pre-image oracle).

My thinking so far is that the error handling is on the other side of an interface (so everything is still in memory + in process) rather than out of process.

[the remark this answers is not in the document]
protolambda: Both will be possible, the separation is optional, we can bind them together the other way for minimal tests as well. In the full production mode however I'd like the code-path during prefetching to be unified with the onchain execution as much as possible, and it's relatively easy to test still (use some pipes, or a sub-process if really unified all the way), and a lot easier than testing what unified execution with memory mapping would look like.

specs/fault-proof.md
protolambda: Definitely, I can add the specific syscall pre-image oracle usage in a separate specs PR, I'll track it in linear. On a related note I need to specify exactly which things the program may use at all; Go itself uses syscalls as well. This crosses over in the FPVM project though, I'll create a linked issue there.

specs/fault-proof.md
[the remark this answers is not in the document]
trianglesphere: Actually, I think if you specify it this way with an ABI you can run the same binary on chain as off chain. That maybe requires some shims off-chain (or implementing a very simple linux ABI on chain), but that's kinda cool.

PR #5417, opened 2023-04-11
[op-batcher] Bug: panics on nil access

mdehoog:
**Description**
Found a couple of potential nil panics. Fix.

**Tests**
No tests added. Tested that the bugs were fixed locally. Was able to reproduce running E2E tests in parallel, when testing #5398.

### DIFF
--- a/op-batcher/batcher/driver.go
+++ b/op-batcher/batcher/driver.go
@@ -191,7 +191,7 @@ func (l *BatchSubmitter) loadBlocksIntoState(ctx context.Context) {
 	if err != nil {
 		l.log.Warn("Error calculating L2 block range", "err", err)
 		return
-	} else if start.Number == end.Number {
+	} else if start.Number >= end.Number {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


op-batcher/batcher/driver.go
```diff
@@ -212,13 +212,15 @@ func (l *BatchSubmitter) loadBlocksIntoState(ctx context.Context) {
 		latestBlock = block
 	}
 
-	l2ref, err := derive.L2BlockToBlockRef(latestBlock, &l.Rollup.Genesis)
-	if err != nil {
-		l.log.Warn("Invalid L2 block loaded into state", "err", err)
-		return
-	}
+	if latestBlock != nil {
```
mdehoog, line 215: this will be nil if `start == end`

tynes: Perhaps good to leave comments in the code in case upstream implementations change, this depends on a particular invariant

mdehoog: @trianglesphere you are totally right... i switched that condition to check for `start >= end` which also fixes the issue

trianglesphere: ah I see that if start > end this would be hit. nice find.

trianglesphere (approved):

PR #5395, opened 2023-04-08
op-program: command to run the host-client separated program

protolambda:
**Description**

Based on #5394 

Implements the main command with hint-based pre-image fetching and KV store. Work in progress. (we need the L1 fetching and l1 oracle based eth client be updated for this to work)

**Tests**

Work in progress.

PR #5394, opened 2023-04-08
op-program: key-value store for pre-images

protolambda:
**Description**

Based on #5393 (for stack consistency with later PRs)

Implements basic key-value stores for pre-image data. Integration follows in separate PR.

**Tests**

KV-stores are tested with basic put/get interactions, the KV interface is pretty simple.

**Metadata**

Fix CLI-3747

PR #5418, opened 2023-04-11
op-program: kv store for pre-images

protolambda:
**Description**

Split out of PR stack, copy of #5394 
See #5395 for example usage (not polished yet), or #5412 for its place in the program: this is holding on to the pre-images (after they are fetched based on hints, or just already there if in read-only mode)

Functionality is pretty basic, we can  consider adding e.g. a LevelDB backed one if we end up having a lot of pre-images.

**Tests**

Added tests for the KV behavior, and those tests can run with any KV implementation.


op-program/host/kvstore/disk.go
```diff
@@ -0,0 +1,64 @@
...
+	return &DiskKV{path: path}
+}
+
+func (d *DiskKV) pathKey(k common.Hash) string {
+	return path.Join(d.path, k.String()+".txt")
+}
+
+func (d *DiskKV) Put(k common.Hash, v []byte) error {
+	// no O_EXCL, the pre-image may already exist. It's fine to overwrite it.
+	f, err := os.OpenFile(d.pathKey(k), os.O_WRONLY|os.O_CREATE|os.O_TRUNC, diskPermission)
```
ajsutton, line 34: Won't this truncate the file on opening and create a potential race condition where the file exists but has no content (where previously it may have existed with content)?

We'd need to be doing atomic writes by writing to a separate file then an atomic move into place (assuming the filesystem supports atomic moves) or add a lock like in the memory kv store.

op-program/host/kvstore/kv.go
```diff
@@ -0,0 +1,16 @@
+package kvstore
+
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// NotFoundErr is returned when a pre-image cannot be found in the KV store.
+var NotFoundErr = errors.New("not found")
```
ajsutton, line 10: Isn't this usually `ErrNotFound`?

ajsutton (commented):
Looks good, just a question about the concurrency of the disk store.

op-program/host/kvstore/disk.go
```diff
@@ -0,0 +1,64 @@
...
+func NewDiskKV(path string) *DiskKV {
+	return &DiskKV{path: path}
+}
+
+func (d *DiskKV) pathKey(k common.Hash) string {
+	return path.Join(d.path, k.String()+".txt")
+}
+
+func (d *DiskKV) Put(k common.Hash, v []byte) error {
+	// no O_EXCL, the pre-image may already exist. It's fine to overwrite it.
```
Inphi, line 33:
```suggestion
	// no O_EXCL, the key file may already exist. It's fine to overwrite it.
```

since in this case it's the hash that already exists and we're overwriting its pre-image with a different value.

Inphi (commented):
the @ajsutton O_TRUNC is indeed an issue. Readers may have a truncated view of the file. If you'd rather not use locks, you can write to write temp, swap and flush.

op-program/host/kvstore/disk.go
trianglesphere: Easy solution is to use a RWMutex - I'd be the performance would be acceptable.

protolambda: Since we're not supposed to overwrite pre-images with different values anyway, I think I'll add `os_O_EXCL` instead. I'll add an already-exists error to the Put interface.

op-program/host/kvstore/kv.go
protolambda: You're right, the standard library prefixes errors like that. With custom error types they're often suffixed like `MyCustomError`, don't think there's a reason to do it one way or the other, but will change it to match stdlib exported-errors naming.

op-program/host/kvstore/disk.go
protolambda: see other comment, changed the behavior here, but thanks for the change suggestion

protolambda:
Updated it it to return `ErrAlreadyExists` when the pre-image already exists in the KV-store. It's up to the user to not write different pre-images with the same key. Shouldn't be an issue for the global pre-image type which has always the same key for the same value based on hash. For inputs that pass through the pre-image oracle we'll be able to catch any error, and shouldn't use the KV store there anyway.

op-program/host/kvstore/disk.go
trianglesphere: Even with O_EXCL I think you can race between reading & writing.

op-program/host/kvstore/disk.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (d *DiskKV) pathKey(k common.Hash) string {
+	return path.Join(d.path, k.String()+".txt")
+}
+
+func (d *DiskKV) Put(k common.Hash, v []byte) error {
+	f, err := os.OpenFile(d.pathKey(k), os.O_WRONLY|os.O_CREATE|os.O_EXCL|os.O_TRUNC, diskPermission)
+	if err != nil {
+		if errors.Is(err, os.ErrExist) {
```
ajsutton, line 36: Even with this it's possible for concurrent calls to `Put` and `Get` to overlap such that the `Get` will return that the file exists but is empty (or get partial content).

To confirm I shoved a `time.Sleep(15 * time.Second)` before the `f.Write` call just below and ran:
```
func TestName(t *testing.T) {
	tmp := t.TempDir() // automatically removed by testing cleanup
	kv := NewDiskKV(tmp)

	hash := common.Hash{0xaa}
	data := []byte{1, 2, 3, 4, 5}
	go func() {
		err := kv.Put(hash, data)
		require.NoError(t, err)
	}()
	time.Sleep(3 * time.Second)
	res, err := kv.Get(hash)
	require.NoError(t, err)
	require.Equal(t, data, res)
}
```
the Get call winds up returning no error and empty data.

op-program/host/kvstore/disk.go
protolambda: Added a rw lock and updated docs to cover the concurrent safety issue if multiple different instances with the same disk directory are used.

op-program/host/kvstore/disk.go
protolambda: Added a rw lock and updated docs to cover usage safety.

op-program/host/kvstore/kv_test.go
```diff
@@ -0,0 +1,67 @@
...
+	})
+
+	t.Run("concurrent", func(t *testing.T) {
+		hash := common.Hash{0xee}
+		data := []byte{1, 2, 3, 4, 5}
+		go func() {
+			err := kv.Put(hash, data)
+			require.NoError(t, err)
+		}()
+		time.Sleep(3 * time.Second)
```
sebastianst, line 62: What's the purpose of this test? And why does it have to have such a high sleep time? Wouldn't something like 100 ms be sufficient? Or even something like
```suggestion
        done := make(chan(struct{}))
		go func() {
			err := kv.Put(hash, data)
			require.NoError(t, err)
			close(done)
		}()
		<-done
```
to avoid sleeping in the test? Although I don't see what the test gives us.

In general I'd try to avoid long sleeps in tests if possible.

If the purpose is to test concurrent access, I'd spawn like 100 goroutines and run the roundtrip test in each, so that concurrent read&write access is tested.

protolambda: Hmm, the proper test of the get-put is only really reproducible by starting the put, then doing the get, and then ending the put. But that can't work with rwlock. And by closing the chan after Put, the Get is already synchronized to happen after put completes. To test we'd have to mock a slow file write, which requires more complexity (have to mock a read-write filesystem (go has read-only FS abstraction, but nothing nice for writes), and it wouldn't be a generic KV test anymore either. I'm dropping the test.

sebastianst (approved):

PR #5420, opened 2023-04-11
op-node: add support for a dynamic sequencer

0xfuturistic:
### Description

This pull request introduces changes enabling the op-node to work with a dynamic system sequencer, improving network fault tolerance and flexibility. This ensures continuous transaction processing during sequencer changes. The update also enables a rotating sequencer system to be implemented by merely changing the block signer's address in the SystemConfig smart contract, facilitating a straightforward sequencer decentralization mechanism. 

Note that a node dynamically reacting to changes of the system's sequencer is optional and enabled or disabled using a new flag of op-node, `sequencer.dynamic.` Currently this behavior is disabled by default.

### Key changes

1. **Update runtime configuration at new L1 heads** 
> The driver now updates the node's runtime configuration at every new L1 head processed, allowing to node to have access to the latest sequencer address of the system. This information is then used to determine which gossip to ignore and, if the node has dynamic sequencing enabled, whether sequencing should be stopped if the node is no longer the system sequencer (and vice-versa). Note that updating the runtime config after the initial loading (specifically at updates for every subsequent L1 block) was mentioned in comments in the code but was missing implementation.
2. **Expose P2P signer address** 
> The node can now access its P2P signer's address during runtime, allowing it to determine if it is the system's sequencer.
3. **Introduce dynamic sequencing** 
> If it has dynamic sequencing enabled, a node can automatically start sequencing if it is the system sequencer and stop if it isn't. This new mechanism ensures that sequencing is stopped if the system's sequencer does not match the node's p2p signer and ensures that sequencing is started if they match (in other words, if the node is the sequencer)
4. **Minor bug fixes:** 
> Devnet contracts deployment command includes a reset flag to avoid a potential error in deploy step 19, and op-e2e's system config was amended to include missing pointers to rollupNodes configs. Line breaks were also introduced in specs/deposits.md to fix lint errors.

### Tests

A new test evaluating how nodes react to changes of the system sequencer was introduced. This test verifies that a node selected as the sequencer gracefully stops when the system sequencer changes and that the node picked by the system automatically starts sequencing.

No further tests are needed since the existing tests of the op-node API cover the start/stop sequencer logic.

Enjoy

maxresdefault

### DIFF
--- a/bedrock-devnet/devnet/__init__.py
+++ b/bedrock-devnet/devnet/__init__.py
@@ -61,7 +61,7 @@ def main():
         addresses = read_json(addresses_json_path)
     else:
         log.info('Deploying contracts.')
-        run_command(['yarn', 'hardhat', '--network', 'devnetL1', 'deploy', '--tags', 'l1'], env={
+        run_command(['yarn', 'hardhat', '--network', 'devnetL1', 'deploy', '--tags', 'l1', '--reset'], env={
             'CHAIN_ID': '900',
             'L1_RPC': 'http://localhost:8545',
             'PRIVATE_KEY_DEPLOYER': 'ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80'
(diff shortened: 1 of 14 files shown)


tynes:
Thanks for the contribution but this will not be merged. For a feature this large, the specs must be proposed first and have rough consensus on the approach. Its not clear what dynamic sequencing means and what problems it solves

0xfuturistic:
Thanks for the feedback @tynes. Will propose specs!

PR #5423, opened 2023-04-12
op-program: Implement caching for the oracle.

ajsutton:
**Description**

Adds a wrapper implementation of the L1 and L2 oracles that adds caching of retrieved objects.  Using the decorator pattern for caching allows reuse of the same caching code regardless of which actual oracle implementation is required and follows the single responsibility principle to make it easier to reason about and test the code.

**Tests**

Added unit tests for the new cache wrappers.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3749/caching


ajsutton:
Currently draft because there's some clean up left to do to extract the stub L2 oracle into a separate file since it's now being reused. That will conflict badly with https://github.com/ethereum-optimism/optimism/pull/5422 though so will wait until it merges.

op-program/client/l1/cache.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"github.com/ethereum-optimism/optimism/op-node/eth"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+// CachingOracle is an implementation of Oracle that delegates to another implementation, adding caching of all results
+type CachingOracle struct {
+	oracle Oracle
+	blocks map[common.Hash]eth.BlockInfo
```
trianglesphere, line 12: Should we use https://pkg.go.dev/github.com/hashicorp/golang-lru/v2 instead of a map here? This will ensure that we don't let the cache grow without bound. We use a mix of v1 and v2 throughout the code base where we want caches (only difference is generic vs casting interfaces)

ajsutton: Yeah I was wondering about that. It's a somewhat unique situation because the VM disables GC anyway but may be worth benefiting from the reduced memory when running it off-chain. Will switch it over and see what @protolambda thinks - easy to revert if we want to keep it simple and cache everything.

ajsutton: Pushed, but I really have no idea what reasonable sizing is for the caches. Will likely need to get some numbers from a busy network at some point so if we do go with an LRU cache, I'll add a separate ticket to gather data and set sensible cache sizes.

protolambda (approved):
LGTM. The cache size can be tweaked later. Avoiding pre-image oracle calls in favor of in-memory data is good, since it saves pre-image proof overhead and additional instructions (assuming we have duplicate reads to hit the cache). The GC being disabled means that reusing old heap memory is indeed troublesome. So a `map` may be better, but we can try that later once we have a VM ready to verify. Just having the caching layer at all is already an improvement, and changing this to `map` later if necessary is easy.

PR #5424, opened 2023-04-12
op-node: finalize while syncing

protolambda:
**Description**

Finalize while syncing. The last finalized L1 signal is now always retained in the engine queue, even if it does not match recently processed L1 blocks (the `finalityData` buffer). When the finalized L1 block is past the current traversed L1 origin, the engine queue will now try to verify it's on the correct chain, and finalize it. It only does so every 64 L1 blocks, based on the last attempt, to avoid frequent repeats and duplicate work.

**Tests**

Added an action test to create a finalized L1 chain, and then make the verifier sync it, and verify it finalizes even after getting a signal for data it didn't traverse to yet at the time of the signal.

**Metadata**

Fix CLI-3794

**TODOs**


trianglesphere (approved):
lgtm

PR #5425, opened 2023-04-12
fix(specs/withdrawals): Sherlock updates

qbzzt:
Closing: DEVRL-844

### DIFF
--- a/specs/withdrawals.md
+++ b/specs/withdrawals.md
@@ -18,8 +18,8 @@ an L2 account to an L1 account.
 more specific terms to differentiate:
 
 - A _withdrawal initiating transaction_ refers specifically to a transaction on L2 sent to the Withdrawals predeploy.
-- A _withdrawal proofing transaction_ refers specifically to an L1 transaction 
-  which proves the withdrawal is correct (that it has been included in a merkle 
+- A _withdrawal proofing transaction_ refers specifically to an L1 transaction
+  which proves the withdrawal is correct (that it has been included in a merkle
   tree whose root is available on L1).
 - A _withdrawal finalizing transaction_ refers specifically to an L1 transaction which finalizes and relays the
   withdrawal.
@@ -27,7 +27,7 @@ more specific terms to differentiate:
 Withdrawals are initiated on L2 via a call to the Message Passer predeploy contract, which records the important
 properties of the message in its storage.
 Withdrawals are proven on L1 via a call to the `OptimismPortal`, which proves the inclusion of this withdrawal message.
-Withdrawals are finalized on L1 via a call to the `OptimismPortal` contract, 
+Withdrawals are finalized on L1 via a call to the `OptimismPortal` contract,
 which verifies that the fault challenge period has passed since the withdrawal message has been proved.
 
 In this way, withdrawals are different from [deposits][g-deposits] which make use of a special transaction type in the
@@ -75,7 +75,7 @@ This is a very simple contract that stores the hash of the withdrawal data.
 1. After the withdrawal is proven, it enters a 7 day challenge period, allowing time for other network participants
    to challenge the integrity of the corresponding output root.
 1. Once the challenge period has passed, a relayer submits a withdrawal finalizing transaction once again to the
-   `OptimismPortal` contract. 
+   `OptimismPortal` contract.
    Again, the relayer is not necessarily the same entity which initiated the withdrawal on L2.
 1. The `OptimismPortal` contract receives the withdrawal transaction data and verifies that the withdrawal has
    both been proven and passed the challenge period.
--- a/specs/withdrawals.md
+++ b/specs/withdrawals.md
@@ -18,14 +18,17 @@ an L2 account to an L1 account.
 more specific terms to differentiate:
 
 - A _withdrawal initiating transaction_ refers specifically to a transaction on L2 sent to the Withdrawals predeploy.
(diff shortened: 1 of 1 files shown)


specs/withdrawals.md
```diff
@@ -59,18 +64,19 @@ This is a very simple contract that stores the hash of the withdrawal data.
...
 1. The `OptimismPortal` contract retrieves the output root for the given block number from the `L2OutputOracle`'s
-   `getL2OutputAfter()` function, and performs the remainder of the verification process internally.
+   `getL2Output()` function, and performs the remainder of the verification process internally.
 1. If proof verification fails, the call reverts. Otherwise the hash is recorded to prevent it from being re-proven.
    Note that the withdrawal can be proven more than once if the corresponding output root changes.
 1. After the withdrawal is proven, it enters a 7 day challenge period, allowing time for other network participants
    to challenge the integrity of the corresponding output root.
-1. Once the challenge period has passed, a relayer submits the withdrawal transaction once again to the
-   `OptimismPortal` contract. Again, the relayer need not be the same entity which initiated the withdrawal on L2.
+1. Once the challenge period has passed, a relayer submits a withdrawal finalizing transaction once again to the
```
ajsutton, line 77: Suspect we should remove "once again" from here.  This is the first time the withdrawal finalising transaction has been sent - it's the withdrawal proving transaction that's sent previously.

specs/withdrawals.md
```diff
@@ -59,18 +64,19 @@ This is a very simple contract that stores the hash of the withdrawal data.
 
 ### On L1
 
-1. A [relayer][g-relayer] submits the required inputs to the `OptimismPortal` contract. The relayer need
-   not be the same entity which initiated the withdrawal on L2.
+1. A [relayer][g-relayer] submits a withdrawal proving transaction required inputs to the `OptimismPortal` contract.
```
ajsutton, line 67:
```suggestion
1. A [relayer][g-relayer] submits a withdrawal proving transaction with the required inputs to the `OptimismPortal` contract.
```

specs/withdrawals.md
```diff
@@ -18,12 +18,17 @@ an L2 account to an L1 account.
 more specific terms to differentiate:
 
 - A _withdrawal initiating transaction_ refers specifically to a transaction on L2 sent to the Withdrawals predeploy.
+- A _withdrawal proofing transaction_ refers specifically to an L1 transaction
```
ajsutton, line 21:
```suggestion
- A _withdrawal proving transaction_ refers specifically to an L1 transaction
```
given we're using `proving` below.

tynes: Also the word `proofing` means something different than `proving`

specs/withdrawals.md
```diff
@@ -59,18 +64,19 @@ This is a very simple contract that stores the hash of the withdrawal data.
...
+   `getL2Output()` function, and performs the remainder of the verification process internally.
 1. If proof verification fails, the call reverts. Otherwise the hash is recorded to prevent it from being re-proven.
    Note that the withdrawal can be proven more than once if the corresponding output root changes.
 1. After the withdrawal is proven, it enters a 7 day challenge period, allowing time for other network participants
    to challenge the integrity of the corresponding output root.
-1. Once the challenge period has passed, a relayer submits the withdrawal transaction once again to the
-   `OptimismPortal` contract. Again, the relayer need not be the same entity which initiated the withdrawal on L2.
+1. Once the challenge period has passed, a relayer submits a withdrawal finalizing transaction once again to the
+   `OptimismPortal` contract.
+   Again, the relayer is not necessarily the same entity which initiated the withdrawal on L2.
```
tynes, line 79: We can probably just say the thing about the relayer once and word it along the line of `the relayer doesn't need to be the same entity that initiated the withdrawal` because in practice it usually is. There are no incentives to spend gas on other people's withdrawals

specs/withdrawals.md
qbzzt: I agree, thank you.

specs/withdrawals.md
qbzzt: With account abstraction coming up, this is sufficiently important that repetition will probably save us some support questions. I rephrased as you suggested.

specs/withdrawals.md
```diff
@@ -175,7 +182,7 @@ The following inputs are required to prove and finalize a withdrawal:
 These inputs must satisfy the following conditions:
 
 1. The `l2BlockNumber` must be the block number that corresponds to the `OutputProposal` being proven.
-1. `L2OutputOracle.getL2OutputAfter(l2BlockNumber)` returns a non-zero `OutputProposal`.
+1. `L2OutputOracle.getL2Output(l2BlockNumber)` returns a non-zero `OutputProposal`.
```
refcell, line 185: The `getL2Output` function takes the `l2OutputIndex` not `l2BlockNumber`.

See [L2OutputOracle.sol#L239-L245](https://github.com/ethereum-optimism/optimism/blob/qbzzt/230410-specs-withdrawals/packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol#L239-L245):
```solidity
function getL2Output(uint256 _l2OutputIndex)
    external
    view
    returns (Types.OutputProposal memory)
{
    return l2Outputs[_l2OutputIndex];
}
```

refcell (changes requested):
Small change on the input to the L2OutputOracle's `getL2Output` function. Otherwise, looks great!

specs/withdrawals.md
qbzzt: Thank you, fixed (it also required a fix a few lines higher)

ajsutton (approved):

PR #5426, opened 2023-04-12
[op-node] Add support for non-batched RPC calls when batchSize == 1

mdehoog:
**Description**

Adds support to op-node's batching implementation that uses unbatched calls when batchSize == 1, to support nodes that don't support JSON-RPC batching.

An alternative implementation would be to add a new node-provider-specific receipt fetcher that avoids RPC batching. This current PR seems more compatible with other future uses of the batching implementation.

**Tests**

Added new tests.

**Additional context**

Some node providers don't support batched JSON-RPC calls, for example AWS's [managed node offering](https://docs.aws.amazon.com/managed-blockchain/latest/ethereum-dev/ethereum-nodes.html). This PR allows running op-stack nodes using AWS as the L1 node.

**TODOs**

### DIFF
--- a/op-node/sources/batching.go
+++ b/op-node/sources/batching.go
@@ -24,6 +24,7 @@ type IterativeBatchCall[K any, V any] struct {
 
 	makeRequest func(K) (V, rpc.BatchElem)
 	getBatch    BatchCallContextFn
+	getSingle   CallContextFn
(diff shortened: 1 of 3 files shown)


protolambda (approved):
LGTM

PR #5427, opened 2023-04-12
op-e2e: Longer seq window for TestSystemMockP2P

trianglesphere:

PR #5428, opened 2023-04-12
ci: Reduce executor size

mslipper:
Our CCI usage more than doubled over the last few months. After looking at the config, we were using xlarge VMs for Docker build/publish steps as well as Slither which cost 40 credits/minute. I reduced those to values I think are more appopriate, and reduced some of our Docker instance sizes as well.


protolambda (approved):

trianglesphere:
op-e2e test is a known flake which should be fixed in #5427

zhwrd (approved):

PR #5429, opened 2023-04-12
op-e2e: fix TestLargeL1Gaps flake

protolambda:
Fix a flaky test by manually incrementing the nonce, instead of relying on the tx pool which may or may not present the right nonce value.


protolambda:
Flaked on the test that the other PR flaked on, that flaked on the test that was fixed here. We'll have to rerun one or the other, or just merge the flake fixes.

trianglesphere (approved):

protolambda:
Will rebase this to get the other flake fix in, to see if CI will pass

PR #5430, opened 2023-04-12
Revert "[op-node] Add support for non-batched RPC calls when batchSize == 1"

protolambda:
Reverts ethereum-optimism/optimism#5426

The change caused CI to fail here: https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/16566/workflows/c3c21446-c755-45e1-9ce9-f2914e3bcc7d/jobs/612077/steps

cc @mdehoog 

We'll look into the CI failure, but reverting for now.

PR #5434, opened 2023-04-13
[op-node] Add support for non-batched RPC calls when batchSize == 1 (flaky test fixed)

mdehoog:
Redo of #5426 (reverted in #5430), with a flaky test fix. The test was flaky because there was a chance that the mocked method could be called early if the `select` didn't return due to the context being closed. The fix is to introduce an early check of the `ctx.Err()` to the `Fetch` method: https://github.com/ethereum-optimism/optimism/blob/916513af6e4d708706bcaf0d2d1ba17b03fe9910/op-node/sources/batching.go#L90-L93

Copied previous description below:

**Description**

Adds support to op-node's batching implementation that uses unbatched calls when batchSize == 1, to support nodes that don't support JSON-RPC batching.

An alternative implementation would be to add a new node-provider-specific receipt fetcher that avoids RPC batching. This current PR seems more compatible with other future uses of the batching implementation.

**Tests**

Added new tests.

**Additional context**

Some node providers don't support batched JSON-RPC calls, for example AWS's [managed node offering](https://docs.aws.amazon.com/managed-blockchain/latest/ethereum-dev/ethereum-nodes.html). This PR allows running op-stack nodes using AWS as the L1 node.

**TODOs**

### DIFF
--- a/op-node/sources/batching.go
+++ b/op-node/sources/batching.go
@@ -24,6 +24,7 @@ type IterativeBatchCall[K any, V any] struct {
 
 	makeRequest func(K) (V, rpc.BatchElem)
 	getBatch    BatchCallContextFn
+	getSingle   CallContextFn
 
 	requestsValues []V
 	scheduled      chan rpc.BatchElem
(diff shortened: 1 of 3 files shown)


protolambda:
Re-running the Go tests for green CI. Something small and unrelated flaked, fixed here: https://github.com/ethereum-optimism/optimism/pull/5436

protolambda (approved):

PR #5437, opened 2023-04-13
contracts-bedrock: update systemconfig resource config validation

tynes:
**Description**

Updates the way that the system config validates the resource config to ensure that its not possible to set config values that break the system. It is possible to set config values that would burn too much gas, rendering the system unusable. Every error cannot be caught, so try to catch the most possible. Altering the default resource config is not recommended, but it is possible.

**Tests**

This also adds fuzz tests that cover the problem, if the change in the require statement in the system config contract is reverted, then the fuzz tests catch the `UNDEFINED` error quite quickly.


clabby (approved):
LGTM. The fuzz test has a lot of `vm.assume`s, which could cause us to reach the max rejection limit. Would be great if we can convert some of these to `bound(...)` if possible. Leaving this note here in case someone is debugging why ctb goes red.

trianglesphere:
@tynes this broke `ci/circleci: contracts-bedrock-checks`

PR #5438, opened 2023-04-13
Add cookies + headers support to op-node, op-batcher, op-proposer L1 client

mdehoog:
**Description**

Adds support for cookies and custom headers on the L1 clients using [cookiejar](https://pkg.go.dev/net/http/cookiejar) and `http.Header`.

This is useful for certain node deployments that use cookies, for example running L1 nodes behind an AWS load balancer using [sticky sessions](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/sticky-sessions.html).

A use for custom headers is passing `Accept: application/gzip` to enable gzip encoding of responses, reducing network throughput. AWS's managed Ethereum nodes require this header, otherwise they return errors if the response size is too large. Another use is auth via headers.

As an example for sticky sessions, we are occasionally seeing batcher transactions being confirmed, but the next transaction uses a nonce that is too low because some nodes behind the load balancer haven't yet processed the block.

Also refactors the txmgr to share an L1 client with the caller, rather than creating it's own internally.

**Tests**

Add test, and modified existing tests.

**TODOs**


trianglesphere:
> If we feel like this adds too much complexity, we can close this (we can implement this as a reverse proxy to manage cookies on a different layer)

I think we're fine with this extra functionality. I've actually been thinking about migrating the `op-node/client` package to `op-service` and using the client in the batcher+proposer. With these extra `l1` flags, I think it'd make sense to do that move & create shared `l1` flags

sebastianst:
Adding do-not-merge until we've decided how to proceed.

op-node/client/rpc.go
```diff
@@ -186,6 +187,13 @@ func (ic *InstrumentedRPCClient) EthSubscribe(ctx context.Context, channel any,
 	return ic.c.EthSubscribe(ctx, channel, args...)
 }
 
+func instrument1(m *metrics.Metrics, name string, cb func() error) error {
+	record := m.RecordRPCClientRequest(name)
+	err := cb()
+	record(err)
+	return err
+}
+
```
mdehoog, line 196: left-over from `op-node/client/client.go` which dooesn't appear to be used at all

mdehoog:
@sebastianst @trianglesphere I implemented the shared client as mentioned, i think this is ready for another review

I avoided trying to pull out too much of the op-node client implementation as there is a lot of op-node specific code there (started [here](https://github.com/ethereum-optimism/optimism/compare/develop...mdehoog:optimism:michael/move-l1-client) but it felt too complex). Instead just opted to implement shared client initialization and CLI flags.

trianglesphere:
> @sebastianst @trianglesphere I implemented the shared client as mentioned, i think this is ready for another review
> 
> I avoided trying to pull out too much of the op-node client implementation as there is a lot of op-node specific code there (started [here](https://github.com/ethereum-optimism/optimism/compare/develop...mdehoog:optimism:michael/move-l1-client) but it felt too complex). Instead just opted to implement shared client initialization and CLI flags.

Thanks for pulling this together. Taking a look at it I think I lean away from bringing this in now. This is at the in between of consolidating the different client package, but I think we'll want to do it in a single place. It's a large set of changes so probably not an immediate priority.

mdehoog:
Too complex for now, may separate into smaller, more manageable PRs in the future

PR #5439, opened 2023-04-13
op-chain-ops: prevent invalid decoding errors

tynes:
**Description**

Invalid decoding errors should not stop the script from progressing. This fix ensures that all possible errors are handled in a way that they can be skipped. The panic conditions should never happen, so we do not need to worry about them. It is permissionless to insert data into the set that is parsed through this function, so errors cannot break it.

**Tests**

Also adds unit tests for the parsing, ensures to cover each relevant branch.

Closes CLI-3830


op-chain-ops/crossdomain/witness.go
```diff
@@ -41,6 +41,40 @@ func NewSentMessageFromJSON(path string) ([]*SentMessage, error) {
...
+	method, err := abi.MethodById(msg[:4])
+	if err != nil {
+		return nil, err
+	}
+
+	if method.Sig != "passMessageToL1(bytes)" {
+		return nil, fmt.Errorf("unknown method: %s", method.Name)
+	}
+
+	out, err := method.Inputs.Unpack(msg[4:])
```
mslipper, line 66: Confirming: if someone calls `passMessageToL1` such that this `Unpack` fails, the call will revert? The withdrawals migration needs to know about any call to `passMessageToL1` that mutates the `sentMessages` mapping.

mslipper (approved):
LGTM pending my comment above.

op-chain-ops/crossdomain/witness.go
tynes: I've added a test showing the functionality here. The 4 byte selector must match and the following calldata must be abi encoded bytes, otherwise the call will revert.

PR #5441, opened 2023-04-13
op-chain-ops: withdrawal script witness file

tynes:
**Description**

Allow for configuration of the withdrawal script using the witness file. The parsing now happens in go instead of in javascript.


clabby (approved):
LGTM- thank you!

mslipper (approved):

PR #5444, opened 2023-04-13
feat(ctb): Fail relayed XDM messages on reentry

maurelian:
**Description**

This PR is an alternative to #5440.

The solution implementation as well as fuzz tests are included in this PR. Invariant testing will be split out to properly iterate and introduce invariant testing.

The sherlock-identified issue in the `CrossDomainMessager` whereby `relayMessage` naively reset the `xDomainMsgSender` after a call. The issue here was that a nested call would reset this value to the `Constants.DEFAULT_L2_SENDER` which deviates from the expected behavior that the `xDomainMsgSender` is set to the `_sender` for the entirety of the subcall.

As opposed to the minimal solution outlined in #5440, this change sets a message as failed upon re-entry to the `relayMessage` function by checking the value of `xDomainMsgSender`.

This check is a simple addition of the following
```Diff
+ // Another message is being relayed, so we just store this for future replay.
+ if (xDomainMsgSender != Constants.DEFAULT_L2_SENDER) {
+     failedMessages[versionedHash] = true;
+     return;
+ }
```

We then also **remove** the `reentrancyLocks` from storage and don't have to worry about re-entrency here since _any_ subcall through the `SafeCall` lib to the XDM `relayMessage` will be caught and recorded as failed by the aforementioned `xDomainMsgSender` check.

**Benefits of this solution**

- Removes the `reentrancyLocks` saving `SSTORE`s and `SLOAD`s as well as removing the storage mapping altogether.
- Removes more complexity and proverbially "bites the bullet" now to deal with re-entry.

**Potential Solution Cons**

- Deviates from the suggested minimal-solution from multiple sherlock finding reports.
- Changes state and the codepaths from what has been audited.

**Metadata**

Fixes CLI-3834


packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -357,6 +345,12 @@ abstract contract CrossDomainMessenger is
             "CrossDomainMessenger: message has already been relayed"
         );
 
+        // Another message is being relayed, so we just store this for future replay.
+        if (xDomainMsgSender != Constants.DEFAULT_L2_SENDER) {
+            failedMessages[versionedHash] = true;
```
tynes, line 350: This needs to also emit an event so we can hold the invariant that an event is emitted each time `failedMessages` has a value set in it

packages/contracts-bedrock/contracts/test/L1CrossDomainMessenger.t.sol
```diff
@@ -295,173 +291,6 @@ contract L1CrossDomainMessenger_Test is Messenger_Initializer {
         assertEq(L1Messenger.failedMessages(hash), true);
     }
 
-    // relayMessage: Should revert if the recipient is trying to reenter with the
```
tynes, line 298: Its really annoying that we have duplicate tests for both L1 and L2 - we should really be testing the base contract directly, there is nothing to do as part of this PR but just wanted to note this

tynes:
This definitely seems quite sane, just be sure to squash commits when ready for review

tynes:
I'm trying to build the bindings locally to see if i get a diff

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -17,3 +23,100 @@ contract CrossDomainMessenger_BaseGas_Test is Messenger_Initializer {
...
+
+    // Internal helper function to relay a message and perform assertions.
+    function internalRelay(address innerSender) internal {
+        assertEq(sender, L1Messenger.xDomainMessageSender());
+
+        bytes memory callMessage = abi.encodeWithSelector(
+            CrossDomainMessenger_RelayMessage_Test.relayMessage.selector
+        );
+
+        bytes32 hash = Hashing.hashCrossDomainMessage(
```
tynes, line 45: I would love it if you used kwargs style syntax here so we know what the 0s are

refcell: Good call - will use kwargs going forward 👍  (maybe we should also consider putting in a gardening ticket or sub-ticket underneath a styleguide compliance ticket)

packages/contracts-bedrock/.storage-layout
```diff
@@ -135,8 +134,7 @@
 | xDomainMsgSender       | address                  | 204  | 0      | 20    | contracts/L2/L2CrossDomainMessenger.sol:L2CrossDomainMessenger |
 | msgNonce               | uint240                  | 205  | 0      | 30    | contracts/L2/L2CrossDomainMessenger.sol:L2CrossDomainMessenger |
 | failedMessages         | mapping(bytes32 => bool) | 206  | 0      | 32    | contracts/L2/L2CrossDomainMessenger.sol:L2CrossDomainMessenger |
-| reentrancyLocks        | mapping(bytes32 => bool) | 207  | 0      | 32    | contracts/L2/L2CrossDomainMessenger.sol:L2CrossDomainMessenger |
-| __gap                  | uint256[41]              | 208  | 0      | 1312  | contracts/L2/L2CrossDomainMessenger.sol:L2CrossDomainMessenger |
```
maurelian, line 139: I'm pretty sure this is fine, but would appreciate if @tynes can explicitly acknowledge it.

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -17,3 +23,100 @@ contract CrossDomainMessenger_BaseGas_Test is Messenger_Initializer {
...
+        );
+
+        vm.expectCall(target, callMessage);
+
+        bytes32 hash = Hashing.hashCrossDomainMessage({
+            _nonce: Encoding.encodeVersionedNonce({ _nonce: 0, _version: 1 }),
+            _sender: sender,
+            _target: target,
+            _value: 0,
+            _gasLimit: 0,
```
maurelian, line 100: Is it worth fuzzing this value too?

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -17,3 +23,100 @@ contract CrossDomainMessenger_BaseGas_Test is Messenger_Initializer {
...
+    // Storage slot of the l2Sender
+    uint256 constant senderSlotIndex = 50;
+
+    address public fuzzedSender;
+    address public sender;
+
+    address public target;
+
+    // Internal helper function to relay a message and perform assertions.
+    function internalRelay(address innerSender) internal {
```
maurelian, line 38: nit, but `_internalRelay` and `_innerSender`

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -17,3 +23,100 @@ contract CrossDomainMessenger_BaseGas_Test is Messenger_Initializer {
...
+        for (uint256 i = 0; i < 10; i++) {
+            address innerSender;
+            unchecked {
+                innerSender = address(uint160(uint256(uint160(fuzzedSender)) + i));
+            }
+            internalRelay(innerSender);
+        }
+    }
+
+    function testFuzz_relayMessageReenter_succeeds(address _sender) external {
```
maurelian, line 83: What's the purpose of fuzzing the _sender value in particular?

packages/contracts-bedrock/.storage-layout
tynes: Yes I agree that this is fine because the mapping was used transiently. The only way that a value would be left in there is if there is a bug in our code (static analysis using my eyes says no altho i didn't read the bytecode) or if ethereum is broken

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
tynes: I think fuzzing the gaslimit here makes sense as well as the gaslimit in the top level call, just in case there is a weird unnoticed interaction. There is also the possibility of making the `relayMessage` call with a certain amount of gas

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -17,3 +23,100 @@ contract CrossDomainMessenger_BaseGas_Test is Messenger_Initializer {
...
+            _message: callMessage
+        });
+
+        assertTrue(L1Messenger.failedMessages(hash));
+        assertFalse(L1Messenger.successfulMessages(hash));
+        assertEq(sender, L1Messenger.xDomainMessageSender());
+    }
+
+    // relayMessage is called by the L1 Cross Domain Messenger.
+    function relayMessage() external payable {
```
tynes, line 73: this can be internal right?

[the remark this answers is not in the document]
tynes: I'm confused where this gets called

packages/contracts-bedrock/.storage-layout
refcell: Resolving to continue reviews and get this across the line 🦾

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
refcell: Since that's the value that is set to state (`xDomainMsgSender`) I think it makes sense to fuzz this value.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -357,6 +345,13 @@ abstract contract CrossDomainMessenger is
             "CrossDomainMessenger: message has already been relayed"
         );
 
+        // Another message is being relayed, so we just store this for future replay.
```
tynes, line 348: I would prefer to include specific reference to this detecting re-entrency.

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -17,3 +23,100 @@ contract CrossDomainMessenger_BaseGas_Test is Messenger_Initializer {
...
+        L1Messenger.relayMessage({
+            _nonce: Encoding.encodeVersionedNonce({ _nonce: 0, _version: 1 }),
+            _sender: sender,
+            _target: target,
+            _value: 0,
+            _minGasLimit: 0,
+            _message: callMessage
+        });
+
+        assert(L1Messenger.successfulMessages(hash));
```
tynes, line 116: nit: mixing up `assert`, `assertEq` and `assertTrue`/`assertFalse`

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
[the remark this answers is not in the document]
maurelian: Just adding that I also had trouble following the logic. The naming collision doesn't help.:)

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
```diff
@@ -17,3 +23,100 @@ contract CrossDomainMessenger_BaseGas_Test is Messenger_Initializer {
         L1Messenger.baseGas(hex"ff", _minGasLimit);
     }
 }
+
+// CrossDomainMessenger_RelayMessage_Test tests re-entrency of relayMessage.
+contract CrossDomainMessenger_RelayMessage_Test is Messenger_Initializer {
+    // Storage slot of the l2Sender
+    uint256 constant senderSlotIndex = 50;
```
tynes, line 30: Just had a though, i wonder if it makes sense to have a lib in `Common.t.sol` that includes constants for storage slots we want to set dynamically so in case they change we can update a single place

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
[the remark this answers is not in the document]
refcell: Yep noted and fixed!

packages/contracts-bedrock/contracts/test/CrossDomainMessenger.t.sol
refcell: Maybe it makes sense to pull slots + constants out further into their own contract? Something like `State.t.sol`?

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
```diff
@@ -1019,7 +1019,8 @@ contract OptimismPortal_FinalizeWithdrawal_Test is Portal_Initializer {
         assertEq(outputRoot, Hashing.hashOutputRootProof(proof));
         assertEq(withdrawalHash, Hashing.hashWithdrawal(_tx));
 
-        // Mock the call to the oracle
+        // Setup the Oracle to return the outputRoot we want as well as a finalized timestamp.
+        uint256 finalizedTimestamp = block.timestamp - oracle.FINALIZATION_PERIOD_SECONDS() - 1;
```
maurelian, line 1023: I don't think this is used anywhere.

maurelian (commented):
I'm unable to approve since I opened the PR, but LGTM other than one nit in a test file.

packages/contracts-bedrock/contracts/test/OptimismPortal.t.sol
refcell: Good catch!

tynes (approved):

PR #5446, opened 2023-04-13
op-program: trie list pre-image read/write util funcs

protolambda:
**Description**

Add `mpt.ReadTrie` and `mpt.WriteTrie` util funcs to load/dump Merkle Patricia Trie lists (as used in tx, receipt, and withdrawal lists in ethereum) from/to a pre-image collection.

TBD:
- These functions use panics currently, because we'd exit the program with critical error anyway if the MPT data is invalid. Do we intend to use these functions elsewhere? We may otherwise want to add the error returns still.
- The `trie.Reader` and `trie.NodeReader` functions in geth seem promising, but use some private types, do we want to change this in op-geth, to get rid of the DB shim workaround (quite hacky TBH)?

**Tests**

- Round trip some random collections of list elements.
- Empty list case
- Basic case for easy debugging

**Metadata**

Fix CLI-3827

**TODOs**

### DIFF
--- a/op-program/client/mpt/trie.go
+++ b/op-program/client/mpt/trie.go
@@ -1,0 +1,120 @@
+package mpt
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/trie"
+)
+
+// ReadTrie takes a Merkle Patricia Trie (MPT) root of a "DerivableList", and a pre-image oracle getter,
+// and traverses the implied MPT to collect all raw leaf nodes in order, which are then returned.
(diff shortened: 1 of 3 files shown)


op-program/client/mpt/trie.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+		Put: func(key []byte, value []byte) {
+			panic("put not supported")
+		},
+		Delete: func(key []byte) {
+			panic("delete not supported")
+		},
+	}}
+
+	// tire.New backed with a trie.NodeReader and trie.Reader seems really promising
```
ajsutton, line 31:
```suggestion
	// trie.New backed with a trie.NodeReader and trie.Reader seems really promising
```

ajsutton (approved):
LGTM. If we need to implement the Database interface for writing anyway, I wouldn't bother trying to simplify the read side. It works well and tests will tell us quickly if an op-geth upgrade ever requires implementing new Database methods.

protolambda:
Dismissing review, will make some more changes, seeing some things that next PR will not work as well with.

protolambda:
Ready for review now, sorry :sweat_smile:

ajsutton (approved):

PR #5449, opened 2023-04-13
op-program: pre-image oracle implementation

protolambda:
**Description**

L1 and L2 pre-image oracle implementations. These implement the `Oracle` interface by mapping the requests to pre-image gets & pre-image hints (all behind interface). The L1/L2 pre-image oracles do the encoding/decoding, and call the MPT utils to load the transactions and receipts lists by tx / receipts roots.

It uses the pre-image key scheme (the prefix byte approach) from the specs in #5412 

This PR depends on #5446 

**Tests**

It create random blocks, receipts, state nodes, contract code and pulls them through the pre-image oracle, incl. the MPT functionality.

**Metadata**

Fix CLI-3824

### DIFF
--- a/op-node/eth/receipts.go
+++ b/op-node/eth/receipts.go
@@ -1,0 +1,54 @@
+package eth
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+// EncodeReceipts encodes a list of receipts into raw receipts. Some non-consensus meta-data may be lost.
+func EncodeReceipts(elems []*types.Receipt) ([]hexutil.Bytes, error) {
+	out := make([]hexutil.Bytes, len(elems))
+	for i, el := range elems {
+		dat, err := el.MarshalBinary()
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal receipt %d: %w", i, err)
+		}
+		out[i] = dat
+	}
+	return out, nil
+}
+
+// DecodeRawReceipts decodes receipts and adds additional blocks metadata.
+// The contract-deployment addresses are not set however (high cost, depends on nonce values, unused by op-node).
+func DecodeRawReceipts(block BlockID, rawReceipts []hexutil.Bytes, txHashes []common.Hash) ([]*types.Receipt, error) {
+	result := make([]*types.Receipt, len(rawReceipts))
+	totalIndex := uint(0)
+	prevCumulativeGasUsed := uint64(0)
+	for i, r := range rawReceipts {
+		var x types.Receipt
+		if err := x.UnmarshalBinary(r); err != nil {
+			return nil, fmt.Errorf("failed to decode receipt %d: %w", i, err)
+		}
+		x.TxHash = txHashes[i]
+		x.BlockHash = block.Hash
+		x.BlockNumber = new(big.Int).SetUint64(block.Number)
+		x.TransactionIndex = uint(i)
+		x.GasUsed = x.CumulativeGasUsed - prevCumulativeGasUsed
+		// contract address meta-data is not computed.
+		prevCumulativeGasUsed = x.CumulativeGasUsed
+		for _, l := range x.Logs {
+			l.BlockNumber = block.Number
+			l.TxHash = x.TxHash
+			l.TxIndex = uint(i)
+			l.BlockHash = block.Hash
+			l.Index = totalIndex
+			totalIndex += 1
+		}
+		result[i] = &x
+	}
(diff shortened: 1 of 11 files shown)


op-program/preimage/iface.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// KeyType is the key-type of a pre-image, used to prefix the pre-image key with.
+type KeyType byte
+
+const (
+	// The zero key type is illegal to use, ensuring all keys are non-zero.
+	_ KeyType = 0
+	// LocalKeyType is for input-type pre-images, specific to the local program instance.
+	LocalKeyType KeyType = 0
```
ajsutton, line 35: Shouldn't this be:
```suggestion
	LocalKeyType KeyType = 1
```

ajsutton (approved):
LGTM. Just the one key prefix that I think was wrong.

op-node/eth/receipts.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal receipt %d: %w", i, err)
+		}
+		out[i] = dat
+	}
+	return out, nil
+}
+
+// DecodeRawReceipts decodes receipts and adds additional blocks metadata.
+// The contract-deployment addresses are not set however (high cost, depends on nonce values, unused by op-node).
```
trianglesphere, line 26: I know we don't use everything in the receipt, but I wonder if we should migrate to a different type (like `ConsensusReceipt`) and use that inside the op-node b/c it worries be that there is a difference here.

op-node/eth/transactions.go
```diff
@@ -0,0 +1,43 @@
...
+		var x types.Transaction
+		if err := x.UnmarshalBinary(data[i]); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal tx %d: %w", i, err)
+		}
+		dest[i] = &x
+	}
+	return dest, nil
+}
+
+func HashTransactions(elems []*types.Transaction) []common.Hash {
```
trianglesphere, line 37: Weird function name - I thought it was gonna do a flat hash over the transactions.. Maybe `SelectTxHashes` or `TransactionsToHashes`

op-program/preimage/iface.go
```diff
@@ -0,0 +1,77 @@
...
+type Key interface {
+	// PreimageKey changes the Key commitment into a
+	// 32-byte type-prefixed preimage key.
+	PreimageKey() common.Hash
+}
+
+type Oracle interface {
+	// Get the full pre-image of a given pre-image key.
+	// This returns no error: the client state-transition
+	// is invalid if there is any missing pre-image data.
```
trianglesphere, line 18: Where does the failure occur? Or is it an error to return an empty array?

op-program/client/l2/oracle_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/ethereum-optimism/optimism/op-node/testutils"
+	"github.com/ethereum-optimism/optimism/op-program/client/mpt"
+	"github.com/ethereum-optimism/optimism/op-program/preimage"
+)
+
+func mockPreimageOracle(t *testing.T) (po *PreimageOracle, hintsMock *mock.Mock, preimages map[common.Hash][]byte) {
+	// Prepare the pre-images
+	preimages = make(map[common.Hash][]byte)
+
+	hintsMock = new(mock.Mock)
```
trianglesphere, line 26: IMO the number of hints called for isn't part of the correctness conditions & this would be a really easy function to fake rather than mock.

ajsutton: I think it is important in this case to be specific about what hints are requested (ideally we'd also assert that the hint must precede requesting the preimages it needs since the host fetching code will depend on that).

op-node/eth/receipts.go
ajsutton: That's a good idea for a larger refactor. Particularly if we can just use the consensus data and not have to fill in extra data from the transactions etc.

op-program/client/l2/oracle_test.go
protolambda: Wanted to be very explicit about which hints were being expected, it's not really complicated functionality, just need to make sure something is called so we know the pre-images can be prepared as necessary.

op-node/eth/receipts.go
protolambda: `ConsensusReceipt` sounds like a good idea, but for now I just generalized the function that the client in the op-node uses when fetching raw receipts (from the geth debug API)

op-node/eth/transactions.go
protolambda: you're right, will rename it. It's meant to be consistent with the other functions (`<Transform>Transactions`) but it can be clearer.

PR #5450, opened 2023-04-14
op-program: pre-image oracle communication

protolambda:
**Description**

Implement pre-image oracle communication. Follows specs in #5412 : this implementation of the raw pre-image interface can function over blocking syscalls. As well as the pre-image hints (one directional and optional, see specs)

**Tests**

Round-trip through the read/write sides, and check unexpected EOF. There's not really any strictly invalid string, other than lengths having to make sense, but since the program is trusted we don't impose limits here.

**Metadata**

Fix CLI-3825

### DIFF
--- a/op-program/preimage/hints.go
+++ b/op-program/preimage/hints.go
@@ -1,0 +1,65 @@
+package preimage
+
+import (
+	"encoding/binary"
+	"fmt"
+	"io"
+)
+
+// HintWriter writes hints to an io.Writer (e.g. a special file descriptor, or a debug log),
+// for a pre-image oracle service to prepare specific pre-images.
+type HintWriter struct {
+	w io.Writer
+}
+
+var _ Hinter = (*HintWriter)(nil)
+
+func NewHintWriter(w io.Writer) *HintWriter {
+	return &HintWriter{w: w}
+}
+
+func (hw *HintWriter) Hint(v Hint) {
+	hint := v.Hint()
+	hintBytes := make([]byte, 4, 4+len(hint)+1)
+	binary.BigEndian.PutUint32(hintBytes[:4], uint32(len(hint)))
+	hintBytes = append(hintBytes, []byte(hint)...)
+	hintBytes = append(hintBytes, 0) // to block writing on
+	_, err := hw.w.Write(hintBytes)
+	if err != nil {
+		panic(fmt.Errorf("failed to write pre-image hint: %w", err))
+	}
+}
+
+// HintReader reads the hints of HintWriter and passes them to a router for preparation of the requested pre-images.
+// Onchain the written hints are no-op.
+type HintReader struct {
+	r io.Reader
+}
+
+func NewHintReader(r io.Reader) *HintReader {
+	return &HintReader{r: r}
+}
+
+func (hr *HintReader) NextHint(router func(hint string) error) error {
+	var lengthPrefix [4]byte
+	if _, err := io.ReadFull(hr.r, lengthPrefix[:]); err != nil {
+		if err == io.EOF {
+			return io.EOF
+		}
+		return fmt.Errorf("failed to read hint length prefix: %w", err)
+	}
+	length := binary.BigEndian.Uint32(lengthPrefix[:])
+	payload := make([]byte, length)
+	if length > 0 {
+		if _, err := io.ReadFull(hr.r, payload); err != nil {
+			return fmt.Errorf("failed to read hint payload (length %d): %w", length, err)
+		}
+	}
+	if err := router(string(payload)); err != nil {
+		return fmt.Errorf("failed to handle hint: %w", err)
+	}
+	if _, err := hr.r.Read([]byte{0}); err != nil {
+		return fmt.Errorf("failed to read trailing no-op byte to unblock hint writer: %w", err)
+	}
+	return nil
+}
--- a/op-program/preimage/hints.go
+++ b/op-program/preimage/hints.go
@@ -20,8 +20,8 @@ func NewHintWriter(w io.Writer) *HintWriter {
 
 func (hw *HintWriter) Hint(v Hint) {
 	hint := v.Hint()
-	hintBytes := make([]byte, 4, 4+len(hint)+1)
-	binary.BigEndian.PutUint32(hintBytes[:4], uint32(len(hint)))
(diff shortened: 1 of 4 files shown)


op-program/preimage/hints.go
```diff
@@ -0,0 +1,65 @@
...
+var _ Hinter = (*HintWriter)(nil)
+
+func NewHintWriter(w io.Writer) *HintWriter {
+	return &HintWriter{w: w}
+}
+
+func (hw *HintWriter) Hint(v Hint) {
+	hint := v.Hint()
+	hintBytes := make([]byte, 4, 4+len(hint)+1)
+	binary.BigEndian.PutUint32(hintBytes[:4], uint32(len(hint)))
```
ajsutton, line 24: There's a lot of `4` as a magic number. Is there a constant for uint32 bytes length we could use somewhere? Otherwise probably worth defining one here just to reduce the cognitive load for understanding what's happening.

Could potentially even do:
```
const uint32Size = 4
const hintOverheadSize = uint32Size + 1 // one extra byte for the trailing 0 byte
func (hw *HintWriter) Hint(v Hint) {
	hint := v.Hint()
	hintBytes := make([]byte, uint32Size, len(hint)+hintOverheadSize)
	binary.BigEndian.PutUint32(hintBytes[:uint32Size], uint32(len(hint)))
	hintBytes = append(hintBytes, []byte(hint)...)
	hintBytes = append(hintBytes, 0) // to block writing on
	_, err := hw.w.Write(hintBytes)
	if err != nil {
		panic(fmt.Errorf("failed to write pre-image hint: %w", err))
	}
}
```

op-program/preimage/oracle.go
```diff
@@ -0,0 +1,80 @@
...
+	if _, err := o.rw.Write(h[:]); err != nil {
+		panic(fmt.Errorf("failed to write key %s (%T) to pre-image oracle: %w", key, key, err))
+	}
+
+	var lengthPrefix [8]byte
+	if _, err := io.ReadFull(o.rw, lengthPrefix[:]); err != nil {
+		panic(fmt.Errorf("failed to read pre-image length of key %s (%T) from pre-image oracle: %w", key, key, err))
+	}
+
+	length := binary.BigEndian.Uint64(lengthPrefix[:])
```
ajsutton, line 34: Any reason we can't just use:
```suggestion
	var length uint64
	if err := binary.Read(o.rw, binary.BigEndian, &length); err != nil {
		panic(fmt.Errorf("failed to read pre-image length of key %s (%T) from pre-image oracle: %w", key, key, err))
	}

```

op-program/preimage/oracle.go
```diff
@@ -0,0 +1,80 @@
...
+		return fmt.Errorf("failed to read requested pre-image key: %w", err)
+	}
+	value, err := getPreimage(key)
+	if err != nil {
+		return fmt.Errorf("failed to serve pre-image %s request: %w", key, err)
+	}
+
+	var lengthPrefix [8]byte
+	binary.BigEndian.PutUint64(lengthPrefix[:], uint64(len(value)))
+	if _, err := o.rw.Write(lengthPrefix[:]); err != nil {
```
ajsutton, line 70: Similarly can we just do:
```suggestion
	if err := binary.Write(o.rw, binary.BigEndian, uint64(len(value))); err != nil {
```

op-program/preimage/hints.go
ajsutton: Or if we're ok with a slightly less optimised version:
```
func (hw *HintWriter) Hint(v Hint) {
	hint := v.Hint()
	var hintBytes []byte
	hintBytes = binary.BigEndian.AppendUint32(hintBytes, uint32(len(hint)))
	hintBytes = append(hintBytes, []byte(hint)...)
	hintBytes = append(hintBytes, 0) // to block writing on
	_, err := hw.w.Write(hintBytes)
	if err != nil {
		panic(fmt.Errorf("failed to write pre-image hint: %w", err))
	}
}
```

op-program/preimage/hints.go
```diff
@@ -0,0 +1,65 @@
...
+	r io.Reader
+}
+
+func NewHintReader(r io.Reader) *HintReader {
+	return &HintReader{r: r}
+}
+
+func (hr *HintReader) NextHint(router func(hint string) error) error {
+	var lengthPrefix [4]byte
+	if _, err := io.ReadFull(hr.r, lengthPrefix[:]); err != nil {
```
ajsutton, line 45: Could potentially use the `binary.Read(hr.r, binary.BigEndian, &length)` approach here too.

ajsutton (approved):
This works, is pretty simple and tests are good. Have made some suggestions to potentially simplify but I'm ok with it as-is if the alternative `binary.Read` approach isn't a good fit for some reason.

op-program/preimage/hints.go
```diff
@@ -0,0 +1,65 @@
...
+	"io"
+)
+
+// HintWriter writes hints to an io.Writer (e.g. a special file descriptor, or a debug log),
+// for a pre-image oracle service to prepare specific pre-images.
+type HintWriter struct {
+	w io.Writer
+}
+
+var _ Hinter = (*HintWriter)(nil)
```
ajsutton, line 15: Wouldn't the `HintWriter` and the `OracleClient` be in the client package so the client program can write out requests and the reader/server would be in the host module?

protolambda: Putting them in the same package makes testing between the client and server easier, since this is the inner-most layer where the client and server actually directly communicate data to each other.

op-program/preimage/hints.go
```diff
@@ -0,0 +1,64 @@
...
+func NewHintWriter(w io.Writer) *HintWriter {
+	return &HintWriter{w: w}
+}
+
+func (hw *HintWriter) Hint(v Hint) {
+	hint := v.Hint()
+	var hintBytes []byte
+	hintBytes = binary.BigEndian.AppendUint32(hintBytes, uint32(len(hint)))
+	hintBytes = append(hintBytes, []byte(hint)...)
+	hintBytes = append(hintBytes, 0) // to block writing on
```
trianglesphere, line 26: If this is already length-prefixed, why do we need the trailing byte?

PR #5452, opened 2023-04-14
op-program: Populate preimage store

ajsutton:
**Description**

Implements a pre-fetcher that takes in the hints from `HintReader.NextHint` and request for preimages from `OracleServer.NextPreimageRequest` and then populates the KV store by fetching required data.

Modifies the op-program main to use this fetcher and an in-memory KV store as the backing for oracle requests. This makes the fetching mode run using the same key/value requests and access as non-fetching mode would just using hints to fetch the required data on demand.

**Additional context**

Builds on #5463 and #5465

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3747/store-oracle-content-in-key-value-store

### DIFF
--- a/op-node/sources/debug_client.go
+++ b/op-node/sources/debug_client.go
@@ -1,0 +1,49 @@
+package sources
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+)
+
+type DebugClient struct {
+	callContext CallContextFn
+}
+
+func NewDebugClient(callContext CallContextFn) *DebugClient {
+	return &DebugClient{callContext}
+}
+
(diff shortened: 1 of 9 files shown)


protolambda (changes requested):
PR looks good, but I think we want to use the L1/L2 interfaces with error handling in the prefetcher, since it's running in the host.

protolambda:
Alternatively we could also add error-returns to the "fetching oracle" code, if there's functionality there that we want to share rather than plugging directly into a L1/L2 client from `sources` package. And then define wrappers in the client package that map an error-returning oracle to one that escalates them to critical errors.

protolambda:
As per Slack discussion: updated the prefetcher to take an interface matching the L1/L2 rpc bindings (and added debug bindings for the MPT-node / code fetching). This way the prefetcher can handle errors and bubble them up to the get-preimage caller (who can then shut down the program from the host side etc.).

protolambda (approved):

protolambda (approved):

PR #5453, opened 2023-04-14
op-program: Switch L1 Oracle to retrieve *types.Header

ajsutton:
**Description**

Change L1 Oracle to return `*types.Header` instead of `eth.BlockInfo` so that the header RLP can be generated and stored in the key value store.

**Additional context**

Requires changing from using `sources.L1Client` to a plain `ethclient.Client` which has fewer features.  We'll either need to reimplement things like advanced receipt checking and the `trustRPC` setting or remove those options from the CLI.

### DIFF
--- a/op-program/client/l1/cache.go
+++ b/op-program/client/l1/cache.go
@@ -1,7 +1,6 @@
 package l1
 
 import (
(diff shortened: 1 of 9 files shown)


ajsutton:
Closing in favour of #5463

PR #5455, opened 2023-04-14
[op-batcher] Add Compressor abstraction, and new implementation with shadow compression

mdehoog:
**Description**

Adds a new Compressor interface which provides an abstraction around compressor details. This allows us to plug-in different compressor implementations to the ChannelOut.

Currently the batcher uses a "target" and "approx compression ratio" to determine how much data to write to a frame. This estimation implementation is hard to get right, and due to the variation of data on the L2, can either cause transactions that are 50% full, or channels to span multiple transactions due to too aggressive compression ratio estimation.

This PR also introduces a second compression implementation with a separate shadow writer that is written to and flushed immediately, which gives better estimation of the compressed size. This means we get much more accurate compression estimation, at the expense of more compute (compressing twice). The `Flush()` call will cause slightly less efficient compression, which is fine because we don't actually use the flushed buffer for writing the TX, only for size estimation.

The compression ratio and target size flags remain to support the old "target" compression style, which is still the default.

**Tests**

Added tests for the shadow compressor.

**TODOs**


mdehoog:
Before I fix all the tests, want to make sure there's alignment that this approach makes sense 😄

op-e2e/actions/l2_batcher.go
```diff
@@ -133,7 +133,7 @@ func (s *L2Batcher) Buffer(t Testing) error {
 		if s.l2BatcherCfg.GarbageCfg != nil {
 			ch, err = NewGarbageChannelOut(s.l2BatcherCfg.GarbageCfg)
 		} else {
-			ch, err = derive.NewChannelOut()
+			ch, err = derive.NewChannelOut(s.l2BatcherCfg.MaxL1TxSize)
```
sebastianst, line 136:
```suggestion
			ch, err = derive.NewChannelOut(s.l2BatcherCfg.MaxL1TxSize-1)
```
That's because we prepend a version byte to the tx data.

sebastianst (commented):
I like the `ChannelCompressor` abstraction. It pushes the actual channel size handling down into the `ChannelOut`, which can make sense. I left one comment where I propose that we inject a `ChannelCompressor` interface into the `NewChannelOut` constructor and have two impls, one the old way and one with the shadow flushing compressor.

If other reviewers don't like the idea of extending the `ChannelOut` this way, we could instead have the batcher's `ChannelBuilder` hold two `ChannelOut`s for each channel, one used in the usual way without flushing and another one taking on the shadowing role, calling flush on it after adding any block.

Is there a way to clone a `zlip` compressor state? If yes, it would be much more optimal to clone the compressor state before adding any block, then flush and then check the output size. This way, the result would be much closer to the actual channel out's output size. I'm also very curious how much the current implementation of flushing after adding every block will deviate from only flushing once on close.

mdehoog:
@sebastianst I originally went about implementing cloning zlib state (pulled it all out into a separate library and implemented clone methods: https://github.com/mdehoog/cloneable-zlib), but when I benchmarked it, the cloning added too much overhead (cloning on every new batch was causing 5-10x overhead, depending on batch size), so quickly abandoned that approach

I think your suggestion of splitting the implementation in two is a good one, will do that today

op-node/rollup/derive/channel_compressor.go
```diff
@@ -0,0 +1,74 @@
+package derive
+
+import (
+	"bytes"
+	"compress/zlib"
+)
+
+type ChannelCompressor struct {
```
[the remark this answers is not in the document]
sebastianst: Just thinking about naming, I realize that it's a plain `Compressor` abstraction that has nothing to do with channels. The output size checking one is just a `LimitedOutputCompressor` :D

op-e2e/actions/l2_batcher.go
mdehoog: this already happens here: https://github.com/ethereum-optimism/optimism/blob/b72138c696f90af23b879338a3fb80e547e74ac1/op-batcher/batcher/driver.go#L90

op-batcher/batcher/shadow_compressor_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+			err = sc.Close()
+			require.NoError(t, err)
+
+			buf, err := io.ReadAll(sc)
+			require.NoError(t, err)
+
+			if inputSize > 0 {
+				require.LessOrEqual(t, float64(len(buf))/float64(inputSize), test.minRatio)
+			}
```
trianglesphere, line 100: What is `minRatio` testing? I see it ranges from `0.02` to `0.80` in the test cases, but I'm not sure why it'd be different.

op-batcher/flags/flags.go
```diff
@@ -135,3 +147,41 @@ func CheckRequired(ctx *cli.Context) error {
...
+var CompressorKinds = []CompressorKind{
+	RatioCompressorKind,
+	ShadowCompressorKind,
+}
+
+func (kind CompressorKind) String() string {
+	return string(kind)
+}
+
+func (kind *CompressorKind) Set(value string) error {
```
trianglesphere, line 172: Is this function used?

op-batcher/batcher/channel_builder.go
```diff
@@ -96,10 +98,20 @@ func (cc *ChannelConfig) Check() error {
...
-func (c ChannelConfig) InputThreshold() uint64 {
-	return uint64(float64(c.TargetNumFrames) * float64(c.TargetFrameSize) / c.ApproxComprRatio)
+func (cc *ChannelConfig) NewCompressor() (derive.Compressor, error) {
+	switch cc.CompressorKind {
+	case flags.ShadowCompressorKind:
+		return NewShadowCompressor(
+			cc.TargetFrameSize,
+			cc.TargetNumFrames,
+		)
+	default:
```
trianglesphere, line 108: This code works, but the default statement is doing a lot of work to handle the cases where the compressor kind is not set while constructing the batcher in process.

trianglesphere (commented):
The compression abstraction & shadow compressor implementation looks good. I'm slightly concerned about how which compressor is selected (lots of tracing config), but that may also be from how it's set up.

One interesting check would be to ensure that if we have the ratio compressor (or not the ratio compressor) that `ApproxCompressionRatio` is set. It should always be set via the CLI flags, but that may not carry through to the config.

op-batcher/batcher/shadow_compressor_test.go
mdehoog: it's different because one is compressing all 0s and the other is compressing random data (which hits 0.67 ratio normally)

however this is kinda just testing zlib, which is useless in these specific tests... i'll just remove this

op-batcher/flags/flags.go
mdehoog: this is required as part of the CLI generic flag: https://github.com/urfave/cli/blob/b158c1095ecce00068fafa6d636752bbaeb5f51d/flag_generic.go#L10

trianglesphere (commented):
changes lgtm. The L1 Ops hive test is failing & I'd like to see it pass before merging. (There are some other failing tests, but that is because you're an external contributor, hive should pass with external contributors).

mdehoog:
Hive test failures seem unrelated to this change, they are failing on `develop` too: https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/18165/workflows/e1b3b550-719b-4919-9cf2-e370c84a680f/jobs/712075

trianglesphere (approved):
lgtm

PR #5457, opened 2023-04-14
CI: fix how slither determines what to detect and when to fail.

maurelian:
**Description**

This change implements the information provided in these slither issues:
https://github.com/crytic/slither/issues/1408
https://github.com/crytic/slither-action/issues/38

Apparently 'fail_pedantic' defaults to true, which overrides the 'exclude_*' settings, causing failures on optimization and informational severity findings.

As is this config should:
1. Fail only on High severity findings
2. Print Medium severity findings but not fail on them.

This PR also deletes any findings in the slither db which
were less than high severity. They are now simply ignored via this corrected config.

**Tests**

I've included a temp change to ensure that slither will run on this PR. Having wiped out most of the `slither.db.json` one would expect it to fail if the config is not. working as expected.


packages/contracts-bedrock/slither.config.json
```diff
@@ -1,9 +1,11 @@
 {
-  "detectors_to_exclude": "assembly-usage,block-timestamp,naming-convention,solc-version,low-level-calls,boolean-equality",
```
maurelian, line 2: This list on the right was growing by a 'whack-a-mole' like process.
The remaining list is deliberate.

packages/contracts-bedrock/slither.config.json
```diff
@@ -1,9 +1,11 @@
...
+  "detectors_to_exclude": "incorrect-shift-in-assembly",
+  "fail_high": true,
+  "fail_pedantic": false,
+  "exclude_optimization": true,
   "exclude_informational": true,
   "exclude_low": true,
   "exclude_medium": true,
-  "exclude_high": false,
```
maurelian, line 6: could leave this in, but the file was getting too verbose.

tynes (approved):

packages/contracts-bedrock/slither.config.json
```diff
@@ -1,9 +1,11 @@
 {
-  "detectors_to_exclude": "assembly-usage,block-timestamp,naming-convention,solc-version,low-level-calls,boolean-equality",
+  "detectors_to_exclude": "incorrect-shift-in-assembly",
+  "fail_high": true,
+  "fail_pedantic": false,
+  "exclude_optimization": true,
   "exclude_informational": true,
   "exclude_low": true,
   "exclude_medium": true,
```
0xalpharush, line 8: > Print Medium severity findings but not fail on them.
Medium detectors are being excluded

maurelian: I think I changed my mind about printing without failing, in order to avoid confusion and reduce the noise level.

PR #5463, opened 2023-04-15
Add a HeaderRLP method to BlockInfo

ajsutton:
**Description**

Adds a `HeaderRLP` method to `BlockInfo` to allow accessing the RLP of the block header which is required by op-program.

To support this the op-node `EthClient` now stores header data internally using the `eth.BlockInfo` interface rather than `HeaderInfo`. The `HeaderInfo` struct has been made internal to its module and is now a small wrapper over the geth `*types.Header` which adds block hash caching and the `BlockInfo` methods.  This ensures the data required to create the header RLP is available and the `types.Header` RLP serialisation can be used without introducing additional code into `op-node` that needs to be updated for hard forks.  The geth header was already being created as part of calculating the block hash when the RPC wasn't trusted so this isn't new functionality.

Additionally, the Geth `*types.Block` no longer implements the full `BlockInfo` interface so a wrapper for that was added similar to the `headerBlockInfo` wrapper. This only affected test code.

**Metadata**

- https://linear.app/optimism/issue/CLI-3747/store-oracle-content-in-key-value-store

### DIFF
--- a/op-node/sources/types.go
+++ b/op-node/sources/types.go
@@ -6,6 +6,7 @@ import (
 	"math/big"
 	"strings"
 
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/holiman/uint256"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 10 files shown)


protolambda (approved):

op-node/sources/types.go
```diff
@@ -149,7 +138,12 @@ func (hdr *rpcHeader) checkPostMerge() error {
 }
 
 func (hdr *rpcHeader) computeBlockHash() common.Hash {
-	gethHeader := types.Header{
+	gethHeader := hdr.createGethHeader()
+	return gethHeader.Hash()
+}
+
+func (hdr *rpcHeader) createGethHeader() *types.Header {
+	return &types.Header{
```
trianglesphere, line 146: Should strongly consider adding an allfields linter here (https://github.com/subtle-byte/allfields)

PR #5466, opened 2023-04-17
op-program: Support running in offline mode.

ajsutton:
**Description**

Update the setup code for op-program to support running in offline mode.  A `--datadir` CLI arg is added to specify the directory to use to store preimage data. If the `--l1` and `--l2` options are specified the preimages are fetched, otherwise only the pre-populated preimages are used to replay the derivation.

In-memory kv store is still supported in fetching mode if no `--datadir` is supplied because it makes testing easier (and the host part of op-program is essentially just a test harness).

**Tests**

Tests for the new CLI arg and config option. Will need to add e2e tests to cover the setup changes.

**Additional context**

Builds on #5463, #5465 and #5452

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3748/read-only-execution-mode

### DIFF
--- a/op-program/host/cmd/main.go
+++ b/op-program/host/cmd/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/ethereum-optimism/optimism/op-program/host/version"
 	"github.com/ethereum-optimism/optimism/op-program/preimage"
 	oplog "github.com/ethereum-optimism/optimism/op-service/log"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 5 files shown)


op-program/host/cmd/main.go
```diff
@@ -99,31 +100,51 @@ func setupLogging(ctx *cli.Context) (log.Logger, error) {
...
-	l1Fetcher, err := l1.NewFetchingOracle(ctx, logger, cfg)
-	if err != nil {
-		return fmt.Errorf("connect l1 fetcher: %w", err)
+	var kv kvstore.KV
+	if cfg.DataDir == "" {
+		logger.Info("Using in-memory storage")
+		kv = kvstore.NewMemKV()
+	} else {
+		logger.Info("Creating disk storage", "datadir", cfg.DataDir)
+		if err := os.Mkdir(cfg.DataDir, 0755); err != nil && !errors.Is(err, os.ErrExist) {
```
Inphi, line 115: nit: A bit nicer to create directory recursively

Inphi (approved):
LGTM

Inphi (approved):

PR #5108, opened 2023-03-10
fix(ctb): Use `callWithMinGas` in the XDM

clabby:
# Overview

Moves from `SafeCall.call` -> `SafeCall.callWithMinGas` in the `CrossDomainMessenger` contract.

**Metadata**
Fixes CLI-3599

PR #5470, opened 2023-04-17
feat(ctb): Account for worst-case `CALL` gas in `SafeCall`

clabby:
## Overview

Adjusts `SafeCall.callWithMinGas` to account for the worst-case cost of 3/5 of the `CALL` opcode's dynamic gas factors:
* `address_access_cost`
* `positive_value_cost`
* `value_to_empty_account_cost`

As for `memory_expansion_cost` & `code_execution_cost`, these are expected to be covered by the `minGasLimit` itself.


clabby:
* develop
  * **PR #5470** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/5470" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/5470?utm_source=stack-comment).

packages/contracts-bedrock/contracts/test/invariants/SafeCall.t.sol
```diff
@@ -78,25 +92,43 @@ contract SafeCaller_Actor is StdUtils {
...
+    function performSafeCallMinGas(
+        uint64 gas,
+        uint64 minGas,
+        address to,
+        uint8 value,
+        uint16 reservedGas
+    ) external {
+        // Only send to EOAs - we exclude the console as it has no code but reverts when called
+        // with a selector that doesn't exist due to the foundry hook.
+        vm.assume(to.code.length == 0 && to != 0x000000000000000000636F6e736F6c652e6c6f67);
```
tynes, line 104: Should be able to access this as `CONSOLE`
https://github.com/foundry-rs/forge-std/blob/fc560fa34fa12a335a50c35d92e55a6628ca467c/src/Base.sol#L11

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -391,9 +391,10 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         //   2. The amount of gas provided to the call to the target contract is at least the gas
         //      limit specified by the user. If there is not enough gas in the callframe to
         //      accomplish this, `callWithMinGas` will revert.
-        // Additionally, if there is not enough gas remaining to complete the execution after the
-        // call returns, this function will revert.
-        bool success = SafeCall.callWithMinGas(_tx.target, _tx.gasLimit, _tx.value, _tx.data);
+        // Additionally, we use 0 `_reservedGas` here because it does not matter if this context
+        // reverts with "Out of Gas." This is because the user will be able to re-attempt
+        // finalization of the withdrawal with a higher gas limit.
+        bool success = SafeCall.callWithMinGas(_tx.target, _tx.gasLimit, _tx.value, 0, _tx.data);
```
tynes, line 397: nit: prefer kwargs syntax when its unclear what the arguments have become, ie the magic `0`

packages/contracts-bedrock/contracts/test/invariants/SafeCall.t.sol
clabby: `CONSOLE` is a part of `Base.sol`, and this actor contract only inherits from `StdUtils`.

maurelian:
Fixes CLI-3832

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -358,7 +361,9 @@ abstract contract CrossDomainMessenger is
         );
 
         xDomainMsgSender = _sender;
-        bool success = SafeCall.callWithMinGas(_target, _minGasLimit, _value, _message);
+        // Add an extra 30_000 reserved gas to retain after the safe call to ensure that there
+        // is enough gas to finish the execution within this context afterwards.
+        bool success = SafeCall.callWithMinGas(_target, _minGasLimit, _value, 30_000, _message);
```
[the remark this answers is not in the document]
maurelian: I propose that we: 

1. Revert #5108. 
2. Rebase on #5444 once it's in.

Combine the old gas buffer check with the new reentrancy protection check to be:

```diff
    if (xDomainMsgSender != Constants.DEFAULT_L2_SENDER 
+	|| gasleft() >= _minGasLimit + RELAY_GAS_REQUIRED
    ) {
            failedMessages[versionedHash] = true;
            emit FailedRelayedMessage(versionedHash);
            return;
        }
```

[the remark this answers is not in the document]
tynes: I do like this direction but am unsure of exactly how it would help to unstick a message as we do not allow for passing in a different `minGasLimit`. The old contracts do allow for this, there is a different function `replayMessage`. This at least would give us better monitoring into the issue, we would be able to observe the set of failed messages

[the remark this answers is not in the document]
tynes: by `RELAY_GAS_REQUIRED` do you mean `MIN_GAS_CONSTANT_OVERHEAD`?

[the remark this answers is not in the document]
maurelian: > ... we do not allow for passing in a different `minGasLimit`

We don't allow modifying `minGasLimit`, but `SafeCall.call()` can accept `gasleft()` as its `_gas` [argument](https://github.com/ethereum-optimism/optimism/blob/d460e7670923a115fe2c3c3af14914df8394b849/packages/contracts-bedrock/contracts/libraries/SafeCall.sol#L17-L26) which it will forward.

`RELAY_GAS_REQUIRED` was the constant value in a previous version of the contract which was "the amount of gas held in reserve to guarantee that relay execution completes".

https://github.com/ethereum-optimism/optimism/blob/608ebc9a8a597d57acc71d4740fe9b1980357380/packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol#L370-L377

There were other issues with the implementation (not accounting for the SSTORE to set `xDomainMessageSender`), but we can address that.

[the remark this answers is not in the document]
maurelian: I think the pattern we need to avoid is any logic that triggers a revert based on the `gasleft()`.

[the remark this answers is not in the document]
clabby: I agree that not reverting here is the safer route. I've rebased the PR on top of the other XDM changes and changed it up.

packages/contracts-bedrock/contracts/L2/L2CrossDomainMessenger.sol
```diff
@@ -17,12 +17,12 @@ import { L2ToL1MessagePasser } from "./L2ToL1MessagePasser.sol";
  */
 contract L2CrossDomainMessenger is CrossDomainMessenger, Semver {
     /**
-     * @custom:semver 1.1.0
+     * @custom:semver 1.2.0
      *
      * @param _l1CrossDomainMessenger Address of the L1CrossDomainMessenger contract.
      */
     constructor(address _l1CrossDomainMessenger)
-        Semver(1, 1, 0)
```
maurelian, line 25: I just realized that we forgot to bump semver on #5444. Do we need to bump this up twice to `1.3.0` to account for that? 

@tynes I think you have the clearest ideas about that.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -152,7 +152,7 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         address _guardian,
         bool _paused,
         SystemConfig _config
-    ) Semver(1, 3, 1) {
+    ) Semver(1, 4, 1) {
```
maurelian, line 155: Need to reset the patch version when the minor version increments:

```suggestion
    ) Semver(1, 4, 0) {
```

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -345,17 +357,22 @@ abstract contract CrossDomainMessenger is
             "CrossDomainMessenger: message has already been relayed"
         );
 
+        // If there is not enough gas left to perform the external call and finish the execution,
+        // silently fail and assign the message to the failedMessages mapping.
```
refcell, line 361: Not a big deal, so just a comment but I think instead of "silently fail", it'd be better verbiage to say something to the tune of "return early" since silently failing implies a resulting unexpected state or state to be out of wack.

packages/contracts-bedrock/contracts/L1/L1CrossDomainMessenger.sol
```diff
@@ -20,12 +20,12 @@ contract L1CrossDomainMessenger is CrossDomainMessenger, Semver {
...
     /**
-     * @custom:semver 1.1.0
+     * @custom:semver 1.2.0
      *
      * @param _portal Address of the OptimismPortal contract on this network.
      */
     constructor(OptimismPortal _portal)
-        Semver(1, 1, 0)
+        Semver(1, 2, 0)
         CrossDomainMessenger(Predeploys.L2_CROSS_DOMAIN_MESSENGER)
```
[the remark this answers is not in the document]
tynes: Let's bump first outside of this PR and then rebase this on top of that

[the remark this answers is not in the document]
maurelian: resolved #5491

refcell (approved):
Looks good to me :handshake:

clabby:
Adding the `mergify-ignore` label so that this doesn't get merged in before we're confident that these changes plug all necessary holes.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -391,8 +391,6 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
         //   2. The amount of gas provided to the call to the target contract is at least the gas
         //      limit specified by the user. If there is not enough gas in the callframe to
         //      accomplish this, `callWithMinGas` will revert.
```
maurelian, line 393: Would the following be slightly more accurate?

```suggestion
        //   2. The amount of gas provided to the execution context of the the target contract is at
        //       least the gas limit specified by the user. If there is not enough gas in the
        //       callframe to accomplish this, `callWithMinGas` will revert.
```

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -391,8 +391,6 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
-        // Additionally, if there is not enough gas remaining to complete the execution after the
-        // call returns, this function will revert.
```
maurelian, line 395: It's slightly unclear to me if this is meant to indicate that this is handled by `SafeCall.callWithMinGas` or if it's just friendly reminder about how the EVM works. 

Also I don't think it'd be obvious to an outsider what the purpose of this statement is. 

Maybe it would help to append it with: 

```
This ensure that calls which do not provide sufficient additional gas will not finalize
the withdrawal, so that it can be attempted again.
```

Honestly maybe I'm just being pedantic and there's a lot of words here already, so I'm not going to block over this.

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
```diff
@@ -53,15 +53,21 @@ library SafeCall {
...
             // addition, exactly 51 gas is consumed between the below `GAS` opcode and the `CALL`
-            // opcode, so it is factored in with some extra room for error.
-            if lt(gas(), div(mul(64, add(_minGas, 200)), 63)) {
+            // opcode, so it is factored in with a good bit of extra room for error. The 40_000
+            // base buffer is to account for the worst case of the dynamic cost of the `CALL`
+            // opcode's `address_access_cost`, `positive_value_cost`, and
+            // `value_to_empty_account_cost` factors with an added buffer of 5,700 gas. It is
+            // still possible to self-rekt by passing a minimum gas limit that does not account
+            // for the `memory_expansion_cost + code_execution_cost` factors of the dynamic
+            // cost of the `CALL` opcode.
```
maurelian, line 69: I think this is a bit clearer

```suggestion
            // still possible to self-rekt by initiating a withdrawal with a minimum gas limit
            // that does not account for the `memory_expansion_cost + code_execution_cost` factors
            // of the dynamic cost of the `CALL` opcode.
```

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -345,17 +357,22 @@ abstract contract CrossDomainMessenger is
...
         );
 
+        // If there is not enough gas left to perform the external call and finish the execution,
+        // return early and assign the message to the failedMessages mapping.
         // If `xDomainMsgSender` is not the default L2 sender, this function
         // is being re-entered. This marks the message as failed to allow it
         // to be replayed.
-        if (xDomainMsgSender != Constants.DEFAULT_L2_SENDER) {
+        if (
+            gasleft() < _minGasLimit + RELAY_RESERVED_GAS ||
```
tynes, line 366: I'm slightly worried about this breaking gas estimation

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
clabby: Added this with a slight modification, this is more clear than the current comment imo :100:

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
clabby: Agree, added

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
clabby: This comment was removed because the `callWithMinGas` function will **not** revert if there is not enough gas remaining to complete the rest of `finalizeWithdrawalTransaction` after the call is performed. In the case that there's not enough gas to complete execution, `finalizeWithdrawalTransaction` will OOG, allowing for the caller to retry with a higher gaslimit.

clabby: This comment was also wrong prior to this change - The behavior of `callWithMinGas` has never attempted to reserve any more gas than was required to forward `_minGasLimit` to the target context.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
clabby: Yeah, this totally breaks gas estimation, good spot. Adding a revert if the `tx.origin` is the estimation address.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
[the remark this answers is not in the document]
tynes: I am starting to feel reluctant about how large of a diff this is

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -345,17 +357,32 @@ abstract contract CrossDomainMessenger is
...
+            if (tx.origin == Constants.ESTIMATION_ADDRESS) {
+                revert("CrossDomainMessenger: failed to relay message");
+            }
+
             return;
         }
 
         xDomainMsgSender = _sender;
-        bool success = SafeCall.callWithMinGas(_target, _minGasLimit, _value, _message);
+        bool success = SafeCall.call(_target, gasleft(), _value, _message);
```
smartcontracts, line 385: This doesn't actually reserve the gas amount

smartcontracts: Still passing in the full gas amount, the gas buffer needs to be reserved explicitly (gasleft() - relay gas buffer). I think best to just use callWithMinGas here and add a new parameter for reservedGas. It's much clearer.

clabby: The reason I didn't fully reserve the remaining gas was because the `RELAY_RESERVED_GAS` constant factors in both the worst case cost of the `CALL` opcode (minus the memory expansion / execution costs) as well as the gas required to finish the execution. We could split it up into 2 constants?

fwiw, I initially went w/ modifying `callWithMinGas`, but @maurelian brought up a great point that we can just not revert in this context if a threshold of remaining gas is not met, which is much safer. See the thread above in this PR ^

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
maurelian: OMG, I was reviewing that comment as if it was an addition. 🫠

packages/contracts-bedrock/contracts/test/invariants/CrossDomainMessenger.t.sol
```diff
@@ -94,12 +100,12 @@ contract RelayActor is StdUtils {
 contract XDM_MinGasLimits is Messenger_Initializer {
     RelayActor actor;
 
-    function setUp() public virtual override {
+    function init(bool doRevert) public virtual {
```
maurelian, line 103: why `init`?

clabby:
```solidity
contract MultipleSetupTest is Test {
    function setUp() public {
        // ...
    }

    function setUp(bool _in) public {
        // ...
    }

    function testExample() public {
        // ...
    }
}
```

```
$ forge test

Running 3 tests for test/Sandbox.t.sol:ChallengeTest
Failing tests:
Encountered 1 failing test in test/Sandbox.t.sol:MultipleSetupTest
[FAIL. Reason: Multiple setUp functions] setUp() (gas: 0)
```

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -345,17 +357,32 @@ abstract contract CrossDomainMessenger is
...
             failedMessages[versionedHash] = true;
             emit FailedRelayedMessage(versionedHash);
+
+            // Revert in this case if the transaction was triggered by the estimation address. This
+            // should only be possible during gas estimation or we have bigger problems. Reverting
+            // here will make the behavior of gas estimation change such that the gas limit
+            // computed will be the amount required to relay the message, even if that amount is
+            // greater than the minimum gas limit specified by the user.
+            if (tx.origin == Constants.ESTIMATION_ADDRESS) {
+                revert("CrossDomainMessenger: failed to relay message");
```
maurelian, line 378: Hmmm, I'm not sure this will work as the Portal's `finalizeWithdrawalTransaction` won't bubble it up.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
maurelian: > See the thread above in this PR ^

@smartcontracts the thread is gone from this view, please take a look at it [here](https://github.com/ethereum-optimism/optimism/pull/5470#discussion_r1169217738).

maurelian: > We could split it up into 2 constants?

I believe it was that way previously and that it makes the most sense, at least to my mind.

packages/contracts-bedrock/contracts/test/invariants/CrossDomainMessenger.t.sol
maurelian: okee dokee. :)

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
smartcontracts: Replacing `callWithMinGas` with `call` just has the effect of not explicitly guaranteeing that the sub-call will receive enough gas. So the minimum gas limit is no longer respected. I feel like this is trying to make a larger diff than the original intention which should be like a 10 line diff (add two buffers, make sure it always respects minimum gas limit).

I do understand that minimum gas limit is a pain here and we should consider modifying things in the future but I feel like this is a large unnecessary diff at this point

maurelian: > Replacing `callWithMinGas` with `call` just has the effect of not explicitly guaranteeing that the sub-call will receive enough gas.

It's true that `callWithMinGas` explicitly guarantees that the sub-call will receive enough gas, but it enforces that guarantee by bricking the withdrawal.

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
```diff
@@ -52,16 +65,23 @@ library SafeCall {
         bytes memory _calldata
     ) internal returns (bool) {
         bool _success;
+        bool _halt = checkMinGas(_minGas, 0);
         assembly {
-            // Assertion: gasleft() >= ((_minGas + 200) * 64) / 63
+            // Assertion: gasleft() >= (_minGas * 64) / 63 + 40_000
```
maurelian, line 70: This comment should be on `checkMinGas` above.

Some or all of the rest comment block below also needs to be moved or edited to acommodate this change.

clabby: Still in the process of splitting these - will do.

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
```diff
@@ -35,6 +35,19 @@ library SafeCall {
...
     }
 
+    /**
+     * @notice Helper function that returns `true` if there is not enough gas remaining to cover
+     *         the minimum gas requirement for a call + an optional reserved gas amount.
+     * @param _minGas      The minimum amount of gas that may be passed to the target context.
+     * @param _reservedGas Optional amount of gas to reserve for the caller after the execution
+     *                     of the target context.
+     */
+    function checkMinGas(uint256 _minGas, uint256 _reservedGas) internal view returns (bool _halt) {
```
maurelian, line 45: Nit:
> returns `true` if there is not enough gas remaining

This is unintuitive to me. For example, if I just encountered this function in the wild (with a less descriptive name for the return value), ie:

```
bool ret = checkMinGas(_minGas, 0)
```
I would assume that `ret == true` means we _do_ have enough gas.

clabby: The parameter is named `_halt` for this reason - glad to add some extra documentation here or flip it around.

maurelian: How about: 
- Flip it
- rename to `hasMinGas()`

maurelian, line 45: Another nit: we've always avoided named vars. We should reenable that slither check.

```suggestion
    function checkMinGas(uint256 _minGas, uint256 _reservedGas) internal view returns (bool) {
```

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
clabby: Comments updated, RFC

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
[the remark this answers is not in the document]
refcell: Think we're in a good place to resolve this convo

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
```diff
@@ -82,13 +111,11 @@ library SafeCall {
...
-            // ((((_minGas + 200) * 64) / 63) - 49) >= _minGas
-            // ((((_minGas + 200) * 64) / 63) - 51) * (63 / 64) >= _minGas
-            // Both inequalities hold true for all possible values of `_minGas`.
+            // The call will be supplied at least ((_minGas * 64) / 63) gas due to the
+            // above assertion. This ensures that, in all circumstances (except for when the
+            // `_minGas` does not account for the `memory_expansion_cost` and `code_execution_cost`
+            // factors of the dynamic cost of the `CALL` opcode), the call will receive at least
+            // the minimum amount of gas specified.
             _success := call(
                 gas(), // gas
```
maurelian, line 120: Comment only:

It's interesting to me that we give `gas()` (or rather `min(gas() * 63 / 64, gas())`) to the actual `CALL`, rather than `_minGas`, which in the Portal is `tx.gasLimit`.

https://github.com/ethereum-optimism/optimism/blob/0470651b3c425a74f9ba24113f0a695645e8a7f9/packages/contracts-bedrock/contracts/L1/OptimismPortal.sol#L394

It seems fine since we do all the work above to ensure that when we get here, `gas()` is going to be plenty higher `tx.gasLimit`. 

Also, for whatever reason `CALL` doesn't seem to fail if `gas()` is less then the stack value consumed by gas.

refcell: I think it makes sense to pass any additional gas a caller provides into the subcontext since the parameter is specifically annotated as a _minimum_ gas limit.

refcell (approved):
This looks pretty ironed out to me at this point pending other's reviews.

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
maurelian: Yeah, I agree. Just needed to think it through.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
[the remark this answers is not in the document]
maurelian: > Think we're in a good place to resolve this convo

Oh, didn't realize this was my comment to begin with. Resolved.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
maurelian: Resolving on the basis that this applies to outdated code.

maurelian (approved):

clabby:
Waiting for @tynes' approval to remove the `do-not-merge` and `mergify-ignore` labels.

tynes:
Do you mind squashing this PR in a commit or a few commits?

packages/contracts-bedrock/contracts/libraries/SafeCall.sol
```diff
@@ -35,6 +35,41 @@ library SafeCall {
...
+     *      3.) Because EIP-150 ensures that a maximum of 63/64ths of the remaining gas in the call
+     *          frame may be passed to a subcontext, we need to ensure that the gas will not be
+     *          truncated.
+     *      4.) Use wisely. This function is not a silver bullet.
+     */
+    function hasMinGas(uint256 _minGas, uint256 _reservedGas) internal view returns (bool) {
+        bool _hasMinGas;
+        assembly {
+            _hasMinGas := iszero(
+                lt(gas(), add(div(mul(_minGas, 64), 63), add(40000, _reservedGas)))
```
tynes, line 67: could you get rid of `iszero` by using `gt` instead of `lt`?

clabby: Mentioned in slack as well, x-posting:
```
iszero(lt(x, y)) == x >= y
gt(x, y) == x > y
```

The off-by-one here isn't a big deal, but imo should stick to the `>=` assertion (6 gas + 3-6 gas for dups isn't a big deal w/ the giant buffers)

tynes (approved):

PR #5459, opened 2023-04-14
sdk: fix withdrawal bug

smartcontracts:
**Description**

Fixes a bug in the sdk that allows it to work with migrated withdrawals. This fix is necessary to ensure that withdrawals initiated with the legacy system can be finalized on bedrock. Includes test coverage of computing the storage slot in the message passer as well as a test using anvil to actually finalize a withdrawal.

Closes CLI-3836

PR #5474, opened 2023-04-17
op-chain-ops: legacy withdrawal script refactoring

tynes:
**Description**

Clean up the script that is used to migrate legacy withdrawals.
This script is very important for testing the legacy withdrawal
migration.

A bug was introduced with #4911
which was a sherlock audit fix. This was a bug because the change
resulted in the withdrawal hashes changing, meaning that the storage
slots computed client side also changed. This resulted in breaking the
script. This commit reverts this change. The changes landing in
#5470 will cause the
buffer in the gas limit for the  migrated withdrawals to be too small,
so we can reintroduce this code behind a switch with the network.
Its not ideal that the migration code isn't going to match 1:1
with goerli but thats ok because we will be able to thoroughly
test it.

### DIFF
--- a/op-chain-ops/cmd/withdrawals/main.go
+++ b/op-chain-ops/cmd/withdrawals/main.go
@@ -22,11 +22,14 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/eth"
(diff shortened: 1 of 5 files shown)


tynes:
If these fixes are merged and then https://github.com/ethereum-optimism/optimism/pull/5459 is rebased on top of this, it should fix @roninjin10's issue. Will need to actually see it work but feeling pretty sure that the combo of these fixes will work

roninjin10:
Amazing!

mslipper:
Is this a pure revert, or was additional functionality introduced?

tynes:
> Is this a pure revert, or was additional functionality introduced?

I left in the unit tests and modularization that enabled the unit tests but the underlying functionally was reverted

mslipper (approved):

tynes:
Can I get a review from @ethereum-optimism/ecopod ?

roninjin10 (approved):

PR #5475, opened 2023-04-17
fix(CI): Fail CI if `contracts-bedrock` tests fail

clabby:
## Overview

Adds a new task to CI, `contracts-bedrock-tests`, which runs `forge test` rather than `forge coverage`. `forge coverage` will not return an error exit code if any tests that it runs fail.


maurelian (commented):
Left a nit, but overall LGTM.

tynes (approved):

mslipper (approved):

PR #5477, opened 2023-04-17
op-program: Hint reader recovery

Inphi:
**Description**

Attempt to recover the hint read stream when the NextHint callback fails

**Metadata**

- Fixes CLI-3848


ajsutton (approved):
LGTM.

PR #5478, opened 2023-04-17
contracts-bedrock: update constructor require statment

tynes:
**Description**

Updates the constructor check in the `L2OutputOracle` so that it is correct. It previously compared two different types of units (blocknumbers and timestamps). The real thing that we need to ensure is that neither is 0 because that will result in a broken system.


refcell (approved):
Nice! Looks good to me!

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -95,11 +95,11 @@ contract L2OutputOracle is Initializable, Semver {
         address _proposer,
         address _challenger,
         uint256 _finalizationPeriodSeconds
-    ) Semver(1, 2, 0) {
+    ) Semver(1, 3, 0) {
```
refcell, line 98: The natspec above uses `1.2.0` - should bump that as well to `1.3.0`

refcell (approved):
Small natspec semver bump, otherwise looks good!

maurelian (approved):
LGTM assuming resolution of the semver in natspec

clabby (changes requested):
LGTM- will approve once the natspec comment above the constructor has been updated from `1.2.0` -> 1.3.0`

tynes:
I've updated the natspec, good call on that. Its very easy to forget to update the natspec tag when updating the version, i wonder if there is a better way to manage this.

maurelian:
> Its very easy to forget to update the natspec tag when updating the version, i wonder if there is a better way to manage this.

@tynes do we really need the tag?

refcell:
> > Its very easy to forget to update the natspec tag when updating the version, i wonder if there is a better way to manage this.
> 
> @tynes do we really need the tag?

Maybe split semver tags into a gardening ticket to explore enforcing a version bump if the file is changed?

tynes:
> > Its very easy to forget to update the natspec tag when updating the version, i wonder if there is a better way to manage this.
> 
> @tynes do we really need the tag?

We are supposed to do a patch semver for any change to the source that doesn't change the bytecode

clabby (approved):

PR #5483, opened 2023-04-18
op-node: Load runtime config at new L1 heads

0xfuturistic:
The op-node's runtime configuration currently fetches the address of the system sequencer from the L1. However, it's only loaded at initialization, causing issues if any underlying variables tracked by it change during runtime. For instance, if the system owner changes the address of the sequencer in the system config, the runtime config of a node already initialized won't reflect this change. This can result in outdated information, leading to errors.

- An outdated runtime config can cause a node to accept gossip from an obsolete sequencer while punishing and rejecting gossip from the current one, since it uses an outdated P2PSequencerAddress for the verification of the signatures of the payloads.
- The same verification function also prevents non-sequencer signers from publishing payloads, since the node ensures the payload’s signature passes verification before publishing blocks. This has the benefit of protecting the node from punishment by other network peers if they accidentally publish blocks while not being the sequencer. However, if the runtime config is outdated, this safeguard fails catastrophically, not only allowing a node to publish payloads when it is not a sequencer but even blocking the node from publishing blocks if it became the system sequencer after the node was initialized. The latter happens because the node's gossip layer may still consider its signer a non-sequencer, making the payload's signature verification fail.

To address these issues, the runtime config needs to be kept up to date. One way to achieve this is to load the runtime config every time a new L1 head is received (something missing implementation but mentioned in the in-line comments of the runtime config). Although this approach adds overhead, it's worth it to avoid operating with outdated data. An implementation of this idea could be simple: the driver can call the Load function of the node's runtime config when it receives a new L1 head in the main event loop. To do this, we could create an interface for a new runtime config in the driver that allows the driver to access the Load function of the runtime config. Then, we can create a new variable inside the struct of the driver that points to an object with this interface, and, upon initialization of the driver by the node, we pass the node's runtime config to it. This approach is identical to the one used by the gossip layer to interact with the runtime config.


0xfuturistic:
shoutout to @tynes for his feedback on PR #5420 (also going even further and splitting the changes of that PR into more manageable proposals like this one)

PR #5486, opened 2023-04-18
chore: Add warning to contract-defs.ts

roninjin10:
Add a comment that warns us to never do this pattern again

### DIFF
--- a/packages/contracts/src/contract-defs.ts
+++ b/packages/contracts/src/contract-defs.ts
@@ -12,6 +12,12 @@ export const getContractDefinition = (name: string): any => {
   // the contracts, but we need the contracts to be compiled before the contract-artifacts.ts
   // file can be generated.
   // eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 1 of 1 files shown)


tynes (approved):

packages/contracts/src/contract-defs.ts
```diff
@@ -12,6 +12,12 @@ export const getContractDefinition = (name: string): any => {
   // the contracts, but we need the contracts to be compiled before the contract-artifacts.ts
   // file can be generated.
   // eslint-disable-next-line @typescript-eslint/no-var-requires
```
tynes, line 14: need to move the eslint hint below the warning

clabby (approved):
LGTM once Mark's comment is resolved.

packages/contracts/src/contract-defs.ts
tynes: hey @roninjin10 are you planning on fixing this?

PR #5487, opened 2023-04-18
op-program: Specify l2 block number to stop derivation at

ajsutton:
**Description**

Adds a `--l2.blocknumber` option that the derivation will stop at and check the claim. This avoids the need to find a L1 head that includes exactly the batches required to derive the L2 block the claim is from and none beyond that.

Also found that `main_test` was getting tests interfering with each other when run in parallel. That shouldn't be possible, but something in the cli lib seems to do it so unmarking them as parallel safe.

**Tests**

Added unit tests.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3818/fpp-halt-execution-when-target-l2-block-reached

### DIFF
--- a/op-program/client/driver/driver.go
+++ b/op-program/client/driver/driver.go
@@ -24,18 +24,20 @@ type L2Source interface {
 }
 
 type Driver struct {
-	logger       log.Logger
-	pipeline     Derivation
-	l2OutputRoot func() (eth.Bytes32, error)
+	logger         log.Logger
+	pipeline       Derivation
+	l2OutputRoot   func() (eth.Bytes32, error)
+	targetBlockNum uint64
 }
 
-func NewDriver(logger log.Logger, cfg *rollup.Config, l1Source derive.L1Fetcher, l2Source L2Source) *Driver {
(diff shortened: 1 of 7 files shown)


Inphi (approved):
Looks good.
At first i thought it would be nicer UX to have the cli immediately halt with an error if there isn't enough data to derive up to the target block number. But this only makes sense to do when the data sources are fully loaded and we've abstracted that aspect away (for good reason). 
The program can still timeout so this works out.

ajsutton:
Yeah> Looks good.
> At first i thought it would be nicer UX to have the cli immediately halt with an error if there isn't enough data to derive up to the target block number. But this only makes sense to do when the data sources are fully loaded and we've abstracted that aspect away (for good reason). 
> The program can still timeout so this works out.

Given the L1 head is fixed it will actually fail if it can't reach the target height. It just currently still tries to compare the claims. 

Potentially we could make the error message more friendly rather than just saying the claim failed but I think it logs the chain head it gets to when the derivation process ends so it should be reasonably clear.

PR #5488, opened 2023-04-19
op-e2e: Extract system test helpers

ajsutton:
**Description**

Reduce duplication in e2e system tests by extracting helper functions for common tasks.

Extracted helpers:
* `InitParallel` - sets parallel (unless configured not to) and handles the verboseGeth logging config.
* `SendDepositTransaction` - sends a deposit transaction and confirms its inclusion on L1 and L2.
* `SendL2Tx` - sends a normal L2 transaction and can verify it shows up correctly on the sequencer and replicas
* Some withdrawal helpers to send a withdrawal tx on L2, then prove and finalise it on L1.

**Metadata**

- Prep work for https://linear.app/optimism/issue/CLI-3855/setup-more-complex-chain-in-fpp-e2e-tests


op-e2e/system_test.go
```diff
@@ -425,19 +360,12 @@ func TestMintOnRevertedDeposit(t *testing.T) {
...
-	tx = types.NewTx(reconstructedDep)
-	receipt, err = waitForTransaction(tx.Hash(), l2Verif, 10*time.Duration(cfg.DeployConfig.L1BlockTime)*time.Second)
-	require.NoError(t, err)
-	require.Equal(t, receipt.Status, types.ReceiptStatusFailed)
+	SendDepositTx(t, cfg, l1Client, l2Verif, opts, func(l2Opts *DepositTxOpts) {
+		l2Opts.ToAddr = toAddr
+		// trigger a revert by transferring more than we have available
+		l2Opts.Value = new(big.Int).Mul(common.Big2, startBalance)
+		l2Opts.ExpectedStatus = types.ReceiptStatusFailed
+	})
```
ajsutton, line 368: Introducing this pattern which essentially lets helper functions accept optional parameters with useful defaults. It doesn't just pass a struct because the `nil` value for a lot of values can be indistinguishable from an explicitly set value (e.g. for a uint64 value) and making things a pointer is pain.
Passing a function that takes the defaults and adjusts them as needed seems like the neatest solution but very happy to hear other ideas.

protolambda (approved):
Nice test-util refactor, this cleans things up a lot.

And new `InitParallel` on `TestRegolith` sub-tests makes sense.

I thought about adding `t.Helper()` on the new util functions, so test reports list the line within the test rather than the util, but we can add that later if necessary, and since the helpers do multiple things it's still useful to locate the error in the helper.

refcell:
@mergifyIO dequeue

refcell:
@mergifyio unqueue

refcell:
FYI @ajsutton removed from the queue since it was holding up other prs on train

trianglesphere:
@ajsutton need to update `system_fpp_test.go` with the new helpers

op-e2e/op_geth_test.go
```diff
@@ -90,6 +90,7 @@ func TestPreregolith(t *testing.T) {
 	for _, test := range tests {
 		test := test
 		t.Run("GasUsed_"+test.name, func(t *testing.T) {
+			InitParallel(t)
```
refcell, line 93: Adding parallelism to subtests is a huge improvement, very nice!

refcell (approved):
This is awesome :fire:

op-e2e/op_geth_test.go
trianglesphere: This actually doesn't make much of a difference when running the entire test suite. Parallelizing these tests takes it from 2.5s -> 1s, however the top level tests are already running in parallel with the other top level tests & the total test time even with maximally parallel tests is order 3-4 minutes.

ajsutton: Yeah its a minor thing, but I thought just the top level call was enough to run them all fully parallel and it turns out its not so... As much as anything its just setting up the pattern that you always call `InitParallel` as the first thing in the test, then if we need to do any common setup like setting up the geth logging we can do it in that one method, plus all tests are built running in parallel from the start to do what we can to keep build times down.

ajsutton:
Thought I'd found a couple of bits of work that finally didn't have to go in a conga line of PRs but forgot about the `InitParallel` call in the new FPP test...

Shame mergify doesn't just automatically dequeue things rather than blocking everything else.

PR #5490, opened 2023-04-19
fix(contracts-bedrock): SafeCall FFI Interface Fuzzing

refcell:
**Description**

Currently ci fails since the `to` address passed into `SafeCall` is fuzzed with the ffi interface address, resulting in the test failing. This simply adds an assumption that the `to` address is not the ffi interface contract address.

### DIFF
--- a/packages/contracts-bedrock/contracts/test/SafeCall.t.sol
+++ b/packages/contracts-bedrock/contracts/test/SafeCall.t.sol
@@ -23,6 +23,8 @@ contract SafeCall_call_Test is CommonTest {
         vm.assume(to != address(0x000000000000000000636F6e736F6c652e6c6f67));
         // don't call the create2 deployer
         vm.assume(to != address(0x4e59b44847b379578588920cA78FbF26c0B4956C));
+        // don't call the ffi interface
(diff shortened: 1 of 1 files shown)


tynes (approved):

packages/contracts-bedrock/contracts/test/SafeCall.t.sol
```diff
@@ -23,6 +23,8 @@ contract SafeCall_call_Test is CommonTest {
         vm.assume(to != address(0x000000000000000000636F6e736F6c652e6c6f67));
         // don't call the create2 deployer
         vm.assume(to != address(0x4e59b44847b379578588920cA78FbF26c0B4956C));
+        // don't call the ffi interface
+        vm.assume(to != address(0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f));
```
clabby, line 27: Would prefer if we didn't hard-code the addresses here - the FFI Interface is available in scope

refcell: For style changes, let's put in gardening. I agree, I was just following the previous work

refcell: This comment also just removed the on-merge-train label. Going to resolve so we don't block others.

PR #5492, opened 2023-04-19
op-program: host-client interaction via I/O pipes

Inphi:
**Description**

Building on #5395, we separate host-client communication to use pipes rather than function calls. This lets us later run the client  on a separate execution environment.

To avoid Oracle requests from racing ahead of a hint prefetch, hint writes now block until a prefetch has completed. This is done by making the hints pipe bidirectional.

**Metadata**

- Fixes CLI-3753

### DIFF
--- a/op-program/host/host.go
+++ b/op-program/host/host.go
@@ -95,10 +95,6 @@ func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
 	routeHints(logger, hHost, hinter)
 	launchOracleServer(logger, oracleServer, getPreimage)
 
-	// TODO(CLI-XXX): This is a hack to wait for the oracle server and hint router to begin polling for requests
-	// before the program starts. This should be replaced with a more robust solution.
-	//time.Sleep(time.Second * 1)
-
 	return cl.ClientProgram(
 		logger,
 		cfg.Rollup,
--- a/op-program/host/host.go
+++ b/op-program/host/host.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"time"
 
 	"github.com/ethereum-optimism/optimism/op-node/chaincfg"
 	"github.com/ethereum-optimism/optimism/op-node/client"
@@ -89,16 +88,16 @@ func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
 	pClientRW, pHostRW := bidirectionalPipe()
 	oracleServer := preimage.NewOracleServer(pHostRW)
 	// Setup pipe for hint comms
-	hHostR, hClientW := io.Pipe()
-	hHost := preimage.NewHintReader(hHostR)
+	hClientRW, hHostRW := bidirectionalPipe()
+	hHost := preimage.NewHintReader(hHostRW)
 	defer pHostRW.Close()
-	defer hHostR.Close()
+	defer hHostRW.Close()
 	routeHints(logger, hHost, hinter)
 	launchOracleServer(logger, oracleServer, getPreimage)
 
 	// TODO(CLI-XXX): This is a hack to wait for the oracle server and hint router to begin polling for requests
 	// before the program starts. This should be replaced with a more robust solution.
-	time.Sleep(time.Second * 1)
+	//time.Sleep(time.Second * 1)
 
 	return cl.ClientProgram(
 		logger,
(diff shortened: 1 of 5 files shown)


ajsutton (commented):
Looking good. I think the testing for this realistically just needs to be via e2e tests so if the new system_fpp_test is passing with these changes I'd say we can go ahead and merge them as a good next step towards splitting the client and host.

Inphi:
This needs a bit more work as the e2e fpp test sometimes fails.

op-program/host/host.go
```diff
@@ -80,54 +74,89 @@ func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
...
-	for {
-		if err = d.Step(ctx); errors.Is(err, io.EOF) {
-			break
-		} else if err != nil {
-			return err
-		}
-	}
-	if !d.ValidateClaim(eth.Bytes32(cfg.L2Claim)) {
-		return ErrClaimNotValid
+	// TODO(CLI-XXX): This is a hack to wait for the oracle server and hint router to begin polling for requests
```
Inphi, line 99: I'll follow up with a proper fix for the full client-host process separation.

ajsutton: I'm unclear why this is required - shouldn't the client program block once it tries to write until these components start reading from the socket?

If it is required I think we should get it right here before merging - adding arbitrary sleeps tends not to actually fix intermittency in CI because CI servers can run surprisingly slowly and unpredictably.

op-program/host/host.go
```diff
@@ -80,54 +74,89 @@ func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
...
+	// Setup pipe for preimage oracle interaction
+	pClientRW, pHostRW := bidirectionalPipe()
+	oracleServer := preimage.NewOracleServer(pHostRW)
+	// Setup pipe for hint comms
+	hHostR, hClientW := io.Pipe()
+	hHost := preimage.NewHintReader(hHostR)
+	defer pHostRW.Close()
+	defer hHostR.Close()
+	routeHints(logger, hHost, hinter)
+	launchOracleServer(logger, oracleServer, getPreimage)
```
trianglesphere, line 97: This means that we have three threads going at once
1. `ClientProgram` (main thread)
2. `routeHints`
3. `oracleServer`
I think we should be able to do it in two - client program & host. The host will have to multiplex between the two different pipes but I think it should at least be spun up as one unit (maybe having two thread internally) to better coordinate between the two FDs

Inphi: It's doable but doing so complicates the host programming. The host oracle and prefetcher will have to be rewritten to support a poll-based interface. And doing so also exposes too much of the internals of pipes (since Go doesn't provide an idiomatic way to select on pipes).

trianglesphere: Ok. That makes sense for why we need three threads. There's another part here where I don't love spwaning them this way. One thing I think that would be interesting is spawning the client program in it's own routine/thread only after these routines & ready

Inphi: Yeah that's the end goal of this host-client I/O separation. Spawning the client program in its own thread will be done in a follow up PR

op-program/host/host.go
Inphi: done

op-program/preimage/hints_test.go
```diff
@@ -73,20 +87,30 @@ func TestHints(t *testing.T) {
...
+			require.ErrorIs(t, err, cbErr)
+			var readHint string
+			err = hr.NextHint(func(hint string) error {
+				readHint = hint
+				return nil
+			})
+			require.NoError(t, err)
+			require.Equal(t, readHint, "two")
+		}()
+		wg.Wait()
```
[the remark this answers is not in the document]
Inphi: Fixed.

op-program/preimage/hints_test.go
```diff
@@ -20,26 +21,39 @@ func TestHints(t *testing.T) {
...
+				hw.Hint(rawHint(h))
 			}
-			require.NoError(t, err)
-		}
+			wg.Done()
+		}()
+
+		got := make(chan string, len(hints))
+		go func() {
+			println("MENA")
```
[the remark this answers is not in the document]
Inphi: Thanks. It's removed now.

op-program/preimage/hints_test.go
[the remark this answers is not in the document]
ajsutton: I think we need to apply the same timeout to this test as is done in the `test` func above.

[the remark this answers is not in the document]
Inphi: whoops! Thanks for pointing that out. Fixed once more.

ajsutton (approved):
Nice. :)

PR #5497, opened 2023-04-19
op-program: Support running program as a separate process

Inphi:
**Description**

Follow-up to #5492 to completely separate the host and client. This PR introduces a simple bootstrap protocol so the client can bootstrap itself from the environment.

**Tests**

`TestVerifyL2OutputRootDetached` - e2e test that runs the program in detached mode

**Metadata**

- Fixes [CLI-3752](https://linear.app/optimism/issue/CLI-3752)


op-program/host/host.go
```diff
@@ -51,83 +58,193 @@ func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
...
+		files[hClientWFd] = hClientW.Fd()
+		files[pClientRFd] = pClientRW.ReadFd()
+		files[pClientWFd] = pClientRW.WriteFd()
+		attr := &syscall.ProcAttr{
+			Env:   env,
+			Files: files,
 		}
-		l2Cl, err := sources.NewL2Client(l2RPC, logger, nil, l2ClCfg)
+		var err error
+		childPID, err = syscall.ForkExec(os.Args[0], os.Args, attr)
```
protolambda, line 153: Instead of forking manually we can also use the `os/exec` package `Command` utils to run a sub-process.

```go
cmd := exec.Command(...) 
cmd.ExtraFiles = []*io.File{pClientRWFile, hClientWFile}` (or something like this, the RW file may require some changes)
err := cmd.Start()
```

op-program/host/host.go
```diff
@@ -51,83 +58,193 @@ func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
...
-			return fmt.Errorf("failed to setup L2 RPC: %w", err)
+			return fmt.Errorf("failed to create preimage pipe: %w", err)
 		}
+		oracleServer := preimage.NewOracleServer(pHostRW)
+		launchOracleServer(logger, oracleServer, getPreimage)
+		defer pHostRW.Close()
+	} else {
+		r := os.NewFile(pClientRFd, "pClientR")
+		w := os.NewFile(pClientWFd, "pClientW")
+		pClientRW = &readWritePair{r, w}
```
protolambda, line 118: Can these not be combined into the same file?

Inphi: It's nicer this way. Allows the `readWritePair` to be abstracted as a `io.ReadWriter` so we just pass a single object to the oracle/prefetcher.

op-program/host/flags/flags.go
```diff
@@ -81,6 +81,11 @@ var (
 			return &out
 		}(),
 	}
+	Detached = cli.BoolFlag{
+		Name:   "detached",
+		Usage:  "Run the program as a separate process detached from the host",
+		EnvVar: service.PrefixEnvVar(envVarPrefix, "DETACHED"),
+	}
```
ajsutton, line 88: We should add tests for this flag in `main_test.go`.

Inphi: done

op-program/client/boot.go
```diff
@@ -0,0 +1,72 @@
...
+	// TODO(CLI-XXX): The rollup config will be hardcoded. It's configurable for testing purposes.
+	Rollup             *rollup.Config      `json:"rollup"`
+	L2ChainConfig      *params.ChainConfig `json:"l2_chain_config"`
+	L1Head             common.Hash         `json:"l1_head"`
+	L2Head             common.Hash         `json:"l2_head"`
+	L2Claim            common.Hash         `json:"l2_claim"`
+	L2ClaimBlockNumber uint64              `json:"l2_claim_block_number"`
+}
+
+type BootstrapOracleWriter struct {
```
Inphi, line 24: Please keep in mind that this bootstrap oracle isn't the final form. I'll follow up with a proper implementation in a later PR. This simple implementation was added just so the client can boot in a detached context.

op-program/client/boot_test.go
```diff
@@ -0,0 +1,51 @@
...
+	r, w := io.Pipe()
+	br := NewBootstrapOracleReader(r)
+	bw := NewBootstrapOracleWriter(w)
+
+	bootInfo := BootInfo{
+		Rollup:             new(rollup.Config),
+		L2ChainConfig:      new(params.ChainConfig),
+		L1Head:             common.HexToHash("0xfffff"),
+		L2Head:             common.HexToHash("0xfffff"),
+		L2Claim:            common.HexToHash("0xfffff"),
```
ajsutton, line 24: We should assign these different values so the test is actually confirming the specific value is preserved for each field and doesn't somehow mix up the fields.

op-program/host/host.go
```diff
@@ -24,8 +26,20 @@ type L2Source struct {
...
+func RunningProgramInClient() bool {
+	value, _ := os.LookupEnv(opProgramChildEnvName)
+	return value == "true"
+}
+
 // FaultProofProgram is the programmatic entry-point for the fault proof program
 func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
+	if RunningProgramInClient() {
+		cl.Main(logger)
+		// unreachable
```
ajsutton, line 40:
```suggestion
		panic("Client main should have exited process")
```
Just because everything unreachable one day gets reached and a clear panic will be so much easier to debug. :)

ajsutton (commented):
Looking good. I notice that when run in detached mode we get errors logged on shutdown which would be nice to avoid:
```
    host.go:196:                ERROR[04-24|08:32:11.194] pre-image hint router error              err="failed to read hint length prefix: read |0: file already closed"
    host.go:211:                ERROR[04-24|08:32:11.194] pre-image server error                   error="failed to read requested pre-image key: read |0: file already closed"
```

I'd be ok with doing that as a follow up if it makes more sense though. It's not a big deal, just that those are the last lines logged so it looks like things failed when they actually succeeded. I suspect we'll need to come up with a good way to do a clean shutdown as part of https://linear.app/optimism/issue/CLI-3868/exit-when-pre-image-data-missing anyway.

ajsutton:
oh those errors are just because we're checking for `io.ErrClosedPipe` but with these changes we actually get `fs.ErrClosed` in `routeHints` and `launchOracleServer` so should be an easy fix.

op-program/client/boot_test.go
Inphi: Good point. Done.

Inphi:
> oh those errors are just because we're checking for `io.ErrClosedPipe` but with these changes we actually get `fs.ErrClosed` in `routeHints` and `launchOracleServer` so should be an easy fix.

Yup that's the issue. It's fixed now.

op-program/host/host.go
Inphi: Done.

ajsutton (approved):

op-program/host/host.go
Inphi: this is no longer relevant. bidirectional pipes for host/client communication are now being used

PR #5498, opened 2023-04-19
op-e2e: Disable TestSystemMockP2P

trianglesphere:
This test is still consistently flaking & jamming up the mergify queue.


ajsutton (approved):

ajsutton:
Oh do we have a ticket open to fix it? Could probably reference it in the skip message as well.

trianglesphere:
> Oh do we have a ticket open to fix it? Could probably reference it in the skip message as well.

Created a ticket for this

PR #5499, opened 2023-04-19
chore(contracts-bedrock): ResourceConfig Sherlock Test

refcell:
**Description**

Adds back a `ResourceConfig` test previously closed in #5396 as part of CLI-3828 since @tynes added the related fix in #5437.

This tests that setting the `ResourceConfig`'s `baseFeeMaxChangeDenominator` to 1 will revert as reported in a sherlock finding.

**Metadata**

- Fixes CLI-3828


tynes:
Looks like gas snapshot needs to be regenerated

packages/contracts-bedrock/contracts/test/ResourceMetering.t.sol
```diff
@@ -134,6 +141,23 @@ contract ResourceMetering_Test is Test {
         assertEq(postBaseFee, 2125000000);
     }
 
+    function test_meter_baseFeeMaxChangeDenominator_reverts() external {
```
maurelian, line 144: This part of the name should be descriptive:

```suggestion
    function test_meter_denominatorEq1_reverts() external {
```

packages/contracts-bedrock/contracts/test/ResourceMetering.t.sol
```diff
@@ -134,6 +141,23 @@ contract ResourceMetering_Test is Test {
...
+        rcfg.baseFeeMaxChangeDenominator = 1;
+        meter.setParams(rcfg);
+        meter.use(target * elasticityMultiplier);
+
+        (, uint64 prevBoughtGas, ) = meter.params();
+        assertEq(prevBoughtGas, target * elasticityMultiplier);
+
+        vm.roll(initialBlockNum + 2);
+
+        vm.expectRevert("UNDEFINED");
```
maurelian, line 157: Is `UNDEFINED` like a magic string in Foundry or Solidity?

packages/contracts-bedrock/contracts/test/ResourceMetering.t.sol
maurelian: This test is a little funny since it's testing behaviour that is only  exposed by the `MeterUser` contract, and shouldn't technically be reachable in the base implementation. 

So, I think it should also have a comment in natspec to explain what it's doing.

maurelian (approved):
Requested some cosmetic changes, but approving now to unblock once resolved.

packages/contracts-bedrock/contracts/test/ResourceMetering.t.sol
refcell: No this is thrown by the call to `powWad` in our `Arithmetic` library since the first param will be `0`. See: https://github.com/ethereum-optimism/2023-04-optimism-judging/blob/e4321d3febf6c6dcd6cf74fd5712c268a0c2e10d/1-processed/2-medium/089.md?plain=1#L79-L87

refcell: Added to the natspec since this isn't obvious :)

packages/contracts-bedrock/contracts/test/ResourceMetering.t.sol
```diff
@@ -134,6 +141,32 @@ contract ResourceMetering_Test is Test {
         assertEq(postBaseFee, 2125000000);
     }
 
+    /**
+     * @notice This tests that the metered modifier reverts if
+     *         the ResourceConfig baseFeeMaxChangeDenominator
+     *         is set to 1.
+     *         Since the metered modifier internally calls
+     *         solmate's powWad function, it will revert
+     *         with the error string "UNEXPECTED" since the
```
maurelian, line 150:
```suggestion
     *         with the error string "UNDEFINED" since the
```

PR #5501, opened 2023-04-19
fix: withdrawal gas limit

tynes:
**Description**

Updates the gas limit in the migrated withdrawals codepath
to ensure that enough gas is used. This is required after https://github.com/ethereum-optimism/optimism/pull/5470
is merged

**Tests**

There is test coverage here that prevents a regression. We will need to run this tooling as part of the migration rehearsals to be super sure.

Closes CLI-3856

### DIFF
--- a/op-chain-ops/crossdomain/migrate.go
+++ b/op-chain-ops/crossdomain/migrate.go
@@ -20,7 +20,13 @@ var (
 )
 
 // MigrateWithdrawals will migrate a list of pending withdrawals given a StateDB.
-func MigrateWithdrawals(withdrawals SafeFilteredWithdrawals, db vm.StateDB, l1CrossDomainMessenger *common.Address, noCheck bool) error {
+func MigrateWithdrawals(
+	withdrawals SafeFilteredWithdrawals,
+	db vm.StateDB,
+	l1CrossDomainMessenger *common.Address,
+	noCheck bool,
+	chainID *big.Int,
+) error {
 	for i, legacy := range withdrawals {
 		legacySlot, err := legacy.StorageSlot()
 		if err != nil {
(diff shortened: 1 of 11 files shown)


op-chain-ops/crossdomain/migrate.go
```diff
@@ -97,13 +107,21 @@ func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger *com
...
+// The chain id is used to determine the overhead.
+func MigrateWithdrawalGasLimit(data []byte, chainID *big.Int) uint64 {
 	// Compute the upper bound on the gas limit. This could be more
 	// accurate if individual 0 bytes and non zero bytes were accounted
 	// for.
 	dataCost := uint64(len(data)) * params.TxDataNonZeroGasEIP2028
+
+	// Goerli has a lower gas limit than other chains.
+	overhead := uint64(200_000)
+	if chainID.Cmp(big.NewInt(420)) != 0 {
```
mslipper, line 119: If we're never running the Goerli migration again, is this even necessary? Is the idea here to preserve deterministic migrations for folks who want to migrate L2 themselves?

tynes: It is for 2 reasons:
- preserve deterministic migrations
- allow go code to be used for creating migrated withdrawal finalizations

op-chain-ops/crossdomain/migrate.go
```diff
@@ -97,13 +107,21 @@ func MigrateWithdrawal(withdrawal *LegacyWithdrawal, l1CrossDomainMessenger *com
...
+		overhead = 1_00_000
```
maurelian, line 120: Presumably meant this to be: 
```suggestion
		overhead = 1_000_000
```

roninjin10 (approved):
approving sdk changes to unblock

op-chain-ops/crossdomain/migrate.go
tynes: Good catch, yes you are right

mslipper (approved):

PR #5502, opened 2023-04-19
FE-1002: supportsTokenPair should return false whenever the check catches an error

tremarkley:
This caused a firefox-specific error where, err.stack was not defined and therefore threw an uncaught exception. Since there is not a reliable way to determine the type of JSON RPC error and since we expect non standard tokens to hit this code path, we should just default to returning false whenever an error is caught in this function.

### DIFF
--- a/packages/sdk/src/adapters/standard-bridge.ts
+++ b/packages/sdk/src/adapters/standard-bridge.ts
@@ -186,14 +186,7 @@ export class StandardBridgeAdapter implements IBridgeAdapter {
       // If the L2 token is not an L2StandardERC20, it may throw an error. If there's a call
       // exception then we assume that the token is not supported. Other errors are thrown. Since
       // the JSON-RPC API is not well-specified, we need to handle multiple possible error codes.
-      if (
-        err.message.toString().includes('CALL_EXCEPTION') ||
-        err.stack.toString().includes('execution reverted')
-      ) {
-        return false
-      } else {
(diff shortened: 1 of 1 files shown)


tremarkley:
* develop
  * **PR #5502** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5502" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5502?utm_source=stack-comment).

packages/sdk/src/adapters/standard-bridge.ts
```diff
@@ -186,14 +186,7 @@ export class StandardBridgeAdapter implements IBridgeAdapter {
...
       // the JSON-RPC API is not well-specified, we need to handle multiple possible error codes.
-      if (
-        err.message.toString().includes('CALL_EXCEPTION') ||
-        err.stack.toString().includes('execution reverted')
-      ) {
-        return false
-      } else {
-        throw err
-      }
+      return false
```
roninjin10, line 189:
```suggestion
      if (!err?.message?.toString().includes('CALL_EXCEPTION') && !err?.stack?.toString().includes('execution reverted') {
        console.error('Unexpected error when checking bridge', e)
      }
      return false
```

roninjin10 (approved):

PR #5503, opened 2023-04-19
op-e2e: Add actual transactions to the FPP system test

ajsutton:
**Description**

Updates the FPP system test so it sends actual TX. Prior to the challenged section of chain it sends some TX to setup world state with some account balances and a withdrawal recorded to the L2 world state.  Then during the challenged period it sends another deposit transaction, a normal transaction and a withdrawal which modifies the state setup for both balances and the L2 world state.

Also stops the batch submitter and proposer before stopping the nodes to avoid them spamming the logs.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3855/setup-more-complex-chain-in-fpp-e2e-tests


trianglesphere (approved):
lgtm

PR #5506, opened 2023-04-20
op-program: Handle pre-images that already exist

ajsutton:
**Description**

The exact same receipt RLP may result from different transactions in different blocks so ErrAlreadyExists must be handled when storing trie nodes.

**Tests**

Added unit test.


trianglesphere (approved):

donatocayurin:
Thank you ❤️🙏

donatocayurin:
❤️🙏

PR #5508, opened 2023-04-20
fix(ctb): Remove `minGas` revert in `relayMessage`

clabby:
## Overview

Removes the revert in `relayMessage` when the minimum gas threshold is not met.

**Metadata**
Fixes CLI-3870


clabby:
* develop
  * **PR #5470** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5470" target="_blank">Graphite</a> 
    * **PR #5508** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5508" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5508?utm_source=stack-comment).

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -345,17 +350,32 @@ abstract contract CrossDomainMessenger is
...
             emit FailedRelayedMessage(versionedHash);
+
+            // Revert in this case if the transaction was triggered by the estimation address. This
+            // should only be possible during gas estimation or we have bigger problems. Reverting
+            // here will make the behavior of gas estimation change such that the gas limit
+            // computed will be the amount required to relay the message, even if that amount is
+            // greater than the minimum gas limit specified by the user.
+            if (tx.origin == Constants.ESTIMATION_ADDRESS) {
+                revert("CrossDomainMessenger: failed to relay message");
+            }
```
maurelian, line 372: Does this work for estimation? The Portal's finalizeWithdrawalTransaction won't bubble it up. 

Are we even sure that previous iterations of this method supported gas estimation?

I don't think I've ever come across this pattern in other production code before.  Feels weird that we're coming up with a whole new method to do this thing in 2023.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -345,17 +350,32 @@ abstract contract CrossDomainMessenger is
...
         );
 
+        // If there is not enough gas left to perform the external call and finish the execution,
+        // return early and assign the message to the failedMessages mapping.
         // If `xDomainMsgSender` is not the default L2 sender, this function
         // is being re-entered. This marks the message as failed to allow it
         // to be replayed.
-        if (xDomainMsgSender != Constants.DEFAULT_L2_SENDER) {
+        if (
+            !SafeCall.hasMinGas(_minGasLimit, RELAY_RESERVED_GAS) ||
```
maurelian, line 359: There's a lot of stuff happening between this check and the CALL...

maurelian: Including the xDomainMessageSender  SSTORE.

refcell: I believe the `RELAY_RESERVED_GAS` includes this plus the cost of execution after the `CALL` subcontext ends.

refcell: That's probably a constraint that should be explicitly stated if it's not already

refcell: Think this can be resolved given the discussion today and the outline in [XDM-Bricked-Withdrawals-Fix](https://www.notion.so/oplabs/XDM-Bricked-Withdrawals-Fix-b738c44a5e6047a687cfc796511108b4?pvs=4).

It would be nice to explicitly break up the implicit assumption that the `5,700` additional gas buffer from the `hasMinGas` check's `40,000` buffer covers these instructions. So it's explicit + clear going forward.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -415,17 +450,23 @@ abstract contract CrossDomainMessenger is
      * @return Amount of gas required to guarantee message receipt.
      */
     function baseGas(bytes calldata _message, uint32 _minGasLimit) public pure returns (uint64) {
```
clabby, line 452: IMO, we should revert in `sendMessage` if sending the result of `baseGas` is unachievable based on the constraints of the L1 block gas limit.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
clabby: I've added an explicit buffer within `baseGas` as well as the call to `hasMinGas`

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -415,17 +450,23 @@ abstract contract CrossDomainMessenger is
...
+                MIN_GAS_DYNAMIC_OVERHEAD_DENOMINATOR) +
+            // Gas reserved for the worst-case cost of 3/5 of the `CALL` opcode's dynamic gas
+            // factors. (Conservative)
+            RELAY_CALL_OVERHEAD +
+            // Relay reserved gas (to ensure execution of `relayMessage` completes after the
+            // subcontext finishes executing) (Conservative)
+            RELAY_RESERVED_GAS +
+            // Gas reserved for the execution between the `hasMinGas` check and the `CALL`
+            // opcode. (Conservative)
+            RELAY_GAS_CHECK_BUFFER;
```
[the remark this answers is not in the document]
refcell: Might want to add as a new linear ticket

packages/contracts-bedrock/tasks/check-l2.ts
```diff
@@ -274,9 +274,9 @@ const check = {
     const MIN_GAS_CALLDATA_OVERHEAD =
       await L2CrossDomainMessenger.MIN_GAS_CALLDATA_OVERHEAD()
     console.log(`  - MIN_GAS_CALLDATA_OVERHEAD: ${MIN_GAS_CALLDATA_OVERHEAD}`)
-    const MIN_GAS_CONSTANT_OVERHEAD =
-      await L2CrossDomainMessenger.MIN_GAS_CONSTANT_OVERHEAD()
-    console.log(`  - MIN_GAS_CONSTANT_OVERHEAD: ${MIN_GAS_CONSTANT_OVERHEAD}`)
+    const RELAY_CONSTANT_OVERHEAD =
+      await L2CrossDomainMessenger.RELAY_CONSTANT_OVERHEAD()
+    console.log(`  - MIN_GAS_CONSTANT_OVERHEAD: ${RELAY_CONSTANT_OVERHEAD}`)
```
refcell, line 279:
```suggestion
    console.log(`  - RELAY_CONSTANT_OVERHEAD: ${RELAY_CONSTANT_OVERHEAD}`)
```

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
refcell: Discussion in slack, but placing here for posterity: We should also consider **not** restricting this and place this logic in the sdk so future teams can adopt these contracts for various system constraints without modification and needing to engage an audit. If the change happens in the sdk, it's much more lightweight to change in the future.

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -345,17 +361,36 @@ abstract contract CrossDomainMessenger is
             "CrossDomainMessenger: message has already been relayed"
         );
 
+        // If there is not enough gas left to perform the external call and finish the execution,
+        // return early and assign the message to the failedMessages mapping.
+        // We are asserting that we have enough gas to:
+        // 1. Call the target contract (_minGasLimit + RELAY_CALL_OVERHEAD + RELAY_GAS_CHECK_BUFFER)
```
refcell, line 367: Should this be updated to:

```suggestion
        // 1. Call the target contract (_minGasLimit + RELAY_CALL_OVERHEAD + RELAY_RESERVED_GAS + RELAY_GAS_CHECK_BUFFER)
```

refcell (changes requested):

packages/contracts-bedrock/contracts/universal/CrossDomainMessenger.sol
clabby: The `RELAY_RESERVED_GAS` is reserved for the execution after the call - see #2 😄

tynes (approved):

refcell (approved):

refcell:
@mergifyio unqueue

maurelian (approved):

PR #5512, opened 2023-04-21
op-program: Cache blocks by number

ajsutton:
**Description**

Avoids repeatedly iterating from the head block back to the requested block number.

**Tests**

Mostly covered by existing tests since this is just an optimisation and externally visible behaviour shouldn't change. Added some extra unit tests to ensure we have coverage of requesting blocks both forwards and backwards (which shows up different potential bugs in the caching).

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3865/fpp-cache-block-by-number-lookups


trianglesphere (approved):
nice - this'll really reduce the number of block lookups we need to do.

PR #5516, opened 2023-04-21
fix(ci): Disable flaky atst test

roninjin10:
A test was flaky because it's using public endpoint for goerli.

This is a low value test so rather than waste time setting up anvil to run with a alchemy key I just quickly disabled it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -561,11 +561,11 @@ jobs:
       - run:
           name: build
           command: yarn build
-          working_directory: packages/atst
+          working_directory: packages/sdk
       - run:
           name: lint
           command: yarn lint:check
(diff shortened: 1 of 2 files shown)


trianglesphere (approved):

.circleci/config.yml
```diff
@@ -561,11 +561,11 @@ jobs:
       - run:
           name: build
           command: yarn build
-          working_directory: packages/atst
+          working_directory: packages/sdk
       - run:
           name: lint
           command: yarn lint:check
-          working_directory: packages/atst
+          working_directory: packages/sdk
```
roninjin10, line 568: this was copy paste pasta that I noticed and fixed while I was here

PR #5524, opened 2023-04-23
op-program: Add separate keystore for local keys

ajsutton:
**Description**

Adds a separate kvstore implementation to allow the host to serve local config via type 1 keys without having to add them to the global KV store.

The read side of this in the client program will come as part of https://linear.app/optimism/issue/CLI-3751/oracle-based-bootstrap 

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3850/fpp-separate-global-and-local-kv-lookups

### DIFF
--- a/op-program/host/host.go
+++ b/op-program/host/host.go
@@ -84,6 +84,9 @@ func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
 		}
 	}
 
+	localPreimageSource := kvstore.NewLocalPreimageSource(cfg)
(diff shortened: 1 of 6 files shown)


ajsutton:
@Inphi we should get https://github.com/ethereum-optimism/optimism/pull/5497/ merged first as there will be some conflicts with this but they should be easy to resolve. Also happy to adjust the details of this based on what you had in mind for the oracle based bootstrap but figured I'd put it up as its reasonably standalone and will help ensure we're on the same page about how that oracle based bootstrap should work in its final form.

Inphi (approved):
This works great. Feel free to merge this in early and I can resolve the conflicts.\

PR #5528, opened 2023-04-24
specs: Update op-program hint protocol

Inphi:
**Description**

Updates the hint protocol as implemented in #5492

**Metadata**

- Fixes CLI-3866


protolambda (approved):
Nice

PR #5529, opened 2023-04-24
contracts-bedrock: update portal min gas limit

tynes:
**Description**

Prevent users from sending cross domain messages with too small of a gas limit. Users must pay for resources they consume. The L2 network must store deposit transactions in their history and they are also included in the p2p gossip of unsafe blocks. A linearly increasing requirement based on the calldata size is introduced. A base of 21k minimum gas limit is required when there is no gas limit, the formula `calldata.length * 16 + 21000` is used to enforce a min gas limit.

https://github.com/ethereum-optimism/optimism/pull/5527 should be merged before this

**Tests**

Test coverage is added for both depositing a transaction and the min gas limit function directly, including fuzz tests.

**Additional context**

The higher the gas limit is, the more gas that is burned on L1

Fixes CLI-3847


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -186,6 +186,15 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
     }
 
+    /**
+     * @notice Computes the minimum gas limit for a deposit. The minimum gas limit
+     *         linearly increases based on the size of the calldata. This is to prevent
+     *         users from creating L2 resource usage without paying for it.
+     */
+    function minimumGasLimit(uint64 _byteCount) public pure returns (uint64) {
+        return _byteCount * 16 + 21000;
+    }
```
maurelian, line 196: nit:

```suggestion
    function minimumGasLimit(uint256 _byteCount) public pure returns (uint64) {
        return uint64(_byteCount) * 16 + 21000;
    }
```

I can take it or leave it, but it feels like we can probably make integration a bit easier to handling the casting here ourselves.

tynes: The cast can cause overflow in this case which is why I didn't go with this pattern

maurelian: Ahh, good call.

maurelian:
Fixes CLI-3847

refcell (approved):
Changes look good 👍

maurelian (approved):

PR #5532, opened 2023-04-25
op-program: Bootstrap program via local oracle

Inphi:
**Description**

Load program initialization parameters from the local oracle. Also moved types around to keep the client imports isolated from the host.

**Metadata**

- Fixes #CLI-3751


op-program/client/program.go
```diff
@@ -37,21 +36,18 @@ func Main(logger log.Logger) {
...
-	logger.Debug("Loaded boot info", "bootInfo", bootInfo)
+func RunProgram(logger log.Logger, preimageOracle io.ReadWriter, preimageHinter io.ReadWriter) error {
 
 	pClient := preimage.NewOracleClient(preimageOracle)
 	hClient := preimage.NewHintWriter(preimageHinter)
 	l1PreimageOracle := l1.NewPreimageOracle(pClient, hClient)
 	l2PreimageOracle := l2.NewPreimageOracle(pClient, hClient)
+
+	bootInfo := NewBootstrapClient(pClient).BootInfo()
+	logger.Debug("Program Bootstrapped", "bootInfo", bootInfo)
```
ajsutton, line 47: This is probably worth making Info level - when the fault proof client runs without the host this would be the only real way to confirm what inputs it is actually running with.

ajsutton (approved):
Looks good.

op-program/client/program.go
Inphi: Good point. Done.

PR #5533, opened 2023-04-25
fix(op-chain-ops): Compute local output root with op-node exported function

refcell:
**Description**

This is a quick-fix discovered during a migration sherlock issue retro. The local output root in the `op-chain-ops` migrate command is computed manually using the `Keccak256Hash` function. This pr updates the code to use the exported `ComputeL2OutputRoot` from the `op-node`.

**Metadata**

- Fixes #CLI-3883


trianglesphere (approved):

tynes:
This originally did not use a function because there was a bug in the function, this script is not covered in CI but will remember if this script breaks this is what caused it

trianglesphere:
> This originally did not use a function because there was a bug in the function, this script is not covered in CI but will remember if this script breaks this is what caused it

IIRC there wasn't a bug in the function, but it was being called with the parameters in the wrong order. https://github.com/ethereum-optimism/optimism/pull/5114/files#diff-8f7bbce915ade563953fda055c7b5c424357da7cc240efd98a70bfce15abf4c1

PR #5534, opened 2023-04-25
op-program: Embed goerli L2 chain config

ajsutton:
**Description**

The op-goerli chain config is now automatically provided when specifying `--network=goerli`.  The OP Goerli config has previously just been embedded in the bootstrap database so has to be defined in this PR.

**Tests**

Updated CLI and config tests.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3873/fpp-make-goerli-chain-config-available

### DIFF
--- a/op-program/host/cmd/main_test.go
+++ b/op-program/host/cmd/main_test.go
@@ -21,19 +21,20 @@ var (
 	l2HeadValue        = common.HexToHash("0x222222").Hex()
 	l2ClaimValue       = common.HexToHash("0x333333").Hex()
 	l2ClaimBlockNumber = uint64(1203)
-	l2Genesis          = core.DefaultGoerliGenesisBlock()
-	l2GenesisConfig    = l2Genesis.Config
+	// Note: This is actually the L1 goerli genesis config. Just using it as an arbitrary, valid genesis config
+	l2Genesis       = core.DefaultGoerliGenesisBlock()
+	l2GenesisConfig = l2Genesis.Config
 )
 
 func TestLogLevel(t *testing.T) {
 	t.Run("RejectInvalid", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


op-program/host/config/chaincfg.go
```diff
@@ -0,0 +1,41 @@
+package config
+
+import (
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+var OPGoerliChainConfig = &params.ChainConfig{
```
ajsutton, line 10: Have defined this here because it's currently only used by the fault proof program.  We possibly should move it into op-geth as the canonical source and it could then be updated there when new hard forks are scheduled.

op-program/host/cmd/main_test.go
```diff
@@ -54,181 +55,196 @@ func TestDefaultCLIOptionsMatchDefaultConfig(t *testing.T) {
...
 
 	for name, cfg := range chaincfg.NetworksByName {
 		name := name
 		expected := cfg
 		t.Run("Network_"+name, func(t *testing.T) {
-			cfg := configForArgs(t, replaceRequiredArg(t, "--network", name))
+			args := replaceRequiredArg("--network", name)
+			if name == "beta-1" {
+				// No built-in config for beta-1 which is fine as it's no longer in use.
+				// Newly added named networks should hook up the L2 chain config
```
Inphi, line 84: nit: Maybe we should remove `beta-1` from the chaincfg networks if it's no longer used.

Inphi (approved):
LGTM

op-program/host/cmd/main_test.go
ajsutton: Tracking in https://linear.app/optimism/issue/CLI-3887/remove-beta-1-network-definition

PR #5536, opened 2023-04-26
op-batcher: Properly drain state on L2 reorgs

trianglesphere:
**Description**
When a L2 reorg occurs the batcher should close out the pending channel and submit everything that it has before clearing the channel builder. The reason for this is that even though it may be submitting incorrect data it should not leave dangling channels. After the pending state is drained it can begin tracking the new L2 chain.

**Tests**

Unfortunately the action tests are not suitable for testing this because they have rewritten the batcher. It might be possible in op-e2e, but I don't see an easy way to trigger a L2 reorg & also asserts that it does this flushing.

**Metadata**

- Fixes CLI-3632


op-batcher/batcher/driver.go
```diff
@@ -294,7 +296,14 @@ func (l *BatchSubmitter) loop() {
...
 		case <-ticker.C:
-			l.loadBlocksIntoState(l.shutdownCtx)
+			if err := l.loadBlocksIntoState(l.shutdownCtx); errors.Is(err, ErrReorg) {
+				err := l.state.Close()
+				if err != nil {
+					l.log.Error("error closing the channel manager to handle a L2 reorg", "err", err)
+				}
+				l.publishStateToL1(queue, receiptsCh, true)
+				l.state.Clear()
+			}
```
ajsutton, line 306: Does it matter that we're going to fall step out of this error handling and call publishStateToL1 again below? I expect it winds up being a no-op but it feels a bit weird to have just reset everything and then carry on like everything's normal.

trianglesphere: I added a `continue` statement to escape the case stmt & continue with the select loop after the error case is handled.

trianglesphere: I think it's a no-op, but definitely weird.

ajsutton (approved):
LGTM.

PR #5545, opened 2023-04-26
feat(proxyd): consensus event dispatcher

felipe-op:
**Description**

This change introduces a hook to listen to events from `ConsensusPoller` that notifies external components when consensus has broken.

**Tests**

Modified current tests to cover the new functionality.

**Invariants**

n/a

**Additional context**

**Metadata**

- Fixes https://linear.app/optimism/issue/INF-150/emit-event-when-consensus-is-broken-in-a-backend-group

**TODOs**


proxyd/cache.go
```diff
@@ -75,6 +81,29 @@ func (c *redisCache) Put(ctx context.Context, key string, value string) error {
 	return err
 }
 
+func (c *redisCache) Clear(ctx context.Context) error {
+	patterns := []string{"lvc:*", "method:*"}
+
+	for _, p := range patterns {
```
Inphi, line 87: This could take a while to run; delaying bootup. In the heyday we'd have over 100k `method:*` keys cached.
Perhaps we can lazily delete `method:` keys by checking its expiration at lookup time. Or inspect the cached data to figure out if it part of a reorg'd chain. For example, using the block number/hash in the `eth_getBlockByNumber`. Hopefully there's a solution we can use that doesn't complicate caching logic.

mslipper: If we set expirations on all keys, then we can increment a "generation" number embedded in the key. Redis can then evict keys based on their expiration/LRU if memory gets low.

mslipper: Another advantage there is that you won't need to modify `patterns` above if there are additional key prefixes stored in Redis.

proxyd/proxyd.go
```diff
@@ -310,6 +310,12 @@ func Start(config *Config) (*Server, func(), error) {
 			if config.BackendGroups[bgName].ConsensusAsyncHandler == "noop" {
 				copts = append(copts, WithAsyncHandler(NewNoopAsyncHandler()))
 			}
+			if config.Cache.Enabled {
+				copts = append(copts, WithListener(func() {
+					err := cache.Clear(context.Background())
+					log.Warn("error invalidating cache", "err", err)
```
mslipper, line 316: [nit] This will always warn.

mslipper (commented):
LGTM once the cache purge comment is addressed.

proxyd/cache.go
Inphi: Cool idea. One caveat though is multiple proxyd instances writing to the same cache will may conflict whenever a new generation is warranted (i.e. when consensus is broken and they all clear the cache at the same time).
One solution is to have a "current generation key" that can be atomically incremented via `INCR` to coordinate writes across proxyd instances.

felipe-op: I'll leave the actual cache invalidation logic out of this PR and merge only the event emitter for now.

felipe-op:
I removed the cache invalidation logic from this PR. This is now only the event dispatcher.
Cache will be handled in future change.

mslipper (approved):

PR #5546, opened 2023-04-27
op-program: Build op-program client binary

ajsutton:
**Description**

Adds an entry point specifically for the fault proof program and updates the makefile to build the host and client. Also always builds a MIPs version of the client binary suitable for execution inside cannon.

Note the change in `ssz.go` to "fix" 32bit compatibility.  It may be better to just error when `ExtraData` is longer than 32 bytes given it's defined as a List with max length 32 bytes but there's a test specifically checking we can marshal a value longer than that.

**Metadata**

- https://linear.app/optimism/issue/CLI-3878/fpp-create-client-executable

### DIFF
--- a/op-node/eth/ssz.go
+++ b/op-node/eth/ssz.go
@@ -62,7 +62,9 @@ func unmarshalBytes32LE(in []byte, z *Uint256Quantity) {
 
 // MarshalSSZ encodes the ExecutionPayload as SSZ type
 func (payload *ExecutionPayload) MarshalSSZ(w io.Writer) (n int, err error) {
-	if len(payload.ExtraData) > math.MaxUint32-executionPayloadFixedPart {
+	// Cast to uint32 to enable 32-bit MIPS support where math.MaxUint32-executionPayloadFixedPart is too big for int
+	// In that case, len(payload.ExtraData) can't be longer than an int so this is always false anyway.
+	if uint32(len(payload.ExtraData)) > math.MaxUint32-uint32(executionPayloadFixedPart) {
 		return 0, ErrExtraDataTooLarge
 	}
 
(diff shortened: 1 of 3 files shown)


op-node/eth/ssz.go
```diff
@@ -62,7 +62,9 @@ func unmarshalBytes32LE(in []byte, z *Uint256Quantity) {
 
 // MarshalSSZ encodes the ExecutionPayload as SSZ type
 func (payload *ExecutionPayload) MarshalSSZ(w io.Writer) (n int, err error) {
-	if len(payload.ExtraData) > math.MaxUint32-executionPayloadFixedPart {
+	// Cast to uint32 to enable 32-bit MIPS support where math.MaxUint32-executionPayloadFixedPart is too big for int
+	// In that case, len(payload.ExtraData) would overflow an int, but the max length for valid ExtraData is 32 bytes
+	if uint32(len(payload.ExtraData)) > math.MaxUint32-uint32(executionPayloadFixedPart) {
```
ajsutton, line 67: Without this change MIPS compilation fails with:
```
../op-node/eth/ssz.go:65:30: math.MaxUint32 - executionPayloadFixedPart (untyped int constant 4294966787) overflows int
```

Inphi (commented):
Could we also update the dockerfile to support a mips target? It'll be nice to be able to confirm a mips client works before the VM is fully ready via qemu.

Inphi (approved):

ajsutton:
There's no mips version of the alpine base image we use so I think making the docker image support mips is going to take a lot more work than its worth unfortunately.

Inphi:
> There's no mips version of the alpine base image we use so I think making the docker image support mips is going to take a lot more work than its worth unfortunately.

Gotcha. This is fine then.

trianglesphere:
> > There's no mips version of the alpine base image we use so I think making the docker image support mips is going to take a lot more work than its worth unfortunately.
> 
> Gotcha. This is fine then.

Debian images might have a mips base, but that needs a larger conversation. https://hub.docker.com/layers/library/golang/latest/images/sha256-bcfd44284b9432724abd99de5b6899b988c24fba7aa427712690b52d4e228701?context=explore

ajsutton:
I think we'll be able to test it in the VM soon enough and that's the key target anyway which is slightly different to a docker mips anyway - especially the on chain version which is very minimal.

PR #5547, opened 2023-04-27
op-program: Specify client program to run as an executable

ajsutton:
**Description**

Require the user to specify the client program to run as an executable when using detached mode. Removes the env var approach to toggling between the host and client executables.

This makes the host program much more similar to the actual VM which executes an arbitrary program and means that it is executing the actual client executable rather than calling it programatically.

Downside is that the e2e test now needs to build the client program executable so it can execute it. This is done by executing `make` from the e2e test which isn't particularly pleasant. It does mean the test remains completely standalone without relying on the user compiling the executable before running it though.

**Tests**

Updated existing tests.

**Additional context**

Builds on #5546 

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3878/fpp-create-client-executable

### DIFF
--- a/op-e2e/system_fpp_test.go
+++ b/op-e2e/system_fpp_test.go
@@ -9,29 +9,15 @@ import (
 	"github.com/ethereum-optimism/optimism/op-node/client"
 	"github.com/ethereum-optimism/optimism/op-node/sources"
 	"github.com/ethereum-optimism/optimism/op-node/testlog"
-	oppcl "github.com/ethereum-optimism/optimism/op-program/client"
 	"github.com/ethereum-optimism/optimism/op-program/client/driver"
 	opp "github.com/ethereum-optimism/optimism/op-program/host"
 	oppconf "github.com/ethereum-optimism/optimism/op-program/host/config"
-	oplog "github.com/ethereum-optimism/optimism/op-service/log"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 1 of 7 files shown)


op-e2e/system_fpp_test.go
```diff
@@ -115,7 +103,19 @@ func testVerifyL2OutputRoot(t *testing.T, detached bool) {
 	fppConfig.L1URL = sys.NodeEndpoint("l1")
 	fppConfig.L2URL = sys.NodeEndpoint("sequencer")
 	fppConfig.DataDir = preimageDir
-	fppConfig.Detached = detached
+	if detached {
+		// When running in detached mode we need to compile the client executable since it will be called directly.
```
trianglesphere, line 107: I wonder eventually pull this out of `op-e2e` and have a CI pipeline which takes care of doing the building & then placing the result in the right place. Instead of a test binary that's run with `go test` it'd just be a binary that we'd have to run from CI.

The other idea I have here is building into a tmp directory & using that binary. This idea probably works better with the current setup.

trianglesphere (approved):
Changes LGTM. I've got a couple ideas around the binary, but nothing blocking.

op-e2e/system_fpp_test.go
ajsutton: Yeah I was very much in two minds about the best way to do this. Part of the challenge currently is that most of our CI tasks have all their commands in the circleCI config and don't go through a makefile where it would be easy to run locally the same way and setup the right dependencies, env vars etc that more complex tests might require.

But I do quite like the way this just automatically works even when I just run it in GoLand.  I will extract the build command from the test itself to a helper function now I think about it though - then the test is clearer about what it wants to achieve and the details are better encapsulated if we want to change how it does it in the future (or if there are other build targets we want to execute they can all live together etc).

PR #5549, opened 2023-04-27
op-program: Add server mode

ajsutton:
**Description**

Adds a server mode to the op-program host so that it inits the prefetchers and key-value store but doesn't run the actual fault proof program. Instead it listens for preimage hints and requests via additional file handles using the same protocol as the fault proof client does.

This is designed so that the FPVM can remain completely agnostic to the program it runs, it simply forwards any preimage requests from the program to the FPP host in server mode.  In on-chain mode the FPVM would ignore all hints and retrieve pre-images from the oracle contracts - this just allows running the off-chain emulation in MIPS mode to create the full trace.

**Tests**

Added a test specifically for server mode and unit tests around the new config and CLI args.

**Additional context**

Builds on #5547 

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3877/fpp-add-server-mode-to-host

### DIFF
--- a/op-program/host/flags/flags.go
+++ b/op-program/host/flags/flags.go
@@ -86,6 +86,11 @@ var (
 		Usage:  "Run the specified client program as a separate process detached from the host. Default is to run the client program in the host process.",
 		EnvVar: service.PrefixEnvVar(envVarPrefix, "EXEC"),
 	}
+	Server = cli.BoolFlag{
+		Name:   "server",
+		Usage:  "Run in pre-image server mode without executing any client program.",
+		EnvVar: service.PrefixEnvVar(envVarPrefix, "SERVER"),
+	}
 )
 
 // Flags contains the list of configuration options available to the binary.
@@ -107,6 +112,7 @@ var programFlags = []cli.Flag{
 	L1TrustRPC,
 	L1RPCProviderKind,
 	Exec,
+	Server,
 }
 
 func init() {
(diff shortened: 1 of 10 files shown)


op-program/host/flags/flags.go
```diff
@@ -107,6 +112,7 @@ var programFlags = []cli.Flag{
 	L1TrustRPC,
 	L1RPCProviderKind,
 	Exec,
+	Server,
```
Inphi, line 115: Since `exec` and `server` are mutually exclusive, perhaps we can simplify configs by having both under a `--mode` option.

ajsutton: hmm, maybe but how do we specify which client executable to run in that case? I guess you could do `--exec <client.exe>` or `--exec server` but it feels a bit awkward and means you can't run a program called "server" as the client.

Inphi: You're right. I assumed `exec` is a boolean flag.

op-program/host/host.go
```diff
@@ -85,8 +66,14 @@ func FaultProofProgram(logger log.Logger, cfg *config.Config) error {
...
 	defer hHostRW.Close()
-	hHost := preimage.NewHintReader(hHostRW)
-	routeHints(logger, hHost, hinter)
+
+	go func() {
+		err := PreimageServer(ctx, logger, cfg, pHostRW, hHostRW)
+		if err != nil {
+			logger.Error("preimage server failed", "err", err)
+		}
+		logger.Debug("Preimage server stopped")
```
Inphi, line 75: move L68 (`defer hHostRW.Close()`)  here so that we close the socket before blocking on child wait. Doing this should also fully resolve [CLI-3868](https://linear.app/optimism/issue/CLI-3868/exit-when-pre-image-data-missing).

ajsutton: Done. I was wondering about that but got tied up in various questions about what the right approach to ensuring things closed and when etc so had left it.

Inphi (approved):

PR #5553, opened 2023-04-27
feat(contracts-bedrock): Gnosis Safe Tx Hash Validation

refcell:
**Description**

This introduces a `SafeBuilder` function to compute the gnosis safe tx hash for given safe and proxy admin addresses. It is wrapped by the downstream child `PostSherlockL1.s.sol` and `PostSherlockL2.s.sol` scripts.

Can be run using `forge script --sig` as shown below.

_NOTE: Alchemy keys need to be supplied to the following commands._

```bash
forge script \
    --rpc-url https://opt-goerli.g.alchemy.com/v2/<API_KEY> \
    --sig "function computeSafeTransactionHash() public returns (bytes32)" \
    scripts/upgrades/PostSherlockL2.s.sol
```

```bash
forge script \
    --rpc-url https://eth-goerli.g.alchemy.com/v2/<API_KEY> \
    --sig "function computeSafeTransactionHash() public returns (bytes32)" \
    scripts/upgrades/PostSherlockL1.s.sol
```

**Metadata**

- Fixes CLI-3769


packages/contracts-bedrock/scripts/upgrades/SafeBuilder.sol
```diff
@@ -57,6 +57,37 @@ abstract contract SafeBuilder is Script {
         return success;
     }
 
+    /**
```
tynes, line 60: Does it make sense to reuse this function as part of the signing codepath?

packages/contracts-bedrock/scripts/upgrades/PostSherlockL2.s.sol
```diff
@@ -299,4 +299,13 @@ contract PostSherlockL2 is SafeBuilder {
         require(set.BaseFeeVault != address(0), "no proxies for this network");
         return set;
     }
+
+    /**
```
tynes, line 303: Ideally we can figure out long term how to not hardcode addresses so these same scripts can work across networks, or we have it so that all networks can use the same impls

packages/contracts-bedrock/scripts/upgrades/SafeBuilder.sol
```diff
@@ -57,6 +57,37 @@ abstract contract SafeBuilder is Script {
...
+     * @notice Computes the safe transaction hash.
+     */
+    function _computeSafeTransactionHash(address _safe, address _proxyAdmin) public returns (bytes32) {
+        // Ensure that the required contracts exist
+        require(address(multicall).code.length > 0, "multicall3 not deployed");
+        require(_safe.code.length > 0, "no code at safe address");
+        require(_proxyAdmin.code.length > 0, "no code at proxy admin address");
+
+        IGnosisSafe safe = IGnosisSafe(payable(_safe));
+        uint256 nonce = safe.nonce();
```
tynes, line 70: This is good but note that there is a race condition here, ie the wrong nonce will be grabbed if another tx frontruns this one. Perhaps a way around that would be to use ffi to `eth_getLogs` and figure out what nonce to use, altho might be annoying

tynes: Might require brute forcing and checking to see with `eth_call` to see which hash is in the mapping then use that nonce

packages/contracts-bedrock/scripts/upgrades/PostSherlockL2.s.sol
refcell: Will add a gardening ticket assigned to myself to design doc multi-network deployment configs. I know there are a variety out there, just need to find something reliable + extensible.

refcell: Created [CLI-3901](https://linear.app/optimism/issue/CLI-3901/featcontracts-design-multi-network-deployment-configs)

packages/contracts-bedrock/scripts/upgrades/SafeBuilder.sol
refcell: Discussion was had offline; closing since we settled on not over-engineering against nonce race conditions for now.

packages/contracts-bedrock/scripts/upgrades/SafeBuilder.sol
refcell: For this, are you suggesting we could use this `computeSafeTransactionHash` as a refactored function inside `_run`?

packages/contracts-bedrock/scripts/upgrades/PostSherlockL2.s.sol
```diff
@@ -126,6 +126,21 @@ contract PostSherlockL2 is SafeBuilder {
         require(PROXY_ADMIN.getProxyImplementation(prox.OptimismMintableERC721Factory).codehash == impl.OptimismMintableERC721Factory.codehash);
     }
 
+    /**
+     * @notice The address of the proxy admin.
+     */
+    function getProxyAdmin() public pure override returns (address) {
+        return 0x4200000000000000000000000000000000000018;
```
tynes, line 133: Hmm the proxy admin and safe were originally the 2 args to the `run` function. Ideally we do not couple these scripts to our network more and we can move towards having scripts that are agnostic to the network. To do this, we should be moving towards less hardcoding and more reading from a config file

refcell: LMK if I should introduce a config file in this pr.

clabby (approved):

tynes:
I'd like to see what this looks like rebased on #5555

packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
```diff
@@ -110,29 +110,40 @@ contract PostSherlockL1 is SafeBuilder {
         require(PROXY_ADMIN.getProxyImplementation(prox.L1ERC721Bridge).codehash == impl.L1ERC721Bridge.codehash, "L1ERC721Bridge codehash");
     }
 
+    /**
+     * @notice Computes the safe transaction hash.
+     */
+    function computeSafeTransactionHash(address _safe, address _proxyAdmin) public override returns (bytes32) {
```
tynes, line 116: This should have no args?

PR #5555, opened 2023-04-28
feat(contracts-bedrock): L2OutputOracle Scripts

refcell:
**Description**

This PR introduces scripts for working with the `L2OutputOracle` including a script that deletes an output for the below ticketed rehearsal runbook.

**Metadata**

- Fixes CLI-3729


packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,225 @@
...
+    function CHALLENGER() external view returns (address);
+    function getL2Output(uint256 _l2OutputIndex) external view returns (Types.OutputProposal memory);
+}
+
+/**
+ * @title DeleteOutput
+ * @notice Deletes an output root from the L2OutputOracle.
+ * @notice Example usage is provided in the README documentation.
+ */
+contract DeleteOutput is Script {
```
tynes, line 23: If this inherits from `SafeBuilder` then we don't need to copy/paste the `run` logic

packages/contracts-bedrock/scripts/interfaces/IL2OutputOracle.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Types } from "../../contracts/libraries/Types.sol";
+
```
tynes, line 5: This will end up out of date - any reason to just not import the actual `L2OutputOracle` instead?

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,314 @@
...
+    function buildCalldata(address _l2oo, uint256 index) internal pure returns (bytes memory) {
+        IMulticall3.Call3[] memory calls = new IMulticall3.Call3[](1);
+
+        IL2OutputOracle l2oo = IL2OutputOracle(_l2oo);
+
+        calls[0] = IMulticall3.Call3({
+            target: _l2oo,
+            allowFailure: false,
+            callData: abi.encodeCall(
+                l2oo.deleteL2Outputs,
```
tynes, line 109: Can't we do this on the type instead of creating a variable?

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,314 @@
...
+import { Types } from "../../contracts/libraries/Types.sol";
+import { FeeVault } from "../../contracts/universal/FeeVault.sol";
+import { Predeploys } from "../../contracts/libraries/Predeploys.sol";
+
+/**
+ * @title DeleteOutput
+ * @notice Deletes an output root from the L2OutputOracle.
+ * @notice Example usage is provided in the README documentation.
+ */
+contract DeleteOutput is EnhancedScript, GlobalConstants {
```
tynes, line 22: Hmm this could just inherit `SafeBuilder` right?

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,314 @@
...
+            ProxyAdmin: 0x01d3670863c3F4b24D7b107900f0b75d4BbC6e0d,
+            L2OutputOracleProxy: 0xE6Dfba0953616Bacab0c9A8ecb3a9BBa77FC15c0
+        });
+    }
+
+    /**
+     * @notice Follow up assertions to ensure that the script ran to completion.
+     */
+    function _postCheck(uint256 index) internal view {
+        IL2OutputOracle l2oo = IL2OutputOracle(contracts[GOERLI].L2OutputOracleProxy);
```
tynes, line 68: We should use `block.chainid` to make this portable instead of hardcoding `GOERLI`

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
tynes: The idea is that scripts don't need to implement `run`, they just implement building the calldata and post checks

tynes: Minimal implementation required:

```solidity
pragma solidity 0.8.15;

import { SafeBuilder } from "./SafeBuilder.sol";

contract BedrockUpgrade is SafeBuilder {
    function setUp() external {
        //
    }

    function buildCalldata(address _proxyAdmin) internal override view returns (bytes memory) {
        //
    }

    function _postCheck() internal override {
        //
    }

    function test_script_succeeds() skipWhenNotForking external {
        //
    }
}
```

packages/contracts-bedrock/scripts/interfaces/IL2OutputOracle.sol
refcell: Yea agreed, easier not to maintain.

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,314 @@
...
+        address _l2oo = contracts[GOERLI].L2OutputOracleProxy;
+
+        return _getTransactionHash(_safe, _l2oo, index);
+    }
+
+    /**
+     * @notice Executes the gnosis safe transaction to delete an L2 Output Root.
+     */
+    function run(uint256 _index) external returns (bool) {
+        address _safe = contracts[GOERLI].Safe;
```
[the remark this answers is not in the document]
refcell: > hh deploy artifacts
> ;-;

But yes agreed, for now should we keep it hardcoded so we can just avoid footgunning ourselves and then post-migration improve scripting? If so let's create a gardening linear ticket to track?

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
refcell: Ah yes forgot to refactor to use the `SafeBuilder`, my b. Will do that in this round of fixes b4 re-requesting a review.

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
refcell: Unless we decide on a good solution here i pretty strongly lean on the side of hardcoding so we:
a) don't footgun ourselves in the short-term
b) more easily track in a gardening ticket to make extensible

If you think otherwise, will change it, lmk

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,208 @@
...
+            ProxyAdmin: 0x01d3670863c3F4b24D7b107900f0b75d4BbC6e0d,
+            L2OutputOracleProxy: 0xE6Dfba0953616Bacab0c9A8ecb3a9BBa77FC15c0
+        });
+    }
+
+    /**
+     * @notice Executes the gnosis safe transaction to delete an L2 Output Root.
+     */
+    function run(uint256 _index) external returns (bool) {
+        address _safe = contracts[GOERLI].Safe;
```
tynes, line 61: I think you should make `contracts` internal and rename to `_contracts` and then make a getter `contracts()` that uses `block.chainid` to look in the mapping and revert if the values in the struct are undefined then use `contracts()` getter everywhere you want to reference the contracts

packages/contracts-bedrock/scripts/outputs/README.md
```diff
@@ -0,0 +1,37 @@
...
+These are outlined below.
+
+### Retrieving an L2 Block Number
+
+The output's associated L2 block number can be retrieved using the following command, where
+`<L2_OUTPUT_INDEX>` is the index of the output in the [L2OutputOracle](../../contracts/L1/L2OutputOracle.sol).
+
+```bash
+$ forge script scripts/output/DeleteOutput.s.sol \
+  --sig "function getL2BlockNumber(uint256) public view returns (uint256)" \
```
tynes, line 32: you can omit `function ... public view returns ...` and just use `getL2BlockNumber(uint256)` right?

packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
```diff
@@ -110,6 +110,16 @@ contract PostSherlockL1 is SafeBuilder {
         require(PROXY_ADMIN.getProxyImplementation(prox.L1ERC721Bridge).codehash == impl.L1ERC721Bridge.codehash, "L1ERC721Bridge codehash");
     }
 
+    /**
```
tynes, line 113: I prefer to leave this `run` in the `SafeBuilder` for now and instead wrap it with overloaded `run` if we need more args. Then in the future we can rearchitect a bit more easily

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,208 @@
...
+        });
+
+        return abi.encodeCall(IMulticall3.aggregate3, (calls));
+    }
+
+    /**
+     * @notice Internal helper function to compute the safe transaction hash.
+     */
+    function _getTransactionHash(address _safe, address _proxyAdmin, address _l2oo, uint256 _index) internal returns (bytes32) {
+        // Ensure that the required contracts exist
```
tynes, line 130: This code should be able to live in the safe builder? `buildCalldata` should read `_l2oo` and `_index` from storage

packages/contracts-bedrock/scripts/outputs/README.md
refcell: Yessir, was a copy-pasta.

packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
refcell: Note this will require then to run the `DeleteOutput` script with `--sig` _or_ to hardcode the index to delete in the script. No way to get around this unless the `uint256 index` is an argument in `run()`.

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
refcell: Good call!

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * @notice Returns the ContractSet for the defined block chainid.
+     *
+     * @dev Reverts if no ContractSet is defined.
+     */
+    function contracts() public view returns (ContractSet memory) {
+        ContractSet memory cs = _contracts[block.chainid];
+        if (cs.Safe == address(0) && cs.ProxyAdmin == address(0) && cs.L2OutputOracleProxy == address(0)) {
```
clabby, line 64: None of these contracts should be at `address(0)` realistically, maybe swap to this?
```suggestion
        if (cs.Safe == address(0) || cs.ProxyAdmin == address(0) || cs.L2OutputOracleProxy == address(0)) {
```

clabby (approved):

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
tynes, line 64: I feel like we want this to be an or condition instead of an and condition, so if any of them are unset

packages/contracts-bedrock/scripts/outputs/DeleteOutput.s.sol
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * @notice Executes the gnosis safe transaction to delete an L2 Output Root.
+     */
+    function run(uint256 _index) external returns (bool) {
+        address _safe = contracts().Safe;
+        address _proxyAdmin = contracts().ProxyAdmin;
+        index = _index;
+        vm.startBroadcast();
```
tynes, line 77: we can dedupe the broadcast if we call `run` instead of `_run` here

tynes: as well as the postcheck

tynes (approved):

PR #5557, opened 2023-04-28
op-program: Script and instructions for running against Goerli

ajsutton:
**Description**

Adds more detailed instructions on how to run the fault proof program along with a script to gather suitable inputs and verify a part of the op-goerli chain.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3896/write-instructions-to-run-fpp-on-goerli

### DIFF
--- a/op-program/.gitignore
+++ b/op-program/.gitignore
@@ -1,1 +1,2 @@
 bin
+/oracledata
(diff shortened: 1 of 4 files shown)


ajsutton:
@qbzzt I could use your help on this one when you get a chance (it's not blocking anything so not a rush).
1. Where's the best place to put this kind of content? It will eventually get replaced by proper documentation on how the dispute game works and this fault proof program just fits into that, but it's useful to show some progress now
2. It could use some beating into shape by someone with better writing skills than me. :)

I'm happy to provide more background or whatever other info is useful as well - hopefully this is at least a reasonable starting point.

PR #5560, opened 2023-04-28
contracts-bedrock: deploy time hacky fix

tynes:
**Description**

Without this fix, things break at deploy time. This is not the ideal fix. Should refactor a bit with a real fix.
Opening this as a draft, will push a better fix


packages/contracts-bedrock/scripts/upgrades/PostSherlock.s.sol
```diff
@@ -61,6 +61,9 @@ contract PostSherlockL1 is SafeBuilder {
      * @notice Place the contract addresses in storage so they can be used when building calldata.
      */
     function setUp() external {
+        // HACK: needs refactor
```
roninjin10, line 64: definitely needs more comment than this if this ships imo

PR #5573, opened 2023-05-02
op-e2e: Add FPP test for empty blocks

Inphi:
**Description**

Add e2e test asserting that the fault proof program handles missing batches and empty blocks.

**Metadata**

- Fixes CLI-3886


op-e2e/system_fpp_test.go
```diff
@@ -27,6 +27,114 @@ func TestVerifyL2OutputRootDetached(t *testing.T) {
...
+	require.Nil(t, err, "Waiting for block on verifier")
+	require.Equal(t, receipt.BlockHash, nonEmptyBlock.Hash(), "send tx should be published")
+
+	t.Log("Capture current L2 head as agreed starting point")
+	l2AgreedBlock, err := l2Seq.BlockByNumber(ctx, nil)
+	require.NoError(t, err, "could not retrieve l2 agreed block")
+	l2Head := l2AgreedBlock.Hash()
+
+	t.Log("=====Stopping batch submitter=====")
+	err = sys.BatchSubmitter.Stop(ctx)
```
ajsutton, line 87: Feels like we should wait for the safe head to update to our new block before stopping the batch submitter. Otherwise we might stop the batcher before it has published the tx we sent.

op-e2e/system_fpp_test.go
```diff
@@ -27,6 +27,114 @@ func TestVerifyL2OutputRootDetached(t *testing.T) {
...
+
+	t.Log("Sending transactions to setup existing state, prior to challenged period")
+	aliceKey := cfg.Secrets.Alice
+	receipt := SendL2Tx(t, cfg, l2Seq, aliceKey, func(opts *TxOpts) {
+		opts.ToAddr = &cfg.Secrets.Addresses().Bob
+		opts.Value = big.NewInt(1_000)
+	})
+	nonEmptyBlock, err := waitForBlock(receipt.BlockNumber, l2Verif, time.Duration(2*cfg.DeployConfig.L1BlockTime)*time.Second)
+	require.Nil(t, err, "Waiting for block on verifier")
+	require.Equal(t, receipt.BlockHash, nonEmptyBlock.Hash(), "send tx should be published")
```
ajsutton, line 79: This seems to be assuming that the verifier can only receive blocks via L1 but wouldn't it also be getting blocks via gossip from the sequencer?

op-e2e/system_fpp_test.go
```diff
@@ -27,6 +27,114 @@ func TestVerifyL2OutputRootDetached(t *testing.T) {
...
+
+	t.Log("Capture current L2 head as agreed starting point")
+	l2AgreedBlock, err := l2Seq.BlockByNumber(ctx, nil)
+	require.NoError(t, err, "could not retrieve l2 agreed block")
+	l2Head := l2AgreedBlock.Hash()
```
ajsutton, line 84: As this is using the unsafe head, it may wind up selecting a block that's after the safe head and winds up being re-orged back out when the sequence window is reached.  We can probably just use `receipt.BlockHash` as our agreed starting point in this case.

op-e2e/system_fpp_test.go
```diff
@@ -27,6 +27,114 @@ func TestVerifyL2OutputRootDetached(t *testing.T) {
...
+// induced by missing batches.
+// Setup is as follows:
+// - create initial conditions and agreed l2 state
+// - stop the batch submitter to induce empty blocks
+// - wait for the seq window to expire so we can observe empty blocks
+// - select an empty block as our claim
+// - reboot the batch submitter
+// - update the state root via a tx
+// - run program
+func testVerifyL2OutputRootEmptyBlock(t *testing.T, detached bool) {
```
ajsutton, line 48: Nothing is using the `detached` bool.

ajsutton (commented):
I had to play with the code directly to think through all the details here and wound up with the code below.  The main change is that I've dropped the verifier node and monitored the safe head to know when empty blocks were inserted because the sequence window elapsed.

```go

// TestVerifyL2OutputRoot_EmptyBlock asserts that the program can verify the output root of an empty block
// induced by missing batches.
// Setup is as follows:
// - create initial conditions and agreed l2 state
// - stop the batch submitter to induce empty blocks
// - wait for the seq window to expire so we can observe empty blocks
// - select an empty block as our claim
// - reboot the batch submitter
// - update the state root via a tx
// - run program
func testVerifyL2OutputRootEmptyBlock(t *testing.T, detached bool) {
	InitParallel(t)
	ctx := context.Background()

	cfg := DefaultSystemConfig(t)
	// We don't need a verifier - just the sequencer is enough
	delete(cfg.Nodes, "verifier")
	// Use a small sequencer window size to avoid test timeout while waiting for empty blocks
	// But not too small to ensure that the updated state after claim selection is published
	cfg.DeployConfig.SequencerWindowSize = 16

	sys, err := cfg.Start()
	require.Nil(t, err, "Error starting up system")
	defer sys.Close()

	log := testlog.Logger(t, log.LvlInfo)
	log.Info("genesis", "l2", sys.RollupConfig.Genesis.L2, "l1", sys.RollupConfig.Genesis.L1, "l2_time", sys.RollupConfig.Genesis.L2Time)

	l1Client := sys.Clients["l1"]
	l2Seq := sys.Clients["sequencer"]
	rollupRPCClient, err := rpc.DialContext(context.Background(), sys.RollupNodes["sequencer"].HTTPEndpoint())
	require.Nil(t, err)
	rollupClient := sources.NewRollupClient(client.NewBaseRPCClient(rollupRPCClient))

	verifierRPCClient, err := rpc.DialContext(context.Background(), sys.RollupNodes["sequencer"].HTTPEndpoint())
	require.Nil(t, err)
	verifierClient := sources.NewRollupClient(client.NewBaseRPCClient(verifierRPCClient))

	t.Log("Sending transactions to setup existing state, prior to challenged period")
	aliceKey := cfg.Secrets.Alice
	receipt := SendL2Tx(t, cfg, l2Seq, aliceKey, func(opts *TxOpts) {
		opts.ToAddr = &cfg.Secrets.Addresses().Bob
		opts.Value = big.NewInt(1_000)
	})

	require.NoError(t, waitForSafeHead(ctx, receipt.BlockNumber.Uint64(), rollupClient))

	t.Log("Capture current L2 head as agreed starting point")
	l2Head := receipt.BlockHash

	t.Log("=====Stopping batch submitter=====")
	err = sys.BatchSubmitter.Stop(ctx)
	require.NoError(t, err, "could not stop batch submitter")

	// Wait for the sequencer to catch up with the current L1 head so we know all submitted batches are processed
	t.Log("Wait for sequencer to catch up with last submitted batch")
	l1HeadNum, err := l1Client.BlockNumber(ctx)
	require.NoError(t, err)
	_, err = waitForL1OriginOnL2(l1HeadNum, l2Seq, 30*time.Second)
	require.NoError(t, err)

	// Get the current safe head now that the batcher is stopped
	safeBlock, err := l2Seq.BlockByNumber(ctx, big.NewInt(int64(rpc.SafeBlockNumber)))
	require.NoError(t, err)

	// Wait for safe head to start advancing again when the sequencing window elapses, for at least three blocks
	t.Log("Wait for safe head to advance after sequencing window elapses")
	require.NoError(t, waitForSafeHead(ctx, safeBlock.NumberU64()+3, rollupClient))

	// Use the 2nd empty block as our L2 claim block
	t.Log("Determine L2 claim")
	l2ClaimBlock, err := l2Seq.BlockByNumber(ctx, big.NewInt(int64(safeBlock.NumberU64()+2)))
	require.NoError(t, err, "get L2 claim block number")
	l2ClaimBlockNumber := l2ClaimBlock.Number().Uint64()
	l2Output, err := verifierClient.OutputAtBlock(ctx, l2ClaimBlockNumber)
	require.NoError(t, err, "could not get expected output")
	l2Claim := l2Output.OutputRoot

	t.Log("=====Restarting batch submitter=====")
	err = sys.BatchSubmitter.Start()
	require.NoError(t, err, "could not start batch submitter")

	t.Log("Add a transaction to the next batch after sequence of empty blocks")
	receipt = SendL2Tx(t, cfg, l2Seq, aliceKey, func(opts *TxOpts) {
		opts.ToAddr = &cfg.Secrets.Addresses().Bob
		opts.Value = big.NewInt(1_000)
		opts.Nonce = 1
	})
	require.NoError(t, waitForSafeHead(ctx, receipt.BlockNumber.Uint64(), rollupClient))

	t.Log("Determine L1 head that includes batch after sequence of empty blocks")
	l1HeadBlock, err := l1Client.BlockByNumber(ctx, nil)
	require.NoError(t, err, "get l1 head block")
	l1Head := l1HeadBlock.Hash()

	testFaultProofProgramScenario(t, ctx, sys, &FaultProofProgramTestScenario{
		L1Head:             l1Head,
		L2Head:             l2Head,
		L2Claim:            common.Hash(l2Claim),
		L2ClaimBlockNumber: l2ClaimBlockNumber,
	})
}
```

I had to increase the timeout in the `waitForSafeHead` as well because it takes more than 30 seconds for the sequence window to elapse. Given the timeout shouldn't be reached just setting a fixed wait period that's well after what we expect to wait is fine IMO.

op-e2e/system_fpp_test.go
Inphi: gossip shouldn't happen directly as the sequencer and verifier are disconnected by default - https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/setup.go#L204. 
And afaict, there isn't a bootnode for autodiscovery.

ajsutton: ah, that's not at all obvious but good to know. :)

ajsutton:
btw the code I posted is just a "this may be helpful" not a "this is definitely the right way to do it" thing.

op-e2e/system_fpp_test.go
Inphi: Good catch. Will fix.

Inphi:
For some reason, waiting for a given safe block number returns before that the block is fully imported by the verifier. Not sure what's going on there, but I've replaced it to poll for the block number directly instead. As long as the verifier is not peered with the sequencer, all blocks retrieved from it are safe.

Inphi:
> I had to play with the code directly to think through all the details here and wound up with the code below. The main change is that I've dropped the verifier node and monitored the safe head to know when empty blocks were inserted because the sequence window elapsed.

This actually works better. Thanks for code suggestion. As for the 30s timeout, it's unavoidable due to seq window waits. Not terribly ideal for local execution. But CI uses a larger timeout so it's fine.

op-e2e/system_fpp_test.go
```diff
@@ -96,19 +209,38 @@ func testVerifyL2OutputRoot(t *testing.T, detached bool) {
 	require.NoError(t, err, "get l1 head block")
 	l1Head := l1HeadBlock.Hash()
 
+	testFaultProofProgramScenario(t, ctx, sys, &FaultProofProgramTestScenario{
+		L1Head:             l1Head,
+		L2Head:             l2Head,
+		L2Claim:            common.Hash(l2Claim),
+		L2ClaimBlockNumber: l2ClaimBlockNumber,
```
ajsutton, line 216: I think we should be passed `detached` through here too.

ajsutton (approved):
LGTM, just need to pass through detached for the existing test as well I think.

op-e2e/system_fpp_test.go
Inphi: whoops. thanks.

op-e2e/system_fpp_test.go
```diff
@@ -27,6 +27,119 @@ func TestVerifyL2OutputRootDetached(t *testing.T) {
 	testVerifyL2OutputRoot(t, true)
 }
 
+func TestVerifyL2OutputRootEmptyBlock(t *testing.T) {
+	testVerifyL2OutputRootEmptyBlock(t, false)
+}
+
+func TestVerifyL2OutputRootEmptyBlockDetached(t *testing.T) {
+	testVerifyL2OutputRootEmptyBlock(t, true)
```
trianglesphere, line 35: This flaked: https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/17578/workflows/48c306ad-b1af-4334-8bbc-a88a4eeeb2a0/jobs/676022

PR #5574, opened 2023-05-02
op-program: Ensure go routines all complete cleanly

ajsutton:
**Description**

Tidy up the handling of goroutines to ensure any created go routines are exited prior to the two main entry point functions (`FaultProofProgram` and `PreimageServer`) returning.

This avoids test failures caused by dangling go routines like:  https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/17480/workflows/9981daa8-1817-4189-8cdb-239bc65f86ab/jobs/670118

Winds up being more difficult than I'd really like because we need to close file channels to signal that the preimage server, hinter and client program should exit, but also ensure we only call `Close()` once because the behaviour is officially undefined when calling it multiple times (and we do use multiple implementations in different modes so can't rely on the fact that File will just return an error that we ignore).

Very open to any suggestions for simplifying.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3910/fix-dangling-go-routines

### DIFF
--- a/op-program/host/host.go
+++ b/op-program/host/host.go
@@ -58,17 +58,28 @@ func FaultProofProgram(ctx context.Context, logger log.Logger, cfg *config.Confi
 	if err != nil {
 		return fmt.Errorf("failed to create preimage pipe: %w", err)
 	}
-	defer pHostRW.Close()
+	pClientCloser := oppio.NewSafeClose(pClientRW)
+	defer pClientCloser.Close()
 
 	// Setup client I/O for hint comms
 	hClientRW, hHostRW, err := oppio.CreateBidirectionalChannel()
 	if err != nil {
 		return fmt.Errorf("failed to create hints pipe: %w", err)
 	}
+	hClientCloser := oppio.NewSafeClose(hClientRW)
+	defer hClientCloser.Close()
 
+	// Use a channel to receive the server result so we can wait for it to complete before returning
+	serverErr := make(chan error)
 	go func() {
-		defer hHostRW.Close()
-		err := PreimageServer(ctx, logger, cfg, pHostRW, hHostRW)
+		serverErr <- PreimageServer(ctx, logger, cfg, pHostRW, hHostRW)
+	}()
+	defer func() {
+		// Ensure the client streams are closed to trigger the server to exit
+		pClientCloser.Close()
+		hClientCloser.Close()
+		logger.Debug("Waiting for preimage server to exit")
+		err := <-serverErr
 		if err != nil {
 			logger.Error("preimage server failed", "err", err)
 		}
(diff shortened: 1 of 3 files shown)


op-program/host/host.go
```diff
@@ -142,8 +164,14 @@ func PreimageServer(ctx context.Context, logger log.Logger, cfg *config.Config,
 	hinterDone := routeHints(logger, hintChannel, hinter)
 	select {
 	case err := <-serverDone:
+		// Close the channels to trigger the hinter to exit and wait for it to complete
+		closeChannels()
+		<-hinterDone
```
Inphi, line 169: rather than close and wait here. Could we use a defer statement that waits on the hinter/server channels? It's well-defined to receive multiple times from a closed channel.

ajsutton: The problem I hit was that `defer` winds up running in the opposite order to what they're called.  So we can't just add another `defer` here because it would wind up executing before the file handles are closed and so we'd deadlock.  But  if we created the `hinterDone` channel here and passed it into `routeHints` so we could setup that `defer` before the file closing defers, it would block forever if we hadn't reached the point where we actually called `routeHints` because  it wouldn't yet be closed.

I kind of hope I'm missing something about how to make this work elegantly though because it's really annoying to not just be able to use `defer` here.

ajsutton: One option would be to do:

```
func PreimageServer(ctx context.Context, logger log.Logger, cfg *config.Config, preimageChannel oppio.FileChannel, hintChannel oppio.FileChannel) error {
	var serverDone chan error
	var hinterDone chan error
	defer func() {
		preimageChannel.Close()
		hintChannel.Close()
		if serverDone != nil {
			// Wait for pre-image server to complete
			<-serverDone
		}
		if hinterDone != nil {
			// Wait for hinter to complete
			<-hinterDone
		}
	}()
	logger.Info("Starting preimage server")
```

and then we wouldn't need to manually close and wait like this.  Same pattern works in `FaultProofProgram` as well and I think it's better.  I've pushed up an additional commit with the new approach - let me know what you think.

Inphi (approved):
Honestly this isn't much better than the previous revision. But I don't have a better suggestion.
The current patch is fine though. I think the added comments makes it clear what's going on.

op-program/host/host.go
Inphi: It's still complex/
I tried to come up with a solution that hides complexity in `launchOracleServer` and `routeHints` functions but this cross-synchronization makes it even more complicated.
In the future, I would like to redo I/O to use polling instead of blocking so we can easily handle cancellations.

Inphi: Resolving. This is fine.

ajsutton: Yeah agreed, I'm not fan particularly.  Not sure how polling will interact with the VM but worth a look I suspect.

PR #5576, opened 2023-05-02
op-program: Setup CI job to periodically verify goerli outputs

ajsutton:
**Description**

Schedules a job in CI ever 4 hours to verify a segment of the Goerli chain with the fault proof program.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3891/fpp-create-goerli-verification-ci-job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,7 @@ version: 2.1
 orbs:
   go: circleci/go@1.5.0
   gcp-cli: circleci/gcp-cli@3.0.1
+  slack: circleci/slack@4.10.1
 commands:
   gcp-oidc-authenticate:
     description: "Authenticate with GCP using a CircleCI OIDC token."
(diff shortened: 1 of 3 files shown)


ajsutton:
Currently the circleci config sets this up to use a free Alchemy plan as the L1 which is slower but works fine (and a good test of the retry logic). I think its worth getting this merged now though and we can just update the config when our L1 node is sync'd.

.circleci/config.yml
```diff
@@ -1105,6 +1106,21 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: verify-goerli
+          command: |
+            make verify-goerli
+          working_directory: op-program
+      - slack/notify:
+          channel: C03N11M0BBN
+          event: fail
```
Inphi, line 1121: Is there a way to know if the job ran at all? Since it runs pretty infrequently, maybe it's fine to get notified all the time.

Inphi (approved):
nice!

.circleci/config.yml
ajsutton: We can check in the CircleCI dashboard. We used to have the hive tests posting to slack on every run and it just created a lot of noise that hid the failures.

Inphi: I see. It's better this way then.

Inphi:
You may need to rebase on the latest develop to fix the link lint

PR #5577, opened 2023-05-02
feat(ctb): Use named vars for reused config addresses

maurelian:
**Description**

Deduplicates repeated strings.

### DIFF
--- a/packages/contracts-bedrock/deploy-config/mainnet.ts
+++ b/packages/contracts-bedrock/deploy-config/mainnet.ts
@@ -3,11 +3,18 @@ import { DeployConfig } from '../src/deploy-config'
 // NOTE: The 'mainnet' network is currently being used for bedrock migration rehearsals.
 // The system configured below is not yet live on mainnet, and many of the addresses used are
 // unsafe for a production system.
+
+// The following addresses are assigned to multiples roles in the system, therfore we save them
+// as constants to avoid having to change them in multiple places.
+const foundationMultisig = '0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266' // hh test signer 0
+const feeRecipient = '0x70997970C51812dc3A010C7d01b50e0d17dc79C8' // hh test signer 1
+const mintManager = '0x5C4e7Ba1E219E47948e6e3F55019A647bA501005'
+
 const config: DeployConfig = {
-  finalSystemOwner: '0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A',
-  controller: '0x78339d822c23d943e4a2d4c3dd5408f66e6d662d',
-  portalGuardian: '0x78339d822c23d943e4a2d4c3dd5408f66e6d662d',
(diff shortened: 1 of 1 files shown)


packages/contracts-bedrock/deploy-config/mainnet.ts
```diff
@@ -3,11 +3,15 @@ import { DeployConfig } from '../src/deploy-config'
...
 // The system configured below is not yet live on mainnet, and many of the addresses used are
 // unsafe for a production system.
+
+const foundationMultisig = 'f39fd6e51aad88f6f4ce6ab8827279cfffb92266'
+const feeRecipient = '0x90F79bf6EB2c4f870365E785982E1f101E93b906'
+
 const config: DeployConfig = {
-  finalSystemOwner: '0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A',
-  controller: '0x78339d822c23d943e4a2d4c3dd5408f66e6d662d',
-  portalGuardian: '0x78339d822c23d943e4a2d4c3dd5408f66e6d662d',
```
tynes, line 9: Did you mean to alter the values here?

maurelian: Yes, `0x783` was a local address I used. `0x9ba` is the real foundation multisig which prevented us from unpausing the portal before. These addresses should all be the same, even when we go to mainnet, but for rehearsal they will just be the hh test addresses.

packages/contracts-bedrock/deploy-config/mainnet.ts
```diff
@@ -26,16 +32,16 @@ const config: DeployConfig = {
...
-  l1FeeVaultRecipient: '0x90F79bf6EB2c4f870365E785982E1f101E93b906',
-  sequencerFeeVaultRecipient: '0x90F79bf6EB2c4f870365E785982E1f101E93b906',
+  baseFeeVaultRecipient: feeRecipient,
+  l1FeeVaultRecipient: feeRecipient,
+  sequencerFeeVaultRecipient: feeRecipient,
 
   governanceTokenName: 'Optimism',
   governanceTokenSymbol: 'OP',
-  governanceTokenOwner: '0x90F79bf6EB2c4f870365E785982E1f101E93b906',
+  governanceTokenOwner: foundationMultisig,
```
tynes, line 44: The governance token owner isn't the foundation multisig, it should be the mint manager. The mint manager is already on L2 and would prefer to use the actual value here as to prevent us from accidentally goofing

tynes: I see that it was never correct

tynes (approved):

tynes:
Preference to set the owner of the token to [0x5C4e7Ba1E219E47948e6e3F55019A647bA501005](https://optimistic.etherscan.io/address/0x5C4e7Ba1E219E47948e6e3F55019A647bA501005). It is what the actual value should be so that we don't need to reconsider it in the future. Ideal if its with a descriptive variable name like the others :)

PR #5581, opened 2023-05-02
contracts-bedrock: finalize deploy config

tynes:
**Description**

Finalized deploy config


packages/contracts-bedrock/deploy-config/mainnet.json
maurelian, line 1: Will also need to update mainnet.ts

maurelian (approved):

PR #5585, opened 2023-05-02
Update docker files to use golang:1.19.9 and alpine3.16

ajsutton:
**Description**

Update docker files to use `golang:1.19.9-alpine3.16` instead of `golang:1.19.0-alpine3.15` as there's a bug in the MIPS compiler in 1.19.0 (https://github.com/golang/go/issues/54991) which causes it to fail with the latest op-geth changes.  There isn't an `alpine3.15` version with the updated golang compiler so updating from `alpine:3.15` to alpine:3.16` for the images as well to be consistent with what we compile on.

**Metadata**

- Required for https://linear.app/optimism/issue/CLI-3867/update-op-geth-to-upstream-1116-release and https://github.com/ethereum-optimism/optimism/pull/5514

**TODOs**

### DIFF
--- a/op-batcher/Dockerfile
+++ b/op-batcher/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM golang:1.19.0-alpine3.15 as builder
+FROM --platform=$BUILDPLATFORM golang:1.19.9-alpine3.16 as builder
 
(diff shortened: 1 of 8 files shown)


trianglesphere (approved):
LGTM. TBH it's slightly concerning that we're building with patch version of go that was so far behind current.

PR #5588, opened 2023-05-03
Update differential testing script to geth 1.11.6

trianglesphere:
**Description**

For whatever reason the contract test op-geth version got overridden so we need to update it here.

**Metadata**

- Fixes CLI-3916


refcell (approved):
Looks good, markdown check blocking

clabby (approved):

PR #5589, opened 2023-05-03
op-chain-ops: check-migration-quick

tynes:
**Description**

Adds a new tool that can be used as part of the migration that checks a minimal amount of information to ensure that the database migration was successful. It checks the extradata in the block header for the bedrock magic and errors if its not found. It will log information about the block header as well.

Closes CLI-3915


trianglesphere (approved):
lgtm

PR #5595, opened 2023-05-03
feat(contracts-periphery): create Faucet contract

tremarkley:
context & discussion
https://www.notion.so/oplabs/Superchain-faucet-backend-changes-ce42ce8e9f3b4725a41b696a822db4af?pvs=4

This contract will be used for distributing testnet funds from the [Superchain faucet](https://www.notion.so/oplabs/Superchain-Faucet-6f9bc01ddbcc441e814ad8636a003b48?pvs=4)

Added Faucet contract
Added tests


packages/contracts-periphery/contracts/universal/faucet/Faucet.sol
```diff
@@ -0,0 +1,273 @@
...
+     * @notice Maps from id to nonces to whether or not they have been used.
+     */
+    mapping(bytes => mapping(bytes32 => bool)) public usedNonces;
+
+	/**
+     * @notice Modifier that makes a function admin priviledged.
+     */
+    modifier priviledged() {
+        require(
+			msg.sender == ADMIN,
```
tynes, line 179: tabs vs spaces, everywhere else is spaces

packages/contracts-periphery/contracts/universal/faucet/Faucet.sol
```diff
@@ -0,0 +1,273 @@
...
+    ) {
+        ADMIN = admin;
+    }
+
+    /**
+     * @notice Allows users to donate ETH to this contract.
+     */
+    receive() external payable {
+	    // Thank you!
+	}
```
tynes, line 199: needs lint

packages/contracts-periphery/contracts/universal/faucet/Faucet.sol
```diff
@@ -0,0 +1,273 @@
...
+     * @notice Allows the admin to withdraw funds.
+     *
+     * @param recipient Address to receive the funds.
+     * @param amount    Amount of ETH in wei to withdraw.
+     */
+    function withdraw(
+        address payable recipient,
+        uint256 amount
+    ) public priviledged {
+		new SafeSend{value: amount}(recipient);
```
tynes, line 211: Why are you using `selfdestruct` to send funds? See the `SafeCall` library in `contracts-bedrock`. There is no guarantee that `selfdestruct` will work like this into the future as its scheduled to be removed from the protocol. It may or may not still send all of the funds depending on the particular EIP that is implemented

smartcontracts: Lawl that was me. Currently the most foolproof way of sending funds without triggering reentrancy.

packages/contracts-periphery/contracts/universal/faucet/Faucet.sol
```diff
@@ -0,0 +1,273 @@
...
+            bool
+        );
+}
+
+/**
+ * @title  AdminFAM
+ * @notice FaucetAuthModule that allows an admin to sign off on a given faucet drip. Takes an admin
+ *         as the constructor argument.
+ */
+contract AdminFAM is FaucetAuthModule, Semver, EIP712Upgradeable {
```
smartcontracts, line 56: I'd break this out into its own file, same with the FaucetAuthModule interface, which I'm realizing should be `IFaucetAuthModule`

packages/contracts-periphery/contracts/universal/faucet/Faucet.sol
smartcontracts: It seems highly likely that any future selfdestruct will still send funds, but fwiw it's easy enough to change this contract later.

packages/contracts-periphery/contracts/universal/faucet/Faucet.sol
```diff
@@ -0,0 +1,172 @@
...
+    constructor(address payable _recipient) payable {
+        selfdestruct(_recipient);
+    }
+}
+
+/**
+ * @title  Faucet
+ * @notice Faucet contract that drips ETH to users.
+ */
+contract Faucet {
```
tremarkley, line 23: Does this contract also need to extend `EIP712Upgradeable` in order for it to be upgradeable in the future?

smartcontracts: I don't think we really need to worry about upgradeability here. It's easy enough to claw back the funds and deploy a new faucet.

packages/contracts-periphery/contracts/universal/faucet/authmodules/AdminFaucetAuthModule.sol
```diff
@@ -0,0 +1,72 @@
...
+import { SignatureChecker } from "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
+import { IFaucetAuthModule } from "./IFaucetAuthModule.sol";
+import { Faucet } from "../Faucet.sol";
+
+/**
+ * @title  AdminFaucetAuthModule
+ * @notice FaucetAuthModule that allows an admin to sign off on a given faucet drip. Takes an admin
+ *         as the constructor argument.
+ */
+contract AdminFaucetAuthModule is IFaucetAuthModule, Semver, EIP712Upgradeable {
```
smartcontracts, line 17: I don't think we need to make this upgradeable. We can always just deploy a new auth module and reconfigure the faucet. Much easier than putting this thing behind a proxy.

packages/contracts-periphery/contracts/universal/faucet/Faucet.sol
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * @notice Emitted on each drip.
+     * @param authModule The type of authentication that was used for verifying the drip.
```
smartcontracts, line 26: Nit: missing whitespace between notice and params

packages/contracts-periphery/contracts/universal/faucet/authmodules/AdminFaucetAuthModule.sol
tremarkley: ok sounds good, I changed this to just use EIP712

packages/contracts-periphery/contracts/universal/faucet/Faucet.sol
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * @notice Mapping of authentication IDs to the next timestamp at which they can be used.
+     */
+    mapping(IFaucetAuthModule => mapping(bytes => uint256)) public timeouts;
+
+    /**
+     * @notice Maps from id to nonces to whether or not they have been used.
+     */
+    mapping(bytes => mapping(bytes32 => bool)) public usedNonces;
```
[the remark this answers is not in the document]
tremarkley: done 👍

packages/contracts-periphery/contracts/universal/faucet/authmodules/AdminFaucetAuthModule.sol
```diff
@@ -0,0 +1,66 @@
...
+    function verify(
+        Faucet.DripParameters memory _params,
+        bytes memory _id,
+        bytes memory _proof
+    ) external view returns (bool) {
+        // Generate a EIP712 typed data hash to compare against the proof.
+        bytes32 digest = _hashTypedDataV4(
+            keccak256(abi.encode(PROOF_TYPEHASH, _params.recipient, _params.nonce, _id))
+        );
+        return SignatureChecker.isValidSignatureNow(ADMIN, digest, _proof);
```
smartcontracts, line 64: tiny style nit, would prefer whatever the linted version is of:

```
return SignatureChecker.isValidSignatureNow(
    ADMIN,
    _hashTypedDataV4(
       keccak256(abi.encode(PROOF_TYPEHASH, _params.recipient, _params.nonce, _id))
    ),
    _proof,
);
```

smartcontracts (approved):
LGTM. We should put like 1k Goerli ETH into this contract and tell people they can keep the Goerli ETH if they break the contract.

PR #5596, opened 2023-05-03
op-program: Allow new named chains to be added without the op-geth config

ajsutton:
**Description**

Allow new named chains to be added without the op-geth config required by the fault proof program.

Simplifies the bedrock migration and the new mainnet config can be added afterwards.

**Metadata**

https://linear.app/optimism/issue/CLI-3936/add-mainnet-chain-config-to-fpp logged to add the config post-migration and re-enable the test since newly created chains won't have the complication of migration so should just be added immediately.


Inphi (approved):
LGTM!

PR #5597, opened 2023-05-04
ci: Change bedrock link checker to run nightly instead of on every PR

ajsutton:
**Description**

Moves the bedrock markdown link check to a daily scheduled job rather than being run on every PR. This should make it less likely that we'll get blocked for spamming sites with suspicious looking requests and will at least reduce the noise on PRs if we do.

**Additional context**

Slack notification of failures is going to the new `notify-link-check` channel.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-3912/move-link-checking-to-a-scheduled-job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -612,10 +612,20 @@ jobs:
       - run:
           name: markdown lint
           command: yarn lint:specs:check
+
+  bedrock-markdown-links:
+    machine:
+      image: ubuntu-2204:2022.07.1
+    steps:
+      - checkout
       - run:
           name: link lint
           command: |
-            docker run --init -it -v `pwd`:/input lycheeverse/lychee --verbose --no-progress --exclude-loopback --exclude twitter.com --exclude-mail /input/README.md "/input/specs/**/*.md"
+            make bedrock-markdown-links
+      - slack/notify:
+          channel: C055R639XT9 #notify-link-check
+          event: fail
+          template: basic_fail_1
 
   fuzz-op-node:
     docker:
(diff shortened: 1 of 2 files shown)


refcell (approved):
Changes look good.

trianglesphere (commented):
Do we also want to remove it from the `main` workflow?

.circleci/config.yml
```diff
@@ -611,10 +612,20 @@ jobs:
...
+
+  bedrock-markdown-links:
+    machine:
+      image: ubuntu-2204:2022.07.1
+    steps:
+      - checkout
       - run:
           name: link lint
           command: |
-            docker run --init -it -v `pwd`:/input lycheeverse/lychee --verbose --no-progress --exclude-loopback --exclude twitter.com --exclude-mail /input/README.md "/input/specs/**/*.md"
```
ajsutton, line 617: @trianglesphere This bit removes it from the main workflow.  We still run this job because it also does linting - it's just the link checking we don't want to run on PRs.

The diff makes it look very confusing but we wind up with two separate jobs, one for linting that's still in the main work flow and one for link checks that's only in the scheduled flow:

```
  bedrock-markdown:
    machine:
      image: ubuntu-2204:2022.07.1
    steps:
      - checkout
      - check-changed:
          patterns: specs/(.*)\.md$
      - run:
          name: yarn dev deps  # todo: what's the best way to pull in the dependencies for linting? yarn install above is using production env without dev dependencies
          command: yarn install --production=false
      - run:
          name: specs toc
          command: yarn lint:specs:toc && git diff --exit-code ./specs
      - run:
          name: markdown lint
          command: yarn lint:specs:check

  bedrock-markdown-links:
    machine:
      image: ubuntu-2204:2022.07.1
    steps:
      - checkout
      - run:
          name: link lint
          command: |
            make bedrock-markdown-links
      - slack/notify:
          channel: C055R639XT9 #notify-link-check
          event: fail
          template: basic_fail_1
```

trianglesphere: Oh, I see that makes sense. I didn't realize it was tied up with the regular lint.

trianglesphere (approved):
lgtm

PR #5598, opened 2023-05-04
ci: Add go.mod and go.sum to rebuild all patterns

ajsutton:
**Description**

Add go.mod and go.sum to rebuild all patterns.

Any new go code winds up using these dependency definitions by default and they don't change often so it's safer to rebuild everything when they change to avoid surprises. For example the last op-geth update let a compile error slip through in contract differential testing (https://github.com/ethereum-optimism/optimism/pull/5588).


tynes (approved):

tynes:
Good catch!

trianglesphere (approved):

PR #5568, opened 2023-05-01
feat: Introduce a new op-challenger service for PoPs

refcell:
**Description**

This PR introduces a naive challenge agent ported from to be run as a sidecar for permissionless output proposal validation. In an effort to break up the challenger and slowly, iteratively introduce the challenger to the monorepo, this pr is kept minimal.

Initially, this will be run by a permissioned set of challengers in the **Attestation Dispute Game**, but will be decentralized over time as a variety of dispute games are introduced (primarily, the fault-based dispute game).

**Metadata**

Fixes CLI-3909

PR #5602, opened 2023-05-04
feat(pops): Challenger Config Logic

refcell:
### Background

Building off #5568, this PR introduces new changes to the `op-challenger` service.

The `op-challenger` is a naive, honest challenge agent that can be run as a standalone service for permissionless output proposal validation.

In the initial iteration of PoPs, only a permissioned set of challenge agents can dispute outputs as part of the **Attestation Dispute Game**. But once Fault Dispute Games are implemented, disputing outputs will be permissionless.

### Description

This PR adds configuration logic to the challenger service, as similar as possible to the `op-proposer` and other op-stack services.

Future PRs will introduce L2OutputOracle event listeners, DisputeGameFactory creation, and DisputeGame responses.

### Metadata

Fixes CLI-3940

### DIFF
--- a/op-challenger/challenger/config.go
+++ b/op-challenger/challenger/config.go
@@ -1,0 +1,92 @@
+package challenger
+
+import (
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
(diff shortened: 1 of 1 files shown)


trianglesphere (approved):
lgtm

PR #5603, opened 2023-05-04
sdk: fixup deposit eth task + hh config

tynes:
**Description**

Fixes hardhat config by adding mainnet network
and uses correct addresses in building the contracts object.


refcell (approved):
Nice!

roninjin10 (approved):

PR #5612, opened 2023-05-04
fix(op-batcher): Unexport Metrics Fields

refcell:
**Description**

Quick fix for the `op-batcher` `metrics.go` to change fields to be un-exported.


trianglesphere (approved):
tyvm

PR #5614, opened 2023-05-04
op-e2e: wait for observed seq reorg in TestMissingBatchE2E

Inphi:
**Description**

Don't rely on timing for the L2 sequencer reorg in TestMissingBatchE2E.

**Metadata**

- Fixes CLI-2998


ajsutton (approved):

trianglesphere (approved):
tyvm

PR #5617, opened 2023-05-04
feat(pops): Main Challenger Binary Logic

refcell:
**Description**

Adds main logic required to run an op-challenger binary stub.

**Metadata**

Fixes CLI-3944


ajsutton (approved):
Looks good. We really need to start building tests for this module though. The more code we add before starting to write tests, the harder it will be to add them later because the design isn't testable.

PR #5622, opened 2023-05-05
feat(op-service): Refactor Service Utilities

refcell:
**Description**

In the `op-batcher`, `op-proposer`, and now `op-challenger`, we've been duplicating utility code in `utils.go`. This is 3 separate instances of redundant, identical code that can be refactored out to reduce maintainability overhead.

This PR refactors and exports these common utilities in a new `op-service` `client` package and moves `parseAddress` into  the `op-service/util.go` file.


ajsutton:
Love this! Great little tidy up.

PR #5624, opened 2023-05-05
feat(pops): Refactor Challenger for Testability

refcell:
**Description**

Cleans up the op-challenger to follow the `op-program`'s format so we can test cli args easily.

NOTE: `challenger/utils.go` will be removed in #5622 :)


ajsutton (approved):
LGTM.  

Worth noting that the `CheckRequired` method in `flags.go` isn't being called now. The config checks pick it up but the error message they give is slightly less useful because it doesn't include the specific CLI option name that's missing (ie you get `missing l1 eth rpc url`instead of `flag --l1-eth-rpc is required`).

I suggest hooking that back up in the next PR and adding tests for the CLI flags to cover it. You do get a bit of duplication between the config tests and flag tests but they're quite simple tests and cli vs config are their own separate concerns so worth testing individually. Plus the CLI tests cover the actual parsing which can provide some surprises at times.

PR #5625, opened 2023-05-05
op-stack: Validate prefixed Environment Variables

trianglesphere:
**Description**

This looks for all environment variables with a prefix like OP_NODE and then validates them against the list of environment variables provided by the flags for the component. It is not a critical error to provide an unused environment error, but we do log at a warn level to suggest to users that the value is not taking an effect.

**Tests**

Tested manually

**Metadata**

- Fixes CLI-3949


refcell (approved):
Since the added `op-service/util.go` ValidateEnvVars function is isolated and uses reflect, might be nice to have a test or tests in a `util_test.go`. Otherwise, looks good.

ajsutton:
Good idea and the implementation is nice, but agree we should have some tests around the validation function.

ajsutton (approved):
Nice. I was expecting a mock log to make it testable but it's nice to have the extra separation of concerns by splitting it up.

PR #5626, opened 2023-05-05
op-node: Don't invoke p2p unsafe sync for old L2 blocks

trianglesphere:
**Description**

Limits when the p2p unsafe sync gets invoked to make sure that we don't request old blocks.
There is another mechanism which will limit how often a p2p request is made, but it is not
guaranteed to work.

**Tests**

Unit tests.


op-node/node/node.go
```diff
@@ -378,6 +378,10 @@ func (n *OpNode) RequestL2Range(ctx context.Context, start, end eth.L2BlockRef)
 		return n.rpcSync.RequestL2Range(ctx, start, end)
 	}
 	if n.p2pNode != nil && n.p2pNode.AltSyncEnabled() {
+		if time.Unix(int64(start.Time), 0).Before(time.Now().Sub(14 * time.Hour)) {
```
trianglesphere, line 381: TODO: Decide on the value here. 12hrs is goerli/mainnet sequencing window. I gave it a buffer on top of that. We could also go down to like 6 hrs.

ajsutton:
I'm missing some context here - why can't we request these blocks from p2p? From what I understand we shouldn't be using alt-sync at all if we're syncing from L1 (https://github.com/ethereum-optimism/optimism/blob/0d7859bbfcc3b8d5c9259f3b5baf2a4b6e7dc7e4/op-node/rollup/driver/state.go#L237-L242 should do that). As long as L1 is advancing we should either get new batches or reach the end of the sequence window - either way the L2 head advances and moves us within this range of blocks to fetch again.  So we'd be hitting this limit only in the case of an extended L1 outage?
I can see that requesting really old blocks would likely cause a lot of blocks to be held in the `quarantine` as p2p sync walked backwards along the chain so some limit likely makes sense there, just want to properly understand the situations we'd hit that limit as it likely affects what sizing is right.

trianglesphere:
> I'm missing some context here - why can't we request these blocks from p2p? From what I understand we shouldn't be using alt-sync at all if we're syncing from L1

So there are two uses for alt-sync
1. Alt sync from a L2 op-geth URL. This uses a trusted url to pull blocks from & then insert them locally. It is more intended to doing a sync from scratch
2. P2P unsafe sync. This requests recent unsafe blocks so a node can keep up with the unsafe portion of the chain. The most common case is when a node restarts. It will miss some of the unsafe blocks that got published & won't be able to sync the new unsafe blocks it has received until the safe portion of the chain catches up.

I put the time check inside the P2P sync because the rpc sync should be able to request old blocks. For the P2P sync it should just fetch blocks from L1 unless they are unsafe. The alt sync ticker mostly does this, but I wanted to add a second check for this so we don't accidentally request blocks that we cannot use.

ajsutton (approved):
Ah yeah that makes sense to me. Looks good.

PR #5629, opened 2023-05-06
fix comment of `ChainStorageContainer`

zhiqiangxu:
There're in fact only two instances of `ChainStorageContainer` deployed on `Layer 1` .

Queued transactions for the Canonical Transaction Chain are stored in `CanonicalTransactionChain.queueElements` instead.


tynes:
Thank you for looking into this but we are no longer accepting changes to the legacy contracts. They will be deprecated very soon

PR #5632, opened 2023-05-08
ci: Exclude explorer.optimism.io from link checker.

ajsutton:
**Description**

It regularly rejects requests from CI due to DoS prevention even though the link is valid.


refcell (approved):
LGTM

PR #5633, opened 2023-05-08
Avoid copying `.git/` to Docker images

saucepoint:
**Description**

The Dockerfiles were copying `optimism/.git/` to the images. This behavior was causing issues with a `docker-compose.yml` file that depended on the Dockerfiles in `optimism/op-*/`

The main concern with removing the `.git/` artifacts from the images is losing version control when accessing the containers in interactive mode

**Tests**

No tests added, the devnet seems to behave as expected. Tested sending ether on both local L1 and local L2. Can do more extensive testing if needed

**Additional context**

Workin' on a simple `docker-compose` which uses anvil as a drop-in replacement for the L1 (instead of geth). Will be useful for cross-chain-comm apps that depend on existing L1 apps 🤔 

**Metadata**

- N/A

**TODOs**

### DIFF
--- a/indexer/Dockerfile
+++ b/indexer/Dockerfile
@@ -3,7 +3,6 @@ FROM golang:1.18.0-alpine3.15 as builder
 RUN apk add --no-cache make gcc musl-dev linux-headers git jq bash
 
 COPY ./indexer /go/indexer
-COPY ./.git /go/.git
 COPY ./indexer/go.mod /go/indexer/go.mod
 COPY ./indexer/go.sum /go/indexer/go.sum
 
--- a/op-batcher/Dockerfile
+++ b/op-batcher/Dockerfile
@@ -13,8 +13,6 @@ COPY ./op-signer /app/op-signer
 COPY ./go.mod /app/go.mod
 COPY ./go.sum /app/go.sum
 
-COPY ./.git /app/.git
-
 WORKDIR /app/op-batcher
 
 RUN go mod download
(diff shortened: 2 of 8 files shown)


ajsutton:
I think this breaks the version reporting. ie running with `--version` now gives:
```
op-node version v0.0.0
```

instead of:
```
op-node version v0.0.0-090644ae-1682359932
```

This version is also logged at startup.

I'm not really sure why including `.git` in the builder container causes problems with docker-compose, it's ultimately just another directory to copy as part of the build process.

tynes:
> Workin' on a simple docker-compose which uses anvil as a drop-in replacement for the L1 (instead of geth). Will be useful for cross-chain-comm apps that depend on existing L1 apps thinking

Anvil must be in a mode where it produces blocks on an interval, lazy block production will break the devnet

saucepoint:
> I'm not really sure why including .git in the builder container causes problems with docker-compose, it's ultimately just another directory to copy as part of the build process.

My errors might be because I have `optimism/` installed as a git submodule. The repo looks like:

```
optimism/  # submodule
docker-compose.yml
```

with services looking like:
```yml
  ...
  op-batcher:
    depends_on:
      - l1
      - l2
      - op-node
    build:
      context: optimism/
      dockerfile: ./op-batcher/Dockerfile
  ...
```

Both `op-proposer` and `op-batcher` fail to build

```
#0 0.674 fatal: not a git repository: /app/../.git/modules/optimism
```

ajsutton:
ah yes, when it's included as a submodule it won't have its own `.git` directory and won't be able to work out versions correctly.  Is there a reason you need to build the optimism docker images from scratch? Every commit to the `develop` branch is published as an image already so you could just include the pre-built images in docker-compose which would be faster and avoid these issues.

saucepoint:
oh wasn't aware of existing images! that'll work

ajsutton:
For the record the images are available from the google cloud repo. e.g. `us-docker.pkg.dev/oplabs-tools-artifacts/images/op-node:latest`

PR #5634, opened 2023-05-08
feat(pops): Flag Tests for the Challenger

refcell:
**Description**

Follows @ajsutton's [suggestion](https://github.com/ethereum-optimism/optimism/pull/5624#pullrequestreview-1415963890) of adding back in the flags `CheckRequired` when creating the config from cli. Also adds flag tests.


ajsutton (approved):
Looks good.  We could add tests to confirm that we get an error if required flags aren't supplied as well.

op-challenger/flags/flags_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+}
+
+func TestCorrectEnvVarPrefix(t *testing.T) {
+	for _, flag := range Flags {
+		envVar := envVarForFlag(flag)
+		if envVar == "" {
+			t.Errorf("Failed to find EnvVar for flag %v", flag.GetName())
+		}
+		if envVar[:len("OP_CHALLENGER_")] != "OP_CHALLENGER_" {
```
trianglesphere, line 43: Can use https://pkg.go.dev/strings#HasPrefix here

PR #5636, opened 2023-05-08
feat(op-challenger): L2OutputOracle Event Listeners

refcell:
**Description**

Introduces an L2OutputOracle event listener for proposed outputs.

Wanted to keep this pr fairly minimal, but if it makes sense, can introduce subcommands here for the various functions to "verify" outputs, "attest" dgs, "create" dispute games, etc, etc.

### DIFF
--- a/op-challenger/challenger/challenger.go
+++ b/op-challenger/challenger/challenger.go
@@ -13,6 +13,7 @@ import (
 	log "github.com/ethereum/go-ethereum/log"
 
 	config "github.com/ethereum-optimism/optimism/op-challenger/config"
+	flags "github.com/ethereum-optimism/optimism/op-challenger/flags"
 	metrics "github.com/ethereum-optimism/optimism/op-challenger/metrics"
 
 	bindings "github.com/ethereum-optimism/optimism/op-bindings/bindings"
@@ -36,6 +37,8 @@ type Challenger struct {
 	l1Client *ethclient.Client
 
 	rollupClient *sources.RollupClient
+
+	gameType flags.GameType
 
 	// l2 Output Oracle contract
 	l2ooContract     *bindings.L2OutputOracleCaller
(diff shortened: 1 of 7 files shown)


ajsutton (commented):
I think there's a few things to consider here:
1. How do we ensure that we don't miss any outputs? e.g. if our logs subscription gets disconnected, we want to reconnect it and start scanning for logs from the last block we had seen. Otherwise a brief disconnection may cause us to miss an output and implicitly accept it as valid. Currently I think it all just dies if there's a disconnection, at some point it probably needs to get some reconnect logic.
2. Similarly, how do we handle errors - say because our trusted RPC node is down or not in sync when we try to validate a claim? We probably want to be able to retry validation of output claims.
3. It makes me a bit nervous to have the dispute game creation in the log handling loop with a 10 minute timeout. If it actually took that long we'd potentially have been disconnected from the logs subscription because we hit read timeouts etc. Depends a lot on the server config and other factors so maybe not actually an issue but events may come through quite fast if the L1 node falls behind and then has to sync back up.

I'm not at all sure, but I wonder if splitting up the tasks would make this easier. So we'd have three components, probably all running in their own go routines:
1. Listen to logs, do the parsing from contract ABI to something nicer to work with (e.g the big.Int parsing etc). This could then also track which block it's up to so it can start from where it left off. 
2. Validate a log. This could handle retrying as needed, potentially by just putting the log back into the input channel, though we need to be careful of deadlocks if that channel fills up. If the output is invalid then it passes it onto the create dispute game step.
3. Create a dispute game. 

Each would be connected by a channel, probably with a small buffer so it doesn't block immediately but still gets back pressure if there's a bunch of events coming through so the queue doesn't get too big.

Depending on what we'd do if validating a log fails maybe listening to logs and validating them should stay in the same component as they may need to coordinate their actions closely.

That said, I think this is a good starting point - it works, just doesn't provide guarantees that it won't miss logs. Probably just needs some more logging when it does skip logs. I don't have a problem with starting with that, and then adding in reliability guarantees later as long as we're prepared to refactor as needed to make them fit well.

trianglesphere (commented):
I know in the op-node we don't use log subscriptions & instead manually inspect every single log. This may be overkill for the op-challenger, but we should characterize this to see if we need to & retain the flexibility to switch over to reading all logs if we need to.

PR #5637, opened 2023-05-08
[op-batcher] Multiple pending channels

mdehoog:
**Description**

Adds support for multiple pending channels to be open in the batcher at once. This implements support for concurrent batch tx sends, even when a channel only has a single frame. This is a continuation of #5398 which only submitted multiple txs if a channel had multiple frames.

**Tests**

Added a new simple E2E test that sets the target tx size to 2 and starts the batcher after 10 L2 blocks have been generated, to ensure a lot of transactions are submitted concurrently to the L1.

**TODOs**


op-batcher/batcher/channel_manager.go
```diff
@@ -34,26 +33,20 @@ type channelManager struct {
...
 
 	// Pending data returned by TxData waiting on Tx Confirmed/Failed
-
-	// pending channel builder
-	pendingChannel *channelBuilder
-	// Set of unconfirmed txID -> frame data. For tx resubmission
-	pendingTransactions map[txID]txData
-	// Set of confirmed txID -> inclusion block. For determining if the channel is timed out
-	confirmedTransactions map[txID]eth.BlockID
+	currentChannel *channel
```
trianglesphere, line 36: comments on the usage of currentChannel + channelQueue would be helpful.
As far as I can tell we only add to the currentChannel & then pull frames from any of the channels in the channelQueue (preferring older channels in the queue).

trianglesphere (approved):
Changes LGTM. Thanks for implementing this.

We'll have to do a follow up PR to figure out the metrics. We've been running with #5865 in testing. Because this now shifts pending bytes from the loaded stage to the sitting in the channel stage, I think we may want to add a metric for how many bytes are submitted / confirmed on L1.

PR #5642, opened 2023-05-08
fix(ctb): Remove unused mainnet artifact

maurelian:
**Description**

The ERC721 Bridge artifacts were copied into the bedrock package
in order to simplify our HH deployment scripts, without needing to read
artifacts from both the legacy contracts and contracts-periphery
packages.

However only the Proxy artifact is necessary. This artifact needs to be
deleted in order to deploy the updated ERC721 bridge implementation
during the setup phase of the mainnet deployments.


clabby:
IMO, we should make an archive directory for past deployments rather than deleting them.

clabby (approved):

PR #5644, opened 2023-05-08
op-node: Precompute sync status per derivation loop

Inphi:
**Description**

This allows the sync status to be retrieved quickly via RPC. Caching the sync status is cheap as it only involves simple state reads.

**Metadata**

- Fixes CLI-3815

### DIFF
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -326,6 +326,8 @@ func (s *Driver) eventLoop() {
 				stepAttempts = 0
 				reqStep() // continue with the next step if we can
 			}
+		case respCh := <-s.stateReq:
+			respCh <- struct{}{}
 		case respCh := <-s.forceReset:
 			s.log.Warn("Derivation pipeline is manually reset")
 			s.derivation.Reset()
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -33,6 +33,9 @@ type Driver struct {
 	// The derivation pipeline determines the new l2Safe.
 	derivation DerivationPipeline
 
+	// Requests to block the event loop for synchronous execution to avoid reading an inconsistent state
+	stateReq chan chan struct{}
+
 	// Upon receiving a channel in this channel, the derivation pipeline is forced to be reset.
 	// It tells the caller that the reset occurred by closing the passed in channel.
 	forceReset chan chan struct{}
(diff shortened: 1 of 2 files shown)


op-node/rollup/driver/state.go
```diff
@@ -426,34 +427,22 @@ func (s *Driver) syncStatus() *eth.SyncStatus {
...
-	case <-ctx.Done():
-		return nil, ctx.Err()
-	}
+	return s.syncStatusCache.Load(), nil
 }
 
-// BlockRefWithStatus blocks the driver event loop and captures the syncing status,
-// along with an L2 block reference by number consistent with that same status.
-// If the event loop is too busy and the context expires, a context error is returned.
+// BlockRefWithStatus returns the syncing status along with an L2 block reference by number consistent with that same status.
```
trianglesphere, line 441: I believe that this should be blocking for the main loop. We rely on the invariant that the L2 block by number is part of the same chain as the sync status.

op-node/rollup/driver/state.go
```diff
@@ -426,34 +427,22 @@ func (s *Driver) syncStatus() *eth.SyncStatus {
...
-// SyncStatus blocks the driver event loop and captures the syncing status.
-// If the event loop is too busy and the context expires, a context error is returned.
+// updateSyncStatusCache updates the cached sync status
+func (s *Driver) updateSyncStatusCache() {
+	ss := s.syncStatus()
+	s.syncStatusCache.Store(ss)
+}
+
+// SyncStatus returns the syncing status.
 func (s *Driver) SyncStatus(ctx context.Context) (*eth.SyncStatus, error) {
```
trianglesphere, line 437: what happens if the returned sync status is nil because it has not been initialized yet?

op-node/rollup/driver/state.go
Inphi: Wasn't sure if we should maintain this behavior. It's a weak invariant because a reorg could in theory occur while the main loop is blocked. I can revert back to the way it was though. Main motivation for this is to prevent the batcher, for example, from being blocked waiting the sync status. Updating `BlockRefWithStatus` is just a bonus.

op-node/rollup/driver/state.go
Inphi: It returns a `null` JSON object, which I think is what we want. Seems less misleading than a zero valued `SyncStatus` struct.

op-node/rollup/driver/state.go
trianglesphere: If anything, I think we need to increase the strength of this invariant for the proposer. Having an L1 reorg so the L1 state is out of sync with the L2 state is fine. The biggest worry is if the L1 reorg has propagated to the state loop, but that L2 state has not been updated.

tynes:
Wondering the status of this PR, should we close it?

Inphi:
> Wondering the status of this PR, should we close it?

Got side tracked with other stuff. I'll wrap this up today.

tynes:
Ping on this again

op-node/rollup/driver/state.go
Inphi: Sorry for the delay. I've reverted this function back to what it was. It avoids desync L2 state and block info desync.

PR #5648, opened 2023-05-08
feat(migration): Mimic `xdm.baseGas` in `migrate.go`

clabby:
## Overview

Changes `migrate.go`'s `MigrateWithdrawalBaseGas` function to mimic `baseGas` in `CrossDomainMessenger.sol`. We set the `_minGasLimit` parameter to a fixed `1_000_000` gas due to the overhead of performing exact simulations on each migrated withdrawal during the live migration.

With a `1_000_000` minimum gas limit for the *outer* call (`OptimismPortal` -> `L1CrossDomainMessenger`) and a `0` minimum gas limit for the *inner* call (`L1CrossDomainMessenger` -> `tx.target` - see [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-chain-ops/crossdomain/migrate.go#L89)), the `SafeCall.hasMinGas` check in the XDM should always pass. In addition, because we reserve roughly `40_000` gas, we will have enough to finish execution of `relayMessage` if the external call from the `L1CrossDomainMessenger` to the target fails in order to guarantee replayability.

**Metadata**
Closes CLI-3969


clabby:
* develop
  * **PR #5648** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/5648" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/5648?utm_source=stack-comment).

tynes:
We will also need to update the sdk to include the same calculation, I can take over this PR and add that

op-chain-ops/crossdomain/migrate.go
```diff
@@ -112,16 +123,34 @@ func MigrateWithdrawalGasLimit(data []byte, chainID *big.Int) uint64 {
...
+	var overhead uint64
+	if chainID.Cmp(big.NewInt(420)) == 0 {
+		overhead = uint64(200_000)
+	} else {
+		// Mimic `baseGas` from `CrossDomainMessenger.sol`
+		overhead = uint64(
+			// Constant overhead
+			RelayConstantOverhead +
+				// Dynamic overhead (EIP-150)
+				(MinGasDynamicOverheadNumerator*1_000_000)/MinGasDynamicOverheadDenominator +
```
refcell, line 138: Maybe just add a little comment here about where the 1m comes from, otherwise LGTM

tynes (approved):

tynes:
Would love a review from @devexpod

tynes:
Can i get a review from @roninjin10 ?

PR #5650, opened 2023-05-08
feat(pops): Dispute Game Factory

refcell:
**Description**

Introduces the `DisputeGameFactory` contract and associated tests.


packages/contracts-bedrock/contracts/dispute/DisputeGameFactory.sol
```diff
@@ -0,0 +1,163 @@
...
+     * @param rootClaim The root claim of the DisputeGame.
+     * @param extraData Any extra data that should be provided to the created dispute game.
+     * @return _uuid The unique identifier for the given dispute game parameters.
+     */
+    function getGameUUID(
+        GameType gameType,
+        Claim rootClaim,
+        bytes memory extraData
+    ) public pure returns (Hash _uuid) {
+        assembly {
```
clabby, line 133: This isn't @refcell's fault, but solhint not allowing `assembly ("memory-safe")` is so sad

clabby (approved):

packages/contracts-bedrock/contracts/dispute/DisputeGameFactory.sol
refcell: ++

PR #5652, opened 2023-05-08
feat(proxyd): add limit to consensus block lag

felipe-op:
**Description**

Add a configurable limit to the maximum number of blocks a backend is allowed to lag behind.
When a backend is lagging too far, it is ignored from the consensus.

**Tests**

Integration tests + manual

**Invariants**

n/a

**Additional context**

This is part of the project Consensus Aware RPC Proxy: https://www.notion.so/oplabs/Consensus-Aware-RPC-Proxy-0138e029af814e4cbca2740c7888e02d

**Metadata**

- Fixes INF-201

**TODOs**
- [x] Author or reviewer has added an entry to the [current release notes draft][RND], if appropriate -- not applicable

### DIFF
--- a/proxyd/consensus_poller.go
+++ b/proxyd/consensus_poller.go
@@ -35,6 +35,7 @@ type ConsensusPoller struct {
 
 	banPeriod          time.Duration
 	maxUpdateThreshold time.Duration
+	maxBlockLag        uint64
 }
 
 type backendState struct {
@@ -160,6 +161,12 @@ func WithMaxUpdateThreshold(maxUpdateThreshold time.Duration) ConsensusOpt {
 	}
 }
 
+func WithMaxBlockLag(maxBlockLag uint64) ConsensusOpt {
+	return func(cp *ConsensusPoller) {
+		cp.maxBlockLag = maxBlockLag
+	}
+}
+
 func WithMinPeerCount(minPeerCount uint64) ConsensusOpt {
 	return func(cp *ConsensusPoller) {
 		cp.minPeerCount = minPeerCount
@@ -181,6 +188,7 @@ func NewConsensusPoller(bg *BackendGroup, opts ...ConsensusOpt) *ConsensusPoller
 
 		banPeriod:          5 * time.Minute,
 		maxUpdateThreshold: 30 * time.Second,
+		maxBlockLag:        50,
 		minPeerCount:       3,
 	}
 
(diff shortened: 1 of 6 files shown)


proxyd/consensus_poller.go
```diff
@@ -267,7 +275,12 @@ func (cp *ConsensusPoller) UpdateBackendGroupConsensus(ctx context.Context) {
 			continue
 		}
 
-		if lowestBlock == 0 || backendLatestBlockNumber < lowestBlock {
+		// find the highest common ancestor, ignoring backends that are too far lagging behind
+		// when the backend is too far ahead from current lowest, the current lowest is ignored
+		// when the backend if too far behind, the backend itself is ignored
+		if lowestBlock == 0 ||
```
mslipper, line 281: Just for my understanding here:

Let's say we have three nodes A, B, and C. A's height is 50, B's height is 100, and C's height is 101. Assume a max block lag of 10. If iterated in order A, B, and C:

1. For A, `lowestBlock` will first be set to 50.
2. For B, `lowestBlock` will be set to 100 because the`backendLatestBlockNumber > lowestBlock && uint64(backendLatestBlockNumber-lowestBlock) > cp.maxBlockLag` check will succeed.
3. For C, `lowestBlock` will be set to 100 because the `backendLatestBlockNumber > lowestBlock && uint64(backendLatestBlockNumber-lowestBlock) > cp.maxBlockLag` check will fail.

proxyd/consensus_poller.go
```diff
@@ -305,12 +318,14 @@ func (cp *ConsensusPoller) UpdateBackendGroupConsensus(ctx context.Context) {
...
 				- with minimum peer count
 				- updated recently
+				- not lagging
 			*/
 			bs := cp.backendState[be]
 			notUpdated := bs.lastUpdate.Add(cp.maxUpdateThreshold).Before(time.Now())
 			isBanned := time.Now().Before(bs.bannedUntil)
 			notEnoughPeers := !be.skipPeerCountCheck && bs.peerCount < cp.minPeerCount
-			if !be.IsHealthy() || be.IsRateLimited() || !be.Online() || notUpdated || isBanned || notEnoughPeers {
+			lagging := (bs.latestBlockNumber < proposedBlock) && uint64(proposedBlock-bs.latestBlockNumber) >= cp.maxBlockLag
```
mslipper, line 327: Shouldn't this just be `bs.latestBlockNumber < proposedBlock`? We don't want to route requests to nodes that don't have the proposed block under any circumstances.

mslipper (changes requested):
Left one clarifying comment, and one commend about a potential logic issue. Otherwise LGTM.

proxyd/consensus_poller.go
felipe-op: Your understanding is correct

felipe-op:
I replaced the lag comparison to work with a global highest known block.

This is better than previous impl proposal because it was ordering-dependent.  Consider the following case with 3 nodes A, B and C, and a max lag of 50:


```
  30   50   80
 --A----B----C--


a,b,c
 a,b, pick a @ 30
 a,c, override a because its lagging too much, pick c @ 80
 result: c @ 80

b,c,a
 b,c, pick b @ 50
 b,a, pick a @ 30
 result: a @ 30

c,b,a
 c,b, pick b @ 50
 b,a, pick a @ 30
 result: a @ 30
```

PR #5653, opened 2023-05-09
Use strings package to check for a prefix

trianglesphere:
**Description**

Use a built in function rather than manually slicing strings.

### DIFF
--- a/op-challenger/flags/flags_test.go
+++ b/op-challenger/flags/flags_test.go
@@ -40,7 +40,7 @@ func TestCorrectEnvVarPrefix(t *testing.T) {
 		if envVar == "" {
 			t.Errorf("Failed to find EnvVar for flag %v", flag.GetName())
 		}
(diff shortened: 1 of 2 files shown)


op-challenger/flags/flags_test.go
```diff
@@ -40,7 +40,7 @@ func TestCorrectEnvVarPrefix(t *testing.T) {
 		if envVar == "" {
 			t.Errorf("Failed to find EnvVar for flag %v", flag.GetName())
 		}
-		if envVar[:len("OP_CHALLENGER_")] != "OP_CHALLENGER_" {
+		if !strings.HasPrefix("OP_CHALLENGER_", envVar) {
```
ajsutton, line 43: I think the args need to go the other way around.

refcell (approved):
GTG once @ajsutton's suggestion is impl'd

op-challenger/flags/flags_test.go
trianglesphere: yep

ajsutton (approved):

PR #5656, opened 2023-05-09
fix(proxyd): eth2 block tags {safe, finalized} should be valid tag values and avoid cache

felipe-op:
**Description**

This patch fixes a bug in which the block tags {safe, finalized} were being rejected. This allows them to be used and avoid them to be cached.

**Tests**

`go test ./...`

**Invariants**

n/a 

**Additional context**

n/a

**Metadata**

- Fixes INF-171

**TODOs**

- [X] Author or reviewer has added an entry to the [current release notes draft][RND], if appropriate. -- not applicable


proxyd/methods.go
```diff
@@ -271,7 +271,11 @@ func (e *EthGasPriceMethodHandler) PutRPCMethod(context.Context, *RPCReq, *RPCRe
 }
 
 func isBlockDependentParam(s string) bool {
-	return s == "latest" || s == "pending"
+	return s == "earliest" ||
```
Inphi, line 274: we can cache `earliest` as that's block 0 which doesn't change.

mslipper (approved):

PR #5672, opened 2023-05-10
op-node: Skip resetting the scheduler timer and channel when a desired action is already pending

ajsutton:
**Description**

Each time the driver goes round it's loop, it asks the sequencer when it wants to perform its next action and then resets the scheduler timer to that delay before it proceeds to a `select` which checks which of the possible channels (including the sequencer timer) is ready for execution.

This introduces a race condition between the sequencer timing firing again after calling `Reset` (which may happen asynchronously even when the delay is 0) and the `select` checking the sequencer timer channel. If the timer doesn't fire in time, the driver goes back round the loop and will reset the sequencer timer again, restarting the race.

To avoid this race, if the sequencer requests a delay of 0 either:
* if the channel already has a pending entry, do nothing - the action is already scheduled correctly
* otherwise create a new channel and synchronously publish to it, guaranteeing it will be ready to execute

**Tests**

**Invariants**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

**TODOs**

### DIFF
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -192,8 +192,23 @@ func (s *Driver) eventLoop() {
 	var sequencerCh <-chan time.Time
 	planSequencerAction := func() {
 		delay := s.sequencer.PlanNextSequencerAction()
+		if delay == 0 {
+			if len(sequencerCh) > 0 {
+				// We want to perform the next action immediately and it's already scheduled so just leave it alone.
+				return
+			}
+			// Otherwise create a channel and trigger it immediately.
+			sequencerTimer.Stop() // Stop the timer as we're not using it now.
+			ch := make(chan time.Time, 1)
+			sequencerCh = ch
+			ch <- time.Now()
+			return
+		}
+
+		// Setup the timer to trigger after the requested delay
 		sequencerCh = sequencerTimer.C
-		if len(sequencerCh) > 0 { // empty if not already drained before resetting
+		if !sequencerTimer.Stop() {
+			// empty channel if the timer had fired before we stopped it
 			<-sequencerCh
 		}
 		sequencerTimer.Reset(delay)


op-node/rollup/driver/state.go
```diff
@@ -192,8 +192,21 @@ func (s *Driver) eventLoop() {
 	var sequencerCh <-chan time.Time
 	planSequencerAction := func() {
 		delay := s.sequencer.PlanNextSequencerAction()
+		if delay == 0 {
+			if len(sequencerCh) > 0 {
+				// We want to perform the next action immediately and it's already scheduled so just leave it alone.
+				return
+			}
+			// Otherwise create a channel and trigger it immediately.
```
trianglesphere, line 200: Should probably also stop the timer in this case.

op-node/rollup/driver/state.go
```diff
@@ -192,8 +192,21 @@ func (s *Driver) eventLoop() {
...
+			ch := make(chan time.Time, 1)
+			sequencerCh = ch
+			ch <- time.Now()
+			return
+		}
+		// Setup the timer to trigger after the requested delay
 		sequencerCh = sequencerTimer.C
-		if len(sequencerCh) > 0 { // empty if not already drained before resetting
+		if len(sequencerCh) > 0 {
```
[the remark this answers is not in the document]
ajsutton: Yeah it's probably worth doing. It probably doesn't make a lot of difference in this case - the worst case is that the timer fires between cleaning the channel and resetting the timer which causes us to run a scheduler action sooner than we should. Depends if the sequencer action is smart enough to handle that case or not, but definitely simpler if we just get it right.

op-node/rollup/driver/state.go
ajsutton: Yeah worth while so we don't leave it hanging around only to ignore its result.

ajsutton:
Add do-not-merge label so that we can be deliberate about when to merge this rather than having mergify kick in.

trianglesphere:
IMO this change is probably net positive, but not strictly needed. I think we'll also want to follow up with this later.

@protolambda what do you think about bringing in this change now?

protolambda:
> IMO this change is probably net positive, but not strictly needed. I think we'll also want to follow up with this later.

Agreed

> @protolambda what do you think about bringing in this change now?

I'd avoid making changes to it now. We are free to update the sequencing timing, but should prefer stability so close to launch.

ajsutton:
Agreed, closing for now to avoid PRs hanging around.

PR #5681, opened 2023-05-10
Add Safe transaction templates to be used in the Bedrock mainnet launch.

zchn:
**Description**

Add Safe transaction templates to be used in the Bedrock mainnet launch.

Questions for reviewers: Is this the right place to add these files? Is there a 
better folder I should move them to? 

**Tests**

No test added since it's just json files. Will manually test by loading them in the 
Safe UI once all smart contract addresses are determined.

**Invariants**

N/A

**Additional context**

None.

**Metadata**

None

**TODOs**

None


tynes:
Do you mind moving the `multisig` directory to `contracts-bedrock` so its `contracts-bedrock/multisig`? I don't think these need to live in the repo long term, but open to the idea. The `deployments` directory is meant for hardhat deploy artifacts and we have code that walks the fs so having them in there may result in unintended consequences

PR #5684, opened 2023-05-10
op-node: Modify p2p unsafe sync rate limits

trianglesphere:
**Description**

This PR adds a global rate limit to outgoing P2P unsafe sync requests in addition to the existing local rate limit. The global rate limit outgoing & ingoing rate limits have the same parameters.

This PR also modifies the rate limit parameters. I have increased the burst values (the size of a full token bucket) and in the case of the per peer rate limit decreased the rate. This was done because the burst values were smaller than the per second limits.

Lastly, `clientErrRateCost` must be the same as the per peer rate limit burst because if you request more tokens than the burst, the request will be rejected & the code will not achieve it's goal of delaying future requests to a peer where the request failed.

**Tests**

No tests.

**Metadata**

- Fixes CLI-3967

### DIFF
--- a/op-node/p2p/sync.go
+++ b/op-node/p2p/sync.go
@@ -48,16 +48,17 @@ const (
 	maxThrottleDelay = time.Second * 20
 	// Do not serve more than 20 requests per second
 	globalServerBlocksRateLimit rate.Limit = 20
-	// Allow up to 5 concurrent requests to be served, eating into our rate-limit
-	globalServerBlocksBurst = 5
-	// Do not serve more than 5 requests per second to the same peer, so we can serve other peers at the same time
-	peerServerBlocksRateLimit rate.Limit = 5
-	// Allow a peer to burst 3 requests, so it does not have to wait
-	peerServerBlocksBurst = 3
+	// Allows a burst of 2x our rate limit
+	globalServerBlocksBurst = 40
+	// Do not serve more than 4 requests per second to the same peer, so we can serve other peers at the same time
+	peerServerBlocksRateLimit rate.Limit = 4
+	// Allow a peer to request 30s of blocks at once
+	peerServerBlocksBurst = 15
 	// If the client hits a request error, it counts as a lot of rate-limit tokens for syncing from that peer:
 	// we rather sync from other servers. We'll try again later,
 	// and eventually kick the peer based on degraded scoring if it's really not serving us well.
-	clientErrRateCost = 100
+	// TODO(CLI-4009): Use a backoff rather than this mechanism.
+	clientErrRateCost = peerServerBlocksBurst
 )
 
 func PayloadByNumberProtocolID(l2ChainID *big.Int) protocol.ID {
@@ -204,6 +205,9 @@ type SyncClient struct {
 
 	receivePayload receivePayloadFn
 
+	// Global rate limiter for all peers.
+	globalRL *rate.Limiter
+
 	// resource context: all peers and mainLoop tasks inherit this, and start shutting down once resCancel() is called.
 	resCtx    context.Context
 	resCancel context.CancelFunc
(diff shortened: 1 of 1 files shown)


op-node/p2p/sync.go
```diff
@@ -231,6 +234,7 @@ func NewSyncClient(log log.Logger, cfg *rollup.Config, newStream newStreamFn, rc
 		requests:        make(chan rangeRequest), // blocking
 		peerRequests:    make(chan peerRequest, 128),
 		results:         make(chan syncResult, 128),
+		globalRL:        rate.NewLimiter(globalServerBlocksRateLimit, globalServerBlocksBurst),
```
ajsutton, line 237: Why would we limit the total number of requests we send? As long as we spread it out across enough peers we don't need to throttle ourselves.  Restricting the number of concurrent requests we have in-flight makes sense (and we do that I believe) but not a global rate limit (unless I'm missing something).

op-node/p2p/sync.go
```diff
@@ -48,16 +48,16 @@ const (
...
+	globalServerBlocksBurst = 40
+	// Do not serve more than 4 requests per second to the same peer, so we can serve other peers at the same time
+	peerServerBlocksRateLimit rate.Limit = 4
+	// Allow a peer to request 30s of blocks at once
+	peerServerBlocksBurst = 15
 	// If the client hits a request error, it counts as a lot of rate-limit tokens for syncing from that peer:
 	// we rather sync from other servers. We'll try again later,
 	// and eventually kick the peer based on degraded scoring if it's really not serving us well.
-	clientErrRateCost = 100
+	clientErrRateCost = peerServerBlocksBurst
```
ajsutton, line 60: hmm, this works if we assume all nodes use these same values, but if we want to tolerate variations then we probably want to wait a decent chunk of time to ensure that we leave enough time for the peer to resume serving us requests.  Given `WaitN` doesn't allow waiting for capacity beyond the burst limit, we probably need to do the wait in a loop or maybe wait until the full burst amount is available again plus an exponential backoff if we get repeated rate limit errors.

trianglesphere: So this is in our internal rate limit for requesting for peers, but yes, we may want to use a different mechanism to back off in the case that there is an error.

ajsutton: From the comment, my understanding is that this value is used to get us to stop making new requests from a peer for a while after we get an error from the client that we've exceeded their rate limiting. So with this reduced to just `peerServerBlocksBurst` we'll only wait until our counter for that peer reaches zero then try again.  If the remote peer has a lower limit or different time window to us though, we'll get another rejection for exceeding their request limit. So we want to wait a longer period to ensure the remote side will allow us to start making requests again - hence this was originally set much higher than the burst limit.
It looks like we're using this value to back off after any error, not just rate exceeded errors, but I suspect we still want to be waiting a reasonable period of time whereas I think we're now making that back-off period quite short.

op-node/p2p/sync.go
ajsutton: Yeah you're right - we have a limit on the number of pending requests in `peerRequests` but only limit to 1 in-flight request per peer.  I'm on the fence about this - what are we trying to protect against with this?
I'm just always nervous about limiting the speed sync can catch up because you generally can't do anything else useful until you're back in sync.

protolambda:
The changes look fine to me, but are the comments of @ajsutton resolved or not?

op-node/p2p/sync.go
trianglesphere: We're trying to limit how many blocks are fetched over P2P. It's set to 20 RPS. In general nodes should try to get data from L1 instead.

op-node/p2p/sync.go
trianglesphere: Yea. We'll consume the full burst & pretty soon be able to retry (250ms until the bucket has a single token in it). We should fix it here & I'll create a ticket to track the full fix.

ajsutton (approved):
Yep looks fine to me given a separate story to increase the retry delay after getting errors. It used to break in that case so this is an improvement. :)

PR #5685, opened 2023-05-10
op-node: Bake in internal devnet to chain cfgs

trianglesphere:
**Description**

Allow the internal devnet to be selected with a `--network` flag.

**Metadata**

- Fixes CLI-3937

### DIFF
--- a/op-node/chaincfg/chains.go
+++ b/op-node/chaincfg/chains.go
@@ -77,8 +77,39 @@ var Goerli = rollup.Config{
 	RegolithTime:           u64Ptr(1679079600),
 }
 
+var Devnet = rollup.Config{
+	Genesis: rollup.Genesis{
+		L1: eth.BlockID{
+			Hash:   common.HexToHash("0xe0473633677bd7a134925b00f86e688655c28352b81d76a966b7a813e0bc4fce"),
+			Number: 8598297,
+		},
+		L2: eth.BlockID{
+			Hash:   common.HexToHash("0x742e03d9c8fe4dfd96f8e2d53d3815ec5537cbe3d88ad28cbb75c17ff2672867"),
+			Number: 0,
+		},
+		L2Time: 1677973680,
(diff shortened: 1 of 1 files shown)


op-node/chaincfg/chains.go
```diff
@@ -77,8 +77,39 @@ var Goerli = rollup.Config{
...
+		},
+	},
+	BlockTime:              2,
+	MaxSequencerDrift:      1200,
+	SeqWindowSize:          3600,
+	ChannelTimeout:         120,
+	L1ChainID:              big.NewInt(5),
+	L2ChainID:              big.NewInt(997),
+	BatchInboxAddress:      common.HexToAddress("0xff00000000000000000000000000000000000997"),
+	DepositContractAddress: common.HexToAddress("0xc6170e048b7daef6d0b6cbbad9aca0e5370ddbbc"),
```
tynes, line 105: This is a bad idea because the contract addresses are not deterministic so any changes to the deployment process will result in these becoming stale and this will be difficult to debug

tynes (changes requested):
plz reconsider not doing this

op-node/chaincfg/chains.go
trianglesphere: This is a deployed devnet, not generic one.

ajsutton (commented):
I'm not sure we should embed internal devnet config into the public release. I can see the desire for it though so would be ok with it as long as we name it something better than just `devnet`. 

If we do add this, it would be great if we could add the execution chain config to `op-program/host/config/chaincfg.go` as well.

op-node/chaincfg/chains.go
tynes: Oh I see. Ok that makes more sense. Thought of the local devnet

trianglesphere:
Closing b/c we will continue to configure this network with a `rollup.json` file.

PR #5686, opened 2023-05-11
op-program: Update goerli verification script to handle long safe head stalls

ajsutton:
**Description**

Uses the latest output if there isn't an output after the current finalized block and checks it's at or before the finalized block.

Avoids failures like https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/17993/workflows/fa162c20-dc53-4c24-b2e3-7312e9574116/jobs/702292


trianglesphere (approved):
lgtm. Will put up another PR to remove the `println` in this file - that's an old builtin that we shouldn't be using

trianglesphere (approved):
lgtm. Will put up another PR to remove the `println` in this file - that's an old builtin that we shouldn't be using

PR #5696, opened 2023-05-12
feat(pops): SystemConfig Signer Set

refcell:
**Description**

Updates the `SystemConfig` to hold a signer set. Adds fuzzed testing for the `SystemConfig`.

**Metadata**

Fixes CLI-3986


tynes:
Can we target a feature branch or make this into a draft? Any production contract is code frozen until after the bedrock upgrade

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -70,6 +74,19 @@ contract SystemConfig is OwnableUpgradeable, Semver {
      */
     ResourceMetering.ResourceConfig internal _resourceConfig;
 
+    /**
+     * @notice The `signerSet` is a set of addresses that are allowed to issue positive attestations
+     *         for alternative output proposals in the `AttestationDisputeGame`.
+     */
+    address[] internal _signerSet;
```
tynes, line 81: nit: I'd prefer to call them attesters because its possible we end up having multiple types of "signers"

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
```diff
@@ -22,12 +22,16 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+     * @custom:value SIGNER_SET           Represents an update to the signer set.
+     * @custom:value SIGNATURE_THRESHOLD  Represents an update to the signature threshold.
      */
     enum UpdateType {
         BATCHER,
         GAS_CONFIG,
         GAS_LIMIT,
-        UNSAFE_BLOCK_SIGNER
+        UNSAFE_BLOCK_SIGNER,
+        SIGNER_SET,
```
tynes, line 33: If the `op-node` experiences an unknown type, it will result in it erroring. The `op-node` doesn't need to know about these changes. Ideally, there are events emitted for all changes to the state of this contract so the set of events can reproduce its state but we cannot do that now so we will need to remove the 2 update types added in this PR

packages/contracts-bedrock/contracts/test/SystemConfig.t.sol
```diff
@@ -143,6 +143,31 @@ contract SystemConfig_Setters_TestFail is SystemConfig_Init {
         vm.expectRevert("SystemConfig: precision loss with target resource limit");
         sysConf.setResourceConfig(config);
     }
+
+    function testFuzz_setSigner_nonOwner_reverts(address _signer, bool _authenticated) external {
+        vm.expectRevert("Ownable: caller is not the owner");
```
tynes, line 148: This should have a `vm.assume(_signer != sysConf.owner())` ?

packages/contracts-bedrock/contracts/test/SystemConfig.t.sol
```diff
@@ -201,4 +226,37 @@ contract SystemConfig_Setters_Test is SystemConfig_Init {
...
+                    emit ConfigUpdate(
+                        0,
+                        SystemConfig.UpdateType.SIGNER_SET,
+                        abi.encode(_signer, _authenticated)
+                    );
+                }
+            }
+        }
+        vm.prank(sysConf.owner());
+        sysConf.setSigner(_signer, _authenticated);
```
tynes, line 248: It would be nice to assert on the post return value of `signerSet` someplace, I don't see thorough test coverage of adding/removing with specific assertions on the signer set. Basically to cover the removal case with the inplace swap of the last element + the pop

packages/contracts-bedrock/contracts/L1/SystemConfig.sol
refcell: Removed and split their canonical updates into new events

refcell:
Closing in favor of #5702

PR #5697, opened 2023-05-12
feat(pops): L2OutputOracle Updates

refcell:
**Description**

Updates the L2OutputOracle contract for permissionless output proposals.

**Metadata**

Fixes CLI-3987

### DIFF
--- a/packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
+++ b/packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
@@ -2,8 +2,16 @@
 pragma solidity 0.8.15;
 
 import { Initializable } from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
-import { Semver } from "../universal/Semver.sol";
+
 import { Types } from "../libraries/Types.sol";
+import { Semver } from "../universal/Semver.sol";
+
+import { GameType } from "../libraries/DisputeTypes.sol";
+import { GameStatus } from "../libraries/DisputeTypes.sol";
+
+import { IDisputeGame } from "../dispute/IDisputeGame.sol";
+import { IBondManager } from "../dispute/IBondManager.sol";
+import { IDisputeGameFactory } from "../dispute/IDisputeGameFactory.sol";
 
 /**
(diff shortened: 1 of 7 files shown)


packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -244,6 +276,25 @@ contract L2OutputOracle is Initializable, Semver {
...
+     * @return The output for the given block number.
+     */
+    function getL2OutputByNumber(uint256 _l2BlockNumber)
+        external
+        view
+        returns (Types.OutputProposal memory)
+    {
+        uint256 index = l2OutputIndices[_l2BlockNumber];
+        if (index == 0) {
+            return Types.OutputProposal({ outputRoot: bytes32(0), timestamp: 0, l2BlockNumber: 0 });
```
tynes, line 293: Should this revert to indicate non existence rather than returning the null value?

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -217,7 +239,13 @@ contract L2OutputOracle is Initializable, Semver {
             );
         }
 
-        emit OutputProposed(_outputRoot, nextOutputIndex(), _l2BlockNumber, block.timestamp);
+        // Post the bond to the bond manager
+        BOND_MANAGER.post{ value: msg.value }(
+            keccak256(abi.encode(_l2BlockNumber)),
```
tynes, line 244: nit: use kwargs syntax so its easier to see what the args are
Also why hash the blocknumber? There can be colliding ids..

packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol
```diff
@@ -217,7 +239,13 @@ contract L2OutputOracle is Initializable, Semver {
...
+            msg.sender,
+            uint256(10 days)
```
tynes, line 246: Why 10 days? It should be a constant

tynes:
This should definitely target a feature branch

PR #5699, opened 2023-05-12
op-chain-ops: better comments

tynes:
**Description**

Add comments to ensure that a known bug is commented. See https://github.com/ethereum-optimism/optimism/pull/5682 for a longer description.

Adds a minimal check to make sure this bug is never encountered in practice.

### DIFF
--- a/op-chain-ops/crossdomain/legacy_withdrawal.go
+++ b/op-chain-ops/crossdomain/legacy_withdrawal.go
@@ -57,7 +57,9 @@ func (w *LegacyWithdrawal) Encode() ([]byte, error) {
 	return out, nil
 }
 
-// Decode will decode a serialized LegacyWithdrawal
+// Decode will decode a serialized LegacyWithdrawal. There is a known inconsistency
+// where the decoded `msg.sender` is not authenticated. A round trip of encoding and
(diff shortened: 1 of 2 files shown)


op-chain-ops/crossdomain/legacy_withdrawal.go
```diff
@@ -57,7 +57,9 @@ func (w *LegacyWithdrawal) Encode() ([]byte, error) {
 	return out, nil
 }
 
-// Decode will decode a serialized LegacyWithdrawal
+// Decode will decode a serialized LegacyWithdrawal. There is a known inconsistency
```
mslipper, line 60: Can we add a panic in the `Encode` method as well to prevent improper usage elsewhere?

mslipper (approved):

op-chain-ops/crossdomain/legacy_withdrawal.go
tynes: I've added this now

tynes:
Will rebase `feat/mainnet` on `develop` when we do the next migration rehearsal

tynes:
@mslipper I reverted the commit that adds the `panic` because it will break the tests in a way that would require a large refactor

PR #5702, opened 2023-05-12
feat(pops): `SystemConfig` Attestor Set

clabby:
**Description**

> **Note**
> This PR is the base of the `pops` project PR stack. Do not merge it yet, we are feature frozen and this project introduces changes to existing contracts. It also replaces #5696 as the bottom of the following stack:

```
└─   `feat/pops-devnet` [DEVNET MVP BRANCH, DEPENDS ON ALL BELOW BRANCHES]
 └─  `feat/pops-proposer` [`op-proposer` changes]
  └─  `feat/pops-l2oo` [`L2OutputOracle` changes]
   └─  `feat/pops-contracts` [New contracts: `AttestationDisputeGame`, `DisputeGameFactory`, etc]
👉  └─  `feat/pops-system-config-attestor-set` [Adds an attestor set maintained by the `SystemConfig`]
     └─  `develop`
```

This PR adds a new attestor set that is maintained by the `SystemConfig` for the Permissionless Output Proposals project.

**Metadata**

Fixes CLI-3986

PR #5704, opened 2023-05-13
feat(pops): POPs contracts

clabby:
**Description**

```
└─   `feat/pops-devnet` [DEVNET MVP BRANCH, DEPENDS ON ALL BELOW BRANCHES]
 └─  `feat/pops-proposer` [`op-proposer` changes]
  └─  `feat/pops-l2oo` [`L2OutputOracle` changes]
👉 └─  `feat/pops-contracts` [New contracts]
    └─  `feat/pops-system-config-attestor-set` [Adds an attestor set maintained by the `SystemConfig`]
     └─  `develop`
```

Adds several smart contracts for the POPs project, namely the `AttestationDisputeGame`. This PR will also contain all modifications to existing contracts in `develop` that associate with POPs, such as the `DisputeGameFactory` & `BondManager` if they are required.

**Metadata**

Fixes CLI-3485


clabby:
* develop
  * **PR #5702** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5702" target="_blank">Graphite</a> 
    * **PR #5704** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5704" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5704?utm_source=stack-comment).

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+            // Store the `DISPUTE_TYPE_HASH`
+            mstore(ptr, disputeTypeHash)
+            // Store the `rootClaim` of the `AttestationDisputeGame`.
+            mstore(add(ptr, 0x20), _rootClaim)
+            // Store the L2 block number that the `rootClaim` commits to.
+            mstore(add(ptr, 0x40), _l2BlockNumber)
+
+            disputeStructHash := keccak256(ptr, 0x60)
+        }
+
```
tynes, line 182: Do we need to update the free memory pointer here? There is used memory past where the free memory pointer is

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+    function version() external pure override returns (string memory) {
+        assembly {
+            // Store the pointer to the string
+            mstore(returndatasize(), 0x20)
+            // Store the version ("0.0.1")
+            // len |   "0.0.1"
+            // 0x05|302E302E31
+            mstore(0x25, 0x05302E302E31)
+            // Return the semantic version of the contract
+            return(returndatasize(), 0x60)
```
tynes, line 209: wen push0

tynes: Not opposed to updating the solc version to 0.8.20 for the new set of contracts, should certainly not do that for the other contracts tho until we are deliberate about it

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
clabby: We should not - we store 96 bytes at the current free memory pointer and hash them, so there will be no references to the data within this memory. It should be completely fine that solc overwrites this memory on the next allocation.

If it'd make you feel better about this, we definitely can though!

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
clabby: Yeah I lean towards keeping 0.8.15 until we've done a mass upgrade. Big push0 fanboy, but introducing another compiler version into our compilation pipeline would be a bit annoying. wdyt?

clabby: (Another note- this function is extra asf. Leaning towards just returning an immutable string lmfao.)

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+     *         necessary parties.
+     * @dev May only be called if the `status` is `IN_PROGRESS`.
+     * @return _status The status of the resolved game.
+     */
+    function resolve() public returns (GameStatus _status) {
+        // Set the status as `CHALLENGER_WINS`.
+        status = GameStatus.CHALLENGER_WINS;
+
+        // Delete all outputs from [l2BlockNumber, currentL2BlockNumber]
+        l2OutputOracle.deleteL2Outputs(l2BlockNumber());
```
tynes, line 226: why not just call `l2BlockNumber` once and put the value on the stack?

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+    function gameType() public pure override returns (GameType) {
+        return GameType.ATTESTATION;
+    }
+
+    /**
+     * @notice Overrides the EIP712 domain information
+     * @return _name The name of the domain.
+     * @return _version The version of the domain.
+     */
+    function _domainNameAndVersion()
```
tynes, line 250: What do you all think about using this eip for retrieving the eip 712 domain?
https://ethereum-magicians.org/t/eip-5267-retrieval-of-eip-712-domain/9951/3

packages/contracts-bedrock/contracts/libraries/Clone.sol
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: BSD
+pragma solidity ^0.8.15;
+
+/**
+ * @title Clone
+ * @author zefram.eth, Saw-mon & Natalie, clabby
+ * @notice Provides helper functions for reading immutable args from calldata
+ * @dev MODIFICATIONS:
```
tynes, line 8: this modifications is only useful relative to something so the original link should be included as well

packages/contracts-bedrock/contracts/test/AttestationDisputeGame.t.sol
```diff
@@ -0,0 +1,245 @@
...
+        systemConfig.transferOwnership(address(this));
+
+        // Add 5 signers to the attestor set
+        for (uint256 i = 1; i < 6; i++) {
+            attestorKeys.push(i);
+            systemConfig.setAttestor(vm.addr(i), true);
+        }
+        systemConfig.setAttestationThreshold(5);
+
+        // {
```
tynes, line 63: dead comments?

clabby: yep- port isn't done yet, @refcell's version used a mock L2OO / SystemConfig, in the process of using the real thing here. Will pull it out of draft when this PR is good to go 💯

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+    /**
+     * @notice The EIP-712 type hash for the `Dispute` struct.
+     */
+    Hash public constant DISPUTE_TYPE_HASH =
+        Hash.wrap(keccak256("Dispute(bytes32 outputRoot,uint256 l2BlockNumber)"));
+
+    /**
+     * @notice The BondManager contract that is used to manage the bonds for this game.
+     */
+    IBondManager public immutable bondManager;
```
tynes, line 38: Immutables are generally snake cased all caps

clabby: ibid ^

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+     *        and `l2BlockNumber` (within the `extraData`) from a key that
+     *        exists within the `attestorSet`.
+     */
+    function challenge(bytes calldata signature) external {
+        // Attempt to recover the signature provided. Solady's ECDSA library
+        // will revert if the signer cannot be recovered from the given signature.
+        address recovered = ECDSA.recoverCalldata(Hash.unwrap(getTypedDataHash()), signature);
+
+        // Check that the recovered address is part of the `attestorSet`.
+        // if (!systemConfig.attestorSet(recovered)) {
```
tynes, line 130: This should only care about looking up in the frozen attestor set right?

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * @notice If all necessary information has been gathered, this function should mark the game
+     *         status as either `CHALLENGER_WINS` or `DEFENDER_WINS` and return the status of
+     *         the resolved game. It is at this stage that the bonds should be awarded to the
+     *         necessary parties.
+     * @dev May only be called if the `status` is `IN_PROGRESS`.
+     * @return _status The status of the resolved game.
+     */
+    function resolve() public returns (GameStatus _status) {
```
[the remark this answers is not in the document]
clabby: This function is left `public` to adhere to the `IDisputeGame` interface. Not all games will be implicitly resolved within `challenge`, such as the fault game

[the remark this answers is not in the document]
tynes: Ok thats fair, I think we need to ensure that this can only be called once tho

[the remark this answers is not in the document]
clabby: Yep- onlyOnce logic not yet implemented for this version :)

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
tynes: I suppose its safe as long as other safe code is written, there exists code that this would create a bug for.

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,292 @@
...
+    /**
+     * @notice The EIP-712 type hash for the `Dispute` struct.
+     */
+    Hash public constant DISPUTE_TYPE_HASH =
+        Hash.wrap(keccak256("Dispute(bytes32 outputRoot,uint256 l2BlockNumber)"));
+
+    /**
+     * @notice The current version of the `AttestationDisputeGame` contract.
+     */
+    string public constant VERSION = "0.0.1";
```
tynes, line 38: I don't think we need a getter for both `version` and `VERSION`. Do you want to move away from `Semver`? Its slightly wasteful, bloating the contract size, but it does enforce standardization

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
clabby: Got it- not worth keeping track of the possible footgun for 96 bytes of memory expansion, will update the pointer.

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
clabby: I agree. This is only to adhere to the interface, which we'd be better off updating.

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+    function resolve() public returns (GameStatus _status) {
+        // Set the status as `CHALLENGER_WINS`.
+        status = GameStatus.CHALLENGER_WINS;
+
+        // Delete all outputs from [l2BlockNumber, currentL2BlockNumber]
+        l2OutputOracle.deleteL2Outputs(l2BlockNumber());
+
+        // Request the `BondManager` to distribute the faulty output bond to the attestors.
+        uint256 _l2BlockNumber = l2BlockNumber();
+        bondManager.seizeAndSplit(keccak256(abi.encode(_l2BlockNumber)), attestationSubmitters);
```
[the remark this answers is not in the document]
tynes: I want to be sure that there are no denial of service sort of attacks here where an attacker can prevent a faulty output from being disputed. This means that the bond id must be reusable. The condition is: 1) invalid output proposed, 2) invalid output disputed successfully, 3) invalid output proposed again at same block number

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
tynes: We could turn `Semver` into an abstract contract in future releases and move away from the current immutable style

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        // Mark that the authorized signer has issued a positive attestation for the `rootClaim`.
+        challenges[recovered] = true;
+
+        // Increment the number of positive attestations that have been issued for the `rootClaim`.
+        attestationSubmitters.push(msg.sender);
+
+        // If the provided signature breaches the signature threshold, resolve the game.
+        if (attestationSubmitters.length == frozenSignatureThreshold) {
```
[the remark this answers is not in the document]
refcell: Done ✅

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
refcell: Resolving since solady gives eip 5267 to us for free: https://github.com/Vectorized/solady/blob/main/src/utils/EIP712.sol#L119-L144

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
```diff
@@ -0,0 +1,297 @@
...
+            let ptr := mload(0x40)
+
+            // Store the `DISPUTE_TYPE_HASH`
+            mstore(ptr, disputeTypeHash)
+            // Store the `rootClaim` of the `AttestationDisputeGame`.
+            mstore(add(ptr, 0x20), _rootClaim)
+            // Store the L2 block number that the `rootClaim` commits to.
+            mstore(add(ptr, 0x40), _l2BlockNumber)
+
+            disputeStructHash := keccak256(ptr, 0x60)
```
clabby, line 189:
```suggestion
            disputeStructHash := keccak256(ptr, 0x60)
            
            // Update the free memory pointer
            mstore(0x40, and(0x7F, not(0x1F))
```

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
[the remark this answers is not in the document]
refcell: This is a test that needs to happen in the l2oo branch cc @clabby - once you submit the branch, i can do this.

packages/contracts-bedrock/contracts/dispute/AttestationDisputeGame.sol
refcell: Marked internal, not going to introduce `Semver` changes to avoid a larger frozen code diff.

PR #5707, opened 2023-05-14
refactor(proxyd): clean up backend rate limiter

felipe-op:
**Description**

This change refactors proxyd removing backend rate limiter, an un-used feature.

**Tests**

Removed associated tests, updated existing tests to comply with new behavior.

**Invariants**

n/a

**Additional context**

**Metadata**

Fixes:
- https://linear.app/optimism/issue/INF-185/low-priority-clean-up-backend-rate-limiter

**TODOs**


proxyd/backend.go
```diff
@@ -340,91 +328,19 @@ func (b *Backend) Forward(ctx context.Context, reqs []*RPCReq, isBatch bool) ([]
...
 
-	b.setOffline()
 	return nil, wrapErr(lastError, "permanent error forwarding request")
 }
 
 func (b *Backend) ProxyWS(clientConn *websocket.Conn, methodWhitelist *StringSet) (*WSProxier, error) {
-	if !b.Online() {
-		return nil, ErrBackendOffline
-	}
-	if b.IsWSSaturated() {
```
mslipper, line 351: I'm fine removing this for now, but I do see a world where we may need the ability to limit concurrent WS connections. MEV bots will make thousands of connections to a single upstream to try and spam requests.

proxyd/server.go
```diff
@@ -586,6 +588,14 @@ func (s *Server) populateContext(w http.ResponseWriter, r *http.Request) context
 	)
 }
 
+func randStr(l int) string {
```
mslipper, line 591: Is this used anywhere?

mslipper (approved):

proxyd/backend.go
felipe-op: I'm sure we'll have to do some refactoring in WS code when we re-enable it. At that time, we can move to an simple atomic integer or the new sliding window impl to track the current number of conns.

proxyd/server.go
felipe-op: Yeah, `populateContext` uses this to generate a `req_id`. It used to live in the file `proxyd/backend_rate_limiter.go` what this PR deleted, so it have to be moved somewhere. I picked `server.go` since its where it is currently used.

PR #5708, opened 2023-05-14
fix(proxyd): add missing shutdown for consensus poller

felipe-op:
**Description**

This change adds a missing shutdown to the consensus poller on service termination.

**Tests**

Manual

**Invariants**

n/a

**Additional context**

**Metadata**

Fixes:
https://linear.app/optimism/issue/INF-222/fix-missing-consensus-poller-shutdown

**TODOs**


mslipper (approved):

proxyd/server.go
```diff
@@ -222,6 +222,11 @@ func (s *Server) Shutdown() {
 	if s.wsServer != nil {
 		_ = s.wsServer.Shutdown(context.Background())
 	}
+	for _, bg := range s.BackendGroups {
+		if bg.Consensus != nil {
```
mslipper, line 226: Do we expect to have to perform additional shutdown procedures on a `BackendGroup` in the future? If so, might be better to add a `Shutdown()` method on the group itself that checks for the internal `Consensus` object and shuts it down. Then you can call `Shutdown()` every time without the additional check.

mslipper (approved):

proxyd/server.go
felipe-op: Yeah, this is a convenient nit. I'll merge this PR and move the backend shutdown logic to its own method.

PR #5711, opened 2023-05-15
contracts-bedrock: deploy time checks

tynes:
**Description**

Add deploy time checks to ensure that the config invariants are sane. There is an invariant such that the L2 gas limit must be larger than the amount of gas used by the system transaction plus the amount of gas that deposits can purchase on L1. This invariant is held checked in the migration code but not the typescript deployment code.

Closes CLI-3774

### DIFF
--- a/packages/contracts-bedrock/src/constants.ts
+++ b/packages/contracts-bedrock/src/constants.ts
@@ -1,3 +1,5 @@
+import { ethers } from 'ethers'
+
 /**
  * Predeploys are Solidity contracts that are injected into the initial L2 state and provide
  * various useful functions.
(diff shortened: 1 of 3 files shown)


packages/contracts-bedrock/src/constants.ts
```diff
@@ -26,3 +28,14 @@ export const predeploys = {
   BaseFeeVault: '0x4200000000000000000000000000000000000019',
   L1FeeVault: '0x420000000000000000000000000000000000001a',
 }
+
+const uint128Max = ethers.BigNumber.from('0xffffffffffffffffffffffffffffffff')
```
clabby, line 32:
```suggestion
const uint128Max = ethers.constants.MaxUint256.mask(128)
```

clabby (approved):
LGTM, left 1 suggestion. Feel free to ignore it if you'd rather hard code uint128 max.

packages/contracts-bedrock/src/constants.ts
tynes: Not super opinionated here

PR #5712, opened 2023-05-15
fix(op-node): Remove Deprecated Public rpc.Api Field

refcell:
**Description**

Removes the deprecated `Public` field in go-ethereum `rpc.Api` instantiation.

Reference: https://github.com/ethereum/go-ethereum/pull/25059


tynes:
Comment from `go-ethereum`

> 	Public        bool        // deprecated - this field is no longer used, but retained for compatibility

tynes (approved):

PR #5713, opened 2023-05-15
feat(op-node): SystemConfig graceful op-node error handling

refcell:
**Description**

Prevent the op-node from erroring on version0 `SystemConfig` unrecognized update types.

**Why?**

This allows us to add update types for the system config that do not affect derivation in version 0 without having the op-node error.

For example, in the attestation dispute game, we want to have a system-wide set of "attestors" be defined in the `SystemConfig` contract. We want to emit events when this set of attestors is updated but don't need the rollup derivation pipeline to pick up these config updates during L1 traversal.

**TODO**

- [ ] Consideration on if this is an acceptable change or we need to explicitly add to the next hardfork. Tradeoffs, tradeoffs...


trianglesphere (commented):
This cleanup looks good, but I have one comment about making sure that the behavior does not change.

Looking closely at the error handling, it's a little inconsistent, but in the L1 traversal stage, we return a critical error if there is an error so we should probably keep that behavior.

refcell:
Closing since we decided to keep this behavior.

PR #5714, opened 2023-05-15
Revert "fix(op-node): Remove Deprecated Public rpc.Api Field"

trianglesphere:
Reverts ethereum-optimism/optimism#5712

I believe that this breaking hive. Need to test to be sure.


tynes:
if this is breaking things then its probably because an older version is being used in hive

trianglesphere:
This is no the PR which cause the failure.

PR #5719, opened 2023-05-16
feat(ctp): Change the type on the auth id field of the Faucet contract to bytes32

tremarkley:
Having the auth id type as `bytes` was causing an issue when using the ethers.js [`_signTypedData`](https://docs.ethers.org/v5/api/signer/#Signer-signTypedData) method to generate an EIP signed message containing the id. The issue was that when recovering the signed message within the `AdminFaucetAuthModule` it would return the incorrect address for the signer of the message. Changing the auth id type to `bytes32` resolves this issue.


smartcontracts (approved):
LGTM. I think this approach is fine because if the ID is longer than 32 bytes we can always just hash it, shove the original into the proof, and verify the hash.

PR #5721, opened 2023-05-16
migration: cleanup rehearsal artifacts

tynes:
**Description**

Cleans up the migration artifacts so that `feat/mainnet` can be merged into `develop`


op-node/chaincfg/chains.go
```diff
@@ -43,36 +43,44 @@ var Mainnet = rollup.Config{
 var Goerli = rollup.Config{
 	Genesis: rollup.Genesis{
 		L1: eth.BlockID{
```
maurelian, line 45: I think you edited the wrong network here.

tynes: Oh yeah i totally did, good catch

ajsutton (approved):

PR #5722, opened 2023-05-16
op-program: Remove unused L1 fetcher

ajsutton:
**Description**

Removes the unused `L1FetchingOracle`. Fetching happens only in the host.


protolambda:
LGTM, can we do the same for `FetchingL2Oracle` (`rm -rf op-program/host/l2`)?

ajsutton:
> LGTM, can we do the same for `FetchingL2Oracle` (`rm -rf op-program/host/l2`)?

Yep good point.  I've done that too and rebased.

trianglesphere (approved):

PR #5726, opened 2023-05-17
contracts-bedrock: mainnet deploy config fee vault recipients

tynes:
**Description**

Update the fee vault recipients to their final value before the upgrade. The value can be changed through an L2 contract upgrade. The values were previously set to a hardhat test account. We never need to recycle the ether back to the test accounts because they already have so much ether that it is fine to set it to its final value.


maurelian (approved):
I verified this address with the right parties at OP Labs.

PR #5732, opened 2023-05-18
op-node: Add persistence of peer scores in peer store

ajsutton:
**Description**

Introduces the concept of `ExtendedPeerstore` which provides a type-safe API for storing additional peer data - initially peer scores from various scoring sources.

Currently it doesn't actually use the information in the store but does update the store periodically with the latest gossip score for each peer. The extended peer store is also passed into the method that creates the connection gater to prove the plumbing all fits together but the gater hasn't been updated to actually use it.

Also simplifies the setup initialisation of the p2p layer by removing the connection gater and manager constructors from the config. These are both created as part of the `config.Host` call anyway so the config has control over them.

Currently scores are stored as meta entries for peers. This is simple but overly naive.  The underlying `datastore.Batching` has been passed into the constructor to ensure that it is available when being wired up as an optimised implementation will likely use that directly. We will have to implement some form of GC if we move out of using meta (but that will give us flexibility on how long to retain scoring information which is likely to be required anyway).

Also pretty sure the API to get scores isn't the most useful thing. For disconnection we'll want to be able to get a total score value out rather than individual components, but not sure the peer store should be involved in applying weightings to calculate that score.

**Tests**

**Invariants**

**Additional context**

**Metadata**

- https://linear.app/optimism/issue/CLI-3246/persist-peer-scores

### DIFF
--- a/op-e2e/setup.go
+++ b/op-e2e/setup.go
@@ -465,6 +465,8 @@ func (cfg SystemConfig) Start(_opts ...SystemConfigOption) (*System, error) {
 			if p, ok := p2pNodes[name]; ok {
 				return p, nil
 			}
+			// TODO: This will need to be modified to use sys.Mocknet.AddPeerWithPeerstore so we can make the peerstore
+			// an extended peerstore.
 			h, err := sys.Mocknet.GenPeer()
 			if err != nil {
 				return nil, fmt.Errorf("failed to init p2p host for node %s", name)
--- a/op-node/p2p/config.go
+++ b/op-node/p2p/config.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"time"
 
+	"github.com/ethereum-optimism/optimism/op-node/p2p/store"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/p2p/discover"
 	"github.com/ethereum/go-ethereum/p2p/enode"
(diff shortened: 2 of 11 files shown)


ajsutton:
Split out the simplification of init code to https://github.com/ethereum-optimism/optimism/pull/5733

op-e2e/setup.go
```diff
@@ -465,6 +465,8 @@ func (cfg SystemConfig) Start(_opts ...SystemConfigOption) (*System, error) {
 			if p, ok := p2pNodes[name]; ok {
 				return p, nil
 			}
+			// TODO: This will need to be modified to use sys.Mocknet.AddPeerWithPeerstore so we can make the peerstore
+			// an extended peerstore.
 			h, err := sys.Mocknet.GenPeer()
```
ajsutton, line 470: This was an unfortunate discovery.  Switching to specifying the peerstore ourselves will require duplicating a bit of code that `GenPeer` previously did.  Alternatively we could introduce the `ExtendedPeerStore` wrapper later rather than it being the actual host `Peerstore` but it feels a lot nicer to have everything go through the wrapper.

ajsutton: I updated the libp2p PR to so that GenPeer can take a PeerStore and/or ConnectionManager so we can duplicate code in the short term but hopefully migrate to those new APIs in a future go-libp2p version.

ajsutton:
Done in separate PRs.

PR #5733, opened 2023-05-18
op-node: Remove ConnGater and ConnMngr constructors from p2p config

ajsutton:
**Description**

Removes the ConnGater and ConnMngr constructors from p2p config. The config was calling these itself from the Host method and nothing actually used the ability to further customize them.

**Tests**

No actual change in behaviour, just simplified init process.


trianglesphere (commented):
Changes seem fine. Note that hive p2p test failed. Rerunning to see if it was a flake or not.

trianglesphere (approved):
P2P hive test passed.

PR #5736, opened 2023-05-18
op-e2e: pending block test + op-geth update

protolambda:
**Description**

Tests that the pending block matches the latest block, a change introduced in https://github.com/ethereum-optimism/op-geth/pull/93

~~Added a temporary `go.mod` change so it runs CI with that op-geth PR. We should merge the geth PR first, and then update the go.mod to point at the main branch.~~ Matches latest `optimism` branch op-geth commit now (previous v1.11.6 changes were included before this PR already, and don't really affect the op-node other than e2e testing, this strictly introduces the receipt L1-fee overhead metadata fix and the pending-block change).

### DIFF
--- a/op-e2e/system_test.go
+++ b/op-e2e/system_test.go
@@ -241,12 +241,14 @@ func TestPendingGasLimit(t *testing.T) {
 	cfg.GethOptions["sequencer"] = []GethOption{
 		func(ethCfg *ethconfig.Config, nodeCfg *node.Config) error {
 			ethCfg.Miner.GasCeil = 10_000_000
+			ethCfg.Miner.RollupComputePendingBlock = true
 			return nil
 		},
 	}
 	cfg.GethOptions["verifier"] = []GethOption{
 		func(ethCfg *ethconfig.Config, nodeCfg *node.Config) error {
 			ethCfg.Miner.GasCeil = 9_000_000
+			ethCfg.Miner.RollupComputePendingBlock = true
 			return nil
 		},
 	}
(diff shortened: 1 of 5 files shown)


op-e2e/system_test.go
```diff
@@ -1371,3 +1371,43 @@ func latestBlock(t *testing.T, client *ethclient.Client) uint64 {
...
+
+	cfg := DefaultSystemConfig(t)
+	sys, err := cfg.Start()
+	require.Nil(t, err, "Error starting up system")
+	defer sys.Close()
+	l2Seq := sys.Clients["sequencer"]
+
+	t.Run("block", func(t *testing.T) {
+		for i := 0; i < 10; i++ {
+			pending, err := l2Seq.BlockByNumber(context.Background(), big.NewInt(-1))
```
tynes, line 1387: We should use `rpc.PendingBlockNumber` over `-1`, based on https://github.com/ethereum/go-ethereum/pull/27219 it seems like the numbers do not have a guarantee of stability

protolambda: I tried that, but the server-side magic numbers do not match the client-side magic numbers.

trianglesphere: Do the numbers match after that PR switches them? This looks like we will have to change this when upstream changes.

op-e2e/system_test.go
```diff
@@ -1371,3 +1373,43 @@ func latestBlock(t *testing.T, client *ethclient.Client) uint64 {
...
+			require.NoError(t, err)
+			latest, err := l2Seq.BlockByNumber(context.Background(), nil)
+			require.NoError(t, err)
+			if pending.NumberU64() == latest.NumberU64() {
```
ajsutton, line 1393: Wouldn't this pass if the sequencer completed building the block between when we requested the pending block and requested the latest block?

Feels like these tests would be better in `op_geth_test` so they're controlling geth to avoid these kinds of race conditions.  Otherwise if we break this functionality in the future the test will fail intermittently (probably most of the time to be fair) rather than being guaranteed to fail.

op-e2e/system_test.go
ajsutton: Probably worth declaring it as a constant/var locally just for clarity of purpose if we can't use an existing constant.

The only real risk is if -1 winds up meaning `latest` at some point in the future but I suspect that would cause a lot of grief for anyone using the `ethclient` module given it does indeed use hard coded `-1` when doing its translation and then it's actually `pending` that gets sent in the RPC request.

ajsutton (commented):
I'm not against this as-is but I worry a bit about the race conditions involved with assertions about the pending block when we aren't controlling block building timing.

protolambda:
- rebased on develop
- updated op-geth reference to latest
- updated e2e tests for receipt metadata fields

Will look into adding the additional geth tests now, but think we can also choose to merge with current tests as-is if there is any urgency, since it covers the pending-block behavior, and we also already did an internal devnet test (see notion doc).

protolambda:
- Added test for pending block result at every stage of Engine API
- Added TODOs for every usage of -1 as pending-block identifier in ethclient. Once we update op-geth with upstream changes we can cleanly import the geth rpc label vars.

op-e2e/system_test.go
protolambda: Added op-geth test

op-e2e/system_test.go
protolambda: Added TODOs with reference to ticket of next geth update, which would include this upstream change.

trianglesphere (approved):
lgtm

PR #5737, opened 2023-05-18
feat(op-node): P2P Subcommand for Requesting an ENR

refcell:
**Description**

Introduces a p2p subcommand `reqenr` that requests an ENR from a local, ephemeral discovery service.

### DIFF
--- a/op-node/cmd/p2p/req_enr.go
+++ b/op-node/cmd/p2p/req_enr.go
@@ -9,7 +9,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/urfave/cli"
+
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/p2p/discover"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/ethereum/go-ethereum/p2p/enr"
 	"github.com/ethereum/go-ethereum/rlp"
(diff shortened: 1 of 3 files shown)


op-node/cmd/p2p/req_enr.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	log.Debug("Found active tcp port", "tcpPort", tcpPort)
+
+	// Notify of any new connections/streams/etc.
+	host.Network().Notify(rollupP2P.NewNetworkNotifier(log.Root(), nil))
+	log.Info("Setup new network notifier")
+
+	// All nil if disabled.
+	rollupCfg := &rollup.Config{
+		L2ChainID: big.NewInt(100),
```
tynes, line 100: The chain id should be configurable?

op-node/cmd/p2p/req_enr.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	_, dv5Udp, err := p2pConfig.Discovery(log.New("p2p", "discv5"), rollupCfg, tcpPort)
+	if err != nil {
+		return fmt.Errorf("failed to start discv5: %w", err)
+	}
+
+	log.Debug("Discv5 started, requesting enr...")
+
+	for {
```
tynes, line 109: this entire `for` loop could go in a goroutine, then we could block like this
https://github.com/ethereum-optimism/optimism/blob/fc56ded13c91ae945f7e35bff0e45f9154fd647e/op-batcher/batcher/batch_submitter.go#L102

op-node/cmd/p2p/req_enr.go
```diff
@@ -0,0 +1,206 @@
...
+	log.Info("dialing static peer", "peer", peerId)
+	if _, err := net.DialPeer(ctx, peerId); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ParseENR parses an enr string into a [wrappedEnr].
+func ParseENR(e string, validSchemes enr.IdentityScheme) (*enode.Node, error) {
+	e = e[len(enrPrefix):]
```
tynes, line 192: could this panic? its untrusted input right?

tynes:
I'm slightly confused what this does, it requests an enr but you pass the enr as an argument? If this used for interop with magi?

PR #5738, opened 2023-05-18
feat(op-challenger): L2OutputOracle Filter Query.

refcell:
**Description**

Split out of #5636, this adds a simple tested function for building a filter query for the `L2OutputOracle` `OutputProposed` event.

### DIFF
--- a/op-challenger/challenger/oracle.go
+++ b/op-challenger/challenger/oracle.go
@@ -1,0 +1,36 @@
+package challenger
+
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var ErrMissingEvent = errors.New("missing event")
+
+// BuildOutputLogFilter creates a filter query for the L2OutputOracle contract.
+//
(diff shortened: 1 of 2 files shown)


op-challenger/challenger/oracle.go
```diff
@@ -0,0 +1,28 @@
...
+// BuildOutputLogFilter creates a filter query for the L2OutputOracle contract.
+//
+// The `OutputProposed` event is encoded as:
+// 0: bytes32 indexed outputRoot,
+// 1: uint256 indexed l2OutputIndex,
+// 2: uint256 indexed l2BlockNumber,
+// 3: uint256 l1Timestamp
+func BuildOutputLogFilter(l2ooABI *abi.ABI) (eth.FilterQuery, error) {
+	// Listen for `OutputProposed` events from the L2 Output Oracle contract
+	event := l2ooABI.Events["OutputProposed"]
```
tynes, line 20: Could we add a check that `event` isn't `nil` or the zero value and return an error if so? This would catch a change in the ABI without a `panic`

refcell: Ah good catch! I actually thought about this but must have just forgotten to impl, thanks.

ajsutton (approved):

PR #5746, opened 2023-05-19
op-node: Add extended peer store to store scores

ajsutton:
**Description**

Starts building an extended peer store that supports persisting peer scores. Not currently hooked up and lacking pruning of old entries, but defines the interface and has a sensible implementation for persisting scores.

**Tests**

Unit tests for persistence functions.

**Invariants**

**Metadata**

- https://linear.app/optimism/issue/CLI-3246/persist-peer-scores

### DIFF
--- a/op-node/p2p/store/scorebook.go
+++ b/op-node/p2p/store/scorebook.go
@@ -21,8 +21,6 @@ type scoreBook struct {
 
 var scoresBase = ds.NewKey("/peers/scores")
 
-type ScoreType string
-
 const (
 	scoreDataV0    = "0"
 	scoreCacheSize = 100
@@ -65,14 +63,19 @@ func (d *scoreBook) getPeerScoresNoLock(id peer.ID) (PeerScores, error) {
 	return scores, nil
 }
 
-func (d *scoreBook) SetGossipScore(id peer.ID, score float64) error {
+func (d *scoreBook) SetScore(id peer.ID, scoreType ScoreType, score float64) error {
 	d.Lock()
 	defer d.Unlock()
 	scores, err := d.getPeerScoresNoLock(id)
 	if err != nil {
 		return err
 	}
-	scores.Gossip = score
+	switch scoreType {
+	case TypeGossip:
+		scores.Gossip = score
+	default:
+		return fmt.Errorf("unknown score type: %v", scoreType)
+	}
 	data, err := serializeScoresV0(scores)
 	if err != nil {
 		return fmt.Errorf("encode scores for peer %v: %w", id, err)
--- a/op-node/p2p/store/scorebook.go
+++ b/op-node/p2p/store/scorebook.go
@@ -1,0 +1,90 @@
+package store
+
+import (
+	"context"
+	"errors"
+	"fmt"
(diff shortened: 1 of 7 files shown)


op-node/p2p/store/scorebook.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	scores, err = deserializeScoresV0(data)
+	if err != nil {
+		return PeerScores{}, fmt.Errorf("invalid score data for peer %v: %w", id, err)
+	}
+	d.cache.Add(id, scores)
+	return scores, nil
+}
+
+func (d *scoreBook) SetGossipScore(id peer.ID, score float64) error {
```
protolambda, line 68: I think we should define a Go enum (typed int, iota definition) and change this to `SetScore(id peer.ID, scoreType ScoreType, score float64) error` and then add a switch to write it to the scores struct. Specifically I'd like the gossip-score to be stored with its original parts, so we can see details when we debug why a score is high/low. And this would make adding more non-gossip score types easy.

ajsutton: hah, funnily enough that's the way I originally wrote it and then changed to add more type safety.  Can switch it back.

op-node/p2p/store/serialize_test.go
```diff
@@ -0,0 +1,45 @@
+package store
+
+import (
+	"testing"
+
+	"github.com/status-im/keycard-go/hexutils"
```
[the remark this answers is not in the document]
ajsutton: ah, it should have been, `Hex2Bytes` not `HexToBytes` - sigh.

op-node/p2p/store/serialize.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"bytes"
+	"encoding/binary"
+
+	pool "github.com/libp2p/go-buffer-pool"
+)
+
+func serializeScoresV0(scores PeerScores) ([]byte, error) {
+	var b pool.Buffer
```
protolambda, line 11: If we don't return this buffer `b`, will it return to the global buffer pool (the default when left unspecified)? The docs say it frees memory on Read, but the `.Bytes()` function isn't a read; it exposes the buffered data without copy, and the buffer cannot be returned to the pool safely/automatically (since the returned slice might theoretically be mutated). In general I'm against using `sync.Pool` without comments, because it's easy to miss a `.Put` otherwise. The libp2p buffer pool seems to be the same.

Docs: https://pkg.go.dev/github.com/libp2p/go-buffer-pool#Buffer

protolambda (commented):
I'm going to change it to `bytes.Buffer` so we can merge this PR without delay and without buffer-pool risk, and can maybe land more of the work that you stacked on top. We can always optimize it in a follow-up PR.

protolambda (approved):

op-node/p2p/store/serialize.go
ajsutton: I hope it's safe - I followed the pattern that the metadata store is using but the unpooled version should be fine and is definitely safer so makes sense.

PR #5748, opened 2023-05-19
feat(op-challenger): Challenger types and enum string refactor

refcell:
**Description**

Split out of #5636, this adds types to the op-challenger as well as a refactored `EnumString` op-service utility.


op-challenger/types/game_type_test.go
```diff
@@ -0,0 +1,63 @@
...
+	require.Equal(t, FaultDisputeGameType.String(), d.selected)
+}
+
+// TestGameType_String tests the String function on the DisputeGameType.
+func TestGameType_String(t *testing.T) {
+	d := NewDisputeGameType()
+	require.Equal(t, DefaultGameType().String(), d.String())
+	require.NoError(t, d.Set("fault"))
+	require.Equal(t, FaultDisputeGameType.String(), d.String())
+}
```
ajsutton, line 53: We probably should make this a table test to check each game type returns the right string. It would be easy to add a new game type but forget to add it to the array of game type names.

Ideally this would iterate all the enum values so the test would automatically pick up any new ones, but I'm not sure if that's possible in Golang.  Second best thing would be to have a list of known game types in this test then this and `TestGameType_Valid` could iterate over the same list.

op-challenger/types/game_type_test.go
```diff
@@ -0,0 +1,63 @@
...
+// TestGameType_Valid tests the Valid function with valid inputs.
+func TestGameType_Valid(t *testing.T) {
+	require.True(t, AttestationDisputeGameType.Valid())
+	require.True(t, FaultDisputeGameType.Valid())
+	require.True(t, ValidityDisputeGameType.Valid())
+}
+
+// TestGameType_Invalid tests the Valid function with an invalid input.
+func TestGameType_Invalid(t *testing.T) {
+	require.False(t, GameType(255).Valid())
```
ajsutton, line 23: I'd probably use `ValidityDisputeGameType + 1` here rather than 255. Then if someone adds a new type this test will fail. Even better if you have the list of valid game types I mention in the other comment it could just use the last entry in that + 1 so it will force people to update the list of game types and the new type will be covered by the other tests automatically.

op-challenger/types/game_type.go
```diff
@@ -0,0 +1,86 @@
...
+func (d DisputeGameType) Type() GameType {
+	if d.selected == DisputeGameTypes[0] {
+		return AttestationDisputeGameType
+	} else if d.selected == DisputeGameTypes[1] {
+		return FaultDisputeGameType
+	} else if d.selected == DisputeGameTypes[2] {
+		return ValidityDisputeGameType
+	} else {
+		return DefaultGameType()
+	}
```
ajsutton, line 85: Couldn't we just make `d.selected` be a `GameType` and return  it directly rather than storing it as a string?  `Set` is already validating the provided string so we'd wind up with:

```
// Set sets the dispute game type.
func (d *DisputeGameType) Set(value string) error {
	for i, enum := range d.Enum {
		if enum.String() == value {
			d.selected = GameType(i)
			return nil
		}
	}

	return fmt.Errorf("allowed values are %s", enum.EnumString(d.Enum))
}

// String returns the selected dispute game type.
func (d DisputeGameType) String() string {
	return d.selected.String()
}

// Type maps the [DisputeGameType] string value to a [GameType] enum value.
func (d DisputeGameType) Type() GameType {
	return d.selected
}
```

If that doesn't work for some reason, we should add a test because they all pass once you remove a couple of now unnecessary `.String()` calls. :)

It does mean you're only guaranteed to get the default game type if you use `NewDisputeGameType()` rather than constructing the enum directly, though currently the default also happens to be the value 0 so it works either way.

op-challenger/types/game_type_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+// TestGameType_Default tests the default value of the DisputeGameType.
+func TestGameType_Default(t *testing.T) {
+	d := NewDisputeGameType()
+	require.Equal(t, DefaultGameType(), d.selected)
+	require.Equal(t, DefaultGameType(), d.Type())
+}
+
+// TestGameType_String tests the Set and String function on the DisputeGameType.
+func TestGameType_String(t *testing.T) {
```
trianglesphere, line 47: This just seems to be testing the values against itself. If we want to test this, I recommend testing against hardcoded values.

op-challenger/types/game_type.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Valid returns true if the game type is within the valid range.
+func (g GameType) Valid() bool {
+	return g >= AttestationDisputeGameType && g <= ValidityDisputeGameType
+}
+
+// DisputeGameType is a custom flag type for dispute game type.
+type DisputeGameType struct {
+	Enum     []enum.Stringered
+	Default  GameType
```
trianglesphere, line 42: why do we care about the default here?

trianglesphere (commented):
Some of the decisions around game_type are a little hard to understand without the context of how this will be used.

op-challenger/types/game_type_test.go
ajsutton: Yeah I'd say we want a list of the expected GameType->String pairs in this test.  This string test would confirm we get the expected values and `TestGameType_Invalid` would get the highest valid `GameType` value from that list. `TestGameType_Valid` could also then be a table test that iterates over the expected game types and checks they're all valid.

It does introduce some duplication because the game types are declared in two places but the production code version separates the definition of GameType from its string name which introduces the risk of them lining up, whereas the test version would have them paired together very clearly so its obvious in a diff if they aren't lining up properly.

ajsutton: Actually, the other possibility here is that `GameType` should just be a string and would be its own name which avoids any chance of them being out of sync and simplifies a whole bunch of stuff.  Not sure if there's a particular reason we need it to be a uint8 though.

refcell:
> Some of the decisions around game_type are a little hard to understand without the context of how this will be used.

The context of `GameType` can be seen in pr #5636 as i put in the pr description. See your previous comment on that stale pr here: https://github.com/ethereum-optimism/optimism/pull/5636/files#r1188018273

ajsutton (approved):
LGTM.

PR #5749, opened 2023-05-19
Add remappings.txt

maurelian:
**Description**

Just writes the output of `forge remappings` to `remappings.txt` to facilitate easier usage of pyrometer.

### DIFF
--- a/packages/contracts-bedrock/remappings.txt
+++ b/packages/contracts-bedrock/remappings.txt
@@ -1,0 +1,9 @@
+@cwia/=node_modules/clones-with-immutable-args/src/
+@openzeppelin/=node_modules/@openzeppelin/
+@openzeppelin/contracts-upgradeable/=node_modules/@openzeppelin/contracts-upgradeable/
+@openzeppelin/contracts/=node_modules/@openzeppelin/contracts/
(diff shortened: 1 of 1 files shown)


tynes (commented):
The annoying thing is that this could get out of date

packages/contracts-bedrock/remappings.txt
```diff
@@ -0,0 +1,9 @@
+@cwia/=node_modules/clones-with-immutable-args/src/
+@openzeppelin/=node_modules/@openzeppelin/
```
tynes, line 2: Is line 2 a dead remapping?

tynes:
Maybe a npm script for generating this would be useful? Do you think its also too much to fail ci if there is a diff?

maurelian:
> Maybe a npm script for generating this would be useful? Do you think its also too much to fail ci if there is a diff?

I'm open to that. My thought was that I wasn't sure about adding that kind of scaffolding around this as there's no firm commitment to using pyrometer, though there does seem to be enthusiasm. 

I think we should either
1. not do this
2. add the npm script and CI check. 

LMK what you think.

refcell:
I think we should try it out and if we run into pain points with this wrt ci, we can remove.

PR #5750, opened 2023-05-19
feat(op-challenger): Subscription Logic

refcell:
**Description**

Split out of #5636, this adds subscription logic to the op-challenger that will be used to gracefully listen to logs.

A follow-up PR will further refactor this implementation to allow for a full traversal client to iterate all logs (warning/erroring on non-sequential blocks) and feed this back to the log store. Should be parallelizable with the log filter.

### DIFF
--- a/op-challenger/challenger/log_store.go
+++ b/op-challenger/challenger/log_store.go
@@ -108,7 +108,7 @@ func (l *logStore) buildBackoffStrategy() backoff.Strategy {
 // resubscribe attempts to re-establish the log store internal
 // subscription with a backoff strategy.
 func (l *logStore) resubscribe(ctx context.Context) error {
-	l.log.Info("resubscribing")
+	l.log.Info("log store resubscribing with backoff")
 	backoffStrategy := l.buildBackoffStrategy()
 	return backoff.DoCtx(ctx, 10, backoffStrategy, func() error {
 		if l.subscription == nil {
--- a/op-challenger/challenger/log_store_test.go
+++ b/op-challenger/challenger/log_store_test.go
@@ -110,7 +110,7 @@ func TestLogStore_Subscribe_ReceivesLogs(t *testing.T) {
 	}
 	client.logs <- mockLog
 
-	timeout, tCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	timeout, tCancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer tCancel()
 	err := e2eutils.WaitFor(timeout, 500*time.Millisecond, func() (bool, error) {
 		result := logStore.GetLogByBlockHash(mockLog.BlockHash)
@@ -126,7 +126,7 @@ func TestLogStore_Subscribe_SubscriptionErrors(t *testing.T) {
 
 	client.sub.errorChan <- ErrTestError
 
-	timeout, tCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	timeout, tCancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer tCancel()
 	err := e2eutils.WaitFor(timeout, 500*time.Millisecond, func() (bool, error) {
 		subcount := client.subcount == 2
@@ -137,13 +137,10 @@ func TestLogStore_Subscribe_SubscriptionErrors(t *testing.T) {
 }
 
 func TestLogStore_Subscribe_NoClient_Panics(t *testing.T) {
-	defer func() {
-		if recover() == nil {
-			t.Error("expected nil client to panic")
-		}
-	}()
-	logStore, _ := newErrorLogStore(t, nil)
-	require.NoError(t, logStore.Subscribe(context.Background()))
+	require.Panics(t, func() {
+		logStore, _ := newErrorLogStore(t, nil)
+		_ = logStore.Subscribe(context.Background())
+	})
 }
 
 func TestLogStore_Subscribe_ErrorSubscribing(t *testing.T) {
@@ -162,11 +159,8 @@ func TestLogStore_Quit_ResetsSubscription(t *testing.T) {
 }
 
 func TestLogStore_Quit_NoSubscription_Panics(t *testing.T) {
-	defer func() {
-		if recover() == nil {
-			t.Error("expected no subscription to panic")
-		}
-	}()
-	logStore, _ := newErrorLogStore(t, nil)
-	logStore.Quit()
+	require.Panics(t, func() {
+		logStore, _ := newErrorLogStore(t, nil)
+		logStore.Quit()
+	})
 }


refcell:
* develop
  * **PR #5750** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5750" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5750?utm_source=stack-comment).

op-challenger/challenger/log_store.go
```diff
@@ -0,0 +1,109 @@
...
+				l.log.Info("resubscribing to subscription", "id", subId)
+				err := subscription.Subscribe()
+				if err == nil {
+					break
+				}
+			}
+		case log := <-subscription.logs:
+			l.mu.Lock()
+			l.logList = append(l.logList, log)
+			l.logMap[log.BlockHash] = log
```
ajsutton, line 102: What if there's more than one log event in a single block?

op-challenger/challenger/log_store.go
```diff
@@ -0,0 +1,109 @@
...
+		select {
+		case err := <-subscription.sub.Err():
+			l.log.Error("log subscription error", "err", err)
+			for {
+				l.log.Info("resubscribing to subscription", "id", subId)
+				err := subscription.Subscribe()
+				if err == nil {
+					break
+				}
+			}
```
ajsutton, line 98: Not sure we want to be in a tight loop retrying the subscription here. Probably need a backoff strategy in some form.

op-challenger/challenger/log_store_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+// TestLogStore_NewLogStore tests the NewLogStore method on a [logStore].
+func TestLogStore_NewLogStore(t *testing.T) {
+	query := ethereum.FilterQuery{}
+	logStore := NewLogStore(query)
+	require.Equal(t, query, logStore.query)
+	require.Equal(t, []types.Log{}, logStore.logList)
+	require.Equal(t, make(map[common.Hash]types.Log), logStore.logMap)
+	require.Equal(t, SubscriptionId(0), logStore.currentSubId)
+	require.Equal(t, make(map[SubscriptionId]Subscription), logStore.subMap)
```
ajsutton, line 22: These tests are probably too tied to the specifics of the implementation rather than testing required functionality. We want the tests to focus on the externally visible behaviour and be as decoupled as possible from the implementation.  The key behaviours of the log store would be:
* `Spawn` creates a subscription and adds any logs received from it to a list of logs
* When an subscription reports an error, it gets resubscribed.
* `Quit` exits all the go routines listening to subscriptions (not sure how to test this...)

The points of visibility we have are the mock client we pass in and the logs that come back out.  So for reading the logs I think we wind up with a test like:
```Golang
func TestReadLogsFromSubscription(t *testing.T) {
	query := ethereum.FilterQuery{}
	logStore := NewLogStore(query)
	// TODO: Pass client into NewLogStore instead of setting internal variable directly.
	client := &mockLogFilterClient{}
	logStore.client = client

	// Create the new subscription.
	err := logStore.Spawn()
	require.NoError(t, err)
	sub := client.LastSubscription()
	require.NotNil(t, sub)
	l := types.Log{BlockHash: common.Hash{0xaa}}
	sub.logs <- l

	timeout, tCancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer tCancel()
	err = e2eutils.WaitFor(timeout, 500*time.Millisecond, func() (bool, error) {
		result, ok := logStore.GetLogByBlockHash(l.BlockHash)
		return ok && result.BlockHash == l.BlockHash, nil
	})
	require.NoError(t, err)
}
```

https://github.com/ethereum-optimism/optimism/commit/9a91c8e4d6022ec1eb196b01104d0b93d794c66a has all the changes I made to support that - the mock client needs to track the subscription calls for example.

I don't really like the fact that the waiter functions are in `e2eutils` - they're useful for any multithreaded test so possibly should be moved out of op-e2e but I'm not really sure where they should live currently.

refcell: Attempted to clean up the tests. Will re-request your review here since a lot has changed if that's ok :)

op-challenger/challenger/log_store_test.go
```diff
@@ -0,0 +1,225 @@
...
+	// Send logs through the subscription.
+	blockHash := common.HexToHash("0x1")
+	logStore.subscription.logs <- types.Log{
+		BlockHash: blockHash,
+	}
+
+	// Wait for the logs to be dispatched.
+	timeout, tCancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer tCancel()
+	err = e2eutils.WaitFor(timeout, 500*time.Millisecond, func() (bool, error) {
```
trianglesphere, line 172: Very low timeout.

trianglesphere (commented):
Code seems fine, but it's hard to review b/c I'm missing lots of context on when a `logStore` vs `Subscription` should be used & what the requirements for these components are.

op-challenger/challenger/log_store.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// resubscribe resubscribes to the log store subscription with a backoff.
+func (l *logStore) resubscribe() error {
+	l.log.Info("resubscribing to subscription", "id", l.subscription.ID())
+	ctx := context.Background()
+	backoffStrategy := l.buildBackoffStrategy()
+	return backoff.DoCtx(ctx, 10, backoffStrategy, func() error {
+		if l.subscription == nil {
+			l.log.Error("subscription zeroed out")
```
ajsutton, line 93: Why do we need to worry about `subscription` being `nil`? It's set in the constructor and never changed. If other code is reaching in and messing with our internal state panicking is probably the right thing to do.

op-challenger/challenger/log_store_test.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// TestLogStore_Subscribe_MissingClient tests the [Subscribe] method on a [logStore]
+// fails when the client is missing.
+func TestLogStore_Subscribe_MissingClient(t *testing.T) {
+	query := ethereum.FilterQuery{}
+	log := testlog.Logger(t, log.LvlError)
+	logStore := NewLogStore(query, nil, log)
+	err := logStore.Subscribe()
+	require.EqualError(t, err, ErrMissingClient.Error())
```
ajsutton, line 87: I suspect panicking is a valid thing to do if someone passing a `nil` client.

op-challenger/challenger/log_store_test.go
ajsutton: I think that 500ms is how often to run the verification and the timeout is actually the 30s on the context.

op-challenger/challenger/log_store_test.go
```diff
@@ -0,0 +1,225 @@
...
+	client := newMockLogStoreClient()
+	log := testlog.Logger(t, log.LvlError)
+	logStore := NewLogStore(query, client, log)
+
+	// Subscribe to the logStore.
+	err := logStore.Subscribe()
+	require.NoError(t, err)
+
+	// Dispatch logs on the logStore.
+	go logStore.dispatchLogs()
```
ajsutton, line 195: shouldn't this be using `logStore.Start()`?

op-challenger/challenger/log_store_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// Send an error through the subscription.
+	client.sub.errorChan <- errors.New("test error")
+	time.Sleep(1 * time.Second)
```
ajsutton, line 199: Need to avoid arbitrary sleeps - there's no guarantee that async tasks will complete in 1s - CI servers can be surprisingly busy and unpredictable. We'd either need the stub subscription to expose a channel that reports reconnections so we can use that to synchronise or we'd need to use a waiter to wait until the desired effect happens.

op-challenger/challenger/log_store.go
```diff
@@ -0,0 +1,146 @@
...
+	l.logList = append(l.logList, log)
+	l.logMap[log.BlockHash] = append(l.logMap[log.BlockHash], log)
+	l.mu.Unlock()
+}
+
+// GetLogs returns all logs in the log store.
+func (l *logStore) GetLogs() []types.Log {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	return l.logList
```
ajsutton, line 116: This isn't thread-safe because the reference to `logList` is returned and can be used by the caller after the lock is released.  Would need to return a copy of the list but worth asking if getting all logs at once is a requirement.

op-challenger/challenger/log_store.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// GetLogByBlockHash returns all logs in the log store for a given block hash.
+func (l *logStore) GetLogByBlockHash(blockHash common.Hash) []types.Log {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	return l.logMap[blockHash]
```
ajsutton, line 123: This will need to return a copy of the log slice as well for thread safety.

ajsutton (commented):
I've left a bunch of comments, but generally the tests are still reaching into implementation details of the log store implementation more than I think they should.  Generally tests should be written using on the public interface of the object because they're trying to test the external behaviour that callers will experience.  Sometimes there's value in testing a private method because it has some particular complexity that's easier to test in isolation but that should be the exception because it ties the test to the specific implementation.

I'm pretty sure in this case the test should only be using the public methods of the log store and it can exercise the behaviour well via that (plus leveraging the stub subscription we inject via the constructor).  The store is actually only doing two things:
1. Adding any logs received from the subscription to the store
2. Reconnecting the subscription when it reports an error with a backoff (though the backoff isn't testable currently because the backoff package doesn't abstract time yet).

Could possibly add that it shuts down and releases resources when `Quit` is called. I think I you work the tests around those scenarios rather than specific methods you'll wind up with better coverage and that tests won't be brittle or require updating if the implementation changes - they'd only fail if the external behaviour changes which is what we want them to warn us about.

op-challenger/challenger/log_store.go
```diff
@@ -0,0 +1,154 @@
...
+		Min:       1000,
+		Max:       20_000,
+		MaxJitter: 250,
+	}
+}
+
+// resubscribe attempts to re-establish the log store internal
+// subscription with a backoff strategy.
+func (l *logStore) resubscribe(ctx context.Context) error {
+	l.log.Info("resubscribing")
```
ajsutton, line 111: nit: Probably need to add a bit more context to this log message to say what its resubscribing to given the log won't show where in code it came from.

op-challenger/challenger/log_store_test.go
```diff
@@ -0,0 +1,172 @@
...
+	logStore, client := newLogStore(t)
+	defer logStore.Quit()
+	require.NoError(t, logStore.Subscribe(context.Background()))
+
+	mockLog := types.Log{
+		BlockHash: common.HexToHash("0x1"),
+	}
+	client.logs <- mockLog
+
+	timeout, tCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
ajsutton, line 113: I'd make these 5 second timeouts more like 30s. They'll only be reached when the test fails and if the server is busy you can get multiple second delays in scheduling (or do weird things with its clock) and we don't want any risk of intermittent failures. While the timeout being reached is extremely low probability, we wind up running a ton of tests each day that do this kind of timeout so even a 0.001% probably can mean it happens annoyingly often.

op-challenger/challenger/log_store_test.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func TestLogStore_Subscribe_NoClient_Panics(t *testing.T) {
+	defer func() {
+		if recover() == nil {
+			t.Error("expected nil client to panic")
+		}
+	}()
+	logStore, _ := newErrorLogStore(t, nil)
+	require.NoError(t, logStore.Subscribe(context.Background()))
```
ajsutton, line 146: Can simplify this to:
```suggestion
	require.Panics(t, func() {
		logStore, _ := newErrorLogStore(t, nil)
		_ = logStore.Subscribe(context.Background())
	})
```

op-challenger/challenger/subscription_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TestSubscription_Subscribe_NilClient_Panics(t *testing.T) {
+	defer func() {
+		if recover() == nil {
+			t.Error("expected nil client to panic")
+		}
+	}()
+	subscription, _ := newSubscription(t, nil)
+	require.NoError(t, subscription.Subscribe())
```
ajsutton, line 53: Can use `require.Panics` here too.

ajsutton (approved):
LGTM. Just a few minor style things that I think are worth tidying up but approving now as the code is good and don't want to block you on timezones.

op-challenger/challenger/log_store_test.go
refcell: Definitely agree, good call

PR #5756, opened 2023-05-22
feat(op-challenger): Full Traversal Client

refcell:
**Description**

Stacked ontop of #5750, this pr introduces a client that traverses all
blocks and scans the block logs to ensure no OutputProposed log can be
missed.

**Metadata**

Fixes CLI-4058


refcell:
* develop
  * **PR #5750** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5750" target="_blank">Graphite</a> 
    * **PR #5756** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5756" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5756?utm_source=stack-comment).

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,238 @@
...
+			return
+		}
+
+		// Fetch the header with a backoff
+		header, err := l.fetchHeaderByNumber(ctx, start)
+		if err != nil {
+			l.log.Error("failed to fetch header", "err", err)
+			return
+		}
+		l.dispatchNewHead(ctx, header, handleLog, false)
```
ajsutton, line 182: I don't think this is going to work out right because it's including all the checks around `lastBlockNumber` etc but we're backfilling and now have multiple threads accessing and updating `lastBlockNumber`.  Given it already knows the backfill is stepping through each block in order, it just needs the end of `dispatchNewHead` that fetches and processes receipts extracted.

Then it could also just retrieve the full block rather than the header to save on RPC calls.

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// NewLogTraversal creates a new log traversal.
+func NewLogTraversal(client TraversalClient, query *ethereum.FilterQuery, log log.Logger) *logTraversal {
+	return &logTraversal{
+		client:          client,
+		query:           query,
+		quit:            make(chan struct{}),
+		log:             log,
+		lastBlockNumber: big.NewInt(0),
```
ajsutton, line 52: I think this is going to iterate over every block from genesis every time we start up which will take days.

ajsutton (commented):
There doesn't appear to be any re-org handling included in this code which is likely to cause problems where it potentially misses some blocks and getting a very inconsistent view of the chain if some tx wind up in different blocks on either side of the reorg.

Handling reorgs could wind up being quite complex here and depending on exactly what's required there's potentially different ways of going about it (possibly even just tracking the finalised block rather than head). Otherwise there will need to be a way to "unprocess" logs when they're re-orged back out which has impacts outside of this client.

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,247 @@
...
+//go:generate mockery --name MinimalEthClient --output ./mocks/
+type MinimalEthClient interface {
+	InfoByHash(ctx context.Context, hash common.Hash) (eth.BlockInfo, error)
+	InfoByNumber(ctx context.Context, number uint64) (eth.BlockInfo, error)
+	FetchReceipts(ctx context.Context, blockHash common.Hash) (eth.BlockInfo, types.Receipts, error)
+	SubscribeNewHead(ctx context.Context, ch chan<- *types.Header) (ethereum.Subscription, error)
+}
+
+// NewMinimalEthClient creates a [sources.EthClient] with a minimal interface.
+func NewMinimalEthClient(client client.RPC, log log.Logger, metrics caching.Metrics, config *sources.EthClientConfig) (MinimalEthClient, error) {
```
trianglesphere, line 49: no need for this - as `*sources.EthClient` satisfies `MinimalEthClient` you can directly replace `NewMinimalEthClient ` with `sources.NewEthClient`. You're following the language idiom of accepting interfaces, but not the idiom of returning structs. In this case there is not really any benefit of returning a interface here.

op-challenger/challenger/traversal.go
trianglesphere: ditto

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+}
+
+// LastBlockNumber returns the last block number that was traversed.
+func (l *logTraversal) LastBlockNumber() *big.Int {
+	return l.lastBlockNumber
+}
+
+// fetchBlockByHash gracefully fetches block info by hash with a backoff.
+func (l *logTraversal) fetchBlockInfoByHash(ctx context.Context, hash common.Hash) (eth.BlockInfo, error) {
```
trianglesphere, line 75: should be part of the client or a wrapper on top of the ethclient instead of being part of the `logTraversal`

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		for _, receipt := range receipts {
+			for _, log := range receipt.Logs {
+				err := handleLog(log)
+				if err != nil {
+					l.log.Error("Failed to handle log", "err", err)
+					return
+				}
+			}
+		}
```
trianglesphere, line 199: this code is repeated & could be labelled `fetchAndProccessBlock`

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// dispatchNewHead dispatches a new head.
+func (l *logTraversal) dispatchNewHead(ctx context.Context, header *types.Header, handleLog func(*types.Log) error, allowCatchup bool) {
+	info, err := l.fetchBlockInfoByHash(ctx, header.Hash())
+	if err != nil {
+		l.log.Error("Failed to fetch block", "err", err)
+		return
+	}
+	expectedBlockNumber := l.lastBlockNumber.Add(l.lastBlockNumber, big.NewInt(1))
```
trianglesphere, line 220: This is complicated multi-threaded code that I'm a little worried about.

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,247 @@
...
+		start = start.Add(start, big.NewInt(1))
+	}
+}
+
+// updateBlockNumber updates the last block number with a mutex lock.
+func (l *logTraversal) updateBlockNumber(blockNumber *big.Int) {
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	l.lastBlockNumber = blockNumber
+}
```
ajsutton, line 211: The locking around `lastBlockNumber` is inconsistent.  It's accessed without locks in `dispatchNewHead` and `LatestBlockNumber` but has locks around it when setting here. Both reads and writes need to use the same lock to ensure memory safety across threads.

But in this case, apart from the `LastBlockNumber` function which doesn't appear to be necessary, `lastBlockNumber` is only ever used from a single thread - the goroutine running `onNewHead`, so if `LastBlockNumber` were removed and the last block number treated entirely as an internal detail then no locks would be required for it.  Previously I think it was also being used by the `spawnCatchup` thread but that's not the case now.

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,247 @@
...
+// dispatchNewHead dispatches a new head.
+func (l *logTraversal) dispatchNewHead(ctx context.Context, header *types.Header, handleLog func(*types.Log) error, allowCatchup bool) {
+	info, err := l.fetchBlockInfoByHash(ctx, header.Hash())
+	if err != nil {
+		l.log.Error("Failed to fetch block", "err", err)
+		return
+	}
+	expectedBlockNumber := l.lastBlockNumber.Add(l.lastBlockNumber, big.NewInt(1))
+	currentBlockNumber := big.NewInt(int64(info.NumberU64()))
+	if l.lastBlockNumber.Cmp(big.NewInt(0)) != 0 && currentBlockNumber.Cmp(expectedBlockNumber) == 1 {
```
ajsutton, line 222: So we're just starting from whatever the first new block we receive after start up is? That will mean we fail to process any blocks that arrive while we're restarting/offline.

op-challenger/challenger/traversal.go
```diff
@@ -0,0 +1,247 @@
...
+
+// updateBlockNumber updates the last block number with a mutex lock.
+func (l *logTraversal) updateBlockNumber(blockNumber *big.Int) {
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	l.lastBlockNumber = blockNumber
+}
+
+// dispatchNewHead dispatches a new head.
+func (l *logTraversal) dispatchNewHead(ctx context.Context, header *types.Header, handleLog func(*types.Log) error, allowCatchup bool) {
```
ajsutton, line 214: allowCatchup is now always true so can be removed.

ajsutton (commented):
I suspect there's a need to take a step back and work out what the algorithm is for ensuring every block is processed _or_ (probably preferably) to find a way to approach the problem that still works even if blocks are missed.  For example if we can read the info we need from the latest world state then we could just periodically check that and not need to track every block.

Either way, we're currently missing any blocks that arrive while we're offline and not handling reorgs. Persistent storage for `lastBlockNumber` could solve the problem of missing blocks while offline, but I'm not sure I see how to adapt this logic to handle reorgs because it's entirely based around using block number and to handle reorgs well you generally need to use block hashes instead of numbers. I'm all for doing things incrementally, but I don't see how to get from here to where we process every block without doing a near complete rewrite of this code.

op-challenger/challenger/traversal.go
refcell: Done, built and restacked onto #5873

op-challenger/challenger/traversal.go
refcell: Definitely don't disagree, let's discuss in our huddle

op-challenger/challenger/traversal.go
refcell: Right, changed back to feeding a block number into the constructor

refcell:
Closing so we can approach the challenge agent as part of the fault proof roadmap.

PR #5758, opened 2023-05-22
feat(ctp): deploy faucet with ChugSplash

smartcontracts:
**Description**
Creates a deployment file so that the Faucet smart contract can be deployed with ChugSplash. Also creates a CI job that we can use to trigger deployments.


smartcontracts:
Pending this PR: https://github.com/ethereum-optimism/optimism/pull/5790

PR #5761, opened 2023-05-23
contracts-bedrock: dispute game factory event

tynes:
**Description**

Adds an event that is emitted when a dispute game implementation is added to the dispute game factory contract. This will make it easier to track the state of the contract.


packages/contracts-bedrock/contracts/dispute/IDisputeGameFactory.sol
```diff
@@ -23,6 +23,13 @@ interface IDisputeGameFactory {
         Claim indexed rootClaim
     );
 
+    /**
+     * @notice Emitted when a new game implementation added to the factory
+     * @param gameType The type of the DisputeGame.
+     * @param impl The implementation contract for the given `GameType`.
+     */
+    event ImplementationSet(address impl, GameType gameType);
```
clabby, line 31: Should probably index these args

clabby (approved):
LGTM apart from the non `indexed` event args

tynes:
@clabby appreciate the review but also enjoy your vacation :)

PR #5770, opened 2023-05-23
op-node: refactor scorer to decouple connection-gater and implement new layered gater

protolambda:
**Description**

As described in the peer-scoring issues notion doc, we should decouple the gating from the gossip scoring.
This updates the scorer not directly interface with the gater, and implements a new connection gater that handles ban-expiry.

Changes:
- Implement gating package:
  - Old default connection gater is now the `BlockingConnectionGater`
  - `ExpiryConnectionGater` can wrap a gater to handle ban-expiry of both peer-bans and IP-bans
  - `ScoringConnectionGater` can wrap a gater to enforce a minimum peer-score
  - `MeteredConnectionGater` can wrap a gater to track metrics of the dials and accepts, and how many of those we allow.
- Use the gater in the host
  - Just the blocking connection gater with metrics wrapper for now. We should only instantiate the expiry and scoring gater add-ons once we hook those up to the peerstore functionality.
- Remove the gater from the peer-scorer
- Create the scorer as part of the host setup instead of gossip setup. It shouldn't be in the gossip setup, because we write more than just gossip scores
  - The scorer is not hooked up to the new extended peerstore yet
  - The scorer connect/disconnect hooks now are actually hooked up as a notifee to libp2p host. Updating the scores based on decay is out of scope for this PR though.

**Also ignore the line-count; there's quite some generated code for mocks to test the gater functionality.**

Note that until we implement https://linear.app/optimism/issue/CLI-4002/op-node-peer-score-based-banskicks the scores won't be reflected in bans yet. So temporarily the (off-by-default) peer-banning is non-functional. Existing bans and bans through RPC api will still be enforced however.

**Tests**

- The new ban-expiry is unit-tested
- The existing scorer tests were updated to not use the old broken gater (which couldn't do unbans etc.)

**Metadata**

- Fixes CLI-3595


protolambda:
Rebasing and splitting this PR

PR #5771, opened 2023-05-23
op-node: implement new layered p2p connection gater

protolambda:
**Description**

Implementing new `op-node/p2p/gating` package:

- Old default connection gater is now the BlockingConnectionGater
- ExpiryConnectionGater can wrap a gater to handle ban-expiry of both peer-bans and IP-bans (with metrics for unbans)
- ScoringConnectionGater can wrap a gater to enforce a minimum peer-score
- MeteredConnectionGater can wrap a gater to track metrics of the dials and accepts, and how many of those we allow.

**Also ignore the line-count; there's quite some generated code for mocks to test the gater functionality.**

**Tests**

The new ban-expiry is unit-tested. The new gater is not integrated yet, e2e testing to follow upon integration.

**Metadata**

Fix CLI-3595

PR #5772, opened 2023-05-23
op-node: P2P scorer refactor

protolambda:
**Description**

This refactors the scorer to decouple from the connection-gater, and to persist more detailed scoring information in the peer-store, with room to grow without version changes, to handle req-resp sync scoring and other scoring functionality that we might add soon.

I changed the score-record to use JSON: it defaults to 0 values upon changes, making things less fragile to changes, as default score components are 0 and I expect more changes. The old approach was also not scalable to add much score detail, which I think we should to be able to debug the final score results.

With the refactor, we now adopt the basic connection-gater that was moved in #5771, but do not yet enable banning or any gating based on peer-score.

Based on top of #5771 

**Tests**

Updated tests to decouple gater and use updated peerstore functionality.

**Metadata**

Fix CLI-4001


op-node/p2p/node.go
```diff
@@ -6,6 +6,10 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/ethereum-optimism/optimism/op-node/p2p/store"
+
```
trianglesphere, line 10: weird goimports formatting

ajsutton (approved):
Looks good. If we can make the json decoder ignore fields in the JSON that don't exist in the struct we should be able to add and remove fields without changing version which would be nice. Can worry about that when/if its necessary though.

op-node/p2p/node.go
ajsutton: Pushed a commit to tidy these up so we can unblock further work.

PR #5775, opened 2023-05-24
feat(ci): use nvm to install node

smartcontracts:
**Description**
Updates the CI builder to use nvm to install node, which allows us to specify a particular version of node.

### DIFF
--- a/ops/docker/ci-builder/Dockerfile
+++ b/ops/docker/ci-builder/Dockerfile
@@ -43,12 +43,13 @@ COPY check-changed.sh /usr/local/bin/check-changed
 
 RUN apt-get update && \
   apt-get install -y bash curl openssh-client git build-essential ca-certificates jq musl gnupg coreutils && \
-  curl -sL https://deb.nodesource.com/setup_16.x -o nodesource_setup.sh && \
+  curl -sL https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh -o nvm_install.sh && \
   curl -sL https://go.dev/dl/go1.19.linux-amd64.tar.gz -o go1.19.linux-amd64.tar.gz && \
   tar -C /usr/local/ -xzvf go1.19.linux-amd64.tar.gz && \
   ln -s /usr/local/go/bin/gofmt /usr/local/bin/gofmt && \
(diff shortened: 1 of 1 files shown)


ops/docker/ci-builder/Dockerfile
```diff
@@ -43,12 +43,13 @@ COPY check-changed.sh /usr/local/bin/check-changed
...
-  curl -sL https://deb.nodesource.com/setup_16.x -o nodesource_setup.sh && \
+  curl -sL https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh -o nvm_install.sh && \
   curl -sL https://go.dev/dl/go1.19.linux-amd64.tar.gz -o go1.19.linux-amd64.tar.gz && \
   tar -C /usr/local/ -xzvf go1.19.linux-amd64.tar.gz && \
   ln -s /usr/local/go/bin/gofmt /usr/local/bin/gofmt && \
-  bash nodesource_setup.sh && \
-  apt-get install -y nodejs && \
+  bash nvm_install.sh && \
+  nvm install 16.16.0 && \
+  nvm use 16.16.0 && \
```
roninjin10, line 52: I have never been able to get this to work myself in a docker container.  If you get this working I'll definitely be copying you in all my stuff

tynes: Also this is still using 16, are there breaking changes between diff versions in the 16.x release?

roninjin10: All 16.x are safe

roninjin10: why not just copy in the .nvmrc though so .nvmrc is source of truth?
```typescript
COPY .nvmrc ./.nvmrc
...
nvm use && \
```

smartcontracts: @roninjin10 I don't think this image has the monorepo as context

smartcontracts: @tynes there's a hyper specific bug in npx that doesn't appear in the version that the monorepo specifies in nvmrc but appears in the version that the builder uses

PR #5777, opened 2023-05-24
op-node: persist p2p bans with expiration time, update ban-expiry gater

protolambda:
**Description**

This generalizes the score-book implementation, so we can use the same pattern with peer-bans and IP-bans.

The new peerstore temp-ban interface is then used by the ban-expiry connection gater (not enabled in p2p node yet).
The connection gater now first checks the inner blocking gater (i.e. perma bans), and then completely independently checks temp-bans. This is important, since we want to prune the temp-ban data in the peerstore, but don't want it to cause inconsistencies with perma-bans, so I decoupled them. The ban-expiration record itself signifies a temporary ban.

This also adds `BanPeer(id, expiration)` and `BanIP(ip, expiration)` helpers to the p2p node. Unban by setting the time to 0 time struct. We should consider exposing these via the p2p RPC like we do with perma-bans.

**mocks are updated to match the new interface, line count is a bit high because of it**

**Tests**

- The score-book tests cover the generalized book, but the the IP and peer ban books don't have tests (yet).
- The connection gater tests were updated to match the new peer/IP ban store interface and independent behavior of inner-gater.

**Metadata**

Fix CLI-4011

### DIFF
--- a/op-node/p2p/gating/expiry.go
+++ b/op-node/p2p/gating/expiry.go
@@ -1,16 +1,17 @@
 package gating
 
 import (
-	"errors"
-	"net"
 	"time"
 
-	"github.com/ethereum-optimism/optimism/op-service/clock"
-	"github.com/ethereum/go-ethereum/log"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr/net"
+
+	"github.com/ethereum/go-ethereum/log"
+
+	"github.com/ethereum-optimism/optimism/op-node/p2p/store"
+	"github.com/ethereum-optimism/optimism/op-service/clock"
 )
 
 type UnbanMetrics interface {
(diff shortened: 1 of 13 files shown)


op-node/p2p/gating/expiry.go
```diff
@@ -47,19 +46,19 @@ func AddBanExpiry(gater BlockingConnectionGater, store ExpiryStore, log log.Logg
 
 func (g *ExpiryConnectionGater) peerBanExpiryCheck(p peer.ID) (allow bool) {
 	// if the peer is blocked, check if it's time to unblock
-	expiry, err := g.store.PeerBanExpiry(p)
+	expiry, err := g.store.GetPeerBanExpiration(p)
+	if err == store.UnknownBanErr {
```
tynes, line 50: Is there a semantic different from doing an equality check with `==` and using `errors.Is`?

tynes: I feel like `errors.Is` will catch wrapped errors where an equality check wont?

ajsutton (commented):
Looks good.  I'd really like to see some tests around the new books go in. They're mostly leveraging the record book base so shouldn't have too many corner cases to try and cover.

op-node/p2p/gating/expiry.go
protolambda: `errors.Is` would work, but similar to `EOF` in many IO stdlib funcs, returning the exact error from `GetPeerBanExpiration` is part of the api here, so it shouldn't make a difference.

ajsutton: True, I'm just paranoid. :)

ajsutton:
Added the tests and made the `errors.Is` change to satisfy my paranoia. :)

PR #5779, opened 2023-05-25
op-node: Add peer monitor to kick/ban peers with low scores

ajsutton:
**Description**

Adds a `PeerMonitor` that runs in the background checking connected peers and disconnecting/banning those with scores below a minimum value.

* Currently peers aren't actually banned - that will call into the code from #5777.
* We are currently persisting peer scores for a day (plus some arbitrary time for GC to actually delete the score) so any peer that gets disconnected for a low score will be banned for at least that long. We need to pass in the actual retention period to the score book and apply that.
* There's more hard coded values in here than I'd like - particularly the min score to disconnect which should be the same as will be used for the score based peer gater and the `static` label for protected peers
* I'm not entirely sure if the adapter pattern is a good idea or not. The ability to list peers, check if a peer is protected, get peer scores and ban peers are all in different components in libp2p which gets pretty unwieldy to pass them all into the `PeerManager` separately. The adapter can hide that complexity a bit which I think is good, but it is another bit of complexity. Opinions welcome.
* It currently checks 1 peer per second regardless of how many peers. We could change that to iterate the peer list faster but then pause for a longer period between checks. I kind of like the simplicity though...

**Tests**

**Invariants**

**Additional context**

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4002/op-node-peer-score-based-banskicks

**TODOs**


protolambda (approved):
Functionality LGTM, small nit on the threshold thing

PR #5781, opened 2023-05-25
op-bindings: error handling

tynes:
**Description**

Ensure that the deployed bytecode is properly formatted hex to ensure that an empty value is not returned. This should never happen but is a sanity check to ensure that unexpected values are not returned.

### DIFF
--- a/op-bindings/bindings/registry.go
+++ b/op-bindings/bindings/registry.go
@@ -2,15 +2,20 @@ package bindings
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/ethereum-optimism/optimism/op-bindings/solc"
 	"github.com/ethereum/go-ethereum/common"
 )
 
+// layouts respresents the set of storage layouts. It is populated in an init function.
 var layouts = make(map[string]*solc.StorageLayout)
 
+// deployedBytecodes represents the set of deployed bytecodes. It is populated
+// in an init function.
 var deployedBytecodes = make(map[string]string)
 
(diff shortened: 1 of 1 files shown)


op-bindings/bindings/registry.go
```diff
@@ -19,11 +24,36 @@ func GetStorageLayout(name string) (*solc.StorageLayout, error) {
...
+	}
+	str = strings.TrimPrefix(str, "0x")
+
+	for _, c := range []byte(str) {
+		if !isHexCharacter(c) {
+			return false
+		}
+	}
+	return true
+}
```
refcell, line 59: I think we should be able to use the builtin golang `encoding/hex` library here to avoid regex typos. See: https://pkg.go.dev/encoding/hex#DecodeString

For example, we could have isHex just check if `hex.DecodeString(s)` returns an error. Something like:
```golang
// isHex validates whether each byte is valid hexadecimal string.
func isHex(str string) bool {
	if len(str)%2 != 0 {
		return false
	}
	str = strings.TrimPrefix(str, "0x")

    _, err := hex.DecodeString(str)
    if err != nil {
         return false
    }
    return true
}
```

refcell (changes requested):

op-bindings/bindings/registry.go
tynes: The above code was used in go-ethereum so I'm not worried about it having a bug

refcell (approved):

PR #5785, opened 2023-05-25
OP Stack Chain Registry

Sabnock01:
**Description**

This PR adds a registry contract for admin addresses of OP Stack deployments to register their addresses for easy query. Resolves [Ecosystem Contribution #47](https://github.com/ethereum-optimism/ecosystem-contributions/issues/47)

**Additional context**

Only ever used Foundry up until now for the most part but gave it a go at writing my first Hardhat task.


packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+pragma solidity 0.8.15;
+
+/**
+ * @title  OpStackChainRegistry
+ * @notice OpStackChainRegistry acts as a way to register and query
+ *         contract addresses for an OP Stack chain
+ */
+contract OpStackChainRegistry {
+
+    error OnlyDeploymentAdmin();
```
tynes, line 11: nit: natspec missing

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @notice Struct representing a deployment
+     *         on an OP Stack chain
+     */
+    struct Deployment {
+        string deploymentName;
```
tynes, line 18: nit: no need to prefix these with `deploymentXXX` as the struct is called `Deployment`

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+     * @notice Mapping of deployments to their chain registry
+     */
+    mapping(string => mapping(string => address)) public registry;
+
+    /**
+     * @notice Claims a deployment
+     *
+     * @param deployment The deployment to claim
+     */
+    function claimDeployment(Deployment calldata deployment) public {
```
tynes, line 46: You should generally be able to recover the state of a smart contract by applying the ordered set of events. No event here. This way of thinking about things enforces thinking of smart contracts as state machines

packages/contracts-bedrock/tasks/update-registry.ts
```diff
@@ -0,0 +1,62 @@
+import fs from 'fs'
+import path from 'path'
+
+import { task } from 'hardhat/config'
+
```
tynes, line 5: Can you add a comment here describing the purpose and usage of this task? 

Also re: moving to foundry, [this](https://github.com/wighawag/forge-deploy) looks promising to move to foundry in a backwards compat way, would love it if you took a look at it and shared thoughts

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
smartcontracts: I think we could also get rid of the `Deployment` struct in general and just have a mapping of `name => admin`. That would simplify things throughout the contract.

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+        deployments[deployment.deploymentName] = deployment.deploymentAdmin;
+    }
+
+    /**
+     * @notice Registers entries in a deployment
+     *
+     * @param deployment The deployment to register entries in
+     * @param entries An array of entries to register
+     */
+    function register(Deployment calldata deployment, DeploymentEntry[] calldata entries) public {
```
smartcontracts, line 56: Removing the `Deployment` struct would, for example, make it so that the function signature here is simply:

```
function register(string calldata deployment, DeploymentEntry[] calldata entries) public
```

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+        for (uint i = 0; i < entries.length; i++) {
+            registry[deployment.deploymentName][entries[i].entryName] = entries[i].entryAddress;
+        }
+    }
+
+    /**
+     * @notice Queries the chain registry for a list of deployment addresses
+     *
+     * @param deployment The deployment to query
+     * @param names An array of names to query the addresses for
```
smartcontracts, line 67: Nit, missing `@return` natspec

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+     */
+    function query(Deployment calldata deployment, string[] calldata names) public view returns (address[] memory) {
```
smartcontracts, line 69: Same comment as above: re just using `string calldata deployment` instead of the `Deployment` struct.

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    error OnlyDeploymentAdmin();
+
+    /**
+     * @notice Struct representing a deployment
+     *         on an OP Stack chain
+     */
+    struct Deployment {
+        string deploymentName;
+        address deploymentAdmin;
```
smartcontracts, line 19: I think it'd be worth having a function that allows an admin to transfer ownership over a deployment to a new address.

clabby: Would be cool to include a pre-commit/timelock to ownership transfers here (maybe now, maybe in the future). Most people will use proxy addresses for the deployment entries, so upgrades can still happen in the meantime via the `ProxyAdmin` owner multisig. Longer term, it'd be awesome to consider the `ProxyAdmin` owner as the `Deployment`'s `deploymentAdmin`.

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+ */
+contract OpStackChainRegistry {
+
+    error OnlyDeploymentAdmin();
+
+    /**
+     * @notice Struct representing a deployment
+     *         on an OP Stack chain
+     */
+    struct Deployment {
```
clabby, line 17: shower thought: Maybe people should also have to claim a `chainId` as well as the name of the deployment? rfc @tynes @smartcontracts

packages/contracts-bedrock/contracts/test/OpStackChainRegistry.t.sol
```diff
@@ -0,0 +1,63 @@
...
+        _entries[0] = OpStackChainRegistry.DeploymentEntry("entry1", makeAddr("entry1"));
+        _entries[1] = OpStackChainRegistry.DeploymentEntry("entry2", makeAddr("entry2"));
+
+        vm.prank(makeAddr("admin"));
+        chainRegistry.register(_deployment, _entries);
+
+        return _entries;
+    }
+
+    function test_claim() public {
```
tynes, line 39: Can you add natspec to these tests so that we can verify against natural language what the test is supposed to do?

packages/contracts-bedrock/contracts/test/OpStackChainRegistry.t.sol
```diff
@@ -0,0 +1,63 @@
...
+        _claim();
+    }
+
+    function test_register() public {
+        OpStackChainRegistry.Deployment memory _deployment = _claim();
+        _register(_deployment);
```
tynes, line 45: why use an internal function here if its only used once?

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
tynes, line 56: style nit: arguments are underscore prefixed throughout the codebase

packages/contracts-bedrock/tasks/update-registry.ts
```diff
@@ -0,0 +1,62 @@
...
+  )
+  .setAction(async (args, hre) => {
+    const OpStackChainRegistry = await hre.ethers.getContractFactory(
+      'OpStackChainRegistry'
+    )
+    const registry = OpStackChainRegistry.attach(args.registry)
+
+    const deploymentFolder = path.resolve(__dirname, args.deploymentFolder)
+
+    const deploymentFiles = fs
```
tynes, line 20: There is a better way to get the deployments using hardhat deploy, `hre.deployments.get`
https://github.com/wighawag/hardhat-deploy#the-deployments-field
be sure to import hh deploy into the file to get the ts lsp to work

packages/contracts-bedrock/tasks/update-registry.ts
```diff
@@ -0,0 +1,62 @@
+import fs from 'fs'
```
tynes, line 1: I wouldn't be opposed to using a foundry script for this, we started adding them to the `scripts` directory

packages/contracts-bedrock/contracts/test/OpStackChainRegistry.t.sol
Sabnock01: `_register()`? It's used in both `test_register()` and `test_query`.

packages/contracts-bedrock/contracts/periphery/ChainRegistry.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * @notice Transfers ownership of a deployment to a new admin
+     *
+     * @param _deployment The deployment to transfer ownership of
+     * @param _newAdmin The new admin to transfer ownership to
+     */
+    function transferAdmin(string calldata _deployment, address _newAdmin) public {
+        if (msg.sender != deployments[_deployment]) revert OnlyDeploymentAdmin();
```
smartcontracts, line 70: Probably worth making into a modifier, our codebase also currently doesn't use custom errors (just `require(...)`). I think for consistency we should use `require` statements and we can move to custom errors at some point in the future.

packages/contracts-bedrock/contracts/periphery/ChainRegistry.sol
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * @notice Mapping of deployments to their chain registry
+     */
+    mapping(string => mapping(string => address)) public registry;
+
+    /**
+     * @notice Claims a deployment
+     *
+     * @param _deployment The deployment to claim
+     */
```
smartcontracts, line 56: Missing the `_admin` param docstring

smartcontracts (commented):
Looking good! A few style nits left but let's get the functionality over the line and then we can clean up the style. We really need a better linter :-(

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
smartcontracts: Hmmm is chain ID part of system config?

packages/contracts-bedrock/tasks/update-registry.ts
smartcontracts: I think probably easiest to keep as hardhat and move all the scripts over to foundry later

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
smartcontracts: Yeah for now this is basically a social consensus hack so there aren't any security assumptions being made in this contract, other than the fact that off-chain infrastructure (like the SDK) may bootstrap itself from these addresses. Deployers probably want to be careful with the owner of this contract no matter what.

packages/contracts-bedrock/contracts/periphery/ChainRegistry.sol
```diff
@@ -0,0 +1,104 @@
...
+    function claimDeployment(string calldata _deployment, address _admin) public {
+        deployments[_deployment] = _admin;
```
MoeyJac, line 58: Shouldn't this check to see if a `_deplyment` already exists in the registry? Otherwise anyone could override `_deployment` in the registry with any `_admin`, allowing them access to edit any of the `DeploymentEntry`s.

```solidity
error DeploymentAlreadyClaimed();
...
if (deployments[_deployment] != 0) revert DeploymentAlreadyClaimed();
```

Sabnock01: Oh wow good catch!

clabby:
> Looking good! A few style nits left but let's get the functionality over the line and then we can clean up the style. We really need a better linter :-(

`forge fmt`

packages/contracts-bedrock/contracts/periphery/ChainRegistry.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * @notice Registers entries in a deployment
+     *
+     * @param _deployment The deployment to register entries in
+     * @param _entries An array of entries to register
+     */
+    function register(string calldata _deployment, DeploymentEntry[] calldata _entries) public {
+        if (msg.sender != deployments[_deployment]) revert OnlyDeploymentAdmin();
+        for (uint i = 0; i < _entries.length; i++) {
```
tynes, line 84: no need to assign i to 0 here, its a slight gas optimization, no need to hyper optimize but its a very low hanging fruit

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
MoeyJac: > Would be cool to include a pre-commit/timelock to ownership transfers here (maybe now, maybe in the future). Most people will use proxy addresses for the deployment entries, so upgrades can still happen in the meantime via the `ProxyAdmin` owner multisig. Longer term, it'd be awesome to consider the `ProxyAdmin` owner as the `Deployment`'s `deploymentAdmin`.

What would you think of a `mapping(string => mapping(address => boo)) = registeredAdmins` to keep track of a list of admins per deployment instead? Not much overhead needed and would allow multiple admins/addresses to manage a given deployment.

Would only additionally require
```solidity
// Checks to see at least 2 admin exist (in case od removal
// so no deployment is left orphaned
modifier twoAdmins(string memory _deployment, address _admin);

// Maintains map of deploymentNames => adminAddress => admin Exists
mapping(string => mapping(address => boo)) = registeredAdmins;

// Maintains map of deploymentName => deploymentExists
mapping(string => bool) deploymentExists;

function addAdmin(string memory _deployment, address _admin);

function revokeAdmin(string memory _deployment, address _admin) twoAdmins(_deployment, _admin);
```

packages/contracts-bedrock/tasks/update-registry.ts
Sabnock01: Doesn't this just get the deployment artifact for a single contract? Wouldn't you still need a list of names? Or do you see that as something you'd provide as an arg?

The way I have it now is it asks if the deployment (eg 'mainnet') is already claimed, if not it claims it. Then it goes through each of the files, asks if it's already registered, if not register it. 

How did you envision this?

smartcontracts (approved):
Nice work! LGTM

packages/contracts-bedrock/contracts/periphery/OpStackChainRegistry.sol
Sabnock01: It is not as far as I can tell.

tynes: I feel like if we cannot answer this question, this contract is far from being used in production

packages/contracts-bedrock/tasks/update-registry.ts
```diff
@@ -0,0 +1,71 @@
...
+import path from 'path'
+
+import { task } from 'hardhat/config'
+
+/**
+ * Updates the ChainRegistry contract by claiming new deployments
+ * and registering new contract addresses for existing deployments
+ */
```
roninjin10, line 9: What do you think about migrating this deploy script to foundry?

roninjin10 (approved):

packages/contracts-bedrock/tasks/update-registry.ts
Sabnock01: Should be trivial. What directory should it go in?

Sabnock01:
Hi. Moved onto other projects after this was initially approved at the end of May but will try to get a new forge script written for this week.

tynes:
Going to close this now because we are going to go with an offchain solution but appreciate the contributions

PR #5793, opened 2023-05-26
op-node: Enable scoring connection gaters

ajsutton:
**Description**

Adds the ban expiry and scoring connection gaters into the p2p stack so that peer bans expiry correctly and peers with low scores are unable to connect (even if they disconnected before actually being banned).

**Tests**

Really need e2e tests to cover how this all hangs together (https://linear.app/optimism/issue/CLI-3603/op-e2e-peer-score-testing).

**Additional context**

Builds on #5779

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4015/op-node-utilize-the-ban-expiry-and-scoring-peer-gaters

### DIFF
--- a/op-node/p2p/host.go
+++ b/op-node/p2p/host.go
@@ -163,9 +163,8 @@ func (conf *Config) Host(log log.Logger, reporter metrics.Reporter, metrics Host
 	if err != nil {
 		return nil, fmt.Errorf("failed to open connection gater: %w", err)
 	}
-	// TODO(CLI-4015): apply connGtr enhancements
-	// connGtr = gating.AddBanExpiry(connGtr, ps, log, cl, reporter)
-	//connGtr = gating.AddScoring(connGtr, ps, 0)
+	connGtr = gating.AddBanExpiry(connGtr, ps, log, clock.SystemClock, metrics)
+	connGtr = gating.AddScoring(connGtr, ps, minAcceptedPeerScore)
 	connGtr = gating.AddMetering(connGtr, metrics)
 
 	connMngr, err := DefaultConnManager(conf)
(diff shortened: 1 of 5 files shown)


protolambda:
Just checked metrics to see what the scores are like, but the information we track is not detailed enough. All peers end up in the "nopx" band (score 0 - 20), no exceptions. Setting a lower bound of -100 is probably fine for now, but I think we need more detailed metrics (heatmap of metered scores over time, without tagging individual peers maybe?) to adjust the range. I think by using the total Gossip score (without req-resp sync score) we end up with a 0-1 range.

ajsutton:
> Just checked metrics to see what the scores are like, but the information we track is not detailed enough. All peers end up in the "nopx" band (score 0 - 20), no exceptions. Setting a lower bound of -100 is probably fine for now, but I think we need more detailed metrics (heatmap of metered scores over time, without tagging individual peers maybe?) to adjust the range. I think by using the total Gossip score (without req-resp sync score) we end up with a 0-1 range.

Yeah we definitely need to improve the metrics. We do expose the actual scores via RPC now which for my local node that's been running for multiple days since last restart the count of scores I get are:
```
cast rpc opp2p_peers false --rpc-url http://localhost:9555 | jq '.peers|to_entries[]|.value.scores.gossip.total' | sort -n | uniq -c
   1 -5.04
  75 0
   1 6.76335245727138
   8 8
   1 8.083129784306813
   1 25.70725152572483
```

Send an invalid messages will get you a very negative score very quickly based on previous testing.  The -100 value may actually be too low but seems like a reasonable starting point at least.

When we have p2p sync scoring we will either have to hook that in as the gossip application score (in which case I'm expecting it to be included in total) or come up with our own system for combining the scores. Plugging into application score certainly looks like the easiest version.

protolambda (approved):
LGTM, not enabling score-gating yet, but peer-monitor (with CLI flag, off by default) + properly enforcing temp-bans is good to merge.

PR #5797, opened 2023-05-26
op-batcher: Add metrics for pending L2 transaction data size

trianglesphere:
**Description**

This adds metrics to track the size of pending blocks (rather than just the number).
It is expressed in a counter & a gauge. There is no new counter for the pending size
in channels because that counter already exists as the `input_bytes` metrics.

Note: This is based off of the `op-batcher/v1.0.9-rc.1` tag.

### DIFF
--- a/op-batcher/metrics/metrics.go
+++ b/op-batcher/metrics/metrics.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/prometheus/client_golang/prometheus"
@@ -30,6 +31,8 @@ type Metricer interface {
 	RecordL2BlocksLoaded(l2ref eth.L2BlockRef)
 	RecordChannelOpened(id derive.ChannelID, numPendingBlocks int)
 	RecordL2BlocksAdded(l2ref eth.L2BlockRef, numBlocksAdded, numPendingBlocks, inputBytes, outputComprBytes int)
+	RecordL2BlockInPendingQueue(block *types.Block)
+	RecordL2BlockInChannel(block *types.Block)
 	RecordChannelClosed(id derive.ChannelID, numPendingBlocks int, numFrames int, inputBytes int, outputComprBytes int, reason error)
 	RecordChannelFullySubmitted(id derive.ChannelID)
 	RecordChannelTimedOut(id derive.ChannelID)
@@ -55,8 +58,10 @@ type Metrics struct {
 	// label by openend, closed, fully_submitted, timed_out
 	channelEvs opmetrics.EventVec
 
-	pendingBlocksCount prometheus.GaugeVec
-	blocksAddedCount   prometheus.Gauge
+	pendingBlocksCount        prometheus.GaugeVec
+	pendingBlocksBytesTotal   prometheus.Counter
+	pendingBlocksBytesCurrent prometheus.Gauge
+	blocksAddedCount          prometheus.Gauge
 
 	channelInputBytes       prometheus.GaugeVec
 	channelReadyBytes       prometheus.Gauge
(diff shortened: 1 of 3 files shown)


op-batcher/metrics/metrics.go
```diff
@@ -109,6 +114,16 @@ func NewMetrics(procName string) *Metrics {
...
+		pendingBlocksBytesCount: factory.NewCounter(prometheus.CounterOpts{
+			Namespace: ns,
+			Name:      "pending_blocks_bytes_count",
+			Help:      "Size of transactions in pending blocks as they are fetched from L2",
+		}),
+		pendingBlocksBytesCurrent: factory.NewGauge(prometheus.GaugeOpts{
+			Namespace: ns,
+			Name:      "pending_blocks_bytes_gauge",
+			Help:      "Current size of transactions in the pending (fetched from L2 but not in a channel) stage.",
+		}),
```
sebastianst, line 126:
```suggestion
		pendingBlocksBytesTotal: factory.NewCounter(prometheus.CounterOpts{
			Namespace: ns,
			Name:      "pending_blocks_bytes_total",
			Help:      "Total size of transactions in pending blocks as they are fetched from L2",
		}),
		pendingBlocksBytesCurrent: factory.NewGauge(prometheus.GaugeOpts{
			Namespace: ns,
			Name:      "pending_blocks_bytes_current",
			Help:      "Current size of transactions in the pending (fetched from L2 but not in a channel) stage.",
		}),
```
I suggest we keep the naming of the metrics and variables more consistent with the other metrics. I like the idea of appending `current` to clearly differentiate from the `total` though. Less confusing.

Note that the [prometheus docs](https://prometheus.io/docs/practices/naming/) suggest that the metric names of counters end in `total`.
> Note that an accumulating count has `total` as a suffix, in addition to the unit if applicable.

op-batcher/metrics/metrics.go
```diff
@@ -243,6 +258,24 @@ func (m *Metrics) RecordChannelClosed(id derive.ChannelID, numPendingBlocks int,
...
+}
+
+func (m *Metrics) RecordL2BlockInChannel(block *types.Block) {
+	var size uint64
+	for _, tx := range block.Transactions() {
+		size += tx.Size()
+	}
+	m.pendingBlocksBytesCurrent.Add(-1 * float64(size))
+	// Refer to RecordL2BlocksAdded to see the current + count of bytes added to a channel
+}
```
sebastianst, line 277: Why did you decide to sum up transaction sizes, instead of the whole block (using rlp-encoded size)? Isn't this slightly less also? The metric would be more comparable to the other `input_bytes` and `input_bytes_total` metrics if you use full rlp block encoding size.

op-batcher/metrics/metrics.go
```diff
@@ -55,8 +58,10 @@ type Metrics struct {
 	// label by openend, closed, fully_submitted, timed_out
 	channelEvs opmetrics.EventVec
 
-	pendingBlocksCount prometheus.GaugeVec
-	blocksAddedCount   prometheus.Gauge
+	pendingBlocksCount        prometheus.GaugeVec
+	pendingBlocksBytesCount   prometheus.Counter
+	pendingBlocksBytesCurrent prometheus.Gauge
+	blocksAddedCount          prometheus.Gauge
```
sebastianst, line 64: I realize that it was a confusing idea to end scalar gauges in `Count` because there's also counters... I like using `Current`. However, I propose we keep it consistent within the file and with the general prometheus convention to suffix counters with `Total`.
```suggestion
	pendingBlocksCount        prometheus.GaugeVec
	pendingBlocksBytesTotal   prometheus.Counter
	pendingBlocksBytesCurrent prometheus.Gauge
	blocksAddedCount          prometheus.Gauge
```

sebastianst (changes requested):

op-batcher/metrics/metrics.go
trianglesphere: FYI `input_bytes` is computed from the actual size of the batches (not the blocks)

sebastianst (approved):

PR #5798, opened 2023-05-27
op-bootnode

mdehoog:
**Description**

Added an implementation of a standalone bootnode that simply runs op-node's discovery process.

**Tests**

Test coverage is mostly provided by op-node's existing P2P package. We may want to add some E2E tests for this wrapper.

**Additional context**

This is the bootnode implementation that Base runs in production. It also supports bootstrapping networks before genesis, as only a small amount of genesis information is required to run a bootnode:
 - L2 chain ID
 - L2 genesis timestamp
 - L2 block time
 - L2 genesis block number (defaults to 0)

### DIFF
--- a/op-bootnode/bootnode/entrypoint.go
+++ b/op-bootnode/bootnode/entrypoint.go
@@ -4,9 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
-	"os/signal"
-	"syscall"
 
 	opnode "github.com/ethereum-optimism/optimism/op-node"
 	"github.com/ethereum-optimism/optimism/op-node/eth"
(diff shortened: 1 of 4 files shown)


op-bootnode/bootnode/entrypoint.go
```diff
@@ -0,0 +1,83 @@
...
+
+	p2pNode, err := p2p.NewNodeP2P(ctx, config, logger, p2pConfig, &gossipNoop{}, &l2Chain{}, &gossipConfig{}, m)
+	if err != nil || p2pNode == nil {
+		return err
+	}
+	if p2pNode.Dv5Udp() == nil {
+		return fmt.Errorf("uninitialized discovery service")
+	}
+
+	p2pNode.DiscoveryProcess(ctx, logger, config, p2pConfig.TargetPeers())
```
tynes, line 67: No need to block until a sigterm or sigkill here after starting the discover process?

mdehoog: good point, fixed

refcell (commented):
Looking good to me 👍

refcell:
I think you might just have to rebase now @mdehoog :)

trianglesphere (approved):

PR #5800, opened 2023-05-27
op-node: fix TestMultiPeerSync race condition

protolambda:
**Description**

Fix dead-lock in op-node p2p sync test, which happened in [this CI run](https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/18639/workflows/0a0ff489-fd1f-4b1a-8c3e-c20b6a95f262/jobs/740880) reported by @tynes 

- fix TestMultiPeerSync race condition, by holding off on the manual re-request until the failed request is actually marked as complete
- add debug log to catch this problem where a sync request is not being made because of an already in-flight request
- add `for i := uint64(89); i > 10; i-- { // wait for all payloads` around first sync work of range `<10, 90>`. This should complete before we do the test part that temporarily hides payload 25. Previously the test would only inspect first synced block result.
- cleanup sync-test-data to handle concurrent payloads access properly: the `map` shouldn't be read/written/deleted to concurrently.


trianglesphere (approved):
lgtm. TY for keeping CI green

PR #5817, opened 2023-05-29
op-node: minimal chain config

tynes:
**Description**

Update the mainnet config for what is already known for the hardcoded mainnet config.
These values should be checked against what is in the mainnet deploy config json and the mainnet deploy artifacts.


mslipper (approved):

mslipper:
Confirmed based on today's values in the Notion table.

PR #5818, opened 2023-05-29
feat(ctb): Generate deployment README files

maurelian:
**Description**

Ports the `autogen:markdown` script from the `contracts` package to `contracts-bedrock` and generates the files.

Draft until the artifacts are renamed to remove the OVM prefix.


tynes:
We should consider usage of `forge doc` to see if it works for this usecase as to minimize additional custom tooling

maurelian:
Maybe we just add the md files for now then

Issue #4980, opened 2023-02-25
Convert Echidna tests to Forge invariantes


maurelian:
We currently have a suite of [Echidna tests](https://github.com/ethereum-optimism/optimism/blob/24edceb652b5a812313b0b742f74acaa0f3ea3a5/packages/contracts-bedrock/contracts/echidna) in our `contracts-bedrock` package which use Echidna for property testing. 

We also have a suite of forge [invariant tests](https://github.com/ethereum-optimism/optimism/blob/24edceb652b5a812313b0b742f74acaa0f3ea3a5/packages/contracts-bedrock/contracts/test/invariants). 

In order to standardize our tooling, we'd like to convert the echidna tests over to Forge invariant tests. External contributors would be very welcome here. 

A suggested workflow would be to open one PR per converted test file: 

- [ ] FuzzAddressAliasing.sol
- [ ] FuzzBurn.sol
- [ ] FuzzEncoding.sol
- [ ] FuzzHashing.sol
- [ ] FuzzOptimismPortal.sol
- [ ] FuzzResourceMetering.sol

The first 4 files listed are much simpler to work with, so I'd suggest starting from the top.

Ratimon:
@maurelian Hi, I think this issue is now resolved, so it might be closed now, or should there be any possible contributions I should make here?

maurelian:
LGTM, thank you so much for your help here!

PR #5822, opened 2023-05-30
contracts-bedrock: Convert Echidna tests  to Forge invariants for Burn library 

Ratimon:
**Description**

I have added `Burn.Eth.t.sol` and `Burn.Gas.t.sol`  files.

**Invariants**

It is noted that:

According to original `FuzzBurn.sol` echidna test file
`
        unchecked {
            if (postBurnGas - preBurnGas > value || preBurnGas - value > preBurnGas) {
                failedGasBurn = true;
            }
        }
`
I notice that the logic should has been that it tries to falsify invariant (at least value gas was burnt). In other words, it should be that if the gas burned is less that the value pass, `failedGasBurn` should then be set to be true. Hence, the invariant cant be hold.

So, I have implement the new logic in `Burn.Gas.t.sol` as followed:

`       
        unchecked {
            if (postBurnGas - preBurnGas <= value && preBurnGas - value > preBurnGas) {
                failedGasBurn = true;
            }
        }
`

Apologize if my understanding is wrong.

**Metadata**

- Fixes #[Convert Echidna tests to Forge invariantes #4980] 

**TODOs**

- [ ] Fix original `FuzzBurn.sol` echidna  if my above understanding is correct.

### DIFF
--- a/packages/contracts-bedrock/contracts/test/invariants/Burn.Eth.t.sol
+++ b/packages/contracts-bedrock/contracts/test/invariants/Burn.Eth.t.sol
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { StdUtils } from "forge-std/StdUtils.sol";
+import { Test } from "forge-std/Test.sol";
+import { Vm } from "forge-std/Vm.sol";
+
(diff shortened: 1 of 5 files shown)


refcell (approved):
Looks great!

PR #5825, opened 2023-05-30
op-wheel: Fix post-shanghai

mslipper:


op-wheel/engine/engine.go
```diff
@@ -18,6 +19,28 @@ import (
...
+		Withdrawals           []*types.Withdrawal `json:"withdrawals"`
+	}
+	var enc PayloadAttributes
+	enc.Timestamp = hexutil.Uint64(p.Timestamp)
+	enc.Random = p.Random
+	enc.SuggestedFeeRecipient = p.SuggestedFeeRecipient
+	enc.Withdrawals = make([]*types.Withdrawal, 0)
+	return json.Marshal(&enc)
+}
+
```
protolambda, line 43: Looks like an extract of the code-generation in geth. Just defining the `Timestamp hexutil.Uint64` field in the actual V2 struct also works, and avoids this extra code.

op-wheel/engine/engine.go
```diff
@@ -112,21 +135,17 @@ func BuildBlock(ctx context.Context, client client.RPC, status *StatusData, sett
...
 	}
 	var pre engine.ForkChoiceResponse
-	if err := client.CallContext(ctx, &pre, "engine_forkchoiceUpdatedV1",
+	if err := client.CallContext(ctx, &pre, "engine_forkchoiceUpdatedV2",
 		engine.ForkchoiceStateV1{
 			HeadBlockHash:      status.Head.Hash,
 			SafeBlockHash:      status.Safe.Hash,
 			FinalizedBlockHash: status.Finalized.Hash,
-		}, engine.PayloadAttributes{
+		}, PayloadAttributesV2{
```
protolambda, line 143: Changing it to V2 is fine, but it does mean it won't be able to build payloads with a pre-shanghai timestamp. Maybe worth a comment, and a change in the CLI help flags to state that it builds post-shanghai blocks only.

Spec:
```
PayloadAttributesV1 MUST be used to build a payload with the timestamp value lower than the Shanghai timestamp,
PayloadAttributesV2 MUST be used to build a payload with the timestamp value greater or equal to the Shanghai timestamp,
```
https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2

protolambda (approved):
Changes look correct, although I think we can get rid of  the`MarshalJSON` by just changing the timestamp type. And a comment and/or note in CLI help that it now supports post-shanghai block-building only would be nice.

tynes (approved):

protolambda:
Resolving my comments, will open a PR to fix the nits

PR #5830, opened 2023-05-30
feat: Refactor interrupt blocking.

refcell:
**Description**

While working on the op-challenger I noticed the interrupt blocking is fairly redundant across multiple services and fairly error prone to typos. The PR refactors interrupt blocking into an `op-service` package called `io` with variadic signal overrides.


ajsutton (approved):
LGTM.

op-service/io/interrupts.go
```diff
@@ -0,0 +1,26 @@
+package io
```
trianglesphere, line 1: only suggestion is that `io` is already overloaded. I might just call this `opio` so we don't have to do the rename.

PR #5831, opened 2023-05-30
feat(op-challenger): Oracle Watcher

refcell:
**Description**

- Sets up the `op-challenger` subcommand dispatch.
- Adds an L2OutputOracle `OutputProposed` event watcher as part of the `watch` subcommand.

**Metadata**

Fixes CLI-4062

### DIFF
--- a/op-challenger/cmd/entrypoint.go
+++ b/op-challenger/cmd/entrypoint.go
@@ -1,4 +1,4 @@
-package challenger
+package main
 
 import (
 	"context"
@@ -10,8 +10,11 @@ import (
 	"github.com/ethereum-optimism/optimism/op-challenger/config"
 	"github.com/ethereum-optimism/optimism/op-challenger/metrics"
 	"github.com/ethereum-optimism/optimism/op-service/opio"
-	oppprof "github.com/ethereum-optimism/optimism/op-service/pprof"
-	oprpc "github.com/ethereum-optimism/optimism/op-service/rpc"
+
+	"github.com/ethereum-optimism/optimism/op-challenger/challenger"
+
+	"github.com/ethereum-optimism/optimism/op-service/pprof"
+	"github.com/ethereum-optimism/optimism/op-service/rpc"
 )
 
 // Main is the entrypoint into the Challenger. This method executes the
@@ -24,7 +27,7 @@ func Main(logger log.Logger, version string, cfg *config.Config) error {
 	m := metrics.NewMetrics("default")
 	logger.Info("Initializing Challenger")
 
-	challenger, err := NewChallenger(*cfg, logger, m)
+	service, err := challenger.NewChallenger(*cfg, logger, m)
 	if err != nil {
 		logger.Error("Unable to create the Challenger", "error", err)
 		return err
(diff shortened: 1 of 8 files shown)


op-challenger/cmd/entrypoint.go
```diff
@@ -1,17 +1,18 @@
...
-	_ "net/http/pprof"
 	"os"
 	"os/signal"
 	"syscall"
 
 	"github.com/ethereum/go-ethereum/log"
 
-	"github.com/ethereum-optimism/optimism/op-challenger/config"
-	"github.com/ethereum-optimism/optimism/op-challenger/metrics"
+	challenger "github.com/ethereum-optimism/optimism/op-challenger/challenger"
```
tynes, line 12: why named imports? they are the same as the package name

op-challenger/cmd/watch/oracle.go
```diff
@@ -0,0 +1,76 @@
...
+
+	log "github.com/ethereum/go-ethereum/log"
+
+	challenger "github.com/ethereum-optimism/optimism/op-challenger/challenger"
+	config "github.com/ethereum-optimism/optimism/op-challenger/config"
+	metrics "github.com/ethereum-optimism/optimism/op-challenger/metrics"
+)
+
+// Oracle listens to the L2OutputOracle for newly proposed outputs.
+func Oracle(logger log.Logger, cfg *config.Config) error {
```
tynes, line 18: This would be cool if it optionally also took an op-node RPC url and then did a comparison, could be used as a detector service to trigger an alert if the mismatched output exists for too long

refcell: Interesting, what do you think about putting this behind a long flag? (Maybe something like "--with-rpc")

op-challenger/cmd/watch/cmd.go
```diff
@@ -0,0 +1,35 @@
...
+
+			return Oracle(logger, cfg)
+		},
+	},
+	{
+		Name:  "factory",
+		Usage: "Watches the DisputeGameFactory for new dispute games",
+		Action: func(ctx *cli.Context) error {
+			panic("factory not implemented")
+		},
```
ajsutton, line 33: I probably just wouldn't introduce this until the next PR where it gets implemented rather than panicing.

ajsutton (commented):
I'm not sure this actually fixes [CLI-3952](https://linear.app/optimism/issue/CLI-3952) which talks about supporting a variety of modes including sending challenges etc.

Same kind of comments as on #5842 - I'm not really sure I understand where this is all going so it's hard to know if this is the right way to go about it or not.  Quite possibly I'm just not the right person to be reviewing this now too. :)

refcell:
> I'm not sure this actually fixes [CLI-3952](https://linear.app/optimism/issue/CLI-3952) which talks about supporting a variety of modes including sending challenges etc.
> 
> Same kind of comments as on #5842 - I'm not really sure I understand where this is all going so it's hard to know if this is the right way to go about it or not. Quite possibly I'm just not the right person to be reviewing this now too. :)

This is a good point, so these PRs set up the foundations for CLI-3952 for the challenger to run in watch mode. But I agree this doesn't directly fix that ticket, so i'll sub itemize that ticket and make it clear 👍

op-challenger/cmd/watch/oracle.go
```diff
@@ -0,0 +1,60 @@
...
+
+	service, err := challenger.NewChallenger(*cfg, logger, m)
+	if err != nil {
+		logger.Error("Unable to create the Challenger", "error", err)
+		return err
+	}
+
+	logger.Info("Listening for OutputProposed events from the L2OutputOracle contract", "l2oo", cfg.L2OOAddress.String())
+
+	subscription, err := service.NewOracleSubscription()
```
ajsutton, line 32: should this have a `defer subscription.Quit()` to ensure the subscription is closed cleanly? The various clients created by challenger aren't closed either.

I don't see where the clients are closed in things like the batcher either so may not matter since the whole process should exit.

ajsutton (approved):
LGTM, just the one question about closing resources.

op-challenger/cmd/watch/oracle.go
refcell: Agree, I think it makes sense unsubscribe gracefully even though the process exits.

PR #5840, opened 2023-05-31
[indexer] consolidate retrieval of finalized headers

hamdiallam:
The Fetcher will be used to drive new batches of headers to check. This can be used against both the L1 and L2 instances. Not super clear why the indexer currently duplicates this logic in both L1/L2

Follow up
1. Implement the processor which will poll the Fetcher
2. Investigate if we can utilize a subscription for new headers instead of polling
3. More robust implementation for the `EthClient` interface. See some of the comments on this PR

### DIFF
--- a/indexer/go.mod
+++ b/indexer/go.mod
@@ -154,6 +154,7 @@ require (
 	github.com/shirou/gopsutil v3.21.11+incompatible // indirect
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
 	github.com/status-im/keycard-go v0.2.0 // indirect
+	github.com/stretchr/objx v0.5.0 // indirect
 	github.com/syndtr/goleveldb v1.0.1-0.20220614013038-64ee5596c38a // indirect
 	github.com/tklauser/go-sysconf v0.3.10 // indirect
 	github.com/tklauser/numcpus v0.5.0 // indirect
--- a/indexer/node/client.go
+++ b/indexer/node/client.go
@@ -24,13 +24,9 @@ type EthClient interface {
 	FinalizedBlockHeight() (*big.Int, error)
 	BlockHeadersByRange(*big.Int, *big.Int) ([]*types.Header, error)
 
-	// TODO: probably will remove this
 	RawRpcClient() *rpc.Client
 }
 
-// TODO:
-//   - Have client transparently support retry semantics
-//   - Members should be private and supply the needed methods
 type client struct {
 	rpcClient *rpc.Client
 }
@@ -103,7 +99,7 @@ func (c *client) BlockHeadersByRange(startHeight, endHeight *big.Int) ([]*types.
 
 		header := batchElem.Result.(*types.Header)
 		if i > 0 && header.ParentHash != headers[i-1].Hash() {
-			// TODO: Log here that we got a bad (malicious?) response
+			// Warn here that we got a bad (malicious?) response
 			break
 		}
 
--- a/indexer/node/client.go
+++ b/indexer/node/client.go
@@ -1,0 +1,131 @@
+package node
+
+import (
+	"context"
+	"math/big"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+const (
+	// defaultDialTimeout is default duration the processor will wait on
+	// startup to make a connection to the backend
+	defaultDialTimeout = 5 * time.Second
+
+	// defaultRequestTimeout is the default duration the processor will
+	// wait for a request to be fulfilled
+	defaultRequestTimeout = 10 * time.Second
+)
+
+type EthClient interface {
+	FinalizedBlockHeight() (*big.Int, error)
+	BlockHeadersByRange(*big.Int, *big.Int) ([]*types.Header, error)
+
+	// TODO: probably will remove this
+	RawRpcClient() *rpc.Client
+}
+
+// TODO:
+//   - Have client transparently support retry semantics
+//   - Members should be private and supply the needed methods
+type client struct {
+	ethClient *ethclient.Client
+	rpcClient *rpc.Client
+}
+
+func NewEthClient(rpcUrl string) (EthClient, error) {
+	ctxwt, cancel := context.WithTimeout(context.Background(), defaultDialTimeout)
+	defer cancel()
+
+	rpcClient, err := rpc.DialContext(ctxwt, rpcUrl)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &client{rpcClient: rpcClient}
+	return client, nil
+}
+
+func (c *client) RawRpcClient() *rpc.Client {
+	return c.rpcClient
+}
+
+// FinalizedBlockHeight retrieves the latest block height in a finalized state
+func (c *client) FinalizedBlockHeight() (*big.Int, error) {
+	ctxwt, cancel := context.WithTimeout(context.Background(), defaultRequestTimeout)
+	defer cancel()
+
+	var block *types.Block
+	err := c.rpcClient.CallContext(ctxwt, block, "eth_getBlockByNumber", "finalized", false)
+	if err != nil {
+		return nil, err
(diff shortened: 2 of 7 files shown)


indexer/node/client.go
```diff
@@ -0,0 +1,129 @@
...
+
+type EthClient interface {
+	FinalizedBlockHeight() (*big.Int, error)
+	BlockHeadersByRange(*big.Int, *big.Int) ([]*types.Header, error)
+
+	// TODO: probably will remove this
+	RawRpcClient() *rpc.Client
+}
+
+// TODO:
```
roninjin10, line 31: same linter might make you add a linear ticket
```suggestion
// TODO https://linear.app/optimism/issue/DX-51/indexer-create-event-processor
```

indexer/node/fetcher.go
```diff
@@ -0,0 +1,63 @@
...
+	// eventually be configurable
+	maxHeaderBatchSize = 50
+)
+
+type Fetcher struct {
+	ethClient EthClient
+
+	// TODO: Store the last header block hash to ensure
+	// the next batch of headers builds on top
+	nextStartingBlockHeight *big.Int
```
roninjin10, line 20: This could be a regular number I think if you prefer.   I personally like just using big.Int without thinking too hard about it though.

indexer/node/client.go
```diff
@@ -0,0 +1,129 @@
...
+func NewEthClient(rpcUrl string) (EthClient, error) {
+	ctxwt, cancel := context.WithTimeout(context.Background(), defaultDialTimeout)
+	defer cancel()
+
+	rpcClient, err := rpc.DialContext(ctxwt, rpcUrl)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &client{rpcClient: rpcClient}
```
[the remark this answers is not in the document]
tynes: Ah batch requests

indexer/node/client.go
```diff
@@ -0,0 +1,129 @@
...
+
+		headers[i] = header
+		size = size + 1
+	}
+	headers = headers[:size]
+
+	return headers, nil
+}
+
+func toBlockNumArg(number *big.Int) string {
```
tynes, line 118: Perhaps using these types could be helpful in case the values under the hood change
https://github.com/ethereum/go-ethereum/blob/ac86547b0154129f4dac865579946bd2e9378fb0/rpc/types.go#L65

roninjin10 (approved):

roninjin10 (approved):

indexer/node/client.go
```diff
@@ -0,0 +1,129 @@
...
+
+	client := &client{rpcClient: rpcClient}
+	return client, nil
+}
+
+func (c *client) RawRpcClient() *rpc.Client {
+	return c.rpcClient
+}
+
+// FinalizedBlockHeight retrieves the latest block height in a finalized state
```
tynes, line 55: Potentially not a bad idea to reuse our backoff tooling for retries
https://github.com/ethereum-optimism/optimism/blob/2d04a15ebde0baf885b17760f11496cf54efe55f/op-service/backoff/operation.go#L27

roninjin10: +1

indexer/node/client.go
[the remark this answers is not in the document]
tynes: We do have tooling for creating an ethclient, not sure if it'll give you access to making batch calls tho
https://github.com/ethereum-optimism/optimism/blob/develop/op-service/client/ethclient.go

tynes:
Could reuse this tooling for the batch calling: https://github.com/ethereum-optimism/optimism/blob/2d04a15ebde0baf885b17760f11496cf54efe55f/op-node/client/rpc.go#L144

indexer/node/client.go
```diff
@@ -0,0 +1,129 @@
...
+
+// TODO:
+//   - Have client transparently support retry semantics
+//   - Members should be private and supply the needed methods
+type client struct {
+	rpcClient *rpc.Client
+}
+
+func NewEthClient(rpcUrl string) (EthClient, error) {
+	ctxwt, cancel := context.WithTimeout(context.Background(), defaultDialTimeout)
```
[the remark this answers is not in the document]
tynes: the request itself will timeout client side after the `defaultDialTimeout`

indexer/node/fetcher.go
tynes: I believe the number in the header is a `big.Int` so it makes sense to use the same type. `big.Int` is a weird type its one of my least fav parts of go, takes awhile to get used to

indexer/go.mod
```diff
@@ -147,6 +147,7 @@ require (
 	github.com/shirou/gopsutil v3.21.11+incompatible // indirect
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
 	github.com/status-im/keycard-go v0.2.0 // indirect
+	github.com/stretchr/objx v0.5.0 // indirect
```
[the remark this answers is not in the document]
tynes: it looks like its a transitive dep so one of the packages that is a dep imports it

indexer/node/fetcher.go
```diff
@@ -0,0 +1,63 @@
...
+// NextConfirmedHeaders retrives the next set of headers that have been
+// marked as finalized by the connected client
+func (f *Fetcher) NextFinalizedHeaders() ([]*types.Header, error) {
+	finalizedBlockHeight, err := f.ethClient.FinalizedBlockHeight()
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO:
+	//  - (unlikely) What do we do if our connected node is suddently behind by many blocks?
```
tynes, line 43: I'd log a warning and return just like how you are

indexer/node/fetcher.go
```diff
@@ -0,0 +1,63 @@
...
+	if f.nextStartingBlockHeight.Cmp(finalizedBlockHeight) >= 0 {
+		return nil, nil
+	}
+
+	// clamp to the max batch size. the range is inclusive so +1 when computing the count
+	endHeight := finalizedBlockHeight
```
tynes, line 49: This block of code might not be bad to break out into a function that can be unit tested

indexer/node/client.go
```diff
@@ -0,0 +1,129 @@
...
+	// defaultRequestTimeout is the default duration the processor will
+	// wait for a request to be fulfilled
+	defaultRequestTimeout = 10 * time.Second
+)
+
+type EthClient interface {
+	FinalizedBlockHeight() (*big.Int, error)
+	BlockHeadersByRange(*big.Int, *big.Int) ([]*types.Header, error)
+
+	// TODO: probably will remove this
```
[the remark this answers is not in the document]
hamdiallam: ah gotcha, let me cut some tickets for these todos

indexer/node/client.go
hamdiallam: sweet! Yea i left this as a TODO since i wans't sure exactly how I wanted to implement the retry functionality. Good to know op-service has this already

indexer/node/client.go
hamdiallam: ohhh, I didn't even realize the finalized block tag was supported via `-3`. Will reuse `ethclient` on some of this functionality

indexer/node/fetcher.go
hamdiallam: +1 I also hate using big int but makes sense to avoid future problems if we always directly convert to `uint64`

indexer/node/fetcher.go
hamdiallam: sounds good

roninjin10:
DX-50

roninjin10:
FE-1197 does it work for ecosystem team?

indexer/node/fetcher.go
hamdiallam: yea this is a good point

hamdiallam:
Updated so that the fetcher holds reference to the last header. Useful so that the fetcher can ensure the next batch is chained correctly on the last header

hamdiallam:
Copying @roninjin10 comment about batch sizes


> Just sharing context on how others do this:
>
> Ponder fetches an extremely aggressive amount at once and then if it fails backs off to more conservative amounts.
> 
> The graph does the opposite where it tries to get more aggressive as it goes

PR #5841, opened 2023-05-31
feat(proxyd): add support to rewrite debug_getRawReceipts

felipe-op:
**Description**

This change adds support for `debug_getRawReceipts` method.

**Tests**

Added new tests.

**Invariants**

We cache only requests that are block hash, block numbers should never be cached.

**Additional context**

https://www.notion.so/oplabs/Consensus-Aware-RPC-Proxy-0138e029af814e4cbca2740c7888e02d

**Metadata**

https://linear.app/optimism/issue/INF-257/add-debug-getrawreceipts-to-rewrite-methods

**TODOs**

### DIFF
--- a/proxyd/rewriter.go
+++ b/proxyd/rewriter.go
@@ -63,24 +63,26 @@ func RewriteRequest(rctx RewriteContext, req *RPCReq, res *RPCRes) (RewriteResul
 	case "eth_getLogs",
 		"eth_newFilter":
 		return rewriteRange(rctx, req, res, 0)
+	case "debug_getRawReceipts":
+		return rewriteParam(rctx, req, res, 0, true)
 	case "eth_getBalance",
 		"eth_getCode",
 		"eth_getTransactionCount",
 		"eth_call":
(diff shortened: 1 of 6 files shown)


proxyd/rewriter.go
```diff
@@ -89,9 +91,9 @@ func rewriteParam(rctx RewriteContext, req *RPCReq, res *RPCRes, pos int) (Rewri
 
 	// we assume latest if the param is missing,
 	// and we don't rewrite if there is not enough params
-	if len(p) == pos {
+	if len(p) == pos && !required {
```
mslipper, line 94: What does this mean?

mslipper (approved):
LGTM, but wondering what that `required` flag means.

proxyd/rewriter.go
felipe-op: it differentiates if the param is required or not

when the param is optional, we assume `latest` as the default value
when it is required, we just prevent rewriting anything

PR #5842, opened 2023-05-31
feat(op-challenger): DisputeGameFactory Watcher

refcell:
**Description**

Stacked ontop of #5831, this pr contains the logic for the `watch factory`
subcommand to listen for newly created dispute games from the dispute
game factory.

**Metadata**

Fixes CLI-4063

### DIFF
--- a/op-challenger/challenger/factory.go
+++ b/op-challenger/challenger/factory.go
@@ -31,8 +31,3 @@ func BuildDisputeGameLogFilter(contract *abi.ABI) (ethereum.FilterQuery, error) 
 
 	return query, nil
 }
-
-// BuildDisputeGameLogFilter creates a filter query for the DisputeGameFactory contract.
-func (c *Challenger) BuildDisputeGameLogFilter() (ethereum.FilterQuery, error) {
-	return BuildDisputeGameLogFilter(c.dgfABI)
-}
--- a/op-challenger/challenger/factory.go
+++ b/op-challenger/challenger/factory.go
@@ -1,0 +1,38 @@
+package challenger
+
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var ErrMissingFactoryEvent = errors.New("missing factory event")
+
+// BuildDisputeGameLogFilter creates a filter query for the DisputeGameFactory contract.
+//
+// The `DisputeGameCreated` event is encoded as:
+// 0: address indexed disputeProxy,
+// 1: GameType indexed gameType,
+// 2: Claim indexed rootClaim,
+func BuildDisputeGameLogFilter(contract *abi.ABI) (ethereum.FilterQuery, error) {
+	event := contract.Events["DisputeGameCreated"]
+
+	if event.ID == (common.Hash{}) {
+		return ethereum.FilterQuery{}, ErrMissingFactoryEvent
+	}
+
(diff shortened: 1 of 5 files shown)


refcell:
* develop
  * **PR #5831** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5831" target="_blank">Graphite</a> 
    * **PR #5842** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5842" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5842?utm_source=stack-comment).

op-challenger/challenger/factory.go
```diff
@@ -0,0 +1,38 @@
...
+		Topics: [][]common.Hash{
+			{event.ID},
+		},
+	}
+
+	return query, nil
+}
+
+// BuildDisputeGameLogFilter creates a filter query for the DisputeGameFactory contract.
+func (c *Challenger) BuildDisputeGameLogFilter() (ethereum.FilterQuery, error) {
```
ajsutton, line 36: Not sure this needs to be a method on `Challenger` given it just delegates to the standalone function.

op-challenger/cmd/watch/factory.go
```diff
@@ -0,0 +1,76 @@
...
+	metricsCfg := cfg.MetricsConfig
+	if metricsCfg.Enabled {
+		log.Info("starting metrics server", "addr", metricsCfg.ListenAddr, "port", metricsCfg.ListenPort)
+		go func() {
+			if err := m.Serve(ctx, metricsCfg.ListenAddr, metricsCfg.ListenPort); err != nil {
+				logger.Error("error starting metrics server", err)
+			}
+		}()
+		m.StartBalanceMetrics(ctx, logger, service.Client(), service.From())
+	}
```
ajsutton, line 56: I'm kind of confused why this subcommand is publishing balance metrics. Though I'm not entirely sure what the aim of this subcommand is exactly - it feels like a debugging kind of tool to just print events as they happen but then it has all this metrics infrastructure.

ajsutton (commented):
This code looks ok, but I'm struggling to understand what the actual requirements are here. That's probably just me lacking context, but it would certainly help me if there were a better description of what this does and what it's building towards in the PR description. Future coders wondering why this code is the way it is may also appreciate that. :)

op-challenger/challenger/factory.go
refcell: We need it to access the internal `c.dgfABI` field

ajsutton: It's only called from the `NewFactorySubscription` method on `Challenger` though so that already has access to `dgfABI`.  I don't _think_ we need to export either of these methods from the module.

op-challenger/cmd/watch/factory.go
```diff
@@ -0,0 +1,60 @@
...
+
+	service, err := challenger.NewChallenger(*cfg, logger, m)
+	if err != nil {
+		logger.Error("Unable to create the Challenger", "error", err)
+		return err
+	}
+
+	logger.Info("Listening for DisputeGameCreated events from the DisputeGameFactory contract", "dgf", cfg.DGFAddress.String())
+
+	subscription, err := service.NewFactorySubscription()
```
ajsutton, line 32: Same question about closing resources as on #5831

ajsutton (approved):
LGTM. Just the couple of questions about whether we can narrow the interface and if we need to close resources. Neither are a big deal though.

PR #5848, opened 2023-05-31
op-wheel: add cheat code command

tynes:
**Description**

Allows it to set into the state arbitrary code at an arbitrary account. This will make it very easy to patch contracts that contain immutables. Then we can patch the contracts to have different immutables without needing to update the proxies and worry about that.

We can also cheat the impl of the multisig to be a call forwarder so that we do not need to cheat so many storage slots and forget about particular slots.

### DIFF
--- a/op-wheel/cheat/cheat.go
+++ b/op-wheel/cheat/cheat.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/consensus/beacon"
 	"github.com/ethereum/go-ethereum/consensus/ethash"
 	"github.com/ethereum/go-ethereum/core"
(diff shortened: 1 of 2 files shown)


op-wheel/cheat/cheat.go
```diff
@@ -379,6 +379,13 @@ func SetBalance(addr common.Address, amount *big.Int) HeadFn {
 	}
 }
 
+func SetCode(addr common.Address, code string) HeadFn {
+	return func(headState *state.StateDB) error {
+		headState.SetCode(addr, common.FromHex(code))
```
protolambda, line 384: it should `code []byte`, otherwise it's going to panic on invalid strings

op-wheel/commands.go
```diff
@@ -271,6 +271,17 @@ var (
...
 	}
+	CheatSetCodeCmd = cli.Command{
+		Name: "code",
+		Flags: []cli.Flag{
+			DataDirFlag,
+			addrFlag("address", "Address to change code of"),
+			cli.StringFlag{Name: "code", Usage: "New code of the account"},
+		},
+		Action: CheatAction(false, func(ctx *cli.Context, ch *cheat.Cheater) error {
+			return ch.RunAndClose(cheat.SetCode(addrFlagValue("address", ctx), ctx.String("code")))
```
protolambda, line 282: Similar to `addrFlag` helper, we should define a `hexutils.Bytes` generic-flag creation helper, to make bytes-flags easy and safe (errors are handled as part of flag-parsing then, and don't cause panics).

protolambda (commented):
It works, but I'd strongly suggest using a more safe approach to flag parsing, so we don't get panics on bad inputs. See comments on how to resolve.

tynes:
Updated based on PR comments

protolambda (approved):

PR #5853, opened 2023-06-01
feat(ctb): Copy goerli Lib_AddressManager deployment to bedrock

maurelian:
**Description**

This change ensures that we're handling deployment artifacts the same between Goerli and Mainnet.

Will make it easier to delete the legacy contracts package in the future as well. 

After this commit both networks have the same number of deployment artifacts:

```
➜ for file in deployments/mainnet/*.json; do echo "$file"; done | wc -l
      19

➜ for file in deployments/goerli/*.json; do echo "$file"; done | wc -l
      19
```


tynes (changes requested):
Can we rename to `AddressManager.json` to be 1:1 with mainnet? The go tooling will break otherwise because duplicate file name

tynes:
We still need to delete the old file with the old name, then LGTM

maurelian:
OK. Changes this around so that it: 
1. Uses the new name
2. Copies the file instead of moving it (I think that might've broken something in the legacy package).

tynes (approved):

PR #5857, opened 2023-06-01
contracts-bedrock: simplify hh config for mainnet

tynes:
**Description**

hardhat barfs if we pass the zero key in, not sure why we do it everywhere. default to using the 'test test ...' key at index 0 if the private key is not passed instead. this is useful for mainnet migration rehearsals and since the mainnet contracts have already been deployed, unlikely this will be used again and if used again the private key env var MUST be passed anyways


refcell (approved):

refcell:
FYI @tynes you're just missing a `npx prettier --write hardhat.config.ts` run for this to pass

PR #5864, opened 2023-06-01
feat(proxyd): add consensus_getReceipts meta method

felipe-op:
**Description**

This change introduces a new meta method called `consensus_getReceipts`. 

It allows for clients to transparently call it and let proxyd decide what is the supported method for the actual selected backend.

**Tests**

New tests added.

**Invariants**

n/a

**Additional context**

This is part of the project Consensus Aware RPC Proxy: https://www.notion.so/oplabs/Consensus-Aware-RPC-Proxy-0138e029af814e4cbca2740c7888e02d

**Fixes**

https://linear.app/optimism/issue/INF-263/meta-method-consensus-getrawreceipt

**TODOs**
- [x] Author or reviewer has added an entry to the [current release notes draft][RND], if appropriate -- not applicable


proxyd/README.md
```diff
@@ -89,6 +89,59 @@ Cache use Redis and can be enabled for the following immutable methods:
...
+{
+  "jsonrpc":"2.0",
+  "id": 1,
+  "params": {
+    "blockNumberOrHash": "0xc6ef2fc5426d6ad6fd9e2a26abeab0aa2411b7ab17f30a99d3cb96aed1d1055b",
+    "transactions": [
+      "0xe670ec64341771606e55d6b4ca35a1a6b75ee3d5145a99d05921026d1527331",
+      "0x88df016429689c079f3b2f6ad39fa052532c56795b733da78a91ebe6a713944b"
+    ]
+  }
```
ajsutton, line 117: To bring my comment on this over from slack:
I wonder if we're better off assuming that there is some method to get receipts by block hash so we can send only the block hash, then have the op-node fall back to requesting individual receipts if there isn't actually one.
That saves us having to fetch the list of transaction hashes before we can make the call even though every provider we currently use only needs the block hash (either geth debug_getRawReceipts or alchemy alchemy_getTransactionReceipts). If none are available proxyd would just return a response indicating that.

felipe-op: absolutely, I'll change the interface and submit a new revision

ajsutton (commented):
Yeah I think that looks good.

ajsutton (approved):

mslipper (approved):

PR #5868, opened 2023-06-02
fix(docs/op-stack): Allow fresh loads to change pages

qbzzt:
I don't know what the problem was, but replacing all the files in `src/.vuepress/theme` and `src/.vuepress/style` with those from the community docs site, which does work, fixed it.

Removing the HTML comment fix it when using `yarn preview`, but not on Netlify. It looks like Netlify does some additional processing that sometimes breaks :-(


docs/op-stack/src/.vuepress/styles/index.styl
```diff
@@ -19,13 +19,13 @@ aside.sidebar {
 p.sidebar-heading {
   color: #323A43 !important;
   font-family: 'Open Sans', sans-serif;
-  font-weight: 600 !important;
```
smartcontracts, line 22: Should be OK to keep the current file as-is here

qbzzt: Resolved by using the current index.styl instead of the one from devdocs.

PR #5870, opened 2023-06-02
op-node: Add a histogram to report current peer scores

ajsutton:
**Description**

Reports the current peer scores in a histogram that is replaced with each update cycle so that it only shows the current score distribution, not observations of each score as it changes.  Preserving the history of observations in the histogram would effectively hide banned peer scores because they get disconnected so quickly and thus spend almost no time with such a low score but the rest of the time their connected would contribute to better score observations.  Similarly "good" peers have their observations smeared from 0 up to their current score, making it impossible to see the current score.

The component scores that make up the total are reported in separate labels. Getting these values requires returning the updated values from `SetScores` so that `SnapshotHook` has the full set of scores instead of just the gossip scores. The full set of scores is already loaded as part of applying the diff so can efficiently be returned.

**Additional context**

Need to do some more serious testing with how this works with prometheus histogram functions given the resetting nature of the histogram is a bit unusual. If it works well, we can then remove the older score band based metrics since that data can be derived from the histogram.

**Metadata**

- Fixes #[Link to Issue]

**TODOs**


ajsutton:
This does work - the heat map takes a bit of fiddling with options to make it work because most peers have a 0 score. Overtime they spread out a bit and will be more visible but there's still a heavy dominance of 0 scores in the data.

The catch is that when working with the data you need to avoid using `rate` even though you would typically use that with histograms (and Grafana's suggestions will insert it).

image

Will try just a plain histogram and see if rate can avoid the smear effect.

ajsutton:
It does work when using `rate` and a normal histogram so that's certainly simpler.

image

But you can get the same heat map from the existing metric and the values displayed in the tooltips are then the actual number of peers in that bucket:
image

ajsutton:
I think the two options we should consider are:
1. Continue with the band based scoring we have now, just update it to report the component values as well as the total
2. Use a standard histogram rather than resetting repeatedly. This will mean the current number of peers in each bucket isn't available but using `rate` you can still get a heat map that gives a good overview of where peer scores sit relative to each other. More specific details can be retrieved via RPC requests. Remove the existing band based scoring. This is then the simplest metric code.

I'm leaning towards option 2 just for the simplicity of the code. I've updated this PR to reflect that but can easily revert the last couple of transactions if we want to keep the band based scores or go with the resetting histogram to only show current peer scores.

op-node/metrics/metrics.go
```diff
@@ -452,9 +447,15 @@ func NewMetrics(procName string) *Metrics {
 
 // SetPeerScores updates the peer score [prometheus.GaugeVec].
 // This takes a map of labels to scores.
```
protolambda, line 449: Outdated doc comment.
Also worth adding that order of input slice doesn't matter

op-node/flags/p2p_flags.go
```diff
@@ -34,15 +34,6 @@ var (
 		Value:    "none",
 		EnvVar:   p2pEnv("PEER_SCORING"),
 	}
-	PeerScoreBands = cli.StringFlag{
-		Name: "p2p.score.bands",
```
protolambda, line 38: We might want to leave a no-op hidden flag for compatibility, so this change can be rolled out more easily.

protolambda (commented):
Changes LGTM, nice to have more detailed score metrics without metrics label churn of peerids.
One nit on doc comment, and I think we need to be more careful with the flag removal.

op-node/flags/p2p_flags.go
ajsutton: Re-added as a hidden option.

op-node/metrics/metrics.go
ajsutton: Updated comment.  I also renamed the metric to just `peer_scores` instead of `peer_scores_histogram` - the `_histogram` was only appended initially to not conflict with the existing metric but now that it's gone we can go with the simpler name.  The histogram already appends things like `_bucket` etc to the individual metrics involved.

op-node/p2p/node.go
```diff
@@ -117,7 +117,7 @@ func (n *NodeP2P) init(resourcesCtx context.Context, rollupCfg *rollup.Config, l
 			return fmt.Errorf("cannot init without extended peerstore: %w", err)
 		}
 		n.store = eps
-		n.scorer = NewScorer(rollupCfg, eps, metrics, setup.PeerBandScorer(), log)
+		n.scorer = NewScorer(rollupCfg, eps, metrics, log)
 		n.host.Network().Notify(&network.NotifyBundle{
 			ConnectedF: func(_ network.Network, conn network.Conn) {
 				n.scorer.OnConnect(conn.RemotePeer())
```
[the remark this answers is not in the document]
ajsutton: As far as I can tell semgrep is drunk...

PR #5873, opened 2023-06-02
feat(op-service): Retrying client

refcell:
**Description**

Refactors a `client.RPC`-compatible client wrapped with a
`backoff.Strategy`. This is used by the downstream `op-challenger` as
part of the log traversal client when iterating over all logs.


refcell:
* develop
  * **PR #5873** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5873" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5873?utm_source=stack-comment).

op-service/client/backoff.go
```diff
@@ -0,0 +1,82 @@
...
+	strategy      backoff.Strategy
+}
+
+// NewBackoffClient creates a new backoff client.
+// The backoff strategy is optional, if not provided, the default exponential backoff strategy is used.
+func NewBackoffClient(c InnerRPC, retries int, strategy ...backoff.Strategy) *backoffEthClient {
+	if len(strategy) == 0 {
+		strategy = []backoff.Strategy{ExponentialBackoff}
+	}
+	return &backoffEthClient{
```
tynes, line 41: Could this use the instrumented client internally?

trianglesphere: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/client/rpc.go
Or https://github.com/ethereum-optimism/optimism/blob/develop/op-node/sources/eth_client.go

refcell: Oh nice! Should I refactor retry.go then into op-service in this pr?

ajsutton: It probably is worth refactoring to make it a common thing. Note that op-program is using `maxAttempts = math.MaxInt` because it needs to keep retrying until it works - returning an error will result in a crash.

The max attempts could certainly be passed in as an argument to the retrying source though and then op-program could retry forever and challenger could use something more sensible.

refcell: > const maxAttempts = math.MaxInt // Succeed or die trying

🔥

op-service/client/backoff.go
```diff
@@ -0,0 +1,82 @@
...
+// backoffEthClient wraps a [InnerRPC] with a backoff strategy.
+type backoffEthClient struct {
+	c             InnerRPC
+	retryAttempts int
+	strategy      backoff.Strategy
+}
+
+// NewBackoffClient creates a new backoff client.
+// The backoff strategy is optional, if not provided, the default exponential backoff strategy is used.
+func NewBackoffClient(c InnerRPC, retries int, strategy ...backoff.Strategy) *backoffEthClient {
```
ajsutton, line 37: nit: I'd be tempted to call this `NewRetryingClient` since the key functionality it adds is the ability to retry. It just happens to use a backoff strategy as part of retrying.

op-program/host/host.go
```diff
@@ -21,10 +22,13 @@ import (
 	oppio "github.com/ethereum-optimism/optimism/op-program/io"
 	"github.com/ethereum-optimism/optimism/op-program/preimage"
 	opservice "github.com/ethereum-optimism/optimism/op-service"
+	opclient "github.com/ethereum-optimism/optimism/op-service/client"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 )
 
+const maxAttempts = math.MaxInt // Succeed or die trying
```
ajsutton, line 30: I'd probably rename this to `maxRPCRetries` since it doesn't have the context of being in a file dedicated to retrying to stuff anymore.

ajsutton (approved):
LGTM. Just a couple of suggested naming changes.

op-service/client/backoff.go
```diff
@@ -0,0 +1,82 @@
...
+
+var (
+	// ExponentialBackoff is the default backoff strategy.
+	ExponentialBackoff = backoff.Exponential()
+)
+
+// InnerRPC is a minimal [client.RPC] interface that is used by the backoff client.
+//
+//go:generate mockery --name InnerRPC --output ./mocks/
+type InnerRPC interface {
```
trianglesphere, line 21: This is the same interface (& serves the same purpose) as https://github.com/ethereum-optimism/optimism/blob/d5ea28af16d92b2a18442d71736b5fba23757033/op-node/client/rpc.go#L21. Fine for now, but it should be removed when we do a larger cleanup of the sources.

op-service/client/backoff.go
refcell: That's good with me :)

PR #5884, opened 2023-06-02
op-chain-ops: Add additional logging to legacy balances migration

sebastianst:


op-chain-ops/ether/migrate.go
```diff
@@ -120,6 +121,7 @@ func doMigration(mutableDB *state.StateDB, dbFactory util.DBFactory, addresses [
 			// why we may have to filter out duplicates.
 			if seenAccounts[account.address] {
 				log.Info("skipping duplicate account during iteration", "addr", account.address)
```
maurelian, line 123: I wonder how noisey this might be if we think we might be skipping many thousand accounts?
I'd suggest getting the CSVs from Mark and testing this.

op-chain-ops/ether/migrate.go
```diff
@@ -198,7 +200,25 @@ func doMigration(mutableDB *state.StateDB, dbFactory util.DBFactory, addresses [
...
+		"diff_count", len(addresses)-count,
+		"diff_total", len(addresses)-(count+dups),
+	)
+
+	// Print first 10 accounts without balance
+	aleft := 10
+	log.Info("Listing first %d accounts without balance", aleft)
+	for i, a := range addresses {
+		if !seenAccounts[a] {
+			log.Info("Account[%d] without balance", i, "addr", a)
```
tynes, line 215: decrementing needs to go inside of the branch, then lgtm

tynes (approved):

PR #5886, opened 2023-06-02
contracts-bedrock: add withdrawal network and make withdrawal minimum configurable

dapperscene6:
**Description**
Modify `L1FeeVault`, `BaseFeeVault`, and `SequencerFeeVault` to have a configurable minimum withdrawal amounts and to be configured to withdraw to L1 or L2.

This changes adds deploy-configuration parameters to already deployed chains which leads to said chains current deploy-configurations no longer being able to recreate the genesis state of said already deployed chains.

Hexadecimals were used for `MinimumWithdrawalAmount` configuration in `deploy-configuration` for consistency though @smartcontracts and I discussed changing all `deploy-configuration` hexadecimals to decimal to increase `deploy-configuration` readability.

Running `make bindings` in `op-binding` created a larger diff than expected which I believe is causing some of the CI tests   to fail.

**Tests**
I've added solidity tests for the smart contract changes and golang tests for the golang changes.

**Invariants**

TBD

**Additional context**

These changes are necessary to enable daily revenue share on L2 as well as to reduce withdrawal fees on L1 by aggregating `FeeVault` earning prior to withdrawing to L1.

**Metadata**

- Fixes #[Link to Issue]

**TODOs**


packages/contracts-bedrock/src/deploy-config.ts
```diff
@@ -136,19 +136,49 @@ interface RequiredDeployConfig {
   proxyAdminOwner: string
 
   /**
-   * L1 address which receives the base fee for the L2 network.
+   * L1 or L2 address which receives the base fee for the L2 network.
```
tynes, line 139: Local or remote

tynes:
Not your fault but this is way too large of a diff for a change like this. Its the fault of how the codebase is architected. I'm thinking of ways to reduce friction in the future so that diffs can be smaller with these sorts of changes

tynes:
I was able to reproduce 7e22976 locally

packages/contracts-bedrock/contracts/universal/FeeVault.sol
```diff
@@ -65,12 +88,16 @@ abstract contract FeeVault {
...
-        emit Withdrawal(value, RECIPIENT, msg.sender);
+        emit Withdrawal(value, RECIPIENT, msg.sender, WITHDRAWAL_NETWORK);
 
-        L2StandardBridge(payable(Predeploys.L2_STANDARD_BRIDGE)).bridgeETHTo{ value: value }(
-            RECIPIENT,
-            WITHDRAWAL_MIN_GAS,
-            bytes("")
-        );
+        if (WITHDRAWAL_NETWORK == WithdrawalNetwork.L2) {
+            SafeCall.call(RECIPIENT, gasleft(), value, bytes(""));
```
tynes, line 94: Should use `SafeCall.send`
https://github.com/ethereum-optimism/optimism/blob/75c00a5091e43cc50d9b33381a166c68c6a465cd/packages/contracts-bedrock/contracts/libraries/SafeCall.sol#L17

packages/contracts-bedrock/deploy-config/mainnet.json
```diff
@@ -22,6 +22,12 @@
   "baseFeeVaultRecipient": "0xa3d596EAfaB6B13Ab18D40FaE1A962700C84ADEa",
   "l1FeeVaultRecipient": "0xa3d596EAfaB6B13Ab18D40FaE1A962700C84ADEa",
   "sequencerFeeVaultRecipient": "0xa3d596EAfaB6B13Ab18D40FaE1A962700C84ADEa",
+  "baseFeeVaultMinimumWithdrawalAmount": '0x8ac7230489e80000',
```
tynes, line 25: This needs to be valid JSON

packages/contracts-bedrock/src/deploy-config.ts
dapperscene6: @tynes could you clarify what you're suggesting? Do you mean changing these comments to be: `Local or remote address which receives the base fee for the L2 network.`?

op-chain-ops/genesis/testdata/test-deploy-config-devnet-l1.json
```diff
@@ -22,6 +22,14 @@
   "baseFeeVaultRecipient": "0xBcd4042DE499D14e55001CcbB24a551F3b954096",
   "l1FeeVaultRecipient": "0x71bE63f3384f5fb98995898A86B02Fb2426c5788",
   "sequencerFeeVaultRecipient": "0x71bE63f3384f5fb98995898A86B02Fb2426c5788",
+  "baseFeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
+  "l1FeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
+  "sequencerFeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
+  "baseFeeVaultWithdrawalNetwork": 0,
+  "l1FeeVaultWithdrawalNetwork": 0,
+  "sequencerFeeVaultWithdrawalNetwork": 0,
+
```
tynes, line 31: nit: newlines

packages/contracts-bedrock/contracts/test/SequencerFeeVault.t.sol
```diff
@@ -11,16 +12,25 @@ contract SequencerFeeVault_Test is Bridge_Initializer {
     SequencerFeeVault vault = SequencerFeeVault(payable(Predeploys.SEQUENCER_FEE_WALLET));
     address constant recipient = address(256);
```
tynes, line 14: The abi change here is breaking backwards compatibility. I recommend we introduce a second event and emit the 2 events for some amount of time and then sunset the legacy event

packages/contracts-bedrock/contracts/test/SequencerFeeVault.t.sol
```diff
@@ -73,4 +88,33 @@ contract SequencerFeeVault_Test is Bridge_Initializer {
         // The withdrawal was successful
         assertEq(vault.totalProcessed(), amount);
     }
+
+    function test_withdraw_toL2_succeeds() external {
```
tynes, line 92: This feels like it should in another test contract with the asymmetry of the `vm.etch` happening in `setUp` vs a test fn

packages/contracts-bedrock/contracts/test/SequencerFeeVault.t.sol
```diff
@@ -73,4 +88,33 @@ contract SequencerFeeVault_Test is Bridge_Initializer {
...
+            FeeVault.WithdrawalNetwork.L2
+        );
+
+        // The entire vault's balance is withdrawn
+        vm.expectCall(recipient, address(vault).balance, bytes(""));
+
+        vault.withdraw();
+
+        // The withdrawal was successful
+        assertEq(vault.totalProcessed(), amount);
```
tynes, line 118: A balance check on the vault having 0 and the recipient having the balance would be good here

packages/contracts-bedrock/contracts/universal/FeeVault.sol
```diff
@@ -40,12 +47,28 @@ abstract contract FeeVault {
...
-     * @param _recipient           Wallet that will receive the fees on L1.
-     * @param _minWithdrawalAmount Minimum balance before a withdrawal can be triggered.
+     * @notice Emitted each time a withdrawal occurs.
+     *
+     * @param value             Amount that was withdrawn (in wei).
+     * @param to                Address that the funds were sent to.
+     * @param from              Address that triggered the withdrawal.
+     * @param withdrawalNetwork Network which the to address will receive funds on.
+     */
+    event Withdrawal(uint256 value, address to, address from, WithdrawalNetwork withdrawalNetwork);
```
tynes, line 57: We should have 2 withdrawal events to stay backwards compatible, then we can announce the sunset of the old one and remove it after some time. This will prevent existing dashboards from breaking, so that they have time to upgrade

packages/contracts-bedrock/src/deploy-config.ts
tynes: Yes exactly, I think using local or remote is better language because it generalizes to L3+. Its def not consistent throughout the codebase

packages/contracts-bedrock/contracts/test/SequencerFeeVault.t.sol
refcell: Agree here

packages/contracts-bedrock/deploy-config/getting-started.json
```diff
@@ -29,9 +29,16 @@
   "finalizationPeriodSeconds": 12,
 
   "proxyAdminOwner": "ADMIN",
+
   "baseFeeVaultRecipient": "ADMIN",
   "l1FeeVaultRecipient": "ADMIN",
   "sequencerFeeVaultRecipient": "ADMIN",
+  "baseFeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
```
refcell, line 36: nit: split into a new group
```suggestion

  "baseFeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
```

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -32,6 +32,12 @@
   "baseFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
   "l1FeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
   "sequencerFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "baseFeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
```
refcell, line 35: ibid

packages/contracts-bedrock/src/deploy-config.ts
refcell: If I were to read "Local or remote address", personally I think I'd be very confused without context. I think a clearer and still more flexible expression may be something like
```
L1 or higher (eg L2) address which receives the base fee for the L2 network.
```

refcell (changes requested):
Mostly echoing @tynes comments here, otherwise almost there!

packages/contracts-bedrock/deploy-config/getting-started.json
```diff
@@ -29,10 +29,18 @@
   "finalizationPeriodSeconds": 12,
 
   "proxyAdminOwner": "ADMIN",
+
```
refcell, line 32: nit: think this newline (between `proxyAdminOwner` and `baseFeeVaultRecipient` was added by mistake
```suggestion
```

packages/contracts-bedrock/deploy-config/goerli.json
```diff
@@ -29,10 +29,18 @@
   "finalizationPeriodSeconds": 12,
 
   "proxyAdminOwner": "0xf80267194936da1E98dB10bcE06F3147D580a62e",
+
```
refcell, line 32: Oh meant to add the newline down below but just removing this now works 👍 
```suggestion
```

refcell (changes requested):
Just two small corrections and I believe we're gtm

refcell (approved):

tynes (approved):

tynes:
This looks good to me!

packages/contracts-bedrock/contracts/universal/FeeVault.sol
```diff
@@ -11,24 +12,30 @@ import { Predeploys } from "../libraries/Predeploys.sol";
...
     /**
-     * @notice Wallet that will receive the fees on L1.
+     * @notice Wallet that will receive the fees.
      */
     address public immutable RECIPIENT;
 
+    /**
+     * @notice Network which the RECIPIENT will receive fees on.
+     */
+    WithdrawalNetwork public immutable WITHDRAWAL_NETWORK;
```
kahuang, line 37: If we want to upgrade existing fee vault contracts, does this variable need to be moved below the `totalProcessed` variable, to preserve storage slots?

tynes: Immutables are inlined in the bytecode, it's not in storage so this doesn't change the storage layout

PR #5899, opened 2023-06-06
op-node: Add scoring for p2p sync request response

ajsutton:
**Description**

Introduce peer scoring for p2p sync req/resp handling.

Currently this is just a spike to see how things can be fit together and demonstrate the overall design.

Key parts:
 * Introduce `peerApplicationScorer` which is responsible for all the details of adjusting and combining non-gossip peer score components
 * The single `ReqRespSync` score component is replaced by a `ReqRespScores` object so we can track individual components. Test confirm this is backwards compatible.
 * Broke the peer store interface a bit by introducing `ReqRespScoresDiff` as an implementation of `ScoreDiff` that takes a function to update the scores. This means the specific changes made now leak outside of the module where previously they were private. It makes application scores much simpler though because the `SetScores` method already loads the previous value, applies the diff then saves it in a thread safe way. This lets us hook into that process to increment the scores as needed without having to add additional thread safety logic.
   *  Currently this passes through the full `PeerScores` struct, but it could be modified to have a separate object for application scores so the gossip scores controlled by libp2p aren't editable. Not sure its worth it though.

Also need to hook this into the request handling side to penalise peers when they hit rate limits and hook up banning for garbage responses (possibly should replace the `ErrorMessage` scores here but need to dig more into the details of what causes errors).

**Tests**

**Invariants**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

**TODOs**

### DIFF
--- a/op-node/flags/p2p_flags.go
+++ b/op-node/flags/p2p_flags.go
@@ -73,6 +73,15 @@ var (
 		Value:    "none",
 		EnvVar:   p2pEnv("TOPIC_SCORING"),
 	}
+
+	ApplicationScoring = cli.StringFlag{
+		Name: "p2p.scoring.application",
+		Usage: "Sets the application scoring strategy. " +
+			"Can be one of: none or light.",
+		Required: false,
+		Value:    "none",
+		EnvVar:   p2pEnv("APPLICATION_SCORING"),
+	}
 	P2PPrivPath = cli.StringFlag{
 		Name: "p2p.priv.path",
 		Usage: "Read the hex-encoded 32-byte private key for the peer ID from this txt file. Created if not already exists." +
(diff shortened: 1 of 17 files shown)


ajsutton:
Suggestion from Proto here is to introduce two different diff types to the scorebook:
* Increment - adds one to whichever field is specified, used for when an event happens
* Decay - given the appropriate decay rates will decay each field

This preserves the atomic nature of the change but more tightly defines the actually diff rather than just calling out to a function to make whatever changes it wants.  The scorebook will know about decay, but just takes the rates as inputs rather than having to own the application scoring params itself.

ajsutton:
This has now all been done via other, more polished PRs.

PR #5910, opened 2023-06-07
feat(indexer) processor module with basic l1Processor & l2Processor

hamdiallam:
Introduce a `processor` modules that exposes `L1Processor` and `L2Processors`. These structs define the core logic that indexes new batches retrieved from the fetcher. All database operations for a given batch are wrapped in a database transaction.

The L1 & L2 processors simply index block headers right now.

## Aside
- `indexer-refresh` command was updated such that the new indexer can now be run e2e.
- `EthClient#FinalizedBlockHeight` switched to "latest" blocktag instead of "finalized" blocktag to avoid dealing with issues with the local devnet. We'll address this
- L1Processor needs some additional config such that it doesn't start from genesis. It should start from the L1 block corresponding to L2 genesis

### DIFF
--- a/indexer/database/blocks.go
+++ b/indexer/database/blocks.go
@@ -46,8 +46,8 @@ type LegacyStateBatch struct {
 }
 
 type BlocksView interface {
-	FinalizedL1BlockHeight() (*big.Int, error)
-	FinalizedL2BlockHeight() (*big.Int, error)
+	FinalizedL1BlockHeader() (*L1BlockHeader, error)
+	FinalizedL2BlockHeader() (*L2BlockHeader, error)
 }
 
 type BlocksDB interface {
@@ -80,9 +80,6 @@ func (db *blocksDB) StoreL1BlockHeaders(headers []*L1BlockHeader) error {
 }
 
 func (db *blocksDB) StoreLegacyStateBatch(stateBatch *LegacyStateBatch) error {
-	// Even though transaction control flow is managed, could we benefit
-	// from a nested transaction here?
-
 	result := db.gorm.Create(stateBatch)
 	if result.Error != nil {
 		return result.Error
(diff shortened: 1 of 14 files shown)


indexer/database/contract_events.go
```diff
@@ -29,8 +29,6 @@ type L2ContractEvent struct {
 }
 
 type ContractEventsView interface {
-	L1ContractEventByGUID(string) (*L1ContractEvent, error)
```
hamdiallam, line 32: Removing these since there's new use for them. Would rather add as-needed

indexer/database/blocks.go
```diff
@@ -111,14 +108,19 @@ func (db *blocksDB) StoreLegacyStateBatch(stateBatch *LegacyStateBatch) error {
...
 
-func (db *blocksDB) FinalizedL1BlockHeight() (*big.Int, error) {
+// FinalizedL1BlockHeader returns the latest L1 block header stored in the database, nil otherwise
+func (db *blocksDB) FinalizedL1BlockHeader() (*L1BlockHeader, error) {
 	var l1Header L1BlockHeader
 	result := db.gorm.Order("number DESC").Take(&l1Header)
 	if result.Error != nil {
+		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+			return nil, nil
+		}
```
roninjin10, line 118: Why is this necessary?
```suggestion
```

indexer/processor/processor.go
```diff
@@ -0,0 +1,68 @@
...
+type processor struct {
+	fetcher *node.Fetcher
+
+	db         *database.DB
+	processFn  processFn
+	processLog log.Logger
+}
+
+// Start kicks off the processing loop. This is a blocking operation and should be run within its own goroutine
+func (p processor) Start() {
```
roninjin10, line 28: What does op-node use for this?

roninjin10 (commented):
need to refresh on gorm later today to give proper review

roninjin10 (commented):
need to refresh on gorm later today to give proper review

indexer/database/blocks.go
hamdiallam: Since we are returning pointers, I think it makes sense to have `nil` represent an non-existent entry when `err == nil`. Useful for the processor when determining if it should start from genesis or a specified height (last indexed block).

Gorm treats non existent entries as errors

indexer/processor/processor.go
hamdiallam: Good question, not entirely sure cc @tynes

hamdiallam:
DX-80

indexer/processor/processor.go
hamdiallam: Looked into op-node. op-node/ defines a driver for walking the L1 chain. It's opinionated and specifically build for the op-node use case. This is simple enough for us to define here

PR #5911, opened 2023-06-07
feat(op-challenger): Parse Logs

refcell:
**Description**

Adds logic to the `op-challenger` allowing it to parse `OutputProposed`
logs using the l2ooABI. This is required downstream in the dispatch
logic to pass the expected output and block number to the validation
function.


refcell:
* develop
  * **PR #5911** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5911" target="_blank">Graphite</a>  👈
    * **PR #5912** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5912" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5911?utm_source=stack-comment).

op-challenger/challenger/output.go
```diff
@@ -15,8 +17,21 @@ var (
 	ErrInvalidBlockNumber = errors.New("invalid block number")
 	// ErrUnsupportedL2OOVersion is returned when the output version is not supported.
 	ErrUnsupportedL2OOVersion = errors.New("unsupported l2oo version")
+	// ErrInvalidOutputLogTopic is returned when the output log topic is invalid.
+	ErrInvalidOutputLogTopic = errors.New("invalid output log topic")
 )
 
+// ParseOutputLog parses a log from the L2OutputOracle contract.
+func (c *Challenger) ParseOutputLog(log *types.Log) (*big.Int, eth.Bytes32, error) {
```
trianglesphere, line 25: I think returning a struct here would be helpful to make sure that these value have names & don't get separated.

op-challenger/challenger/output_test.go
```diff
@@ -10,13 +10,49 @@ import (
...
+
+func TestParseOutputLog_WrongLogTopic_Errors(t *testing.T) {
+	challenger := newTestChallenger(t, eth.OutputResponse{}, true)
+	_, _, err := challenger.ParseOutputLog(&types.Log{
+		Topics: []common.Hash{{0x01}, {0x02}, {0x03}, {0x04}},
+	})
+	require.ErrorIs(t, err, ErrInvalidOutputLogTopic)
+}
+
+func TestParseOutputLog_BadLog_Panics(t *testing.T) {
```
trianglesphere, line 45: We should not panic in parsing code, please return an error instead. Note that we don't always need named errors unless we expect the end user of this API to inspect the retuned error & do something based on the specific (rather than presence) error value.

op-challenger/challenger/output_test.go
```diff
@@ -10,13 +10,49 @@ import (
...
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
 
+	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
 	"github.com/ethereum-optimism/optimism/op-challenger/metrics"
 	"github.com/ethereum-optimism/optimism/op-node/eth"
 	"github.com/ethereum-optimism/optimism/op-node/testlog"
 )
 
+func TestParseOutputLog_Succeeds(t *testing.T) {
```
trianglesphere, line 23: Can we test against a real log?

refcell: I'm not really sure how best to do this. Is there a way to encode a sol `emit OutputProposed(...)` using the abi encoder?

trianglesphere: hmm, it doesn't look like there is an easy way to do this. We could run the EVM in memory. Another option is to hook this up to an e2e test. I think this log looks correct, but I'm not gonna be 100% sure until I see it running in the full system

op-challenger/challenger/output.go
```diff
@@ -15,38 +17,67 @@ var (
...
 	ErrUnsupportedL2OOVersion = errors.New("unsupported l2oo version")
+	// ErrInvalidOutputLogTopic is returned when the output log topic is invalid.
+	ErrInvalidOutputLogTopic = errors.New("invalid output log topic")
+	// ErrInvalidOutputTopicLength is returned when the output log topic length is invalid.
+	ErrInvalidOutputTopicLength = errors.New("invalid output log topic length")
 )
 
+// OutputProposal is a proposal for an output root
+// in the L2OutputOracle for a given L2 block number.
+type OutputProposal struct {
```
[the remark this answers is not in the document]
trianglesphere: the type with the bindings also includes the raw log.

op-challenger/challenger/output_test.go
refcell: I think this is sufficiently tested to continue work ontop in trunk for now

trianglesphere (approved):

PR #5914, opened 2023-06-07
Fix indexer release action

mslipper:


zhwrd (changes requested):
looks good but also needs to be added to the `hold` job on circleci conf line 1532

mslipper:
Nice catch, fixed.

zhwrd (approved):

roninjin10 (approved):

PR #5923, opened 2023-06-08
Fixes #5703

### LINKED ISSUE #5703: BUG: the .sample.env should have DISABLE_LIVE_DEPLOYER=false
**Describe the bug**

The .sample.env in packages/bedrock-contracts/ folder should have DISABLE_LIVE_DEPLOYER=false instead of true. This is causing the deploy contracts script to get stuck.

**To Reproduce**
In https://stack.optimism.io/docs/build/getting-started/#deploy-the-l1-contracts, trying the "npx hardhat deploy --network getting-started --tags l1" command get stuck.

**Expected behavior**
 The script should finish deploying the contracts and run the l1 rpc.

**Screenshots**
Screenshot 2023-05-13 at 17 38 52

**System Specs:**
 - OS: MacOs
 - Package Version (or commit hash): 4611198bf8bfd16563cc6bdf49bb35eed2e46801

**Additional context**

chrisciszak:
**Description**

Fixes #5703

process.env will always output strings, !! in typescript will convert any non empty string to true. Adding a proper check for this resolves this problem. Also, changed the env.sample to reflect the already existing boolean logic which is easier to maintain.

Another issue that was causing the hang was related to the difference in the case that is possible to arise.

### DIFF
--- a/packages/contracts-bedrock/src/deploy-utils.ts
+++ b/packages/contracts-bedrock/src/deploy-utils.ts
@@ -364,7 +364,7 @@ export const doOwnershipTransfer = async (opts: {
  */
 export const liveDeployer = async (opts: {
   hre: HardhatRuntimeEnvironment
-  disabled: string | undefined
+  disabled: boolean | undefined
 }): Promise<boolean> => {
   if (!!opts.disabled) {
     console.log('Live deployer manually disabled')
--- a/packages/contracts-bedrock/.env.example
+++ b/packages/contracts-bedrock/.env.example
@@ -18,7 +18,7 @@ CAST_COMMANDS=1
 SAVE_DEPLOYMENTS=1
 
 # Disable the live deployer so that transactions must be submitted manually
-DISABLE_LIVE_DEPLOYER=true
+DISABLE_LIVE_DEPLOYER=0
 
 # Sets the deployer's key to match the first default hardhat account
 PRIVATE_KEY_DEPLOYER=ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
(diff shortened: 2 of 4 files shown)


packages/contracts-bedrock/deploy/019-SystemDictatorInit.ts
```diff
@@ -176,7 +176,7 @@ const deployFn: DeployFunction = async (hre) => {
   if (
     (await SystemDictatorProxy.callStatic.admin({
       from: ethers.constants.AddressZero,
-    })) !== hre.deployConfig.controller
```
tynes, line 179: This is an unrelated change, please remove it and propose in its own PR

packages/contracts-bedrock/src/deploy-utils.ts
```diff
@@ -366,7 +366,7 @@ export const liveDeployer = async (opts: {
   hre: HardhatRuntimeEnvironment
   disabled: string | undefined
 }): Promise<boolean> => {
-  if (!!opts.disabled) {
```
tynes, line 369: Perhaps parse the env var into a bool higher up the call stack and have `disabled` be a bool arg

chrisciszak: @tynes done

packages/contracts-bedrock/deploy/019-SystemDictatorInit.ts
chrisciszak: moved to a new PR https://github.com/ethereum-optimism/optimism/pull/5993

upnodedev:
We have tried your fix and it still doesn't work until I comment these lines out

image

Credit: https://github.com/ethereum-optimism/optimism/issues/5703#issuecomment-1556188148

chrisciszak:
The fix works if you run initial deployment.

You shouldn't need to comment this.

If it failed already, then you need this too: https://github.com/ethereum-optimism/optimism/pull/5993

PR #5926, opened 2023-06-08
integration-tests: deprecate

tynes:
**Description**

The legacy integration tests are no longer required. Delete the directory. This commit is helpful because it will reduce the amount of tests that need to run in CI.

The legacy integration tests ran against the `ops` docker compose setup, which brought up the legacy system.


trianglesphere (approved):
bedrock ftw

PR #5931, opened 2023-06-08
op-e2e: Fix bug in TestWithdrawals

jyellick:
The existing TestWithdrawals test attempts to assert that the total fees of the prove and finalize transactions are accounted for in the L1 balance.  But, it re-uses the `calcGasFees` function and passes in the `GasTipCap` and the `GasFeeCap` from the L2 transaction, not from the actual L1 transactions.  In the event that the gas fee differs on the L2 from the L1, then the test will fail.

To keep the diff small, this change simply multiplies the GasUsed and `EffectiveGasPrice` from the receipts.  This could be moved into a helper function like `calcGasFees`.  Using the `EffectiveGasPrice` seems just as good or better than relying on the parameters of the original transaction, and could be similarly utilized higher in the test for the L2 fees.  If we really want to get the parameters from the tx, then `ProveAndFinalizeWithdrawal` would need to be modified to return the txes.

FWIW, I encountered this when using Erigon for the L2 nodes, which compute the SuggestedGasTip slightly differently from Geth resulting in a discrepancy between the L1 and L2 gas tips, but this could conceivably cause flakiness even when using Geth.

### DIFF
--- a/op-e2e/system_test.go
+++ b/op-e2e/system_test.go
@@ -1086,11 +1086,6 @@ func TestWithdrawals(t *testing.T) {
 	// Verify balance after withdrawal
 	ctx, cancel = context.WithTimeout(context.Background(), 1*time.Second)
 	defer cancel()
-	header, err = l1Client.HeaderByNumber(ctx, finalizeReceipt.BlockNumber)
-	require.Nil(t, err)
-
-	ctx, cancel = context.WithTimeout(context.Background(), 1*time.Second)
-	defer cancel()
 	endBalance, err = l1Client.BalanceAt(ctx, fromAddr, nil)
 	require.Nil(t, err)
 
@@ -1098,7 +1093,9 @@ func TestWithdrawals(t *testing.T) {
 	// Fun fact, the fee is greater than the withdrawal amount
 	// NOTE: The gas fees include *both* the ProveWithdrawalTransaction and FinalizeWithdrawalTransaction transactions.
 	diff = new(big.Int).Sub(endBalance, startBalance)
-	fees = calcGasFees(proveReceipt.GasUsed+finalizeReceipt.GasUsed, tx.GasTipCap(), tx.GasFeeCap(), header.BaseFee)
+	proveFee := new(big.Int).Mul(new(big.Int).SetUint64(proveReceipt.GasUsed), proveReceipt.EffectiveGasPrice)
+	finalizeFee := new(big.Int).Mul(new(big.Int).SetUint64(finalizeReceipt.GasUsed), finalizeReceipt.EffectiveGasPrice)
+	fees = new(big.Int).Add(proveFee, finalizeFee)
 	withdrawAmount = withdrawAmount.Sub(withdrawAmount, fees)
 	require.Equal(t, withdrawAmount, diff)
 }


trianglesphere:
hmm, `proveReceipt` & `finalizedReceipt` are pulled from L1, not L2.

How is the difference in SuggestedGasTip between erigon/geth showing up here?

op-e2e/system_test.go
```diff
@@ -1098,7 +1098,9 @@ func TestWithdrawals(t *testing.T) {
 	// Fun fact, the fee is greater than the withdrawal amount
 	// NOTE: The gas fees include *both* the ProveWithdrawalTransaction and FinalizeWithdrawalTransaction transactions.
 	diff = new(big.Int).Sub(endBalance, startBalance)
-	fees = calcGasFees(proveReceipt.GasUsed+finalizeReceipt.GasUsed, tx.GasTipCap(), tx.GasFeeCap(), header.BaseFee)
```
jyellick, line 1101: Here, in the original code, we are invoking `calcGasFees` with `proveReceipt` (L1), `finalizeReceipt` (L1), and the gas parms froom `tx` (L2).

So, when the SuggestedTipPrice for the L2 Tx, and the L1 Tx differ, (if using Erigon for the L2, and Geth for the L1), then `calcGasFees` will return the L1 Gas Usage * the L2 Gas cost, which is wrong.

jyellick: I can't comment there as it's not part of the diff, but line 1055 is where `tx` is set (and it is never overridden), and it's an L2 transaction.

We use it in `calcGasFees` on line 1073 with the receipt from L2 which is correct.  Then, again here at 1101 with the L1 receipts, which is the bug.

jyellick:
Maybe the Erigon comment was more confusing than it was helpful.  To be clear, the bug is that the fee computation is using the gas usage from L1, with the gas pricing from an L2 tx.  That is the bug this PR fixes.

The test currently passes because it is using a Geth for both L1 and L2 nodes, which returns the same SuggestedGasTip (1 Gwei) when there is no tip history to infer from.  Because the SuggestedGasTip is the same for both L1 and L2, the actual GasTip for the L2 transaction ends up being the same as the GasTip for the L1 transaction.  Therefore the test passes even though we're mixing up the L1 and L2 GasTips.

When using Erigon for the L2 nodes, the SuggestedGasTip returns a different suggested gas tip (0 Gwei).  So, now the GasTip on the L1 txes (1 Gwei) is not the same as the GasTip on the L2 tx (0 Gwei), so the test fails.  Erigon is not breaking anything, it's just the test's assertion that has the bug.

Personally, I'm not entirely sure why `calcGasFees` takes the parameters from the tx (tip and fee cap), the receipt (gas used), and the header (baseFee) to compute the gas fees.  We could instead simply take the gas used and effective gas price from the receipt as this change does.  To my understanding, Optimism does not modify the EIP-1559 logic in any way, so I'm not sure what value we're getting by manually computing the gas fee ourselves.  I'd be happy to modify `calcGasFees` and its callers if you agree.

ajsutton (approved):
Funnily enough, hit this with the latest op-geth as well because it changes the fee estimation for OP networks so L1 and L2 wind up returning different GasTips (https://github.com/ethereum-optimism/optimism/pull/5938).

The test is confirming that the L1 contract transfers the correct amount of ETH so I agree its reasonable to trust the `EffectiveGasPrice` from the receipt rather than having to calculate from the transaction.

Thanks.

PR #5932, opened 2023-06-08
cleanup: delete l2geth-exporter

tynes:
**Description**

Removes the `l2geth-exporter` package from the monorepo. This exposed metrics for the legacy system and is no longer necessary.


refcell (commented):
Looks good, but I think @protolambda wanted to copy the entire monorepo to a new repo to isolate + freeze pre-bedrock code. That should probably be done before this is merged.

refcell:
Update: this will be part of https://linear.app/optimism/issue/CLI-4053/monorepo-post-migration-cleanup

zhwrd (approved):

PR #5937, opened 2023-06-08
Make optimism portal address overridable in withdrawal-mon

kahuang:
This makes withdrawal-mon accept an override for the OptimismPortal contract so it can run against other op-chains.

**Tests**

I tested this against a Conduit op-chain, and saw it process a valid withdrawal.

**Invariants**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

**TODOs**


packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -98,12 +104,35 @@ export class WithdrawalMonitor extends BaseServiceV2<Options, Metrics, State> {
       name: 'L2',
     })
 
-    this.state.messenger = new CrossChainMessenger({
-      l1SignerOrProvider: this.options.l1RpcProvider,
-      l2SignerOrProvider: this.options.l2RpcProvider,
-      l1ChainId: await getChainId(this.options.l1RpcProvider),
-      l2ChainId: await getChainId(this.options.l2RpcProvider),
-    })
+    if (this.options.optimismPortalAddress) {
```
tynes, line 107: This sort of branching logic isn't going to scale well. In theory any address should be able to be overridden

tynes: Are you planning on following up with this? I believe the best way to do this is that there is a networks object in the sdk where you can go from chain id to the `contracts` object, then you can overwrite the portal address on that object and use it to instantiate the cross chain messenger

tynes:
@kahuang Are you planning on updating this based on the feedback?

packages/chain-mon/src/wd-mon/service.ts
kahuang: Right I guess if the chain id is one the SDK doesn't know about the addresses just come back as 0, or does it throw an error?

The only address that needs to be overriden here is portal, because its the only one that's used. Would you just prefer I make the "l1" addresses outside the if? I kinda just copied the same pattern from https://github.com/ethereum-optimism/optimism/blob/c7949cb3651c4ab07e44b26b1f2f95b795dc5e52/packages/fault-detector/src/service.ts#L130

PR #5940, opened 2023-06-09
op-service: Update retryingClient to handle individual elements in a batch failing

ajsutton:
**Description**

Update retryingClient to handle individual elements in a batch failing. Previously it would only retry if there was an IO error with the batch request. Errors such as hitting Alchemy request limits are reported as errors on the individual BatchElem and weren't triggering a retrying.

The retry mechanism ensures that only the individual elements that fail are retried. That avoids double execution (it doesn't know if the requests have side effects) but also avoids wasting CPU repeating requests that already passed.

**Tests**

Updates tests to cover both IO errors and individual request failures.

Will also fix the scheduled cpp-verify job which was failing because it was hitting rate limits and not retrying.


op-service/client/retry.go
```diff
@@ -53,12 +54,68 @@ func (b *retryingClient) CallContext(ctx context.Context, result any, method str
...
+					// Replace with errors.Join when we move to go 1.20
+					combinedErr = errors.New(combinedErr.Error() + "\n" + elem.Error.Error())
+				}
+			}
+		}
+		if len(failed) > 0 {
+			pending = failed
+			return combinedErr
+		}
+		return nil
```
refcell, line 118: I think we can use [multierror](https://pkg.go.dev/github.com/hashicorp/go-multierror#section-readme) here to append errors nicely. I believe it's a fairly standard golang package also used in geth (if my memory's not failing me).
```suggestion
		var combinedErr error
		for i, elem := range batch {
			req := pending[i]
			idx := req.idx // Index into input of the original BatchElem

			// Set the result on the original batch to pass back to the caller in case we stop retrying
			input[idx].Error = elem.Error
			input[idx].Result = elem.Result

			// If the individual request failed, add it to the list to retry
			if elem.Error != nil {
				// Need to retry this request
				failed = append(failed, req)
				combinedErr = multierror.Append(combinedErr, elem.Error)
			}
		}
		if len(failed) > 0 {
			pending = failed
			return combinedErr
		}
		return nil
```

refcell (changes requested):
Small suggestion to use multierror.

op-service/client/retry.go
trianglesphere: IDK if geth uses multierror, but we used hashicorp's version before & there is now a version in the standard library.

ajsutton: Standard library has `errors.Join` but only in go 1.20 which we don't use yet.
More recent go also lets you use two `%w` in `fmt.Errorf` but the version of go we use doesn't. (I had some arguments with CI about these things....)

We do use hashicorp multierror already so have switched to that which is a nice cleanup.

refcell (approved):

Inphi (approved):
LGTM!

op-service/client/retry.go
trianglesphere: maybe time to get onto go 1.20. 1.19 will EOL this fall

PR #5954, opened 2023-06-09
feat(op-service): Mock TxManager

refcell:
**Description**

Adds a new `MockTxManager` to the `op-service` `txmgr` package for use in services when in dry-run mode.

PR #5955, opened 2023-06-09
feat(op-batcher): Dry Run

refcell:
**Description**

Adds a `--dry-run` flag for the op-batcher using a mock tx manager backend.

**Metadata**

Fixes CLI-3243

### DIFF
--- a/op-batcher/batcher/config.go
+++ b/op-batcher/batcher/config.go
@@ -86,6 +86,7 @@ type CLIConfig struct {
 	MaxL1TxSize uint64
 
 	Stopped bool
+	DryRun  bool
 
 	TxMgrConfig      txmgr.CLIConfig
 	RPCConfig        rpc.CLIConfig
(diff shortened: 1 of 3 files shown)


refcell:
* develop
  * **PR #5954** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5954" target="_blank">Graphite</a> 
    * **PR #5955** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5955" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5955?utm_source=stack-comment).

protolambda:
Does dry-run mode work if the batcher is started with a different 0-balance key?

refcell:
> Does dry-run mode work if the batcher is started with a different 0-balance key?

It should because the key is passed through to the tx manager, and since the mock tx manager in #5954 only logs, the balance is never checked.

trianglesphere (commented):
Two things that'd be helpful with the dry-run mode: 1) being able to replay this historically (fake clock which matches the L1 conditions at the time + the L2 blocks being loaded in) and 2) good logging / metrics around the cost of submitting transactions.

refcell:
> Two things that'd be helpful with the dry-run mode: 1) being able to replay this historically (fake clock which matches the L1 conditions at the time + the L2 blocks being loaded in) and 2) good logging / metrics around the cost of submitting transactions.

I'm not entirely sure what the implementation would look like to meet these two points.
- Is is possible to split into another pr so we don't bloat this one?
- Can we create linear tickets for these with a bit more concrete steps?
- Are these blockers to having this stack merged?

trianglesphere:
> Is is possible to split into another pr so we don't bloat this one?

Yep, we can do that.

> Can we create linear tickets for these with a bit more concrete steps?

Also yes. We also need to build a shared understanding of what we want here. We've had various ideas floating around, but nothing super concrete.

> Are these blockers to having this stack merged?

No, but I personally wouldn't call the dry run feature done at this point.

PR #5964, opened 2023-06-09
cleanup: delete message-relayer

tynes:
**Description**

The `message-relayer` was a legacy server that
was responsible for relaying withdrawals. It has
not been maintained and is no longer used as part
of the devnet system. Keeping it around adds
additional maintainance burden for the team, especially as things like 2 step withdrawals require it to be refactored.

Replaces https://github.com/ethereum-optimism/optimism/pull/4387


tynes:
The `geth tests` are just a flake

refcell (approved):

PR #5965, opened 2023-06-09
cleanup: delete migration-data

tynes:
**Description**

Deletes the `migration-data` package from the `packages` directory. This package is no longer useful and will exist in a previous commit/fork of the repo. It will be very painful to maintain the ability to run the migration indefinitely into the future, it should be safe to assume that users that want to perform the migration can check out the legacy commit and run it from that point.


refcell (approved):

refcell:
Let the record show the diff
```diff
+0
-398,352
```

PR #5967, opened 2023-06-09
feat(cmon): fix time string formatting bug

maurelian:
**Description**

An error was made which resulted in the logger printing a boolean rather than a formatted date string.


packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -191,16 +191,14 @@ export class WithdrawalMonitor extends BaseServiceV2<Options, Metrics, State> {
         await this.state.messenger.contracts.l2.BedrockMessagePasser.sentMessages(
           proofEvent.args.withdrawalHash
         )
-      const provenAt = `${
-        (dateformat(
-          new Date(
-            (await this.options.l1RpcProvider.getBlock(proofEvent.blockHash))
-              .timestamp * 1000
-          )
+      const provenAt = `${dateformat(
```
tynes, line 194: This is why nested expressions is an antipattern

maurelian: as opposed to something like this?

```
const date = new Date(
  (await this.options.l1RpcProvider.getBlock(proofEvent.blockHash))
    .timestamp * 1000
)
const dateString = dateformat(
  d,
  'mmmm dS, yyyy, h:MM:ss TT',
  true
)
const provenAt = dateString + "UTC"
```
      

My impression was that they are preferred based on the frequency of their occurrence.

tynes: I recommend a refactor with no nesting of expressions, it makes it much easier to read. Only solidity gets a pass on this because solc sucks and nesting expressions often is an optimization

tynes: Kelvin dies on a hill re: nested expressions and he is no longer on the eng team day to day

maurelian: lol

tynes (approved):

tynes:
semgrep scan looks like its a non issue

tynes:
You should add a changeset then lgtm for merge

packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -187,20 +187,19 @@ export class WithdrawalMonitor extends BaseServiceV2<Options, Metrics, State> {
     }
 
     for (const proofEvent of proofEvents) {
+      console.log('proofEvent:', proofEvent)
```
smartcontracts, line 190: Old log?

maurelian: oops, removed.

smartcontracts (approved):

PR #5968, opened 2023-06-09
feat(indexer): index all L1 & L2 contract events

hamdiallam:
When starting the processor, the optimism contracts should be supplied. We watch these contracts and index as batches are processed.

## Aside
- Skip L1 blocks that do not have optimism activity.

## Followup
- Contract configuration. Right now the L1/L2 contracts are the hardcoded local devnet addresses

### DIFF
--- a/indexer/database/bridge.go
+++ b/indexer/database/bridge.go
@@ -3,9 +3,12 @@ package database
 import (
 	"errors"
 
+	"gorm.io/gorm"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
-	"gorm.io/gorm"
+
+	"github.com/google/uuid"
 )
 
 /**
@@ -26,7 +29,7 @@ type TokenPair struct {
 }
 
 type Deposit struct {
-	GUID                 string `gorm:"primaryKey"`
+	GUID                 uuid.UUID `gorm:"primaryKey"`
 	InitiatedL1EventGUID string
 
 	Tx        Transaction `gorm:"embedded"`
@@ -39,7 +42,7 @@ type DepositWithTransactionHash struct {
 }
 
 type Withdrawal struct {
-	GUID                 string `gorm:"primaryKey"`
+	GUID                 uuid.UUID `gorm:"primaryKey"`
 	InitiatedL2EventGUID string
 
 	WithdrawalHash       common.Hash `gorm:"serializer:json"`
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ethereum-optimism/optimism/indexer/node"
 	"github.com/ethereum-optimism/optimism/indexer/processor"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 
 	"github.com/urfave/cli"
@@ -72,22 +73,30 @@ func NewIndexer(ctx *cli.Context) (*Indexer, error) {
 		return nil, err
 	}
 
-	// L1 Processor
+	// L1 Processor (hardhat devnet contracts). Make this configurable
+	l1Contracts := processor.L1Contracts{
+		OptimismPortal:         common.HexToAddress("0x6900000000000000000000000000000000000000"),
+		L2OutputOracle:         common.HexToAddress("0x6900000000000000000000000000000000000001"),
+		L1CrossDomainMessenger: common.HexToAddress("0x6900000000000000000000000000000000000002"),
+		L1StandardBridge:       common.HexToAddress("0x6900000000000000000000000000000000000003"),
+		L1ERC721Bridge:         common.HexToAddress("0x6900000000000000000000000000000000000004"),
+	}
 	l1EthClient, err := node.NewEthClient(ctx.GlobalString(flags.L1EthRPCFlag.Name))
 	if err != nil {
 		return nil, err
 	}
-	l1Processor, err := processor.NewL1Processor(l1EthClient, db)
+	l1Processor, err := processor.NewL1Processor(l1EthClient, db, l1Contracts)
 	if err != nil {
 		return nil, err
 	}
 
 	// L2Processor
+	l2Contracts := processor.L2ContractPredeploys() // Make this configurable
 	l2EthClient, err := node.NewEthClient(ctx.GlobalString(flags.L2EthRPCFlag.Name))
 	if err != nil {
 		return nil, err
 	}
-	l2Processor, err := processor.NewL2Processor(l2EthClient, db)
+	l2Processor, err := processor.NewL2Processor(l2EthClient, db, l2Contracts)
 	if err != nil {
 		return nil, err
 	}
--- a/indexer/processor/l1_processor.go
+++ b/indexer/processor/l1_processor.go
@@ -94,23 +94,12 @@ func l1ProcessFn(processLog log.Logger, ethClient node.EthClient, l1Contracts L1
 
 		/** Index Blocks **/
 
-		l1Headers := make([]*database.L1BlockHeader, numHeaders)
 		l1HeaderMap := make(map[common.Hash]*types.Header)
-		for i, header := range headers {
-			blockHash := header.Hash()
-			l1Headers[i] = &database.L1BlockHeader{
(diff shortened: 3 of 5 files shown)


hamdiallam:
DX-81

indexer/processor/l1_processor.go
```diff
@@ -1,18 +1,52 @@
...
+	L1ERC721Bridge         common.Address
+
+	// Some more contracts -- ProxyAdmin, SystemConfig, etcc
+	// Ignore the auxiliary contracts?
+
+	// Legacy contracts? We'll add this in to index the legacy chain.
+	// Remove afterwards?
+}
+
+func (c L1Contracts) toArray() []common.Address {
```
hamdiallam, line 33: Dont like how this has to be redefined for both `L1Contracts` and `L2Contracts`. Would like to eventually find a way to DRY this code

indexer/processor/l1_processor.go
```diff
@@ -1,18 +1,52 @@
...
+
+	return contracts
+}
+
 type L1Processor struct {
 	processor
 }
 
-func NewL1Processor(ethClient node.EthClient, db *database.DB) (*L1Processor, error) {
+func NewL1Processor(ethClient node.EthClient, db *database.DB, l1Contracts L1Contracts) (*L1Processor, error) {
```
hamdiallam, line 49: will add some safegaurds to make sure the supplied addresses in `L1Contracts` are deployed on the chain pointed to in `ethClient` through `eth_getCode` calls

indexer/processor/l2_processor.go
```diff
@@ -1,18 +1,60 @@
...
+func L2ContractPredeploys() L2Contracts {
+	return L2Contracts{
+		L2CrossDomainMessenger: common.HexToAddress("0x4200000000000000000000000000000000000007"),
+		L2StandardBridge:       common.HexToAddress("0x4200000000000000000000000000000000000010"),
+		L2ERC721Bridge:         common.HexToAddress("0x4200000000000000000000000000000000000014"),
+		L2ToL1MessagePasser:    common.HexToAddress("0x4200000000000000000000000000000000000016"),
+	}
+}
+
+func (c L2Contracts) toArray() []common.Address {
```
tynes, line 41: nit: this is technically a slice, i usually dont see array language usage in go

hamdiallam: ah yep good point

indexer/database/bridge.go
```diff
@@ -39,7 +42,7 @@ type DepositWithTransactionHash struct {
 }
 
 type Withdrawal struct {
-	GUID                 string `gorm:"primaryKey"`
+	GUID                 uuid.UUID `gorm:"primaryKey"`
```
roninjin10, line 45: What do these label things actually do?  I used them myself in the cli part without even really fully understanding this language feature

indexer/indexer.go
```diff
@@ -72,22 +73,30 @@ func NewIndexer(ctx *cli.Context) (*Indexer, error) {
...
 
-	// L1 Processor
+	// L1 Processor (hardhat devnet contracts). Make this configurable
+	l1Contracts := processor.L1Contracts{
+		OptimismPortal:         common.HexToAddress("0x6900000000000000000000000000000000000000"),
+		L2OutputOracle:         common.HexToAddress("0x6900000000000000000000000000000000000001"),
+		L1CrossDomainMessenger: common.HexToAddress("0x6900000000000000000000000000000000000002"),
+		L1StandardBridge:       common.HexToAddress("0x6900000000000000000000000000000000000003"),
+		L1ERC721Bridge:         common.HexToAddress("0x6900000000000000000000000000000000000004"),
+	}
```
roninjin10, line 83: Specifying contract addresses like this is a mistake Optimism consistently makes imo.   doesn't have to be this pr but we really need to make it a point to try to use deployment artifacts as a single source of truth

roninjin10: might go to client office hours to ask about this.  Easiest solution imo is to have a go package which is generated from the artifacts

roninjin10: There might actually already exist a solution like getContractArftifacts in js that I'm just not aware of as well

roninjin10: also one more comment about this.   These are actually coming from the config.toml file.   It has a section in it called [chain] which would be where these are configured.    I'm including a property though called `preset` that allows you to just use known chain presets for known chains like OP and base without having to specify everything though

indexer/processor/l1_processor.go
```diff
@@ -1,18 +1,52 @@
...
+	fields := reflect.VisibleFields(reflect.TypeOf(c))
+	v := reflect.ValueOf(c)
+
+	contracts := make([]common.Address, len(fields))
+	for i, field := range fields {
+		contracts[i] = (v.FieldByName(field.Name).Interface()).(common.Address)
+	}
+
+	return contracts
+}
```
roninjin10, line 43: Aren't all the contract objects ect. we need generated similar to typechain in op-bindings @tynes ?   I think we can just use that.

roninjin10: If we do need this in the package would prefer this to be in it's own folder called contracts rather than here

roninjin10: oh nvm my first comment I misread what this is doing

roninjin10 (approved):
lots of comments but since this is a WIP happy either addressing htem here or in future pr

indexer/database/bridge.go
hamdiallam: gorm will reflect on these tags and uses them to build the appropriate query string.

For example without the primaryKey tag, it wouldnt know how to properly insert or fetch rows and would tread the intended id as a regular column

indexer/indexer.go
hamdiallam: I'm all in for having presets that can be selected. Kinda left that open ended on the configuration side of things

The reason its important for the processor to not bias towards preset it so that it can work with any OP chain

roninjin10: tbh that part is intuitive it's more like "wtf is this language feature" that's confusing me.  Seeing here it's kinda an extra property Tag ont he thing though which makes it look less wierd/magical https://medium.com/golangspec/tags-in-golang-3e5db0b8ef3e

indexer/database/bridge.go
roninjin10: tbh that part is intuitive it's more like "wtf is this language feature" that's confusing me. Seeing here it's kinda an extra property Tag ont he thing though which makes it look less wierd/magical https://medium.com/golangspec/tags-in-golang-3e5db0b8ef3e

indexer/indexer.go
roninjin10: presets aren't doing anything magical.   A preset is just something that specifies all the contract addresses for you without having to be verbose about it.   None of the application will have any idea that presets exist

roninjin10: What I was getting at when mentioning the config and presets is that this code you wrote here should not be here.  It has tob e passed in from teh cli app as a param

roninjin10: essentially nothing about the app other than the config should be aware of contract addresses and other configurable things

hamdiallam: gotcha i see what you mean. On the same page that the config the sole entry point for how these addresses are generated.

We definitely don't have a good approach to this now. I can also smell future complexity once we go through additional contract upgrades where the ABI has to be versioned

roninjin10: Yup that's one reason I really want the build pipeline between contract artifacts and services like this to be as tight as possible.   If it's tight, we avoid versioning issues because the contracts upgrade and the indexer upgrade are forced to stay in sync

hamdiallam: >  If it's tight, we avoid versioning issues because the contracts upgrade and the indexer upgrade are forced to stay in sync

I don't think this enforcement would be desired for the indexer. For example, if someone is resyncing the indexer from genesis, using the latest version isn't actually what you want. The lifecycle of an op-chain will consist of many versions/upgrades

hamdiallam:
> lots of comments but since this is a WIP happy either addressing htem here or in future pr

Sounds, lets merge this and continue iterating

indexer/indexer.go
hamdiallam: resolving this to merge PR but good conversation to follow up on

PR #5974, opened 2023-06-09
make extradata configurable for the genesis block generator

roberto-bayardo:
**Description**

Make the ExtraData field of the genesis block configurable in the genesis generator by adding an optional L2GenesisBlockExtraData config option. Previous behavior (which is to use []byte(BEDROCK) for genesis ExtraData) is preserved when the option is unspecified.

**Tests**

Added a unit test to confirm that marshal/unmarshal works when the new option is present.

**Additional context**

We imagine non-Optimism superchains will want to customize ExtraData in their genesis blocks.


op-chain-ops/genesis/config.go
```diff
@@ -72,6 +72,9 @@ type DeployConfig struct {
 	// Seconds after genesis block that Regolith hard fork activates. 0 to activate at genesis. Nil to disable regolith
 	L2GenesisRegolithTimeOffset *hexutil.Uint64 `json:"l2GenesisRegolithTimeOffset,omitempty"`
 
+	// Configurable extradata. Default to []byte("BEDROCK") if empty.
+	L2GenesisBlockExtraData []byte `json:"l2GenesisBlockExtraData,omitempty"`
```
tynes, line 76: fwiw, none of the others use `omitempty` so i think its fine to leave that out

op-chain-ops/genesis/config_test.go
```diff
@@ -26,6 +26,20 @@ func TestConfigMarshalUnmarshal(t *testing.T) {
 	require.JSONEq(t, string(b), string(encoded))
 }
 
+func TestConfigWithExtraDataMarshalUnmarshal(t *testing.T) {
```
tynes, line 29: What do you think about not adding an additional json file for this and instead updating an existing json file?

op-chain-ops/genesis/genesis.go
```diff
@@ -74,16 +74,21 @@ func NewL2Genesis(config *DeployConfig, block *types.Block) (*core.Genesis, erro
 		difficulty = newHexBig(0)
 	}
 
+	extraData := config.L2GenesisBlockExtraData
+	if extraData == nil {
```
tynes, line 78: Should this check for `len(extraData) == 0`?

tynes:
Not sure why the `op-node-lint` is failing

op-chain-ops/genesis/config.go
roberto-bayardo: All except L2GenesisRegolithTimeOffset which I guess I was copying.  Let me know which you prefer? if I get rid of it here I can also get rid of it for L2GenesisRegolithTimeOffset for consistency I guess. Unless there's some good reason it should be different.

tynes: Ah, I feel like it was added there because its a pointer and `nil` can mean something different than `0`. I believe `nil` means never and `0` means now

tynes: I prefer leaving out `omitempty` in the case of `L2GenesisBlockExtraData`, curious what you think

op-chain-ops/genesis/genesis.go
roberto-bayardo: Chose this way because I wondered if there might be a use case where someone would want to genuinely set extradata to 0 bytes.  I think if someone explicitly has empty bytes for extradata in their json file they are probably not wanting the default?

op-chain-ops/genesis/config_test.go
roberto-bayardo: Yeah this could be overkill.  I created a separate one so I could have both default and non-default code paths getting exercised but the code is pretty simple so probably not so important.

roberto-bayardo: Have consolidated it to a single test.

op-chain-ops/genesis/config.go
roberto-bayardo: removed omitempty here, but left it with l2genesisregolithtimeoffset, since that one seems like it will in general be unspecified.

op-chain-ops/genesis/config_test.go
tynes: Thank you!

op-chain-ops/genesis/genesis.go
tynes: Ah ok this totally makes sense

tynes:
Looks good to me pending the tests passing

tynes:
Some sort of race condition in the `op-node` tests

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12f4da8]

goroutine 8926 [running]:
testing.tRunner.func1.2({0x15796c0, 0x275d240})
	/usr/local/go/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1399 +0x39f
panic({0x15796c0, 0x275d240})
	/usr/local/go/src/runtime/panic.go:884 +0x212
sync/atomic.(*Bool).Load(...)
	/usr/local/go/src/sync/atomic/type.go:17
github.com/ethereum-optimism/optimism/op-node/p2p.TestMultiPeerSync(0xc003c06d00?)
	/root/project/op-node/p2p/sync_test.go:254 +0xb68
testing.tRunner(0xc003c07040, 0x1b06c88)
	/usr/local/go/src/testing/testing.go:1446 +0x10b
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1493 +0x35f
```

cc @protolambda

op-chain-ops/genesis/config.go
ajsutton: Confirming your understanding of `L2GenesisRegolithTimeOffset` is correct - not setting it is a valid option (never) and different to 0 (starting from bedrock block).

roberto-bayardo:
> Looks good to me pending the tests passing

PR to fix the broken test: https://github.com/ethereum-optimism/optimism/pull/5995

PR #5986, opened 2023-06-11
Support non-legacy L2 deployments with optional governance

roberto-bayardo:
Extend the genesis utilities to support fresh (no migration requirement) L2 deployments, with or without governance token.


op-bindings/predeploys/addresses.go
```diff
@@ -3,66 +3,72 @@ package predeploys
...
+	GovernanceTokenAddr               = "0x4200000000000000000000000000000000000042"
+	LegacyMessagePasserAddr           = "0x4200000000000000000000000000000000000000"
+	L2ERC721BridgeAddr                = "0x4200000000000000000000000000000000000014"
+	OptimismMintableERC721FactoryAddr = "0x4200000000000000000000000000000000000017"
+	ProxyAdminAddr                    = "0x4200000000000000000000000000000000000018"
+	BaseFeeVaultAddr                  = "0x4200000000000000000000000000000000000019"
+	L1FeeVaultAddr                    = "0x420000000000000000000000000000000000001a"
 )
 
+type Predeploy struct {
```
tynes, line 26: What is the purpose of this struct?

op-chain-ops/genesis/layer_two.go
```diff
@@ -46,3 +54,74 @@ func BuildL2DeveloperGenesis(config *DeployConfig, l1StartBlock *types.Block) (*
...
+
+	// Set up the proxies
+	depBytecode, err := bindings.GetDeployedBytecode("Proxy")
+	if err != nil {
+		return nil, err
+	}
+	if len(depBytecode) == 0 {
+		return nil, errors.New("Proxy has empty bytecode")
+	}
+	for i := uint64(0); i <= 2048; i++ {
```
tynes, line 86: This code is duplicated in a few places, perhaps we can reuse this code?

op-bindings/predeploys/addresses.go
tynes: Ah ok I see its used in building the genesis. I'd like to delete the migration code from the codebase to prevent such large diffs

roberto-bayardo: Right, this PR has two commits, the first is the extension of predeploy type with these extra metadata fields, and the second is the crux of the change that implements the new non-migration L2 genesis w/ optional governance (which has a much smaller diff).  Consider the first commit tentative -- if predeploy metadata isn't likely to be useful beyond genesis we can revert & keep that metadata local to package genesis.

op-chain-ops/genesis/layer_two.go
roberto-bayardo: The other proxy deployment loops are slightly different since they also embed the handling of migration cases. I'll take another look and see if that can be cleanly separated out, but I figured the migration code might be short lived anyway.

op-bindings/predeploys/addresses.go
roberto-bayardo: Should I go ahead and delete the migration code? That would indeed simplify a lot of things....

op-chain-ops/genesis/config.go
```diff
@@ -216,14 +216,15 @@ func (d *DeployConfig) Check() error {
 	if d.L2GenesisBlockBaseFeePerGas == nil {
 		return fmt.Errorf("%w: L2 genesis block base fee per gas cannot be nil", ErrInvalidDeployConfig)
 	}
```
tynes, line 218: My preference would be to add a bool that is enable/disable gov token instead of making the config implicit on each other, the other config values only apply if the bool value is set

op-chain-ops/genesis/layer_two.go
tynes: Ah got it, I'd like to remove the migration code to reduce diff in these sorts of PRs since it can live in our legacy repo, then hopefully we can settle on 1 implementation of creating the L2 genesis. I believe there is an `op-node` genesis command that will generate the `genesis.json`, that was the UX we came up with for creating a genesis file

roberto-bayardo: op-node's genesis command calls the function above this one (BuildL2DeveloperGenesis), which we should probably replace with one that doesn't include the deprecated stuff.

tynes:
Are you ok with us merging https://github.com/ethereum-optimism/optimism/pull/5996 to help reduce the diff of this PR? Will make it easier to review for correctness

roberto-bayardo:
> Are you ok with us merging #5996 to help reduce the diff of this PR? Will make it easier to review for correctness

yes, let me do that and make another pass over this to see what I can clean up / simplfiy.

roberto-bayardo:
Updated this PR to make it quite a bit more surgical, e.g. it now only excludes a single "legacy" contract (LegacyWETH).

Still TODO: change the config to a boolean to enable/disable governance.

tynes:
Amazing! This diff is so much more manageable

op-chain-ops/genesis/config.go
roberto-bayardo: done

op-chain-ops/genesis/layer_two.go
```diff
@@ -46,3 +54,74 @@ func BuildL2DeveloperGenesis(config *DeployConfig, l1StartBlock *types.Block) (*
 
 	return db.Genesis(), nil
 }
+
+// BuildL2MainnetGenesis will build an L2 Genesis suitable for a Superchain mainnet that does not
+// require a pre-bedrock migration & supports optional governance token predeploy.
```
tynes, line 59: It would be great to have a spec of sorts in this godoc so that we can be sure that the impl is correct.
- Proxies in the address space `[0x4200000000000000000000000000000000000000, 0x4200000000000000000000000000000000000800)` for predeploys
- All predeploy proxies owned by the `ProxyAdmin`
- Predeploys as per the [spec](https://github.com/ethereum-optimism/optimism/blob/develop/specs/predeploys.md)
  - except for no  `LegacyERC20ETH` at `0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000`
  - optional governance token at `0x4200000000000000000000000000000000000042`

op-chain-ops/genesis/layer_two.go
```diff
@@ -46,3 +54,74 @@ func BuildL2DeveloperGenesis(config *DeployConfig, l1StartBlock *types.Block) (*
...
+	deployResults, err := immutables.BuildOptimism(immutable)
+	if err != nil {
+		return nil, err
+	}
+	for name, predeploy := range predeploys.Predeploys {
+		addr := *predeploy
+		if predeploys.IsDeprecated(addr) {
+			continue
+		}
+		if addr == predeploys.GovernanceTokenAddr && storage["GovernanceToken"] == nil {
```
tynes, line 104: I don't love how this is implicit on an upstream function mutating state, perhaps we can use the bool on the deploy config itself instead here?

tynes: I think it would be good to have a logline saying governance disabled or something

op-bindings/predeploys/addresses.go
```diff
@@ -46,6 +46,29 @@ var (
 	Predeploys = make(map[string]*common.Address)
 )
 
+// IsProxied returns true for predeploys that will sit behind a proxy contract
+func IsProxied(predeployAddr common.Address) bool {
+	switch predeployAddr {
+	case LegacyERC20ETHAddr:
```
tynes, line 52: The false cases can be collapsed into a single branch

op-chain-ops/genesis/layer_two.go
```diff
@@ -46,3 +54,74 @@ func BuildL2DeveloperGenesis(config *DeployConfig, l1StartBlock *types.Block) (*
 
 	return db.Genesis(), nil
 }
+
+// BuildL2MainnetGenesis will build an L2 Genesis suitable for a Superchain mainnet that does not
+// require a pre-bedrock migration & supports optional governance token predeploy.
+func BuildL2MainnetGenesis(config *DeployConfig, l1StartBlock *types.Block) (*core.Genesis, error) {
```
tynes, line 60: I don't love the name of this function, ideally this is what the `op-node` uses for its genesis command for devnets long term, i think its ok for now

tynes (commented):
Do you plan on generating the file using the test? Or the `op-node` command? I'm fine with either, the long term ideal is the `op-node` command so that we can have a single way of doing it

op-chain-ops/genesis/layer_two_test.go
```diff
@@ -94,3 +93,110 @@ func TestBuildL2DeveloperGenesisDevAccountsFunding(t *testing.T) {
...
+		if !predeploys.IsProxied(addr) {
+			continue
+		}
+
+		adminSlot, ok := account.Storage[genesis.AdminSlot]
+		require.Equal(t, true, ok)
+		require.Equal(t, predeploys.ProxyAdminAddr.Hash(), adminSlot)
+		require.Equal(t, depB, account.Code)
+	}
+	require.Equal(t, 2063, len(gen.Alloc)) // TODO: confirm 2063 is correct!
```
tynes, line 141: Do you think a manual inspection of the output would be appropriate to accomplish this todo?

op-chain-ops/genesis/layer_two.go
```diff
@@ -46,3 +54,74 @@ func BuildL2DeveloperGenesis(config *DeployConfig, l1StartBlock *types.Block) (*
...
+	genspec, err := NewL2Genesis(config, l1StartBlock)
+	if err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
roberto-bayardo: Right, I intended this to be appropriate for a mainnet so I left that out.  It's a tiny amount though so maybe it's OK?  Did you leave it in your bedrock mainnet?  I was also wondering why it is that finding it like that makes calls cheaper?

op-bindings/predeploys/addresses.go
roberto-bayardo: You mean via fallthrough?

op-chain-ops/genesis/layer_two.go
roberto-bayardo: I called it a "mainnet" config since I don't prefund any accounts like the dev-genesis function was doing.  We could add a bool for prefunding dev accounts here, or make it a config option?

op-chain-ops/genesis/layer_two.go
roberto-bayardo: yes should have used the config in the first place: fixed.  Also added a log line.

op-bindings/predeploys/addresses.go
tynes: Like this example:

```golang
func WhiteSpace(c rune) bool {
    switch c {
    case ' ', '\t', '\n', '\f', '\r':
        return true
    }
    return false
}
```

I believe you can also have something like:
```
case foo:
case bar:
    return true
```

op-chain-ops/genesis/layer_two.go
tynes: I think its fine as it is, we don't need to worry about it too much right now

op-bindings/predeploys/addresses.go
roberto-bayardo: ah got it, don't need a fallthrough with a return statement outside the case. fixed.

op-chain-ops/genesis/layer_two_test.go
```diff
@@ -94,3 +93,107 @@ func TestBuildL2DeveloperGenesisDevAccountsFunding(t *testing.T) {
...
+		if !predeploys.IsProxied(addr) {
+			require.NotEqual(t, proxyBytecode, account.Code)
+			continue
+		}
+
+		adminSlot, ok := account.Storage[genesis.AdminSlot]
+		require.Equal(t, true, ok)
+		require.Equal(t, predeploys.ProxyAdminAddr.Hash(), adminSlot)
+		require.Equal(t, proxyBytecode, account.Code)
+	}
```
tynes, line 138: This could also get the impl and then get the code of the impl to ensure that code is set at the impl

op-chain-ops/genesis/layer_two_test.go
```diff
@@ -94,3 +93,107 @@ func TestBuildL2DeveloperGenesisDevAccountsFunding(t *testing.T) {
...
+	if writeFile {
+		file, _ := json.MarshalIndent(gen, "", " ")
+		_ = os.WriteFile("genesis.json", file, 0644)
+	}
+}
+
+// Same test as TestBuildL2MainnetGenesis, only we blow away the governance token config and
+// confirm the governance predeploy doesn't exist (or more precisely, there's an unused proxy
+// contract at its address instead).
+func TestBuildL2MainnetNoGovernanceGenesis(t *testing.T) {
```
tynes, line 150: Perhaps could combine some of the business logic of the 2 tests added

op-chain-ops/genesis/layer_two_test.go
roberto-bayardo: hmm, not seeing a straightforward way to do that?

op-chain-ops/genesis/layer_two_test.go
roberto-bayardo: done.

PR #5987, opened 2023-06-11
feat(contracts-bedrock): Validate forge version before building

roninjin10:
- Validate forge version before building
- Opened issue with foundry [here](https://github.com/foundry-rs/foundry/issues/5133)

### DIFF
--- a/packages/contracts-bedrock/scripts/verify-foundry-install.ts
+++ b/packages/contracts-bedrock/scripts/verify-foundry-install.ts
@@ -1,8 +1,24 @@
+import { execSync } from 'child_process'
+
 import { sync as commandExistsSync } from 'command-exists'
 
 if (!commandExistsSync('forge')) {
   console.error(
-    'Command failed. Is Foundry not installed? Consider installing via `curl -L https://foundry.paradigm.xyz | bash` and then running `foundryup` on a new terminal. For more context, check the installation instructions in the book: https://book.getfoundry.sh/getting-started/installation.html.'
+    `Command failed. Is Foundry not installed?
+Consider installing via \`curl -L https://foundry.paradigm.xyz | bash\` and then running \`foundryup -C da2392e58bb8a7fefeba46b40c4df1afad8ccd22\` on a new terminal.
+For more context, check the installation instructions in the book: https://book.getfoundry.sh/getting-started/installation.html.`
   )
   process.exit(1)
+}
+
+const version = execSync('forge --version').toString()
+
+
+if (!version.includes('da239')) {
+  console.warn(
+    `Detected forge version ${version}. This may work, but if you run into any issues, please upgrade to forge 0.2.0 or higher.
+Consider installing via \`curl -L https://foundry.paradigm.xyz | bash\` and then running \`foundryup -C da2392e58bb8a7fefeba46b40c4df1afad8ccd22\` on a new terminal.
+For more context, check the installation instructions in the book: https://book.getfoundry.sh/getting-started/installation.html.`
+  )
+  process.exit(0)
 }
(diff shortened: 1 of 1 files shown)


roninjin10:
* develop
  * **PR #5987** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5987" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/5987?utm_source=stack-comment).

clabby:
I like the idea here. What do you think about a more specific check given that we pin a specific version of foundry? i.e. right now we're on `da2392e`, and there are many nightly releases within `0.2.0`.

roninjin10:
@clabby done and I made it a warning instead of an error

packages/contracts-bedrock/scripts/verify-foundry-install.ts
```diff
@@ -1,8 +1,24 @@
...
+const version = execSync('forge --version').toString()
+
+
+if (!version.includes('da239')) {
+  console.warn(
+    `Detected forge version ${version}. This may work, but if you run into any issues, please upgrade to forge 0.2.0 or higher.
+Consider installing via \`curl -L https://foundry.paradigm.xyz | bash\` and then running \`foundryup -C da2392e58bb8a7fefeba46b40c4df1afad8ccd22\` on a new terminal.
+For more context, check the installation instructions in the book: https://book.getfoundry.sh/getting-started/installation.html.`
+  )
+  process.exit(0)
```
roninjin10, line 23:
```suggestion
```

roninjin10:
@clabby I opened an issue about this in foundry repo btw https://github.com/foundry-rs/foundry/issues/5133

packages/contracts-bedrock/scripts/verify-foundry-install.ts
```diff
@@ -1,8 +1,23 @@
...
   )
   process.exit(1)
 }
+
+const version = execSync('forge --version').toString()
+
+
+if (!version.includes('da239')) {
+  console.warn(
+    `Detected forge version ${version}. This may work, but if you run into any issues, please upgrade to forge 0.2.0 or higher.
```
roninjin10, line 19:
```suggestion
    `Detected forge version ${version}. This may work, but if you run into any issues, please upgrade to forge da2392e58bb8a7fefeba46b40c4df1afad8ccd22.
```

clabby: I wonder if we should put this inside of a configuration file somewhere. At the moment, the commit hash of the foundry version we pin is in `ci-builder`, the contracts package README, and now here as well. @maurelian - any issues with us putting it in one spot?

clabby: If the foundry guys decide to allow for pinning the version within the TOML, we can read it from there, but we may need a temp workaround.

roninjin10: The markdown file makes it tough to get it down to below 2 spots.  Best we could do for a markdown file is link out to another file since github doesn't support any fancy import directives.   The ci-builder and this could share though

tynes:
Looks like lint error

packages/contracts-bedrock/scripts/verify-foundry-install.ts
maurelian: How about: 
1. create a file `echo da2392e58bb8a7fefeba46b40c4df1afad8ccd22 > .forge-version`
2. I think that this script and the ci-builder could just read that file, and the markdown could like to it.

roninjin10:
Got 3 p0s atm will circle back to this likely thursday

packages/contracts-bedrock/scripts/verify-foundry-install.ts
roninjin10: I like this.   The alternative to linking to it would be to write a very simple ci check that checks that readme has the forge version in it

tynes:
Will reopen this PR after the recent changes to how foundry versions are managed in the repo

PR #5990, opened 2023-06-12
README: Remove broken link to infra/op-replica

ajsutton:
**Description**

Remove broken link to `infra/op-replica` in the README.  op-replica was removed in 23150b8b85c20a83782c3b5a73ced9bad527bc0d


tynes (approved):
Good catch

PR #5991, opened 2023-06-12
op-node: Require peer scoring params to be valid if set

ajsutton:
**Description**

If peer scoring params are not `nil`, activate peer scoring. Previously additional sanity checks were done to ignore params that weren't valid. `p2p.Config` now stores the params as pointers so they default to `nil`, avoiding the invalid default values that previously required these additional checks.

**Metadata**

- Trying to pave the way for https://linear.app/optimism/issue/CLI-3603/op-e2e-peer-score-testing


op-node/p2p/peer_scores.go
```diff
@@ -12,14 +12,13 @@ func ConfigurePeerScoring(gossipConf GossipSetupConfigurables, scorer Scorer, lo
 	peerScoreParams := gossipConf.PeerScoringParams()
 	peerScoreThresholds := NewPeerScoreThresholds()
 	opts := []pubsub.Option{}
-	// Check the app specific score since libp2p doesn't export it's [validate] function :/
-	if peerScoreParams != nil && peerScoreParams.AppSpecificScore != nil {
```
Inphi, line 16: does libp2p still complain if the AppSpecificScore is nil? If not, we should validate the peer score params settings ourselves

ajsutton: Yes it does.

protolambda (approved):

PR #5993, opened 2023-06-12
case check for 019 deploy-utils

chrisciszak:
**Description**

As in the title.


tynes:
This code is going to be refactored to remove the dictator as part of the deploy path, will make the deployment of the system cheaper

chrisciszak:
@tynes you want close this PR then or merge and then refactor. It's causing issues for people though.

tynes:
We have removed this code as part of the deployment process, please open an issue if you are still having problems

PR #5998, opened 2023-06-12
indexer: cleanup legacy code

tynes:
**Description**

Deletes some legacy code from the indexer. No need to keep around multiple copies of the legacy bindings, so migrate to the legacy bindings kept in `op-bindings` Had to update some deps to get this to build correctly. Replace all imports to the `op-bindings` scc bindings instead of the one internal to this package.


hamdiallam (approved):

tynes:
The following tests should have been skipped and are failing because their dependency (building the monorepo with yarn) never ran:

- ci/circleci: contracts-tests
- ci/circleci: fault-detector-tests
- ci/circleci: sdk-next-tests
- ci/circleci: sdk-tests

PR #6002, opened 2023-06-13
op-node: Fix TestMultiPeerSync flakiness

ajsutton:
**Description**

Waits for the block to be requested before checking the request has completed.

There are three potential ways the race condition in this test could resolve:

* Test checks before the request is made - should wait for inflight request to appear (currently panics)
* Test checks for the request while its in flight - should wait for it to complete (currently works)
* Test checks for the request after it has completed and been removed from inflight map (currently panics)

Sadly, for case 1 you need to wait if the request isn't present in the inflight map and for case 3 you need to not wait if it isn't present.  When I solved for case 1, sometimes the test would just hang forever (we should add a timeout on that wait as well).

So this introduces a separate channel that tracks which blocks have been requested (for convenience it's shared across all peers). That gets us past case 1, and we can handle 2&3 by waiting for the request to be marked as complete or removed from the inflight map.

Also added timeouts when waiting so the test can't hang forever.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4086/fix-testmultipeersync


op-node/p2p/sync_test.go
```diff
@@ -251,7 +269,15 @@ func TestMultiPeerSync(t *testing.T) {
...
-	for !clB.inFlight[25].Load() {
+	ctx, cancelFunc = context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancelFunc()
+	for clB.inFlight[25] != nil && !clB.inFlight[25].Load() {
+		select {
+		case <-ctx.Done():
+			t.Fatal("Did not complete request for block 25 in timely manner")
+		default:
+			// Carry on, timeout not yet reached
+		}
```
protolambda, line 280: Part of the problem is that the original fix of mine was wrong here, and we shouldn't read from `inFlight` map at all here, because it can be concurrently changed

protolambda (commented):
Think we need to change the sync client to safely expose the right information to do the test.

op-node/p2p/sync_test.go
ajsutton: hmm, good point - while our code can handle the race conditions involved with two threads modifying the map, I suspect go's map implementation may not like it.

I've rebased the original commit onto latest develop and pushed up an extra commit that changes it to use a new `isInFlight` method on `SyncClient`. It means adding new production code but I think it makes sense and its the least intrusive thing I could think of to be able to know that the response has actually been processed.

protolambda (approved):

PR #6007, opened 2023-06-13
improvement(fault-detector) quickly decrease block range size when polling

hamdiallam:
the fault detector will reduce the block range size when polling for events then quickly increase it when syncing events from genesis.

as a patch, lets reduce the block range faster than we increase it -- only increasing when the step size has been successful for at least 5 iterations.

** Longer term fix **
- Once support for `StateCommitmentChain` chain is removed, the fault detector should be smart enough to binary search the l1 block space for a starting L1 height since the `l2Outputs` stored in `L2OutputOracle` contains timestamps of the l1 block for a given output. We cant do this right now since `StateCommitmentChain` batches does have this

### DIFF
--- a/packages/fault-detector/src/helpers.ts
+++ b/packages/fault-detector/src/helpers.ts
@@ -67,16 +67,13 @@ export const updateOracleCache = async <TSubmissionEventArgs>(
   })
 
   let failures = 0
+  let success = 0
   let currentBlock = cache.highestBlock + 1
   let step = endBlock - currentBlock
   while (currentBlock < endBlock) {
+    const rangeInfo = { node: 'l1', startBlock: currentBlock, rangeSize: step }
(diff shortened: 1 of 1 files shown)


packages/fault-detector/src/helpers.ts
```diff
@@ -117,7 +121,12 @@ export const updateOracleCache = async <TSubmissionEventArgs>(
 
       // We've failed 3 times in a row, we're probably stuck.
       if (failures >= 3) {
```
roninjin10, line 123: Likely never becomes relavent but likely good to at least anchor at the alchemy limit
`
```suggestion
      if (failures >= 3 && step < 2000) {
```

roninjin10 (approved):
lgtm

PR #6010, opened 2023-06-13
devnet: unify tooling for how to bring up devnet

tynes:
**Description**

The bash script is removed in favor of unifying on the python code. In the future we should only have 1 way to bring up the devnet because it is difficult to maintain 2 different ways. It adds additional CI runs plus the need to add functionality to 2 different things when contracts are added to the system.


ajsutton (approved):
LGTM.  Was able to spin up the devnet locally with both pre-deploy and deploy modes fine. We do still have two quite divergent paths to spin up the devnet since nearly all the python code is split to different methods depending on the deployment type, but it absolutely makes sense to have them both in python at least.  Would have saved me a lot of confusion when I was first understanding how the devnet gets spun up.

tynes:
We should eliminate one of the paths and imo it should be the contracts in genesis path, maintaining the L1 portion is not ideal. The same code that deploys to networks should be used for integration testing

PR #6013, opened 2023-06-14
unify genesis factories

roberto-bayardo:
**Description**

Unify develop & mainnet genesis factories into a single factory & remove more migration related legacy code.

There are only 2 behavior changes between the previous BuildL2Dev factory and the factory method replacing it from this PR:

- Predeploy balances are now only funded when config.FundDevAccounts is true. Previously they were funded regardless.

- LegacyERC20Eth predeploy is removed entirely as it should have only been needed for migrating pre-bedrock Optimism.
 
**Tests**

Existing test cases were appropriately updated to cover the new behavior changes.

**Additional context**

Follow up to https://github.com/ethereum-optimism/optimism/pull/5986


op-chain-ops/genesis/layer_two.go
```diff
@@ -1,83 +1,29 @@
...
 	genspec, err := NewL2Genesis(config, l1StartBlock)
 	if err != nil {
 		return nil, err
 	}
 
 	db := state.NewMemoryStateDB(genspec)
-
 	if config.FundDevAccounts {
 		FundDevAccounts(db)
+		SetPrecompileBalances(db)
```
roberto-bayardo, line 25: Protected this call within the FundDevAccounts since the comment for it says "This should only be used for devnets."

roberto-bayardo:
Made ProxyAdmin a proxied contract and updated the documentation [here](https://github.com/ethereum-optimism/optimism/blob/develop/specs/predeploys.md) (which had it listed is NOT proxied in the matrix, but proxied in the description -- matrix seemed to be what was wrong).

tynes:
This PR is looking good so far!

roberto-bayardo:
> This PR is looking good so far!

thanks, I think it's good to go if only the CI would pass. I think it's just flakiness -- anything I can do to help it along?

tynes:
> > This PR is looking good so far!
> 
> thanks, I think it's good to go if only the CI would pass. I think it's just flakiness -- anything I can do to help it along?

The only thing in CI that is failing is due to the env vars not being populated in the CI run, I manually inspected the logs of the CI runs that are impacted by this change (both devnets) and they look good to me

roberto-bayardo:
> > > This PR is looking good so far!
> > 
> > 
> > thanks, I think it's good to go if only the CI would pass. I think it's just flakiness -- anything I can do to help it along?
> 
> The only thing in CI that is failing is due to the env vars not being populated in the CI run, I manually inspected the logs of the CI runs that are impacted by this change (both devnets) and they look good to me

ok, tagged as ready for review.

tynes (approved):

PR #6016, opened 2023-06-14
feat(ctb): `toNibbles` optimization.

clabby:
## Overview

Revisits #4340 after _6 months_ of the change being differential tested against the current implementation and an added memory safety cheatcode.

**Description**

Saves roughly ~15.5% of the cost of an average `proveWithdrawalTransaction` call to the `OptimismPortal`.

**Tests**

* Memory safety test for the inline-asm version of the `toNibbles` function, ported from [substratum](https://github.com/clabby/substratum)

**Additional context**
This change has never failed a differential test after months of inclusion in every single CI run. In addition, we now have memory safety assertions by forge.


op-bindings/bindings/optimismportal.go
maurelian, line 1: Portal is gonna need a semver bump

clabby: Good callout, bumped.

tynes:
> This change has never failed a differential test after months of inclusion in every single CI run

Looks good to me

tynes (approved):

maurelian:
> This change has never failed a differential test after months of inclusion in every single CI run

Where is the diff testing setup for this? Was there any difference between the first and the 1000th time that test ran?

PR #6019, opened 2023-06-14
op-node: handle -32600 receipts err code, and always fallback to err msg

protolambda:
**Description**

Alchemy returns `-32600` (generic) instead of `-32601` ("unknown method") for unsupported RPC methods. We should make the op-node fall-back to the wider-supported receipt fetching methods when it hits this error, to avoid sync stalling. It already will log warnings about it falling back, and tries the advanced receipt fetching methods later again.


protolambda:
update: lint issue, fixed now

trianglesphere (approved):

PR #6026, opened 2023-06-15
feat(ctb): DisputeGameFactory Deploy Scripts

refcell:
**Description**

Adds a deploy scripts for the `DisputeGameFactory` that are only run on the devnet (as matched by the deploy config l1 chain id).

**Metadata**

Fixes CLI-4093


packages/contracts-bedrock/deploy/020-DisputeGameFactoryImpl.ts
```diff
@@ -0,0 +1,32 @@
+import assert from 'assert'
+
+import { DeployFunction } from 'hardhat-deploy/dist/types'
+
+import { deploy } from '../src/deploy-utils'
+
+const deployFn: DeployFunction = async (hre) => {
+  // We only want to deploy the dgf on devnet for now
+  if (hre.deployConfig.l1ChainID === 900) {
+    await deploy({
```
tynes, line 10: One thing I realized is that if the contract deploys and then the tx fails inside of `postDeployAction`, the `postDeployAction` will not run again. We could assign the result of `deploy` to a variable and then send the txs that way. Also realizing that we should call `initialize` in the constructor so we don't need to call it afterwards

PR #6028, opened 2023-06-15
fix(ctb): Initialize DisputeGameFactory in Constructor

refcell:
**Description**

Fixes the `DisputeGameFactory` to be initialized in the constructor with `address(0)`.


packages/contracts-bedrock/contracts/dispute/DisputeGameFactory.sol
```diff
@@ -43,18 +43,18 @@ contract DisputeGameFactory is OwnableUpgradeable, IDisputeGameFactory, IVersion
...
     /**
-     * @notice Constructs a new DisputeGameFactory contract. Set the owner
-     *         to `address(0)` to prevent accidental usage of the implementation.
+     * @notice Constructs a new DisputeGameFactory contract.
+     * @param _owner The owner of the contract.
      */
-    constructor() OwnableUpgradeable() {
-        _transferOwnership(address(0));
+    constructor(address _owner) OwnableUpgradeable() {
+        initialize(_owner);
```
tynes, line 50: What if we call init with `address(0)` here to preserve the property as before?

tynes (approved):

tynes:
Once CI passes, we should merge this to unblock https://github.com/ethereum-optimism/optimism/pull/6029

tynes:
Ok waiting on CI to pass here

PR #6031, opened 2023-06-15
require fee increase & gas re-estimation for retried transactions

roberto-bayardo:
**Description**

Currently the txmgr retry logic will not bump a transaction's tip and max-fee in the case the latest suggested values are lower than the original values.  When a transaction is stuck in the mempool (e.g. due to too low fees or bad gaslimit estimate) this regularly results in retry failures with error message "resubmitted already known transaction" or "replacement transaction underpriced".

This PR imposes a mandatory fee increase with each retry (up to a point) to ensure a retried tx isn't immediately rejected in this manner.

We also see txs get stuck in the mempool because of extremely high gaslimit estimation (e.g. gaslimit nearly equals block gas limit).  In these cases, increasing fees alone does not solve the problem. To fix this scenario, this change makes the txmgr re-estimate gas with each retry.  We have also created a [geth PR](https://github.com/ethereum/go-ethereum/pull/27502) to fix what we suspect is causing the high gaslimit estimates.  (Fee re-estimation also seems to be helpful in detecting if resubmitting the tx would just fail anyway due to a state change.)

**Tests**

Updated unit tests to capture the new behavior.

**Additional context**

The issues addressed here have caused the Base goerli proposer to get stuck multiple times. Historically we have resorted to unjamming the proposer by manually submitting a transaction with a higher fee and lower gaslimit to replace the stuck transaction.

This issue does not happen with the batcher because batcher txs are simple sends, and we believe the underlying cause of the bad gas usage estimation only arises for reverting contract txs. Proposer txs revert quite frequently with the error message "block hash does not match the hash at the expected height".

### DIFF
--- a/op-service/txmgr/txmgr.go
+++ b/op-service/txmgr/txmgr.go
@@ -21,9 +21,8 @@ import (
 )
 
 const (
-	// Geth defaults the priceBump to 10
-	// Set it to 15% to be more aggressive about including transactions
-	priceBump int64 = 15
+	// Geth requires a minimum fee bump of 10% for tx resubmission
+	priceBump int64 = 10
 
 	// The multiplier applied to fee suggestions to put a hard limit on fee increases
 	feeLimitMultiplier = 5
@@ -285,7 +284,15 @@ func (m *SimpleTxManager) sendTx(ctx context.Context, tx *types.Transaction) (*t
 				return nil, errors.New("aborted transaction sending")
 			}
 			// Increase the gas price & submit the new transaction
-			tx = m.increaseGasPrice(ctx, tx)
+			newTx, err := m.increaseGasPrice(ctx, tx)
+			if err != nil || sendState.IsWaitingForConfirmation() {
+				// there is a chance the previous tx goes into "waiting for confirmation" state
+				// during the increaseGasPrice call. In some (but not all) cases increaseGasPrice
+				// will error out during gas estimation. In either case we should continue waiting
+				// rather than resubmit the tx.
+				continue
+			}
+			tx = newTx
 			wg.Add(1)
 			bumpCounter += 1
 			go sendTxAsync(tx)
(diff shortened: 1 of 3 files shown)


op-service/txmgr/txmgr.go
```diff
@@ -418,32 +418,42 @@ func (m *SimpleTxManager) queryReceipt(ctx context.Context, txHash common.Hash,
...
+		tip = new(big.Int).Set(tx.GasTipCap())
+		basefee = new(big.Int)
 	}
-	gasTipCap, gasFeeCap := updateFees(tx.GasTipCap(), tx.GasFeeCap(), tip, basefee, m.l)
+	bumpedTip, bumpedFee := updateFees(tx.GasTipCap(), tx.GasFeeCap(), tip, basefee, m.l)
 
-	if tx.GasTipCapIntCmp(gasTipCap) == 0 && tx.GasFeeCapIntCmp(gasFeeCap) == 0 {
-		return tx
+	// Make sure increase is at most 5x the suggested values
+	maxTip := new(big.Int).Mul(tip, big.NewInt(5))
```
tynes, line 442: The `big.NewInt(5)` could be a top level `var` with a godoc

tynes:
Generally looks good to me so far but tagging @trianglesphere who wrote this code

roberto-bayardo:
Testing update: this is running on our alpha goerli sequencer/proposer/batcher since since yesterday evening. Since then not a single "already submitted" error. Before we'd see ~5 of these errors per hour across the proposer and the batcher. This is normal operation though, not the error situation we are hoping to prevent. I still have some suspicion the failure case may be related to gaslimit, which is why I added the gaslimit logging each time it bumps fees.

roberto-bayardo:
Re: gaslimit, I see a lot of these errors in the logs that arise from the txmgr call to EstimateGas. Anything to be concerned about?

`failed to create the tx: failed to estimate gas: execution reverted: L2OutputOracle: block hash does not match the hash at the expected height
`

op-service/txmgr/txmgr.go
roberto-bayardo: done

op-service/txmgr/txmgr.go
```diff
@@ -20,9 +20,14 @@ import (
 	"github.com/ethereum-optimism/optimism/op-service/txmgr/metrics"
 )
 
-// Geth defaults the priceBump to 10
-// Set it to 15% to be more aggressive about including transactions
-const priceBump int64 = 15
+const (
+	// Geth defaults the priceBump to 10
+	// Set it to 15% to be more aggressive about including transactions
+	priceBump int64 = 15
```
roberto-bayardo, line 26: We should consider reducing this to 10 (the min required by geth) with this change. I don't see a good reason to go above the minimum since if it's still too low we'll end up retrying again anyway with another bump.

op-service/txmgr/txmgr.go
```diff
@@ -418,46 +427,67 @@ func (m *SimpleTxManager) queryReceipt(ctx context.Context, txHash common.Hash,
 	return nil
 }
 
-// increaseGasPrice takes the previous transaction & potentially clones then signs it with a higher tip.
-// If the tip + basefee suggested by the network are not greater than the previous values, the same transaction
-// will be returned. If they are greater, this function will ensure that they are at least greater by 15% than
-// the previous transaction's value to ensure that the price bump is large enough.
-//
-// We do not re-estimate the amount of gas used because for some stateful transactions (like output proposals) the
-// act of including the transaction renders the repeat of the transaction invalid.
```
roberto-bayardo, line 427: I was confused by the above comment: the fact that a tx can become invalid seems a reason why you *should* re-estimate gas here, as you detect the tx is invalid earlier before even trying to execute it.

op-service/txmgr/txmgr.go
```diff
@@ -418,46 +427,67 @@ func (m *SimpleTxManager) queryReceipt(ctx context.Context, txHash common.Hash,
...
-		return tx
+	// Make sure increase is at most 5x the suggested values
+	maxTip := new(big.Int).Mul(tip, big.NewInt(feeLimitMultiplier))
+	if bumpedTip.Cmp(maxTip) > 0 {
+		m.l.Warn("bumped tip getting capped at 5x the suggested value", "bumped", bumpedTip, "suggestion", tip)
+		bumpedTip.Set(maxTip)
+	}
+	maxFee := calcGasFeeCap(new(big.Int).Mul(basefee, big.NewInt(feeLimitMultiplier)), maxTip)
+	if bumpedFee.Cmp(maxFee) > 0 {
+		m.l.Warn("bumped fee getting capped at 5x the implied suggested value", "bumped", bumpedFee, "suggestion", maxFee)
```
sebastianst, line 452: Now that `5` is in a `const feeLimitMultiplier`, should we print the const value here instead? If we ever change the const, the message will show a wrong number.

sebastianst (commented):
decreasing bump to 10% from 15% also makes sense to me 👍🏻

op-service/txmgr/txmgr.go
roberto-bayardo: done

op-service/txmgr/txmgr.go
```diff
@@ -418,32 +418,42 @@ func (m *SimpleTxManager) queryReceipt(ctx context.Context, txHash common.Hash,
...
-// We do not re-estimate the amount of gas used because for some stateful transactions (like output proposals) the
-// act of including the transaction renders the repeat of the transaction invalid.
-//
-// If it encounters an error with creating the new transaction, it will return the old transaction.
+// We do not re-estimate the amount of gas used because for some stateful transactions (like output
+// proposals) the act of including the transaction renders the repeat of the transaction invalid.
 func (m *SimpleTxManager) increaseGasPrice(ctx context.Context, tx *types.Transaction) *types.Transaction {
+	m.l.Info("bumping gas price for tx", "hash", tx.Hash(), "tip", tx.GasTipCap(), "fee", tx.GasFeeCap(), "gaslimit", tx.Gas())
+	var tip, basefee *big.Int
 	tip, basefee, err := m.suggestGasPriceCaps(ctx)
```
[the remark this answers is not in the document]
roberto-bayardo: indeed, fixed

op-service/txmgr/txmgr.go
trianglesphere: the tx is invalidated by itself. Specifically if the proposal transaction is included on chain, but the confirmation depth has not been hit, replaying the transaction in the pending state will alway fail because the output has already been proposed.

roberto-bayardo: and in this case re-estimating gas would return an error and abort the replay, which seems the desirable outcome? (The original tx should eventually confirm and then progress will resume from there.)

roberto-bayardo: oh I see, the error could cause the original confirmation wait loop to abort before reaching confirmation & reading the receipt, resulting in an unnecessary re-submision by the proposer?

roberto-bayardo: Oh and indeed I do see this happen in our logs, once in the last 24 hours. I think all we need to do here is to continue looping waiting on the previous tx should increaseGasPrice error out.

roberto-bayardo:
I think this PR is ready to go. Running since yesterday proposing to goerli and in the last 12 hours we see this from the proposer transactions:

- 0 tx underpriced / already known errors
- 1 case of super-high gas limit corrected by gas-re-estimation
- 0-4 cases of retrying a tx with higher fees every hour.  Almost all txs only that were retried were retried only once. One tx was retried 3 times but fee suggestions were _extremely_ low at the time. The 3rd retry cleared with feecap of 600 wei and a tip of 400 wei (note that's wei not gwei!)
- 4 cases of tx retry resulted in "gas estimation failure", causing the retry to abort and allowing the original tx to confirm successfully.  The remaining ones resulted in successful tx replacement.
- other than the tx re-estimation related aborts above, only other reason for a tx failing without receipt was because it was sucessfully replaced

Let me know if there's anything else I can do?

sebastianst (approved):
LGTM thanks!

I'd also be fine with just adjusting the tests to the 10% price bump, instead of doing
```
func init() {
	// fix price bump here (15%) so tests won't break if the default changes
	priceBumpPercent = big.NewInt(100 + 15)
}
```

roberto-bayardo:
> LGTM thanks!
> 
> I'd also be fine with just adjusting the tests to the 10% price bump, instead of doing

Tests adjusted for a 10% price bump, and added a require guard on priceBump for tests that rely on it so it's easy to know why things break if priceBump does get modified.

trianglesphere (approved):
lgtm. Thanks for these changes

roberto-bayardo:
@mergifyio requeue

PR #6034, opened 2023-06-16
ops: bring back scripts

tynes:
**Description**

Bring back `ci-builder` to the monorepo as it should have never been deleted. Also bring back various tagging scripts. It is not clear which tagging script is used at the moment, need clarity on which should be actually brought back before merging this. `ci-versions.js` is required for changesets, so bring that back for sure. `update-op-geth.py` is used but unclear about the 2 other shell scripts.


ops/scripts/tag-bedrock-go-modules.sh
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env bash
```
trianglesphere, line 1: Not used as far as I know.

ops/scripts/update-op-geth.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
```
trianglesphere, line 1: I believe this may still be used.

zhwrd:
I believe `tag-bedrock-go-modules.sh` is unused, but [ops/scripts/ci-docker-tag-op-stack-release.sh](https://github.com/ethereum-optimism/optimism/pull/6034/files#diff-a6781a83c6d4ad088a7959d923836ca5aeb5288a9b5a1792ff91faa964bbb618) is used.

ops/scripts/ci-docker-tag-op-stack-release.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
trianglesphere, line 1: This is used in CI IIRC.

zhwrd (approved):

PR #6039, opened 2023-06-16
ops: update L1 geth for devnet

tynes:
**Description**

Updates to the latest geth for the devnet to ensure that our system works against the latest changes
to L1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1008,7 +1008,7 @@ jobs:
     steps:
       - checkout
       - check-changed:
-          patterns: op-(.+),packages
+          patterns: op-(.+),packages,ops-bedrock
       - run:
           name: Install latest golang
           command: |
--- a/ops-bedrock/Dockerfile.l1
+++ b/ops-bedrock/Dockerfile.l1
@@ -1,4 +1,4 @@
-FROM ethereum/client-go:v1.11.2
+FROM ethereum/client-go:v1.12.0
 
 RUN apk add --no-cache jq
 


tynes:
An interesting error in the devnet logs: `execution reverted: L2OutputOracle: block hash does not match the hash at the expected height`

Inphi:
> An interesting error in the devnet logs: `execution reverted: L2OutputOracle: block hash does not match the hash at the expected height`

Does this occur in a test we have?

ajsutton:
Yeah that's not a one off.  Geth 1.12.0 has changed the default for `estimateGas` from using `pending` to `latest` so the current chain head block hash isn't available on-chain.

ajsutton:
Ran into the same thing in https://github.com/ethereum-optimism/optimism/pull/6138 and pulling the 1.12.0 changes into op-geth https://github.com/ethereum-optimism/optimism/pull/6155/

roberto-bayardo:
BTW this EstimateGas behavior update is why we were seeing the stuck proposal transaction issue and you were not (we were running the newer geth all along).  The race condition situation we identified in EstimateGas could only be triggered by the output proposal transaction initially everting in EstimateGas and then suddenly not reverting at the end of the binary search.  Since your EstimateGas calls were running on pending state (old geth behavior) the proposal transactions generally wouldn't revert in this manner, so problem avoided.

TL;DR: you should pair this geth update with https://github.com/ethereum-optimism/optimism/pull/6031 in your proposer to avoid getting hit with stuck transactions!

ajsutton:
@roberto-bayardo I'm also leaning towards making `op-proposer` explicitly specify `pending` when estimating gas. It uses the op-geth ethclient so I can add the `EstimateGasAt` method there without having to wait for it to be merged into upstream geth.  Would be interested to know your thoughts on that.

roberto-bayardo:
> @roberto-bayardo I'm also leaning towards making `op-proposer` explicitly specify `pending` when estimating gas. It uses the op-geth ethclient so I can add the `EstimateGasAt` method there without having to wait for it to be merged into upstream geth. Would be interested to know your thoughts on that.

Oh yes, if you're willing to diverge ethclient (which seems fine?) then that makes a lot of sense. I don't think there's any downside to using pending over latest, and as we see there is a downside.  I think the wait loop I added in https://github.com/ethereum-optimism/optimism/pull/6138 is still helpful even with operating on pending state, since it seems the head of the chain can in general be a bit unstable, but running over pending state would at least allow us to safely remove the "+1" I have in there.

ajsutton:
With #6138 now merged we should be able to go ahead with this change. I'm leaning back away from explicitly doing estimation using the `pending` block because not all clients support it and we can avoid increasing the diff from geth, but #6138 should keep things working properly anyway.  Would be worth pulling in the latest changes from master.

protolambda (approved):

tynes:
If this passes in CI, it should auto merge now

PR #6044, opened 2023-06-16
feat(ctb): Move the Style Guide into a new doc

maurelian:
**Description**

We need to start treating our style guide as a first class citizen. This will be the place we reach agreement on style conventions, which inform the custom rules we'll eventually write to enforce them.

This is a minimal PR to set it off.


packages/contracts-bedrock/STYLE_GUIDE.md
```diff
@@ -0,0 +1,130 @@
+# Smart Contract Style Guide
+
+This document providing guidance on how we organize and write our smart contracts. For cases where
+this document does not provide guidance, please refer to existing contracts for guidance,
+with priority on the `L2OutputOracle` and `OptimismPortal`.
```
maurelian, line 5: this is the only part of the PR that could be at all controversial, but I think these contracts have had the most attention up until now and should be the final deciders if anything is unclear.

refcell (approved):

clabby (approved):

PR #6046, opened 2023-06-16
feat(ctb): Refactor LibPosition.sol to rep gindex

Inphi:
**Description**

Generalized Indices are a more succinct and easy to understand model for positions in the dispute game tree. Implementation is often more efficient too and can avoid branching logic for tree traversal.

**Tests**

The already existing LibPosition.t.sol


packages/contracts-bedrock/contracts/test/LibPosition.t.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity ^0.8.15;
 import { Test } from "forge-std/Test.sol";
 import { LibPosition } from "../dispute/lib/LibPosition.sol";
 import "../libraries/DisputeTypes.sol";
+import { console } from "forge-std/console.sol";
```
tynes, line 7: dead import, tbh solc should warn at this

packages/contracts-bedrock/contracts/dispute/lib/LibPosition.sol
```diff
@@ -91,13 +83,14 @@ library LibPosition {
...
     function rightIndex(Position _position, uint256 _maxDepth) internal pure returns (uint64 rightIndex_) {
+        uint256 msb = depth(_position);
         assembly {
-            rightIndex_ := shr(0xC0, shl(0xC0, _position))
-
-            // Walk down to the max depth by moving right
-            for { let i := shr(0x40, _position) } lt(i, sub(_maxDepth, 0x01)) { i := add(i, 0x01) } {
-                rightIndex_ := add(0x01, shl(0x01, rightIndex_))
+            let descent := sub(_maxDepth, msb)
+            for { let i := 0 } lt(i, descent) { i := add(i, 1) } {
```
Inphi, line 89: not very happy with the looping in this implementation. But I couldn't figure out a more efficient way.

clabby: Yeah, good question.. we can get rid of the loop with this:
```solidity
    function rightIndex(Position _position, uint256 _maxDepth) internal pure returns (uint64 rightIndex_) {
        uint256 msb = depth(_position);
        assembly {
            switch eq(msb, _maxDepth)
            case true {
                rightIndex_ := _position
            }
            default {
                let remaining := sub(_maxDepth, msb)
                rightIndex_ := or(shl(remaining, _position), sub(shl(remaining, 1), 1))
            }
        }
        rightIndex_ = indexAtDepth(Position.wrap(rightIndex_));
    }
```
but I also don't like how we have to resort to calling `indexAtDepth` again at the bottom. Do we want the gindex, or the trace index (index at depth) here? If we want the gindex, we can delete the final line of this func.

clabby:
There's also a small fix that will need to be made for the `test_gameDepthExceeded_reverts` in `FaultDisputeGame.t.sol`, just the off-by-one bug you found 😄

packages/contracts-bedrock/contracts/dispute/lib/LibPosition.sol
```diff
@@ -32,10 +35,11 @@ library LibPosition {
...
      */
     function indexAtDepth(Position _position) internal pure returns (uint64 indexAtDepth_) {
-        // Clean the high-order 192 bits by shifting the position left and then right again, leaving
-        // only the `indexAtDepth`.
+        // Return bits p_{msb}...p_{0}
+        uint256 msb = depth(_position);
         assembly {
-            indexAtDepth_ := shr(0xC0, shl(0xC0, _position))
+            let mask := sub(shl(msb, 1), 1)
+            indexAtDepth_ := and(_position, mask)
```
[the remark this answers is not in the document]
Inphi: yup, good catch.

packages/contracts-bedrock/contracts/dispute/lib/LibPosition.sol
Inphi: Thank you, this is what I wanted.
We actually can avoid calling `indexAtDepth` and thus avoid running another bitscan on the `rightIndex_`. We know that `rightIndex_` is at max depth right after the switch statement. All that's left is to clear the `_maxDepth` bit to get the actual index.

In any case, the gindex should suffice if `rightIndex` will only be used during resolution.

packages/contracts-bedrock/contracts/dispute/lib/LibPosition.sol
```diff
@@ -20,9 +20,12 @@ library LibPosition {
      * @return depth_ The `depth` of the `position`.
      */
     function depth(Position _position) internal pure returns (uint64 depth_) {
-        // Shift the high-order 64 bits into the low-order 64 bits, leaving only the `depth`.
+        // Return the most significant bit position
```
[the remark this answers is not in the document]
Inphi: ehh what the heck. let's go with this. The impl comes from an audited repo so I'm less afraid.

clabby (approved):
Nice!

PR #6055, opened 2023-06-18
feat(ctb): Add vizualizer script for the `FaultDisputeGame`

clabby:
## Overview

Adds a simple forge script that calls out to a python script to generate a visualization of a dispute game's state. Can be modified to pull a visualization from a game on a remote chain in the future.

### Sample (Mock values)
Screenshot 2023-06-17 at 7 55 57 PM


packages/contracts-bedrock/scripts/FaultDisputeGameViz.s.sol
```diff
@@ -0,0 +1,96 @@
...
+                position: position,
+                clock: clock
+            });
+        }
+
+        string[] memory commands = new string[](3);
+        commands[0] = "python3";
+        commands[1] = "scripts/dag-viz.py";
+        commands[2] = vm.toString(abi.encode(gameData));
+        vm.ffi(commands);
```
refcell, line 94: I wish there was a common library that encapsulated commands like python, rust etc etc and could relativize paths gracefully.

refcell: Something where I could just go `External.python3(...)` or something

packages/contracts-bedrock/scripts/FaultDisputeGameViz.s.sol
```diff
@@ -0,0 +1,96 @@
...
+                parentIndex: parentIndex,
+                countered: countered,
+                claim: claim,
+                position: position,
+                clock: clock
+            });
+        }
+
+        string[] memory commands = new string[](3);
+        commands[0] = "python3";
```
refcell, line 91: Actually, to avoid using python in its entirety, we could rip the script up into a new golang `package` inside `packages/` called something like `op-visual` that has a solidity wrapper library to run a go binary. Then this sol script could just import `op-visual` and generate the graph in a single invocation line. wdyt?

refcell (approved):

clabby:
@mergifyio queue

PR #6056, opened 2023-06-19
[txmgr] Call estimate gas twice

mdehoog:
**Description**
This PR adds an additional call to `estimateGas` in the `txmgr` transaction creation, and uses the minimum value returned. This is to ensure we use a correct gas limit for proposer transactions. This will cause slight overhead on tx submission, as well as increased load on the L1 node.

**Tests**
No tests added, this is a simple logic change.

**Additional context**
We believe Geth has a race condition bug where, if a contract call switches from reverting to passing during the `estimateGas` call, the gas calculation binary search can return very high values (see [this PR](https://github.com/ethereum/go-ethereum/pull/27502) for more details).

As an example, given an L1 block limit of 30,000,000, and the following binary search code: https://github.com/ethereum/go-ethereum/blob/46ec972c9c56a4e0d97d812f2eaf9e3657c66276/internal/ethapi/api.go#L1202-L1218

The binary search variables might look like:
```
lo = 15010500, hi = 30000000 # contract call reverts, thus gas is increased
lo = 22505250, hi = 30000000
lo = 26252625, hi = 30000000
lo = 28126312, hi = 30000000
lo = 29063156, hi = 30000000
lo = 29531578, hi = 30000000
lo = 29765789, hi = 30000000
lo = 29882894, hi = 30000000 # contract call starts working here due to internal state transition
lo = 29882894, hi = 29941447
lo = 29882894, hi = 29912170
lo = 29882894, hi = 29897532
lo = 29882894, hi = 29890213
lo = 29882894, hi = 29886553
lo = 29882894, hi = 29884723
lo = 29882894, hi = 29883808
lo = 29882894, hi = 29883351
lo = 29882894, hi = 29883122
lo = 29882894, hi = 29883008
lo = 29882894, hi = 29882951
lo = 29882894, hi = 29882922
lo = 29882894, hi = 29882908
lo = 29882894, hi = 29882901
lo = 29882894, hi = 29882897
lo = 29882894, hi = 29882895
```

Thus the tx is submitted with a gas limit of `29882895`. This is rejected by the network because of how high the limit is. We are experiencing a stuck proposer because of this bug, using values like this.

### DIFF
--- a/op-service/txmgr/txmgr.go
+++ b/op-service/txmgr/txmgr.go
@@ -192,17 +192,30 @@ func (m *SimpleTxManager) craftTx(ctx context.Context, candidate TxCandidate) (*
 		rawTx.Gas = candidate.GasLimit
 	} else {
 		// Calculate the intrinsic gas for the transaction
-		gas, err := m.backend.EstimateGas(ctx, ethereum.CallMsg{
+		call := ethereum.CallMsg{
 			From:      m.cfg.From,
 			To:        candidate.To,
 			GasFeeCap: gasFeeCap,
 			GasTipCap: gasTipCap,
 			Data:      rawTx.Data,
-		})
+		}
+		// Some implementations of EstimateGas have a race condition that can cause extremely high
+		// estimates for contract interactions. This can cause stuck transactions, because the
+		// returned estimated gas is close to the block limit, and nodes will ignore the tx. Here
+		// we call EstimateGas twice, and use the minimum of the two values.
+		gas, err := m.backend.EstimateGas(ctx, call)
 		if err != nil {
 			return nil, fmt.Errorf("failed to estimate gas: %w", err)
 		}
-		rawTx.Gas = gas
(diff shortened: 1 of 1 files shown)


tynes:
It looks like the fix has been merged into upstream geth and will be released in 1.12.1 - https://github.com/ethereum/go-ethereum/pull/27505

We have no guarantee this will run in prod until the next hardfork. I don't love calling estimate gas twice, would the correct solution just be to have a sanity cap and not allow the gas limit to be higher than that value?

mdehoog:
Agree I don't love this either. We do have the ability to hard code the gas, but we're not using it yet: https://github.com/ethereum-optimism/optimism/blob/e1571d740059d75585ee0e34eadcf4a7fb7ac2a3/op-proposer/proposer/l2_output_submitter.go#L331

Maybe a good fix would be to simply expose this as a command line flag. We can consider this if we continue to see issues.

Closing this for now.

PR #6059, opened 2023-06-19
feat(ctb): Implement fault dispute game resolution

Inphi:
**Description**

Implement resolution for fault dispute games.

**Metadata**

- Fixes CLI-4101


packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -268,6 +268,61 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
...
+     * @dev Static unit test for the correctness of resolution.
+     */
+    function test_resolve_team_deathmatch() public {
+        gameProxy.attack(0, Claim.wrap(bytes32(uint256(5))));
+        gameProxy.attack(0, Claim.wrap(bytes32(uint256(4))));
+        gameProxy.defend(1, Claim.wrap(bytes32(uint256(6))));
+        gameProxy.defend(1, Claim.wrap(bytes32(uint256(7))));
+
+        GameStatus status = gameProxy.resolve();
+        assertEq(uint(status), uint(GameStatus.DEFENDER_WINS));
```
clabby, line 323: The `GameStatus` should be `CHALLENGER_WINS` due to the uncontested attack, this test is currently passing due to an error in the `resolve` function where the index at which we're querying for the `winner` never gets updated.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -221,8 +221,38 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
...
+            }
+        }
+
+        ClaimData memory winner = claimData[leftMostIndex];
+        if (LibPosition.depth(winner.position) % 2 == 0) {
+            status = GameStatus.DEFENDER_WINS;
+        } else {
+            status = GameStatus.CHALLENGER_WINS;
+        }
         status_ = status;
```
clabby, line 256: This is a great start. Found 1 error and optimized this a bit, feel free to pick and choose:
```suggestion
        if (status != GameStatus.IN_PROGRESS) {
            revert GameNotInProgress();
        }

        // Search for the left-most dangling non-bottom node
        // The most recent claim is always a dangling, non-bottom node so we start with that
        uint256 leftMostIndex = claimData.length - 1;
        uint256 leftMostTraceIndex = LibPosition.rightIndex(
            claimData[leftMostIndex].position,
            MAX_GAME_DEPTH
        );
        for (uint256 i = leftMostIndex; i < type(uint64).max; ) {
            // Fetch the claim at the current index.
            ClaimData storage claim = claimData[i];

            // Decrement the loop counter; If it underflows, we've reached the root
            // claim and can stop searching.
            unchecked {
                --i;
            }

            // If the claim is not a dangling node above the bottom of the tree,
            // we can skip over it. These nodes are not relevant to the game resolution.
            Position claimPos = claim.position;
            if (LibPosition.depth(claimPos) == MAX_GAME_DEPTH || claim.countered) {
                continue;
            }

            // If the claim is a dangling node, we can check if it is the left-most
            // dangling node we've come across so far. If it is, we can update the
            // left-most trace index.
            uint256 traceIndex = LibPosition.rightIndex(claimPos, MAX_GAME_DEPTH);
            if (traceIndex < leftMostTraceIndex) {
                leftMostTraceIndex = traceIndex;
                leftMostIndex = i + 1;
            }
        }

        // If the left-most dangling node is at an even depth, the defender wins.
        // Otherwise, the challenger wins and the root claim is deemed invalid.
        if (LibPosition.depth(claimData[leftMostIndex].position) % 2 == 0) {
            status_ = GameStatus.DEFENDER_WINS;
        } else {
            status_ = GameStatus.CHALLENGER_WINS;
        }

        // Update the game status
        status = status_;
        emit Resolved(status_);
```

clabby (changes requested):

clabby (approved):
This is looking great to me - verified the correctness of this implementation vs. the python impl laid out in the spike.

I'd like to see an invariant test that uses `n` dishonest actors alongside an honest actor to fuzz its correctness, but that can be left for another PR.

Inphi:
> Need to rebase to pass `yarn-monorepo` task - couple JS dependencies have been added / removed the past week or so.

Looks like that task is also failing on develop. This PR is rebased on recent changes, including the change that removed JS deps.
@tynes any ideas why we're seeing failures here?

tynes:
Will take a look

tynes:
https://github.com/ethereum-optimism/optimism/pull/6078 fixes this issue

tynes:
@Inphi you can rebase this PR now and it should be good

tynes:
Looks like the gas snapshot is off. If you can't recreate the diff, its prob stale deps. Delete the node modules and then reinstall

```
rm -rf node_modules && rm -rf packages/**/node_modules && yarn --ignore-engines
```

Inphi:
> Looks like the gas snapshot is off. If you can't recreate the diff, its prob stale deps. Delete the node modules and then reinstall
> 
> ```
> rm -rf node_modules && rm -rf packages/**/node_modules && yarn --ignore-engines
> ```

didn't work. Something is off for me locally that causes a different gas snapshot. The gas snapshot diff only contain changes to the `GasBenchMark_` tests. I've manually reverted the gas used for those tests as they have nothing to do with this change.

clabby:
@mergifyio queue

PR #6060, opened 2023-06-19
fix: overwrite bootnodes if --p2p.bootnodes provided

keroro520:
Based on https://github.com/ethereum-optimism/optimism/pull/4621, the bootnodes should be overriden by `--p2p.bootnodes` but not appended. And it's useless to append the default bootnodes for customized networks.


protolambda:
We can look into changing the default behavior here (will discuss with team), but note that bootnodes are DiscV5 nodes, and the Discv5 DHT is shared between all networks. These nodes do not have to be different between networks.

ajsutton:
For the record, every other client I'm aware of replaces the boot nodes when `--bootnodes` is specified. It's very useful when running a local testnet to keep your discovery completely separate - otherwise it becomes very hard to find your local nodes because they get drowned out by all the remote ones and the remote nodes won't accept them into the DHT because they're not publicly reachable so it all kind of breaks down.

protolambda (approved):

protolambda:
Agree with the argument from @ajsutton . Rerunning CI from where it failed. Might have to merge-in `develop` if that doesn't fix CI.

PR #6069, opened 2023-06-20
op-node: Add application specific peer scores for p2p alt-sync

ajsutton:
**Description**

Updates application specific scores based on p2p alt sync events. Include application specific scores in overall gossip scores.

The app score params are pretty arbitrary at the moment as we don't have any real data but are designed to be very forgiving to avoid banning peers unnecessarily.  In particular, we never ban a peer based on error responses alone because p2p alt sync isn't enabled by default and so most peers rejects requests to open the stream which counts as an error response.

**Additional context**

Builds on #6068 

**Metadata**

- Fixes #[Link to Issue]


ajsutton:
Running this on a local node connected to goerli with alt-p2p sync enabled, there is a bunch of initial downscoring because peers don't support p2p sync requests but not enough to disconnect peers. Then once settles into sync properly (suspect after the safe head updates again so it isn't missing gossip) that stops and all the peers wind up back to positive scores (so decaying works well).

protolambda (approved):

PR #6077, opened 2023-06-20
specs: span-batches spec draft

protolambda:
**Description**

Draft of new batch type: the "span-batch". This is a consensus-breaking change, and can be rolled out through a future network upgrade.

This reduces the overhead significantly when submitting large ranges of empty/low-volume/sparse/volatile blocks: enabling new use-cases of the OP-Stack through improved efficiency.

**Metadata**

Fix CLI-4137


specs/span-batches.md
```diff
@@ -0,0 +1,203 @@
...
+
+[protobuf spec]: https://protobuf.dev/programming-guides/encoding/#varints
+
+Where:
+
+- `prefix = rel_timestamp ++ parent_check + l1_origin_check`
+  - `rel_timestamp`: relative time since genesis, i.e. `anchor.timestamp - config.genesis.timestamp`.
+  - `parent_check`: first 20 bytes of parent hash, i.e. `anchor.parent_hash[:20]`.
+  - `l1_origin_check`: to ensure the intended L1 origins of this span of
+        L2 blocks are consistent with the L1 chain, the blockhash of the last L1 origi is referenced.
```
refcell, line 75:
```suggestion
        L2 blocks are consistent with the L1 chain, the blockhash of the last L1 origin is referenced.
```

refcell, line 75:
```suggestion
        L2 blocks are consistent with the L1 chain, the blockhash of the last L1 origin is referenced.
```

specs/span-batches.md
```diff
@@ -0,0 +1,203 @@
...
+              without the next L1 origin we cannot yet determine if time invariant could have been kept.
+            - If `batch.timestamp >= next_epoch.time` -> `drop`:
+              the batch could have adopted the next L1 origin without breaking the `L2 time >= L1 time` invariant.
+        - `len(batch.transactions) > 0`: -> `drop`:
+          when exceeding the sequencer time drift, never allow the sequencer to include transactions.
+- And for all transactions:
+  - `drop` if the `batch.transactions` list contains a transaction
+    that is invalid or derived by other means exclusively:
+    - any transaction that is empty (zero length `tx_data`)
+    - any [deposited transactions][g-deposit-tx-type] (identified by the transaction type prefix byte in `tx_data`)
```
refcell, line 181: I think the `g-deposit-tx-type` inline link is missing

specs/span-batches.md
```diff
@@ -0,0 +1,203 @@
...
+### Batcher
+
+Instead of transforming L2 blocks into batches a channel,
+the blocks should be buffered to form a span-batch.
+
+Ideally the L2 blocks are buffered as block-inputs, to maximize the span of blocks covered by the span-batch:
+span-batches of single L2 blocks do not increase efficiency as much as with larger spans.
+
+This means that the `(c *channelBuilder) AddBlock` function is changed to
+not directly call `(co *ChannelOut) AddBatch` but defer that until a minimum number of blocks have been buffered.
```
refcell, line 196: Does this need to account for the timestamp drift rule now?

specs/span-batches.md
protolambda: fixed, thank you

specs/span-batches.md
```diff
@@ -0,0 +1,206 @@
...
+Introduce version `1` to the [batch-format](./derivation.md#batch-format) table:
+
+| `batch_version` | `content`           |
+|-----------------|---------------------|
+| 1               | `prefix ++ payload` |
+
+Notation:
+`++`: concatenation of byte-strings.
+`anchor`: first L2 block in the span
+`uvarint`: unsigned Base128 varint, as defined in [protobuf varint spec]
```
refcell, line 68: Think the `varint` was missing in the reference link below
```suggestion
`uvarint`: unsigned Base128 varint, as defined in [protobuf spec]
```

refcell, line 68: Think the `varint` was missing in the reference link below
```suggestion
`uvarint`: unsigned Base128 varint, as defined in [protobuf spec]
```

specs/span-batches.md
```diff
@@ -0,0 +1,206 @@
...
+
+[protobuf spec]: https://protobuf.dev/programming-guides/encoding/#varints
+
+Where:
+
+- `prefix = rel_timestamp ++ parent_check + l1_origin_check`
```
refcell, line 74:
```suggestion
- `prefix = rel_timestamp ++ parent_check ++ l1_origin_check`
```

specs/span-batches.md
```diff
@@ -0,0 +1,206 @@
...
+  - `rel_timestamp`: relative time since genesis, i.e. `anchor.timestamp - config.genesis.timestamp`.
+  - `parent_check`: first 20 bytes of parent hash, i.e. `anchor.parent_hash[:20]`.
+  - `l1_origin_check`: to ensure the intended L1 origins of this span of
+        L2 blocks are consistent with the L1 chain, the blockhash of the last L1 origin is referenced.
+        The hash is truncated to 20 bytes for efficiency, i.e. `anchor.l1_origin.hash[:20]`.
+- `payload = block_count ++ block_tx_counts ++ tx_data_headers ++ tx_data + tx_sigs`:
+  - `block_count`: `uvarint` number of L2 blocks.
+  - `origin_bits`: bitlist of `block_count` bits, right-padded to a multiple of 8 bits:
+    1 bit per L2 block, indicating if the L1 origin changed this L2 block.
+  - `block_tx_counts`: for each block, a `uvarint` of `len(block.transactions)`.
```
refcell, line 84: Is it more efficient to have the `block_tx_counts` represented by a bitlist as well rather than a list of `varint`s?
Also, then would it be possible to completely remove the need for the independent `block_count` variable?

specs/span-batches.md
```diff
@@ -0,0 +1,206 @@
...
+  - `tx_data_headers`: lengths of each `tx_data` entry, encodes as concatenated `uvarint` entries, (empty if there are
+    no entries).
+  - `tx_data`: [EIP-2718] encoded transactions.
+    - The `tx_signature` is truncated from each [EIP-2718] encoded tx. To be reconstructed from `tx_sigs`.
+      - `legacy`: starting at `v` RLP field
+      - `1` ([EIP-2930]): starting at `signatureYParity` RLP field
+      - `2` ([EIP-1559]): starting at `signature_y_parity` RLP field
```
ajsutton, line 91: To confirm, the signature fields are completely omitted and the `1`, `2` prefix here is referring to the EIP-2718 transaction type prefix which is already present.

specs/span-batches.md
```diff
@@ -0,0 +1,206 @@
...
+  - `tx_sigs`: concatenated list of fixed-length transactions:
+    - `legacy` txs `v` value is reduced to `y_parity` (chain ID is already known)
+    - `v` is encoded as `uint8`
```
ajsutton, line 94: Legacy transactions aren't required to encode the chain ID in v. So I think we ether need the full `v` or both `y_parity` and a second boolean for whether the chain ID was specified or not.

But I'm not understanding why `v` is encoded as `uint8` if it's being reduced to a bool...

ajsutton (commented):
Makes sense to me. Really just a question around how the `v` value for legacy transactions is encoded and recovered.

specs/span-batches.md
protolambda: The bitlist would still need something to delimit the end of the bits, which effectively results in a similar format. Definitely can experiment more with this format, but think we need a back-testing tool first.

specs/span-batches.md
protolambda: yes, the idea is to separate signature data, so the random signatures don't interrupt the list of compressible tx data

specs/span-batches.md
protolambda: Removed the assumption around chain-id, now just encoding the full `v` value.

refcell (approved):

PR #6079, opened 2023-06-20
feat(ctb): Style Guide Updates

refcell:
***Description***

Update the style guide to use triple slash natspec instead of seaport. Pending approval, subsequent prs will migrate the contract comment styling across `contracts-bedrock` following this PR being merged.

### DIFF
--- a/packages/contracts-bedrock/STYLE_GUIDE.md
+++ b/packages/contracts-bedrock/STYLE_GUIDE.md
@@ -10,8 +10,8 @@ with priority on the `L2OutputOracle` and `OptimismPortal`.
 
 #### Comments
 
-We use [Seaport](https://github.com/ProjectOpenSea/seaport/blob/main/contracts/Seaport.sol)-style comments with some minor modifications.
-Some basic rules:
+Optimism smart contracts follow the triple-slash [solidity natspec comment style](https://docs.soliditylang.org/en/develop/natspec-format.html#documentation-example)
+with additional rules. These are:
 
 - Always use `@notice` since it has the same general effect as `@dev` but avoids confusion about when to use one over the other.
 - Include a newline between `@notice` and the first `@param`.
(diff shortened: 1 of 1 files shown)


packages/contracts-bedrock/STYLE_GUIDE.md
```diff
@@ -28,30 +28,41 @@ We also have the following custom tags:
 #### Errors
 
 - Use `require` statements when making simple assertions.
-- Use `revert` if throwing an error where an assertion is not being made (no custom errors). See [here](https://github.com/ethereum-optimism/optimism/blob/861ae315a6db698a8c0adb1f8eab8311fd96be4c/packages/contracts-bedrock/contracts/L2/OVM_ETH.sol#L31) for an example of this in practice.
+- Use `revert` if throwing an error where an assertion is not being made (no custom errors).
```
clabby, line 31:
```suggestion
- Use `revert(string)` if throwing an error where an assertion is not being made (no custom errors).
```

maurelian:
Assuming this will require a bit of discussion, so disabling Mergify. 

Also when we make changes to the guide, should we apply them to the existing code at the same time?

refcell:
> Assuming this will require a bit of discussion, so disabling Mergify.
> 
> Also when we make changes to the guide, should we apply them to the existing code at the same time?

Thanks for adding the label! And good question - good practices probably warrant changes alongside or stacked ontop. Since the styling is not perfectly consistent across the codebase right now and we're not deploying anything, I think it's _ok_ to split it all up. That said, I could create a big stack ontop of this to apply the associated changes, it will likely just take a bit of work to keep the stack in sync with trunk.

packages/contracts-bedrock/STYLE_GUIDE.md
```diff
@@ -60,9 +71,12 @@ Unless explicitly discussed otherwise, you MUST include the following basic upgr
 
 ### Versioning
 
-All (non-library and non-abstract) contracts MUST extend the `Semver` base contract which exposes a `version()` function that returns a semver-compliant version string.
-During the Bedrock development process the `Semver` value for all contracts SHOULD return `0.0.1` (this is not particularly important, but it's an easy standard to follow).
-When the initial Bedrock upgrade is released, the `Semver` value MUST be updated to `1.0.0`.
+All (non-library and non-abstract) contracts MUST extend the `Semver` base contract which
+exposes a `version()` function that returns a semver-compliant version string.
+
+Contracts with a `Semver` value of less than `1.0.0` were pre-bedrock upgrade. When the
```
tynes, line 77: Perhaps we can generalize this by saying something along the lines of "contracts pre `1.0.0` are not production ready"

PR #6080, opened 2023-06-20
feat(ctb): Alphabet `step` function in `FaultDisputeGame`

clabby:
## Overview

Implements the step function for the alphabet dispute.

**Metadata**
Closes CLI-4102
Closes CLI-4142


clabby:
* develop
  * **PR #6080** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6080" target="_blank">Graphite</a>  👈
    * **PR #6081** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6081" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6080?utm_source=stack-comment).

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -92,12 +92,67 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
...
+            postStateClaim = parent.claim;
+        } else {
+            // If the step is a defense, the poststate exists elsewhere in the game state,
+            // and the parent claim is the expected pre-state.
+            preState = parent;
+            postStateClaim = state.claim;
+        }
+
+        // Assert that the prestate commits to the instruction at `gindex - 1`
+        if (preState.position.rightIndex(MAX_GAME_DEPTH) != Position.unwrap(stepPos) - 1) {
```
clabby, line 142: This will need to be altered to account for the step from the absolute prestate -> instruction #1.

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -326,34 +348,337 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
...
+                    // Walk up until the valid position that commits to the prestate's
+                    // trace index is found.
+                    while (
+                        Position.unwrap(statePos.parent().rightIndex(MAX_DEPTH)) ==
+                        Position.unwrap(statePos)
+                    ) {
+                        statePos = statePos.parent().parent().parent();
+                    }
+                } else {
+                    // TODO - Defend step logic.
```
clabby, line 453: This is not currently covered; Will implement a test case tonight after dinner.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -27,12 +27,12 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
...
      * @notice The current Semver of the FaultDisputeGame implementation.
      */
-    string internal constant VERSION = "0.0.1";
+    string internal constant VERSION = "0.0.2";
 
     /**
      * @notice The max depth of the game.
      */
-    uint256 internal constant MAX_GAME_DEPTH = 63;
+    uint256 internal constant MAX_GAME_DEPTH = 4;
```
Inphi, line 35: Could we make this `immutable` rather than a `constant`? This way it can be set correctly in various contexts including tests and in prod.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -27,12 +27,12 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
     /**
      * @notice The current Semver of the FaultDisputeGame implementation.
      */
-    string internal constant VERSION = "0.0.1";
+    string internal constant VERSION = "0.0.2";
```
refcell, line 30: nice

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
refcell, line 35: Why do we change this?

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -45,6 +45,12 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
      */
     Position internal constant ROOT_POSITION = Position.wrap(1);
 
+    /**
+     * @notice The absolute prestate of the instruction trace. This is a constant that is defined
+     *         by the program that is being used to execute the trace, in this case, Cannon.
+     */
```
refcell, line 51: I think we should just not mention cannon at all here so we won't need to change in the future on name change. Keep it abstractly interfaced.

refcell: Would just leave it at:
```suggestion
    /**
     * @notice The absolute prestate of the instruction trace. This is a constant that is defined
     *         by the program that is being used to execute the trace.
     */
```

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -92,12 +105,73 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
...
+        bytes calldata
     ) external {
-        // TODO - Call the VM to perform the execution step.
+        // Steps cannot be made unless the game is currently in progress.
+        if (status != GameStatus.IN_PROGRESS) {
+            revert GameNotInProgress();
+        }
+
+        // Get the parent. If it does not exist, the call will revert with OOB.
+        ClaimData storage parent = claimData[_parentIndex];
```
refcell, line 120: nice

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -92,12 +105,73 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
...
+
+        // Pull the parent position out of storage.
+        Position parentPos = parent.position;
+        // Determine the position of the step.
+        Position stepPos = _isAttack ? parentPos.attack() : parentPos.defend();
```
refcell, line 125: nit
```suggestion
        // Pull the parent position out of storage.
        Position parentPos = parent.position;

        // Determine the position of the step.
        Position stepPos = _isAttack ? parentPos.attack() : parentPos.defend();
```

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -219,17 +300,23 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
      * @inheritdoc IDisputeGame
      */
     function resolve() external returns (GameStatus status_) {
+        // TODO: Do not allow resolution before clocks run out.
+
         if (status != GameStatus.IN_PROGRESS) {
+            // If the game is not in progress, it cannot be resolved.
             revert GameNotInProgress();
+        } else if (!claimData[0].countered) {
```
Inphi, line 308: I think this else case is already covered by the for loop. Is this just optimizing for the (hopefully) common case?

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -27,7 +24,7 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
     /**
      * @dev The current version of the `FaultDisputeGame` contract.
      */
-    string internal constant VERSION = "0.0.1";
+    string internal constant VERSION = "0.0.2";
```
refcell, line 27: Don't think this is used anywhere?

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -279,10 +291,20 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
     }
 
     /**
-     * @dev Static unit test asserting that resolve reverts when the game is not in progress.
+     * @dev Static unit test asserting that resolve reverts when the game state is
+     *      not in progress.
```
refcell, line 295: Style guide is to not use the `@dev` tag (see: https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/STYLE_GUIDE.md#style)
```suggestion
     * @notice Static unit test asserting that resolve reverts when the game state is
     *      not in progress.
```

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -326,34 +348,337 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
...
+        Vm _vm
+    ) public virtual {
+        gameProxy = _gameProxy;
+        counterParty = _counterParty;
+        vm = _vm;
+    }
+
+    /**
+     * @notice Perform the next move in the game.
+     * @dev This is very janky atm.
```
refcell, line 380:
```suggestion
```

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -326,34 +348,337 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
...
+        // Resolve the game and assert that the honest player challenged the root
+        // claim successfully.
+        assertEq(uint8(gameProxy.resolve()), uint8(GameStatus.DEFENDER_WINS));
+        assertTrue(dishonest.failedToStep());
+    }
+}
+
+////////////////////////////////////////////////////////////////
+//                           ACTORS                           //
+////////////////////////////////////////////////////////////////
```
refcell, line 634: This might be good candidates to move into another file, but not strongly opinionated on this. Better ask @tynes or @maurelian

refcell (changes requested):

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -92,12 +105,73 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
...
+            revert InvalidParent();
+        }
+
+        // Determine the expected pre & post states of the step.
+        Claim preStateClaim;
+        Claim postStateClaim;
+        if (stepPos.indexAtDepth() == 0) {
+            // If the step position's index at depth is 0, the prestate is the absolute prestate
+            // and the post state is the parent claim.
+            preStateClaim = ABSOLUTE_PRESTATE;
```
Inphi, line 138: It makes sense that the `_stateIndex` ends up unused. But maybe we could also assert that input `_stateIndex` equals `ABSOLUTE_PRESTATE`, for extra safety?

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
clabby: Testing purposes; Tree padding isn't implemented yet 😄

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
clabby: See L89

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
clabby: This is within tests; I think the `@dev` tag is fine here personally.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -92,12 +105,73 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
      * @inheritdoc IFaultDisputeGame
      */
     function step(
-        uint256 _prestateIndex,
+        uint256 _stateIndex,
         uint256 _parentIndex,
```
Inphi, line 109: super nit: Would it be better to rename `_parentIndex`, maybe `claimIndex` instead? The naming makes less sense from a defense pov.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
clabby: This case isn't always covered; if the root claim was never contested, the loop will revert.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
clabby: Is there a need to do this other than for sanity purposes? If the step position's index at depth is zero, there's no other option for the prestate - doing this assertion just adds to the tx cost.

Inphi: It's just for sanity checking. I'm totally also OK with not doing this given the added gas cost.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
refcell: Sweet, thanks 👍

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
refcell: Ah missed because of gh collapse, thanks!

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
Inphi: What changed? [test_resolve_rootUncontested](https://github.com/ethereum-optimism/optimism/blob/37d0332ddc7547b2bdbf90d6bb16063cc2f8681c/packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol#L287) covered this case before.

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
refcell: Ok, I think we should then update the style guide - created a [linear ticket](https://linear.app/optimism/issue/CLI-4144/fixctb-style-guide-dev-targ) to track :)

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
clabby: Screenshot 2023-06-21 at 6 54 08 PM

Inphi: ahh gotcha. Thanks for explaining.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
tynes: We currently have a diff between these contracts and the other contracts where they are `Semver` and these are not. I'm not opposed to moving away from the immutable approach of `Semver` and just using `ISemver` but we need to standardize. This contract should be `ISemver`

clabby: @tynes Can do in a separate PR 👍

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -219,17 +301,23 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone {
...
+
         if (status != GameStatus.IN_PROGRESS) {
+            // If the game is not in progress, it cannot be resolved.
             revert GameNotInProgress();
+        } else if (!claimData[0].countered) {
+            // If the root claim has never been countered, it implicitly wins.
+            status_ = GameStatus.DEFENDER_WINS;
+            status = status_;
+            emit Resolved(status_);
+            return status_;
```
tynes, line 314: this is kinda weird, named return arg and also `return`

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
tynes, line 27: i dont think we need assertions on the version, it adds overhead

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -184,10 +196,10 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
     function test_gameDepthExceeded_reverts() public {
         Claim claim = Claim.wrap(bytes32(uint256(5)));
 
-        for (uint256 i = 0; i < 63; i++) {
+        for (uint256 i = 0; i < 5; i++) {
```
tynes, line 199: `MAX_GAME_DEPTH` instead of hardcoding 5?

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
tynes: Not super opposed, i think its fine how it is

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
clabby: Do you prefer the updated version?

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
refcell: Can we put up a linear ticket for this? I think whenever we split work out into another pr, we should get in the habit of writing a linear ticket so we don't forget/others can pick up the work

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
refcell: Agree, LGTM

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -326,34 +339,336 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
...
+        GamePlayer dishonest = new FullyDivergentPlayer();
+        super.init(honest, dishonest, Claim.wrap(bytes32(uint256(31))));
+    }
+
+    function test_resolvesCorrectly_succeeds() public {
+        // Play the game until a step is forced.
+        dishonest.play(0);
+
+        // Resolve the game and assert that the honest player challenged the root
+        // claim successfully.
```
Inphi, line 580:
```suggestion
        // Resolve the game and assert that the dishonest player challenged the root
        // claim unsuccessfully.
```

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -326,34 +339,336 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
...
+        GamePlayer dishonest = new HalfDivergentPlayer();
+        super.init(honest, dishonest, Claim.wrap(bytes32(uint256(31))));
+    }
+
+    function test_resolvesCorrectly_succeeds() public {
+        // Play the game until a step is forced.
+        dishonest.play(0);
+
+        // Resolve the game and assert that the honest player challenged the root
+        // claim successfully.
```
Inphi, line 616:
```suggestion
        // Resolve the game and assert that the dishonest player challenged the root
        // claim unsuccessfully.
```

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -326,34 +339,336 @@ contract FaultDisputeGame_Test is DisputeGameFactory_Init {
...
+                for (uint256 i = 0; i < len; i++) {
+                    (, , , Position pos, ) = gameProxy.claimData(i);
+                    if (Position.unwrap(pos) == Position.unwrap(statePos)) {
+                        stateIndex = i;
+                        break;
+                    }
+                }
+            }
+
+            // Perform the step and hault recursion.
```
Inphi, line 458:
```suggestion
            // Perform the step and halt recursion.
```

Inphi (commented):
LGTM! I can approve once the `GamePlayer` defense covers steps.

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
refcell: Created one [here](https://linear.app/optimism/issue/CLI-4145/featctb-inherited-semver-contract) :)

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
refcell: I think this is reasonable. If we find it to be illegible later on we can change, but for now it works and doesn't waste gas iiuc

refcell (approved):

PR #6082, opened 2023-06-20
Split out initial files

trianglesphere:
**Description**

Splits the code from the initial file into several.


op-challenger/fault/types.go
```diff
@@ -0,0 +1,28 @@
...
+	Get(i int) (common.Hash, error)
+}
+
+type Claim struct {
+	Value common.Hash
+	Position
+}
+
+type Response struct {
+	Attack bool // note: can we flip this to true == going right / defending??
```
ajsutton, line 26: Yes, I think it makes a lot of sense to just use move left/right instead of attack/defend - it's much clearer.  Probably worth thinking about renaming things on the contract side as well.

ajsutton (approved):

PR #6081, opened 2023-06-20
chore(ctb): `LibPosition` documentation, tests, perf

clabby:
## Overview

Adds some documentation to `LibPosition` as well as several enhancements suggested by @trianglesphere.

Also adds a new static unit test, `test_pos_correctness`, that asserts the correctness of all gindex, (depth, indexAtDepth), and trace index combinations for a tree with a max depth of 4.

PR #6086, opened 2023-06-20
game solver: Add generalized index & tests to position

trianglesphere:
**Description**

This adds support for the generalized index & tests for the position object.

**Tests**

Adds lots of tests.

### DIFF
--- a/op-challenger/fault/position.go
+++ b/op-challenger/fault/position.go
@@ -11,8 +11,7 @@ type Position struct {
 	IndexAtDepth int
 }
 
-// TraceIndex calculates the what the index of the claim value
-// would be inside the trace.
+// TraceIndex calculates the what the index of the claim value would be inside the trace.
 func (p *Position) TraceIndex(maxDepth int) int {
 	lo := 0
 	hi := 1 << maxDepth
@@ -30,6 +29,7 @@ func (p *Position) TraceIndex(maxDepth int) int {
 	return mid
 }
 
+// move goes to the left or right child in the generalized index tree.
 func (p *Position) move(right bool) {
 	p.Depth++
 	p.IndexAtDepth = (p.IndexAtDepth << 1) | boolToInt(right)
@@ -43,15 +43,18 @@ func boolToInt(b bool) int {
 	}
 }
 
+// parent moves up to the parent in the generalized index tree.
 func (p *Position) parent() {
 	p.Depth--
 	p.IndexAtDepth = p.IndexAtDepth >> 1
 }
 
+// Attack moves this position to a position to the left which disagrees with this position.
 func (p *Position) Attack() {
 	p.move(false)
 }
 
+// Defend moves this position to the right which agrees with this position. Note:
 func (p *Position) Defend() {
 	p.parent()
 	p.move(true)
@@ -59,5 +62,30 @@ func (p *Position) Defend() {
 }
 
 func (p *Position) Print(maxDepth int) {
-	fmt.Printf("Trace Position is %04b\tTrace Depth is: %d\tTrace Index is: %d\n", p.IndexAtDepth, p.Depth, p.TraceIndex(maxDepth))
+	fmt.Printf("GIN: %4b\tTrace Position is %4b\tTrace Depth is: %d\tTrace Index is: %d\n", p.ToGIN(), p.IndexAtDepth, p.Depth, p.TraceIndex(maxDepth))
+}
+
+func (p *Position) ToGIN() uint64 {
+	return uint64(1<<p.Depth | p.IndexAtDepth)
+}
+
+func FromGIN(x uint64) Position {
+	depth := MSBIndex(x)
+	indexAtDepth := ^(1 << depth) & x
+	return Position{
+		Depth:        depth,
+		IndexAtDepth: int(indexAtDepth),
+	}
+}
+
+// MSBIndex returns the index of the most significant bit
+func MSBIndex(x uint64) int {
+	if x == 0 {
+		return 0
+	}
+	out := 0
+	for ; x != 0; out++ {
+		x = x >> 1
+	}
+	return out - 1
 }
--- a/op-challenger/fault/position.go
+++ b/op-challenger/fault/position.go
@@ -14,7 +14,7 @@ type Position struct {
 // TraceIndex calculates the what the index of the claim value would be inside the trace.
 func (p *Position) TraceIndex(maxDepth int) int {
 	lo := 0
-	hi := 1 << maxDepth
+	hi := (1 << maxDepth) - 1
 	mid := hi
 	path := p.IndexAtDepth
 	for i := p.Depth - 1; i >= 0; i-- {
--- a/op-challenger/fault/position.go
+++ b/op-challenger/fault/position.go
@@ -3,36 +3,42 @@ package fault
 import "fmt"
 
 // Position is a golang wrapper around the dispute game Position type.
-// Depth refers to how many bisection steps have occurred.
-// IndexAtDepth refers to the path that the bisection has taken
-// where 1 = goes right & 0 = goes left.
 type Position struct {
-	Depth        int
-	IndexAtDepth int
+	depth        int
+	indexAtDepth int
+}
+
+func NewPosition(depth, indexAtDepth int) Position {
+	return Position{depth, indexAtDepth}
+}
+
+func NewPositionFromGIndex(x uint64) Position {
+	depth := MSBIndex(x)
+	indexAtDepth := ^(1 << depth) & x
+	return NewPosition(depth, int(indexAtDepth))
+}
+
+func (p *Position) Depth() int {
+	return p.depth
+}
+
+func (p *Position) IndexAtDepth() int {
+	return p.indexAtDepth
 }
 
 // TraceIndex calculates the what the index of the claim value would be inside the trace.
+// It is equivalent to going right until the final depth has been reached.
 func (p *Position) TraceIndex(maxDepth int) int {
-	lo := 0
-	hi := (1 << maxDepth) - 1
-	mid := hi
-	path := p.IndexAtDepth
-	for i := p.Depth - 1; i >= 0; i-- {
-		mid = (lo + hi) / 2
-		mask := 1 << i
-		if path&mask == mask {
-			lo = mid
-		} else {
-			hi = mid
-		}
-	}
-	return mid
+	// When we go right, we do a shift left and set the bottom bit to be 1.
+	// To do this in a single step, do all the shifts at once & or in all 1s for the bottom bits.
+	rd := maxDepth - p.depth
+	return p.indexAtDepth<<rd | ((1 << rd) - 1)
 }
 
-// move goes to the left or right child in the generalized index tree.
+// move goes to the left or right child.
 func (p *Position) move(right bool) {
-	p.Depth++
-	p.IndexAtDepth = (p.IndexAtDepth << 1) | boolToInt(right)
+	p.depth++
+	p.indexAtDepth = (p.indexAtDepth << 1) | boolToInt(right)
 }
 
 func boolToInt(b bool) int {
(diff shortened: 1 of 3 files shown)


op-challenger/fault/position.go
```diff
@@ -30,6 +29,7 @@ func (p *Position) TraceIndex(maxDepth int) int {
 	return mid
 }
 
+// move goes to the left or right child in the generalized index tree.
```
ajsutton, line 32: I'm not sure what you mean by "generalised index tree" here. `Position` isn't using generalised indices - it's using separate Depth and IndexAtDepth.  You can convert between the two easily but they are different concepts.

op-challenger/fault/position_test.go
```diff
@@ -1 +1,122 @@
...
+			idx: 2,
+		},
+		{
+			pos: Position{Depth: 2, IndexAtDepth: 1}, // 0 or 1?
+			idx: 2,
+		},
+		{
+			pos: Position{Depth: 2, IndexAtDepth: 2}, // 2 or 3?
+			idx: 6,
+		},
```
ajsutton, line 100: `IndexAtDepth` has to monotonically increase for each node at a specific depth. It doesn't really matter if it starts from 0 or 1 but there has to be a power of 2 number of nodes at each depth except 0 and 1 which are the weird bit before the actual binary tree gets started. So I think we have an extra node at depth 2 which should only have 2. Should be `Depth: 2, IndexAtDepth: 0 -> idx: 2` and `Depth: 2, IndexAtDepth: 1 -> idx: 6`. 

I was going to suggest we steal the tests from `LibPosition` but it only has fuzz tests so is just repeating the production logic in the tests and not really testing anything...

So I suspect we need to get a small, concrete example of a tree with the correct generalised indices, depth/indexAtDepth and trace index for each node. Then we can plug them into both this test and LibPosition tests to ensure the two implementations actually match.

ajsutton: cc @clabby

clabby: Yep - spot on with the first paragraph.

Can work on this after #6080. I have a PR open (#6081) to update some documentation, could add it there. Some of the fuzz tests (`testFuzz_defend_correctness` & `testFuzz_rightIndex_correctness`) are moreso differential tests, but others are just sanity checks.

ajsutton: In my mind the tree looks like:
image
Where the `2,1` means depth 2, index 1. I'm not sure how you make generalised indices work in a non-binary tree like this but I'm guessing you pretend there's two children of the root and so wind up never using gIndex 3. So from depth, index to gIndex would be:

```
0,0 -> 1
1,0 -> 2
(1,1 should exist in a binary tree and would be gIndex 3, but can never exist in the game)
2,0 -> 4
2,1 -> 5
3,0 -> 6
3,1 -> 7
3,2 -> 8
3,3 -> 9
```

ajsutton: I'm assuming in that tree that the trace index rounds up when it winds up at a non-integer value but rounding down would also work.

op-challenger/fault/position.go
trianglesphere: I think they are different representations of the same concept. `generalized index = 1 << depth | IndexAtDepth`. Maybe generealized index tree isn't correct, but there is a tree that it's moving through.

op-challenger/fault/position_test.go
clabby: Not quite - You are correct that gindex 3 cannot exist, but it's left child can (gindex = 6). If the initial move at gindex = 2 is defended, the counter claim will be made at gindex = 6.

If we want to be particular, the number of nodes in each depth in the game is `2^d - 1`. In the tree, it's `2^d`. Effectively, none of these nodes can exist in the game because they all commit to the same trace index that the root claim does:
Screenshot 2023-06-20 at 7 07 37 PM

Depth to gIndex would be:
```
0,0 -> 1
1,0 -> 2
(1,1 should exist in a binary tree and would be gIndex 3, but can never exist in the game)
2,0 -> 4
2,1 -> 5
2,2 -> 6
2,3 -> 7
3,0 -> 8
3,1 -> 9
3,2 -> 10
3,3 -> 11
3,4 -> 12
3,5 -> 13
3,6 -> 14
3,7 -> 15
... and so on
```

op-challenger/fault/position.go
ajsutton: They are both ways of referring to nodes in a tree, but it's important not to confuse the two and a "generalised index tree" isn't really a thing.  You have a tree, and you can use generalised indices to refer to nodes.  I suspect the name we want is something like "bisection tree" but not really sure.  Would definitely be good to get some clear language agreed on this to make discussions easier.

op-challenger/fault/position_test.go
ajsutton: ahhh ok, there's a few concepts in there that I had wrong so this diagram is really useful. If we can add the trace indices onto that we'd have the data we need for the tests.

trianglesphere: That tree makes sense. It's been implemented with the high being 8 instead of 7, but I can change that. Maybe that also fixes the indexing.

I think what tripped me up with the generalized index <-> position conversion where the base cases. I think there might still be a bug in the position <-> trace index conversion.

ajsutton: I would suggest when we're converting between position and trace index we add tests for all nodes in that binary tree, even if the game will never hit them. It makes things so much easier to understand.  Then we can have separate tests for how attack/defend should move in the tree which will demonstrate that some of those nodes can never actually be reached.

trianglesphere: hmm, I'm also pretty sure that 2,2 & 2,3 should be able to exist. Depth 0 has 0 bits for the index. Depth 1 has 1 bit for the index at depth (i.e. 2^1 = 0b10) & Depth 2 has 2 bits for the index (2^2 = 0b100)

clabby: Just for my clarity / to align vocab: What do we mean when we say "trace index"?
1. The index at depth of the leaf node (i.e. for a tree of max depth = 4, `[0,2^4)`)
2. The gindex of the leaf node (i.e. for a tree of max depth = 4, `[16,32)`)

We plan on having the `rightIndex` function in `LibPosition` return the gindex of the leaf node (it currently returns the index at depth), but the index at depth can still be extracted from the gindex regardless.

ajsutton: By "trace index" I mean the index in the trace that we should get the claim value from.

clabby: Got it, so that'd be #1. In the tree I posted, the trace index by that definition would be `gIndex - depth << 1` (or more simply, the gindex with the most significant bit unset).

ajsutton: Yes. I'd like to be very clear about what the trace index is for non-leaf nodes as well so that we know we're picking the right value for each claim.

ajsutton: Right, so that's gIndex 6 has a trace index of 11 (max depth 4). Which is another difference to my previous understanding - I was thinking that the trace index was the middle of the descendant leaf nodes.  And that also explains why the attack/defend rules for movement are different to what I expect as well.

ajsutton: So in Go style I see that tree as:
```

type testNodeInfo struct {
	GIndex     uint64
	Position   Position
	TraceIndex uint64
}

var treeNodesMaxDepth4 = []testNodeInfo{
	{GIndex: 1, Position: Position{Depth: 0, IndexAtDepth: 0}, TraceIndex: 15},

	{GIndex: 2, Position: Position{Depth: 1, IndexAtDepth: 0}, TraceIndex: 7},
	{GIndex: 3, Position: Position{Depth: 1, IndexAtDepth: 1}, TraceIndex: 15},

	{GIndex: 4, Position: Position{Depth: 2, IndexAtDepth: 0}, TraceIndex: 3},
	{GIndex: 5, Position: Position{Depth: 2, IndexAtDepth: 1}, TraceIndex: 7},
	{GIndex: 6, Position: Position{Depth: 2, IndexAtDepth: 2}, TraceIndex: 11},
	{GIndex: 7, Position: Position{Depth: 2, IndexAtDepth: 3}, TraceIndex: 31},

	{GIndex: 8, Position: Position{Depth: 3, IndexAtDepth: 0}, TraceIndex: 1},
	{GIndex: 9, Position: Position{Depth: 3, IndexAtDepth: 1}, TraceIndex: 3},
	{GIndex: 10, Position: Position{Depth: 3, IndexAtDepth: 2}, TraceIndex: 5},
	{GIndex: 11, Position: Position{Depth: 3, IndexAtDepth: 3}, TraceIndex: 7},
	{GIndex: 12, Position: Position{Depth: 3, IndexAtDepth: 4}, TraceIndex: 9},
	{GIndex: 13, Position: Position{Depth: 3, IndexAtDepth: 5}, TraceIndex: 11},
	{GIndex: 14, Position: Position{Depth: 3, IndexAtDepth: 6}, TraceIndex: 13},
	{GIndex: 15, Position: Position{Depth: 3, IndexAtDepth: 7}, TraceIndex: 15},

	{GIndex: 16, Position: Position{Depth: 4, IndexAtDepth: 0}, TraceIndex: 0},
	{GIndex: 17, Position: Position{Depth: 4, IndexAtDepth: 1}, TraceIndex: 1},
	{GIndex: 18, Position: Position{Depth: 4, IndexAtDepth: 2}, TraceIndex: 2},
	{GIndex: 19, Position: Position{Depth: 4, IndexAtDepth: 3}, TraceIndex: 3},
	{GIndex: 20, Position: Position{Depth: 4, IndexAtDepth: 4}, TraceIndex: 4},
	{GIndex: 21, Position: Position{Depth: 4, IndexAtDepth: 5}, TraceIndex: 5},
	{GIndex: 22, Position: Position{Depth: 4, IndexAtDepth: 6}, TraceIndex: 6},
	{GIndex: 23, Position: Position{Depth: 4, IndexAtDepth: 7}, TraceIndex: 7},
	{GIndex: 24, Position: Position{Depth: 4, IndexAtDepth: 8}, TraceIndex: 8},
	{GIndex: 25, Position: Position{Depth: 4, IndexAtDepth: 9}, TraceIndex: 9},
	{GIndex: 26, Position: Position{Depth: 4, IndexAtDepth: 10}, TraceIndex: 10},
	{GIndex: 27, Position: Position{Depth: 4, IndexAtDepth: 11}, TraceIndex: 11},
	{GIndex: 28, Position: Position{Depth: 4, IndexAtDepth: 12}, TraceIndex: 12},
	{GIndex: 29, Position: Position{Depth: 4, IndexAtDepth: 13}, TraceIndex: 13},
	{GIndex: 30, Position: Position{Depth: 4, IndexAtDepth: 14}, TraceIndex: 14},
	{GIndex: 31, Position: Position{Depth: 4, IndexAtDepth: 15}, TraceIndex: 15},
}
```

And I think we should be able to iterate through that data to test both GIndex<->Position conversion and Position<->TraceIndex conversions.

clabby: Sol impl of the above test can be found here:
https://github.com/ethereum-optimism/optimism/blob/6153d03df17e005be3a4cb095c718974a8a6a6d7/packages/contracts-bedrock/contracts/test/LibPosition.t.sol#L157

trianglesphere: I've coded it to work with the following tree. Red means unreacheable. T is trace index. P is position. G is generalized index.
download

ajsutton: I think we're getting ahead of ourselves thinking about how we can move around the tree.  For any node we should be able to convert between gIndex, position and trace index.  That's what we need to get working as a primitive construct - it doesn't matter if movement rules allow us to get to a node or not, the conversions should still work.

If we mix movement rules into the conversion logic or tests, then we tightly couple movement and conversion when they should be independent.

trianglesphere: Ok, one test case was wrong (GIndex: 7 should be trace index 15, not 31).
Otherwise tests are all passing.

ajsutton (approved):
LGTM.

PR #6089, opened 2023-06-21
op-program/op-service: Revert retry client

ajsutton:
**Description**

Reverts the change to introduce a new, common retrying client at the RPC layer. The change breaks op-program because fetching L2 contract code requires trying an RPC call under the new key and then falling back to the legacy key. With the retry being done at the RPC level, it just retries requesting the new key forever and has no way to fall back to the legacy key.

So op-program needs to go back to handling retries at the client level, not the RPC level.  Since the retrying RPC client wasn't used anywhere else we may as well remove it entirely.

Note: There was a fix to the lint config for op-service mixed in with one of the PRs being reverted. Reverting here so the revert is clean, but will restore that fix in a followup.


trianglesphere:
> fetching L2 contract code requires trying an RPC call under the new key and then falling back to the legacy key

Where is this done? I'm not familiar with this part of the system.

ajsutton:
> Where is this done? I'm not familiar with this part of the system.

https://github.com/ethereum-optimism/optimism/blob/a7e24c937de307a7a746e9aa9ba894d003c5092e/op-node/sources/debug_client.go#L29-L40

refcell (approved):
LGTM, apologies my retrying refactored client broke things

ajsutton:
> LGTM, apologies my retrying refactored client broke things

Nah, I approved the initial change. :) It was a good idea but CI has done its job and told us about some extra complexity that we weren't aware of.

trianglesphere:
> > Where is this done? I'm not familiar with this part of the system.
> 
> https://github.com/ethereum-optimism/optimism/blob/a7e24c937de307a7a746e9aa9ba894d003c5092e/op-node/sources/debug_client.go#L29-L40

Oh I see - the prefetcher retries, but the L1 & L2 RPC doesn't anymore

trianglesphere (approved):

PR #6094, opened 2023-06-21
feat(op-challenger): AlphabetProvider + Naive Solver

refcell:
**Description**

- Refactors fault examples into `cmd/examples/`
- Added a new `Solver` example in `cmd/examples/`
- New game `Solver` instance
- Bunch of new alphabet and solver tests

**Metadata**

Fixes CLI-4122

### DIFF
--- a/op-challenger/fault/alphabet_provider_test.go
+++ b/op-challenger/fault/alphabet_provider_test.go
@@ -8,6 +8,38 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestAlphabetProvider_Get_ClaimsByTraceIndex tests the [fault.AlphabetProvider] Get function.
+func TestAlphabetProvider_Get_ClaimsByTraceIndex(t *testing.T) {
+	// Create a new alphabet provider.
+	canonicalProvider := NewAlphabetProvider("abcdefgh", uint64(3))
+
+	// Build a list of traces.
+	traces := []struct {
+		traceIndex   uint64
+		expectedHash common.Hash
+	}{
+		{
+			7,
+			common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000768"),
+		},
+		{
+			3,
+			common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000364"),
+		},
+		{
+			5,
+			common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000566"),
+		},
+	}
+
+	// Execute each trace and check the alphabet provider returns the expected hash.
+	for _, trace := range traces {
+		expectedHash, err := canonicalProvider.Get(trace.traceIndex)
+		require.NoError(t, err)
+		require.Equal(t, trace.expectedHash, expectedHash)
+	}
+}
+
 // FuzzIndexToBytes tests the IndexToBytes function.
 func FuzzIndexToBytes(f *testing.F) {
 	f.Fuzz(func(t *testing.T, index uint64) {
--- a/op-challenger/fault/cmd/examples/position.go
+++ b/op-challenger/fault/cmd/examples/position.go
@@ -1,0 +1,35 @@
+package examples
+
+import (
+	"github.com/ethereum-optimism/optimism/op-challenger/fault"
+)
+
+func PositionExampleOne() {
+	// Example 1
+	// abcdefgh
+	// abcdexyz
+	// go left to d, then right to f, then left to e
+	p := fault.Position{}
+	p.Print(3)
+	p.Attack()
+	p.Print(3)
+	p.Defend()
+	p.Print(3)
+	p.Attack()
+	p.Print(3)
+}
+
+func PositionExampleTwo() {
+	// Example 2
+	// abcdefgh
+	// abqrstuv
+	// go left r, then left to b, then right to q
+	p := fault.Position{}
+	p.Print(3)
+	p.Attack()
+	p.Print(3)
+	p.Attack()
+	p.Print(3)
+	p.Defend()
+	p.Print(3)
+}
--- a/op-challenger/fault/cmd/main.go
+++ b/op-challenger/fault/cmd/main.go
@@ -1,43 +1,12 @@
 package main
 
 import (
-	"github.com/ethereum-optimism/optimism/op-challenger/fault"
+	"github.com/ethereum-optimism/optimism/op-challenger/fault/cmd/examples"
 )
 
 func main() {
-	// Example 1
-	// abcdefgh
-	// abcdexyz
-	// go left to d, then right to f, then left to e
-	p := fault.Position{}
-	p.Print(3)
-	p.Attack()
-	p.Print(3)
-	p.Defend()
-	p.Print(3)
-	p.Attack()
-	p.Print(3)
+	examples.SolverExampleOne()
 
-	// GIN:    1	Trace Position is    0	Trace Depth is: 0	Trace Index is: 8
-	// GIN:   10	Trace Position is    0	Trace Depth is: 1	Trace Index is: 4
-	// GIN:  110	Trace Position is   10	Trace Depth is: 2	Trace Index is: 6
-	// GIN: 1100	Trace Position is  100	Trace Depth is: 3	Trace Index is: 5
-
-	// Example 2
-	// abcdefgh
-	// abqrstuv
-	// go left r, then left to b, then right to q
-	p = fault.Position{}
-	p.Print(3)
-	p.Attack()
-	p.Print(3)
-	p.Attack()
-	p.Print(3)
-	p.Defend()
-	p.Print(3)
-
-	// Trace Position is 0000	Trace Depth is: 0	Trace Index is: 8
-	// Trace Position is 0000	Trace Depth is: 1	Trace Index is: 4
-	// Trace Position is 0000	Trace Depth is: 2	Trace Index is: 2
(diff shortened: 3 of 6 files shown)


op-challenger/fault/cmd/examples/position.go
```diff
@@ -0,0 +1,45 @@
...
+	p := fault.Position{}
+	p.Print(3)
+	p.Attack()
+	p.Print(3)
+	p.Attack()
+	p.Print(3)
+	p.Defend()
+	p.Print(3)
+
+	// Trace Position is 0000	Trace Depth is: 0	Trace Index is: 8
```
trianglesphere, line 41: Can probably delete the comments at this point.

op-challenger/fault/cmd/examples/alphabet.go
```diff
@@ -0,0 +1,97 @@
...
+	canonical := "abcdefgh"
+	disputed := "abcdexyz"
+	maxDepth := 3
+	position := fault.Position{}
+	canonicalProvider := fault.NewAlphabetProvider(canonical, uint64(maxDepth))
+	disputedProvider := fault.NewAlphabetProvider(disputed, uint64(maxDepth))
+
+	// Print the initial state.
+	fmt.Println("Canonical state: ", canonical)
+	fmt.Println("Disputed state: ", disputed)
```
trianglesphere, line 26:
```suggestion
	fmt.Println("Disputed state:  ", disputed)
```

op-challenger/fault/cmd/examples/alphabet.go
```diff
@@ -0,0 +1,97 @@
...
+	fmt.Println()
+	fmt.Println("Proceeding with the following moves:")
+	fmt.Println("go left to d, then right to f, then left to e")
+
+	// The first move should be to go left to d (ie the claims shouldn't match).
+	firstIndex := position.TraceIndex(maxDepth) - 1
```
trianglesphere, line 32: Needs a rebase on top of #6086. That will fix the off by one.

op-challenger/fault/cmd/main.go
```diff
@@ -1,43 +1,12 @@
 package main
 
 import (
```
tynes, line 3: its kinda weird to import from another `cmd` package

refcell: definitely, most of this will be deleted though - just a demo / reference for now

op-challenger/fault/solver.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// agreeWithClaim returns true if the [Claim] is correct according to the internal [TraceProvider].
+func (s *Solver) agreeWithClaim(claim Claim) bool {
+	ourValue := s.traceAtPosition(&claim.Position)
+	return ourValue == claim.Value
+}
+
+// traceAtPosition returns the [common.Hash] from internal [TraceProvider] at the given [Position].
+func (s *Solver) traceAtPosition(p *Position) common.Hash {
```
trianglesphere, line 78: Should return the error here instead of suppressing it.

trianglesphere, line 78: Should return the error here instead of suppressing it.

op-challenger/fault/solver.go
```diff
@@ -0,0 +1,86 @@
...
+	// This should not be reached
+	return nil, ErrNoNextMove
+}
+
+func (s *Solver) doNothing() *Response {
+	return nil
+}
+
+// attack returns a response that attacks the claim.
+func (s *Solver) attack(claim Claim) *Response {
```
trianglesphere, line 58: Ditto on returning an error.

op-challenger/fault/solver.go
```diff
@@ -0,0 +1,86 @@
...
+// agreeWithClaim returns true if the [Claim] is correct according to the internal [TraceProvider].
+func (s *Solver) agreeWithClaim(claim Claim) bool {
+	ourValue := s.traceAtPosition(&claim.Position)
+	return ourValue == claim.Value
+}
+
+// traceAtPosition returns the [common.Hash] from internal [TraceProvider] at the given [Position].
+func (s *Solver) traceAtPosition(p *Position) common.Hash {
+	index := p.TraceIndex(s.gameDepth)
+	hash, err := s.Get(uint64(index))
```
trianglesphere, line 80: Can update `p.TraceIndex` in a different PR to return a uint64 to keep types consistent.

op-challenger/fault/solver.go
```diff
@@ -0,0 +1,86 @@
...
+		// Do nothing, we disagree with the parent, but this claim has correctly countered it
+		return s.doNothing(), nil
+	} else if !parentCorrect && !claimCorrect {
+		// We disagree with the parent so want to counter it (which the claim is doing)
+		// but we also disagree with the claim so there must be a difference to the left of claim
+		// Note that we will create the correct counter-claim for parent when it is evaluated, no need to do it here
+		return s.attack(claim), nil
+	}
+	// This should not be reached
+	return nil, ErrNoNextMove
```
trianglesphere, line 50: Do we expect people to handle the case `ErrNoNextMove`? If not we should create a named error object & instead return `errors.New("no next move")` (or some other way to saying that this is invalid).

op-challenger/fault/solver_test.go
```diff
@@ -0,0 +1,108 @@
...
+func TestSolver_NextMove(t *testing.T) {
+	tests := []struct {
+		claim  Claim
+		parent Claim
+		res    *Response
+	}{
+		{
+			claim: Claim{
+				Value:    common.HexToHash("0x000000000000000000000000000000000000000000000000000000000000077a"),
+				Position: NewPosition(0, 0),
```
trianglesphere, line 21: claim & parent should not have the same position.

op-challenger/fault/solver_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+// TestSolver_NextMove tests the NextMove function of the [Solver]
+// when the [Claim] is incorrect and the [Parent] is correct.
```
trianglesphere, line 11: `when the [Claim] is incorrect and the [Parent] is correct` should be part of the test name because there are multiple other options here.

op-challenger/fault/alphabet_provider_test.go
```diff
@@ -8,6 +8,58 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestAlphabetProvider_Positions tests the [fault.AlphabetProvider] interfaces correctly
+// with the [fault.Position] type.
+func TestAlphabetProvider_Positions(t *testing.T) {
```
trianglesphere, line 13: Unless you use `NextMove` you're only really testing the `Position` object here.
The expectedHash check is really testing the AlphabetProvider.

op-challenger/fault/cmd/examples/alphabet.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// Print the initial state.
+	fmt.Println("Canonical state: ", canonical)
+	fmt.Println("Disputed state: ", disputed)
+	fmt.Println()
+	fmt.Println("Proceeding with the following moves:")
+	fmt.Println("go left to d, then right to f, then left to e")
+
+	// The first move should be to go left to d (ie the claims shouldn't match).
+	firstIndex := position.TraceIndex(maxDepth)
```
trianglesphere, line 32: Can we use `NextMove` here?

op-challenger/fault/solver.go
refcell: No, so will just inplace `errors.New("no next move")`.

op-challenger/fault/solver.go
refcell: Added a comment note :)

op-challenger/fault/solver_test.go
```diff
@@ -0,0 +1,107 @@
...
+			},
+			&Response{
+				Attack: true,
+				Value:  common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000465"),
+			},
+		},
+	}
+
+	for _, test := range indices {
+		res, err := solver.NextMove(test.claim, parentClaim)
```
trianglesphere, line 103: Parent Claim should not be static - it should be layer higher than the claim. It's probably fine to skip the first test case because that's the root claim which is weird.

op-challenger/fault/solver_test.go
```diff
@@ -0,0 +1,107 @@
...
+		res    *Response
+	}{
+		{
+			claim: Claim{
+				Value:    common.HexToHash("0x000000000000000000000000000000000000000000000000000000000000077a"),
+				Position: NewPosition(0, 0),
+			},
+			parent: Claim{
+				Value:    common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000768"),
+				Position: NewPosition(0, 0),
```
trianglesphere, line 25: parent claim cannot be at the same position as the claim

op-challenger/fault/solver_test.go
refcell: It isn't static since the current position API mutates this variable inplace through the `Attack` and `Defend` methods.

refcell: Oops im incorrect, changing.

trianglesphere (approved):

PR #6095, opened 2023-06-21
fault-detector: remove pre-bedrock support and update tests

zchn:


packages/fault-detector/src/helpers.ts
```diff
@@ -1,11 +1,11 @@
 import { Contract, BigNumber } from 'ethers'
 import { Logger } from '@eth-optimism/common-ts'
 
-export interface OutputOracle<TSubmissionEventArgs> {
+export interface OutputOracle {
```
hamdiallam, line 4: not super important but technically this interface is no longer needed. My understanding of this interface was to abstract the differences`L2OutputOracle` and `StateCommitmentChain` into one.

Can we simply utilize the typescript ABI for this contract directly? It would provide better type safety as well

hamdiallam: fwiw this is not a blocking suggestion!

hamdiallam (approved):
LGTM! Few comments on removing the bedrock option and related flags

tynes:
@zchn Pushed a commit with the diff to the `yarn.lock` at 3b5bcc8

PR #6098, opened 2023-06-21
go: update to urfave v2, merge endpoint-monitor and op-heartbeat into main go module

protolambda:
**Description**

- Merge endpoint-monitor into root Go module
  - Updated to no longer depend on legacy l2geth.
- Merge op-heartbeat into root Go module
- Update all main Go-mod packages to urfave CLI v2.
   - op-batcher
   - op-bootnode
   - op-challenger
   - op-heartbeat
   - op-node
   - op-program/host
   - op-proposer
   - op-service
   - op-signer/client
   - op-wheel

Cannon + Geth are already using urfave CLI v2.

With this upgrade, in separate later PRs, we can improve op-service without complicating a later v2 migration, and utilize the improved context/shutdown functionality of urfave CLI v2.

In practice, most of the v2 CLI migration changes are:
- commands/flags now with pointers
- EnvVars instead of EnvVar
- import-path updates

**Tests**

No functionality changes

**Metadata**

Fix CLI-3946


op-service/log/cli.go
```diff
@@ -93,10 +93,10 @@ func ReadLocalCLIConfig(ctx *cli.Context) CLIConfig {
 
 func ReadCLIConfig(ctx *cli.Context) CLIConfig {
```
trianglesphere, line 94: Can we deprecate `ReadLocalCLIConfig`? Do we have any other spots where we had a difference between local & global?

trianglesphere (changes requested):
I think the flipped CheckRequired is the reason the tests are failing.
Lots of code to review, but it was pretty straightforwards.

op-batcher/compressor/cli.go
```diff
@@ -15,30 +15,33 @@ const (
 )
 
 func CLIFlags(envPrefix string) []cli.Flag {
+	prefixEnvVars := func(name string) []string {
```
tynes, line 18: Could this function be in `op-service`?

op-batcher/flags/flags.go
```diff
@@ -110,8 +114,8 @@ var Flags []cli.Flag
 
 func CheckRequired(ctx *cli.Context) error {
 	for _, f := range requiredFlags {
-		if !ctx.GlobalIsSet(f.GetName()) {
-			return fmt.Errorf("flag %s is required", f.GetName())
+		if !ctx.IsSet(f.Names()[0]) {
```
tynes, line 117: Could this out of bounds?

tynes (commented):
Generally looks good to me

op-batcher/compressor/cli.go
protolambda: there already is `opservice.PrefixEnvVar`. We can make that output a string-slice, but we'd still need to bind it to `envPrefix`, so it doesn't get any shorter, or we have to repeat `envPrefix` for every call.

op-batcher/flags/flags.go
protolambda: There's always at least 1 name in a flag

op-service/log/cli.go
protolambda: will remove `ReadLocalCLIConfig`, there are not many usages of it, and it shouldn't make a difference anymore.

op-node/flags/flags.go
```diff
@@ -258,8 +258,8 @@ func init() {
 
 func CheckRequired(ctx *cli.Context) error {
 	for _, f := range requiredFlags {
-		if !ctx.GlobalIsSet(f.GetName()) {
-			return fmt.Errorf("flag %s is required", f.GetName())
+		if ctx.IsSet(f.Names()[0]) {
```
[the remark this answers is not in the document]
protolambda: ty, nice catch

op-batcher/compressor/cli.go
trianglesphere: we didn't have a problem repeating `envPrefix`. IMO `opservice.PrefixEnvVar` should return a string slice now

protolambda: And rename it to?

protolambda: Also I don't like repeating envPrefix, seems more error prone :/

protolambda: updated to return in a slice

trianglesphere (approved):

PR #6096, opened 2023-06-21
fix(ctb): Triple Slash Natspec Port

refcell:
**Description**

Following the decision to migrate to triple slash natspec comment styling,
this pr is the base of a port stack.
Specifically, this pr ports all `contracts-bedrock/contracts/L1/*` contracts
to use the triple slash natspec format, bumping contract semver patches where
necessary.

PR #6099, opened 2023-06-21
fix(ctb): Port L2 Contracts to Triple Slash Natspec Styling

refcell:
**Description**

Ports all `contracts-bedrock/contracts/L2/*` contracts and their related test
to use the triple slash natspec styling.


refcell:
* develop
  * **PR #6096** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6096" target="_blank">Graphite</a> 
    * **PR #6099** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6099" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6099?utm_source=stack-comment).

refcell:
If you love the red @tynes, you'll love this pr :P

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
```diff
@@ -17,7 +17,7 @@ contract L1StandardBridge_Getter_Test is Bridge_Initializer {
         assert(L1Bridge.OTHER_BRIDGE() == L2Bridge);
         assert(L1Bridge.messenger() == L1Messenger);
         assert(L1Bridge.MESSENGER() == L1Messenger);
-        assertEq(L1Bridge.version(), "1.1.0");
+        assertEq(L1Bridge.version(), "1.1.1");
```
tynes, line 20: we could prob delete this assertion, its kind of overhead

tynes:
Looks like this needs a rebase

maurelian:
A thought about how to make it totally obvious that this PR does not change anything in the impl (which might be a common occurrence if we're making style changes) would be to do it in two commits: 

1. A commit which only changes the comments. This commit should pass the build process. 
2. Another that updates the semvers and bindings.

refcell:
Re @maurelian, that makes sense, but I usually tend to squash commits so we don't end up merging without a squash by accident. I think it's on the impetus of the implementor to do as their pr description/title say and the reviewer to verify these are the only changes.

This feels more like a process for PR standards as opposed to the actual commits. More generally - I'm quite opposed to dictating how folks perform commits since it severely limits the flexibility in their development styles.

packages/contracts-bedrock/contracts/test/L1StandardBridge.t.sol
refcell: Split out into #6119

maurelian:
@refcell TBH I think we should/will move towards the impetus being on the author to make the review easy by enabling more things to be automatically verifed. 

Regardless this is easy enough to review so I'm not gonna block this now. Seems like we could take it out of draft and get it in.

tynes (approved):

tynes:
Didn't notice any implementation changes here besides the semver bumps

PR #6101, opened 2023-06-21
op-node: log shutdown before releasing client waitgroup

protolambda:
**Description**

Fix test issues where the test-logger was being used after the test shut down.

The `TestMultiPeerSync` defers the `.Close()` calls on the sync clients, but those then completed the wait-group in the sync client, before logging about the shutdown. This caused a race where the logger would panic because of the unexpected log.

The fix is to move the log before the `wg.Done()`.

Reviewing the code further there also seems to be a case where we could have a warn-log while the client was already shutting down. We now return early in `AddPeer` if we're closing already, to ensure post-shutdown AddPeer calls cannot accidentally create logs that make the test flake.


ajsutton (approved):
LGTM.

PR #6105, opened 2023-06-22
op-node: Persist synchroniser state

ajsutton:
**Description**

Adds a `--rpc.admin-state` CLI option to specify a file to persist config changes made via the RPC Admin APIs to. Currently that's just whether the sequencer is started or stopped.

* The file is only created when the first change is made via RPC
* If the file doesn't exist, the other CLI args determine if the sequencer starts or not (`--sequencer.stopped` in particular)
* If the file does exist and is valid, the persisted value overrides the` --sequencer.stopped` option
* If the file exists and is invalid, op-node will refuse to start
* If writing the file fails, the admin API call will report an error and the sequencer state will be unchanged

The state is persisted as a JSON file so that we can extend it with additional options in the future if need be. The code for persistence has been kept simple and takes advantage of there only being one value currently - it can be refactored to be more complex and merge multiple values in the future if needed.

The file is written with maximal safety - updates are first written to a temp file, fsync'd and then renamed into place.

**Tests**

Unit tests added around the persistence code. Manual testing for the actual CLI interactions. Will investigate cli tests or e2e tests for this in a follow up PR.

**Metadata**

- https://linear.app/optimism/issue/CLI-4094/persistence-for-zdd-flags


op-node/node/config_persistence.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("create config dir (%v): %w", p.file, err)
+	}
+	// Write the new content to a temp file first, then rename into place
+	// Avoids corrupting the content if the disk is full or there are IO errors
+	tmpFile := p.file + ".tmp"
+	file, err := os.OpenFile(tmpFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return fmt.Errorf("open file (%v) for writing: %w", tmpFile, err)
+	}
+	defer file.Close()
```
protolambda, line 72: Maybe we should close the file before renaming it?

op-node/node/config_persistence.go
```diff
@@ -0,0 +1,134 @@
...
+		return StateStopped, nil
+	}
+}
+
+func (p *ActiveConfigPersistence) read() (persistedState, error) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	data, err := os.ReadFile(p.file)
+	if errors.Is(err, os.ErrNotExist) {
+		return persistedState{}, nil
```
protolambda, line 108:
```suggestion
		// persistedState.SequencerStarted == nil: SequencerState() will return StateUnset if no state is found
		return persistedState{}, nil
```
Maybe worth a comment about the `nil` boolean value. It doesn't default to false like someone might expect.

op-node/rollup/driver/state.go
```diff
@@ -334,6 +339,10 @@ func (s *Driver) eventLoop() {
 			} else if !bytes.Equal(unsafeHead[:], resp.hash[:]) {
 				resp.err <- fmt.Errorf("block hash does not match: head %s, received %s", unsafeHead.String(), resp.hash.String())
 			} else {
+				if err := s.sequencerNotifs.SequencerStarted(); err != nil {
+					resp.err <- fmt.Errorf("sequencer start notification: %w", err)
+					continue
+				}
```
protolambda, line 345: I'd suggest always persisting it when the signal is received, even if the `driverConfig.SequencerStopped` state already is set.

op-node/rollup/driver/state.go
```diff
@@ -88,6 +91,8 @@ type Driver struct {
 func (s *Driver) Start() error {
 	s.derivation.Reset()
 
+	log.Info("Starting driver", "sequencerEnabled", s.driverConfig.SequencerEnabled, "sequencerStopped", s.driverConfig.SequencerStopped)
+
```
protolambda, line 95: Maybe we should add a call to persist the sequencer state here, to ensure the file exists, if the sequencer is started through CLI rather than through RPC?

op-node/flags/flags.go
```diff
@@ -59,6 +59,11 @@ var (
 		Usage:   "Enable the admin API (experimental)",
 		EnvVars: prefixEnvVars("RPC_ENABLE_ADMIN"),
 	}
+	RPCAdminPersistence = &cli.StringFlag{
+		Name:    "rpc.admin-state",
+		Usage:   "File path used to persist state changes made via the admin API so they persist across restarts",
+		EnvVars: prefixEnvVars("RPC_ADMIN_STATE"),
+	}
 
 	/* Optional Flags */
```
protolambda, line 68: It's an optional flag and in the optional-flags list, but was grouped with the required-flag definitions here.

op-node/service.go
```diff
@@ -143,14 +148,29 @@ func NewL2SyncEndpointConfig(ctx *cli.Context) *node.L2SyncEndpointConfig {
...
+	return node.NewConfigPersistence(stateFile)
+}
+
+func NewDriverConfig(ctx *cli.Context, config node.ConfigPersistence) (*driver.Config, error) {
+	sequencerStopped := ctx.Bool(flags.SequencerStoppedFlag.Name)
+	if state, err := config.SequencerState(); err != nil {
+		return nil, err
+	} else if state != node.StateUnset {
+		sequencerStopped = state == node.StateStopped
+	}
```
protolambda, line 165: - It may be worth a warn-log when the CLI flag disagrees with the persisted state
- It may be worth an info-log when no persisted state was found

op-node/node/config_persistence.go
```diff
@@ -0,0 +1,134 @@
...
+	} else if err != nil {
+		return persistedState{}, fmt.Errorf("read config file (%v): %w", p.file, err)
+	}
+	var config persistedState
+	if err = json.Unmarshal(data, &config); err != nil {
```
protolambda, line 113: If we want this to be extra strict:
- With `dec := json.NewDecoder()` + `dec.DisallowUnknownFields()`, we can specify it to error in case of unknown keys in the read json file.
- We could also return an error if it loads a json value with a nil `SequencerStarted` field.

op-node/node/config_persistence.go
ajsutton: Yeah that's fair - added. `persistedState` is a bit weird using `*bool` but that's why it's deliberately internal only.

op-node/flags/flags.go
ajsutton: ah, I just put it with `rpc.enable-admin` which is also optional.  I'll move the optional flags comment above both.

op-node/rollup/driver/state.go
ajsutton: hmm, that winds up being a more intrusive change for starting because we need to persist only after validating that the unsafe head is the expected hash, but don't want to report a mismatched head to the user because the sequencer is already running.  So on one hand we want to check the hash matches, then persist state, then check if a change is required and the other we want to check if the change is required, then the hash then persist.

Given that we only change the state if the persistence works, if we also do the persistence on startup as you suggest they'll always be in sync anyway and we shouldn't need to do this.  And if they do somehow get out of sync somehow we can manually edit the file since it's never cached in memory.

op-node/rollup/driver/state.go
ajsutton: That's a good idea - it also means that startup fails if we don't have permissions to write the file rather than later discovering that we can't use the admin RPC functions.

op-node/node/config_persistence.go
ajsutton: We may need to loosen up a bit again if we wind up persisting other values to this file but makes sense to start really strict.

op-node/flags/flags.go
```diff
@@ -54,13 +54,18 @@ var (
...
 	RPCEnableAdmin = &cli.BoolFlag{
 		Name:    "rpc.enable-admin",
 		Usage:   "Enable the admin API (experimental)",
 		EnvVars: prefixEnvVars("RPC_ENABLE_ADMIN"),
 	}
-
-	/* Optional Flags */
+	RPCAdminPersistence = &cli.StringFlag{
+		Name:    "rpc.admin-state",
+		Usage:   "File path used to persist state changes made via the admin API so they persist across restarts",
```
keroro520, line 66: It would be better to note that "disable by default".

op-node/node/config_persistence.go
```diff
@@ -0,0 +1,143 @@
...
+	"sync"
+)
+
+type RunningState int
+
+const (
+	StateUnset RunningState = iota
+	StateStarted
+	StateStopped
+)
```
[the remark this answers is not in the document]
keroro520: BTW, what about replace "unset" with "SequencerDisabled"?

op-node/node/config_persistence.go
```diff
@@ -0,0 +1,143 @@
...
+
+type persistedState struct {
+	SequencerStarted *bool `json:"sequencerStarted,omitempty"`
+}
+
+type ConfigPersistence interface {
+	SequencerStarted() error
+	SequencerStopped() error
```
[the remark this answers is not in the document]
ajsutton: The new names suggest that the calls perform an action to start or stop the sequencer which is incorrect.  They are notifications that the sequencer has been started or stopped so the listener can react appropriately.

op-node/node/config_persistence.go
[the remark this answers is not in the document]
ajsutton: I'm not sure I see the advantage here given the code always uses the constants anyway.  It's pretty typically to define an enum like this in go via the `iota` pattern.

In terms of the renaming, `Unset` doesn't mean the sequencer is disabled - it means no state has been persisted, the stored state has not yet been set.

op-node/node/config_persistence.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (p *ActiveConfigPersistence) SequencerStopped() error {
+	return p.persist(false)
+}
+
+// persist writes the new config state to the file as safely as possible.
+// It uses sync to ensure the data is actually persisted to disk and initially writes to a temp file
+// before renaming it into place. On UNIX systems this rename is typically atomic, ensuring the
+// actual file isn't corrupted if IO errors occur during writing.
+func (p *ActiveConfigPersistence) persist(sequencerStarted bool) error {
```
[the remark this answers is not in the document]
ajsutton: Not a bad idea, but in this case I think it has some downsides.
I'm deliberately avoiding having `persistedState` as part of the public interface so the `SequencerState` method can't return a `persistedState` as suggested in the proposed `SequencerStarted` code.  That would mean creating a new `persistedState` in `SequencerStarted` which is problematic if additional settings are persisted in the future because it would silently removed them from the persisted version.  By having the sequencer state be passed as a boolean to `persist` we avoid that and ensure that if a field is added in the future we'll get a compile error to force us to update the code.  There are a few ways we might resolve that in the future depending on what the additional setting is, how frequently it changes and whether it changes at the same time as `SequencerStarted`.

op-node/node/config_persistence.go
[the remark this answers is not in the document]
keroro520: Thanks for your explanation. Now I understand that `ConfigPersistence` is supposed to act as a listener and implement `SequencerStateListener`.

Then what about changing to `OnStartSequencer`/`OnStopSequencer`, typically event-driven naming style.

op-node/node/config_persistence.go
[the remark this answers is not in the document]
keroro520: Oh, I just wanted to make the persisted file easier to read. When users open the persisted file, they may be confused about what these integers indicates?

op-node/node/config_persistence.go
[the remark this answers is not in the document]
keroro520: Feel free to resolve this comment. @ajsutton

protolambda (approved):

PR #6108, opened 2023-06-22
op-challenger API Cleanups

trianglesphere:
**Description**

Simplify the API by using uint64 as the trace index everywhere & return a new position object instead of mutating it when attacking or defending.


ajsutton (approved):
Oh yeah I'm a sucker for a good immutable object. :)

refcell (approved):
Very nice

PR #6109, opened 2023-06-22
feat(op-challenger): Game State Interface

refcell:
**Description**

Creates a `Game` state interface that will be used by an `op-challenger` `Agent` to interact with the game state.

### DIFF
--- a/op-challenger/fault/game.go
+++ b/op-challenger/fault/game.go
@@ -1,0 +1,13 @@
+package fault
+
+// Game is an interface that represents the state of a dispute game.
+type Game interface {
+	// Put adds a claim into the game state and returns its parent claim.
+	Put(claim Claim) (Claim, error)
+
+	// ClaimPairs returns a list of claim pairs.
+	ClaimPairs() []struct {
+		claim  Claim
+		parent Claim
+	}
+}


op-challenger/fault/game.go
```diff
@@ -0,0 +1,13 @@
+package fault
+
+// Game is an interface that represents the state of a dispute game.
```
trianglesphere, line 3: If all we have is the root claim & get parent, how do we iterate the state?

`Put` is good, but instead of `GetParent` + `RootClaim` we need `ClaimParis() []{claim: Claim, parent: Claim}`

trianglesphere (approved):
interface LGTM

op-challenger/fault/game.go
```diff
@@ -0,0 +1,13 @@
...
+type Game interface {
+	// Put adds a claim into the game state and returns its parent claim.
+	Put(claim Claim) (Claim, error)
+
+	// ClaimPairs returns a list of claim pairs.
+	ClaimParis() []struct {
```
tynes, line 9: Pairs not Paris

tynes (changes requested):
typo

PR #6113, opened 2023-06-23
feat(op-challenger): Game State

refcell:
**Description**

Implements the `Game` state which an `Agent` will use to store `Claim`s and track their parents.

The `gameState` struct adheres to a slightly modified `Game` interface with the addition of a `GetClaimParent(claim Claim) Claim` function that is used to retrieve the parent of a specific claim.

This is exposed externally since it is used in testing and may prove useful for the `Agent` to query the parent of a specific claim without needed to query for all `[Claim, Parent]` pairs and iterate the list.

**Metadata**

Fixes CLI-4116


op-challenger/fault/game.go
```diff
@@ -10,4 +19,65 @@ type Game interface {
...
+// gameState is a struct that implements the [Game] interface.
+type gameState struct {
+	// claims is a map of claim gindex to
+	claims map[uint64]Claim
+}
+
+// NewGameState returns a new game state.
+func NewGameState() *gameState {
+	return &gameState{
+		claims: make(map[uint64]Claim),
```
trianglesphere, line 36: There are allowed to be multiple claims at a specific gindex. The claim index + value is what must be unique (& claim index really means attacking/defending a specific node). I'd probably follow the contract representation a little more & have have claims relate to each other with custom metadata. It's fine to make a `node` struct & use it here.

op-challenger/fault/game.go
```diff
@@ -10,4 +19,65 @@ type Game interface {
...
+func (g *gameState) GetClaimParent(claim Claim) Claim {
+	// If the claim is the root, we don't need to return a parent.
+	parent := Claim{}
+	if claim.Depth() != 0 {
+		parent = g.claims[claim.GetParentGIndex()]
+	}
+	return parent
+}
+
+// Put adds a claim into the game state and returns its parent claim.
```
trianglesphere, line 49: Why do we return the parent claim here?

op-challenger/fault/position_test.go
```diff
@@ -95,3 +96,14 @@ func TestTraceIndex(t *testing.T) {
 		require.Equal(t, test.TraceIndex, result)
 	}
 }
+
+// TestPosition_GetParentGIndex_Tabled tests the GetParentGIndex function on the treeNodesMaxDepth4 data.
+// The parent index of a node is the floor of the node's index divided by 2 (since the first index is 1).
+func TestPosition_GetParentGIndex_Tabled(t *testing.T) {
```
trianglesphere, line 102: no need for the `_Tabled` comment. TBH parent gindex should be added to the table & we should manually assert that specific values are correct.

op-challenger/fault/position.go
```diff
@@ -49,6 +49,14 @@ func boolToInt(b bool) int {
 	}
 }
 
+// GetParentGIndex returns the gindex of the parent.
+func (p *Position) GetParentGIndex() uint64 {
```
trianglesphere, line 53: ParentGIndex is not the same as the gindex of the node that we are countering when we defend (only when we attack).

trianglesphere (commented):
Likely a follow on, but we will need the ability to check if a claim is a duplicate.

Looking at this more closely, I don't think that the generalized index is a good key to work with. I'd probably just build a tree with the pointers going from parent to child (opposite how it's laid out in the contract storage).

refcell:
Moving to draft to implement a tree like structure rather than gindex reliant component

refcell:
Closing in favor of #6161 6161

PR #6117, opened 2023-06-23
fault-detector: replace OutputProposed log cache with direct call to getL2Output.

zchn:
Previously, we query L2OutputOracle's OutputProposed to build a mapping between batch index and the output proposal. This has 
two problems:

1. The log will get bigger over time, for example, on Goerli we are already querying more than 10k logs at the same time which 
hits Alchemy RPC's limit.
2. It doesn't work when output proposals are deleted using deleteL2Output, because the current version does not process 
OutputsDeleted events.

This PR replaces the log based mapping with a direct call to L2OutputOracle.getL2Output, because there is already a mapping 
built in the contract.


hamdiallam (approved):
This looks great! Thanks for doing this!

PR #6118, opened 2023-06-23
POC: turning lib clock into a struct

maurelian:
Pros:
- Deletes two files
- Solidity structs don't require testing
- Uses a standard field access syntax (`c.duration` rather than `c.duration()`)
- Removes some of the wrap/unwrap boiler plate

Cons:
- Less gas efficient
- Ugly under the hood to use memory for what should be simple stack values

I haven't looked closely at the other custom types we've implemented, but I wouldn't be shocked if we can save a few hundred more LOC.


maurelian:
Closing. The POC served its purpose.

PR #6119, opened 2023-06-23
fix(ctb): Remove Semver Check L1StandardBridge.t.sol

refcell:
**Description**

Following @tynes [comment](https://github.com/ethereum-optimism/optimism/pull/6099#discussion_r1239095697), this pr splits out the change to remove the semver check on the `L1StandardBridge`.


maurelian (approved):
But, but... how will we know that the number is the number?

refcell:
> But, but... how will we know that the number is the number?

Let's discuss a better approach offline with @tynes :)

PR #6120, opened 2023-06-23
feat: migrate from yarn to pnpm

roninjin10:
Migrates package manager to pnpm
- lots of documentation updated
- lots of scripts updated
- sometimes I remove the cache completely because pnpm is much faster and wasn't worth optimizing for cache like in case of github actions
- prefer to use npx or let package runner infer the script runner when possible
- pnpm does not allow hoisted node modules. Must be explicit about ALL deps for all packages
- pnpm doesn't run prefoo scripts. WE depend on them so we override this default
- Some packages didn't exist in yarn workspaces before. I added them to pnpm workspaces though
- rari-capital/solmate redirecting to transmissions11/solmate caused pnpm issues so made it use t11 instead
- ds-test was using a commit 1 commit before @smartcontracts added a package.json. this made pnpm angry so I bumped commit one commit to @smartcontracts commit
- had to add prettier-plugin-solidity to prettier config
- pnpm to reduce duplication and be space efficient puts all packages in the pnpm store at the root of the repo and symlinks them out to create the node_modules structure.  Other package managers just copy the same files over and over again.  Becasue of this we need to update foundry to allow these files

Pnpm saves space and time of installing node modules. 
It installs node modules into a pnpm store and then symlinks them in the places npm would copy them to. 
By default, it does not hoist node modules
Hoisting is something npm and yarn do to save time/space that pnpm avoids needing to do
Hoisting makes it so dependencies must always be explicit.  With yarn and npm contracts-bedrock at any time can import a dependency only declared in sdk.  With pnpm this is impossible.  More robust code
Pnpm in docker is better because you can run pnpm fetch after only COPY pnpm-lock.yaml. Pnpm will then[ install node modules](https://pnpm.io/cli/fetch) with only the lockfile for maximum cache hits.
[Pnpm has a robust escape hatch](https://pnpm.io/pnpmfile) that allows you to modify package.json of dependent packages that comes in handy compared to trying to bang head against the wall with yarn resolutions
Pnpm commands are the same as npm but with a p in front for the most part


roninjin10:
* develop
  * **PR #6120** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6120" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6120?utm_source=stack-comment).

tynes:
should this base into the pnpm pr?

tynes:
@roninjin10 i pushed a couple of commits, some things are still failing but otherwise its mostly working, idk how to fix some of this stuff

roninjin10:
Nice I think we can get this ready by monday I'm pretty motivated this weekend to ship this and some other things I'm working on

tynes:
We should squash this and remove the changes to `op-bindings/ast` and then this should be ready for review/merge

Will fix the bindings stuff independently

roninjin10:
Changes made on last rebase were big enough that this may break and will require another self review which is why I'm keeping new commits seperate until they are ready

tynes (approved):

hamdiallam (approved):

package.json
```diff
@@ -32,15 +35,14 @@
     "lint:ts:check": "npx nx run-many --target=lint:ts:check",
     "lint:check": "npx nx run-many --target=lint:check",
     "lint:fix": "npx nx run-many --target=lint:fix",
-    "lint:specs:fix": "yarn run markdownlint-cli2-fix \"./specs/**/*.md\"",
-    "lint:specs:check": "yarn run markdownlint-cli2  \"./specs/**/*.md\"",
-    "lint:specs:toc": "yarn run doctoc '--title=**Table of Contents**' ./specs",
+    "lint:specs:fix": "npx markdownlint-cli2-fix \"./specs/**/*.md\"",
+    "lint:specs:check": "npx markdownlint-cli2  \"./specs/**/*.md\"",
+    "lint:specs:toc": "npx doctoc '--title=**Table of Contents**' ./specs",
```
refcell, line 40: Since we're using `pnpm`, we should migrate away from `npx` and use `pnpm dlx` instead.
See: https://pnpm.io/cli/dlx

clabby (approved):
Double checked docker image builds and scanned through the changes; LGTM! Thanks for getting this over the line @roninjin10 @tynes 🚀

packages/contracts-bedrock/package.json
```diff
@@ -48,17 +48,17 @@
     "@eth-optimism/core-utils": "^0.12.1",
     "@openzeppelin/contracts": "4.7.3",
     "@openzeppelin/contracts-upgradeable": "4.7.3",
-    "@rari-capital/solmate": "https://github.com/rari-capital/solmate.git#8f9b23f8838670afda0fd8983f2c41e8037ae6bc",
-    "clones-with-immutable-args": "https://github.com/Saw-mon-and-Natalie/clones-with-immutable-args.git#105efee1b9127ed7f6fedf139e1fc796ce8791f2",
+    "@rari-capital/solmate": "github:transmissions11/solmate#8f9b23f8838670afda0fd8983f2c41e8037ae6bc",
```
refcell, line 51: Finally, thank you for this 🙏

packages/contracts-periphery/package.json
```diff
@@ -60,20 +60,18 @@
     "@nomiclabs/hardhat-ethers": "^2.0.2",
     "@nomiclabs/hardhat-etherscan": "^3.0.3",
     "@nomiclabs/hardhat-waffle": "^2.0.3",
-    "@rari-capital/solmate": "7.0.0-alpha.3",
     "@openzeppelin/contracts": "4.7.3",
     "@openzeppelin/contracts-upgradeable": "4.7.3",
-    "@types/node": "^17.0.21",
+    "@rari-capital/solmate": "7.0.0-alpha.3",
```
refcell, line 65: Can we also change this to `transmissions11`
```suggestion
    "@transmissions11/solmate": "7.0.0-alpha.3",
```

roninjin10: hmm on latest it's just called `solmate` https://github.com/transmissions11/solmate/blob/main/package.json

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "author": "Optimism PBC",
   "license": "MIT",
+  "private": true,
```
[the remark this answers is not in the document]
roninjin10: Don't remember if this broke something but the root package.json in a monorepo should always be private

package.json
roninjin10: Oh this is dope because it would save space on our machine compared to npx.  I'll actually do this in it's own pr.  TIL

.circleci/config.yml
```diff
@@ -367,41 +379,49 @@ jobs:
...
       - run:
           name: lint
+          command: pnpm lint:check || echo "export LINT_STATUS=1" >> "$BASH_ENV"
+          working_directory: packages/contracts-bedrock
+      - run:
+          name: slither
           command: |
-            yarn lint:check || echo "export LINT_STATUS=1" >> "$BASH_ENV"
+            pnpm lint:check || echo "export LINT_STATUS=1" >> "$BASH_ENV"
           working_directory: packages/contracts-bedrock
```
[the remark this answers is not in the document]
roninjin10: Good catch this looks like a bad rebase caused a duplicate to happen

.circleci/config.yml
```diff
@@ -367,41 +379,49 @@ jobs:
...
       - check-changed:
           patterns: contracts-bedrock,hardhat-deploy-config
+      - run:
+          name: Install pnpm package manager
+          command: |
+            npm i pnpm --global
       - run:
           name: lint
+          command: pnpm lint:check || echo "export LINT_STATUS=1" >> "$BASH_ENV"
+          working_directory: packages/contracts-bedrock
```
roninjin10, line 394:
```suggestion
```
Remove duplicate lint step

tynes:
Not sure what they deal with netlify is here

roninjin10:
@tynes I know how to fix netlify from the gateway version of the pr

roninjin10:
ugghhh netlify not sure looking into it

roninjin10:
Ok netlify should be good now had to change some settings on netlify

zhwrd (approved):

roninjin10:
I want to see this go green, then squash, then merge

PR #6121, opened 2023-06-23
feat(ctb): Emit tx hash in wallet-mon detection logs

maurelian:
**Description**

Adds the tx hash to the log emitted if an errant call is detected.

### DIFF
--- a/packages/chain-mon/src/wallet-mon/service.ts
+++ b/packages/chain-mon/src/wallet-mon/service.ts
@@ -7,7 +7,7 @@ import {
   waitForProvider,
 } from '@eth-optimism/common-ts'
 import { getChainId, compareAddrs } from '@eth-optimism/core-utils'
-import { Provider } from '@ethersproject/abstract-provider'
+import { Provider, TransactionResponse } from '@ethersproject/abstract-provider'
 import mainnetConfig from '@eth-optimism/contracts-bedrock/deploy-config/mainnet.json'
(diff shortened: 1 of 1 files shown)


tynes:
The way that this logic works, it cannot work with all possible contracts, if somebody is able to cause a call down the call stack to call something else, this will not catch it, since it only looks at the top level `tx.to`

maurelian:
> The way that this logic works, it cannot work with all possible contracts, if somebody is able to cause a call down the call stack to call something else, this will not catch it, since it only looks at the top level `tx.to`

Ack, but the intent here is to detect unauthorized access to hot wallet keys.

When we have contracts where we might need to monitor for such a thing, I hope that we'll add some events in order to facilitate it.

tynes:
> > The way that this logic works, it cannot work with all possible contracts, if somebody is able to cause a call down the call stack to call something else, this will not catch it, since it only looks at the top level `tx.to`
> 
> Ack, but the intent here is to detect unauthorized access to hot wallet keys.
> 
> When we have contracts where we might need to monitor for such a thing, I hope that we'll add some events in order to facilitate it.

Feels like we should rely on call tracing for this rather than events

maurelian:
I still want this, it would have been useful in debugging a recent issue. 
This service is currently intended to support hot wallets (EOAs), and has no need to look through the call trace.

sebastianst:
merging after discussion with @maurelian

PR #6122, opened 2023-06-23
feat(ctb): Introduce `Semver` on `FaultDisputeGame` + `DisputeGameFactory`

clabby:
## Overview

Removes the hardcoded `VERSION` constant from the dispute contracts in favor of `Semver`, which is the standard in the rest of the contracts codebase.


op-bindings/bindings/disputegamefactory.go
maurelian, line 1: Should there be a binding for the `FaultDisputeGame` itself?

maurelian (approved):
LGTM pending resolution of one comment.

op-bindings/bindings/disputegamefactory.go
clabby: Yes there should - we haven't added it just yet. Can do in this PR or a separate one.

maurelian: Separate is fine.

clabby:
@mergifyio queue

PR #6123, opened 2023-06-23
wip: geth cli

tynes:
**Description**

Right now there are 2 codepaths for deploying the L1 contracts. There is a "predeployed" setup where go code deploys and sets up each smart contract and then there is the "deployed" setup that is used in production. We do not want to maintain both setups for deploying contracts.

The solution is to set up an instance of geth, use the foundry deploy script and then call the debug endpoint to dump the state as a JSON file and then use that as input to op-e2e to set up the L1 contracts. All of the codepath for predeploying the L1 contracts can be removed and then the exact state produced by the deploy script that is used in production will be used by op-e2e

This PR refactors some functionality out of op-e2e into op-chain-ops so that it can be reused in a geth cli meant to build a genesis.json file for op-e2e


op-chain-ops/cmd/geth/main.go
```diff
@@ -0,0 +1,149 @@
...
+			c := make(chan any)
+
+			rpcCfg := oprpc.ReadCLIConfig(ctx)
+			server := oprpc.NewServer(
+				rpcCfg.ListenAddr,
+				rpcCfg.ListenPort,
+				version,
+				oprpc.WithLogger(log.New("rpc_server")),
+			)
+			server.AddAPI(gethrpc.API{
```
tynes, line 108: The idea here was to add a new rpc endpoint `admin_stop` so that we could remotely shut down this script. Open to better ideas but the flow is:
- run this command
- forge deployment script
- dump state
- call rpc to shut down this binary
- move state to location where op-e2e can pull it in

PR #6124, opened 2023-06-23
feat(op-challenger): Tree-like Game State

refcell:
**Description**

Implements the `Game` state which an `Agent` will use to store `Claim`s and track their parents. The `gameState` struct adheres to the `Game` interface.

As opposed to #6113, this PR uses a "tree-like" structure whereby nodes directly contains a list to their children.

This PR also slightly modifies the `Claim` struct as discussed offline to contain a `Parent` reference and a `DefendParent` boolean.

**Metadata**

Fixes CLI-4116

### DIFF
--- a/op-challenger/fault/game.go
+++ b/op-challenger/fault/game.go
@@ -1,13 +1,151 @@
 package fault
 
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var (
+	// ErrClaimExists is returned when a claim already exists in the game state.
+	ErrClaimExists = errors.New("claim exists in game state")
+
+	// ErrClaimNotFound is returned when a claim does not exist in the game state.
+	ErrClaimNotFound = errors.New("claim not found in game state")
+)
+
 // Game is an interface that represents the state of a dispute game.
 type Game interface {
-	// Put adds a claim into the game state and returns its parent claim.
-	Put(claim Claim) (Claim, error)
+	// Put adds a claim into the game state.
+	Put(claim Claim) error
 
 	// ClaimPairs returns a list of claim pairs.
 	ClaimPairs() []struct {
 		claim  Claim
 		parent Claim
 	}
+}
+
+// Node is a node in the game state tree.
+type Node struct {
+	self     Claim
+	children []Node
+}
+
+// NodeKey is the key for a given list of [Claim] children.
+type NodeKey struct {
+	// Position is embedded to pass through its methods.
+	Position
+	// Value is the claim hash.
+	Value common.Hash
+}
+
+// FromClaim creates a [NodeKey] from a given [Claim].
+func FromClaim(claim *Claim) NodeKey {
+	return NodeKey{
+		Position: claim.Position,
+		Value:    claim.Value,
+	}
+}
+
+// gameState is a struct that represents the state of a dispute game.
+// The game state implements the [Game] interface.
+type gameState struct {
+	nodes map[NodeKey]Node
+}
+
+// NewGameState returns a new game state.
+func NewGameState() *gameState {
+	return &gameState{
+		nodes: make(map[NodeKey]Node),
+	}
+}
+
+// getParent returns the parent of a given [Claim].
+func (g *gameState) getParent(claim Claim) (Claim, error) {
+	// Create a new NodeKey from the claim.
+	key := FromClaim(&claim)
+
+	// Get the node from the game state.
+	node, ok := g.nodes[key]
+	if !ok {
+		return Claim{}, ErrClaimNotFound
+	}
+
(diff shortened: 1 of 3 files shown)


refcell:
Throwing a `mergify-ignore` on here until @trianglesphere signs off on the interface since he's tackling the agent and orchestrator

op-challenger/fault/game.go
```diff
@@ -1,13 +1,151 @@
...
+	return NodeKey{
+		Position: claim.Position,
+		Value:    claim.Value,
+	}
+}
+
+// gameState is a struct that represents the state of a dispute game.
+// The game state implements the [Game] interface.
+type gameState struct {
+	nodes map[NodeKey]Node
```
trianglesphere, line 54: Why do we have a map of nodes? The root node is all we need in the game state to be able to find every other node.

refcell: I think it allows for simple, easy access to nodes and their parents.

A root node _should_ allow us to find every other node **iff** the agent gracefully decides not confuse the state by first posting an intermediary node.

I think this is a straightforward approach, and don't see why we'd limit ourselves by only containing the root node when storage isn't the limiting factor here. Lookup speed is far more important and is O(1) with the mapping. With a rootnode it's Θ(d/2)

refcell:
re: already signed off on - meant to make sure we shouldn't change it up at all and can leave as is. Just wanted to give us the flexibility to change it up if need be... this is an mvp after all.

op-challenger/fault/game.go
trianglesphere: > A root node should allow us to find every other node iff the agent gracefully decides not confuse the state by first posting an intermediary node.

This isn't something that happens - the contract requires that claims point to a parent claim (the root claim is initialized at contract creation time).

op-challenger/fault/game_test.go
```diff
@@ -0,0 +1,121 @@
...
+func createTestClaims() []Claim {
+	top := Claim{
+		Value:         common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000768"),
+		Position:      NewPosition(0, 0),
+		Parent:        nil,
+		DefendsParent: false,
+	}
+
+	middle := Claim{
+		Value:         common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000768"),
```
trianglesphere, line 19: shouldn't be the same as the top claim.

op-challenger/fault/game_test.go
```diff
@@ -0,0 +1,121 @@
...
+	g := NewGameState()
+
+	// Create test claims
+	claims := createTestClaims()
+	top := claims[0]
+	middle := claims[1]
+	bottom := claims[2]
+
+	// Put the middle claim into the game state.
+	// We should expect no parent to exist, yet.
```
trianglesphere, line 68: This doesn't make sense - we should always be putting in the parent first.

op-challenger/fault/game_test.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, claim := range claims {
+		err := g.Put(claim)
+		require.NoError(t, err)
+	}
+
+	// Get the list of claim pairs.
+	pairs := g.ClaimPairs()
+
+	// Validate the pairs parents are correct.
+	for _, pair := range pairs {
```
trianglesphere, line 118: Can create a list of expected outputs (i.e. each pair) & assert that the contents of the  returned list is the same as the expected output. You can https://pkg.go.dev/github.com/stretchr/testify/require#ElementsMatch

op-challenger/fault/game.go
```diff
@@ -1,13 +1,151 @@
...
+}
+
+// NewGameState returns a new game state.
+func NewGameState() *gameState {
```
trianglesphere, line 58: Games should probably be initialized with the root claim for the time being

op-challenger/fault/game.go
refcell: Sure, it's still directly limiting though and provides zero benefits.

op-challenger/fault/game_test.go
refcell: If we don't allow this, then there's this strange implicit guarantee that the game state will be constructed walking down from the root.

If that's the way we want to go, we should make this guarantee explicit and then require that the game state be constructed that way. This would probably change the api/interface of the game state though with the agent (especially it's construction).

trianglesphere: On the agent side I doubt it will change much. I guess it wasn't explicit, but I am fine saying that parents must be added before their children. If we iterate the state in the L1 contract we have this property

refcell:
Closing in favor of #6161

PR #6133, opened 2023-06-24
contracts-bedrock: delete dead ci check

tynes:
**Description**

This hh task only queried the op-node and logged its config. If it was useful to print the config, it would be easy to do something like `cast rpc` instead of needing to have a hardhat task for this. No assertions are made against the return values, they were just printed.


refcell (approved):
Very nice

PR #6134, opened 2023-06-24
contracts-bedrock: optimize storage layout checks

tynes:
**Description**

Optimize the storage layout checks by removing a dep on hardhat. Instead of reading massive build-info files from disk, simply read the forge artifacts directly and then filter through the interesting ones.
This now only takes a few seconds to run when the compilation is cached.

Also delete a dead hh task for doing deposits, this was built as a PoC to trigger a deposit.


refcell (approved):
LGTM - just needs a lint fix

packages/contracts-bedrock/scripts/validate-spacers.ts
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Returns true if the contract should be skipped when inspecting its storage layout.
+ * This is useful for abstract contracts that are meant to be inherited.
+ * The two particular targets are:
+ * - CrossDomainMessengerLegacySpacer0
+ * - CrossDomainMessengerLegacySpacer1
+ */
+const skipped = (contractName: string): boolean => {
+  return contractName.includes('CrossDomainMessengerLegacySpacer')
```
clabby, line 21: Not necessary, but would be nice to have an array of regexes up at the top of the file where this function just checks if the any patterns within it match `contractName`.

clabby (approved):
LGTM 👑

Needs lint ^

packages/contracts-bedrock/scripts/validate-spacers.ts
tynes: This is true, i was just lazy with this using higher level apis. We could do this in the future if we need to update this contract, also i'd prefer to have this check somehow built into foundry longer term

PR #6135, opened 2023-06-25
fix(op-node): save remaining batches when BatchUndecided occurs

keroro520:
**Description**

When L1 information is insufficient to decide whether these batches are valid, we save them and wait for more information.

### DIFF
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -194,7 +194,9 @@ batchLoop:
 			remaining = append(remaining, candidates[i+1:]...)
 			break batchLoop
 		case BatchUndecided:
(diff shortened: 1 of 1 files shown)


keroro520:
ping @protolambda @trianglesphere

protolambda:
We are looking into this, but suspect it's dead / non-functional code, due to the way it interacts with the outside Batch Queue, even though it can be reached in the unit-test case of the batch-filtering in isolation.

keroro520:
> We are looking into this, but suspect it's dead / non-functional code, due to the way it interacts with the outside Batch Queue, even though it can be reached in the unit-test case of the batch-filtering in isolation.

Great
Should I close this PR?

PR #6139, opened 2023-06-25
feat(devnet): Set implementation of `FaultDisputeGame` in the factory

clabby:
## Overview

Deploys an alphabet dispute game and sets the `GameType.FAULT` implementation to it in the `DisputeGameFactory` on the local devnet. Used to test the [`op-challenger-rs`](https://github.com/clabby/op-challenger) work this weekend, but will also be useful in testing the reference implementation.


packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -444,11 +450,21 @@ contract Deploy is Deployer {
                 _implementation: disputeGameFactory,
                 _data: abi.encodeCall(
                     DisputeGameFactory.initialize,
-                    (cfg.finalSystemOwner())
+                    (tx.origin)
```
tynes, line 453: We should be able to use `msg.sender` - other deployment functions use it. Perhaps we want to have a global `internal sender` or something like that and use that variable everywhere since a design goal is to always use the same deployer for simplicity, we dont need to make that change in this PR

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -665,5 +681,16 @@ contract Deploy is Deployer {
             console.log("ProxyAdmin ownership transferred to: %s", finalSystemOwner);
         }
     }
+
+    /// @notice Transfer ownership of the DisputeGameFactory contract to the final system owner
+    function transferDisputeGameFactoryOwnership() broadcast() public {
+        DisputeGameFactory disputeGameFactory = DisputeGameFactory(mustGetAddress("DisputeGameFactoryProxy"));
```
tynes, line 687: This needs to be behind the chain id check

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -444,11 +450,21 @@ contract Deploy is Deployer {
                 _implementation: disputeGameFactory,
                 _data: abi.encodeCall(
                     DisputeGameFactory.initialize,
-                    (cfg.finalSystemOwner())
+                    (tx.origin)
                 )
             });
 
-            string memory version = DisputeGameFactory(disputeGameFactoryProxy).version();
+            DisputeGameFactory factory = DisputeGameFactory(disputeGameFactoryProxy);
```
tynes, line 457: This should be in its own function tbh, if we ever upgrade the dgf we wont need to redeploy the fault vm necessarily

refcell (approved):
Looks good to me 🔥

ops-bedrock/docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
       dockerfile: Dockerfile.l1
     ports:
       - "8545:8545"
+      - "8546:8546"
```
trianglesphere, line 20: Why export the WS ports?

tynes: The rust challenger agent uses it

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -665,5 +673,33 @@ contract Deploy is Deployer {
             console.log("ProxyAdmin ownership transferred to: %s", finalSystemOwner);
         }
     }
+
+    /// @notice Transfer ownership of the DisputeGameFactory contract to the final system owner
+    function transferDisputeGameFactoryOwnership() broadcast() public {
+        if (block.chainid == 900) {
```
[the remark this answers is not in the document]
tynes: We do not want it to deploy to prod networks, this is the easiest way to ensure its devnet only for now. We can add additional networks like internal devnet when we are ready

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -665,5 +673,33 @@ contract Deploy is Deployer {
...
+            }
+        }
+    }
+
+    /// @notice Sets the implementation for the `FAULT` game type in the `DisputeGameFactory`
+    function setFaultGameImplementation() broadcast() public {
+        if (block.chainid == 900) {
+            DisputeGameFactory factory = DisputeGameFactory(mustGetAddress("DisputeGameFactoryProxy"));
+            Claim absolutePrestate = Claim.wrap(bytes32(cfg.faultGameAbsolutePrestate()));
+            IBigStepper faultVm = IBigStepper(new AlphabetVM(absolutePrestate));
```
tynes, line 695: I think this is fine but to be idempotent we want to check that the impl isn't already set before setting it

ops-bedrock/docker-compose.yml
clabby: ^ We won't always depend on it, but for now, we use an event subscription with a fallback to a timer-based check on the `claimData` / `disputeGameList` array lengths.

packages/contracts-bedrock/scripts/Deploy.s.sol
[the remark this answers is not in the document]
clabby: As a side-note - it'd be nice to get a library with named constants for the chainids so that these are more human-readable cc @tynes.

ops-bedrock/docker-compose.yml
trianglesphere: Ok. On the go side we've built a polling layer which will allow HTTP clients to create subscriptions, but I guess we don't have that on the rust side yet.

packages/contracts-bedrock/scripts/Deploy.s.sol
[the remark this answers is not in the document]
refcell: +1 we should follow up this pr with that and refactor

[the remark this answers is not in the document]
tynes: Def open to this as a follow up

tynes (approved):

PR #6140, opened 2023-06-26
op-program: Increase log level to make debugging fpp goerli-verify failures easier

ajsutton:
**Description**

Increase log level for op-program in the verify-goerli job to debug so it logs hints it ignores in offline mode.  Convert key to hex before logging so the output is readable.


op-preimage/oracle.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+
+	"github.com/ethereum/go-ethereum/common"
```
protolambda, line 8: `op-preimage` is used by Cannon example programs which otherwise don't depend on geth. Can we avoid the geth import, and just hex-encode using standard utils instead?

ajsutton: ah, that would explain the cannon failure.  Nice to see that doing its job. Will fix.

trianglesphere (approved):

PR #6142, opened 2023-06-26
feat(specs): Remove duplicate config information

maurelian:
**Description**

Removes a bit of spec content which we were using to specify config values, however this just duplicates information that appears elsewhere, and is not applied consistently in the rest of the specs.


tynes:
Where exactly are those values duplicated?

maurelian:
> Where exactly are those values duplicated?

in mainnet.json, goerli.json etc. 

Also these values are totally outdated, the submission interval is in seconds not blocks, which I think makes a strong case for not including it here to keep it up to date.

trianglesphere:
@maurelian bedrock markdown lint is failing. Please don't merge until that is fixed.

maurelian:
@trianglesphere thanks, fixed. Removing the do-not-merge label while allowing the checks to run.

PR #6145, opened 2023-06-26
feat: golang checkl2 script

tynes:
**Description**

Migrate away from the hh version of check-l2 to a go implementation.
The go implementation is much faster so this will help to reduce time
in CI. It also removes the final dep on hh in the contracts-bedrock
package.


op-chain-ops/util/util.go
```diff
@@ -0,0 +1,175 @@
...
+	L2GethClient *gethclient.Client
+}
+
+// NewClients will create new RPC clients from a CLI context
+func NewClients(ctx *cli.Context) (*Clients, error) {
+	clients := Clients{}
+
+	l1RpcURL := ctx.String("l1-rpc-url")
+	if l1RpcURL != "" {
+		l1Client, err := ethclient.Dial(l1RpcURL)
```
trianglesphere, line 37: This + the following block is a function which creates an ethclient, rpc.Client & gethclient from a RPC url

op-chain-ops/cmd/check-l2/main.go
```diff
@@ -0,0 +1,624 @@
...
+			log.Info("Checking predeploy proxy config")
+			g := new(errgroup.Group)
+
+			// All admin slots should be set
+			count := uint64(2048)
+			for i := uint64(0); i < count; i++ {
+				i := i
+				if i%4 == 0 {
+					log.Info("Checking proxy", "index", i, "total", count)
+					if err := g.Wait(); err != nil {
```
trianglesphere, line 58: why wait every 4 invocations?

op-chain-ops/cmd/check-l2/main.go
```diff
@@ -0,0 +1,624 @@
+package main
+
+import (
+	// "errors"
```
refcell, line 4:
```suggestion
```

op-chain-ops/cmd/check-l2/main.go
```diff
@@ -0,0 +1,624 @@
...
+
+		}
+		return nil
+	})
+
+	if err := g.Wait(); err != nil {
+		return err
+	}
+
+	// foo
```
trianglesphere, line 587: ?

refcell, line 587:
```suggestion
```

op-chain-ops/cmd/check-l2/main.go
```diff
@@ -0,0 +1,624 @@
...
+			return nil
+		})
+
+	}
+
+	//
+	g.Go(func() error {
+		switch p {
+		case predeploys.LegacyMessagePasserAddr:
+			contract, err := bindings.NewLegacyMessagePasser(p, client)
```
trianglesphere, line 140: can have a function to verify each type of contract

trianglesphere (commented):
This is a welcome change, but is missing comments & functions. Some of the parallelism is also applied in weird places to me.

op-chain-ops/cmd/check-l2/main.go
tynes: It hits rate limits if increased

trianglesphere: what rate limit? One in our code or on an external RPC?

op-chain-ops/cmd/check-l2/main.go
```diff
@@ -0,0 +1,624 @@
...
+				if i%4 == 0 {
+					log.Info("Checking proxy", "index", i, "total", count)
+					if err := g.Wait(); err != nil {
+						return err
+					}
+				}
+				g.Go(func() error {
+					return checkPredeploy(clients.L2Client, i)
+				})
+			}
```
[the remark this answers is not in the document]
tynes: Its way faster now so we dont need to worry, CI runs this in 2 seconds

op-chain-ops/cmd/check-l2/main.go
tynes: Running this against alchemy

tynes:
> This is a welcome change, but is missing comments & functions. Some of the parallelism is also applied in weird places to me.

The parallelism is done in this way to enable it to run against rate limiting RPCs. I refactored a bunch from how i was originally doing it so that it would work against alchemy.

In CI it only takes 2 seconds
https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/20559/workflows/00bf0ea9-8f1b-403a-8749-45a1f26c38df/jobs/862919

trianglesphere (approved):
lgtm

refcell (approved):

PR #6147, opened 2023-06-26
fix(docs/op-stack): Changes related to the move from hardhat to foundry

qbzzt:
- Restrict the not for production warning. It's not `cast wallet` that is insecure, just `cast wallet new`.
- Fix numbering in #configure-your-network
- Add warning related to `direnv` installation
- Remove no longer relevant request (to replace <RPC>, which no longer appears in a command line)
- Add creating a deployment directory for getting-started
- Fix the `L2OO_ADDR` table line
- Update file name for bridge

Closing: DEVRL-1020

### DIFF
--- a/docs/op-stack/src/docs/build/getting-started.md
+++ b/docs/op-stack/src/docs/build/getting-started.md
@@ -544,7 +544,7 @@ Once you’ve connected your wallet, you’ll probably notice that you don’t h
 1. Grab the address of the proxy to the L1 standard bridge contract:
 
     ```bash
-    cat deployments/getting-started/Proxy__OVM_L1StandardBridge.json |  jq -r .address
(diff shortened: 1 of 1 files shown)


docs/op-stack/src/docs/build/getting-started.md
```diff
@@ -237,7 +239,14 @@ Once you’ve built both repositories, you’ll need head back to the Optimism M
 
 Once you’ve configured your network, it’s time to deploy the L1 smart contracts necessary for the functionality of the chain.
 
-1. Once you’re ready, deploy the L1 smart contracts:
+1. Create a `getting-started` deployment directory.
+
+   ```bash
+   mkdir deployments/getting-started
```
tynes, line 245: I don't think you should need to do this assuming the version of `foundry` is modern enough

tynes (approved):

PR #6150, opened 2023-06-26
contracts-bedrock: move forge check script to bash

tynes:
**Description**

Makes it much faster, will save a lot of time over many CI runs. The startup time of `ts-node` vs `bash` is just too large for such a simple script. We depend on unix filesystems and bash for other commands.


refcell (approved):
cooking

PR #6148, opened 2023-06-26
contracts-bedrock: delete dead typescript

tynes:
**Description**

This commit deletes a bunch of dead typescript code from the contracts-bedrock package.

This deletes bundling in all of the artifacts into the npm package so it will make the npm package must smaller. It will also reduce the build time by removing some steps from the build.

PR #6151, opened 2023-06-26
chore(ctb): Remove TS Utils

refcell:
**Description**

Removes `contract-defs.ts` utilities and moves them into the sdk.


refcell:
Closing for @tynes to remove in #6148

packages/sdk/src/index.ts
```diff
@@ -3,3 +3,4 @@ export * from './utils'
 export * from './cross-chain-messenger'
 export * from './adapters'
 export * from './l2-provider'
+export * from './contract-defs'
```
roninjin10, line 6: Let's not export it for now and isntead plan on removing it completely in a future pr.  This abstraction is a bad abstraction for a bunch of reasons

- Hardcoded to only work on CJS doesn't work on any project using the more modern and recommended ESM
- Not typesafe
- Not code splittable, biggest tech debt the sdk has for frontend

roninjin10 (commented):
See it closed but submitting review in case it reopens

PR #6154, opened 2023-06-26
op-bindings: cleanup gen script

tynes:
**Description**

Does a little refactor to the gen script to dedupe some code and clean things up.


clabby (approved):
Changes to bindgen script look good; Putting `mergify-ignore` on temporarily while we look for why the preimage oracle source map changed.

op-bindings/gen/main.go
```diff
@@ -97,18 +130,7 @@ func main() {
 
 		deployedSourceMap := ""
 		if _, ok := sourceMapsSet[name]; ok {
-			// directory has .sol extension
```
tynes, line 100: My guess is that this got out of date due to `err != nil` and this is resulting in a diff for the preimage oracle source map

clabby:
@mergifyio queue

PR #6157, opened 2023-06-27
op-node: Remove RPC flags from the required flag list

trianglesphere:
**Description**

These flags are	not actually required to run the op-node. They now default to 0.0.0.0 for the address & 8545 for the port.

### DIFF
--- a/op-node/flags/flags.go
+++ b/op-node/flags/flags.go
@@ -44,18 +44,19 @@ var (
 		Usage:   fmt.Sprintf("Predefined network selection. Available networks: %s", strings.Join(chaincfg.AvailableNetworks(), ", ")),
 		EnvVars: prefixEnvVars("NETWORK"),
 	}
+	/* Optional Flags */
 	RPCListenAddr = &cli.StringFlag{
 		Name:    "rpc.addr",
 		Usage:   "RPC listening address",
 		EnvVars: prefixEnvVars("RPC_ADDR"),
+		Value:   "127.0.0.1",
 	}
 	RPCListenPort = &cli.IntFlag{
 		Name:    "rpc.port",
 		Usage:   "RPC listening port",
 		EnvVars: prefixEnvVars("RPC_PORT"),
+		Value:   9545, // Note: op-service/rpc/cli.go uses 8545 as the default.
 	}
-
-	/* Optional Flags */
 	RPCEnableAdmin = &cli.BoolFlag{
 		Name:    "rpc.enable-admin",
 		Usage:   "Enable the admin API (experimental)",
@@ -153,7 +154,7 @@ var (
 	MetricsAddrFlag = &cli.StringFlag{
 		Name:    "metrics.addr",
 		Usage:   "Metrics listening address",
-		Value:   "0.0.0.0",
+		Value:   "0.0.0.0", // TODO(CLI-4159): Switch to 127.0.0.1
 		EnvVars: prefixEnvVars("METRICS_ADDR"),
 	}
 	MetricsPortFlag = &cli.IntFlag{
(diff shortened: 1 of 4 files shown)


op-node/flags/flags.go
```diff
@@ -44,23 +44,24 @@ var (
 		Usage:   fmt.Sprintf("Predefined network selection. Available networks: %s", strings.Join(chaincfg.AvailableNetworks(), ", ")),
 		EnvVars: prefixEnvVars("NETWORK"),
 	}
+	/* Optional Flags */
 	RPCListenAddr = &cli.StringFlag{
 		Name:    "rpc.addr",
 		Usage:   "RPC listening address",
 		EnvVars: prefixEnvVars("RPC_ADDR"),
+		Value:   "0.0.0.0",
```
ajsutton, line 52: We should not default to exposing the RPC port publicly.  Defaulting to `127.0.0.1` would be the safe choice.

trianglesphere: There's a couple other places we need to update stuff then.

op-node/flags/flags.go
```diff
@@ -44,23 +44,24 @@ var (
...
 		Name:    "rpc.addr",
 		Usage:   "RPC listening address",
 		EnvVars: prefixEnvVars("RPC_ADDR"),
+		Value:   "127.0.0.1",
 	}
 	RPCListenPort = &cli.IntFlag{
 		Name:    "rpc.port",
 		Usage:   "RPC listening port",
 		EnvVars: prefixEnvVars("RPC_PORT"),
+		Value:   8545,
```
ajsutton, line 58: This is the same default value as op-geth will use so will make port conflicts very common.  I'd suggest we pick a different default (e.g. 9545).

op-node/flags/flags.go
```diff
@@ -148,7 +149,7 @@ var (
 	MetricsAddrFlag = &cli.StringFlag{
 		Name:    "metrics.addr",
 		Usage:   "Metrics listening address",
-		Value:   "0.0.0.0",
+		Value:   "127.0.0.1",
```
ajsutton, line 152: I like this change, but it's a breaking change so we'll need to be careful with it and make sure it's very clearly communicated. Probably worth considering flagging it as an upcoming breaking change for a release or two before actually changing it so people can specify it explicitly if they do want it to be publicly available.

op-node/flags/p2p_flags.go
```diff
@@ -96,7 +96,7 @@ var (
 		Name:     "p2p.listen.ip",
 		Usage:    "IP to bind LibP2P and Discv5 to",
 		Required: false,
-		Value:    "0.0.0.0",
+		Value:    "0.0.0.0", // TODO: SHould this be 0.0.0.0? I think so b/c it is public
```
ajsutton, line 99: Yes, p2p port should default to 0.0.0.0 because it's completely useless if it's not publicly available.

ajsutton (commented):
I like the look of this but we need to be really clear about the roll out plan and communication strategy or people will suddenly find metrics/pprof/rpc don't work after upgrading.

op-node/flags/flags.go
trianglesphere: Switched. Annoyingly we default to 8545 in op-service. Maybe we need to change that.

op-node/flags/flags.go
ajsutton: We still seem to have the breaking change to defaults in this PR though. Suspect we need to go back to 0.0.0.0 as the default so we can get this merged.

trianglesphere: nice catch, I forgot to revert the change for these flags

ajsutton (approved):
LGTM.

PR #6159, opened 2023-06-27
cannon: Fix oracle tests for mipsevm

Inphi:
Fix up the oracle.bin test, a tiny program that asserts the behavior of the preimage oracle for quick testing.

This PR also fixes a bug in MIPS.sol where preimage reads for partial data (i.e. < 4 bytes) were not properly shifted before 
the value is written to memory.


protolambda (approved):

Inphi:
@Mergifyio queue

PR #6162, opened 2023-06-27
Add a .foundryrc file for setting the version

maurelian:
**Description**

This is currently a PoC. It adds a `.foundryrc` file, and a new `yarn update:foundry` script in the root package.json.

One thing I'm unsure of is how best to handle this in the ci-builder Dockerfile.


ops/docker/ci-builder/Dockerfile
```diff
@@ -11,12 +11,15 @@ RUN apt-get update && \
     chmod +x ./rustup.sh && \
     ./rustup.sh -y
 
+# move the foundryrc file to the foundry dir
+COPY .foundryrc /opt/foundry
```
maurelian, line 15: This doesn't work, just my best guess at it.

packages/contracts-bedrock/README.md
```diff
@@ -56,11 +56,13 @@ npm install @eth-optimism/contracts-bedrock
...
-   ```shell
-   foundryup -C 3b1129b5bc43ba22a9bcf4e4323c5a9df0023140
-   ```
+1. Install the correct version of foundry:
+
+```shell
+curl -L https://foundry.paradigm.xyz | bash
+yarn update:foundry
+```
+
```
refcell, line 65: 🔥 🔥 🔥

ops/docker/ci-builder/Dockerfile
maurelian: Looks like we could pass in the version using `docker build --build-arg foundryrc=$(cat .foundryrc)`.

packages/contracts-bedrock/README.md
refcell: Maybe we can just add `curl -L https://foundry.paradigm.xyz | bash` as a `yarn install:foundry` target as well?

refcell: Makes this really idiomatic then with a `yarn install:foundry && yarn update:foundry`

refcell: or even have `yarn install:foundry` target `yarn update:foundry` as well so you can just run a `yarn install:foundry`

ops/docker/ci-builder/Dockerfile
refcell: Probably worth setting this cmd as a `Makefile` target then so we don't need to think about it

packages/contracts-bedrock/README.md
```diff
@@ -56,11 +56,13 @@ npm install @eth-optimism/contracts-bedrock
...
-1. Install Foundry by following [the instructions located here](https://getfoundry.sh/).
-   A specific version must be used.
-   ```shell
-   foundryup -C 3b1129b5bc43ba22a9bcf4e4323c5a9df0023140
-   ```
+1. Install the correct version of foundry (defined in the .foundryrc file in the root of this repo.
+
+```shell
+curl -L https://foundry.paradigm.xyz | bash
+yarn update:foundry
```
refcell, line 63: We should be able to just run a `yarn install:foundry` now right?
```suggestion
yarn install:foundry
```

refcell: Oh this might also have to be swapped with step 2 since yarn might complain about missing deps

packages/contracts-bedrock/README.md
```diff
@@ -56,16 +56,18 @@ npm install @eth-optimism/contracts-bedrock
...
+
+```shell
+yarn install
+```
+
+1. Install the correct version of foundry (defined in the .foundryrc file in the root of this repo.
+
+```shell
+curl -L https://foundry.paradigm.xyz | bash
+yarn install:foundry
```
refcell, line 69: Think the suggestion borked - both lines 68 + 69 should just be `yarn install:foundry`

refcell (approved):
LGTM

.circleci/config.yml
```diff
@@ -843,7 +843,7 @@ jobs:
           command: |
             curl -L https://foundry.paradigm.xyz | bash
             source $HOME/.bashrc
-            foundryup
```
tynes, line 846: This is going to make it very slow because it will build forge, we should not do this

tynes (changes requested):
I'm only open to the change in ci if caching is added so that it doesnt need to build forge each time

.circleci/config.yml
refcell: We probably should be using the "correct" commit of foundry though no?

maurelian: I reverted this change as it seems to be working now.
If/when it breaks for being the wrong version we can handle that separately.

refcell:
Nice, PR looks good now 🤝

maurelian:
> I'm only open to the change in ci if caching is added so that it doesnt need to build forge each time

@tynes I reverted that change for simplicity.

tynes (approved):

PR #6170, opened 2023-06-27
chore(ci): Fix slither timeout

clabby:
## Overview

The `contracts-bedrock-slither` job has been timing out due to a lack of output from slither until compilation has completed. This PR adds the `--ignore-compile` flag to slither and builds the contracts prior to invoking it so that the build progress is emitted to stdout as it goes.

### DIFF
--- a/packages/contracts-bedrock/scripts/slither.sh
+++ b/packages/contracts-bedrock/scripts/slither.sh
@@ -4,8 +4,10 @@ rm -rf artifacts forge-artifacts
 
 # See slither.config.json for slither settings
 if [[ -z "$TRIAGE_MODE" ]]; then
+  echo "Building contracts"
+  forge build --build-info --force
   echo "Running slither"
-  slither .
+  slither --ignore-compile .
 else
   echo "Running slither in triage mode"
   # Slither's triage mode will run an 'interview' in the terminal, allowing you to review each of


clabby:
* develop
  * **PR #6170** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6170" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6170?utm_source=stack-comment).

packages/contracts-bedrock/scripts/slither.sh
```diff
@@ -4,8 +4,10 @@ rm -rf artifacts forge-artifacts
 
 # See slither.config.json for slither settings
 if [[ -z "$TRIAGE_MODE" ]]; then
+  echo "Building contracts"
+  forge build --build-info --force
```
refcell, line 8: Should this also be run in triage mode?

clabby: This fix is primarily to make sure that slither doesn't time out in CI due to lack of output - triage mode is normally ran locally, so we shouldn't have to do this there.

maurelian (approved):

PR #6177, opened 2023-06-27
op-challenger: Add in memory orchestrator

trianglesphere:
**Description**

This shows multiple agents wired up & playing against each other in an in-memory game.

**Tests**

Manually verified the output of the example.

**Metadata**

- Fixes CLI-4147


op-challenger/fault/cmd/examples/full.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func FullGame() {
+	log.Root().SetHandler(
+		log.LvlFilterHandler(log.LvlInfo, log.StreamHandler(os.Stdout, log.TerminalFormat(true))),
+	)
+
+	canonical := "abcdefgh"
+	disputed := "abcdexyz"
+	maxDepth := 3
```
ajsutton, line 18: nit:
```suggestion
	maxDepth uint64 := 3
```
and then you don't need to cast it multiple times below.

ajsutton (approved):
LGTM. We probably don't actually need this to be multithreaded - could just let the agents take turns at evaluating the game state and making their moves.  But it's done and working now so may as well stick with it.

I suspect semgrep will want you to remove the TODOs though. :)

op-challenger/fault/cmd/examples/full.go
sebastianst: or use a `const` for this and the strings above, then you never have to convert `maxDepth`.

op-challenger/fault/orchestrator.go
```diff
@@ -0,0 +1,87 @@
...
+		agent.PerformActions()
+		// TODO: Multiple claims / how to balance performing actions with
+		// accepting new claims
+		claim := <-claimCh
+		_ = agent.AddClaim(claim)
+
+	}
+}
+
+func (o *Orchestrator) reponderThread() {
```
sebastianst, line 59:
```suggestion
func (o *Orchestrator) responderThread() {
```
typo

refcell (approved):
Looks good aside from @sebastianst + @ajsutton 's comments :)

trianglesphere:
> We probably don't actually need this to be multithreaded

I made it multithreaded b/c there is a lock around the game state in each of the agents so doing the simple thing of piping the respond straight into the add claim would deadlock, but I think I probably could have done it single threaded by storing the responses in an array somewhere.

op-challenger/fault/cmd/examples/full.go
trianglesphere: Switched to starting it as uint64. We definitely use both uint64 & int's which makes it kinda annoying. I'll have to go through & unify the types.

PR #6183, opened 2023-06-28
feat(op-challenger): FaultResponder

refcell:
**Description**

Implements the naive fault responder. The fault responder satisfies the `Responder` interface and engages a fault dispute game via the tx manager to attack or defend a given response claim.

This is a very basic first iteration whereby the fault responder's only course of action is to `Attack` or `Defend`.

**Tests**

Multi-path unit tests around the `faultResponder`'s `Respond` function exist in `responder_test.go`. This is the only public facing function for the `Responder` interface and both the happy and sad paths are tested.

**Metatada**

Fixes CLI-4127

### DIFF
--- a/op-challenger/fault/responder.go
+++ b/op-challenger/fault/responder.go
@@ -1,0 +1,100 @@
+package fault
+
+import (
+	"context"
+	"math/big"
+
+	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
+	"github.com/ethereum-optimism/optimism/op-service/txmgr"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// faultResponder implements the [Responder] interface to send onchain transactions.
+type faultResponder struct {
+	log log.Logger
+
+	txMgr txmgr.TxManager
+
+	fdgAddr common.Address
+	fdgAbi  *abi.ABI
+}
+
+// NewFaultResponder returns a new [faultResponder].
+func NewFaultResponder(logger log.Logger, txManagr txmgr.TxManager, fdgAddr common.Address) (*faultResponder, error) {
+	fdgAbi, err := bindings.FaultDisputeGameMetaData.GetAbi()
+	if err != nil {
+		return nil, err
+	}
+	return &faultResponder{
+		log:     logger,
+		txMgr:   txManagr,
+		fdgAddr: fdgAddr,
+		fdgAbi:  fdgAbi,
+	}, nil
+}
+
+// buildFaultDefendData creates the transaction data for the Defend function.
+func (r *faultResponder) buildFaultDefendData(parentContractIndex int, pivot [32]byte) ([]byte, error) {
+	return r.fdgAbi.Pack(
+		"defend",
+		big.NewInt(int64(parentContractIndex)),
+		pivot,
+	)
+}
+
+// buildFaultAttackData creates the transaction data for the Attack function.
+func (r *faultResponder) buildFaultAttackData(parentContractIndex int, pivot [32]byte) ([]byte, error) {
+	return r.fdgAbi.Pack(
+		"attack",
+		big.NewInt(int64(parentContractIndex)),
+		pivot,
+	)
+}
+
+// BuildTx builds the transaction for the [faultResponder].
+func (r *faultResponder) BuildTx(ctx context.Context, response Claim) ([]byte, error) {
+	if response.DefendsParent() {
+		txData, err := r.buildFaultDefendData(response.ParentContractIndex, response.ValueBytes())
+		if err != nil {
+			return nil, err
+		}
+		return txData, nil
+	} else {
+		txData, err := r.buildFaultAttackData(response.ParentContractIndex, response.ValueBytes())
+		if err != nil {
+			return nil, err
+		}
+		return txData, nil
+	}
+}
+
+// Respond takes a [Claim] and executes the response action.
+func (r *faultResponder) Respond(ctx context.Context, response Claim) error {
+	// Build the transaction data.
+	txData, err := r.BuildTx(ctx, response)
+	if err != nil {
+		return err
+	}
+
+	// Send the transaction through the [txmgr].
+	receipt, err := r.txMgr.Send(ctx, txmgr.TxCandidate{
+		To:     &r.fdgAddr,
+		TxData: txData,
+		// Setting GasLimit to 0 performs gas estimation online through the [txmgr].
+		GasLimit: 0,
+	})
+	if err != nil {
+		return err
+	}
+	if receipt.Status == types.ReceiptStatusFailed {
+		r.log.Error("responder tx successfully published but reverted", "tx_hash", receipt.TxHash)
+	} else {
+		r.log.Info("responder tx successfully published", "tx_hash", receipt.TxHash)
+	}
+
+	return nil
+}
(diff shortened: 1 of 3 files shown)


refcell:
* develop
  * **PR #6183** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6183" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6183?utm_source=stack-comment).

op-challenger/fault/responder.go
```diff
@@ -0,0 +1,163 @@
...
+		txdata, err := r.fdgContract.Attack(&bind.TransactOpts{}, bigParentIndex, response.ValueBytes())
+		if err != nil {
+			return nil, err
+		}
+		return NewFaultResponderTxData(txdata), nil
+	}
+}
+
+// Respond takes a [Claim] and executes the response action.
+func (r *faultResponder) Respond(ctx context.Context, response Claim) error {
```
trianglesphere, line 74: Let's start with this function being single threaded & blocking. Doing the simplest thing means that we can wait for the transaction to be included on chain before performing the next move.

op-challenger/fault/responder.go
```diff
@@ -0,0 +1,163 @@
...
+func (r *faultResponder) Quit() {
+	r.log.Info("Shutting down fault responder")
+	r.cancelShutdownCtx()
+}
+
+// buildTxData builds the transaction data for the [faultResponder].
+func (r *faultResponder) buildTxData(ctx context.Context, response Claim) (TxData, error) {
+	// Get the parent index as a big integer.
+	parentIndex := response.Parent.ToGIndex()
+	bigParentIndex := big.NewInt(int64(parentIndex))
```
trianglesphere, line 55: This is the wrong index. I guess we need to get some more information from the contract.
The index we need to include on chain is the index of the claim inside the ClaimdData array.
https://github.com/ethereum-optimism/optimism/blob/f419bd1514ee8ad5115ea52c2c09698f59eb48a9/packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol#L53

op-challenger/fault/responder.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// sendTransaction creates & submits a transaction using the provided [TxData].
+// Under the hood, it uses the challenger's [txmgr] to handle transaction sending.
+// Gas estimation is performed using go-ethereum's core IntrinsicGas estimation function.
+// This is a blocking method. It should not be called concurrently.
+func sendTransaction(txdata TxData, queue TxQueue, receiptsCh chan txmgr.TxReceipt[TxData]) error {
+	// Do the gas estimation offline. A value of 0 will cause the [txmgr] to estimate the gas limit.
+	data := txdata.Bytes()
+	intrinsicGas, err := core.IntrinsicGas(data, nil, false, true, true, false)
```
trianglesphere, line 195: We shouldn't be using `IntrinsicGas` but instead do gas estimation. The op-proposer is a better example than the batcher for the challenger.

op-challenger/fault/responder.go
trianglesphere: not implemented yet

refcell: I thought I did make it single threaded and blocking by waiting for the receipt?

op-challenger/fault/responder_test.go
```diff
@@ -0,0 +1,42 @@
...
+	t.Skip("Responder not tested yet")
+
+	receiptChan := make(chan txmgr.TxReceipt[TxData])
+	err := fmt.Errorf("error")
+	// go func() {
+	// 	err = sendTransaction(&mockTxData{}, &mockTxQueue{}, receiptChan)
+	// }()
+	<-receiptChan
+	require.NoError(t, err)
+}
```
ajsutton, line 42: We need to be writing tests for all the parts as we go and makings sure the design is actually testable.  Needing to spawn a thread in a unit test is a sign we haven't got that design right yet.

refcell: Agreed - have been for most. Some other game solver components haven't been tested and are merged since we are still trying to get the design right i believe - architecture hasn't been sussed out yet.

That said, I can write more tests here, they will likely be deleted or clobbered though if the architecture changes so it'd take quite a bit of my time

ajsutton: Ultimately you'll go faster if you write tests as you go. Yes you might rework some of them but if you write tests that operate on the public interfaces rather than depending on internal details they won't be brittle and can handle having things change later without needing major rework.  Having tests will also push you to use a better design because code is only testable when you have appropriate coupling and a whole bunch of other key properties of high quality code.  ie writing tests is actually a key part of the design process, not something to tack on the end.

op-challenger/fault/responder.go
trianglesphere: it's not returning the error from `Send`. Also there is no reason to have the tx queue at this point b/c it exists to send multiple transactions at once.

op-challenger/fault/responder_test.go
trianglesphere: Most of the changes we've made have been tested & then only time we've updated the tests is when we change the types of stuff. The couple things that don't have unit tests probably should, but have been tested manually / are exercised through the examples.

For now I think we probably don't need concurrency in the transaction sending which will make testing much easier.

op-challenger/fault/responder.go
```diff
@@ -0,0 +1,160 @@
...
+func (r *faultResponder) Respond(ctx context.Context, response Claim) error {
+	// Build the transaction data.
+	txData, err := r.buildTxData(ctx, response)
+	if err != nil {
+		return err
+	}
+
+	// Send the transaction through the txmgr queue.
+	receiptsCh := make(chan txmgr.TxReceipt[TxData])
+	queue := txmgr.NewQueue[TxData](ctx, r.txMgr, r.MaxPendingTransactions)
```
trianglesphere, line 94: No need for the queue, just return `r.txmgr.Send(candidate)`

op-challenger/fault/responder.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// buildTx builds the transaction for the [faultResponder].
+func (r *faultResponder) buildTx(ctx context.Context, response Claim) (*types.Transaction, error) {
+	// Parent Claim index in the contract claimData array.
+	bigIndex := big.NewInt(int64(response.ParentContractIndex))
+
+	// Build the transaction data using the response claim.
+	if response.DefendsParent() {
+		tx, err := r.fdgContractTransactor.Defend(&bind.TransactOpts{}, bigIndex, response.ValueBytes())
```
trianglesphere, line 45: Take a look at https://github.com/ethereum-optimism/optimism/blob/7af004263048d6a8f1de832967769fc2ec7b5945/op-proposer/proposer/l2_output_submitter.go#L313-L320

Basically `TransactOpts` requires a signer & that doesn't work with our transaction management. In addition `TransactOpts` defaults to sending the transaction. As such, we'll have to manually pack the transaction data here.

op-challenger/fault/responder.go
```diff
@@ -0,0 +1,100 @@
...
+func (r *faultResponder) buildFaultAttackData(parentContractIndex int, pivot [32]byte) ([]byte, error) {
+	return r.fdgAbi.Pack(
+		"attack",
+		big.NewInt(int64(parentContractIndex)),
+		pivot,
+	)
+}
+
+// buildTx builds the transaction for the [faultResponder].
+func (r *faultResponder) buildTx(ctx context.Context, response Claim) ([]byte, error) {
```
trianglesphere: IMO this is the most important function to test. If we test that it's calling to the mock send function, we should also be asserting that the data it is sending is correct, not just that it made the correct number of calls.

trianglesphere (approved):
lgtm

PR #6184, opened 2023-06-28
op-challenger: Counter the root claim

trianglesphere:
**Description**

This adds the ability for the `Move` function to counter the root claim. Because the root claim does not have a parent & we no longer looked at the parent pair, we now iterate claims, not pairs of claims.

This also touches a lot of the example code to remove setting it up with the initial counter claim.

**Tests**

Unit tests to be able to counter the root claim.


trianglesphere:
> This could have just been stacked ontop of https://github.com/ethereum-optimism/optimism/pull/6178 i think

It's similar changes, but it wasn't because this work is a little bit older, but hasn't been pushed as I've been waiting for other PRs to land.

refcell (approved):
Nice, LGTM

PR #6186, opened 2023-06-28
feat(op-challenger): Claim Loader

refcell:
**Description**

Creates a net-new component to the game-solver/challenger to load and track
`Claim`s from a fault dispute game contract through polling.

**Metadata**

Fixes CLI-4126

### DIFF
--- a/op-challenger/fault/loader.go
+++ b/op-challenger/fault/loader.go
@@ -1,0 +1,101 @@
+package fault
+
+import (
+	"context"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// ClaimFetcher is a minimal interface around [bindings.FaultDisputeGameCaller].
+// This needs to be updated if the [bindings.FaultDisputeGameCaller] interface changes.
+type ClaimFetcher interface {
+	ClaimData(opts *bind.CallOpts, arg0 *big.Int) (struct {
+		ParentIndex uint32
+		Countered   bool
+		Claim       [32]byte
+		Position    *big.Int
+		Clock       *big.Int
+	}, error)
+	ClaimDataLen(opts *bind.CallOpts) (*big.Int, error)
+}
+
+// Loader is a minimal interface for loading onchain [Claim] data.
+type Loader interface {
+	FetchClaims(ctx context.Context) ([]Claim, error)
+}
+
+// loader pulls in fault dispute game claim data periodically and over subscriptions.
+type loader struct {
+	log log.Logger
+
+	state Game
+
+	claimFetcher ClaimFetcher
+}
+
+// NewLoader creates a new [loader].
+func NewLoader(log log.Logger, state Game, claimFetcher ClaimFetcher) *loader {
+	return &loader{
+		log:          log,
+		state:        state,
+		claimFetcher: claimFetcher,
+	}
+}
+
+// fetchClaim fetches a single [Claim] with a hydrated parent.
+func (l *loader) fetchClaim(ctx context.Context, arrIndex uint64) (Claim, error) {
+	callOpts := bind.CallOpts{
+		Context: ctx,
+	}
+
+	fetchedClaim, err := l.claimFetcher.ClaimData(&callOpts, new(big.Int).SetUint64(arrIndex))
+	if err != nil {
+		return Claim{}, err
+	}
+
+	claim := Claim{
+		ClaimData: ClaimData{
+			Value:    fetchedClaim.Claim,
+			Position: NewPositionFromGIndex(fetchedClaim.Position.Uint64()),
+		},
+	}
+
+	if !claim.IsRootPosition() {
+		parentIndex := uint64(fetchedClaim.ParentIndex)
+		parentClaim, err := l.claimFetcher.ClaimData(&callOpts, new(big.Int).SetUint64(parentIndex))
+		if err != nil {
+			return Claim{}, err
+		}
+		claim.Parent = ClaimData{
+			Value:    parentClaim.Claim,
+			Position: NewPositionFromGIndex(parentClaim.Position.Uint64()),
+		}
+	}
+
+	return claim, nil
+}
+
+// FetchClaims fetches all claims from the fault dispute game.
+func (l *loader) FetchClaims(ctx context.Context) ([]Claim, error) {
+	// Get the current claim count.
+	claimCount, err := l.claimFetcher.ClaimDataLen(&bind.CallOpts{
+		Context: ctx,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Fetch each claim and build a list.
+	claimList := make([]Claim, claimCount.Uint64())
+	for i := uint64(0); i < claimCount.Uint64(); i++ {
+		claim, err := l.fetchClaim(ctx, i)
+		if err != nil {
+			return nil, err
+		}
+		claimList[i] = claim
+	}
+
+	return claimList, nil
+}
--- a/op-challenger/fault/loader_test.go
+++ b/op-challenger/fault/loader_test.go
@@ -1,0 +1,180 @@
+package fault
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+	"testing"
+
+	"github.com/ethereum-optimism/optimism/op-node/testlog"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	mockClaimDataError = fmt.Errorf("claim data errored")
+	mockClaimLenError  = fmt.Errorf("claim len errored")
+	mockPutError       = fmt.Errorf("put errored")
+)
+
+type mockGameState struct {
+	putCalled int
+	putErrors bool
+}
+
+func (m *mockGameState) Put(claim Claim) error {
+	m.putCalled++
+	if m.putErrors {
+		return mockPutError
+	}
+	return nil
+}
+
+func (m *mockGameState) PutAll(claims []Claim) error {
+	m.putCalled += len(claims)
+	if m.putErrors {
+		return mockPutError
+	}
+	return nil
+}
+
+func (m *mockGameState) Claims() []Claim {
+	return []Claim{}
+}
+
+func (m *mockGameState) IsDuplicate(claim Claim) bool {
+	return false
+}
+
+type mockClaimFetcher struct {
+	claimDataError bool
+	claimLenError  bool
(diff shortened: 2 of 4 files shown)


refcell:
* develop
  * **PR #6186** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6186" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6186?utm_source=stack-comment).

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+
+	// Grab each claim and push it to the [gameState].
+	for i := uint64(0); i < claimCount.Uint64(); i++ {
+		claim, err := l.fetchClaim(ctx, i)
+		if err != nil {
+			return err
+		}
+		if err := l.state.Put(*claim); err != nil {
```
trianglesphere, line 85: `FetchClaims` should return the list of claims & then you should insert the claims into the state separately.

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return []Claim{}, err
+	}
+
+	// Fetch each claim and build a list.
+	claimList := make([]Claim, claimCount.Uint64())
+	for i := uint64(0); i < claimCount.Uint64(); i++ {
+		claim, err := l.fetchClaim(ctx, i)
+		if err != nil {
+			return []Claim{}, err
```
ajsutton, line 84: nit:
```suggestion
			return nil, err
```

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return &claim, nil
+}
+
+// FetchClaims fetches all claims from the fault dispute game.
+func (l *loader) FetchClaims(ctx context.Context) ([]Claim, error) {
+	// Get the current claim count.
+	claimCount, err := l.claimFetcher.ClaimDataLen(&bind.CallOpts{})
+	if err != nil {
+		return []Claim{}, err
```
ajsutton, line 76: nit:
```suggestion
		return nil, err
```

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,100 @@
...
+
+// PushClaims pushes a list of claims into the [Game] state.
+func (l *loader) PushClaims(ctx context.Context, claims []Claim) error {
+	for _, claim := range claims {
+		if err := l.state.Put(claim); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
ajsutton, line 100: Why does this method exist in the claim fetcher? Seems out of scope for it.

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,100 @@
...
+func NewLoader(log log.Logger, state Game, claimFetcher ClaimFetcher) *loader {
+	return &loader{
+		log:          log,
+		state:        state,
+		claimFetcher: claimFetcher,
+	}
+}
+
+// fetchClaim fetches a single [Claim] with a hydrated parent.
+func (l *loader) fetchClaim(ctx context.Context, arrIndex uint64) (*Claim, error) {
```
ajsutton, line 43: `ctx` isn't actually being used so can be removed from the API and from `FetchClaims`.

I'd probably modify it to just return `Claim` rather than a pointer as well given the only usage immediately dereferences it.  Just means returning `Claim{}, err` in a few places but that's fine.

op-challenger/fault/loader_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+// TestLoader_FetchClaims_Succeeds tests [loader.FetchClaims].
+func TestLoader_FetchClaims_Succeeds(t *testing.T) {
+	log := testlog.Logger(t, log.LvlError)
+	mockClaimFetcher := &mockClaimFetcher{}
+	loader := NewLoader(log, &mockGameState{}, mockClaimFetcher)
+	claims, err := loader.FetchClaims(context.Background())
+	require.NoError(t, err)
+	require.Equal(t, len(claims), 1)
+}
```
ajsutton, line 126: This isn't really testing all that much.  I'd say we want to be able to create `mockClaimFetcher` with a specified list of claims that it will return and then we should be able to assert that `loader.FetchClaims` returns that same list of claims.  Currently this test would pass even if `FetchClaims` was just:
```
		claim, _ := l.fetchClaim(ctx, 0)
		return []Claim{*claim}, nil
```
Admittedly you'd need to handle the error to make the next test pass too but there's nothing checking that it's actually retrieving the right number of claims.

ajsutton (commented):
Left some comments, but I think the big issue here is that this code isn't reorg safe. It's quite possible that when `ClaimDataLen` is called there are 6 claims but a reorg then happens and there are then only 5 claims when `ClaimData` is called or that the order of claims in the array changes etc.

Each fetch needs to be given a specific block hash and use that in all calls to the contract to ensure it has a consistent view.  The code that calls this would likely just get the block hash of `latest` or possibly a block close to head but with a small follow distance and pass that it, but the fetcher shouldn't be using `latest` or it will get an inconsistent view.

op-challenger/fault/loader.go
trianglesphere: Because this code does network calls, I think we should add the `ctx` to `CallOpts` (which is how it is uses in this API) instead of removing it.

op-challenger/fault/loader.go
trianglesphere: also unfortunately go sometimes but not always treats `nil` differently from `[]Claim{}` depending on the circumstance.

ajsutton: True, but given we're returning an error there shouldn't be any assumptions made about the value returned here and `nil` is the more typical pattern at least from what I've seen.

op-challenger/fault/loader.go
trianglesphere: maybe this could get moved to the game agent & the game agent takes a `FetchAllClaims` interface & then puts all of the claims into the game state

ajsutton: Yeah agreed - either in the game agent or make this a `AddAll` method on game state.  I lean towards the latter because that feels like a nicer API for the game state to expose and minimises the glue code required.

refcell:
> Left some comments, but I think the big issue here is that this code isn't reorg safe. It's quite possible that when `ClaimDataLen` is called there are 6 claims but a reorg then happens and there are then only 5 claims when `ClaimData` is called or that the order of claims in the array changes etc.
> 
> Each fetch needs to be given a specific block hash and use that in all calls to the contract to ensure it has a consistent view. The code that calls this would likely just get the block hash of `latest` or possibly a block close to head but with a small follow distance and pass that it, but the fetcher shouldn't be using `latest` or it will get an inconsistent view.

Good callout - pinned claim fetching to a given `*big.Int` block number as that's what the CallOpts accepts.

I will leave the "latest" block number fetching out of the scope of this PR and up to the downstream usage of the claim loader to provide.

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,104 @@
...
+		state:        state,
+		claimFetcher: claimFetcher,
+	}
+}
+
+// fetchClaim fetches a single [Claim] with a hydrated parent.
+func (l *loader) fetchClaim(ctx context.Context, arrIndex uint64, blockNum *big.Int) (*Claim, error) {
+	fetchedClaim, err := l.claimFetcher.ClaimData(&bind.CallOpts{
+		Context:     ctx,
+		BlockNumber: blockNum,
```
trianglesphere, line 51: don't use `BlockNumber` here because we don't want to accidentally require archive state.

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,104 @@
...
+			Value:    fetchedClaim.Claim,
+			Position: NewPositionFromGIndex(fetchedClaim.Position.Uint64()),
+		},
+	}
+
+	if !claim.IsRootPosition() {
+		parentIndex := uint64(fetchedClaim.ParentIndex)
+		parentClaim, err := l.claimFetcher.ClaimData(&bind.CallOpts{
+			Context:     ctx,
+			BlockNumber: blockNum,
```
trianglesphere, line 68: ditto

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,104 @@
...
+	return &loader{
+		log:          log,
+		state:        state,
+		claimFetcher: claimFetcher,
+	}
+}
+
+// fetchClaim fetches a single [Claim] with a hydrated parent.
+func (l *loader) fetchClaim(ctx context.Context, arrIndex uint64, blockNum *big.Int) (*Claim, error) {
+	fetchedClaim, err := l.claimFetcher.ClaimData(&bind.CallOpts{
```
trianglesphere, line 49: can share `bind.CallOpts`

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,104 @@
...
+func NewLoader(log log.Logger, state Game, claimFetcher ClaimFetcher) *loader {
+	return &loader{
+		log:          log,
+		state:        state,
+		claimFetcher: claimFetcher,
+	}
+}
+
+// fetchClaim fetches a single [Claim] with a hydrated parent.
+func (l *loader) fetchClaim(ctx context.Context, arrIndex uint64, blockNum *big.Int) (*Claim, error) {
```
trianglesphere, line 48: generally returning a `Claim` is easier than returning `*Claim`

op-challenger/fault/loader.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return &claim, nil
+}
+
+// FetchClaims fetches all claims from the fault dispute game.
+func (l *loader) FetchClaims(ctx context.Context, blockNum *big.Int) ([]Claim, error) {
+	// Get the current claim count.
+	claimCount, err := l.claimFetcher.ClaimDataLen(&bind.CallOpts{
+		Context:     ctx,
+		BlockNumber: blockNum,
```
trianglesphere, line 87: ditto

refcell:
Removing the `blockNumber` pin from callopts right now and tracking reorg hardening in a [new linear ticket](https://linear.app/optimism/issue/CLI-4164/harden-claim-loading).
Currently, we're working towards the MVP op-challenger/game solver and this will be a later task.

trianglesphere (approved):
lgtm

ajsutton:
> Good callout - pinned claim fetching to a given `*big.Int` block number as that's what the CallOpts accepts.
> 
> I will leave the "latest" block number fetching out of the scope of this PR and up to the downstream usage of the claim loader to provide.

Pinning to a block number is not sufficient to avoid reorg issues unfortunately. It has to be a block hash.

PR #6190, opened 2023-06-29
op-node: Add `admin_sequencerActive` RPC method

ajsutton:
**Description**

Add `admin_sequencerActive` RPC method.

Returns true if the node is actively sequencing, otherwise false. This is essentially a parallel with the `eth_mining` API from execution clients.

**Tests**

Updated e2e tests for admin API to cover the new endpoint.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4160/rpc-to-get-current-sequencer-state


trianglesphere (approved):
LGTM

PR #6191, opened 2023-06-29
cannon: Fix unaligned pre-image key writes in Go VM

Inphi:
Fixes a mips.go bug where pre-image key writes from unaligned memory would always start at the aligned offset.

For example, given a buffer containing `0xdeADbeEF` (big-endian) located at `0x32000001`, writing 1 byte to the pre-image request stream should only write a `0xAD` value.


protolambda (approved):
nice find!

PR #6196, opened 2023-06-29
Fault Proof VM Specs

Inphi:
**Description**

Fault Proof VM specification.

**Metadata**

Fixes CLI-4149


specs/fault-proof-vm.md
```diff
@@ -0,0 +1,177 @@
+# Fault Proof Virtual Machine Specification
```
protolambda: I would clarify this is the Cannon FPVM. There may be multiple FPVMs in the future, and we don't want to enshrine MIPS forever.

specs/fault-proof-vm.md
```diff
@@ -0,0 +1,177 @@
...
+  - [Hint Communication](#hint-communication)
+  - [Pre-image Communication](#pre-image-communication)
+    - [Pre-image I/O Alignment](#pre-image-io-alignment)
+- [Exceptions](#exceptions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+This is a description of the Fault Proof Virtual Machine (FPVM). The FPVM emulates a minimal Linux-based system
```
protolambda: Same here, clarify it's the Cannon Fault Proof Virtual Machine.

specs/fault-proof-vm.md
```diff
@@ -0,0 +1,177 @@
...
+
+The FPVM has a fixed program break at `0x40000000`. However, the FPVM is permitted to extend the
+heap beyond this limit via mmap syscalls.
+For simplicity, there are no memory protections against "heap overruns" against other memory segments.
+Such VM steps are still considered valid state transitions.
+
+Specification of memory mappings is outside the scope of this document as it is irrelevant to
+the VM state. FPVM implementers may refer to the Linux/MIPS kernel for inspiration.
+
+## Delay Slots
```
protolambda: Maybe add here that two consecutive delay-slot related instructions in memory are invalid MIPS (since there would be undefined behavior, with the delay slot being set already but being modified within the same delay).

specs/fault-proof-vm.md
```diff
@@ -0,0 +1,177 @@
...
+| 4090 | mmap | uint32 addr | uint32 len | | Allocates a page from the heap. See [heap](#heap) for details. |
+| 4045 | brk | | | | Returns a fixed address for the program break at `0x40000000` |
+| 4120 | clone | | | | Returns 1 |
+| 4246 | exit_group | uint8 exit_code | | | Sets the Exited and ExitCode states to `true` and `$a0` respectively. |
+| 4003 | read | uint32 fd | char *buf | uint32 count | Similar behavior as Linux/MIPS with support for unaligned reads. See [I/O](#io) for more details. |
+| 4004 | write | uint32 fd | char *buf | uint32 count | Similar behavior as Linux/MIPS with support for unaligned writes. See [I/O](#io) for more details. |
+| 4055 | fcntl | uint32 fd | int32 cmd | | Similar behavior as Linux/MIPS. Only the `F_GETFL` (3) cmd is supported. Sets errno to `0x16` for all other commands |
+
+For all of the above syscalls, an error is indicated by setting the return register (`$v0`) to
+`0xFFFFFFFF` and `errno` (`$a3`) is set accordingly.
```
protolambda:
```suggestion
`0xFFFFFFFF` (-1) and `errno` (`$a3`) is set accordingly.
```

It's a signed integer, -1, in the linux syscall docs.

specs/fault-proof-vm.md
```diff
@@ -0,0 +1,177 @@
...
+| ---- | --------------- | ----------- |
+| stdin | 0 | read-only standard input stream. |
+| stdout | 1 | write-only standaard output stream. |
+| stderr | 2 | write-only standard error stream. |
+| hint response | 3 | read-only. Used to read the status of [pre-image hinting](./fault-proof.md#hinting). |
+| hint request | 4 | write-only. Used to provide [pre-image hints](./fault-proof.md#hinting) |
+| pre-image response | 5 | read-only. Used to [read pre-images](./fault-proof.md#pre-image-communication). |
+| pre-image request | 6 | write-only. Used to [request pre-images](./fault-proof.md#pre-image-communication). |
+
+Syscalls referencing unnkown file descriptors fail with an `EBADF` errno as done on Linux/MIPS.
```
protolambda:
```suggestion
Syscalls referencing unnkown file descriptors fail with an `EBADF` errno as done on Linux.
```

specs/fault-proof-vm.md
```diff
@@ -0,0 +1,177 @@
...
+Writing to and reading from standard output, input and error streams have no effect on the FPVM state.
+FPVM implementations may use them for debugging purposes as long as I/O is stateless.
+
+All I/O operations are restricted to a maximum of 4 bytes per operation.
+Any read or write syscall request exceeding this limit will be truncated to 4 bytes.
+Consequently, the return value of read/write syscalls is at most 4, indicating the actual number of bytes read/written.
+
+### Standard Streams
+
+Writing to stderr/stdout standard stream always succeeds with the write count input returned.
```
protolambda:
```suggestion
Writing to stderr/stdout standard stream always succeeds with the write count input returned, effectively continuing execution without writing work.
```

specs/fault-proof-vm.md
```diff
@@ -0,0 +1,177 @@
...
+Reading from stdin has no effect other than to return zero and errno set to 0.
```
protolambda:
```suggestion
Reading from stdin has no effect other than to return zero and errno set to 0, signalling that there is no input.
```

specs/fault-proof-vm.md
Inphi: Good point. The VM behavior here should be deterministic though rather than undefined. I'll update the spec to indicate that an Exception should be triggered, preventing a VM step to occur.
The Go and Solidity VMs may need to be updated to adhere to this.

protolambda: I think it can be implemented by checking if already `NextPC != PC + 4` during a jump/branch/etc. and then raising an exception. The Go compiler should never output such sequence though.

refcell (approved):
LGTM 🔥

PR #6199, opened 2023-06-29
feat(ctb): Time-box `FaultDisputeGame::resolve()`

clabby:
## Overview

Time-boxes the `resolve` function in the `FaultDisputeGame` contract and adds several tests for the correctness of the chess clock logic. During initial iteration we left the `resolve` function without necessary clock checks, but we need to ensure that the game may not be resolved unless the clock of the path that opposes the winning node has ran out.

This change is safe for several reasons:
1. If the winning node is the root claim, it means that either:
    1. The root claim was never countered.
    1. All counters to the root claim have been pushed to a step, where they were disproven and marked as `countered`.

    In this situation, we can check if more than 3.5 days has passed since the root claim's creation. Any potential challenger at this point that wants to provide an alternative attack against the root no longer can due to their clock running out.
2. If the winning node is not the root claim, it may only be considered for resolution if its parent has run out of time on their clock. The winning node can change as the game goes on, but we can safely use this heuristic due to the fact that clocks always are inherited from their grandparents when a fork is created.

**Note**
Also includes a hacky fix for the bindings where we were skipping over storage types that had ast ids during the canonicalization process.


packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -34,7 +34,6 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
     IBigStepper public immutable VM;
 
     /// @notice The duration of the game.
-    /// @dev TODO: Account for resolution buffer. (?)
```
refcell, line 37: We're comfortable ignoring this for now?

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -315,12 +311,29 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
...
+        Clock opposingClock = parentIndex == type(uint32).max
+            ? leftMostUncontested.clock
+            : claimData[parentIndex].clock;
+        if (
+            Duration.unwrap(opposingClock.duration()) +
+                (block.timestamp - Timestamp.unwrap(opposingClock.timestamp())) <=
+            Duration.unwrap(GAME_DURATION) >> 1
+        ) {
+            revert ClockNotExpired();
+        }
```
refcell, line 330: Looks good to me 👍

packages/contracts-bedrock/contracts/test/FaultDisputeGame.t.sol
```diff
@@ -237,11 +280,19 @@ contract FaultDisputeGame_Test is FaultDisputeGame_Init {
...
         assertEq(uint8(gameProxy.status()), uint8(GameStatus.DEFENDER_WINS));
     }
 
+    /// @dev Static unit test for the correctness an uncontested root resolution.
+    function test_resolve_rootUncontestedClockNotExpired_succeeds() public {
+        vm.warp(block.timestamp + 3 days + 12 hours);
+        vm.expectRevert(ClockNotExpired.selector);
+        gameProxy.resolve();
+    }
+
```
refcell, line 295: nice

refcell (approved):

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
clabby: We've decided to add no resolution buffer for the game. There's a direct incentive to resolve the game in order to make bonds available, so there's not a concern about a game being left unresolved for long at all if we maintain a liveness assumption on the honest challenger 😄

clabby:
@mergifyio queue

PR #6207, opened 2023-06-29
op-challenger: Get Pre/Post claim for Step function

trianglesphere:
**Description**

This adds the ability to fetch the Claim which has a trace index before & after a specific claim. This is the claim that serves as the pre or post state to a leaf claim.

**Tests**

Unit tests.


ajsutton (approved):
LGTM.

PR #6208, opened 2023-06-30
feat: processor retries. index L2 ouputs posted on l1

hamdiallam:
1. The processor should be able to retry headers on errors. In order to support this, we need the ability for the processor to advance through buffered headers. Currently the processor panics when it errors 
2. The L1 processor should index L2 outputs

## Notes
- Not too happy with the buffered_traversal abstraction. We can probably merge this with header_traversal? Going with this for now to unblock e2e functionality of the indexer.


hamdiallam:
Per conversation with @roninjin10

It doesn't make sense for the `l2_processor` to lag behind the `l1_processor` since we still end up with the same problem described in reverse.
 - When indexing a withdrawal on L1 we want to make sure that we've seen the corresponding initiation on L2. With the L2 processor lagging behind L1, this becomes challenging to do

A better way to solve this problem is to actually have bridge processing work independently of the L1 and L2 processors. The bridge processor will work in lockstep with indexed l1 and l2 blocks. This design is also easier to parallelize

hamdiallam:
This still addresses DX-79 so i'll keep this published but remove the constraint on the L2 processor

hamdiallam:
closing this and will re-open (new pr) when done with the changes

PR #6210, opened 2023-06-30
feat: Improve release process

roninjin10:
- Move to a one branch model
  - We don't want to do any non revertable changes like modify github branch settings and delete branches yet
  - Instead we want to start using develop as the one branch to do deploys from
  - After stable we then permanently move to whatever our final branch is `main` or `master`
- Publish snapshot releases on every merge to develop
  - This means we are publishing with changesets every release
  - This removes the common occurance of changeset releases being broken and not finding out until attempting to do a real release
- Replace our old (semibroken) canary release workflow with snapshot releases
  - Canary release is broken because there is a bug (always existed) where if you publish multiple new versions of dependant packages the integrity gets messed up on one of the packages
  - Snapshot releases are the recommended way to do these releases with changesets
- Create a reusable setup action
- Small improvements


roninjin10:
* develop
  * **PR #6287** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6287" target="_blank">Graphite</a> 
    * **PR #6210** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6210" target="_blank">Graphite</a>  👈
      * **PR #6303** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6303" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6210?utm_source=stack-comment).

.changeset/config.json
```diff
@@ -1,10 +1,11 @@
 {
-  "$schema": "https://unpkg.com/@changesets/config@1.6.0/schema.json",
-  "changelog": "@changesets/cli/changelog",
+  "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
+  "changelog": ["@changesets/changelog-github", { "repo": "latticexyz/mud" }],
```
roninjin10:
```suggestion
  "changelog": ["@changesets/changelog-github", { "repo": "eth-optimism/optimism" }],
```

.changeset/config.json
```diff
@@ -1,10 +1,11 @@
...
   "commit": false,
+  "fixed": [],
   "linked": [],
   "access": "public",
-  "baseBranch": "master",
+  "baseBranch": "develop",
```
roninjin10, line 8: we are treating our release process as a single branch with develop

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
+name: Setup
```
roninjin10, line 1: This is a reusable setup action we should use in other workflows

roninjin10: To keep pr focused I didn't change other workflows to using it yet but will in future pr

.github/workflows/release-docker-canary.yml
```diff
@@ -1,4 +1,4 @@
-name: Publish Packages (canary)
+name: Publish Docker images (canary)
```
roninjin10, line 1: This only publishes docker now. NPM releases are published everytime we merge to develop

roninjin10: We will want to validate this works as soon as it merges and that it didn't have some hidden dependency on the npm workflow running.  It shouldn't though because all the other releases are purely 100% docker

.github/workflows/release-snapshot.yml
```diff
@@ -0,0 +1,38 @@
+name: Release snapshot
+
+# Releases a snapshot release when new commits merge to develop
+# This ensures the release process is working as expected as well as always gives us a release
+# we should move to a single branch in near future
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - develop
```
roninjin10, line 10: This replaces our canary releases and this will happen automatically everytime we merge to develop

.github/workflows/release-snapshot.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  snapshot-snapshot:
+    name: Publish snapshot release to npm
+    if: github.repository == 'eth-optimism/optimism'
```
roninjin10, line 15: This makes it so this workflow doesn't run on forks of optimism

.github/workflows/release.yml
```diff
@@ -1,9 +1,16 @@
...
+    inputs:
+      branch:
+        description: "Trigger a release workflow"
+        type: string
+        default: "develop"
+        required: true
   push:
     branches:
-      - master
+      - develop
```
roninjin10, line 13: Moving to a single branch model this works purely off of develop

.github/workflows/release.yml
```diff
@@ -12,6 +19,7 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
+    if: github.repository == 'eth-optimism/optimism'
```
roninjin10, line 22: This makes it so forks of optimism don't run this workflow

.github/workflows/release.yml
```diff
@@ -1,9 +1,16 @@
...
+name: Release and version
 
 on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Trigger a release workflow"
+        type: string
+        default: "develop"
+        required: true
```
roninjin10, line 10: Can be manually run too (though we woudln't need to manually run it). Manually running it can be necessary if networking flake happens for example

package.json
```diff
@@ -25,11 +25,15 @@
...
     "release": "npx nx run-many --target=build --skip-nx-cache && pnpm changeset publish",
+    "release:check": "changeset status --verbose --since=origin/main",
+    "release:publish": "npx nx run-many --target=build && pnpm install && pnpm build && changeset publish",
+    "release:version": "changeset version && pnpm install --lockfile-only",
     "install:foundry": "curl -L https://foundry.paradigm.xyz | bash && pnpm update:foundry",
     "update:foundry": "foundryup -C $(cat .foundryrc)"
   },
   "devDependencies": {
     "@babel/eslint-parser": "^7.18.2",
+    "@changesets/changelog-github": "^0.4.8",
```
roninjin10, line 36: This changelong is better because it will actually link to github commits

.changeset/config.json
```diff
@@ -1,10 +1,11 @@
 {
-  "$schema": "https://unpkg.com/@changesets/config@1.6.0/schema.json",
-  "changelog": "@changesets/cli/changelog",
+  "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
+  "changelog": ["@changesets/changelog-github", { "repo": "eth-optimism/optimism" }],
```
roninjin10, line 3: this puts git commit links in our release notes

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+      run: forge build
+
+    - name: Derive appropriate SHAs for base and head for `nx affected` commands
+      uses: nrwl/nx-set-shas@v3
+
+    - run: |
+        echo "nx using following shas:"
+        echo "BASE: ${{ env.NX_BASE }}"
+        echo "HEAD: ${{ env.NX_HEAD }}"
+      shell: bash
```
roninjin10, line 38: @tynes this should fix the git sha issue for nx

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Setup node 16.x
+      uses: actions/setup-node@v3
+      with:
+        node-version-file: .nvmrc
+        registry-url: https://registry.npmjs.org
+        cache: pnpm
+
+    - name: Get foundry version
+      id: foundry-version
+      run: echo ::set-output name=version::$(cat .foundryrc)
```
roninjin10, line 20: @maurelian this the .foundryrc is making the devops-y part of my brain very happy this is great hermetic building here

maurelian: 🫡

package.json
```diff
@@ -25,11 +25,15 @@
     "ready": "pnpm lint && pnpm test",
     "prepare": "husky install",
     "release": "npx nx run-many --target=build --skip-nx-cache && pnpm changeset publish",
+    "release:check": "changeset status --verbose --since=origin/main",
```
roninjin10, line 28:
```suggestion
    "release:check": "changeset status --verbose --since=origin/develop",
```
copy pasta

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+description: Common setup steps used by our workflows
+runs:
+  using: composite
+  steps:
+    - name: Setup pnpm
+      uses: pnpm/action-setup@v2
+      with:
+        version: 8.6.5
+
+    - name: Setup node 16.x
```
roninjin10, line 11:
```suggestion
    - name: Setup node with nvmrc
```
Shouldn't have to remember to update this if nvmrc changes

nickbalestra (commented):
Thanks for this PR @roninjin10 ! 
To be able to provide better feedback, as there are a lot of files touched and a lot of comments, would be useful if you could add more details in the description on how the new improved release process work in detail. 
Here few pointers that would be helpful for me to better understand in order to be able to provide the feedback you request:

- How does the canary release workflow currently work in details? and why is currently semibroken?
- How does the new snapshot release proposed with this PR work in details? and what problem this address on the current canary release
- How does moving to a one branch model work within a codebase which is not running on a one branch model yet

Excited to learn more and thanks for clarifying!!

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    # Workaround for parallel building with forge:
+    # we need to run a build on a single package to make sure
+    # that no race conditions happen with multiple forge instances
+    # later when we call nx build
+    # see https://github.com/foundry-rs/foundry/issues/4736
+    - name: Forge issue 4736 workaround
+      shell: bash
+      working-directory: packages/contracts-bedrock
+      run: forge build
```
roninjin10: without this bug if you have forge ever run in parallel (which happens easily with nx) the process crashes

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    - name: Derive appropriate SHAs for base and head for `nx affected` commands
+      uses: nrwl/nx-set-shas@v3
```
roninjin10, line 32: This is how github actions derive the nx shas. This is necessary to always have nx commands reliably work

.github/workflows/release-snapshot.yml
mslipper: Don't you mean ethereum-optimism?

nickbalestra (approved):
Thanks for explaining the context Will. Approving this as a first step toward trunk-base developement... super stoked to see this happening

.github/workflows/release-snapshot.yml
roninjin10:
```suggestion
    if: github.repository == 'ethereum-optimism/optimism'
```

mslipper (approved):

package.json
```diff
@@ -25,11 +25,15 @@
     "ready": "pnpm lint && pnpm test",
     "prepare": "husky install",
     "release": "npx nx run-many --target=build --skip-nx-cache && pnpm changeset publish",
+    "release:check": "changeset status --verbose --since=origin/develop",
```
roninjin10:
```suggestion
    "release:status": "changeset status --verbose --since=origin/develop",
```
small nit will do in follow up pr

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Setup foundry
+      uses: foundry-rs/foundry-toolchain@v1
+      with:
+        version: $${ steps.foundry-version.outputs.version }
+
+    - name: Install node dependencies
+      shell: bash
+      run: pnpm install --frozen-lockfile
+
+    # Workaround for parallel building with forge:
```
tynes: If you rebase this, we no longer need this because there is only 1 package using forge build

tynes:
Do we still need the patch on changesets in the `patches` directory after this?

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        node-version-file: .nvmrc
+        registry-url: https://registry.npmjs.org
+        cache: pnpm
+
+    - name: Get foundry version
+      id: foundry-version
+      run: echo ::set-output name=version::$(cat .foundryrc)
+
+    - name: Setup foundry
+      uses: foundry-rs/foundry-toolchain@v1
```
tynes, line 23: Is this going to cache? Its very slow to build foundry

roninjin10: It's pretty fast everywhere else I used it. I think it's just pulling the image not building from source

roninjin10:
`Do we still need the patch on changesets in the patches directory after this?`
Hmm I don't believe so.  I wasn't aware of this patch so thanks for bringing it up. In my experience changesets did attempt to publish private packages but I will have to check

tynes:
> Do we still need the patch on changesets in the `patches` directory after this?

I don't think so, since the

> `Do we still need the patch on changesets in the patches directory after this?` Hmm I don't believe so. I wasn't aware of this patch so thanks for bringing it up. In my experience changesets did attempt to publish private packages but I will have to check

I don't think we need it either, if i understand correctly we are updating the version of changesets used so the patch will be a noop?

.github/actions/setup/action.yml
tynes: Foundry only makes a certain number of images available, as new ones are published it deletes old ones. I am pretty sure this will build from source given we are not using a super recent version

roninjin10:
Yea testing out on my own repo it seems like changesets doesn't block attempting to publish private packages so I believe we can remove patches. (alternatively we'd have to update it as it is a no-op as is @tynes)

roninjin10:
Actually just realized the changeset version didn't change (only the schema). I believe we can remove patches though just form my own testing this bug doesn't seem to exist. Will make that it's own pr though after this lands as is

.github/actions/setup/action.yml
roninjin10: Ok we can look at that later. I won't consider this a blocker because I didn't add this this already existed and was just moved from a different file.

roninjin10: Oh wait I think I did add the versioning though nvm one sec

roninjin10: Ok confirmed the action is smart enough to cache https://github.com/foundry-rs/foundry-toolchain/blob/master/action.yml#L9 .

roninjin10:
Ok after this merges gotta verify all the changeset workflows are working

PR #6217, opened 2023-06-30
cannon: Allow EOF pre-image reads

Inphi:
**Description**

This patch fixes a discrepancy in the Go and Solidity FPVM implementations. The set of states, wherein their pre-image offsets are at the pre-image EOF, should be considered valid. The transition for these states is a zero-byte pre-image read.
This behavior is similar to the way Linux handles read(2) syscalls where reading at an EOF stream yields zero bytes read.

PreimageOracle.sol is updated to accommodate this by allowing pre-image parts to be loaded at the pre-image EOF.

**Tests**

- oracle.asm

PreimageOracle.sol foundry tests will be added at a later PR.


protolambda (approved):
LGTM, please rebase / fix merge conflicts

clabby (approved):
LGTM!

clabby:
@mergifyio queue

PR #6218, opened 2023-06-30
feat: Use pnpm fetch for maximum docker cache hits

roninjin10:
- Before I added a giant hack to make a maintainable dockerfile that had maximum cache hits
- now that we are using pnpm we can delete that hack and just use pnpm fetch
- pnpm fetch fetches node_modules using only the lockfile so the package.json don't need to be copied in
- this solves problem of either needing to have poor cache hits or having an unmaintainable dockerfile wehre you need to explicitly copy every package.json

### DIFF
--- a/ops/docker/Dockerfile.packages
+++ b/ops/docker/Dockerfile.packages
@@ -1,35 +1,6 @@
 # This Dockerfile builds all the dependencies needed by the monorepo, and should
 # be used to build any of the follow-on services
 #
-
-# Stage 0 (named `manifests`) collects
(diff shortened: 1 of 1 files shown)


roninjin10:
* develop
  * **PR #6218** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6218" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6218?utm_source=stack-comment).

roninjin10:
This is odd I'll circle back to this later but this doesn't actually improve anything in any way other than deleting code

PR #6220, opened 2023-06-30
op-challenger: Add AttemptStep function

trianglesphere:
**Description**

This adds the ability to get most (but not quite all) of the data required to execute the `Step` function on the contract.

**Tests**

Unit test for the attack case. None for the defend case yet (needs a different game state)


op-challenger/fault/solver.go
```diff
@@ -25,19 +25,56 @@ func NewSolver(gameDepth int, traceProvider TraceProvider) *Solver {
...
+
+type StepData struct {
+	LeafClaim  Claim
+	StateClaim Claim
+	IsAttack   bool
+}
+
+func (s *Solver) AttemptStep(claim Claim, state Game) (StepData, error) {
+	if claim.Depth() != s.gameDepth {
+		return StepData{}, errors.New("cannot step on non-leaf claims")
```
refcell, line 43: I think we should create a unit test around this path as well to make it explicit.

refcell (commented):
Request for 1 unit test, otherwise LGTM

op-challenger/fault/solver.go
trianglesphere: done.

refcell (approved):

PR #6224, opened 2023-06-30
op-challenger: Simplify init

trianglesphere:
**Description**

Removes lot of code that isn't needed for V0. Some code will be resurrected, some won't need to be.

**Tests**

Manually tested that the simplified binary interface still works.


op-challenger/cmd/main.go
```diff
@@ -35,47 +33,33 @@ var VersionWithMeta = func() string {
...
-	// otherwise the final critical log won't show what the parsing error was.
 	oplog.SetupDefaults()
 
 	app := cli.NewApp()
 	app.Version = VersionWithMeta
 	app.Flags = flags.Flags
 	app.Name = "op-challenger"
-	app.Usage = "Challenge Invalid L2OutputOracle Outputs"
-	app.Description = "A modular op-stack challenge agent for dispute games written in golang."
+	app.Usage = "Play the fault game"
```
refcell: I don't understand tightly coupling the op-challenger with the fault game if we explicitly have a `fault` subdir and know it is one way of deleting outputs.
Think we should keep this more simple to start with.
```suggestion
	app.Usage = "Challenge outputs"
```

op-challenger/config/config.go
```diff
@@ -17,79 +17,38 @@ import (
...
-		return ErrInvalidNetworkTimeout
-	}
-	if c.TxMgrConfig == nil {
-		return ErrMissingTxMgrConfig
+		return errors.New("empty l1 rpc")
 	}
-	if c.RPCConfig == nil {
-		return ErrMissingRPCConfig
+	if c.GameAddress == (common.Address{}) {
+		return errors.New("empty game address")
```
refcell: If we keep these errors defined, we can keep some config tests which I believe are still somewhat useful and follow `op-program` testing precedent.

op-challenger/cmd/main.go
trianglesphere: Sure, we can change this. I mean it doesn't actually challenge outputs & there's only one current way to challenge outputs.

op-challenger/config/config.go
trianglesphere: Restored these errors & the tests. Tests did change b/c the configs really should have been structs, not pointers.

refcell (approved):
LGTM

PR #6227, opened 2023-07-03
fixup(ctb): `step` bug

clabby:
## Overview

Fixes a bug in the `FaultDisputeGame`'s step function design where a malicious party could supply a pre/post state index that was invalid. While there was a check for whether or not the supplied DAG index had a claim that committed to the correct trace index, it could not verify that the claim was in the correct path.
Gindices
This could enable a malicious actor to trivially disprove a valid leaf node claim.

**Solution**
Move the computation for finding the pre/post state for the step on chain so that we can verify that the pre & post state for `step` exist within the path that is currently being disputed. The current implementation of this unfortunately introduces a loop into the `step` function. There may be a way to verify the inclusion of a node in a specific path in constant time if we add some extra metadata into `ClaimData`, but the core issue is fixed.

**Cons**
Introduces a loop into `step`. This could have an impact on bonding in the future.


clabby:
* clabby/ctb/fault-game-invariants
  * **PR #6227** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6227" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6227?utm_source=stack-comment).

refcell (approved):
LGTM

Inphi (approved):
LGTM!

PR #6229, opened 2023-07-04
feat(contracts-bedrock): Add ERC1155 bridge

patitonar:
From [BootNode](https://www.bootnode.dev/) we are opening this PR to start the conversation on whether having an ERC1155 bridge as part of the official stack, to complement the existing ETH, ERC20, and ERC721 bridge, makes sense to the Optimism community.

### Motivation

There are several reasons why it is a good idea to have an ERC1155 bridge on Optimism:

- 🧐**Missing piece of the stack**: Currently, there is no way to bridge ERC1155 tokens to Optimism.
- 👀**Other L2s have it**: Other L2 and sidechains already offer an ERC1155 bridge as part of their smart contract infrastructure. This puts Optimism at a competitive disadvantage, as developers may build their applications on other L2s with this functionality.
- ✊**Popular protocols are embracing it**: Popular protocols, such as Uniswap v4, are starting to embrace ERC1155 beyond current adoption. This means that there is a growing demand for ERC1155 tokens, and having an ERC1155 bridge on Optimism will make it easier for users to access these tokens.
- 🧩**Better than having individual teams create it**: If each team creates its own ERC1155 bridge, this will lead to fragmentation and a lack of standardization. Having a single, well-maintained ERC1155 bridge on Optimism will be better for the community as a whole.

### Solution
This pull request adds a new set of smart contracts for an ERC1155 bridge. We followed the structure of the ERC721 and Standard bridges for this first draft implementation.

- **OptimismMintableERC1155**:  Remote representation for some ERC1155 that lives on another network. Gives minting and burning capabilities to a bridge contract.
- **IOptimismMintableERC1155**: Interface for contracts that are compatible with the OptimismMintableERC1155 standard.
- **ERC1155Bridge**: Base abstract contract for the L1 and L2 ERC1155 bridges that includes the shared logic to pass messages to the other network using the messenger contract.
- **L1ERC1155Bridge**: Escrow for ERC1155 tokens deposited into L2. Works together with the L2 ERC1155 bridge to make it possible to transfer ERC1155 tokens from Ethereum to Optimism.
- **L2ERC1155Bridge**: Minter for new OptimismMintableERC1155 tokens when it hears about deposits into the L1ERC1155bridge. Also, acts as a burner for tokens being withdrawn. 

In case the solution is accepted, we can continue iterating on top of this: first obtaining some feedback, comments, and recommendations and then completing the pending tasks such as:
- Add support for ERC1155 batch functionality into the bridge contracts
- Perform internal review for improvements/fixes
- Add required tests
- Add/update docs
- Add deploy scripts
- Any other required tasks

### DIFF
--- a/packages/contracts-bedrock/contracts/L1/L1ERC1155Bridge.sol
+++ b/packages/contracts-bedrock/contracts/L1/L1ERC1155Bridge.sol
@@ -1,0 +1,109 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { IERC1155 } from "@openzeppelin/contracts/token/ERC1155/IERC1155.sol";
+import { ERC1155Holder } from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
+import { L2ERC1155Bridge } from "../L2/L2ERC1155Bridge.sol";
+import { ERC1155Bridge } from "../universal/ERC1155Bridge.sol";
+import { Semver } from "../universal/Semver.sol";
+
+/// @title L1ERC1155Bridge
+/// @notice The L1 ERC1155 bridge is a contract which works together with the L2 ERC1155 bridge to
+///         make it possible to transfer ERC1155 tokens from Ethereum to Optimism. This contract
+///         acts as an escrow for ERC1155 tokens deposited into L2.
+contract L1ERC1155Bridge is ERC1155Bridge, ERC1155Holder, Semver {
+    /// @notice Mapping of L1 token to L2 token to type ID to deposits, indicating the amount of L1 tokens
+    ///         by type ID deposited for L2 tokens.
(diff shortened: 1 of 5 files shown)


tynes:
Thank you for implementing this. You will need to make a post on the gov forum and get support to have a feature like this "officially" supported. Note that it's permissionless to create a bridge, ie anybody can make an erc20 bridge with a shorter fault proof window but it'll create fungibility issues and be more risky.

Also are people going to be bridging Uniswap LP tokens? Idk if that example really makes sense unless there is some sort of cross domain liquidity thing where you can add liquidity on one domain and remove on another

You can find the gov forum at https://gov.optimism.io

patitonar:
Thanks for the reply. I think it makes sense to discuss this on the governance forum. We will work on a proposal.

PR #6235, opened 2023-07-06
ci: move devnet test to make target

jyellick:
When encountering devnet test failures in CI, they're challenging to resolve and reproduce locally because the commands are embedded into the CI configuration and not the Makefile.

This PR simply extracts the commands from the CI steps and embeds them into the devnet python module, allowing for the python script to be invoked with a new `--test` flag to execute those steps.  It also adds a `devnet-test` make target to allow simple invocation of the script in test mode.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -989,17 +989,8 @@ jobs:
           name: Bring up the stack
           command: make devnet-up
       - run:
-          name: Check L2 config
-          command: go run cmd/check-l2/main.go --l2-rpc-url http://localhost:9545 --l1-rpc-url http://localhost:8545
-          working_directory: op-chain-ops
-      - run:
-          name: Deposit ERC20 through the bridge
-          command: timeout 8m npx hardhat deposit-erc20 --network devnetL1 --l1-contracts-json-path ../../.devnet/addresses.json
-          working_directory: packages/sdk
-      - run:
-          name: Deposit ETH through the bridge
-          command: timeout 8m npx hardhat deposit-eth --network devnetL1 --l1-contracts-json-path ../../.devnet/addresses.json
-          working_directory: packages/sdk
+          name: Test the stack
+          command: make devnet-test
       - run:
           name: Dump op-node logs
           command: |
--- a/bedrock-devnet/devnet/__init__.py
+++ b/bedrock-devnet/devnet/__init__.py
@@ -18,6 +18,7 @@ pjoin = os.path.join
 parser = argparse.ArgumentParser(description='Bedrock devnet launcher')
 parser.add_argument('--monorepo-dir', help='Directory of the monorepo', default=os.getcwd())
 parser.add_argument('--allocs', help='Only create the allocs and exit', type=bool, action=argparse.BooleanOptionalAction)
+parser.add_argument('--test', help='Tests the deployment, must already be deployed', type=bool, action=argparse.BooleanOptionalAction)
 
 log = logging.getLogger()
 
@@ -57,6 +58,8 @@ def main():
     ops_bedrock_dir = pjoin(monorepo_dir, 'ops-bedrock')
     deploy_config_dir = pjoin(contracts_bedrock_dir, 'deploy-config'),
     devnet_config_path = pjoin(contracts_bedrock_dir, 'deploy-config', 'devnetL1.json')
+    ops_chain_ops = pjoin(monorepo_dir, 'op-chain-ops')
+    sdk_dir = pjoin(monorepo_dir, 'packages', 'sdk')
 
     paths = Bunch(
       mono_repo_dir=monorepo_dir,
@@ -68,6 +71,8 @@ def main():
       devnet_config_path=devnet_config_path,
       op_node_dir=op_node_dir,
       ops_bedrock_dir=ops_bedrock_dir,
+      ops_chain_ops=ops_chain_ops,
+      sdk_dir=sdk_dir,
       genesis_l1_path=pjoin(devnet_dir, 'genesis-l1.json'),
       genesis_l2_path=pjoin(devnet_dir, 'genesis-l2.json'),
       allocs_path=pjoin(devnet_dir, 'allocs-l1.json'),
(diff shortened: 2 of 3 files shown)


bedrock-devnet/devnet/__init__.py
```diff
@@ -43,13 +46,20 @@ def main():
...
       genesis_l1_path=pjoin(devnet_dir, 'genesis-l1.json'),
       genesis_l2_path=pjoin(devnet_dir, 'genesis-l2.json'),
       addresses_json_path=pjoin(devnet_dir, 'addresses.json'),
       sdk_addresses_json_path=pjoin(devnet_dir, 'sdk-addresses.json'),
       rollup_config_path=pjoin(devnet_dir, 'rollup.json')
     )
 
+    if args.test:
+      log.info('Testing deployed devnet')
+      devnet_test(paths)
```
tynes, line 86: The idea here is that this test command is independent of bringing up the stack?

.circleci/config.yml
```diff
@@ -924,17 +924,9 @@ jobs:
...
-                working_directory: op-chain-ops
-            - run:
-                name: Deposit ERC20 through the bridge
-                command: timeout 8m npx hardhat deposit-erc20 --network devnetL1 --l1-contracts-json-path ../../.devnet/sdk-addresses.json
-                working_directory: packages/sdk
-            - run:
-                name: Deposit ETH through the bridge
-                command: timeout 8m npx hardhat deposit-eth --network devnetL1 --l1-contracts-json-path ../../.devnet/sdk-addresses.json
-                working_directory: packages/sdk
+                name: Test the stack
```
tynes: This doesn't look like it's running in CI, not sure if it's a permissions issue

tynes (commented):
Generally in favor of the idea, not sure how I feel about the approach but could be open to it

bedrock-devnet/devnet/__init__.py
jyellick: Correct.  To run the test manually, you'd do something like:

```
make devnet-up-deploy
make devnet-test
```

I'm not sure how everyone is using the `devnet` targets, but it seems plausible that some users are running other loads against the devnet, so I was wary of including the test execution with bringing the devnet up or down.

.circleci/config.yml
jyellick: I'm certainly not a circle CI expert, but it did execute in the source repo https://app.circleci.com/pipelines/github/bobanetwork/v3-anchorage/97/workflows/8f93a8dd-291f-4452-ad14-c47411905136/jobs/6089 so I believe the definition should be alright.

jyellick:
> Generally in favor of the idea, not sure how I feel about the approach but could be open to it

I'm certainly happy to rework in some other way.  As best as I could tell, the assorted build scripts are being migrated to python (vs. bash), so I tried to keep everything in the python, even though it's simply executing commands.

If there's a different preferred approach, please let me know.  My goal here is primarily to allow developers to iterate while running tests without having to push it out and wait on CI.

tynes:
> > Generally in favor of the idea, not sure how I feel about the approach but could be open to it
> 
> I'm certainly happy to rework in some other way. As best as I could tell, the assorted build scripts are being migrated to python (vs. bash), so I tried to keep everything in the python, even though it's simply executing commands.
> 
> If there's a different preferred approach, please let me know. My goal here is primarily to allow developers to iterate while running tests without having to push it out and wait on CI.

I am generally on board with this change after thinking about it more

ajsutton (commented):
Love this change - moving stuff from `.circleci/config.yml` into makefiles is a long term goal.  Ideally we'd also move the `devnet (with genesis contracts)` tests as well, though that could be a separate PR as this is definitely an improvement.

If we do move the genesis tests, it would be great to remove the duplication as well so `make devnet-test` could just work for both genesis and deploy modes. I think the only difference is if `--l1-contracts-json-path` is specified or not so possibly it could switch based on whether that file exists or not.

bedrock-devnet/devnet/__init__.py
trianglesphere: you would have `devnet-up-deploy` be a pre-requisite for `devnet-test` if you wanted `devnet-test` to automatically bring it up. The only potential issue is that it might always restart the devnet.

jyellick: Right.  Conceivably, we could have multiple targets if we wanted to allow both, say `devnet-test-only` and `devnet-test` which required the `devnet-up-deploy` followed by `devnet-test-only`.  If there's some consensus, I'm happy to modify.

jyellick:
I've updated the `--test` path of the python to add or omit the sdk address path depending on whether the file exists.  Now, things should work both for `make devnet-up` and `make devnet-up-deploy` -- I've also updated the circle CI definition to use it in both paths.

jyellick:
Is there anything preventing this from merging?  Happy to make any desired changes if so.

jyellick:
Rebased to try to keep this current (though no conflicts existed).  Any updates from the maintainers? @tynes @ajsutton ?

.circleci/config.yml
```diff
@@ -989,22 +989,8 @@ jobs:
...
-          working_directory: packages/sdk
-      - run:
-          name: Deposit ETH through the bridge
-          command: timeout 8m npx hardhat deposit-eth --network devnetL1 --l1-contracts-json-path ../../.devnet/addresses.json
-          working_directory: packages/sdk
-      - run:
-          name: Dump op-node logs
-          command: |
-            docker logs ops-bedrock-op-node-1 || echo "No logs."
-          when: on_fail
```
ajsutton: This piece doesn't seem to be in the python script so should probably stay here.

ajsutton (approved):
LGTM, other than I think we want to keep the capturing of op-node logs. I'm not an infra reviewer though.

ajsutton:
Oh turns out I am an infra reviewer. :) @jyellick Would you mind restoring the capturing of op-node logs and then I'll get this merged.

Thank you so much for your incredible patience - we've been really slow on a bunch of your PRs.

.circleci/config.yml
jyellick: Oops! Good catch.  I hadn't originally removed this, but looks like it got clobbered in one of the rebases.  Fixed.

ajsutton:
@jyellick The world is apparently out to get us today - the devnet build failed because of a network error downloading node. Can you kick it to rerun and then I think we should be good to go. It won't let me rerun because it look like its running in your circle ci org.

jyellick:
Thanks @ajsutton, looks like the re-run fixed things, very happy to see this merge!

PR #6240, opened 2023-07-10
contracts-bedrock: remove glob dependency

tynes:
**Description**

The script that ensures that all tests have the correct naming scheme used an external package named glob. This package was not actually necessary, altho it was convenient to use. This commit refactors the script to not use the glob dependency. It also now logs all of the files it successfully processes, so that we can be sure that it is running on all of the expected files. It uses the `IS_TEST` getter in the abi to determine if something is a test, which comes from forge-std `Test` contract. It also now uses `forge config --json` to parse the `out` location for the artifacts, so that it can be changed without needing to update the script.


refcell (approved):
LGTM

PR #6243, opened 2023-07-10
op-node: Engine P2P sync 

ImTei:
Good morning! This is Tei from Test in Prod. This PR contains **two topics:**

- Triggering the **Execution Engine’s P2P sync** for a faster sync and [Happy-path sync](https://github.com/ethereum-optimism/optimism/blob/develop/specs/exec-engine.md#happy-path-sync).
- **Fixing UX issues** caused by P2P Sync (distance from unsafe to safe is too far).

And there is another related [PR](https://github.com/ethereum-optimism/op-geth/pull/105) in op-geth repository to support snap sync.

### Specs

- New flag added
    - `--l2.engine-p2p.enabled=false|true`
    - default: `false`
    - will be stored in `sync.Config.EngineP2PEnabled` struct, and passed to `EngineQueue`
- New fields added in `EngineQueue`
    - `syncCfg sync.Config`
        - Stores `EngineP2PEnabled` and `SkipSanityCheck`
    - `engineSyncTarget eth.L2BlockRef`
        - Target L2 block the engine is currently syncing to.
        - If the engine p2p sync is enabled, it can be different with `unsafeHead`. Otherwise, it must be the same with `unsafeHead`.
        - Added to `SyncStatus`
- If `EngineQueue` is reset, set `engineSyncTarget` to unsafe head.
- If the `EngineP2PEnabled` flag is `true`, `EngineQueue.tryNextUnsafePayload()` should behave as follows:
    - Do not check if the current unsafe head matches the next unsafe payload’s parent when the client receives the next unsafe payload.
    - Allow `SYNCING` and `ACCEPTED` status from `engine_newPayloadV1`
    - Allow `SYNCING` status from `engine_forkchoiceUpdatedV1`
    - If `engine_forkchoiceUpdatedV1` returns `SYNCING` or `VALID`, update `eq.engineSyncTarget` to the new unsafe block.
    - If `engine_forkchoiceUpdatedV1` returns `VALID`, update `eq.unsafeHead` to the new unsafe block.
- If the `eq.unsafeHead` and `eq.engineSyncTarget` are not matched, consider the engine currently syncing.
- If the engine is syncing, `EngineQueue` does not do anything except notify new unsafe payloads to the engine.
    - It just returns `EngineP2PSyncing` error if there’s no unsafe payload in the queue. This error makes the pipeline goes idle.
- `EngineQueue` tries to apply the next unsafe head before applying the next safe attributes in `eq.Step()` function.
    - Current implementations prioritize safe sync over unsafe sync. So engine queue does not try to apply the next unsafe payload until derivates all batches from the current L1 chain.
    - After the engine P2P sync is done, there should be a very long gap between the unsafe and safe heads. We have to wait a very long time to consolidate all blocks in the gap by L1 derivation, and then we can continue unsafe sync.
    - So we need this change to advance the unsafe head even while L1 derivation is going on in the background.
- If `eq.tryNextUnsafePayload()` returns an EOF error, do not return immediately, and continue remaining tasks in `Step()` , such as safe sync.
    - EOF error is returned when all the below conditions are satisfied:
        - Engine P2P sync is disabled
        - The next unsafe payload is at a higher number than the current unsafe head.
        - The unsafe payload does not build upon the current unsafe head
    - So the EOF error means the engine queue must advance the chain head via L1 derivation.

### Test

- Added test cases in op-e2e:
    - Test unsafe sync when engine P2P is disabled.
    - Test unsafe sync when engine P2P is enabled. P2P between execution engines could not be tested in op-e2e. so it checks if only engineSyncTarget is advanced.
- Actual P2P sync can be tested in Hive. But Hive must be updated to use the latest version of op-node before implementing the engine P2P sync test.
- Engine P2P sync test on real chains (connected to one static peer)
    - OP Goerli
        - Snap sync: 2h20m
        - Full sync: 6h40m
    - OP Mainnet
        - Snap sync: 2h20m
        - Full sync: 12h50m

## Fixing UX Issues

This part might be better to open up as another PR. Please let me know if you prefer that way!

### Background

- The safe head is too far behind the unsafe head when the engine p2p sync is finished (~7M blocks in Goerli).
- op-node fails to launch when the user restarts the process after the p2p sync in the current scheme.
    - `reorg is too deep` error comes out while resetting the pipeline.
- The following descriptions explain the details & solution 👇.

### Details & Solution

- op-node resets the derivation pipeline upon launch to find the sync starting point. [Spec](https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md#finding-the-sync-starting-point).
- Finding the sync starting point is implemented in `sync.FindL2Heads()` function.
- `FindL2Heads()` function has two purposes: Finding unsafe and safe heads to start the sync.
- `FindL2Heads()` function has one main loop as follows:
    - Traverse blocks backward from the current unsafe block
    - Find the unsafe starting point.
    - Continues traversing back to find the safe starting point.
    - It checks how far the iteration goes from the initial unsafe head to calculate the L2 reorg depth. If the reorg is too deep, raise `reorg is too deep` error.
        - **Problem:** It calculates reorg depth in every iteration, even if there’s no L2 reorg.
        - **Solution:** Calculating reorg depth should be done only if the unsafe starting point is updated to an older block, which causes L2 reorg.

### Remaining issues

- There’s another UX-related issue though the reorg depth issue is resolved: pipeline reset takes a too long time if the safe and unsafe heads are very far apart.
    - Reason: `FindL2Heads()` iterates every block between unsafe and safe heads to find the safe starting point, which delays the process starts. We speculated intentions for the iteration:
        - Even if we found an unsafe starting point candidate, one of its ancestors may have a non-candidate L1 origin. We should update the unsafe starting point to the older block.
            - However, we found conflicts in the code—[L197](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/sync/start.go#L197-L199) allows the update, but [L232](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/sync/start.go#L231-L234) doesn’t.
        - The iteration should not only check the L1 origin but also other values (i.e. if L1 origins are contiguous, if sequence numbers are right)
    - We implemented `--skip-sanity-check` flag to avoid the issue and accelerate the process start speed.

### DIFF
--- a/op-node/flags/flags.go
+++ b/op-node/flags/flags.go
@@ -221,6 +221,13 @@ var (
 		Required: false,
 		Value:    false,
 	}
+	SkipSanityCheck = &cli.BoolFlag{
+		Name:     "skip-sanity-check",
+		Usage:    "Skip chain sanity check on pipeline reset",
+		EnvVars:  prefixEnvVars("SKIP_SANITY_CHECK"),
+		Required: false,
+		Value:    false,
+	}
 )
 
 var requiredFlags = []cli.Flag{
@@ -260,6 +267,7 @@ var optionalFlags = []cli.Flag{
 	BackupL2UnsafeSyncRPC,
 	BackupL2UnsafeSyncRPCTrustRPC,
 	L2EngineP2PEnabled,
+	SkipSanityCheck,
 }
 
 // Flags contains the list of configuration options available to the binary.
--- a/op-node/flags/flags.go
+++ b/op-node/flags/flags.go
@@ -214,17 +214,18 @@ var (
 		EnvVars:  prefixEnvVars("L2_BACKUP_UNSAFE_SYNC_RPC_TRUST_RPC"),
 		Required: false,
 	}
-	L2EngineP2PEnabled = &cli.BoolFlag{
-		Name:     "l2.engine-p2p.enabled",
+	L2EngineSyncEnabled = &cli.BoolFlag{
+		Name:     "l2.engine-sync",
 		Usage:    "Enables or disables execution engine P2P sync",
-		EnvVars:  prefixEnvVars("L2_ENGINE_P2P_ENABLED"),
+		EnvVars:  prefixEnvVars("L2_ENGINE_SYNC_ENABLED"),
 		Required: false,
 		Value:    false,
 	}
-	SkipSanityCheck = &cli.BoolFlag{
-		Name:     "skip-sanity-check",
-		Usage:    "Skip chain sanity check on pipeline reset",
-		EnvVars:  prefixEnvVars("SKIP_SANITY_CHECK"),
+	SkipSyncStartCheck = &cli.BoolFlag{
+		Name: "l2.skip-sync-start-check",
+		Usage: "Skip sanity check of consistency of L1 origins of the unsafe L2 blocks when determining the sync-starting point. " +
+			"This defers the L1-origin verification, and is recommended to use in when utilizing l2.engine-sync",
+		EnvVars:  prefixEnvVars("L2_SKIP_SYNC_START_CHECK"),
 		Required: false,
 		Value:    false,
 	}
@@ -266,8 +267,8 @@ var optionalFlags = []cli.Flag{
 	HeartbeatURLFlag,
 	BackupL2UnsafeSyncRPC,
 	BackupL2UnsafeSyncRPCTrustRPC,
-	L2EngineP2PEnabled,
-	SkipSanityCheck,
+	L2EngineSyncEnabled,
+	SkipSyncStartCheck,
 }
 
 // Flags contains the list of configuration options available to the binary.
--- a/op-node/flags/flags.go
+++ b/op-node/flags/flags.go
@@ -214,6 +214,13 @@ var (
 		EnvVars:  prefixEnvVars("L2_BACKUP_UNSAFE_SYNC_RPC_TRUST_RPC"),
 		Required: false,
 	}
+	L2EngineP2PEnabled = &cli.BoolFlag{
+		Name:     "l2.engine-p2p.enabled",
+		Usage:    "Enables or disables execution engine P2P sync",
+		EnvVars:  prefixEnvVars("L2_ENGINE_P2P_ENABLED"),
+		Required: false,
+		Value:    false,
+	}
 )
 
 var requiredFlags = []cli.Flag{
@@ -252,6 +259,7 @@ var optionalFlags = []cli.Flag{
 	HeartbeatURLFlag,
 	BackupL2UnsafeSyncRPC,
 	BackupL2UnsafeSyncRPCTrustRPC,
+	L2EngineP2PEnabled,
 }
 
 // Flags contains the list of configuration options available to the binary.
--- a/op-node/rollup/derive/pipeline.go
+++ b/op-node/rollup/derive/pipeline.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/ethereum-optimism/optimism/op-node/eth"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum-optimism/optimism/op-node/rollup/sync"
 )
 
 type Metrics interface {
@@ -46,6 +47,7 @@ type EngineQueueStage interface {
 	Finalized() eth.L2BlockRef
 	UnsafeL2Head() eth.L2BlockRef
 	SafeL2Head() eth.L2BlockRef
+	EngineSyncTarget() eth.L2BlockRef
 	Origin() eth.L1BlockRef
 	SystemConfig() eth.SystemConfig
 	SetUnsafeHead(head eth.L2BlockRef)
@@ -75,7 +77,7 @@ type DerivationPipeline struct {
 }
 
 // NewDerivationPipeline creates a derivation pipeline, which should be reset before use.
-func NewDerivationPipeline(log log.Logger, cfg *rollup.Config, l1Fetcher L1Fetcher, engine Engine, metrics Metrics) *DerivationPipeline {
+func NewDerivationPipeline(log log.Logger, cfg *rollup.Config, l1Fetcher L1Fetcher, engine Engine, metrics Metrics, syncCfg *sync.Config) *DerivationPipeline {
 
 	// Pull stages
 	l1Traversal := NewL1Traversal(log, cfg, l1Fetcher)
@@ -89,7 +91,7 @@ func NewDerivationPipeline(log log.Logger, cfg *rollup.Config, l1Fetcher L1Fetch
 	attributesQueue := NewAttributesQueue(log, cfg, attrBuilder, batchQueue)
 
 	// Step stages
-	eng := NewEngineQueue(log, cfg, engine, metrics, attributesQueue, l1Fetcher)
+	eng := NewEngineQueue(log, cfg, engine, metrics, attributesQueue, l1Fetcher, syncCfg)
 
 	// Reset from engine queue then up from L1 Traversal. The stages do not talk to each other during
 	// the reset, but after the engine queue, this is the order in which the stages could talk to each other.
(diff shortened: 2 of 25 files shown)


op-node/flags/flags.go
```diff
@@ -214,6 +214,20 @@ var (
...
+		Required: false,
+		Value:    false,
+	}
+	SkipSanityCheck = &cli.BoolFlag{
+		Name:     "skip-sanity-check",
+		Usage:    "Skip chain sanity check on pipeline reset",
+		EnvVars:  prefixEnvVars("SKIP_SANITY_CHECK"),
+		Required: false,
+		Value:    false,
+	}
```
protolambda:
```suggestion
	L2EngineSyncEnabled = &cli.BoolFlag{
		Name:     "l2.engine-sync",
		Usage:    "Enables or disables execution engine P2P sync",
		EnvVars:  prefixEnvVars("L2_ENGINE_SYNC_ENABLED"),
		Required: false,
		Value:    false,
	}
	SkipSyncStartCheck = &cli.BoolFlag{
		Name:     "l2.skip-sync-start-check",
		Usage:    "Skip sanity check of consistency of L1 origins of the unsafe L2 blocks when determining the sync-starting point. This defers the L1-origin verification, and is recommended to use in when utilizing l2.engine-sync",
		EnvVars:  prefixEnvVars("L2_SKIP_SYNC_START_CHECK"),
		Required: false,
		Value:    false,
	}
```
I would like to rename these for clarity to the user.
And then reflect the renamed variables in the usages in the code.

op-node/rollup/sync/config.go
```diff
@@ -0,0 +1,8 @@
+package sync
+
+type Config struct {
+	// EngineP2PEnabled is true when the EngineQueue can trigger execution engine P2P sync.
+	EngineP2PEnabled bool `json:"engine_p2p_enabled"`
+	// SkipSanityCheck is true when the EngineQueue does not do sanity check on pipeline reset.
+	SkipSanityCheck bool `json:"skip_sanity_check"`
```
protolambda:
```suggestion
	// EngineSync is true when the EngineQueue can trigger execution engine P2P sync.
	EngineSync bool `json:"engine_sync"`
	// SkipSyncStartCheck skip the sanity check of consistency of L1 origins of the unsafe L2 blocks when determining the sync-starting point. This defers the L1-origin verification, and is recommended to use in when utilizing l2.engine-sync
	SkipSyncStartCheck bool `json:"skip_sync_start_check"`
```

op-node/rollup/derive/pipeline.go
```diff
@@ -199,6 +205,8 @@ func (dp *DerivationPipeline) Step(ctx context.Context) error {
 	if err := dp.eng.Step(ctx); err == io.EOF {
 		// If every stage has returned io.EOF, try to advance the L1 Origin
 		return dp.traversal.AdvanceL1Block(ctx)
+	} else if err == EngineP2PSyncing {
```
protolambda: using `errors.Is` to check the error would be preferred here.

op-node/rollup/driver/state.go
```diff
@@ -313,8 +313,8 @@ func (s *Driver) eventLoop() {
 			s.log.Debug("Derivation process step", "onto_origin", s.derivation.Origin(), "attempts", stepAttempts)
 			err := s.derivation.Step(context.Background())
 			stepAttempts += 1 // count as attempt by default. We reset to 0 if we are making healthy progress.
-			if err == io.EOF {
-				s.log.Debug("Derivation process went idle", "progress", s.derivation.Origin())
+			if err == io.EOF || (err != nil && errors.Is(err, derive.EngineP2PSyncing)) {
+				s.log.Debug("Derivation process went idle", "progress", s.derivation.Origin(), "err", err)
 				stepAttempts = 0
 				s.metrics.SetDerivationIdle(true)
 				continue
```
protolambda: Instead of adding it to this `if`, can you instead add it as its own `else if` below, and clarify in the log message that it went idle because the engine is syncing? And then include the sync-target in the log?

protolambda (commented):
This looks great! I have a few suggestions to make the functionality/flags clearer to the user. Can you implement those?

ImTei:
@protolambda Thank you for the review! I have applied your suggestions :)

protolambda (approved):

ghost:
@ImTei @protolambda How do I enable this feature, seems like it is not documented yet?

ajsutton:
It's been enabled by default for a fair few releases now.  Just make sure you're using the latest release and then don't disable it or disable networking in op-node and it will work.

ghost:
@ajsutton The sync process on op-geth will not conflict with op-node right?

ghost:
@ajsutton I have found that if snap sync would be possible, it will require any kind of bootnodes or static nodes to connect and sync right? Is there any?

ajsutton:
@kaliubuntu0206 My apologies - there are two similarly named features and I was thinking of the wrong one.  p2p sync in op-node has been enabled by default to fill in any missed unsafe blocks.  This is actually the op-geth snap sync functionality which is _not_ yet enabled by default. The two systems don't interfere with each other.

To enable snap sync (this feature), for op-geth, set `--syncmode snap` and ensure you have not disabled discovery (ie don't have the `--nodiscover` flag).  For op-mainnet and op-goerli you will still need to start from the downloaded data directory, for other networks that weren't migrated from pre-bedrock state you just need to ensure op-geth is configured for the right network.

Then for op-node set `--l2.engine-sync=true`.

Bootnodes can be shared across networks so the default op-geth boot nodes should work fine. However it is worth noting that currently there are extremely few peers available on most OP stack networks so it may be difficult to find peers that you can snap sync from.

ghost:
@ajsutton Is there any bootnodes available for optimism mainnet to connect op-geth nodes? Also, I think it is important feature since no one wants to spend extra dollars for waiting the chain retrieval from L1

ajsutton:
Bootnodes are designed so that they can be shared between networks so you don't need different boot nodes and are most likely to find peers using the default boot nodes.  However there are just very very few peers available on OP stack networks currently.

ghost:
@ajsutton Yes but with my understanding in order to sync between L1 nodes you will need a bootnode to find each other, and I see none here https://github.com/ethereum-optimism/op-geth/blob/optimism/params/bootnodes.go

ghost:
Geth also has an history of very poor P2P discovery and many users aren't able to find even with the default bootnode exists. It would be great to have a list of optimism foundation op-geth nodes listed somewhere

ajsutton:
The Mainnet boot nodes are used by default: https://github.com/ethereum-optimism/op-geth/blob/ceb5f3201eb74e9c32594596ea15755ec94f99c1/cmd/utils/flags.go#L1098-L1124

ghost:
@ajsutton They are for `--networkid=1` with my knowledge. I mean, are they able to share op-geth nodes as well? I think they will disconnect immediately.

ghost:
They are also mainnet running nodes, you can see `EF-Bootnode` here https://ethstats.dev/

ajsutton:
Yes, boot nodes are designed to be able to be shared across networks and the sets of peers almost always winds up being shared across bootnodes for different networks just from the way the distributed hash table works for discovery.

That's why the network ID is checked by clients as part of finding peers - to quickly detect peers that are found from the boot nodes that aren't on the same chain. There are a number of other other checks that happen like that as part of the initial handshake process.

There are very very few peers available on OP stack networks currently, you will see a lot of peers being immediately connected as irrelevant and it may take a long time to find ones that are useful because the vast majority of OP stack nodes currently are configured to disable the peer to peer networking.

ghost:
@ajsutton I see, since I haven't really read the specification for P2P communication of ethereum nodes. I hope the feature could be enabled by default and the exclusive bootnodes could be provided for faster synchronization. ( And I think it is crucial since fetching blocks and transactions from Geth is very, very slow ). Thank you for answering my question

PR #6206, opened 2023-06-29
feat: port contracts periphery to contracts bedrock

tynes:
**Description**

Moves all of the contracts to a single repo so that the same tooling and standards can be applied to them. This PR migrates `contracts-periphery` to `contracts-bedrock`, including all of the tests.

PR #6246, opened 2023-07-10
fix(pnpm): Don't build node-hid

roninjin10:
- Skip building node-hid on pnpm install
- Fails for various reasons all the time in Op repo gateway repo and other repos
- Many versions of node-hid exist via recursive ethers related deps. The older ones are problematic
- Also problematic on macos
- Node-hid is not necessary to be built and is only a dev dependency for the libraries. Not necessary for our purposes just using it as an npm package we only need the production deps

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,11 @@
     "node": ">=16",
     "pnpm": ">=8"
   },
+  "pnpm": {
+    "neverBuiltDependencies": [
+      "node-hid"
+    ]
+  },
   "scripts": {
     "clean": "npx nx run-many --target=clean",
     "build": "npx nx run-many --target=build",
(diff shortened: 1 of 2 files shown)


roninjin10:
* develop
  * **PR #6246** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6246" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6246?utm_source=stack-comment).

roninjin10:
Failure looks like flaky networking but will check in on it more later

tynes:
we can prob remove ethers ledger completely since its no longer used at all anymore

tynes:
https://github.com/ethereum-optimism/optimism/pull/6206 removes ethers ledger dep

roninjin10:
I think this is not an issue anymore

PR #6248, opened 2023-07-10
ci: Check OpenMIPS binaries

Inphi:
Pre-built freestanding MIPS binaries are manually added to the repo whenever the cannon MIPS test vectors are updated.

This PR adds a ci step to assert that the binaries are up to date. The `maketests.py` script, which builds all MIPS programs, runs in under 2 seconds on my machine.


Inphi:
The CI fails since it relies on the updated ci-builder image. But subsequent PRs shouldn't fail once the ci-builder image is updated.

trianglesphere (commented):
missing the `capstone` module as a dependency in the CI builder. Can we test this with an updated CI image to make sure that it is working prior to merging?

Inphi:
> 

@trianglesphere  Done. Please take a look at https://github.com/ethereum-optimism/optimism/pull/6259.

Inphi:
@trianglesphere CI builds fine now with the updated ci-builder. Please take another look.

trianglesphere (approved):

PR #6250, opened 2023-07-10
feat: bi-directionally index StandardBridge contract events

hamdiallam:
closes ethereum-optimism/devx#24

Fully index deposits & withdrawals reading from the standardized bridge contract. This will enable us to also automatically index 3rdparty bridges that follow the same ABI with little extra work

_**Note:**_ For withdrawals, we have to handle the bedrock two-step withdrawal process. For any 3rdparty bridge, we'll solely rely on the finalization marker and will not support custom two-step withdrawal implementations. If a 3rdparty bridge wishes to have the same two-step withdrawal process, they should reuse the deployed`L2CrossDomainMessenger` in their standard bridge extension as the underlying message passer.

TODO:
- update `integration_tests` so that the bridge can be tested e2e without manual tests
- create a legacy adapter such we can index legacy deposit/withdrawals pre-bedrock
- If a deposit/withdrawal cant be indexed due to L1/L2 processor needing to catchup, we should instead pause the processor at that step and wait for the precondition rather than failing the entire batch and continually retrying which results in unnecessarily extra work
- Improve the abstraction for the processor. `processor/contract_events.go` is a step in this direction. The processor should be pipelined `ProcessHeaders -> ProcessContractEvents -> ... forward contract events to fill custom tables...`. Rather than figure out this abstraction upfront, lets unblock e2e functionality and let the right abstraction reveal itself

### DIFF
--- a/indexer/database/blocks.go
+++ b/indexer/database/blocks.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+
 	"github.com/google/uuid"
 
 	"gorm.io/gorm"
@@ -19,6 +21,15 @@ type BlockHeader struct {
 	ParentHash common.Hash `gorm:"serializer:json"`
 	Number     U256
 	Timestamp  uint64
+}
+
+func BlockHeaderFromHeader(header *types.Header) BlockHeader {
+	return BlockHeader{
+		Hash:       header.Hash(),
+		ParentHash: header.ParentHash,
+		Number:     U256{Int: header.Number},
+		Timestamp:  header.Time,
+	}
 }
 
 type L1BlockHeader struct {
--- a/indexer/database/blocks.go
+++ b/indexer/database/blocks.go
@@ -23,7 +23,7 @@ type BlockHeader struct {
 	Timestamp  uint64
 }
 
-func BlockHeaderFromHeader(header *types.Header) BlockHeader {
+func BlockHeaderFromGethHeader(header *types.Header) BlockHeader {
 	return BlockHeader{
 		Hash:       header.Hash(),
 		ParentHash: header.ParentHash,
--- a/indexer/database/blocks.go
+++ b/indexer/database/blocks.go
@@ -52,8 +52,9 @@ type LegacyStateBatch struct {
 }
 
 type OutputProposal struct {
-	OutputRoot          common.Hash `gorm:"primaryKey;serializer:json"`
-	L2BlockNumber       U256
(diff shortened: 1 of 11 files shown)


indexer/database/blocks.go
```diff
@@ -21,6 +23,15 @@ type BlockHeader struct {
 	Timestamp  uint64
 }
 
+func BlockHeaderFromHeader(header *types.Header) BlockHeader {
```
roninjin10:
```suggestion
func BlockHeaderFromGethHeader(header *types.Header) BlockHeader {
```
More explicit what this means?

indexer/processor/standard_bridge.go
```diff
@@ -0,0 +1,247 @@
...
+	erc20BridgeFinalizedEvents, err := _standardBridgeFinalizedEvents[bindings.L1StandardBridgeERC20BridgeFinalized](rawEthClient, events)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(ethBridgeFinalizedEvents, erc20BridgeFinalizedEvents...), nil
+}
+
+// parse out eth or erc20 bridge initiated events
+func _standardBridgeInitiatedEvents[BridgeEvent bindings.L1StandardBridgeETHBridgeInitiated | bindings.L1StandardBridgeERC20BridgeInitiated](
```
roninjin10, line 71: I like this straight forward imperative way of doing. This could be generalized to kinda registering handlers in future but this is easier to maintain to start

roninjin10 (approved):
lgtm

indexer/processor/l1_processor.go
```diff
@@ -194,63 +185,188 @@ func l1ProcessFn(processLog log.Logger, ethClient node.EthClient, l1Contracts L1
...
-		numLegacyStateBatches := len(legacyStateBatches)
-		if numLegacyStateBatches > 0 {
-			latestBatch := legacyStateBatches[numLegacyStateBatches-1]
-			latestL2Height := latestBatch.PrevTotal + latestBatch.Size - 1
-			processLog.Info("detected legacy state batches", "size", numLegacyStateBatches, "latest_l2_block_number", latestL2Height)
-		}
+		// Check if the L2Processor is behind or really has missed an event. We can compare against the
+		// OptimismPortal#ProvenWithdrawal on-chain mapping relative to the latest indexed L2 height
+		if withdrawal == nil {
+			var bridgeAddress common.Address
```
hamdiallam: Oops, need to update this

indexer/processor/standard_bridge.go
hamdiallam: yea i definitely agree

indexer/database/blocks.go
hamdiallam: sgtm

PR #6257, opened 2023-07-11
op-challenger: Main loop

trianglesphere:
**Description**

Main loop for the op-challenger. Wires everything together.
Lots of random changes to properly wire everything together.

**Tests**

Tested against defend of the alphabet script.
TODO: op-e2e 

**Metadata**

- Fixes CLI-4125


op-challenger/challenger.go
```diff
@@ -1,12 +1,50 @@
...
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/log"
 )
 
 // Main is the programmatic entry-point for running op-challenger
 func Main(logger log.Logger, cfg *config.Config) error {
+	client, err := ethclient.Dial(cfg.L1EthRpc)
+	if err != nil {
+		return fmt.Errorf("failed to dial L1: %w", err)
+	}
```
ajsutton, line 21: Can we log an issue to make the challenger handle starting even if it's L1 isn't available. Don't need it straight away but worth doing fairly early before it becomes hard to handle properly. k8s doesn't do ordering of pods well and it's really annoying to have programs exit because a dependency isn't fully started yet.

ajsutton (approved):
LGTM.

op-challenger/challenger.go
trianglesphere: Really just needs to use the pre-built dial with retry

PR #6260, opened 2023-07-11
feat: Add ts contract package for TS

roninjin10:
- Add a ts interface for contracts-bedrock
- use wagmi cli to do it
- I have contributed to wagmi cli in the past so we should have an easy time making custom plugins or upstreaming changes if we need them long term. The code is pretty simple for others to pick up as well and very well tested.
- We need this badly internally. We are currently copy pasting abis on the ecopod team and in the sdk because abis imported as JSON do not get correct type inference.  This fixes that and.
- This allows us to smartly configure contract addresses in our utils in a generated way
- this exports a react interface to our contracts, something optimism never had
- wagmi actions allow use in javascript with vanilla js or non-react frameworks
- the constants are exported from the main entrypoint
- we need to make entrypoints separate or else everybody would be forced to install react, wagmi etc. even if you are just using constants
- We want to mostly generate this kind of code for the best maintence cost. This should eventually completely replace the SDK if all goes well. Through plugins we can generate anything and then expose a superset to the API when necessary
- I added a custom plugin to generate an addresses mapping
- I added a custom plugin to append an eslint ignore
- We are generating and commiting instead of generating in the build. The generation is not hermetic (it has a time stamp) which means it would bust the nx cache everytime it built.  This also makes the build faster.
- I will add CI in a follow up pr that will check that the config is up to date. I'm working with wagmi to make this a wagmi cli option `npx wagmi check`
- The generated code has unnecessary duplication of the abis and the addresses. I will open issue to find a way to fix this as well
- I am building the package with tsup because it's fast and it builds cjs and esm correctly and handles multiple entrypoints with ease


roninjin10:
* develop
  * **PR #6260** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6260" target="_blank">Graphite</a>  👈
    * **PR #6298** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6298" target="_blank">Graphite</a> 
    * **PR #6267** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6267" target="_blank">Graphite</a> 
      * **PR #6407** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6407" target="_blank">Graphite</a> 
        * **PR #6301** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6301" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6260?utm_source=stack-comment).

.changeset/config.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://unpkg.com/@changesets/config@1.6.0/schema.json",
   "changelog": "@changesets/cli/changelog",
   "commit": false,
-  "linked": [],
+  "linked": [["contracts-bedrock", "contracts-ts"]],
```
roninjin10, line 6: Link these packages together so they always have the same version

packages/contracts-ts/README.md
```diff
@@ -0,0 +1,51 @@
+## Contracts TS
```
roninjin10, line 1: This package will end up replacing the SDK in many ways.  It's the contract interface to contracts-bedrock and for low maintenance cost it is mostly generated

.changeset/config.json
roninjin10: This makes sense only while this is mostly generated code.  Should that not become the case this would be less useful

nx.json
```diff
@@ -54,6 +54,9 @@
       "inputs": ["default", "testing", "^production"],
       "dependsOn": ["^build"]
     },
+    "generate": {
+      "dependsOn": ["^build"]
+    },
```
roninjin10, line 61: I'm purposefully not caching this because it's writing to src folder which is normally a folder meant to break the cache and it's not hermetic (includes a timestamp). It doesn't play nice with caching so from nx point of view this generation step is uncachable and we check in the changes rather than doing it in on every build.

packages/contracts-ts/CODE_GEN.md
```diff
@@ -0,0 +1,14 @@
...
+Summary - 
+- This package is generated from [contracts-bedrock](../contracts-bedrock/)
+- It's version is kept in sync with contracts bedrock via the [changeset config](../../.changeset/config.json) e.g. if contracts-bedrock is `4.2.0` this package will have the same version.
+
+## Code gen instructions
+
+To run the code gen run the `generate` script from [package.json](./package.json). Make sure node modules is installed. Use nx to guarantee that contracts-bedrock is fully built.
+
+```bash
+pnpm i && pnpm nx generate
```
roninjin10:
```suggestion
pnpm i && pnpm generate
```

packages/contracts-ts/CODE_GEN.md
```diff
@@ -0,0 +1,14 @@
+# Code gen
+
+Summary - 
+- This package is generated from [contracts-bedrock](../contracts-bedrock/)
+- It's version is kept in sync with contracts bedrock via the [changeset config](../../.changeset/config.json) e.g. if contracts-bedrock is `4.2.0` this package will have the same version.
+
+## Code gen instructions
+
+To run the code gen run the `generate` script from [package.json](./package.json). Make sure node modules is installed. Use nx to guarantee that contracts-bedrock is fully built.
```
roninjin10:
```suggestion
To run the code gen run the `generate` script from [package.json](./package.json). Make sure node modules is installed.
```

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+      "require": "./dist/react.cjs"
+    }
+  },
+  "files": [
+    "dist/",
+    "src/"
+  ],
+  "scripts": {
+    "build": "tsup",
+    "generate": "wagmi generate",
```
roninjin10:
```suggestion
    "generate": "nx build @eth-optimism/contracts-bedrock && wagmi generate",
```

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "test": "NODE_OPTIONS=--experimental-fetch vitest"
```
roninjin10: This needs experimental-fetch because fetch doesn't exist in node 16 without this flag. It's added as non experimental in node 18.

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "./actions": {
+      "types": "./src/actions.ts",
+      "import": "./dist/actions.js",
+      "require": "./dist/actions.cjs"
+    },
+    "./react": {
+      "types": "./src/react.ts",
+      "import": "./dist/react.js",
+      "require": "./dist/react.cjs"
+    }
```
roninjin10: These need to be seperate exports or else someone trying to use our ABI or constants would be forced to install react for example even if they aren't using react

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+  "name": "@eth-optimism/contracts-ts",
+  "version": "0.15.0",
+  "description": "TypeScript interface for Contracts Bedrock",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ethereum-optimism/optimism.git",
+    "directory": "packages/contracts-ts"
+  },
+  "homepage": "https://optimism.io",
```
roninjin10, line 11: We are missing these fields in our other packages and should add them

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@eth-optimism/contracts-ts",
+  "version": "0.15.0",
```
roninjin10, line 3: This version matches contracts-bedrock

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@eth-optimism/contracts-ts",
```
roninjin10, line 2: Better name?  what about `@eth-optimism/ts`?

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+      "types": "./src/index.ts",
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
+    },
+    "./actions": {
+      "types": "./src/actions.ts",
+      "import": "./dist/actions.js",
+      "require": "./dist/actions.cjs"
+    },
+    "./react": {
```
roninjin10, line 27: React would be better ux and more discoverable if it was in it's own package. We can consider moving it later

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "wagmi": ">1.0.0"
+  },
+  "peerDependenciesMeta": {
+    "wagmi": {
+      "optional": true
+    },
+    "@wagmi/core": {
+      "optional": true
+    }
+  },
```
roninjin10, line 76: These need to be optional because you can use this package just using the constants

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "@types/glob": "^8.1.0",
+    "@typescript-eslint/eslint-plugin": "^5.60.1",
+    "@typescript-eslint/parser": "^5.60.1",
+    "@wagmi/cli": "^1.3.0",
+    "abitype": "^0.9.0",
+    "ds-test": "github:dapphub/ds-test#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
+    "forge-std": "github:foundry-rs/forge-std#e8a047e3f40f13fa37af6fe14e6e06283d9a060e",
+    "glob": "^7.2.3",
+    "solhint": "^3.4.1",
+    "solhint-plugin-prettier": "^0.0.5",
```
roninjin10:
```suggestion
```
Unused

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+  },
+  "devDependencies": {
+    "@eth-optimism/contracts-bedrock": "workspace:0.15.0",
+    "@types/glob": "^8.1.0",
+    "@typescript-eslint/eslint-plugin": "^5.60.1",
+    "@typescript-eslint/parser": "^5.60.1",
+    "@wagmi/cli": "^1.3.0",
+    "abitype": "^0.9.0",
+    "ds-test": "github:dapphub/ds-test#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
+    "forge-std": "github:foundry-rs/forge-std#e8a047e3f40f13fa37af6fe14e6e06283d9a060e",
```
roninjin10:
```suggestion
```
Unused

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+  "scripts": {
+    "build": "tsup",
+    "generate": "wagmi generate",
+    "test": "NODE_OPTIONS=--experimental-fetch vitest"
+  },
+  "devDependencies": {
+    "@eth-optimism/contracts-bedrock": "workspace:0.15.0",
+    "@types/glob": "^8.1.0",
+    "@typescript-eslint/eslint-plugin": "^5.60.1",
+    "@typescript-eslint/parser": "^5.60.1",
```
roninjin10:
```suggestion
```
Unused

packages/contracts-ts/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "@wagmi/cli": "^1.3.0",
+    "abitype": "^0.9.0",
+    "ds-test": "github:dapphub/ds-test#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
+    "forge-std": "github:foundry-rs/forge-std#e8a047e3f40f13fa37af6fe14e6e06283d9a060e",
+    "glob": "^7.2.3",
+    "solhint": "^3.4.1",
+    "solhint-plugin-prettier": "^0.0.5",
+    "ts-node": "^10.9.1",
```
roninjin10:
```suggestion
```
Unused

packages/contracts-ts/CODE_GEN.md
```diff
@@ -0,0 +1,14 @@
...
+Summary - 
+- This package is generated from [contracts-bedrock](../contracts-bedrock/)
+- It's version is kept in sync with contracts bedrock via the [changeset config](../../.changeset/config.json) e.g. if contracts-bedrock is `4.2.0` this package will have the same version.
+
+## Code gen instructions
+
+To run the code gen run the `generate` script from [package.json](./package.json). Make sure node modules is installed.
+
+```bash
+pnpm i && pnpm generate
```
roninjin10, line 13: pnpm generate builds contracts bedrock if not already built first

packages/contracts-ts/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+[![codecov](https://codecov.io/gh/ethereum-optimism/optimism/branch/develop/graph/badge.svg?token=0VTG7PG7YR&flag=contracts-bedrock-tests)](https://codecov.io/gh/ethereum-optimism/optimism)
+
+ABI and Address constants + generated code from [@eth-optimism/contracts-bedrock/](../contracts-bedrock/) for use in TypeScript.
+
+Much of this package is generated. See [CODE_GEN.md](./CODE_GEN.md) for instructions on how to generate.
+
+#### @eth-optimism/contracts-ts
```
roninjin10, line 9: I'm working on fixing a bug with wagmi that is preventing me from autogenerating all the documentation

packages/contracts-ts/addresses.json
```diff
@@ -0,0 +1,86 @@
...
+    "5": "0x4200000000000000000000000000000000000018"
+  },
+  "SystemConfig": {
+    "1": "0x229047fed2591dbec1eF1118d64F7aF3dB9EB290",
+    "5": "0xAe851f927Ee40dE99aaBb7461C00f9622ab91d60"
+  },
+  "SystemDictator": {
+    "1": "0xB4453CEb33d2e67FA244A24acf2E50CEF31F53cB"
+  },
+  "SystemDictator_goerli": {
```
roninjin10, line 79: This seperation here between SystemDictator and SystemDictator_goerli happens whenever the code gen detects the abis don't match between networks

packages/contracts-ts/wagmi.config.ts
```diff
@@ -0,0 +1,315 @@
...
+  name: 'eslintIgnore',
+  run: async () => {
+    return {
+      prepend: `/* eslint-disable */`,
+      content: ``
+    }
+  },
+}
+
+export default defineConfig([{
```
roninjin10:
```suggestion
// @see https://wagmi.sh/cli
export default defineConfig([{
```

packages/contracts-ts/wagmi.config.ts
```diff
@@ -0,0 +1,316 @@
...
+  if (!contractConfig.name) {
+    throw new Error('Unable to identify the name of the contract at ' + artifactPath)
+  }
+  return contractConfig
+})
+
+/**
+ * Returns the contracts for the wagmi cli config
+ */
+export const getContractConfigs = () => {
```
roninjin10:
```suggestion
const getContractConfigs = () => {
```

packages/contracts-ts/addresses.json
```diff
@@ -0,0 +1,86 @@
+{
```
roninjin10, line 1: We are missing tons of contracts that we can find ways to add in follow up pr. Will likely add them with etherscan plugin https://wagmi.sh/cli/plugins/etherscan

packages/contracts-ts/package.json
hamdiallam: what if we call it `op-bindings-ts`?

hamdiallam (approved):

roninjin10:
@tynes mentioned wanting to be mindful of buildtime impact of new packages. Checking turned out to be a good idea because the build time performance here was massive and easily fixed.
Including the benchmarks and explanation of them here. If not interested skip all the way to the bottom for the outcome.
Some benchmarks
:fire::fire::fire:
pnpm build - tsup   - .55s
pnpm test   - vitest - 2s
pnpm lint   - prettier - 3.4s
:snail::snail::snail:
pnpm typecheck - tsc - 23s

Why fast?

build is fast because it uses tsup.
tsup uses esbuild.
Esbuild doesn't typecheck
Esbuild is built in golang
should use tsup everywhere IMO
Fixes source maps
Super fast
Supports es modules (we currently don't ship modern js only commonjs)

test is fast because it uses vite.
vite uses esbuild.
Esbuild doesn't typecheck
Esbuild is built in golang
Should use vitest everywhere IMO

prettier is fast-ish because it isn't typechecking and it's time scales linearlly with lines of code
Should use rome instead IMO
This is a hot take most people would say we shouldn't
Rome is built in rust and magnitudes faster
Rome replaces esbuild and prettier both at same time
This is a hot take and has a downside of most folks don't have rome configured in vscode and other editors

Why slow

typecheck is slow because:
we are generating thousands of lines of code
Those lines of code are all using wagmi
Wagmi uses abitype to infer contract type safety from abis is one of the most complicated type libraries ever and the type inference comes at a cost of TS doing more work
Typescript is generally really slow
Typescript is written in JavaScript

Action

Because typechecking isn't giving us much in this case (code is generated) I'm going to disable typechecker in CI

packages/contracts-ts/addresses.json
```diff
@@ -0,0 +1,86 @@
...
+  "OptimismMintableERC20Factory": {
+    "1": "0x4200000000000000000000000000000000000012",
+    "5": "0x4200000000000000000000000000000000000012",
+    "420": "0x4200000000000000000000000000000000000012"
+  },
+  "OptimismPortal": {
+    "1": "0xbEb5Fc579115071764c7423A4f12eDde41f106Ed",
+    "5": "0x5b47E1A08Ea6d985D6649300584e6722Ec4B1383"
+  },
+  "PortalSender": {
```
tynes, line 64: We don't care about this contract, it was part of the migration and will never be used again

tynes:
We would like to move to forge fmt for the contracts, it's basically instant where prettier takes 10s plus

packages/contracts-ts/package.json
roninjin10: Hmmm. I like this the best internally don't think it's a good external name though not having Optimism in the namespace

packages/contracts-bedrock/package.json
```diff
@@ -9,7 +9,9 @@
     "contracts/**/*.sol"
   ],
   "scripts": {
-    "bindings": "cd ../../op-bindings && make",
+    "bindings": "pnpm bindings:ts && pnpm bindings:go",
+    "bindings:ts": "nx generate @eth-optimism/contracts-ts",
```
refcell: Very nice, can we also make a pass-through script from the top-level package.json that runs this script or you prefer to just keep locally here?

packages/contracts-ts/package.json
refcell: I think `@eth-optimism/contracts-ts` is pretty clear and not that bad of a name tbh

packages/contracts-ts/package.json
refcell: change-set links ftw

packages/contracts-ts/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './constants'
```
refcell: Should we be cherry-picking constants so we don't accidentally expose un-released values?

packages/contracts-ts/wagmi.config.ts
```diff
@@ -0,0 +1,365 @@
+import { ContractConfig, defineConfig, Plugin } from '@wagmi/cli'
+import { actions, react } from '@wagmi/cli/plugins'
+import glob from 'glob'
+import { readFileSync, writeFileSync } from 'fs'
+import type { Abi, AbiFunction, Address } from 'abitype'
+import { isDeepStrictEqual } from 'util'
+
+/**
+ * Predeployed contract addresses
+ * In future it would be nice to have a json file in contracts bedrock be generated as source of truth
```
refcell, line 10: +1 on this - splitting this metadata out from the contracts-bedrock source adds more work on contracts-bedrock maintainers

packages/contracts-ts/package.json
nickbalestra: yeah contracts-ts is boring and clear enough

packages/contracts-ts/src/index.ts
nickbalestra: agree with @refcell  I don't dislike being more explicit on what we exports in here, also being this the main entry file, otherwise we just proxying constants.ts and not sure about the need of this file?

nickbalestra (approved):
Couple of small non blocking comments, but LGTM!

packages/contracts-ts/src/index.ts
roninjin10: Good point actually there is no need for this file to exist

packages/contracts-ts/wagmi.config.ts
roninjin10: Yea I'll figure this out in a follow up pr today that will also try to add the missing contract addresses

packages/contracts-bedrock/package.json
roninjin10: oohhh good idea

packages/contracts-ts/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+- All [wagmi actions](https://wagmi.sh/react/actions) for use in Vanilla JS or non react code
+
+```typescript
+import { readSystemConfig } from '@eth-optimism/contracts-ts/actions'
+console.log(await readSystemConfig())
+```
+
+#### See Also
+
```
annieke, line 62: this README is very brief and to the point, which may be what we want for now, but I see one day where this is the main TS package people use, then this will need more docs + examples

the temporary solution is we write a snazzy dev blog for it and use it as the "use case" and "why" docs, then extend the README to a full on intro if we see traction

roninjin10: Agreed what I want to do is generate the entire documentation with link to wagmi docs. I'm blocked atm from doing that by this bug https://github.com/wagmi-dev/wagmi/issues/2724

zhwrd (approved):

mslipper (approved):

packages/contracts-ts/wagmi.config.ts
```diff
@@ -0,0 +1,365 @@
...
+ * Keep this in sync with op-bindings/predeploys/addresses.go in meantime
+ */
+const predeployContracts = {
+  LegacyMessagePasser: {
+    address: '0x4200000000000000000000000000000000000000',
+    introduced: 'Legacy',
+    deprecated: true,
+    proxied: true,
```
tynes, line 18: Do we use these additional values?

tynes (approved):

packages/contracts-ts/wagmi.config.ts
roninjin10: Not yet I included them here to encourage the docs team to generate docs from this since the docs use them

packages/contracts-ts/addresses.json
roninjin10: Oh nice call will do a follow up pr to remove this

roninjin10: Adding a TODO.md to the top of the stack to not forget

PR #6264, opened 2023-07-12
feat(ctb): Persist init data on game creation

clabby:
## Overview

> **Note**
> WIP

Implements the init data population on `FaultDisputeGame` creation. `extraData` passed to the `DisputeGameFactory` will contain the [`BootInfo`](https://github.com/ethereum-optimism/optimism/blob/develop/op-program/client/boot.go#L22-L29) struct:
```go
type BootInfo struct {
    L1Head             common.Hash
    L2Head             common.Hash
    L2Claim            common.Hash
    L2ClaimBlockNumber uint64
    L2ChainConfig      *params.ChainConfig
    RollupConfig       *rollup.Config
}
```
which contains necessary bootstrap information for the off-chain challenge agents to run the `op-program`. This information is forwarded to the `PreimageOracle` contract and stored for when the `MIPS.sol` VM needs to access local data for the game.

Keys in the preimage oracle have several different types:
* `0` - Zero; Illegal.
* `1` - Local data, context specific
* `2` - Global keccak data, unspecific to context
* `3` - Global reserved; Future compatibility.

To support storing local keys in the `PreimageOracle` that are context specific, I've also overhauled the `PreimageOracle` contract to do several things:
1. Rather than having hard-coded mappings for whether a preimage part exists, its length, and the part itself, the oracle now uses a custom storage layout that determines these mappings slots as follows:
    1. `keccak256(type)` = `preimagePartOk` mapping for the type. Under the hood this is a 2d mapping of `key => partOffset => bool`.
    2. `keccak256(type | 1 << 8)` = `preimageLengths` mapping for the type. Under the hood this is a 1d mapping of `key => length`.
    3. `keccak256(type | 2 << 8)` = `preimageParts` mapping for the type. Under the hood this is a 2d mapping of `key => partOffset => part`.
2. To accomodate the context-specific requirements for the `1` key type, when a local key is mapped to data in the oracle, the address of the game that the data pertains to is placed in the `[1, 21)` bytes of the key when it is stored. When a key with type = `1` is read in `MIPS.sol`, the `msg.sender` of the `step` call (the `FaultDisputeGame` contract) is set in the key. Because no off chain actors will call `step` in `MIPS.sol` directly and because `MIPS.sol` can't do anything on its own, this is safe.

**Metadata**
Closes CLI-4158


packages/contracts-bedrock/contracts/cannon/PreimageOracle.sol
```diff
@@ -65,8 +140,129 @@ contract PreimageOracle {
...
+
+            storeInNestedMapping(key, partOffset, true, preimagePartOkSlot)
+            storeInNestedMapping(key, partOffset, part, preimagePartsSlot)
+            storeInMapping(key, size, preimageLengthsSlot)
+        }
+    }
+
+    // loadLocalBootData prepares the boot data for a game to be read by the local key,
+    // starting at the given offset, up to 32 bytes (clipped at boot data length, if out of data).
+    function loadLocalBootData(
```
Inphi, line 146: I misspoke earlier regarding the encoding of the `BootInfo`. The `BootInfo` used by the fault proof program isn't encoded per se, but rather it's the components of that struct that are encoded. However, this function loads a "boot data" buffer as a single pre-image, which isn't how the FPP will use it.

At startup, the fault proof program queries the pre-image oracle for each boot info component (rollup config, l1 head, etc) separately to construct an internal `BootInfo` configuration. So the BootData calldata here should consist of a list of _opaque_ pre-image tuples for each component rather than a single pre-image. Each tuple should also contain the local key index that the FPP will use to query the associated pre-image. Unlike `loadKeccak256PreimagePart`, we cannot derive the pre-image key as the local key mapping depends on the program. The user will have to supply the local key indices and their pre-images according to the semantics of the program.

The local pre-images here are opaque to be forwards-compatible with the program. We want to load local preimages generically to support changes to the fault proof program without updating the `PreimageOracle`. In the future, the program may require additional local pre-images for its `BootInfo`.

The `PreimageOracle` should be agnostic to the number and lengths of local pre-images loaded. It shouldn't even have a notion of "boot data" imo to emphasize this point.

Inphi (commented):
Would it be simpler to compute slots by hand only when dealing with local pre-images? So we have both regular Solidity mappings for keccak pre-images and yul for local pre-images.

clabby:
Closing in favor of a different approach.

PR #6277, opened 2023-07-12
Add EAS and SchemaRegistry contracts (and their go bindings) so that they are built as part of the build system

lbeder:
**Description**

This PR adds the EAS and the SchemaRegistry contracts (and their go bindings) so that they are built as part of the build system

**Additional context**

The contracts were ported almost entirely from https://github.com/ethereum-attestation-service/eas-contracts-upgradeable, which is the ERC1967 compliant variant of https://github.com/ethereum-attestation-service/eas-contracts. 

The following minor modifications were made:
- Importing the internal version of `Semver`
- Importing `@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol` instead of `@openzeppelin/contracts/utils/cryptography/EIP712.sol` since the (slightly) older of OZ is being used by the rest of the contracts.


packages/contracts-bedrock/contracts/EAS/EAS.sol
```diff
@@ -0,0 +1,853 @@
...
+            revert InvalidRegistry();
+        }
+
+        _schemaRegistry = registry;
+    }
+
+    /**
+     * @dev Initializes the contract and its parents.
+     */
+    function initialize() external initializer {
```
tynes: Why does this need an initialize function?

tynes: You should always call initialize in the constructor so that the impl cannot be initialized by an adversary. There is no storage set in the constructor unless something is set inside of the eip712 verifier but then the upgradable version is needed

tynes: It looks like the eip712 verifier has no storage based on https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.7.3/contracts/utils/cryptography/draft-EIP712.sol

Would like to confirm with the following command:

```
forge inspect EAS storageLayout
```

lbeder: Yes, the initializer isn't really needed here (since there is no storage initialization, neither in its constructor nor in one of its parents). We are following a convention that every implementation contract should have `initialize` in its interface, but perhaps we should just remove it from EAS and SchemaRegistry. Furthermore, since there are no assignment of privileges or storage initialization in its initializer - initializing it by an adversary doesn't mean anything (although, it's very common for people to report this is a super-duper vulnerability via Immunifi 😅).

I'll remove this code for now.

packages/contracts-bedrock/contracts/EAS/SchemaRegistry.sol
```diff
@@ -0,0 +1,95 @@
...
+    // The global mapping between schema records and their IDs.
+    mapping(bytes32 uid => SchemaRecord schemaRecord) private _registry;
+
+    // Upgrade forward-compatibility storage gap
+    uint256[MAX_GAP - 1] private __gap;
+
+    /**
+     * @dev Creates a new SchemaRegistry instance.
+     */
+    constructor() Semver(1, 0, 0) {}
```
tynes, line 28: Confused as to why this is `Initializable`  when no storage is set in the constructor

lbeder: See the previous comment re: EAS (i.e., it's just a convention), but in any case I'll remove this.

packages/contracts-bedrock/contracts/EAS/resolver/SchemaResolver.sol
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * @title A base resolver contract
+ */
+abstract contract SchemaResolver is ISchemaResolver, Semver {
+    error AccessDenied();
+    error InsufficientValue();
+    error NotPayable();
+
+    // The global EAS contract.
+    IEAS internal immutable _eas;
```
tynes, line 20: This can be a constant and the constant can be defined in `Predeploys.sol` so no need for a constructor argument

packages/contracts-bedrock/contracts/EAS/EAS.sol
```diff
@@ -0,0 +1,853 @@
...
+
+    // Upgrade forward-compatibility storage gap
+    uint256[MAX_GAP - 3] private __gap;
+
+    /**
+     * @dev Creates a new EAS instance.
+     *
+     * @param registry The address of the global schema registry.
+     */
+    constructor(ISchemaRegistry registry) Semver(1, 0, 0) EIP712Verifier("EAS", "1.0.0") {
```
tynes: The registry can be removed as an argument and the hardcoded predeploy constant can be used, see my comment in `SchemaRegistry`

packages/contracts-bedrock/contracts/EAS/resolver/SchemaResolver.sol
lbeder: The only downside is that it will make the interfaces much different than other deployments, while gas-wise, there's (almost) no difference. This can hinder projects that develop for other chains as well, so I'm not sure if it's a good fit here.

packages/contracts-bedrock/contracts/EAS/SchemaRegistry.sol
lbeder: I've removed the initializers. Please take a look.

packages/contracts-bedrock/contracts/EAS/EAS.sol
lbeder: I've removed the initializers. Please take a look.

lbeder:
I've removed the unnecessary initializers, added both of the contracts to the predeploys (and added custom natspec documentation with the predeployed addresses).

packages/contracts-bedrock/contracts/EAS/EAS.sol
tynes: We should def remove `initialize` if its a noop, its just a maintenance burden. You are right that not calling `initialize` in the constructor in this case doesn't cause security issues but its best practice in my opinion to always call `initialize` in the constructor in cases where `initialize` is required

packages/contracts-bedrock/contracts/EAS/EAS.sol
```diff
@@ -0,0 +1,853 @@
...
+     * @dev Refunds remaining ETH amount to the attester.
+     *
+     * @param remainingValue The remaining ETH amount that was not sent to the resolver.
+     */
+    function _refund(uint256 remainingValue) private {
+        if (remainingValue > 0) {
+            // Using a regular transfer here might revert, for some non-EOA attesters, due to exceeding of the 2300
+            // gas limit which is why we're using call instead (via sendValue), which the 2300 gas limit does not
+            // apply for.
+            payable(msg.sender).sendValue(remainingValue);
```
tynes, line 753: Using our safecall library is safer and cheaper than `sendValue` because it is vulnerable to returndata bombs, altho in this case there is no incentive to do so

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/121be5dd09caa2f7ce731f0806b0e14761023bd6/contracts/utils/Address.sol#L41

packages/contracts-bedrock/contracts/EAS/resolver/SchemaResolver.sol
tynes: Since this code lives in both the Optimism monorepo and your github repo, projects that are not OP Stack chains should use the code in your github repo. Using the hardcoded values is how we manage other predeploys and it will make deployment more simple

tynes (approved):

packages/contracts-bedrock/contracts/EAS/EAS.sol
```diff
@@ -0,0 +1,824 @@
...
+    error InvalidRevocation();
+    error InvalidRevocations();
+    error InvalidSchema();
+    error InvalidVerifier();
+    error Irrevocable();
+    error NotPayable();
+    error WrongSchema();
+
+    // The global schema registry.
+    ISchemaRegistry private immutable _schemaRegistry = ISchemaRegistry(Predeploys.SCHEMA_REGISTRY);
```
tynes: this should be a constant now instead of immutable

tynes (changes requested):
needs to make a change and rebuild the bindings

tynes:
Closing in favor of #6328

PR #6278, opened 2023-07-13
cannon: Fix unaligned preimage read in MIPS.sol

Inphi:
Fix a MIPS.sol bug where pre-image reads to unaligned memory creates invalid writes.


packages/contracts-bedrock/contracts/cannon/MIPS.sol
```diff
@@ -145,7 +145,7 @@ contract MIPS {
           if lt(space, datLen) { datLen := space } // if less space than data, shorten data
           if lt(a2, datLen) { datLen := a2 } // if requested to read less, read less
           dat := shr(sub(256, mul(datLen, 8)), dat) // right-align data
-          dat := shl(mul(add(sub(4, datLen), alignment), 8), dat) // position data to insert into memory word
+          dat := shl(mul(sub(sub(4, datLen), alignment), 8), dat) // position data to insert into memory word
```
Inphi: For example, when writing a one-byte 0xAA pre-image (repr as 0x000000AA since it's right-aligned on L147) to an address that's 1-byte misaligned, we want to left shift the pre-image buffer by across 16 bits.
If it's a 2-byte 0xAABB (0x0000AABB) buffer, then we left shift by 8 bits.
And so on.

The expression `sub(sub(4, datLen), alignment))` cannot be less than zero since `datLen = min(datLen, space)` and `space + alignment = 4` (L144).

clabby (approved):
LGTM; Needs rebase

Inphi:
@Mergifyio queue

PR #6280, opened 2023-07-13
op-e2e: add test option to utilize external eth clients for L2

jyellick:
**Description**

This change includes a new path to test op-erigon in the op-e2e tests in a minimally invasive way.  Almost all of the tests are unchanged, with some small structural changes to the setup to allow for utilization of a Geth or Erigon client as the L2 Ethereum client.

Because client diversity is so important to the Ethereum ecosystem, ensuring that clients other than Geth can interoperate with Optimism is an important task.  As the op-e2e tests are primarily testing the Optimism rollup protocol at a high level, nothing prevents these tests from applying to any Ethereum client capable of operating as an L2.  This set of changes should help these clients ensure they are truly implementing the Optimism protocol spec correctly which should benefit the ecosystem as a whole.

Although perhaps this work could have been discussed in an issue, hopefully having working code in a PR can help to focus the discussion.

If or how interoperability with other clients factors into the Optimism release process is an open question, and so actual execution of the op-e2e with Erigon is omitted in this change from CI.  However, by bringing these changes in, we can help ensure that it's possible to continue to develop other clients such as Erigon while leveraging the power of these e2e tests.

Possible extensions to this work could include providing shims for other Ethereum clients, or, for executing op-geth in an extra-process manner.

**Details**

More information will be added in the comments below, but, the basic approach taken is to use the existing HTTP/WS connections already utilized within the tests but instead of wiring them through IPC, they are connected to ephemeral ports provided by an external Erigon process.

Running Erigon as an external process rather than repeating the existing IPC mechanisms for Geth is done for a few different reasons.
* First and foremost, Erigon and Geth share many dependencies (and much of their code), it's seemingly impossible to build a single Go binary containing both op-geth and op-erigon.  So, the IPC approach is not (at least easily) possible.
* Secondly, by utilizing only the standardized RPC and WS mechanisms, the same approach can be generalized to other future Optimism compatible Ethereum clients.
* Finally, although the resulting external process test path is slower and more resource intensive than the Geth IPC path, when utilizing parallelism the test suite still successfully executes in approximately 120 seconds on my laptop, compared with about 90 seconds for the Geth tests.

To execute these tests simply run `make test-erigon` in the `op-e2e` directory.  Or, alternatively pass the `--erigon` flag to your `go test` command.

**Which op-erigon?**

This PR does not use the fork of erigon from https://github.com/testinprod-io/op-erigon but instead utilizes a fork that was developed separately but in parallel at https://github.com/bobanetwork/v3-erigon 

Having reviewed the testinprod-io version of op-erigon, it's a great piece of work, and there are actually some pieces that I like the implementation of better than the bobanetwork fork (for instance how some of the mining stages are handled) and there's also a lot of overlap with pieces that have been copied from the op-geth diff into each.  I'd actually be thrilled if the two forks could some day be combined or at least the diff between them minimized, but, that's a discussion for a different venue.

The important thing to note, is that nothing about the approach taken here precludes simply changing the replace directives in the `go.mod` to point at any other Erigon fork.  However, for the moment, there are some important changes around [ephemeral ports for p2p networking](https://github.com/ledgerwatch/erigon/pull/7503) that have been pushed upstream from the bobanetwork fork but not yet propagated into the testinprod-io fork that are required to prevent port conflicts when running multiple erigon instances.

Just like some of the other requisite changes such as the [storage proofs for eth_getProof](https://github.com/ledgerwatch/erigon/pull/7202) have already bubbled from the bobanetwork fork into erigon and back into the testinprod-io version, I expect it's only a matter of time until these tests can successfully execute against the testinprod-io version if that's ultimately the fork the community chooses to align with. 

**Alternatives**

This PR tries to stick with idiomatic go, avoiding unnecessary abstractions and directly referencing 'real' implementations until there's a need to do something else.  But, the changes could be accomplished in other ways.

For instance, where this PR chooses to have an `EthInstance` type with a pointer either to a Geth instance or an Erigon instance, we could instead create an interface which either the Geth instance or Erigon instance could satisfy, and have the test setup inject a function to construct the instance.  This would have the benefit of not requiring any direct references to Erigon in op-e2e (and, if so desired to op-geth as well).  Personally, this seems like abstraction with no real benefit that will make the code less clear and more fragile and once additional clients are available the abstraction can always be added at a later date.  Of course, if the reviewers feel this is the better approach, I'm happy to modify the PR accordingly.

**Problems**

There are a few tests that depend on Geth specific implementation details.  This includes RPCs that are geth specific (like the assorted `debug_` APIs) as well as certain Geth block building options.  These are skipped for the time being and noted in the tests.

Additionally, although the tests run fairly reliably on my laptop, when running with Erigon, flakes are more likely in a couple particular tests.  This is likely due to the additional overhead of having many processes and real network connections, but I'm actively working to eliminate these flakes.

**Conclusion**

Sorry for the small novel that is this PR text, and hopefully you've made it this far.  I'll be adding code comments below.  Thanks for reviewing!


go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	github.com/multiformats/go-multiaddr v0.8.0
 	github.com/multiformats/go-multiaddr-dns v0.3.1
 	github.com/olekukonko/tablewriter v0.0.5
+	github.com/onsi/gomega v1.26.0
```
jyellick: This is a new dependency, which I would usually try to avoid in a PR like this.  But, gomega provides some very nice asynchronous polling primitives that reduce the code complexity of dealing with external processes.  In particular waiting for the process to become ready by watching the log output can be accomplished safely and without long sleeps in a single line (I'll note this below).

If we're opposed to bringing in this dependency, it can be written out, but will require more code.

op-e2e/bridge_test.go
```diff
@@ -76,7 +76,7 @@ func TestERC20BridgeDeposits(t *testing.T) {
 	// Approve WETH9 with the bridge
 	tx, err = WETH9.Approve(opts, predeploys.DevL1StandardBridgeAddr, new(big.Int).SetUint64(math.MaxUint64))
 	require.NoError(t, err)
-	_, err = waitForTransaction(tx.Hash(), l1Client, 3*time.Duration(cfg.DeployConfig.L1BlockTime)*time.Second)
+	_, err = waitForTransaction(tx.Hash(), l1Client, 6*time.Duration(cfg.DeployConfig.L1BlockTime)*time.Second)
```
jyellick: Without bumping this timeout, this test becomes fairly flaky with Erigon.  Since this is a timeout, and not a sleep, it should not impact the Geth test execution time.

op-e2e/erigon.go
```diff
@@ -0,0 +1,157 @@
...
+		"--datadir", er.DataDir,
+		"init", genesisPath,
+	)
+	sess, err := gexec.Start(
+		cmd,
+		gexec.NewPrefixedWriter(er.Name, os.Stdout),
+		gexec.NewPrefixedWriter(er.Name, os.Stderr),
+	)
+	gt.Expect(err).NotTo(gomega.HaveOccurred())
+	gt.Eventually(sess.Err, time.Minute).Should(gbytes.Say("Successfully wrote genesis state"))
```
jyellick: This sort of construct is why gomega is being pulled in, as this one line will block execution, polling every 10ms, waiting for a particular substring from the process output.  It can obviously be done by spinning up another go routine, creating channels, etc. but, the simplicity and readability make it worth the import for me.

op-e2e/geth.go
```diff
@@ -300,23 +300,25 @@ func createGethNode(l2 bool, nodeCfg *node.Config, ethCfg *ethconfig.Config, pri
...
-		act, err := ks.ImportECDSA(pk, password)
-		if err != nil {
-			n.Close()
-			return nil, nil, err
-		}
-		err = ks.Unlock(act, password)
-		if err != nil {
-			n.Close()
-			return nil, nil, err
+	if !l2 {
```
jyellick: This block of changes is essentially a no-op, and could be removed.  But, I thought it made it more clear that only the L1 node actually needs a signing key, and I found the fact that all of the L2 nodes were getting an identical signing key to the L1 was confusing.

op-e2e/system_test.go
```diff
@@ -1279,7 +1323,7 @@ func TestStopStartBatcher(t *testing.T) {
 	receipt := sendTx()
 
 	// wait until the block the tx was first included in shows up in the safe chain on the verifier
-	safeBlockInclusionDuration := time.Duration(3*cfg.DeployConfig.L1BlockTime) * time.Second
+	safeBlockInclusionDuration := time.Duration(4*cfg.DeployConfig.L1BlockTime) * time.Second
```
jyellick: Another change to reduce flakiness.  The 3 second time window isn't always sufficient when there is lots of process contention.

ajsutton:
Wow, there's a lot to digest here but let me just say upfront that this is amazing work. Very exciting to be thinking through how a multi-client test suite might work. I think there's some important questions for how we should do this in a way that scales well but it will take me a bit of time to try and put together a coherent set of thoughts and know what the right questions to ask are. :)

op-e2e/bridge_test.go
ajsutton: Yeah timeouts have often been set too short. I'm fine with bumping them up. May even be worth splitting the duration increases into a separate PR as it's worth making them plenty long even just for op-geth. CI servers can be amazingly slow at times.

op-e2e/geth.go
ajsutton: I just made a similar change to the devnet startup scripts so glad to see redundant options being removed from here too.  Might be another one worth splitting into its own PR and we can just get it merged now.

op-e2e/system_test.go
```diff
@@ -1351,22 +1395,21 @@ func TestBatcherMultiTx(t *testing.T) {
...
+	// possible additional L1 blocks will be created before the batcher starts,
+	// so we wait additional blocks.
+	for i := int64(0); i < 10; i++ {
+		block, err := waitForBlock(big.NewInt(int64(l1Number)+i), l1Client, time.Duration(cfg.DeployConfig.L1BlockTime*5)*time.Second)
+		require.Nil(t, err, "Waiting for l1 blocks")
 		totalTxCount += len(block.Transactions())
+
+		if totalTxCount >= 10 {
+			return
+		}
```
ajsutton: I like this change as a way to reduce flakiness in general as well - worth merging it in regardless of how we move forward with multiclient support.

ajsutton (commented):
Definitely agree that being able to reuse these e2e system tests is very useful and so we should do something to make them work in a multi-client way. There is also a possibility of using hive to provide cross-client testing which is what it's designed for, but personally I find way too complex and difficult to work with which is why the op-geth tests wound up being added here rather than in hive.

Right now, there's a weird mix of actual e2e system tests and actions tests which are more like integration tests in the `op-e2e` directory. The actions tests are less suitable for cross-client testing as they stub a lot of stuff out and depend a fair bit more on internal details. The system tests are a better fit and especially the `op_geth_test.go` file. There's some deposit and withdrawal flow tests and a few overall sanity tests that make a lot of sense to run cross-client too. Other system tests are generally focussed on testing op-node or the batcher and proposer components so don't really need to run on all clients, though it wouldn't necessarily be bad if they did.

There's some rearranging we should do to make the distinctions between the different types of tests clearer. That's outside the scope of this PR, just noting it because doing multi-client tests well is likely going to be a journey with many PRs over a period of time as we learn and improve things. We don't have to have a perfect solution before we start merging stuff.

So I think the biggest question here is around whether we want the monorepo to wind up with a dependency on all the different clients. I don't think that's particularly viable given the rate at which new op clients are turning up. Apart from the time and cost of running the tests for each client in CI, we need to make sure the maintenance burden of the integrations and handing failures is shared "right". There's two sides to that:
 1. it should be easy for each client team to update the integration when they change their client (eg if an CLI option changes or there's a new dependency or whatever)
 2. if client A commits a change that breaks something, development on client B shouldn't be affected and client B shouldn't be living with tests failing in CI.

As much as possible we want it to be permissionless for clients to integrate and start running these tests too. Anyone should be able to plug their new client into the tests and be able to run them. Ideally they wouldn't have to maintain a fork of the monorepo to do that but at minimum adding a new integration should require very few changes to existing code - you might need to introduce a new module in the monorepo with the adapter logic (like what `erigon.go` provides here) but we want to minimise the required changes outside that module as much as possible.


With that in mind, looking at this specific PR…

The big concern for me is having to add a dependency on op-erigon to the monorepo. It means any updates to the integration or Erigon build process require a PR to the monorepo which will be a pain for you and we wind up with some maintenance work to keep the Erigon supporting code working even though we don’t really know much about it.

I suspect we could avoid that by adding support for a generic external client rather than op-erigon specifically. I hacked something together to show how this would work in https://github.com/ethereum-optimism/optimism/compare/aj/generic-l2-client It’s not right by any means but talking in code can be useful. :) The key idea is that rather than an `—erigon` flag we add a `—l2.client` flag that lets you specify the path to an arbitrary binary to execute. The e2e tests will then run that binary when they need to start a client and pass in the options as CLI args. The binary is then expected to write the ports to use to a file (and should only do that when the client is up and running properly). So to hook up a new client you’d write a custom runner that maps those input arguments to the appropriate CLI options for that client. You can see an example of the runner implementation in `op-erigon/runner/main.go` in that branch. It’s basically the same as the `erigon.go` in your code. We’d write a spec to define the CLI args that must be supported and the contents of the ports file to write.  We’ll have to come up with some way of specifying specific requirements so that we can skip tests that need `debug_getStorageAt` for example or that don’t work for some other reason.  Needs a little thought but should be doable.

What do you think of that kind of approach?

I also think we want to take the branch by abstraction approach mentioned in the description:
> For instance, where this PR chooses to have an EthInstance type with a pointer either to a Geth instance or an Erigon instance, we could instead create an interface which either the Geth instance or Erigon instance could satisfy, and have the test setup inject a function to construct the instance.

That makes things much more client agnostic but also just better encapsulates the differences between clients rather than needing to scatter if statements through the code.

I did notice that most of the op_geth_test cases aren’t using Erigon because they use `NewOpGeth` rather than a full deploy config setup (the only need geth, not the rest of the op-stack).  They’re likely to be some of the most useful tests for compatibility so we should hook into there as well (though probably as a separate PR).

jyellick:
Thanks @ajsutton for the very thoughtful response!

> The big concern for me is having to add a dependency on op-erigon to the monorepo. It means any updates to the integration or Erigon build process require a PR to the monorepo which will be a pain for you and we wind up with some maintenance work to keep the Erigon supporting code working even though we don’t really know much about it.

I totally agree, and I'd really have preferred to avoid bringing the dependency along, so finding a path forward that removes it would make me happy.

> As much as possible we want it to be permissionless for clients to integrate and start running these tests too. Anyone should be able to plug their new client into the tests and be able to run them. Ideally they wouldn't have to maintain a fork of the monorepo to do that but at minimum adding a new integration should require very few changes to existing code - you might need to introduce a new module in the monorepo with the adapter logic (like what erigon.go provides here) but we want to minimise the required changes outside that module as much as possible.

This is really great.  If we're willing to explore the idea of making these op-e2e tests consumable by external implementations without requiring a fork of the monorepo (or at least without modifying the op-e2e tests themselves) that does feel like the best of all worlds.  (I'd considered adding this in the 'alternatives' section but things were already getting to be quite long, and I thought perhaps it was too invasive of a change).

I tried to think of any ways that we could allow these tests to be pulled in as a Go import for other projects, but the more I've thought about it the less this makes sense.  Allowing as a Go import doesn't help non-Go clients, and in general the Go test framework isn't really amenable to sharing test cases across packages.  Of course it can be done, but it's not elegant and would require a more substantial refactoring of the tests.  Additionally, anyone developing another client is likely to want to take other parts of the monorepo, like the devnet tests, which probably ensures they're already maintaining a fork anyway.

So, below, I'm writing under the assumption that consuming projects will need to fork the monorepo and modify the build configuration to inject an external process as you've described.

> I also think we want to take the branch by abstraction approach mentioned in the description:

So the big question to me here, is whether we want to leave the in-process approach for the Geth instance, or write an adapter for op-geth as an external process or both.

The obvious pro to leaving things in process for op-geth is that iterating on the tests locally is going to be a bit faster and less resource intensive.  And, the other bonus here is that for tests which can benefit from more direct Geth manipulation, they can still potentially cast down to the underlying instance and sidestep the abstraction.

The benefits to running op-geth as an external process would be to force test authors to ensure that new tests can be leveraged by multiple clients by avoiding exactly those direct manipulation techniques that can occur by drilling into the client.  It would also ensure that those working in the monorepo feel the same pain as external users that might be trying to consume it in terms of timeouts etc.

Perhaps the best of both worlds would be to leave the in-memory op-geth path, and to create an external adapter for op-geth that executes in CI.  That way the external-adapter path is at least tested with a real client in the monorepo, while developers can still iterate more quickly when executing tests locally.  Of course, the downside to this approach is longer CI run times and more code to maintain.

> I did notice that most of the op_geth_test cases aren’t using Erigon because they use NewOpGeth rather than a full deploy config setup (the only need geth, not the rest of the op-stack). They’re likely to be some of the most useful tests for compatibility so we should hook into there as well (though probably as a separate PR).

This is good info, as despite hacking around inside these tests a fair bit, I can't claim to have a good understanding of what all the test subsets are doing, so if these others are useful, adapting them to the external runner is something I can prioritize.

**Next Steps**

1. ~I'll watch for you reply here particularly on the open question of whether we want to retain the in-process op-geth path or not.~ (Done)
2. ~I'll rework this PR to eliminate the dependency on op-erigon and create a generic external adapter~ (Done)
3. ~I'll take the bits you referenced as good to merge on their in your comments and submit them as separate PRS.~ (Done)

op-e2e/geth.go
jyellick: Done in #6291

op-e2e/system_test.go
jyellick: Done in #6292

op-e2e/bridge_test.go
jyellick: Done in #6294

op-e2e/system_test.go
jyellick: Done in #6294

protolambda:
One more argument to preserve in-process op-geth is that it allows to set breakpoints in op-node and op-geth in the same debugger, making test debugging a lot more pleasant.

ajsutton:
Yeah I think it's worth keeping the in-process `op-geth` version. As Proto mentions it makes debugging a lot easier but also it just minimises the scope of change and risk while we get this all set up and working. I'm keen to keep things reasonable light-weight until we've demonstrated that this approach works and these tests are useful for a range of clients. I think they will be and am excited to do this in a way that should be a lot simpler than hive, but you never know until you actually build it...

So yeah if you can get the external adapter setup that would be a great start and we can get that merged. Then we can iterate from there for things like how to ignore tests that use features not supported by the external client and so on.

jyellick:
I'm not entirely happy with the way I hacked in the CI definition, and there might be a few things left to polish.  That being said, I wanted to push this update because it demonstrates a working end to end flow with the 'external process model' discussed above.

Note that this also includes the conversion of `EthInstance` to an interface (that was previously a struct).

The overall approach taken was very similar to what was suggested, with a few notable deviations:
* Rather than passing many CLI flags to the external process, we opt to pass all of the arguments in as a single JSON file.  This feels like it's likely to be more compatible with other languages, and will be easier to extend over time.
* There are two modes of operation for the external process shim.  One with an `--init` flag, which allows for the external process to perform test-suite wide setup.  In particular, this allows the shim to do heavy lifting like builds, clones, etc. that would be prohibitive to repeat.  The `--config` flag operates to pass in the JSON and actually run the shim.
* Rather than trying to scrape the output of the external process (though the example shim still does this), we determine readiness based on waiting for a JSON file to appear.

Even though we're leaving the in process mode, I thought it was important that there be a way to truly test the external process style in tree.  So, there is a shim for op-geth included in the second commit, and wired into CI.  It's also a nice place to serve as an example and or starting place for others.  The README in that dir may also be helpful.

Importantly, all imports of erigon are gone, and, the only remaining references should be in comments/doc especially around test skips.

As always, I'm happy to adjust according to the feedback here, but I'm hoping that this is pretty close to what was being envisioned.

op-e2e/system_test.go
```diff
@@ -37,13 +46,55 @@ import (
...
+		externalL2Nodes = shimPath
+
+		_, err = os.Stat(externalL2Nodes)
+		if err != nil {
+			fmt.Printf("Failed to stat externalL2Nodes path: %s\n", err)
+			os.Exit(3)
+		}
+
+		cmd := exec.Command(externalL2Nodes, "--init")
+		cmd.Dir = filepath.Dir(shimPath)
```
jyellick: One issue with the way that this currently works, is that there's no real way for `--init` to pass artifacts/state to the other invocations of the shim other than to write files into the source tree.  For the geth external adapter, this just means writing the `op-geth` binary, which doesn't seem egregious.  But, it also feels a bit weird, and wouldn't play nicely if we ever wanted to run multiple `--init`s concurrently.

One option would be to create a working directory for the test suite execution, pass it in as an argument to init, and then as a parameter of the `config.json`.  This would allow for multiple inits to execute at once and generally feels a bit cleaner to me, but, also adds a bit of complexity.

op-e2e/external.go
```diff
@@ -0,0 +1,162 @@
...
+		gexec.NewPrefixedWriter("[exterr:"+er.Name+"]", os.Stderr),
+	)
+	gt.Expect(err).NotTo(gomega.HaveOccurred())
+
+	// 2 minutes may seem like a long timeout, and, it definitely is.  That
+	// being said, when running these tests with high parallelism turned on, the
+	// node startup time can be substantial (remember, this usually is a
+	// multi-step process initializing the database and then starting the
+	// client).
+	gt.Eventually(config.EndpointsReadyPath, 2*time.Minute).Should(gomega.BeARegularFile(), "external runner did not create ready file at %s within timeout", config.EndpointsReadyPath)
```
jyellick: Since we wanted to have a file conveying the endpoint info to the test suite anyway, I thought it could serve double-duty here.  This line is going to poll every 10ms, looking for the existence of this file, and once it exists, it indicates that the backing process must be alive and available at these endpoints.

Alternatively, we could have watched for some emitted log message like the earlier incarnation did, but, given that this file needs to exist in either case (and it would be a failure if the message were emitted and this file did not exist), this approach felt like a reasonably robust way to check for readiness.

.circleci/config.yml
```diff
@@ -938,15 +941,27 @@ jobs:
       - run:
           name: prep results dir
           command: mkdir -p /tmp/test-results
+      - when:
```
jyellick, line 786: It's hard for me to figure out what exactly the cohesive strategy for the CI files are.

Personally, I'd love to see all of this setup pushed out of CI and into the Makefile, that way we could easily execute these test permutations locally.

For the moment, I put in this mishmash of CircleCI conditions and shell scripting, but as mentioned in the comment, I don't really care for it, and think it can be done better, but it does work.  If there's a direction we'd like to take CI, I can refactor here, or in a future PR.

op-e2e/Makefile
```diff
@@ -4,6 +4,12 @@ test:
 lint:
 	golangci-lint run -E goimports,sqlclosecheck,bodyclose,asciicheck,misspell,errorlint -e "errors.As" -e "errors.Is"
 
+test-external-%: external_%/shim
```
jyellick: My make skills aren't terrific, but, the idea here is that if anyone drops another `external_<name>` directory into the `op-e2e` folder, they can execute the test with:

```
make test-external-<name>
```

See https://github.com/bobanetwork/v3-anchorage/commit/d6451b7b5dc5d3fc9817a8184dc7f7c3449bbd55

op-e2e/helper.go
```diff
@@ -1,30 +1,16 @@
...
-	"flag"
 	"os"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/log"
 )
 
-var enableParallelTesting bool = true
-
-// Init testing to enable test flags
```
jyellick, line 13: This code has moved around a bit since I've been following the e2e tests.

I'm not really sure why we've opted to move it out of a `_test.go` file and into `helper.go`.  It has the very unfortunate side effect of forcing anyone who includes this package to get all of the test flags added to their execution environment, which strikes me as rather odd.  But, odd things are often done for a purpose, so hopefully someone can let me know if/how this change breaks others.

One alternative, which I considered but didn't love, would be to take the external data structures and helper functions that the shims are also consuming, and put them into their own package.  That way the shims could get access to them without calling this init code, but if we don't need to, it seems better not to.

op-e2e/op_geth_test.go
```diff
@@ -44,9 +44,17 @@ func TestMissingGasLimit(t *testing.T) {
 // TestTxGasSameAsBlockGasLimit tests that op-geth rejects transactions that attempt to use the full block gas limit.
 // The L1 Info deposit always takes gas so the effective gas limit is lower than the full block gas limit.
 func TestTxGasSameAsBlockGasLimit(t *testing.T) {
+	if externalL2Nodes != "" {
```
jyellick: I've considered a bit how we might have clients advertise what test classes they should support.  My first inclination would be to have the execution of the `--init` step write some sort of compatibility info (probably a JSON file?).  The `TestMain` could consume that compatibility info and set package variables to selectively skip incompatible tests.

For the moment, as suggested, I'd like to ignore this portion of the problem, but wanted to mention the above as a possible future path forward.

op-e2e/helper.go
jyellick: Looks like this code has been pulled into `op-e2e/config/config.go`.  So, I've gone ahead and done the alternative discussed above, of moving the common structures into their own package.

FWIW, I think it's a bad idea to call `testing.Init()` and parsing flags in `func init()`, because it ensures anyone who wants to link to your package ends up having their flags broken.  I don't quite understand the big picture as to why these changes are being done, but if anyone agrees, I'm happy to submit a separate PR moving the flag parsing out of the package initialization.

op-e2e/system_test.go
```diff
@@ -1093,19 +1164,29 @@ func TestFees(t *testing.T) {
...
+	}
+
+	sga := &stateGetterAdapter{
+		ctx:    context.Background(),
+		t:      t,
+		client: l2Seq,
+	}
 
-	l1CostFn := types.NewL1CostFunc(seqBackend.BlockChain().Config(), seqState)
+	l1CostFn := types.NewL1CostFunc(config, sga)
```
jyellick, line 1179: I've gone ahead and written the adapter above, to allow for us to use this `types.NewL1Cost` function from inside this test.  That being said... I'm quite confused as to _why_ we're using it.  To me, the goal of these tests is to confirm that the system is working as expected.  By pulling in this internal function, it seems like we are testing only that the output of this function is being propagated.  If for instance, a bug were introduced changing the behavior of this function, it seems this test would not catch it?

jyellick (commented):
Fixed the conflicts.  It would be great if we could get some eyes on this, as it touches enough that conflicts arise easily and not all of them are easy to resolve.

jyellick:
@ajsutton What can I do to help move this forward? I think all of the initial comments have been addressed, and I'm unaware of anything in this that is blocking, but if there is, please let me know.  I'd really like to start working on further enhancements like moving the `NewOpGeth` based tests to be more generic and support external clients as well, but don't want to do so until I'm sure this approach is accepted.

.circleci/config.yml
ajsutton: Yeah I'm definitely a fan of moving stuff out of the circleci config and into a Makefile - so much easier to maintain and you can run it locally etc. Probably too much for this PR though.

.circleci/config.yml
```diff
@@ -1307,10 +1318,22 @@ workflows:
...
+      - go-e2e-test:
+          name: op-e2e-WS-tests-external-geth
+          module: op-e2e
+          use_http: "false"
+          use_external: "external_geth/shim"
+      - go-e2e-test:
+          name: op-e2e-HTTP-tests-external-geth
+          module: op-e2e
+          use_http: "true"
+          use_external: "external_geth/shim"
```
ajsutton: I suspect it's not worth the cost of running all the e2e tests again with external for http and ws. I'd probably just run the ws version using external and skip the http version. e2e tests are one of the more expensive things we run in CI so increasing the executor minutes will add up pretty fast.

op-e2e/external.go
ajsutton: Yeah I like using the endpoints file to signal readiness.

op-e2e/Makefile
```diff
@@ -13,6 +13,12 @@ test: pre-test
 lint:
 	golangci-lint run -E goimports,sqlclosecheck,bodyclose,asciicheck,misspell,errorlint -e "errors.As" -e "errors.Is"
 
+test-external-%: pre-test external_%/shim
+	go test -v --externalL2 ./external_$*/shim
+
+external_%/shim: external_%/*.go
+	go build -o external_$*/shim ./external_$*/
```
ajsutton: We probably shouldn't assume the external shim is written in go.  I'd just have this execute `make` in the shim's directory, then it can control how it builds and build any dependencies etc as well.

op-e2e/external_geth/main.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	os.Exit(0)
+}
+
+func build() error {
+	cmd := exec.Command("go", "build", "-o", "op-geth", "github.com/ethereum/go-ethereum/cmd/geth")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
```
ajsutton: I don't think I'd include the building logic here, but just make it a requirement that the shim and client are built before the tests are executed. You can usually get a command to execute before a test runs fairly easily  - especially with make but even in most IDEs.

op-e2e/helper.go
ajsutton: The init behaviour is pretty foreign to me (I'm really a Java dev pretending to be a go dev...) - it's been shoved around a bit as part of trying to separate and simplify the test code and then have implementation details in helper functions.  There wasn't really an intention for the e2e package to be needed to be imported by anyone else to be honest because it's really all just tests. There's some movement towards the helpers being in separate modules which I do think make sense so I think moving the common structures to their own module makes sense too.
I'm not really sure where the init code should live but I don't think it's in the current location by design, more just that's where it wound up because it worked at the time.

op-e2e/op_geth_test.go
```diff
@@ -44,9 +45,17 @@ func TestMissingGasLimit(t *testing.T) {
...
 // The L1 Info deposit always takes gas so the effective gas limit is lower than the full block gas limit.
 func TestTxGasSameAsBlockGasLimit(t *testing.T) {
+	if config.ExternalL2Nodes != "" {
+		// Some clients, such as Erigon will not discard txes that exceed the
+		// block gas limit, and instead retains them to re-evaluate with the
+		// next block whose gas limit may have increased.  We should enable this
+		// tests for external clients which exhibit this behavior, but currently
+		// we are disabling it for all external clients.
+		t.Skip("external client may not discard txes which exceed the gas limit")
+	}
```
ajsutton: I suspect that we're better off omitting this currently and then when running tests with erigon skip tests it can't handle with the `-skip` option.

op-e2e/system_test.go
ajsutton: I think making it a pre-requisite that the client is built avoids this uncertainty as it can be built into a normal `bin` directory just like all the other executables.  That's the approach we're using for things like `op-program` and `cannon` which are executed as part of e2e tests in separate processes.

op-e2e/system_test.go
ajsutton: There are (or at least should be) unit tests around the `NewL1Cost` function - the key aim of this test is to ensure that it's calculation is being added in at all the right places (and only added in once along the way). At least that's my understanding, I probably need to dig into it more closely to be sure I understand exactly what the aim is.

ajsutton (commented):
Sorry for the delay in getting back to you on this and for all the reworking of e2e that's gone on in the mean time.

The broader picture here is that we currently have some tests that use hive which is designed to test multiple clients but it does so by throwing docker at everything and so it gets very complex to debug things when tests or clients aren't working. Most of our tests wind up being written as these e2e tests because it's so much easier. Ultimately, having two different approaches to e2e testing doesn't make much sense so we're going to review what we have in hive vs these e2e tests and the trade offs between the two both for op-geth tests and for other cross-client testing. I've been slower than expected getting to doing the work to review everything but it's firmly on my todo list now. Having this around to demonstrate working with multiple clients is a really useful input to that process. Getting that review done is the key piece we're missing to be able to fully commit to an approach for cross client testing.

Thanks for bearing with me and again sorry for the delays on our side.

.circleci/config.yml
jyellick: Great, I'll try to put together a separate PR for this.

op-e2e/helper.go
jyellick: Great, I won't mess with it here, but a much more idiomatic pattern would be to expose a function that others can call to initialize these flags (much like `testing.Init()`) and then to call that setup in the `TestMain` as needed. I'll put together a separate PR that cleans this up.

op-e2e/op_geth_test.go
jyellick: I'll remove these checks.  If we want to skip them with the `-skip` option, I think it would be nice to have a generic way to do it.  Maybe a file or something that can be dropped into the `external_*/skip.json` or `external_/skip.env` or similar that the make target will automatically consume and build the skip flags from.  Will need to experiment with that, but will remove for now.

op-e2e/system_test.go
jyellick: I'll go ahead and remove this, though, for some of the same reasons that leaving the in process op-geth, I think it's nice to have the binary rebuilt from `go test` rather than `make`.  Let's say my binary is currently failing the test `TestFees`, it's nice to be able to run `go test -v -run='TestFees' --externalL2 external_erigon/shim` and iterate, modifying the code, adding tracing, etc. and simply re-executing the test command.  By forcing the binary to be built via `make`, there's an additional step and there's no assurance that the thing being tested is the latest code.  My suspicion is anyone who wants to use this external client support for development will end up hacking something similar in to keep their binary up to date.  As you point out, this could be accomplished in an IDE with a pre-build or pre-test hook, or on the CLI with shell scripts or aliases, but, to me it's nicer if the developer doesn't have to hack this together manually.

op-e2e/system_test.go
jyellick: Yes, I believe `NewL1Cost` does have some unit tests associated with it.  It still struck me as an odd modification to this test.  If we're wanting to ensure that the fields are being set, it's easy enough to check.  If we're wanting to make sure they're being set correctly, then I'm not sure what assurances calling the implementation and comparing the values gives us.  Regardless, not something I'm trying to insist change, but the addition of this call requires the somewhat hacky cludge here, so I thought maybe it might make more sense to derive it more directly in the test.

op-e2e/Makefile
jyellick: My suspicion is, that regardless of the language of the underlying client implementation, it's going to be easiest if the shim is done in Go.  Simply because the sample implementation is in Go, and all of the functions of parsing flags, reading/writing the JSON files, and scraping logs is all very generic and re-implementing is unlikely to provide any real value.

That being said, it's easy enough to convert this into a make invocation, and I don't see any real harm in doing so, so, will do.

.circleci/config.yml
jyellick: Yes, I was on the fence about adding both of these.  Will remove the HTTP one.

op-e2e/external.go
jyellick: Great -- I actually noticed that testify has an `Eventually` construct now.  I'll take a look and see if I can refactor things to remove this gomega dependency.

op-e2e/external_geth/main.go
jyellick: I discussed this a little more in my response to the `TestMain`.  Doing golang development, I generally expect to be able to run `go test` and have things work.  Actually, many of the recent changes forcing `make devnet-allocs` to be a prerequisite and requiring `make` to be used to invoke the tests feels like a significant regression to me, especially considering that all of these tests used to run just fine with simple `go test` invocations on the command line.

To me, doing the build from the shim is nicer, as it ensures that whether using IDE, `go test`, or `make`, that your test invocation will work with no additional setup or configuration.  I'll go ahead and move the external client build to `make`, but I'm unconvinced it's the right thing to do.

jyellick (commented):
Thanks for the reply @ajsutton -- I certainly understand that there are many different priorities competing for your attention.

As you mention, it does seem that most of the meaningful tests are here in the e2e and not in hive, which is why we chose to hook in here when doing our client development.  Even if this test function is ultimately migrated towards hive, it would be great if we could support external clients in the e2e in the meantime.

op-e2e/op_geth_test.go
ajsutton: I suspect the key will be to put the external client code in charge of kicking of the test run rather than the `op-e2e` module itself.  So rather than running `cd op-e2e; go test . -- --externalL2 external_geth/shim` you'd do something more like `cd external_geth; make test-e2e` and then make could ensure the client is built and run `go test` with the appropriate tests skipped. I'm not a make expert but ideally it would let you supply other test args so you could run a single test easily with the `-run <test>` option. `make` becomes the canonical entry point and external clients then have maximal flexibility to configure how the tests run, how the client builds, what to skip etc.

jyellick:
@ajsutton Everything's addressed, I'll continue rebasing, but please let me know if there's anything I missed

op-e2e/external_geth/main.go
ajsutton: Yeah I hear your concerns. I just think that `go test` is too low level a tool once you get into the realm of end-to-end tests. We're effectively using `make` as the canonical entry point as its that bit higher level and gives us better control over dependencies and the overall build pipeline. In an ideal world we should be testing the already built binaries - a build pipeline should build once and then the rest of the pipeline tries to find fault in that build.  We're a fair way off that right now but having e2e tests build their own dependencies moves us even further away from it.

op-e2e/system_test.go
ajsutton: You may well be right - for the purposes of this PR we're probably better off leaving what it tests the same and so stick with this shim.

op-e2e/external_geth/main.go
jyellick: I think this might be the slight schizophrenia I feel with the direction and goals of the e2e.

In our initial discussions there was a strong desire to keep the in-process op-geth because of the ease of iterative development it facilitates.  But, necessarily by testing op-geth by compiling it into the test binary created by `go test`, we're preventing these tests from truly executing against a release.  The same could be said for the other assorted components.

If the goal of the op-e2e is to execute against release binaries built elsewhere, then I totally agree that `make` is a fine tool to ensure they're there (though you'll have trouble convincing this gopher that `go test` isn't also fine e2e tool).  That just feels in conflict with those other discussions. 

But, regardless of the build implementation, I _do_ feel like allowing external binaries to be leveraged in the e2e is valuable -- both for op-geth itself and for other clients wishing to leverage these tests -- so I'd love to see this merged.

ajsutton (approved):
ok, thanks again for your patience on this. I think we've got all our ducks in a row in terms of the broader approach to testing so we can get this merged and use op-e2e is the primary end to end tests including for multi client. Separately we'll wind up migrating the few unique tests we currently have running in hive over to op-e2e as well and stop maintaining our hive fork.

I think I'll have to track down a second approval for contracts but since its just a simple test update affecting that code that should be straight forward.

This is a huge contribution so thank you for the amount of effort you've put in here.

mslipper (approved):

PR #6283, opened 2023-07-13
op-challenger: Add Defender of the Alphabet scripts

trianglesphere:
**Description**

Add Defender of the Alphabet scripts

Usage:
- Run `./init_game` then wait for it to complete
- Run `./mallory` & `./charlie` in parallel - they will play against each other
- Stop `./mallory` & `./charlie` when they no longer make progress - they will not stop themselves
- Run `./visualize` to generate `dispute-game.svg` which shows the claims which have been made.

**Tests**

Manually tested.


op-challenger/charlie
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
ajsutton, line 6: For each of the bash files can we add:
```suggestion
set -euo pipefail
```

Will catch a lot of common bash coding errors and ensure the script fails if a command fails rather than silently continuing.

op-challenger/init_game
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+# build the challenger to keep it up to date
+make
```
ajsutton, line 8: We probably shouldn't assume we're run from within the `op-challenger` dir (ie lots of people will run `./op-challenger/init_game`).
```suggestion
DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
cd "$DIR"

# build the challenger to keep it up to date
make
```

Can use that same construct in all the scripts to make them simpler to use.

op-challenger/visualize
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+DISPUTE_GAME_PROXY="0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"
+
+FAULT_GAME_ADDRESS="0x8daf17a20c9dba35f005b6324f493785d239719d"
+
+dir=$PWD
```
ajsutton:
```suggestion
dir=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
cd "$dir"
```

ajsutton (commented):
Personally I'd be very tempted to add a `.sh` suffix to each filename - it tends to make life easier with editors etc but I'm not particularly opinionated about it.

trianglesphere:
> it tends to make life easier with editors

I can make the change. I write all of these in vim which doesn't seem to care. `cat` also really doesn't care.

ajsutton (approved):
LGTM.

ajsutton:
> > it tends to make life easier with editors
> 
> I can make the change. I write all of these in vim which doesn't seem to care. `cat` also really doesn't care.

VS Code and GoLand are very slightly easier with the suffix but it's not a big deal.  I'm not sure if we've used the suffix consistently elsewhere or not - consistency is probably the main aim.

PR #6284, opened 2023-07-13
cannon: Differential Fuzzing syscalls

Inphi:
Main purpose of this is to differential fuzz the Go and Solidity VMs. These tests won't always assert what the correct behavior should be, but rather ensure that both VMs behave the same way.


cannon/mipsevm/evm_test.go
```diff
@@ -281,30 +280,14 @@ func TestClaimEVM(t *testing.T) {
...
-		logs := evmState.Logs()
-		require.Equal(t, 1, len(logs), "expecting a log with post-state")
-		evmPost := logs[0].Data
-		require.Equal(t, crypto.Keccak256Hash(evmPost), postHash, "logged state must be accurate")
-		env.StateDB.RevertToSnapshot(snap)
+		evm := NewMIPSEVM(contracts, addrs)
+		evm.SetTracer(tracer)
+		evmPost := evm.Step(t, stepWitness)
+
+		uniPost := us.state.EncodeWitness()
```
protolambda: All the `uniPost` occurrences may be worth renaming to e.g. `goPost`, since we don't use unicorn anymore.

protolambda (approved):

clabby (approved):

zhwrd (approved):

Issue #6092, opened 2023-06-21
op-node: unsafeL2Payloads blocks when receiving a block prior to unsafeL2Head


Kelvyne:
**Describe the bug**
When activating the unsafeL2Payload queue, op-node stops processing the queue if a payload is older than the rollup's unsafeL2Head.

**To Reproduce**

It happens more frequently when the batches are pushed slower. 

**Expected behavior**

I'd expect the payload to be dropped and op-node would continue processing the queue

**System Specs:**
 - OS: Mac M1 inside Docker
 - Package Version (or commit hash): 62c7f3b05a70027b30054d4c8974f44000606fb7

**Additional context**

A way I found to fix this issue is to pop these payloads when processing the queue. I don't know if there is bad side-effects from this diff.

```dif
diff --git a/op-node/rollup/derive/engine_queue.go b/op-node/rollup/derive/engine_queue.go
index 6f6c2fa5e..0e5f67ac3 100644
--- a/op-node/rollup/derive/engine_queue.go
+++ b/op-node/rollup/derive/engine_queue.go
@@ -428,6 +428,11 @@ func (eq *EngineQueue) tryNextUnsafePayload(ctx context.Context) error {
 			eq.log.Info("skipping unsafe payload, since it does not build onto the existing unsafe chain", "safe", eq.safeHead.ID(), "unsafe", first.ID(), "payload", first.ID())
 			eq.unsafePayloads.Pop()
 		}
+
+    if uint64(first.BlockNumber) <= eq.unsafeHead.Number {
+      eq.log.Info("skipping unsafe payload, since it is older than unsafe head", "unsafe", eq.unsafeHead.ID(), "unsafe_payload", first.ID())
+      eq.unsafePayloads.Pop()
+    }
 		return io.EOF // time to go to next stage if we cannot process the first unsafe payload
 	}
```
I would love to submit a PR if this is the way to go.

ajsutton:
@Kelvyne Apologies for the delay in responding to this. I agree this appears to be an issue - having p2p alt sync enabled can also trigger this I think because it's trying to backfill blocks and so there's multiple threads trying to fetch and queue blocks so could wind up re-adding a block to the queue that has already been imported.

And I think your fix makes sense as well. The unsafe head is strictly equal to or later than the safe head so we could just replace the safe head check with this unsafe check.
We'd also need to add a test for this case.

If you're still keen to submit a PR for this it would be appreciated and feel free to tag me to review.

[...]

PR #6290, opened 2023-07-14
op-node: Enable p2p alt sync by default.

ajsutton:
**Description**

Enable p2p sync by default. This has been in use on our Goerli nodes for some time and has been a net positive. While there are still potential cases like https://github.com/ethereum-optimism/optimism/issues/6092 which may mean it doesn't currently solve every problem, in those cases the node wouldn't be tracking unsafe head correctly without p2p sync anyway so it's never worse and solves unsafe head stalls in the majority of cases.

At this stage not enabling peer scoring by default as well though we're close to being able to do that. Peer scoring is more difficult when p2p sync isn't enabled on the majority of nodes because they get downscored for not supporting p2p sync. We have appropriate rate limiting in place to prevent DoS vectors even without peer scoring being enabled (and it's just a flag away if we do need it).

**Metadata**

- https://linear.app/optimism/issue/CLI-3899/rollout-and-test-unsafe-p2p-backup


op-node/flags/p2p_flags.go
```diff
@@ -292,6 +292,7 @@ var (
 	SyncReqRespFlag = &cli.BoolFlag{
 		Name:     "p2p.sync.req-resp",
 		Usage:    "Enables experimental P2P req-resp alternative sync method, on both server and client side.",
```
maurelian: Probably shouldn't be a default if considered experimental. :)

```suggestion
		Usage:    "Enables P2P req-resp alternative sync method, on both server and client side.",
```

sebastianst (approved):

op-node/flags/p2p_flags.go
ajsutton: hah - good spot. :)

mslipper (approved):

PR #6291, opened 2023-07-14
op-e2e: only add a signer key to l1

jyellick:
The current code adds the same signing key to all Ethereum clients (including the L2 nodes).  I found this to be quite confusing when parsing the code, and, it's unnecessary for the L2s to have signing keys.

cc @ajsutton


protolambda (approved):

ajsutton:
@jyellick Looks like CI is having a hard time. I don't really understand why `yarn` doesn't exist but there's been a bunch of changes and fixes around pnpm. Could you try rebasing this on top of latest `develop` and see if that helps please?

PR #6292, opened 2023-07-14
op-e2e: make TestBatcherMultiTx more resilient

jyellick:
TestBatcherMultiTx disables the batch submitter at the beginning of the test, then starts it later to ensure that it catches up and appropriately submits the transactions.

However, currently it does this by checking the block number before starting the batcher, then waiting for 3 blocks before checking for the transactions.  Because there is no synchronization, it's possible that the block number advances before the batcher starts, which can lead to the test flaking.

This commit changes the approach to fetching each of the next 10 blocks, and succeeding once the transactions are found, or failing if none are found within the next 10 blocks.

cc @ajsutton

### DIFF
--- a/op-e2e/system_test.go
+++ b/op-e2e/system_test.go
@@ -1351,22 +1351,21 @@ func TestBatcherMultiTx(t *testing.T) {
 	err = sys.BatchSubmitter.Start()
 	require.Nil(t, err)
 
-	// wait for 3 L1 blocks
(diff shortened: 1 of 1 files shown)


trianglesphere (approved):

jyellick:
Is there anything outstanding I need to address before getting this merged?

tynes:
Can confirm that the tests are passing in CI. Thank you for this contribution!

tynes (approved):

PR #6298, opened 2023-07-14
feat: Add contracts-periphery to contracts-ts

roninjin10:
- add contracts-periphery to the generated contracts-ts package

PR #6267, opened 2023-07-12
feat: Add gas estimation utilities

roninjin10:
- Add simple gas estimation tools such as the ability to estimate l1 gas with a minimal api
- This replaces l2Provider.ts version which was tightly coupled to ethers.js. This one you don't even realize you are using viem as the API is minimal
- Don't document this as it's a one off tool and the plan is to make normal tools "just work"
  - we should document it in a guide/tutorial instead after this lands

PR #6301, opened 2023-07-14
feat: Add address.json generation to contracts-bedrock

roninjin10:
- Add address.json generation to contracts-bedrock
- This localizes the source of truth for addresses to contracts-bedrock for easier maintainability
- @eth-optimism/contracts-ts will use these addresses as source of truth
- In future both op-bindings and docs can use it too


roninjin10:
* develop
  * **PR #6260** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6260" target="_blank">Graphite</a> 
    * **PR #6298** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6298" target="_blank">Graphite</a> 
    * **PR #6267** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6267" target="_blank">Graphite</a> 
      * **PR #6407** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6407" target="_blank">Graphite</a> 
        * **PR #6301** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6301?utm_source=stack-comment).

packages/contracts-bedrock/scripts/generate-addresses.ts
```diff
@@ -0,0 +1,249 @@
...
+    for (const chainId of Object.keys(chains)) {
+      addresses[contractName] = {
+        ...addresses[contractName],
+        [Number.parseInt(chainId)]: address,
+      }
+    }
+  }
+
+  for (const artifact of artifacts) {
+    if (addresses[artifact.contractName]?.[artifact.chainId]) {
```
roninjin10:
```suggestion
    // the preference here is:
    // 1. Predeploy address
    // 2. contracts-bedrock
    // 3. contracts-periphery
    if (addresses[artifact.contractName]?.[artifact.chainId]) {
```

packages/contracts-bedrock/package.json
```diff
@@ -10,25 +10,27 @@
...
-    "build": "npx nx build:contracts",
+    "build": "nx build:contracts",
     "prebuild:contracts": "./scripts/verify-foundry-install.sh",
     "build:contracts": "pnpm build:forge",
     "build:forge": "forge build",
     "build:differential": "go build -o ./scripts/differential-testing/differential-testing ./scripts/differential-testing",
     "build:fuzz": "(cd test-case-generator && go build ./cmd/fuzz.go)",
-    "autogen:invariant-docs": "ts-node scripts/invariant-doc-gen.ts",
+    "generate:addresses": "tsx scripts/generate-addresses.ts",
+    "generate:addresses:check": "pnpm generate:addresses && git diff --exit-code ./addresses.json && git diff --exit-code ./addresses.json",
```
refcell: Looks like the git diff was duplicated here.
We should just be able to do
```suggestion
    "generate:addresses:check": "pnpm generate:addresses && git diff --exit-code ./addresses.json",
```

packages/contracts-bedrock/scripts/generate-addresses.ts
```diff
@@ -0,0 +1,249 @@
...
+const deploymentValidator = z.object({
+  address: AddressValidator,
+  abi: AbiValidator
+})
+type Deployment = z.infer<typeof deploymentValidator>
+type Address = z.infer<typeof AddressValidator>
+
+/**
+ * Predeployed contract addresses
+ * In future it would be nice to have a json file in contracts bedrock be generated as source of truth
```
refcell: Think we can remove these comments now :D

refcell (changes requested):

tynes:
We were trying to remove all JS deps from the contracts package

tynes (changes requested):
We don't need any new deps in this package to accomplish this functionality. Once we remove all JS deps from this package and use foundry style dep management, then we are free from needing to wait for JS deps to be installed + built when running the contracts tests in CI

.circleci/config.yml
```diff
@@ -383,10 +383,24 @@ jobs:
...
+            - pnpm-packages-v2-{{ checksum "pnpm.lock.yaml" }}
+      - check-changed:
+          patterns: sdk,contracts-bedrock,contracts
       - run:
-          name: Check addresses.json is up to date
-          command: pnpm nx build && pnpm generate:addresses:check
-          working_directory: packages/contracts-bedrock
+          name: Check generated addresses.ts file
+          command: pnpm nx run-many --target=generate:addresses:check
+
```
roninjin10, line 432: @tynes moved this to it's own build pipeline.  It blocks nothing and depends on nothing so it's what I would consider "free".

Also nx makes it so after we turn on the cache it doesn't even run unless the artifacts actually change

nx.json
```diff
@@ -59,6 +63,9 @@
     "generate": {
       "dependsOn": ["^build"]
     },
+    "generate:addresses:check": {
+      "dependson": ["deployments", "^deployments", "scripts"]
+    },
```
roninjin10, line 68: This config makes nx smart enough to know that the address generation script check only ever needs to run when the artifacts actually have changed. Otherwise cache test result

packages/contracts-bedrock/package.json
```diff
@@ -18,8 +18,8 @@
     "build:forge": "forge build",
     "build:differential": "go build -o ./scripts/differential-testing/differential-testing ./scripts/differential-testing",
     "build:fuzz": "(cd test-case-generator && go build ./cmd/fuzz.go)",
-    "generate:addresses": "tsx scripts/generate-addresses.ts",
-    "generate:addresses:check": "pnpm generate:addresses && git diff --exit-code ./addresses.json && git diff --exit-code ./addresses.json",
+    "generate:addresses": "pnpm --filter=@eth-optimism/contracts-ts generate:addresses $PWD/addresses.json",
+    "generate:addresses:check": "pnpm generate:addresses && git diff --exit-code ./addresses.json",
```
roninjin10, line 22: Since this script doesn't have a build step it can run without adding @eth-optimism/contracts-ts as a dep
If I added the script to .bin it would have even been able to reference it directly in path without needing to call with pnpm. I chose pnpm because it's explicit though

packages/contracts-ts/scripts/generate-addresses.ts
```diff
@@ -3,6 +3,9 @@ import fs from 'fs'
 import path from 'path'
 import { z } from 'zod'
 import { Abi as AbiValidator, Address as AddressValidator } from 'abitype/zod'
+import { fileURLToPath } from 'url'
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
```
roninjin10, line 8: This is a es module package __dirname isn't defined in modern javascript

packages/contracts-ts/scripts/generate-addresses.ts
```diff
@@ -241,8 +246,8 @@ const generate = async () => {
   }
 
   await fs.promises.writeFile(
-    path.join(__dirname, '..', 'addresses.json'),
-    JSON.stringify(addresses, null, 2)
+    process.argv[2] || path.join(__dirname, '..', 'addresses.json'),
```
roninjin10, line 266: making the path we write to configurable so contracts-bedrock can call it

packages/contracts-bedrock/package.json
```diff
@@ -46,13 +48,14 @@
     "ethers": "^5.7.0"
   },
   "devDependencies": {
+    "@types/glob": "^8.1.0",
```
roninjin10, line 53:
```suggestion
```
And run pnpm i

roninjin10:
This pr is stacked last in the stack of prs since it was blocking higher priority changes

packages/contracts-bedrock/scripts/generate-addresses.ts
roninjin10: Good call lol

packages/contracts-ts/scripts/generate-addresses.ts
```diff
@@ -0,0 +1,254 @@
...
+})
+type Deployment = z.infer<typeof deploymentValidator>
+type Address = z.infer<typeof AddressValidator>
+
+/**
+ * Predeployed contract addresses
+ * In future it would be nice to have a json file in contracts bedrock be generated as source of truth
+ * Keep this in sync with op-bindings/predeploys/addresses.go in meantime
+ */
+const predeployContracts = {
```
roninjin10, line 26:
```suggestion
const predeployContracts = {
```
stale

packages/contracts-ts/wagmi.config.ts
```diff
@@ -326,8 +326,6 @@ const addressesByContractByNetworkPlugin: Plugin = {
     const addresses = Object.fromEntries(
       contracts.map((contract) => [contract.name, contract.address ?? {}])
     )
-    // write to json file so it's easy to audit in prs relative to the generated file diff
-    writeFileSync('./addresses.json', JSON.stringify(addresses, null, 2))
```
roninjin10, line 303: Not updating this script yet since the prs are in flux to reduce rebase effort

packages/contracts-ts/scripts/generate-addresses.ts
roninjin10: Later I will make this uses the addresses.json as source of truth

packages/contracts-bedrock/scripts/generate-addresses.ts
roninjin10: Left a different comment so closing this one

packages/contracts-ts/scripts/generate-addresses.ts
```diff
@@ -0,0 +1,254 @@
...
+  LegacyMessagePasser: {
+    address: '0x4200000000000000000000000000000000000000',
+    introduced: 'Legacy',
+    deprecated: true,
+    proxied: true,
+  },
+  DeployerWhitelist: {
+    address: '0x4200000000000000000000000000000000000002',
+    introduced: 'Legacy',
+    deprecated: true,
```
tynes, line 36: Why does this have multiple fields? `deprecated` isn't used, it looks like only the address is needed. The current source of truth of the predeploy addresses is in `core-utils` for typescript

packages/contracts-ts/scripts/generate-addresses.ts
```diff
@@ -0,0 +1,254 @@
...
+  throw new Error(
+    'No node_modules found. Please run `pnpm install` before running this script'
+  )
+}
+
+const nodeModules = path.join(__dirname, '..', 'node_modules')
+
+const bedrockDeployments = {
+  [1]: glob.sync(
+    path.join(nodeModules, '@eth-optimism', 'contracts-bedrock', 'deployments', chains[1], '*.json')
```
tynes: lots of duplicated code here, should be a function

packages/contracts-ts/scripts/generate-addresses.ts
```diff
@@ -0,0 +1,254 @@
...
+  [420]: [...bedrockDeployments[420], ...peripheryDeployments[420]],
+}
+
+const getArtifacts = async () => {
+  const artifactPromises: Promise<Deployment & { chainId: number, contractName: string }>[] = []
+  // same loop as 2 for loops 
+  for (const [chainId, deploymentFiles] of Object.entries(deployments)) {
+    for (const artifactPath of deploymentFiles) {
+      const contractName = artifactPath.split('/').reverse()[0]?.replace('.json', '')
+      const artifact = fs.promises.readFile(artifactPath, 'utf8').then((a) => deploymentValidator.parse(JSON.parse(a))).then((deployment) => ({
```
tynes: the deployment validator is overengineering imo, all you need is something like `if typeof artifact.address != 'string' || !ethAddressRegex.matches(artifact.address) { throw new Error() } `

Also the type of the entire deployment file isn't verified, there are a bunch more top level fields. This script really only needs to know that there are json files where a single key is `address` with an eth addy as the value. Validating unused things adds tech debt because now this script depends on the shape of the deployment file when it doesn't need to. There is a future world where our deployment artifacts look different

packages/contracts-ts/scripts/generate-addresses.ts
```diff
@@ -0,0 +1,254 @@
...
+const generate = async () => {
+  const artifacts = await getArtifacts()
+
+  const addresses: {
+    [contractName: string]: {
+      [chainId: number]: Address
+    }
+  } = {}
+
+  for (const [contractName, { address }] of Object.entries(predeployContracts)) {
```
tynes: can you add a logline here that logs the contract name?

packages/contracts-ts/scripts/generate-addresses.ts
tynes: why use `.then` instead of `await`?

packages/contracts-ts/scripts/generate-addresses.ts
tynes: Why is this going thru node modules instead of the local package?

packages/contracts-ts/scripts/generate-addresses.ts
```diff
@@ -0,0 +1,254 @@
...
+  [5]: glob.sync(
+    path.join(nodeModules, '@eth-optimism', 'contracts-bedrock', 'deployments', chains[5], '*.json')
+  ),
+  [420]: glob.sync(
+    path.join(nodeModules, '@eth-optimism', 'contracts-bedrock', 'deployments', chains[420], '*.json')
+  ),
+}
+
+// These are here instead of contracts-periphery because contracts-periphery is getting moved to contracts-bedrock eventually so they will be here after that happens
+const peripheryDeployments = {
```
tynes: perhaps we can get merge https://github.com/ethereum-optimism/optimism/pull/6206 before this so we can delete this temp code and never need to merge it

.circleci/config.yml
tynes: Restore cache wont make it depend on js dep install?

packages/contracts-bedrock/addresses.json
```diff
@@ -0,0 +1,253 @@
+{
+  "LegacyMessagePasser": {
```
tynes, line 2: Why does this file exist in 2 packages?

packages/contracts-bedrock/addresses.json
```diff
@@ -0,0 +1,253 @@
...
+    "1": "0x4200000000000000000000000000000000000000",
+    "5": "0x4200000000000000000000000000000000000000",
```
tynes, line 4: Should sort the keys alphabetically

.circleci/config.yml
roninjin10: Correct. Restore cache depends on pnpm_lock so the js dep install only ever happens when a new dep is added or removed.  This is true for all workflows. Linking pnpm is instant and it's perfectly well cached so in general pnpm is an unproblematic bottleneck in all pipelines

packages/contracts-bedrock/addresses.json
roninjin10: It exists in contracts-bedrock because that's where it belongs
It exists in contracts-ts so it's there for folks installing contracts-ts also have an easy to reference json file

packages/contracts-ts/scripts/generate-addresses.ts
roninjin10: Touching code outside of your own package in an unexplicit way in a package based monorepo is a bad practice. When you go through node modules you are guaranteeing you correctly declared your dependency instead of it being implicit

roninjin10: Is this blocking?  My standard for stuff like this in scripts is much lower than actual code in a larger system. Duplication in a script is usually easier to maintain than the alternative.

packages/contracts-ts/scripts/generate-addresses.ts
roninjin10: I don't think merging either pr first has any value.  If this merged first then #6206 we can then just do a tiny follow up pr to clean up.

packages/contracts-ts/scripts/generate-addresses.ts
roninjin10: `if typeof artifact.address != 'string' || !ethAddressRegex.matches(artifact.address) { throw new Error() }`
This is harder to maintain harder to read less idiomatic and just in general worse code and also not typesafe. This is a downgrade. Just use zod as a primitive is generally what folks like myself do these days because it's so simple and used everywhere.

`Also the type of the entire deployment file isn't verified`

This is on purpose we only want to validate the fields we actually use so that if the other fields change everything still just works.

`Validating unused things adds tech debt because now this script depends on the shape of the deployment file when it doesn't need to. There is a future world where our deployment artifacts look different`

I feel like this is what I just said. Missing piece of context is I am using the ABI in future prs

roninjin10: I don't have a personal preference either way

roninjin10: `This is on purpose we only want to validate the fields we actually use so that if the other fields change everything still just works.`

`Duck-typing` is the word I couldn't think of this is just duck typing. We need only the address and the abi nothing else is used

packages/contracts-ts/scripts/generate-addresses.ts
tynes: contracts periphery no longer exists so we should remove all refs to it in this script

packages/contracts-ts/scripts/generate-addresses.ts
tynes: I didn't realize you are using the abi in future PRs. I think we can agree that we have different opinions on javascript and thats fine. If you strongly prefer using zod i'm ok with it but I still think we should really only be using the standard library for these scripts. In my eyes thats an upgrade because no deps is very simple and portable, i also find it easier to read and much more explicit than zod which who knows what it does by its name and all the extra typescript syntax. It also allows the package to operate outside of the context of the monorepo fully, altho in this case this script isn't particularly useful when outside of the monorepo, but eventually something would be introduced that makes operations difficult outside of this repo if we are not careful

roninjin10:
This is a nice-to-have pr so will resurect it later when we actually use this to generate docs. Devrel team has plans to update docs next quarter at which time we can bring this back. 

Generally good to not merge code that is not used anyways

PR #6287, opened 2023-07-14
fix(ci): Fix CI issues

roninjin10:
Update ci builder and also do a temporary fix to fix ci while ci builder is redeployed

PR #6303, opened 2023-07-15
chore: More robust caching strategy for pnpm installs in circle ci

roninjin10:
Changes our caching strategy to only cache based on pnpm-lockfile.yaml instead of packages/*/package.json. This has 2 advantages:
- Removes false cache misses when the package.json changes but the deps didn't
- Makes the circle config more robust to breakage. We don't have to manually keep it in sync everytime we add and remove a package
  - This protects us since the first time we add a new package it will work and pass checks and possibly merge if we forget and then fail on second run when cache is badly populated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,20 +115,23 @@ jobs:
           name: Restore PNPM Package Cache
           keys:
             - pnpm-packages-v2-{{ checksum "pnpm-lock.yaml" }}
+      # Fetch node_modules into the pnpm store
+      # This will cache node_modules based on pnpm-lock so other steps can instantly install them with `pnpm install --prefer-offline`
+      # --prefer-offline installs node_modules instantly by just reading from cache if it exists rather than fetching from network
+      # when installing node_modules pnpm simply adds symlinks instead of copying the files which is why it is pretty much instant to run --prefer-offline
+      # this allows a caching strategy of only checking pnpm-lockfile so we don't have to keep it in sync with our packages
+      # For more information see https://pnpm.io/cli/fetch
       - run:
-          name: Install dependencies
-          command: pnpm install --frozen-lockfile
+          name: Fetch dependencies
+          command: pnpm fetch --frozen-lockfile --prefer-offline
       - save_cache:
           name: Save PNPM Package Cache
           key: pnpm-packages-v2-{{ checksum "pnpm-lock.yaml" }}
           paths:
             - "node_modules"
-            - "packages/chain-mon/node_modules"
-            - "packages/common-ts/node_modules"
-            - "packages/contracts-bedrock/node_modules"
-            - "packages/core-utils/node_modules"
-            - "packages/sdk/node_modules"
-            - "packages/contracts-ts/node_modules"
+      - run:
+          name: Install dependencies
+          command: pnpm install --frozen-lockfile --offline
       - run:
           name: print forge version
           command: forge --version
@@ -327,6 +330,10 @@ jobs:
           name: print forge version
           command: forge --version
           working_directory: packages/contracts-bedrock
+      # populate node modules from the cache
+      - run:
+          name: Install dependencies
+          command: pnpm install --frozen-lockfile --prefer-offline
       - run:
           name: test and generate coverage
           command: pnpm coverage:lcov
@@ -348,6 +355,10 @@ jobs:
       - checkout
       - check-changed:
           patterns: contracts-bedrock,op-node
+      # populate node modules from the cache
+      - run:
+          name: Install dependencies
+          command: pnpm install --frozen-lockfile --prefer-offline
       - run:
           name: print forge version
           command: forge --version
@@ -372,6 +383,10 @@ jobs:
             - pnpm-packages-v2-{{ checksum "pnpm-lock.yaml" }}
       - check-changed:
           patterns: contracts-bedrock,op-node
+      # populate node modules from the cache
+      - run:
+          name: Install dependencies
+          command: pnpm install --frozen-lockfile --prefer-offline
       - run:
           name: build contracts
           command: pnpm build
@@ -454,6 +469,10 @@ jobs:
       - checkout
       - check-changed:
           patterns: contracts-bedrock
+      # populate node modules from the cache
+      - run:
+          name: Install dependencies
+          command: pnpm install --frozen-lockfile --prefer-offline
       - run:
           name: slither
           command: |
@@ -469,6 +488,10 @@ jobs:
           name: Restore PNPM Package Cache
           keys:
             - pnpm-packages-v2-{{ checksum "pnpm-lock.yaml" }}
+      # populate node modules from the cache
+      - run:
+          name: Install dependencies
+          command: pnpm install --frozen-lockfile --prefer-offline
       - check-changed:
           patterns: contracts-bedrock
       - run:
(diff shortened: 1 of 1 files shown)


roninjin10:
* develop
  * **PR #6287** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6287" target="_blank">Graphite</a> 
    * **PR #6210** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6210" target="_blank">Graphite</a> 
      * **PR #6303** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6303" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6303?utm_source=stack-comment).

roninjin10:
CI won't actually pass until ci builder is updated to have yarn

mslipper (approved):

tynes (approved):

roninjin10:
This looks good to merge since CI passed

roninjin10:
It's stacked on my own pr though so don't merge it til that one merges

tynes (changes requested):
This has merge conflicts in it now

tynes:
@roninjin10 I just rebased this

roninjin10:
@tynes oops I likely just force pushed over your changes

I noticed a bunch of unnecessary stuff in the config. Will likely reset to your rebase @tynes and then open a new pr for that. e.g. stuff that doesn't depend on node_modules was still restoring cache etc.

Don't merge this til I remove the save-cache. The save-cache should only exist in the monorepo job

.circleci/config.yml
```diff
@@ -115,34 +115,14 @@ jobs:
...
-      # packages installed via npm via github automatically run postpack scripts
-      # their postpack script happens to use yarn so we need it here
-      - run:
-          name: Install yarn package manager
-          command: |
-            npm i yarn@1 --global
       - run:
           name: Install dependencies
-          command: pnpm install --frozen-lockfile
+          command: pnpm install --frozen-lockfile --prefer-offline
```
roninjin10, line 120: This is unecessary pretty sure actually could ust delete

.circleci/config.yml
```diff
@@ -115,34 +115,14 @@ jobs:
...
           key: pnpm-packages-v2-{{ checksum "pnpm-lock.yaml" }}
           paths:
             - "node_modules"
-            - "packages/chain-mon/node_modules"
-            - "packages/common-ts/node_modules"
-            - "packages/contracts-bedrock/node_modules"
-            - "packages/core-utils/node_modules"
-            - "packages/replica-healthcheck/node_modules"
-            - "packages/sdk/node_modules"
-            - "packages/contracts-ts/node_modules"
```
roninjin10, line 131: these can go away and then anything that needs node modules can just `pnpm i --offline-only` to instantly install from cache

roninjin10: will move to it's own pr though

.circleci/config.yml
roninjin10: --prefer-offline is faster if there is any cache

roninjin10: again will move to it's own pr

tynes:
Lemme know when this is ready for review again

roninjin10:
I think someone got to this before I got back?

roninjin10:
will do some of the other cleanups I added

roninjin10:
Dismissed @mslipper review since this is a different pr now
@tynes I'm likely adding a `pnpm i --prefer-offline` in places it isn't necessary and relying on you to point them out

.circleci/config.yml
roninjin10: This own pr is now this pr

roninjin10: Doing --prefer-offline instead of --offline-only just because it makes it less fragile to breaking with the same upside

mslipper (approved):

.circleci/config.yml
```diff
@@ -327,6 +330,10 @@ jobs:
           name: print forge version
           command: forge --version
           working_directory: packages/contracts-bedrock
+      # populate node modules from the cache
```
tynes: I don't think we need JS deps in this task

.circleci/config.yml
```diff
@@ -348,6 +355,10 @@ jobs:
       - checkout
       - check-changed:
           patterns: contracts-bedrock,op-node
+      # populate node modules from the cache
+      - run:
+          name: Install dependencies
```
tynes: We def don't need JS deps in this task

.circleci/config.yml
```diff
@@ -372,6 +383,10 @@ jobs:
             - pnpm-packages-v2-{{ checksum "pnpm-lock.yaml" }}
       - check-changed:
           patterns: contracts-bedrock,op-node
+      # populate node modules from the cache
```
tynes: We need JS deps in this task but will not need them in the future

.circleci/config.yml
```diff
@@ -454,6 +469,10 @@ jobs:
       - checkout
       - check-changed:
           patterns: contracts-bedrock
+      # populate node modules from the cache
```
tynes: We don't need JS deps in this task

.circleci/config.yml
```diff
@@ -484,6 +507,10 @@ jobs:
       - attach_workspace: { at: "." }
       - check-changed:
           patterns: contracts-bedrock
+      # populate node modules from the cache
+      - run:
+          name: Install dependencies
```
tynes: Don't need JS deps, tbh i think we can remove this entire task

.circleci/config.yml
```diff
@@ -515,6 +542,10 @@ jobs:
             - pnpm-packages-v2-{{ checksum "pnpm-lock.yaml" }}
       - check-changed:
           patterns: contracts-bedrock
+      # populate node modules from the cache
+      - run:
```
tynes: Same, don't need JS deps, can prob remove the entire task

.circleci/config.yml
```diff
@@ -527,6 +558,10 @@ jobs:
     resource_class: medium
     steps:
       - checkout
+      # populate node modules from the cache
+      - run:
+          name: Install dependencies
```
tynes: Don't need JS

.circleci/config.yml
```diff
@@ -643,22 +690,17 @@ jobs:
...
           name: Install pnpm package manager
           command: |
             npm i pnpm --global
-      # TODO remove me after ci builder updated
-      # A github dep clones-with-immutable-args is installed via github
-      # packages installed via npm via github automatically run postpack scripts
-      # their postpack script happens to use yarn so we need it here
       - run:
           name: Install yarn package manager
           command: |
```
tynes: We now install solidity deps with git submodules so I don't think we need yarn here unless for a diff reason from the removed comment

.circleci/config.yml
```diff
@@ -862,6 +907,10 @@ jobs:
       - checkout
       - check-changed:
           patterns: <<parameters.working_directory>>,<<parameters.dependencies>>
+      # populate node modules from the cache
+      - run:
```
tynes: No need for JS deps here

tynes (changes requested):
A lot of additions that we don't need, would prefer to keep as much JS out of the contracts and Go CI as possible due to it not being necessary

roninjin10:
Will get back to this this week

.circleci/config.yml
roninjin10: Oh yea likely fixed now

roninjin10:
This pr blocks @spacesailor24 pr https://github.com/ethereum-optimism/optimism/pull/6619

.circleci/config.yml
tynes: Confused why this is here, wont this slow down things? no js is needed here

tynes (approved):

PR #6305, opened 2023-07-16
op-service,batcher,proposer: Retry when dialing RPC URLs

trianglesphere:
**Description**

This adds retries when dialing ethclient or rollupclient RPC urls. Because this now does up to 10 exponential backoffs, the dial timeout has been increased to 15s (though this may not be enough).

I have also modified the API to DialClient to no longer take a context. This is because we are not sharing the context & if we want to pass in a context, we should probably set the timeout outside the dial function.

**Tests**

None

**Metadata**

- Fixes CLI-3985


op-service/client/dial.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return sources.NewRollupClient(client.NewBaseRPCClient(rpcCl)), nil
+}
+
+// Dials a JSON-RPC endpoint repeatedly, with a backoff, until a client connection is established. Auth is optional.
+func dialRPCClientWithBackoff(ctx context.Context, log log.Logger, addr string) (*rpc.Client, error) {
+	bOff := backoff.Exponential()
```
ajsutton: The issue specifically requested linear backoff rather than exponential - we don't need exponential backoff here because the cost of making an attempt is very low. Plus we're dead in the water until this succeeds so don't want to wait any longer than we have to.

ajsutton (commented):
There's a big open question here of whether we want the node to continue trying to connect forever or if we just wanted a longer period.  Typically in my experience you want the node to continue trying but not report as healthy, whereas this will still give up and crash after a while.

Either way it's a net improvement but we probably want to check in with dev-infra before actually closing the issue. Mind you, I note that you logged the original ticket so...

op-service/client/dial.go
ajsutton: Technically it should be a fixed backoff - just retry every 2s or so.

ajsutton:
Also even with this change op-node immediately exits if L1 isn't available:
```
t=2023-07-17T09:46:35+1000 lvl=info msg="Initializing Rollup Node"
INFO [07-17|09:46:35.599] No persisted sequencer state loaded 
INFO [07-17|09:46:35.599] Rollup Config                            l2_chain_id=901 l2_network="unknown L2" l1_chain_id=900 l1_network="unknown L1" l2_start_time=1,689,213,885 l2_block_hash=0x5872a64759f8482601b6175d334dfb3c9bb7043ec6495f5dcf7c2f347b8d1a31 l2_block_number=0 l1_block_hash=0x1337f5077c1e465e123901c331822728094082a73063161d1df3f57856913e2d l1_block_number=0 regolith_time="@ genesis"
ERROR[07-17|09:46:35.603] Error initializing the rollup node       err="Post \"http://localhost:8545\": dial tcp 127.0.0.1:8545: connect: connection refused"
ERROR[07-17|09:46:35.603] Unable to create the rollup node         error="Post \"http://localhost:8545\": dial tcp 127.0.0.1:8545: connect: connection refused"
CRIT [07-17|09:46:35.603] Application failed                       message="Post \"http://localhost:8545\": dial tcp 127.0.0.1:8545: connect: connection refused"
```

I don't think `Dial` even attempts a connection for http. Seems to work for ws connections though.

trianglesphere:
> change op-node immediately exits if L1 isn't available:

The op-node has a different dial code path :(. My other PR attempts to make it slightly better, though maybe that codepath isn't fully working yet

ajsutton:
> > change op-node immediately exits if L1 isn't available:
> 
> The op-node has a different dial code path :(. My other PR attempts to make it slightly better, though maybe that codepath isn't fully working yet

It did retry when I used a `ws` URL though. I'm pretty sure `Dial` just creates the object and will never fail with http because it only creates connections on demand when requests are made. Which makes it much more complex to add retry at startup sadly.

That said, I think this is a net improvement, so if we can switch to a fixed retry schedule and edit the description to not mark the issue as fixed it's worth merging.

op-service/client/dial_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsURLAvailable(t *testing.T) {
+	go func() {
+		_ = http.ListenAndServe(":8989", nil)
+	}()
```
ajsutton: Shouldn't something stop this server again at the end of the test?

And this test is not going to be flaky when it starts hitting port conflicts. You probably need to listen on port 0 which will guarantee selecting a free port and you can then get the effective port to connect to.  Checking if a URL isn't available I'd probably use an unresolvable domain - sadly `example.com` actually resolves even though it's reserved for examples but can probably just use a particularly long root domain - we can probably handle updating the test if it ever gets actually registered.

ajsutton (commented):
I'm not entirely convinced about the approach of checking the port is open as that doesn't actually mean the endpoint is ready to accept requests properly so we may still get some startup failures when the first request is made but its probably good enough.

We do need to make the test safe though.

op-service/client/dial.go
```diff
@@ -0,0 +1,79 @@
...
+		return client, nil
+	})
+}
+
+func IsURLAvailable(address string) bool {
+	u, err := url.Parse(address)
+	if err != nil {
+		return false
+	}
+	conn, err := net.DialTimeout("tcp", u.Host, 1*time.Second)
```
protolambda: This timeout seems very short to me

op-service/client/dial_test.go
protolambda: +1 this, we shouldn't hardcode ports in tests.

op-service/client/dial_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+	require.True(t, IsURLAvailable("http://localhost:8989"))
+	require.False(t, IsURLAvailable("http://localhost:9898"))
```
protolambda: For the false case maybe use an IPv6 black-hole IP address instead? Then the dial will always fail.

protolambda (changes requested):
Backoff/retry code looks good, but would change the timeout on the URL check + avoid hardcoded ports in the testing.

op-service/client/dial.go
trianglesphere: Dial's should be fast & we're explicitly doing retries. I can raise it a little bit, but I tend to keeping it smaller.

trianglesphere:
> I'm not entirely convinced about the approach of checking the port is open as that doesn't actually mean the endpoint is ready to accept requests properly so we may still get some startup failures when the first request is made but its probably good enough.


Definitely worth checking in on if this works in prod. I looked through the logs in the linked ticket & the issue is that host is not available which this should fix.

op-service/client/dial_test.go
trianglesphere: Fixed.

op-service/client/dial_test.go
trianglesphere: Using port 0 instead.

op-service/client/dial.go
ajsutton: There's nothing that guarantees a dial is fast. You need to go through a full TCP handshake which requires multiple round trips. If you have high ping times, some packet loss or an overloaded server it can easily exceed 1s.  Gotta think of us Aussies.... :) Remember that if this timeout is too short the node is dead in the water.

op-service/client/dial.go
```diff
@@ -0,0 +1,79 @@
...
```
protolambda:
```suggestion
	conn, err := net.DialTimeout("tcp", u.Host, 5*time.Second)
```

See other older review comment, if we fix this short timeout, e.g. by increasing it to 5 seconds (the old default dial timeout) we can merge this.

protolambda (approved):

protolambda:
```
# github.com/ethereum-optimism/optimism/op-program/host/prefetcher
host/prefetcher/retry.go:100:17: undefined: backoff.DoCtx
make[1]: *** [Makefile:17: op-program-host] Error 1
make[1]: Leaving directory '/home/circleci/project/op-program'
make: *** [Makefile:59: op-program] Error 2

Exited with code exit status 2
```
rebase missed an issue with op-program usage of the same API

ajsutton (approved):
Looks good once the new call gets updated.

hamdiallam (approved):
Looks good on the indexer changes

roninjin10 (approved):

PR #6306, opened 2023-07-16
op-challenger: Properly implement stepping against trace index 0

trianglesphere:
**Description**

When performing an attack step against trace index 0 we need to provide the absolute pre-state rather than pulling the value from the middle of the trace.

This PR also modifies the absolute pre-state that is stored in the devnet to match what the op-challenger is playing.

**Tests**

Unit tests for the op-challenger.

**Metadata**

- Fixes CLI-4198


refcell (approved):
LGTM

PR #6309, opened 2023-07-16
op-challenger: Cleanup initialization & add Act command

trianglesphere:
**Description**

This factors our some of the challenger setup code & adds an `Act` command which performs all actions against the current view of the contract & then exits. This is contrast to the primary loop mode which does this on a loop.

### DIFF
--- a/op-challenger/challenger.go
+++ b/op-challenger/challenger.go
@@ -13,38 +13,75 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
-// Main is the programmatic entry-point for running op-challenger
-func Main(logger log.Logger, cfg *config.Config) error {
+type Challenger struct {
+	agent  fault.Agent
+	logger log.Logger
+}
+
+// NewChallenger creates a new challenger agent.
+func NewChallenger(logger log.Logger, cfg *config.Config) (*Challenger, error) {
 	client, err := ethclient.Dial(cfg.L1EthRpc)
 	if err != nil {
-		return fmt.Errorf("failed to dial L1: %w", err)
+		return nil, fmt.Errorf("failed to dial L1: %w", err)
 	}
 
 	txMgr, err := txmgr.NewSimpleTxManager("challenger", logger, &metrics.NoopTxMetrics{}, cfg.TxMgrConfig)
 	if err != nil {
-		return fmt.Errorf("failed to create the transaction manager: %w", err)
+		return nil, fmt.Errorf("failed to create the transaction manager: %w", err)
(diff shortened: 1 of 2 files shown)


op-challenger/challenger.go
```diff
@@ -13,38 +13,75 @@ import (
...
+func MainAct(logger log.Logger, cfg *config.Config) error {
+	c, err := NewChallenger(logger, cfg)
+	if err != nil {
+		return err
+	}
+
+	logger.Info("Fault game started in single action mode")
+	c.Act()
+	return nil
+}
```
ajsutton, line 87: This seems a bit weird. It's not really a main method. I'd say the programmatic entry point is just `MainLoop` because the functionality of the program is to constantly watch games and perform actions.  If a single step is desired, it's probably better to just create the challenger and call `Act` once.

For the loop version it probably is worth passing in a `Context` as  well the log and config - then you could programmatically exit the loop by cancelling the context.  And the CLI library conveniently provides a Context as well so that would then hook into it nicely to handle graceful shutdown.

op-challenger/cmd/main.go
```diff
@@ -65,6 +65,24 @@ func run(args []string, action ConfigAction) error {
...
+				logger.Info("Starting op-challenger", "version", VersionWithMeta)
+
+				cfg, err := config.NewConfigFromCLI(ctx)
+				if err != nil {
+					return err
+				}
+				return op_challenger.MainAct(logger, cfg)
+			},
+		},
+	}
```
ajsutton, line 85: ah, I see why there's a second main entry point now...

I wonder if it would be simpler to just have a `--one-shot` or `--single` CLI option that then goes into the config and the main method switches between calling Loop or Act.  Would save duplicating the init code and isn't an uncommon pattern.

ajsutton:
I definitely like having the logic moved into `Challenger` with the `Act` and `Loop` functions. Both of those are business logic and so good to separate from the wiring type code that should live in Main.

op-challenger/cmd/main.go
trianglesphere: yea, this was the thing that was slightly weird. There's also the fact that `op_chalenger.MainLoop` is passed to `run`, but `op_challenger.MainAct` is directly called here.

My preference is to having different sub-commands. I think it's more likely that we want the ability to run an instance of the challenger on a specific game or only have it watch the output. By default it might bundle everything, but I do want to be able to unbundle it.

trianglesphere:
This seems to have been superseded by other PRs.

PR #6310, opened 2023-07-17
op-challenger: Add tests for Position attack and defend

ajsutton:
**Description**

Added tests for `Position` `Attack` and `Defend` methods.


trianglesphere (approved):
Lgtm

PR #6313, opened 2023-07-17
Revert "fix: Add missing contract deployments to contracts-bedrock"

tynes:
**Description**

This reverts commit 360cb821c3133b27f8d2605183781a93ed73e30d.

The artifacts here are all legacy and should not be used, the set of remaining artifacts are the contracts that we care about


maurelian (approved):
This list of contracts is safe to remove. 
They are all either unused, or have been renamed.

PR #6314, opened 2023-07-17
feat(op-challenger): FDG Caller

refcell:
**Description**

Adds a simple contract caller with logs for visibility into the FDG contract during calls. Makes debugging a little bit easier.


refcell:
* develop
  * **PR #6314** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6314" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6314?utm_source=stack-comment).

op-challenger/challenger.go
```diff
@@ -40,11 +41,26 @@ func Main(logger log.Logger, cfg *config.Config) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to bind the fault contract: %w", err)
+	}
+
 	logger.Info("Fault game started")
 
 	for {
 		logger.Info("Performing action")
 		_ = agent.Act()
+		status, err := caller.GetGameStatus(context.Background())
```
trianglesphere: This is a good thing to pull out into a function which either returns the info or does the logging.

refcell: Ah I like that design, will do.

op-challenger/fault/caller.go
```diff
@@ -0,0 +1,79 @@
...
+
+// GetClaimDataLength returns the number of claims in the game.
+func (fc *FaultCaller) GetClaimDataLength(ctx context.Context) (*big.Int, error) {
+	return fc.ClaimDataLen(&bind.CallOpts{Context: ctx})
+}
+
+func (fc *FaultCaller) LogClaimDataLength() {
+	claimLen, err := fc.GetClaimDataLength(context.Background())
+	if err != nil {
+		fc.log.Error("failed to get claim count", "err", err)
```
ajsutton, line 86: Should return here rather than proceeding to log an incorrect number of claims.

op-challenger/fault/caller.go
```diff
@@ -0,0 +1,79 @@
...
+// 1: Challenger Won
+// 2: Defender Won
+func (fc *FaultCaller) GetGameStatus(ctx context.Context) (uint8, error) {
+	return fc.Status(&bind.CallOpts{Context: ctx})
+}
+
+func (fc *FaultCaller) LogGameStatus() {
+	status, err := fc.GetGameStatus(context.Background())
+	if err != nil {
+		fc.log.Error("failed to get game status", "err", err)
```
ajsutton, line 72: Should return here instead of proceeding to log incorrect game status.

op-challenger/fault/caller.go
```diff
@@ -0,0 +1,79 @@
...
+	return &FaultCaller{
+		caller,
+		log,
+	}, nil
+}
+
+// LogGameInfo logs the game info.
+func (fc *FaultCaller) LogGameInfo() {
+	fc.LogGameStatus()
+	fc.LogClaimDataLength()
```
ajsutton: Could I suggest we reduce this down to a single log line. So we get something like `Game status addr=0x1234 claims=10 status=In Progress`. Makes it much easier to find and read the log lines.  The address isn't really important now but may as well include it since we'll eventually support multiple games and it's easy to include now.

I'd just not log the status if either piece of info can't be retrieved for simplicity.

ajsutton (commented):
Good idea - these kind of logs will be really useful in production.

op-challenger/fault/caller.go
refcell: 100% always happy to simplify

ajsutton (approved):
LGTM.

PR #6316, opened 2023-07-17
feat(ctb): Support variable length traces in `FaultDisputeGame`

clabby:
# Overview

Adds support for variable length traces in a fixed-depth `FaultDisputeGame`. A requirement of the game is that it properly handles players disputing traces of different lengths within the same depth tree, and previously, using a trace that was shorter than `2^{MAX_GAME_DEPTH}` instructions caused funky behavior.

In the games below, we have 2 players:
* Charlie (green): `abcdefgh`
* Mallory (red): `abcdexyz`

In this game, Mallory creates an incorrect root claim committing to `z`. Charlie attacks this claim with `h`, which is his full trace. This is because the tree supports a trace length up to 16, so the actual bisection of their trace starts at d=2 rather than d=1.
Gindices (11)
Once the players make it to the point where they can perform an instruction step, there was a bug pertaining to defense steps. Here, the step from `E` (agreed) -> `F` (charlie's claim @ trace index = 5) produces the correct post state, but Mallory's `E` needs to be countered due to it agreeing with the incorrect root claim. In the current contract, this invalid claim is unable to be countered because the step function reverted every time it produced the expected post state.

To fix this, we can follow @ajsutton's decision tree for defense steps by including information about the difference between the parent / post state claims' positions:
| Condition | Result |
| ---------- | ------ |
| VM Step produces correct post state & depth diff is odd | Leaf can be countered |
| VM Step produces correct post state & depth diff is even | Valid step; Revert |
| VM Step produces incorrect post state & depth diff is odd | Valid step; Revert |
| VM Step produces incorrect post state & depth diff is even | Leaf can be countered |

The above rules also work for a flipped game where the root claim is valid:
Gindices (12)

**Tests**
* Added arena tests for half-length and quarter-length traces with both correct & incorrect root claims.

**Metadata**
Fixes CLI-4197


refcell (approved):
LGTM

PR #6324, opened 2023-07-18
fix(op-node): Correctly pop unsafe payload queue when payload is too …

### LINKED ISSUE #6092: op-node: unsafeL2Payloads blocks when receiving a block prior to unsafeL2Head

Kelvyne:
…older

Fixes #6092

**Description**

See #6092

**Tests**

The test is minimalist. It only ensures that the unsafe payload queue is pop'd 

**Additional context**

**Metadata**

- Fixes #6092

### DIFF
--- a/op-node/rollup/derive/engine_queue.go
+++ b/op-node/rollup/derive/engine_queue.go
@@ -420,6 +420,11 @@ func (eq *EngineQueue) tryNextUnsafePayload(ctx context.Context) error {
 		eq.unsafePayloads.Pop()
 		return nil
 	}
+	if uint64(first.BlockNumber) <= eq.unsafeHead.Number {
+		eq.log.Info("skipping unsafe payload, since it is older than unsafe head", "unsafe", eq.unsafeHead.ID(), "unsafe_payload", first.ID())
+		eq.unsafePayloads.Pop()
+		return nil
+	}
 
 	// Ensure that the unsafe payload builds upon the current unsafe head
 	// TODO: once we support snap-sync we can remove this condition, and handle the "SYNCING" status of the execution engine.
--- a/op-node/rollup/derive/engine_queue_test.go
+++ b/op-node/rollup/derive/engine_queue_test.go
@@ -1107,3 +1107,88 @@ func TestResetLoop(t *testing.T) {
 	l1F.AssertExpectations(t)
 	eng.AssertExpectations(t)
 }
+
+func TestEngineQueue_StepPopOlderUnsafe(t *testing.T) {
+	logger := testlog.Logger(t, log.LvlInfo)
+	eng := &testutils.MockEngine{}
+	l1F := &testutils.MockL1Source{}
+
+	rng := rand.New(rand.NewSource(1234))
+
+	refA := testutils.RandomBlockRef(rng)
+	refA0 := eth.L2BlockRef{
+		Hash:           testutils.RandomHash(rng),
+		Number:         0,
+		ParentHash:     common.Hash{},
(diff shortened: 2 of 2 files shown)


Kelvyne:
cc @ajsutton

ajsutton:
Ooo, thank you. I'll get @protolambda to take a look since he knows this code really well and it is quite an important area.

One additional bit of context is that p2p sync is now enabled by default, so while we previously went to almost any length to avoid dropping gossip we might need, now we can be a bit more relaxed knowing that p2p sync can fetch the blocks if needed. I think there may be a case where a L1 reorg causes the safe head to drop back and so there's potentially a period where we haven't seen that reorg yet but the sequencer has published new unsafe payloads that are prior to our current unsafe head (but after the post-reorg unsafe head). This change would then drop those payloads and depend on p2p sync filling them in. However that's likely less common and resolves faster than the case where p2p sync winds up adding a payload just as the safe head advances and unsafe processing stalls until the next safe head update which is what this fixes.

protolambda:
> This change would then drop those payloads and depend on p2p sync filling them in. 

Note that it would only be able to fill them in after first reorging the chain using the safe-head reorg mechanism. Since we do not really have a forkchoice other than following the latest canonical L1 chain, we do not allow unsafe-head reorgs until seeing the new L1 chain.

protolambda (approved):

ajsutton:
> > This change would then drop those payloads and depend on p2p sync filling them in.
> 
> Note that it would only be able to fill them in after first reorging the chain using the safe-head reorg mechanism. Since we do not really have a forkchoice other than following the latest canonical L1 chain, we do not allow unsafe-head reorgs until seeing the new L1 chain.

Yep agreed. Which is similar to the current setup where importing unsafe payloads stops because the next payload is prior to the current unsafe head, until the L1 reorg is process and then that first pending unsafe payload should now build on the new safe head and be able to be imported.

PR #6326, opened 2023-07-18
contracts-bedrock: delete deploy config type

tynes:
**Description**

The deploy config type in typescript is no longer required. The Go version should now be considered the canonical version. This is prep work for creating validation scripts in `op-chain-ops` for deploy config JSON files that are added to the repository to ensure that they have sane values, are not missing any values and do not have any extra values.

This will make it easier to add or remove values over time, reducing the number of implementations from 3 to 2 as there is also a solidity version. The solidity version currently has additional config that is not in the Go version, a follow up PR will add those values to the Go version.

Also remove `ethers` from `contracts-bedrock` as a dep as it is no longer needed.


maurelian (approved):

op-chain-ops/genesis/config.go
```diff
@@ -69,64 +96,76 @@ type DeployConfig struct {
...
-	// L1StandardBridge proxy address on L1
+	// L1StandardBridgeProxy represents the address of the L1StandardBridgeProxy on L1 and is used
+	// as part of building the L2 genesis state.
 	L1StandardBridgeProxy common.Address `json:"l1StandardBridgeProxy"`
-	// L1CrossDomainMessenger proxy address on L1
+	// L1CrossDomainMessengerProxy represents the address of the L1CrossDomainMessengerProxy on L1 and is used
+	// as part of building the L2 genesis state.
 	L1CrossDomainMessengerProxy common.Address `json:"l1CrossDomainMessengerProxy"`
-	// L1ERC721Bridge proxy address on L1
+	// L1ERC721Bridge represents the address of the L1ERC721Bridge on L1 and is used
```
refcell:
```suggestion
	// L1ERC721BridgeProxy represents the address of the L1ERC721Bridge on L1 and is used
```

op-chain-ops/genesis/config.go
```diff
@@ -69,64 +96,76 @@ type DeployConfig struct {
...
 	GovernanceTokenOwner common.Address `json:"governanceTokenOwner"`
-
+	// DeploymentWaitConfirmations is the number of confirmations to wait during
+	// deployment. This is DEPRECATED and should be removed in a future PR.
 	DeploymentWaitConfirmations int `json:"deploymentWaitConfirmations"`
-
-	EIP1559Elasticity  uint64 `json:"eip1559Elasticity"`
+	// EIP1559Elasticity is the elasticity of the EIP1559 fee market.
+	EIP1559Elasticity uint64 `json:"eip1559Elasticity"`
+	// EIP1559BaseFeeMaxChangeDenominator is the denominator of EIP1559 base fee market.
```
refcell:
```suggestion
	// EIP1559Denominator is the denominator of EIP1559 base fee market.
```

refcell (changes requested):
Two small comment nits, otherwise very, very nice 😄

refcell (approved):

PR #6328, opened 2023-07-18
feat: EAS contracts

tynes:
**Description**

Replacement of https://github.com/ethereum-optimism/optimism/pull/6277 for CI
This PR adds the EAS and the SchemaRegistry contracts (and their go bindings) so that they are built as part of the build system

**Additional context**

The contracts were ported almost entirely from https://github.com/ethereum-attestation-service/eas-contracts-upgradeable, which is the ERC1967 compliant variant of https://github.com/ethereum-attestation-service/eas-contracts. 

The following minor modifications were made:
- Importing the internal version of `Semver`
- Importing `@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol` instead of `@openzeppelin/contracts/utils/cryptography/EIP712.sol` since the (slightly) older of OZ is being used by the rest of the contracts.


packages/contracts-bedrock/contracts/EAS/Common.sol
```diff
@@ -0,0 +1,102 @@
+// SPDX-License-Identifier: MIT
+
```
refcell, line 2:
```suggestion
```

packages/contracts-bedrock/contracts/EAS/EAS.sol
```diff
@@ -0,0 +1,824 @@
+// SPDX-License-Identifier: MIT
+
```
refcell, line 2:
```suggestion
```

packages/contracts-bedrock/contracts/EAS/IEAS.sol
```diff
@@ -0,0 +1,465 @@
+// SPDX-License-Identifier: MIT
+
```
refcell, line 2:
```suggestion
```

packages/contracts-bedrock/contracts/EAS/ISchemaRegistry.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: MIT
+
```
refcell, line 2:
```suggestion
```

packages/contracts-bedrock/contracts/EAS/SchemaRegistry.sol
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: MIT
+
```
refcell, line 2:
```suggestion
```

packages/contracts-bedrock/contracts/EAS/resolver/ISchemaResolver.sol
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: MIT
+
```
refcell, line 2:
```suggestion
```

packages/contracts-bedrock/contracts/EAS/resolver/SchemaResolver.sol
```diff
@@ -0,0 +1,180 @@
+// SPDX-License-Identifier: MIT
+
```
refcell, line 2:
```suggestion
```

packages/contracts-bedrock/scripts/Deployer.sol
```diff
@@ -103,20 +103,20 @@ abstract contract Deployer is Script {
...
+            bytes memory code = _getCode(contractName);
+            bytes memory deployedCode = _getDeployedCode(contractName);
             string memory receipt = _getDeployReceiptByContractAddress(addr);
 
             string memory artifactPath = string.concat(deploymentsDir, "/", deploymentName, ".json");
 
             uint256 numDeployments = 0;
             try vm.readFile(artifactPath) returns (string memory res) {
                 numDeployments = stdJson.readUint(string(res), "$.numDeployments");
+                numDeployments++;
```
refcell: Why do we need this? I though the `numDeployments` is set to X in the artifact, and we want to put that in the solidity Artifact struct below? Is it because of 0 vs 1-indexing?

refcell (changes requested):
Very small changes - Happy to follow this PR up with style changes :)

tynes:
Regarding all the newline suggestions, was thinking that we get their code checked in as is as much as possible and then fix the style part of `forge fmt` migration

packages/contracts-bedrock/scripts/Deployer.sol
tynes: The idea is that each time that the same deployment is deployed, you increment the `numDeployments` key in the JSON. This should really increment no matter what tbh, so after the try/catch

refcell (approved):
LGTM

PR #6329, opened 2023-07-18
Add Public Goods Network to types

vacekj:
**Description**

Adds Public Goods Network (https://publicgoods.network/) to Chain IDs.


tynes:
Looks like there is a build issue due to typing

PR #6330, opened 2023-07-18
feat(indexer): e2e blocks & bridge integration tests

hamdiallam:
e2e integration tests for blocks & bridge deposits/withdrawals! 🎉
- wire up indexer config. Catch interrupts and gracefully close. Configurable logger useful in tests. utilize op-service helpers for this
- e2e testing harness setup
- bugfix: OutputProposal was incorrectly deserializing  topics. Utilize generic`UnpackLog` rather than manually parsing topics. Always use this now that the utility is available
- Since the nonce isn't explicitly emitted via events, we need to handle extract the nonce from `OptimismPortal#finalizeWithdrawalTransaction` when parsing through standard bridge finalization events

We'll call this Part (1) of  ethereum-optimism/devx#26 with a few followups:
- e2e test ERC20 deposit/withdrawal (simple)
- index all lowlevel messages passed through (L1/L2CrossDomainMessenger &  L2ToL1MessagePasser/OptimismPortal) separately from the bridge to create cleaner abstractions (relatively simply but M sized)
    - Important to identify bridge and non-bridge messages that is currently hardcoded and has baked in assumptions


indexer/cli/cli.go
```diff
@@ -20,16 +25,29 @@ type Cli struct {
...
 
+	indexer, err := indexer.NewIndexer(cfg)
 	if err != nil {
-		log.Crit("Failed to load config", "message", err)
+		return err
 	}
 
-	// finish me
-	return nil
+	signalChannel := make(chan os.Signal, 1)
```
tynes: I believe there is a helper inside of `op-service` for this functionality

hamdiallam: ah i see it now in opio

hamdiallam: before fixing this i'm going to put up a quick pr so that the indexer doesn't consume the monorepo as a dependency. Currently not recognizing the op-service dir.

Also explains why the bindings update for CrossDomainMessenger & StandardBridge weren't showing

hamdiallam: updated as well as utilize op-service/log for cli flags

roninjin10 (approved):
LGTM but @tynes can likely give a better review

PR #6332, opened 2023-07-18
feat(sdk): Add event filter params to sdk as options

wilsoncusack:
**Description**

This PR adds optional args `fromBlockOrBlockHash` and `toBlockOrHash` to `getMessageStatus`, `getMessageReceipt`, `waitForMessageReceipt`, `waitForMessageStatus`, and `estimateMessageWaitTimeSeconds`.

This was motivated by having issues when bridging due to the massive block range. 

**Tests**

I am not sure how we can test this. I ran the test suite locally. Maybe I can build the SDK from this branch in my project and try it?

**Additional context**

**Metadata**

- Fixes #[Link to Issue]


lauchness (commented):
🚀

tynes:
There is currently no sane way to test the sdk in some cases, def talk to @roninjin10 about the sdk future, he made the sdk-next tests

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -1026,10 +1057,17 @@ export class CrossChainMessenger {
     /**
      * The index of the withdrawal if multiple are made with multicall
      */
-    messageIndex = 0
+    messageIndex = 0,
```
roninjin10, line 1054:
```suggestion
    // consider making this an options object next breaking release
    messageIndex = 0,
```

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -659,10 +659,17 @@ export class CrossChainMessenger {
...
```
roninjin10, line 656:
```suggestion 
    // consider making this an options object next breaking release
    messageIndex = 0,
```

roninjin10 (approved):
image

roninjin10 (changes requested):
Ah good call out @tynes! Would it be possible to add a test? There are existing sdk tests in `packages/sdk/test-next/message-status.spec.ts` would just need to use any message from goerli/op-goerli older than the blocks we fork
9256679 for l1
11276409 for l2

roninjin10:
Testing waitForMessageStatus is a little involved so maybe just test that getMessageStatus returns a tx when in the block range and then doesn't return a tx when the blockrange is outside of the block the tx is in

wilsoncusack:
> Testing waitForMessageStatus is a little involved so maybe just test that getMessageStatus returns a tx when in the block range and then doesn't return a tx when the blockrange is outside of the block the tx is in 

Like a manual test or are you making real network calls in your tests?

packages/sdk/test-next/messageStatus.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+     * @see https://goerli-optimism.etherscan.io/tx/0x8fb235a61079f3fa87da66e78c9da075281bc4ba5f1af4b95197dd9480e03bb5
+     */
+    const txWithdrawalHash =
+      '0x8fb235a61079f3fa87da66e78c9da075281bc4ba5f1af4b95197dd9480e03bb5'
+
+    const txReceipt = await l2Provider.getTransactionReceipt(txWithdrawalHash)
+
+    expect(txReceipt).toBeDefined()
+
+    expect(await crossChainMessenger.getMessageStatus(txWithdrawalHash, 0, 9370789 - 1000, 9370789)).toBe(
```
wilsoncusack: specifying a block range so that this works with more public providers

packages/sdk/test-next/messageStatus.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { describe, expect, it } from 'vitest'
```
wilsoncusack, line 1: new file because old tests weren't running without .spec ?

packages/sdk/test-next/messageStatus.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    const txWithdrawalHash =
+      '0x8fb235a61079f3fa87da66e78c9da075281bc4ba5f1af4b95197dd9480e03bb5'
+
+    const txReceipt = await l2Provider.getTransactionReceipt(txWithdrawalHash)
+
+    expect(txReceipt).toBeDefined()
+
+    expect(
+      await crossChainMessenger.getMessageStatus(txWithdrawalHash, 0, 0, 0)
+    ).toBe(MessageStatus.READY_FOR_RELAY)
```
wilsoncusack, line 48: This is READY_FOR_RELAY because, though the query filter failed, we found the proved withdrawal. Kinda odd so not sure if we want to change behavior here https://github.com/ethereum-optimism/optimism/blob/develop/packages/sdk/src/cross-chain-messenger.ts#L729

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -2240,16 +2252,14 @@ export class CrossChainMessenger {
...
+     * @param messageIndex The index of the message, if multiple exist from multicall
      * @returns Gas estimate for the transaction.
      */
     finalizeMessage: async (
       message: MessageLike,
       opts?: {
         overrides?: CallOverrides
       },
-      /**
-       * The index of the withdrawal if multiple are made with multicall
```
wilsoncusack, line 2251: I took the liberty to move these up into `param`, can revert if not wanted

packages/sdk/example.env
```diff
@@ -0,0 +1,2 @@
+VITE_E2E_RPC_URL_L1=https://ethereum-goerli.publicnode.com
```
roninjin10, line 3:
```suggestion
// public rpcs are heavily throttled/rate limited so replace these with rpcs with apikeys. These are meant to be testnet rpcs
// in future these will get renamed to VITE_E2E_RPC_URL_GOERLI etc.
VITE_E2E_RPC_URL_L1=https://ethereum-goerli.publicnode.com
```

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -326,13 +326,11 @@ export class CrossChainMessenger {
    * Transforms a legacy message into its corresponding Bedrock representation.
    *
    * @param message Legacy message to transform.
+   * @param messageIndex The index of the message, if multiple exist from multicall
    * @returns Bedrock representation of the message.
```
roninjin10, line 330: Not from this pr but we should move away from using jsdoc comments because typescript can handle this better

packages/sdk/src/cross-chain-messenger.ts
roninjin10: In future will get rid of jsdoc comments but this is more consistent with now so I think this is a nice change

packages/sdk/test-next/messageStatus.spec.ts
roninjin10: Correct without a .spec they were never running and we thought they were. Thank you so much for this fix!

packages/sdk/test-next/messageStatus.spec.ts
roninjin10: I think this is correct.  Within this block range it is READY_FOR_RELAY so even though it's slightly confusing it feels correct to me. You could imagine using the filters to debug the state of a tx in a block-range even though you know it's state changed more recently.

It's not perfect though because it can still return the wrong state if the event happened BEFORE not after your blockrange but we will deal with that when we add better documentation

roninjin10 (approved):
image

packages/sdk/test-next/messageStatus.spec.ts
wilsoncusack: actually within the given block range it doesn't even exist 😬 , but up to you what to resolve here vs. follow up

wilsoncusack:
@roninjin10 could we get this merged pls 🙏

roninjin10 (approved):

PR #6333, opened 2023-07-18
fix(op-challenger): Agent Stepping

refcell:
**Description**

Fixes the op-challenger Agent step logic to only step on claims for which we don't agree with their level.

To prevent continuously submitting duplicate step transactions, the Agent now checks the newly added `Countered` field on the `Claim` which is set to true on the leaf node iff it has been stepped on through the FaultDisputeGame contract.

Additionally, this PR "loads" the `Clock` field into the `Claim` struct so a follow on PR can build off the `Clock` to perform Alphabet Fault Dispute Game resolution.

A demonstration running `charlie.sh` and `mallory.sh` scripts with a populated alphabet fault dispute game on devnet is depicted below.

Screenshot 2023-07-18 at 3 48 05 PM

**Tests**

The Loader unit tests required alteration to accomodate the affixed `Clock` and `Countered` `Claim` fields.

### DIFF
--- a/op-challenger/fault/agent.go
+++ b/op-challenger/fault/agent.go
@@ -94,6 +94,17 @@ func (a *Agent) step(claim Claim, game Game) error {
 	if claim.Depth() != a.maxDepth {
 		return nil
 	}
+
+	if game.AgreeWithClaimLevel(claim) {
+		a.log.Warn("Agree with leaf claim, skipping step", "claim_depth", claim.Depth(), "maxDepth", a.maxDepth)
+		return nil
+	}
+
+	if claim.Countered {
+		a.log.Info("Claim already stepped on", "claim_depth", claim.Depth(), "maxDepth", a.maxDepth)
+		return nil
+	}
+
 	a.log.Info("Attempting step", "claim_depth", claim.Depth(), "maxDepth", a.maxDepth)
 	step, err := a.solver.AttemptStep(claim)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


refcell:
* develop
  * **PR #6333** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6333" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6333?utm_source=stack-comment).

op-challenger/fault/agent.go
```diff
@@ -94,6 +94,17 @@ func (a *Agent) step(claim Claim, game Game) error {
...
 		return nil
 	}
+
+	if game.AgreeWithClaimLevel(claim) {
+		a.log.Warn("Agree with leaf claim, skipping step", "claim_depth", claim.Depth(), "maxDepth", a.maxDepth)
+		return nil
+	}
+
+	if claim.Countered {
+		a.log.Warn("Claim already stepped on", "claim_depth", claim.Depth(), "maxDepth", a.maxDepth)
```
ajsutton: I don't think this are unusual enough to be a `Warn` - probably should just be `Info` level.

op-challenger/fault/loader.go
```diff
@@ -56,6 +56,8 @@ func (l *loader) fetchClaim(ctx context.Context, arrIndex uint64) (Claim, error)
 			Value:    fetchedClaim.Claim,
 			Position: NewPositionFromGIndex(fetchedClaim.Position.Uint64()),
 		},
+		Countered:           fetchedClaim.Countered,
```
ajsutton, line 59: I'm slightly nervous about starting to use the `Countered` value from the contract as it's one of the mutable fields so we're breaking our data access plan.  That said, at the moment we're not attempting to cache any game state and its looking likely that we won't need to so this isn't currently going to cause any issues.

Probably just worth a comment to note that we're reading a mutable field from the contract and will need to stop doing that if we ever start caching the game state and only processing new claims rather than starting from scratch each time.

ajsutton (approved):
LGTM, just a couple of minor nits.

PR #6343, opened 2023-07-19
chain-mon: migrate fault detector

tynes:
**Description**

This commit migrates the fault-detector service to the chain-mon package. This is useful because it moves towards standardizing all of the offchain monitoring services into a single typescript package, following the pattern that is already used in the repo.

This simplifies all of the linting and removes boilerplate, as the services in chain-mon are all very similar and use the same dependencies.

This also ports the tests from fault-detector to chain-mon, making it the first tests in the repo. Longer term we could likely remove the need to test using hardhat, but no need to do that for now. This will make it easier to upgrade to a newer version of hardhat as the version we are currently on only supports up to node v16 which is no longer LTS.

This change will save a bit of build time and a bit of time before each commit because it will be one less package that needs to be built or linted.


docs/op-stack/src/docs/releases/bedrock/README.md
```diff
@@ -29,6 +29,6 @@ Dive deep into the specifications for the Bedrock release in the [specs folder o
 - [`op-batcher`](https://github.com/ethereum-optimism/optimism/tree/develop/op-batcher)
 - [`op-proposer`](https://github.com/ethereum-optimism/optimism/tree/develop/op-proposer)
 - [`contracts-bedrock`](https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock)
-- [`fault-detector`](https://github.com/ethereum-optimism/optimism/tree/develop/packages/fault-detector)
+- [`fault-mon`](https://github.com/ethereum-optimism/optimism/tree/develop/packages/packages/chain-mon/src/fault-mon)
```
maurelian:
```suggestion
- [`fault-mon`](https://github.com/ethereum-optimism/optimism/tree/develop/packages/chain-mon/src/fault-mon)
```

maurelian (approved):

PR #6344, opened 2023-07-19
op-e2e: Support L1 time travel

ajsutton:
**Description**

Introduces `AdvancingClock` which is a clock that keeps moving forward at the same rate as real time (but in more discrete steps) but also supports time travelling arbitrary distances into the future.

That clock is then used in the `fakePoS` we already used to drive L1 geth so that we can cause it to time travel. Combined with an upper limit on how many blocks it tries to catch up with consistent times, we can now time travel an arbitrary distance into the future with L1 keeping up practically instantly.

Time travel is only enabled for tests that opt in to it in the system config, avoiding any potential slow downs or effects on non-time travel tests.

L2 doesn't time travel as nicely, but it was able to catch back up to a 24 hour jump in L1 time in under 3 minutes. For the dispute game we only need to time travel L1, but there's potential for this to be able to accelerate withdrawal tests given it would only need to jump forward by a few minutes to an hour.

### DIFF
--- a/op-e2e/fakepos.go
+++ b/op-e2e/fakepos.go
@@ -1,0 +1,124 @@
+package op_e2e
+
+import (
+	"time"
+
+	"github.com/ethereum-optimism/optimism/op-service/clock"
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/beacon/engine"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/eth"
+	"github.com/ethereum/go-ethereum/eth/catalyst"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// fakePoS is a testing-only utility to attach to Geth,
+// to build a fake proof-of-stake L1 chain with fixed block time and basic lagging safe/finalized blocks.
+type fakePoS struct {
+	clock     clock.Clock
+	eth       *eth.Ethereum
+	log       log.Logger
+	blockTime uint64
+
+	finalizedDistance uint64
+	safeDistance      uint64
+
+	engineAPI *catalyst.ConsensusAPI
+	sub       ethereum.Subscription
+}
+
+func (f *fakePoS) Start() error {
+	if advancing, ok := f.clock.(*clock.AdvancingClock); ok {
+		advancing.Start()
+	}
+	f.sub = event.NewSubscription(func(quit <-chan struct{}) error {
+		// poll every half a second: enough to catch up with any block time when ticks are missed
+		t := f.clock.NewTicker(time.Second / 2)
+		for {
+			select {
+			case now := <-t.Ch():
+				chain := f.eth.BlockChain()
+				head := chain.CurrentBlock()
+				finalized := chain.CurrentFinalBlock()
+				if finalized == nil { // fallback to genesis if nothing is finalized
+					finalized = chain.Genesis().Header()
+				}
+				safe := chain.CurrentSafeBlock()
+				if safe == nil { // fallback to finalized if nothing is safe
+					safe = finalized
+				}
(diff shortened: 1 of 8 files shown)


ajsutton:
Time travel adds a small overhead since it needs to setup a ticker to keep advancing the clock manually at specific intervals.  We probably want to make time travel support opt-in for L1 instances and just use a plain `SystemClock` when it's not required to avoid that. Will take a bit of jiggling but should be doable.
And obviously we need a better way to get access to the clock to advance time rather than that global variable...

op-service/clock/advancing.go
```diff
@@ -0,0 +1,71 @@
...
+	ticker       Ticker
+	advanceEvery time.Duration
+	quit         chan interface{}
+	running      atomic.Bool
+
+	lastTick time.Time
+}
+
+// NewAdvancingClock creates a clock that, when started, advances at the same rate as the system clock but
+// can also be advanced arbitrary amounts using the AdvanceBy method.
```
refcell: I think this is referencing the `AdvanceTime` method in `DeterministicClock`
```suggestion
// can also be advanced arbitrary amounts using the AdvanceTime method.
```

refcell (commented):
Very nice, LGTM

op-e2e/fakepos.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/eth"
+	"github.com/ethereum/go-ethereum/eth/catalyst"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// fakePoS is a testing-only utility to attach to Geth,
+// to build a fake proof-of-stake L1 chain with fixed block time and basic lagging safe/finalized blocks.
+type fakePoS struct {
```
tynes, line 18: This is a second impl of fake pos in this package?

protolambda (approved):
Generally looks good, but since the L2 driver is still based on a real clock the L2 side might behave weird when it doesn't advance like the L1: e.g. it may catch up to the L1, but with terrible L2 block-building timing.

ajsutton:
> Generally looks good, but since the L2 driver is still based on a real clock the L2 side might behave weird when it doesn't advance like the L1: e.g. it may catch up to the L1, but with terrible L2 block-building timing.

I did a test with the L2 side - you're right, it's not particularly pretty when the L1 skips a long way into the future though it caught up again eventually. We'll need to investigate further before we try to time travel L2 in tests. It may be that we skew the L1 time more slowly over a number of blocks rather than a direct jump for example.  We also don't need to jump too far for withdrawals anyway because we can configure the settings to make withdrawals fast which should help.

trianglesphere:
> We'll need to investigate further before we try to time travel L2 in tests

It'd be cool run both the L1 & the L2 on the same deterministic clock. I guess we'd have to figure out what that skew rate should be (i.e. advance, then how long should we wait before doing another advance) if we don't want to do a huge advance all at once.

ajsutton:
> > We'll need to investigate further before we try to time travel L2 in tests
> 
> It'd be cool run both the L1 & the L2 on the same deterministic clock. I guess we'd have to figure out what that skew rate should be (i.e. advance, then how long should we wait before doing another advance) if we don't want to do a huge advance all at once.

I think we need to understand what's going on better than I currently do really.  It's possible we'll have to use the `Clock` abstraction everywhere in op-node and we should be aiming for that eventually anyway for unit testing but that's a lot of work and maybe we don't need it.  L2 blocks have to base their concept of time off the L1 timestamp so theoretically it shouldn't matter if op-node thinks the L1 time is wildly out of sync, it should just follow it and everything works but we wouldn't have ever tested cases where L1 time is that far out of sync with reality, let alone if it's moving at a different pace to the system clock. And with PoS on L1 now, I guess that's not really a real world risk because Ethereum L1 breaks if it's time drifts from reality even by a couple of seconds.

PR #6345, opened 2023-07-19
feat(indexer): move into the monorepo go module

hamdiallam:
The indexer consumes the monorepo as a dependency. Lets update the indexer to be apart of the monorepo so it can use the latest and greatest code

Current indexer code depends on op-node/withdrawals which causes chain dependencies on op-e2e/e2eutils and op-chain-ops which is unnecessary.
 - `op-e2e/e2eutils/waits.go` is a useful enough utility to lift it into op-service. This alleviates the dependency issue


hamdiallam:
Hm, looks like i need to fix some of the circleci configuration. Will address this tomorrow

indexer/services/l2/bridge/standard_bridge.go
```diff
@@ -45,6 +44,11 @@ func (s *StandardBridge) GetWithdrawalsByBlockRange(ctx context.Context, start,
 		return nil, err
 	}
 
+	l2L1MpABI, err := bindings.L2ToL1MessagePasserMetaData.GetAbi()
```
hamdiallam: Using the ABI removes the dependency on op-node/withdrawals

indexer/services/l2/bridge/standard_bridge.go
```diff
@@ -76,11 +80,7 @@ func (s *StandardBridge) GetWithdrawalsByBlockRange(ctx context.Context, start,
 				}
 			}
 
-			hash, err := withdrawals.WithdrawalHash(withdrawalInitiated)
-			if err != nil {
-				return nil, err
-			}
-
+			hash := common.Hash(withdrawalInitiated.WithdrawalHash)
```
hamdiallam: This is also available in the event. Rederiving it is cool but the extra deps for it isn't worth it

tynes:
> However I still think op-node/withdrawals has useful utilities. @tynes @mslipper thoughts on moving op-e2e/e2eutils/waits.go into op-service? That's the only utility that op-node/withdrawals is using that's useful to extract out

Definitely on board with moving this to `op-service`, seems like it could be used in many places

indexer/op-legacy/predeploys.go
```diff
@@ -0,0 +1,11 @@
+package op_legacy
+
+import "github.com/ethereum/go-ethereum/common"
+
+const (
+	LegacyERC20ETH = "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000"
```
tynes: Not the end of the world but did like having this in `op-bindings` near the other predeploy addresses

indexer/services/l2/bridge/standard_bridge.go
tynes: You could hash the withdrawal locally and compare it against the event's hash to be sure that you are using software that supports the correct version of withdrawal serialization but maybe its not useful for generalized indexing, this is more useful for software that builds proofs

indexer/op-legacy/predeploys.go
hamdiallam: ah i can definitely put it back in there. It was removed so I wasn't 100% sure

indexer/services/l2/bridge/standard_bridge.go
hamdiallam: Yeah i agree the integrity check is good. I'll probably utilize this in the new indexer code pathway. This change is reverted now that waits.go is in op-service

indexer/op-legacy/predeploys.go
hamdiallam: moved it back

hamdiallam:
> > However I still think op-node/withdrawals has useful utilities. @tynes @mslipper thoughts on moving op-e2e/e2eutils/waits.go into op-service? That's the only utility that op-node/withdrawals is using that's useful to extract out
> 
> Definitely on board with moving this to `op-service`, seems like it could be used in many places

Moved it into `op-service/client/utils`. Lmk it makes sense to just have it be top-level within client

op-bindings/predeploys/addresses.go
```diff
@@ -56,6 +58,7 @@ func IsProxied(predeployAddr common.Address) bool {
 }
 
 func init() {
+	Predeploys["LegacyERC20ETHAddr"] = &LegacyERC20ETHAddr
```
tynes: Sorry if my comment was unclear, it can't go in the mapping below, I think it makes sense to put it in a different file, could be in this package tho

hamdiallam: no worries, i just realized when ci ran (also TIL the eth address is not same post-bedrock). Adjusting it rn

op-bindings/predeploys/legacy_addresses.go
```diff
@@ -0,0 +1,11 @@
+package predeploys
+
+import "github.com/ethereum/go-ethereum/common"
+
+const (
+	LegacyERC20ETH = "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000"
```
roninjin10, line 6: Just a comment I'm sure this is just following convention, but seems unnecessary and also slightly confusing to make this one public

roninjin10 (approved):

op-bindings/predeploys/legacy_addresses.go
hamdiallam: It needs to be public so that other packges can reference it. Otherwise we can't reference it from the indexer package

protolambda (approved):

op-bindings/predeploys/legacy_addresses.go
roninjin10: Why would we need to reference this though instead of the geth Address object thing?  Doesn't the geth Address object have access to this string?

zhwrd (approved):

op-bindings/predeploys/legacy_addresses.go
hamdiallam: ahh i see what you're saying, yea the convention in the other file follows this format. And I think the constant probably gets inlined on builds whenever it's used making it preferrable for arguments that take the string form

PR #6302, opened 2023-07-15
(feat): make supportsInterface function on OptimismMintableERC20 virtual

KNWR:
**Description**

Let's say I want to extend OptimismMintableERC20 to support additional functionality with corresponding interfaces. Currently if I want to extend the contract and have the `supportsInterface` support additional interfaces I'm adding, I can't do that. This change makes it possible by marking`supportsInterface` as virtual, so that other contracts can inherit from OptimismMintableERC20 and register as supporting additional interfaces by including a `supportsInterface` method marked as `override`. 

**Tests**

No additional tests needed as this doesn’t affect runtime functionality. Tests still passing.

PR #6348, opened 2023-07-19
feat: make mintable erc20 inheritable

tynes:
**Description**

Taking over https://github.com/ethereum-optimism/optimism/pull/6302

> Let's say I want to extend OptimismMintableERC20 to support additional functionality with corresponding interfaces. Currently if I want to extend the contract and have the `supportsInterface` support additional interfaces I'm adding, I can't do that. This change makes it possible by marking`supportsInterface` as virtual, so that other contracts can inherit from OptimismMintableERC20 and register as supporting additional interfaces by including a `supportsInterface` method marked as `override`.


maurelian (approved):

KNWR:
Thanks @tynes !

PR #6349, opened 2023-07-19
feat(op-challenger): EthCall Alphabet FDG Resolve

refcell:
**Description**

Introduces a `Resolver` fault subcomponent for the op-challenger agent to call the `resolve` method of the FaultDisputeGame.

This is the alternative method to checking if the challenger can resolve a dispute game instead of tracking the leftmost,
uncontested dispute game node and its clock.

**Tests**

Unit tests are performed using the minimal `ContractCaller` interface that implments `CallContext`.

**Metadata**

Fixes CLI-4232

### DIFF
--- a/op-challenger/fault/resolver.go
+++ b/op-challenger/fault/resolver.go
@@ -1,0 +1,57 @@
+package fault
+
+import (
+	"context"
+	"math/big"
+
+	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// ContractCaller is a minimal interface of [ethclient.Client] to allow for mocking.
+type ContractCaller interface {
+	CallContract(ctx context.Context, msg ethereum.CallMsg, blockNumber *big.Int) ([]byte, error)
+}
+
+type Resolver struct {
+	ContractCaller
+	fdgAddr common.Address
+	fdgAbi  *abi.ABI
+}
+
+// NewResolver creates a new [Resolver] instance.
+func NewResolver(client ContractCaller, fdgAddr common.Address) (*Resolver, error) {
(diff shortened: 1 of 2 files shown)


refcell:
* develop
  * **PR #6349** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6349" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6349?utm_source=stack-comment).

op-challenger/fault/resolver.go
```diff
@@ -0,0 +1,57 @@
...
+		return false, err
+	}
+	_, err = r.CallContract(ctx, ethereum.CallMsg{
+		To:   &r.fdgAddr,
+		Data: resolveData,
+	}, nil)
+	if err != nil {
+		return false, err
+	}
+	return true, nil
```
ajsutton, line 56: Shouldn't this be going via the tx manager like when we call claims? It seems to be a very different pattern to how we're doing calls in `faultResponder`.

Given `faultResponder` already has the ability to build attack, defend and step calls it seems like maybe we should be just adding the ability to call `resolve` in there so we have one place that's responsible for dispute game contract calls?

ajsutton (commented):
You may have a grander vision than I have so happy to talk through how this should fit together and will go look at your other PRs, but my intuition is we should just make this part of the existing `faultResponder`.

op-challenger/fault/resolver.go
trianglesphere: I'm not 100% sure if we should be going through the tx manager for performing the `eth_call`, but we should be using `buildResolveData()` and then doing an `eth_call`. Maybe we extend the txmanager to be able to do that `eth_call` b/c it already has the L1 connection or we can pass in a L1 client from somewhere.

refcell: Refactored the `eth_call` to pass through the `TxManager` in #6371 - let me know what you think there. Will close this PR out now since it's sufficiently outdated.

refcell:
Closing in favor of #6371

PR #6358, opened 2023-07-19
feat(proxyd): sender rate limiter should check chain id

felipe-op:
**Description**

This adds the ability to check the transaction chain id before rate limiting the sender. 

**Tests**

Added new test cases.

**Metadata**

- Fixes https://linear.app/optimism/issue/INF-343/address-proxyd-dos


maurelian (approved):

proxyd/server.go
```diff
@@ -674,6 +685,18 @@ func (s *Server) rateLimitSender(ctx context.Context, req *RPCReq) error {
 	return nil
 }
 
+func (s *Server) isAllowedChainId(chainId *big.Int) bool {
+	if s.allowedChainIds == nil || len(s.allowedChainIds) == 0 {
+		return true
+	}
+	for _, id := range s.allowedChainIds {
```
mslipper, line 692: [nit] most of the time `s.allowedChainIds` will be just one element, so we can save some CPU cycles by checking `if len(s.allowedChainIds) == 1 && s.allowedChainIds[0].Cmp(id) == 0` outside the for loop.

felipe-op: the loop is short circuited, so it will return as soon as the first chain id matches
as we know, this array will always with one element or be super short

proxyd/server.go
```diff
@@ -674,6 +685,18 @@ func (s *Server) rateLimitSender(ctx context.Context, req *RPCReq) error {
 	return nil
 }
 
+func (s *Server) isAllowedChainId(chainId *big.Int) bool {
+	if s.allowedChainIds == nil || len(s.allowedChainIds) == 0 {
```
tynes, line 689: I believe `len(nil)  == 0` so its a bit redundant to check both

mslipper (approved):

PR #6362, opened 2023-07-20
op-challenger: Introduce cannon trace provider

ajsutton:
**Description**

Currently only loads existing proofs and can't run cannon to generate them. Also not yet hooked in - will need a follow up PR to add CLI options to select the cannon trace provider instead of the alphabet one.

**Tests**

Unit tests added for new provider.

**Metadata**

- https://linear.app/optimism/issue/CLI-4225/cannon-trace-provider-load-existing-proof-files


refcell (approved):
LGTM

PR #6363, opened 2023-07-20
specs: span-batch spec fixes

protolambda:
**Description**

Fixes issues in the span-batches spec, found by @testinprod-io @ImTei

Fixes:
- Correct `l1_origin` reference of span-batch to point to the end of the span, not the start.
- Check the span `future` case by comparing the start, not the end. And Flip the `>` to verify it's past the L1 origin of the L2 safe head (the `epoch`).
  - Based on further review: add `+1` to the L2 safe-head L1 origin, to handle the case where the start of the span continues with the L1 block after the L1 origin of the L2 safe head.
- Check the span past/duplicate case by checking if the span is older than the L1 origin of the L2 safe head (the `epoch`)
  - Based on further review: use the start of the span, and not the end, to detect when a span partially overlaps older already processed block data. The span is not guaranteed to reference the same chain, and consistency-checks of the resulting partially valid span would require access to more of the L2 chain history, so we drop the span instead. This matches the behavior of ignoring partially invalid spans that's already in the spec.
- Based on further review: make sure the L1 origin check of the span-batch is explicitly checked in the span-batch validation rules. And use the relationship to the inclusion in the L1 chain to drop span-batches that reference impossible L1 blocks.

Changes:
- change `anchor` to `span_start` for consistency with `span_end` name.

Update: continued spec improvements by @ImTei 
- Updated span-batch encoding
- Fixed derivation rules,  also updated to match new encoding
- Added section about optimization strategy

### DIFF
--- a/specs/span-batches.md
+++ b/specs/span-batches.md
@@ -64,7 +64,8 @@ Introduce version `1` to the [batch-format](./derivation.md#batch-format) table:
 
 Notation:
 `++`: concatenation of byte-strings.
-`anchor`: first L2 block in the span
+`span_start`: first L2 block in the span
+`span_end`: last L2 block in the span
 `uvarint`: unsigned Base128 varint, as defined in [protobuf spec]
 
 [protobuf spec]: https://protobuf.dev/programming-guides/encoding/#varints
@@ -72,11 +73,11 @@ Notation:
 Where:
 
 - `prefix = rel_timestamp ++ parent_check ++ l1_origin_check`
-  - `rel_timestamp`: relative time since genesis, i.e. `anchor.timestamp - config.genesis.timestamp`.
-  - `parent_check`: first 20 bytes of parent hash, i.e. `anchor.parent_hash[:20]`.
+  - `rel_timestamp`: relative time since genesis, i.e. `span_start.timestamp - config.genesis.timestamp`.
+  - `parent_check`: first 20 bytes of parent hash, i.e. `span_start.parent_hash[:20]`.
   - `l1_origin_check`: to ensure the intended L1 origins of this span of
         L2 blocks are consistent with the L1 chain, the blockhash of the last L1 origin is referenced.
-        The hash is truncated to 20 bytes for efficiency, i.e. `anchor.l1_origin.hash[:20]`.
+        The hash is truncated to 20 bytes for efficiency, i.e. `span_end.l1_origin.hash[:20]`.
 - `payload = block_count ++ block_tx_counts ++ tx_data_headers ++ tx_data ++ tx_sigs`:
   - `block_count`: `uvarint` number of L2 blocks.
   - `origin_bits`: bitlist of `block_count` bits, right-padded to a multiple of 8 bits:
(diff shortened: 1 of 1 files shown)


ajsutton (approved):
Maks sense to me. The docker build looks like an intermittent failure to download dependencies.

specs/span-batches.md
```diff
@@ -149,11 +150,16 @@ Span-batch rules, in validation order:
   - Rules:
     - `start_epoch_num + sequence_window_size < inclusion_block_number` -> `drop`:
       i.e. the batch must be included timely.
-    - `end_epoch_num < epoch.number` -> `future`: i.e. all referenced L1 epochs must be there.
-    - `end_epoch_num == epoch.number`:
-      - If `batch.l1_origin_check != epoch.hash[:20]` -> `drop`: verify the batch is intended for this L1 chain.
-    - `end_epoch_num > epoch.number` -> `drop`: must have been duplicate batch,
-      we may be past this L1 block in the safe L2 chain.
+    - `start_epoch_num > epoch.number + 1` -> `future`:
```
ImTei: This condition is always false. 

Definitions:
`start_epoch_num = safe_l2_head.origin.block_number + (origin_changed_bit ? 1 : 0)`.
`epoch = safe_l2_head.l1_origin` (From [derivation specs](https://github.com/ethereum-optimism/optimism/blob/develop/specs/derivation.md?plain=1#L584))

Lets plug in each variable to the inequality.

`start_epoch_num > epoch.number + 1`
becomes
`safe_l2_head.origin.block_number + (origin_changed_bit ? 1 : 0) > safe_l2_head.l1_origin.number + 1`
which can never be satisfied.

specs/span-batches.md
```diff
@@ -149,11 +150,16 @@ Span-batch rules, in validation order:
...
+      i.e. the L1 origin may jump to the next L1 block, but not further.
+    - If `end_epoch_num >= inclusion_block_number` -> `drop`:
+      if the end of the span is past the L1 block it was included in,
+      it cannot possibly reference the chain it was included in, and is thus non-canonical.
+    - If `batch.l1_origin_check` does not match the canonical L1 chain at `end_epoch_num` -> `drop`:
+      verify the batch is intended for this L1 chain.
+    - `start_epoch_num < epoch.number` -> `drop`: must have been duplicate batch,
```
ImTei, line 277: Also here.

`start_epoch_num < epoch.number`
becomes
`safe_l2_head.origin.block_number + (origin_changed_bit ? 1 : 0) < safe_l2_head.l1_origin.number`

which can never be satisfied.

ImTei (commented):
Previous v0 batch included epoch number. 
Current v1 batch does not include the epoch number, but the spec derives the epoch number from the safe head with the assumption that new batch's epoch is continued from the safe head(`start_epoch_num = safe_l2_head.origin.block_number + (origin_changed_bit ? 1 : 0)`). 

The following rules are not proper as I mentioned because of the change.

Epoch number should be included in v1 span batch to check the continuity of epoch number.
Or, we can find L1 block number from L1 node using the hash. 

What do you think?

tynes:
Just wondering if this is good to merge?

protolambda:
@tynes Thanks for the reminder. We can merge it and defer remaining review changes to a separate PR, but I think I prefer to update this one, once I have time to properly review it (I lost track of it while being swamped with other work).

refcell:
(rebased on develop to fix build issues)

protolambda:
Rebased on `develop` and added latest spec-fixes from https://github.com/testinprod-io/optimism/pull/6

protolambda:
Updated to fix linting and table-of-contents

PR #6366, opened 2023-07-20
op-e2e: Add helper to run op-challenger in e2e tests

ajsutton:
**Description**

Enhance test by running a challenger that defends the output and confirm it wins. Still only has the root claim and a single counter-claim before resolving the game.

Modifies op-challenger `Main` to take a `Context` so that it can be stopped instead of running forever.

**Metadata**

- https://linear.app/optimism/issue/CLI-4133/e2e-tests

### DIFF
--- a/op-challenger/challenger.go
+++ b/op-challenger/challenger.go
@@ -1,6 +1,7 @@
 package op_challenger
 
 import (
+	"context"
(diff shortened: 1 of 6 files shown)


refcell (approved):
LGTM

PR #6369, opened 2023-07-20
Fix Alphabet game pre-state

trianglesphere:
**Description**

I accidentally used the number "140" to the character "`". 140 is in octal & it should be 96 in decimal.


refcell (approved):
Makes sense

PR #6373, opened 2023-07-20
feat(ctb): Local preimage key support

clabby:
## Overview

> **Note**
> Several tests will fail here until the necessary changes to the `op-program` have been introduced.

V2 of #6264, covers the points in the design doc written for local data handling.

**Metadata**
Closes CLI-4254


clabby:
* develop
  * **PR #6373** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6373" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6373?utm_source=stack-comment).

packages/contracts-bedrock/contracts/dispute/FaultDisputeGame.sol
```diff
@@ -378,6 +418,9 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
                 countered: false
             })
         );
+
+        // Set the L1 head hash at the time of the game's creation.
+        l1Head = Hash.wrap(blockhash(block.number - 1));
```
clabby, line 429: Note: This will need to change, but can be done in another PR.

refcell: Good candidate for stacking 😉

packages/contracts-bedrock/contracts/cannon/PreimageOracle.sol
```diff
@@ -1,15 +1,15 @@
...
+pragma solidity 0.7.6;
+
+import { PreimageKeyLib } from "./PreimageKeyLib.sol";
 
 /// @title PreimageOracle
 /// @notice A contract for storing permissioned pre-images.
 contract PreimageOracle {
     /// @notice Mapping of pre-image keys to pre-image lengths.
     mapping(bytes32 => uint256) public preimageLengths;
-
```
refcell: Why do we delete these lines, it makes it a lot less legible?

clabby: I'm of the opposite opinion, but nbd on my end. You want the space between vars back?

cannon/mipsevm/witness.go
```diff
@@ -49,18 +49,19 @@ func (wit *StepWitness) EncodePreimageOracleInput() ([]byte, error) {
...
 		input = append(input, wit.PreimageKey[:]...)
+
+		preimagePart := wit.PreimageValue[wit.PreimageOffset:]
 		var tmp [32]byte
-		copy(tmp[:], wit.PreimageValue[wit.PreimageOffset:])
+		copy(tmp[:], preimagePart)
 		input = append(input, tmp[:]...)
-		input = append(input, uint32ToBytes32(uint32(len(wit.PreimageValue))-8)...)
+		input = append(input, uint32ToBytes32(32)...)
+		input = append(input, uint32ToBytes32(wit.PreimageOffset)...)
```
Inphi: Just want to note that while this isn't correct, the other PR fixes this.

Inphi (approved):

tynes:
Need to regen bindings and contracts lint

```
src/dispute/FaultDisputeGame.sol

║ Line     │ Column   │ Type     │ Message                                                │ Rule ID              ║
╟──────────┼──────────┼──────────┼────────────────────────────────────────────────────────┼──────────────────────╢
║ 18       │ 1        │ warning  │ global import of path                                  │ no-global-import     ║
║          │          │          │ ../libraries/DisputeTypes.sol is not allowed.          │                      ║
║          │          │          │ Specify names to import individually or bind all       │                      ║
║          │          │          │ exports of the module into a name (import "path"       │                      ║
║          │          │          │ as Name)                                               │                      ║
║ 19       │ 1        │ warning  │ global import of path                                  │ no-global-import     ║
║          │          │          │ ../libraries/DisputeErrors.sol is not allowed.         │                      ║
║          │          │          │ Specify names to import individually or bind all       │                      ║
║          │          │          │ exports of the module into a name (import "path"       │                      ║
║          │          │          │ as Name)                                               │                      ║
║ 68       │ 2        │ error    │ Line length must be no more than 100 but current       │ max-line-length      ║
║          │          │          │ length is 103.    
```

tynes (approved):

clabby:
@mergifyio requeue

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -241,6 +257,38 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
         move(_parentIndex, _claim, false);
     }
 
+    /// @inheritdoc IFaultDisputeGame
+    function addLocalData(uint256 _ident, uint256 _partOffset) external {
```
ajsutton, line 261: The challenger will need to be able to access each piece of data this can load so would be nice to have getters for them exposed. I think the key ones are the l2 output roots for the agreed starting point and the claim being challenged.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -241,6 +257,38 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
...
+        } else if (_ident == 2) {
+            // Load the earliest output root that commits to the passed L2 block number
+            // into the game's local context in the preimage oracle.
+            Types.OutputProposal memory proposal = L2_OUTPUT_ORACLE.getL2OutputAfter(
+                l2BlockNumber()
+            );
+            oracle.loadLocalData(_ident, proposal.outputRoot, 32, _partOffset);
+        } else if (_ident == 3) {
+            // Load the root claim into the game's local context in the preimage oracle.
+            oracle.loadLocalData(_ident, Claim.unwrap(rootClaim()), 32, _partOffset);
```
ajsutton, line 278: I think this is the wrong value to load - the root claim of the dispute game is the final state of the cannon VM, but the l2.claim we need in the pre-image oracle is the output root that is being challenged by the game.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -241,6 +257,38 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
...
+        IPreimageOracle oracle = VM.oracle();
+        if (_ident == 1) {
+            // Load the L1 head hash into the game's local context in the preimage oracle.
+            oracle.loadLocalData(_ident, Hash.unwrap(l1Head), 32, _partOffset);
+        } else if (_ident == 2) {
+            // Load the earliest output root that commits to the passed L2 block number
+            // into the game's local context in the preimage oracle.
+            Types.OutputProposal memory proposal = L2_OUTPUT_ORACLE.getL2OutputAfter(
+                l2BlockNumber()
+            );
```
ajsutton, line 274: Shouldn't this be the last output prior to the l2 block number rather than the first after it? Note that this is the agreed l2 output root to start execution from. I'm a little unclear exactly what the l2 block number being passed in is representing here though - is it what's being challenged or the agreed starting point?

Ultimately we want cannon to run from the output root immediately prior to the one being challenged. If you disagree with the immediately prior root you should just challenge it instead of the later one as they'll then both be removed. I wonder if it's easier to just pass in the output root index being challenged from the oracle and then we can load the claim being challenged, the block number its from and the prior claim all from the output oracle.

ajsutton: Ah, `getL2OutputAfter` is after or equal to the block number.  So if the passed in block number is the agreed starting point, then this would work.  The claimed output root should then be from the output immediately after this.

It would be more future proof to specify the output being challenged by index and so that when outputs are permissionless it can handle having multiple conflicting outputs for the same block number and get the "parent" of the claim as the starting point since there may be multiple competing children of an output.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
ajsutton: On second thoughts, it's not unreasonable for the challenger to just retrieve the data from the L2OO itself rather than having to expose getters here.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
ajsutton: ah, the catch with using the output index is that if output roots are deleted the indices may be reused. That would be ok if the deletion checks the output root matches but not ok if we are only loading the data on demand in either the game contract or the challenger since the index may have changed in the meantime.

Using block number has the same issue though since `getL2OutputAfter` could return a different output root to what it did when the root claim was calculated.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
clabby: The idea was that we would expose getters for the data within the dispute game (`rootClaim`, `l2BlockNumber`, `l1Head`, etc.) and data like the output root can be fetched through the oracle or the local `op-node`'s `optimism_outputAtBlock` endpoint.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
clabby: Hm this is interesting, should probably be addressed in another PR. The `L2OutputOracle` probably needs a rework that will allow for this - great points. Ideally we do away with the array and move to a mapping or other data structure. This could future proof it for permissionless output proposals as well.

Alternatively we can add the deletion check if we don't want to dig into the L2OO for the alpha - this was the current plan. Prior to deleting the output root, we'd check if the output root at the given index has changed, and if so, the game would be invalidated. We'd need to persist the output root on game creation rather than lazily fetching it in this function, but that's no problem.

PR #6371, opened 2023-07-20
feat(op-challenger): Responder CallResolve

refcell:
**Description**

This PR replaces #6349 by using the `TxManager`'s `Call` method exposed by
the base pr of this stack.

A `CallResolve` function is added to the `Responder` fault subcomponent of
the op-challenger in this pr which `eth_call`'s the `resolve()` function
of the fault dispute game to check if the `Resolve` tx will succeed or not.

This is the alternative method to checking if the challenger can resolve a
dispute game instead of tracking the leftmost, uncontested dispute game node
and its clock.

**Tests**

Unit tests are performed using the minimal `mockTxManager` interface that
mocks `Call`.

**Metadata**

Fixes CLI-4232

PR #6370, opened 2023-07-20
feat(op-service): TxManager Call Method

refcell:
**Description**

This is the base PR in a stack to replace PR #6349.
It is rebuilt following comments on that pr.

It introduces a `Call` method on the `TxManager` interface in order to pass through
a `CallContract` onto the `TxManager`'s internal backend. This is done so that the
op-challenger can check if a tx resolving a dispute game will succeed or not.

**Tests**

TODO

PR #6374, opened 2023-07-20
feat(op-challenger): Top Level Status and Resolution

refcell:
**Description**

Performs fault dispute game resolution.

**Metadata**

Fixes CLI-4250


refcell:
* develop
  * **PR #6370** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6370" target="_blank">Graphite</a> 
    * **PR #6371** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6371" target="_blank">Graphite</a> 
      * **PR #6374** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6374" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6374?utm_source=stack-comment).

op-challenger/challenger.go
```diff
@@ -48,8 +49,15 @@ func Main(logger log.Logger, cfg *config.Config) error {
 
 	for {
 		logger.Info("Performing action")
-		_ = agent.Act()
-		caller.LogGameInfo()
+		_ = agent.Act(context.Background())
+		status, _ := caller.GetGameStatus(context.Background())
+		if status != 0 {
+			logger.Info("Game ended!")
+			caller.LogGameStatus()
```
ajsutton: nit: We probably don't need to log both game ended and the status since the status will include the game state. Looking at your demo video, challenger logging is clearly way too verbose currently. :)

op-challenger/fault/agent.go
```diff
@@ -50,6 +53,18 @@ func (a *Agent) Act() error {
 	return nil
 }
 
+// resolveAndExit resolves the game if it is in a terminal state
+// and returns true if the game resolves successfully.
+func (a *Agent) resolveAndExit(ctx context.Context) bool {
```
ajsutton: maybe this should be `tryResolve`? It doesn't actually exit so not sure that should be part of the name and it doesn't always resolve so that should be reflected in the name somehow.

ajsutton (approved):
LGTM. Just a couple of minor suggestions.

op-challenger/fault/agent.go
```diff
@@ -29,7 +29,10 @@ func NewAgent(loader Loader, maxDepth int, trace TraceProvider, responder Respon
 }
 
 // Act iterates the game & performs all of the next actions.
-func (a *Agent) Act() error {
+func (a *Agent) Act(ctx context.Context) error {
+	if a.tryResolve(ctx) {
+		return nil
+	}
 	game, err := a.newGameFromContracts(context.Background())
```
tynes: any reason we don't pass thru the `ctx` that is passed as an arg here and instead make a new background context?

op-challenger/fault/agent.go
```diff
@@ -50,6 +53,18 @@ func (a *Agent) Act() error {
 	return nil
 }
 
+// tryResolve resolves the game if it is in a terminal state
+// and returns true if the game resolves successfully.
+func (a *Agent) tryResolve(ctx context.Context) bool {
+	if a.responder.CanResolve(ctx) {
+		if err := a.responder.Resolve(ctx); err != nil {
```
tynes: The error context is lost here, perhaps it could be logged as an error?

op-challenger/fault/agent.go
refcell: Oops, good catch!

tynes (approved):

op-challenger/challenger.go
```diff
@@ -48,8 +49,14 @@ func Main(logger log.Logger, cfg *config.Config) error {
 
 	for {
 		logger.Info("Performing action")
-		_ = agent.Act()
-		caller.LogGameInfo()
+		_ = agent.Act(context.Background())
+		status, _ := caller.GetGameStatus(context.Background())
+		if status != 0 {
+			caller.LogGameStatus()
+			return nil
```
tynes, line 56: wait - should this return here?

ajsutton: Yes. At this stage once the game is no longer in process the challenger should exit.

That won't last forever - once it supports multiple games it should carry on, but it doesn't support that yet.

refcell: Yea so right now the challenger is run on a specific instance of a fault dispute game - with the address passed in as a cli flag. So, right now we want to exit the binary once that game is resolved. This will be changed/refactored in the future.

PR #6375, opened 2023-07-20
cannon: Invalid STF for branch/jump in delay slots

Inphi:
Per the Cannon VM spec, implement well-defined behavior for branches and jumps in delay slots.


clabby (approved):
LGTM

Inphi:
@Mergifyio queue

Inphi:
@tynes I think I need your approval since this touches protocol critical code.

cannon/mipsevm/mips.go
```diff
@@ -246,6 +250,9 @@ func (m *InstrumentedState) handleHiLo(fun uint32, rs uint32, rt uint32, storeRe
 }
 
 func (m *InstrumentedState) handleJump(linkReg uint32, dest uint32) error {
+	if m.state.NextPC != m.state.PC+4 {
+		panic("jump in delay slot")
```
trianglesphere, line 254: why not return an error? Do we ever return a non-nil error?

Inphi: We don't ever return a non-nil error. It's not ideal, but we panic instead to isolate effects on the VM state. An invalid step may mess up the VM state.
We could save the pre-state and roll back when there's an invalid step, but doing so right now would be inefficient and double memory usage. Efficient VM snapshotting could be implemented but it's not a priority right now.

trianglesphere (approved):

PR #6383, opened 2023-07-20
specs: Clarify Cannon VM syscall behavior

Inphi:
**Description**

- Clarify the syscall behavior of the Cannon FPVM regarding non-caller-saved registers in Linux/MIPS.
- Replace duplicate sections in the Cannon README with references to the Cannon VM specs.
- Exclude Linux/MIPS linkref from markdown link check.


refcell (approved):
LGTM

protolambda (approved):

PR #6386, opened 2023-07-20
feat(ufm): add op-ufm to tag-service

felipe-op:
**Description**

This change adds op-ufm to ops tag-service used by github actions.


ops/tag-service/tag-service.py
```diff
@@ -10,12 +10,14 @@
 
 # Minimum version numbers for packages migrating from legacy versioning.
 MIN_VERSIONS = {
+    'ci-builder': '0.6.0'
+    'fault-detector': '0.6.3',
```
tynes: This should remove fault-detector as it is no longer a top level package but instead part of `chain-mon` since its very similar code, all of our typescript services that monitor the chain will will in `chain-mon` and use the same deps so that its easier to keep the deps up to date

tynes (approved):

mslipper (approved):

twoshark (approved):

PR #6387, opened 2023-07-20
contracts-bedrock: deploy EAS to optimism goerli

tynes:
**Description**

Deploys the EAS and SchemaRegistry contracts to optimism-goerli. Also makes slight modifications to the deployer and deploy config contracts for simplicity. In reality, we can remove some fields from the deploy config parsing that are not required but we first need better deploy config validation.


refcell (approved):
LGTM

PR #6388, opened 2023-07-21
specs: superchain target and protocol version

protolambda:
**Description**

This is specs change does not directly affect the state-transition function, but does enshrine the concept of a Protocol Version and a Superchain Target, to make future protocol upgrades more manageable & consistent across the growing number of different OP-Stack chains.

The Protocol Version helps version the specifications, and reason about canonical and experimental/proposed protocol specifications.

The Superchain Target unifies configuration and upgrades between chains. The mainnet Superchain Target is updated through protocol-level changes. Modded OP-Stack chains can subscribe to an isolated Superchain Target, but are encouraged to join the mainnet Superchain for increased interoperability.

This introduces the `ProtocolVersions` contract to the specs, a source of information to nodes on the superchain to use as recommendation for upgrades. The "`required`" signal indicates a protocol change is near or past activation, but is opt-in to nodes, and may be ignored if the user prefers to follow protocol changes through offchain signals. The `recommended` signal helps alert nodes in advance of upgrades.

**Metadata**

Fix CLI-4204


specs/superchain-upgrades.md
```diff
@@ -0,0 +1,199 @@
...
+This enables an iterative integration process into the canonical set of specifications,
+but does not guarantee the proposed specifications become canonical.
+
+Note that the Protocol Version only applies to the Protocol specifications with the Superchain Targets specified within.
+This versioning is independent of the [Semver] versioning used in OP Stack smart-contracts,
+and the [Semver]-versioned reference software of the OP-Stack.
+
+### Protocol Version Format
+
+The Protocol Version is [Semver]-compatible.
```
sebastianst, line 62: Now that we use sermver-compatible protocol versions instead of a simple integer, do we want to recommend different behavior of node implementations when e.g. the major vs minor version number is increased? For example, that we recommend halting the node if the supported major version is lower than the required, but not necessarily recommend this for minor version bumps.

tynes (approved):

tynes:
Looks good to me, makes sense. The endpoint is optional so nothing bad should happen if say op-reth doesn't have it implemented

PR #6389, opened 2023-07-21
contracts-bedrock: eas predeploy upgrade script

tynes:
**Description**

Quick script to be used for upgrading predeploys on op goerli.

Run with the following command:

```bash
forge script scripts/upgrades/EASUpgrade.s.sol --rpc-url $ETH_RPC_URL
```


refcell (approved):
LGTM

PR #6390, opened 2023-07-21
op-node: log error when op-node is configured with conflicting network and rollup-config

protolambda:
**Description**

The `--network=...` and `--rollup.config=...` CLI flags of the op-node are mutually exclusive. Log an error to warn the user that something is wrong with their configuration. This preserves the old behavior, to prevent existing CLI configurations from failing. In the future we can make it a hard error that stops the node.

Implements suggestion from @felipe-op


ajsutton (approved):
LGTM.

felipe-op (approved):
awesome!

PR #6391, opened 2023-07-21
op-e2e: Add tests with two op-challenger instances completing the entire dispute game.

ajsutton:
**Description**

Adds e2e tests to confirm an alphabet dispute game with two op-challengers will always resolve in favour of the correct alphabet.

Fixes some confusion around what the absolute pre-state is for the alphabet VM and when the pre-image should be used vs the hashed claim. As part of this switched to always using a hard coded hex value for the absolute pre-state so it's easy to see what the value should be and compare to what's actually included in the transaction.

The alphabet is defended.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4133/e2e-tests

### DIFF
--- a/op-challenger/fault/alphabet_provider.go
+++ b/op-challenger/fault/alphabet_provider.go
@@ -48,9 +48,7 @@ func (ap *AlphabetProvider) Get(i uint64) (common.Hash, error) {
 }
 
 func (ap *AlphabetProvider) AbsolutePreState() []byte {
-	out := make([]byte, 32)
-	out[31] = 96 // ascii character 96 is "`"
-	return out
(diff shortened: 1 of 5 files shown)


refcell (approved):
LGTM - Only note would be maybe to further abstract resolution logic isn't of only waiting for a claim at the max depth (in case we do more extensive tests) but not necessary right mow.

ajsutton:
> LGTM - Only note would be maybe to further abstract resolution logic isn't of only waiting for a claim at the max depth (in case we do more extensive tests) but not necessary right mow.

Yeah I think we want to keep the e2e test really simple in this regard.  When we add tests that don't have a single direct path down to the max depth then the test can wait for the specific claim it expects last pretty easily rather than trying to automate detection of when the game resolves.

PR #6392, opened 2023-07-21
op-challenger: Provide proof data when calling step

ajsutton:
**Description**

Trace providers now provide proof data along with pre-images that are passed through to the `step` function. For the alphabet trace provider, the proof data is always empty. For cannon the proof data is loaded from cannon's JSON proof output.

**Tests**

Updated unit tests for new functionality.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4244/support-proof-data-in-op-challenger


refcell (approved):
LGTM

PR #6393, opened 2023-07-21
op-challenger: Pass context through correctly

ajsutton:
**Description**

Fix op-challenger to pass through the context to where its needed instead of using `context.Background()` or `context.TODO()`.


refcell (approved):
Very nice, LGTM

PR #6394, opened 2023-07-21
op-challenger: Improve logging

ajsutton:
**Description**

* Fix handling of transaction failures when resolving the game. Previously it inverted the return value and logged only when there was no error.
* Avoid duplicate logging of errors
* Reduce expected situations to debug level
* Log an error if the game is lost
* Consistently start log messages with a capital letter

It does still spam the logs with the game status every 300ms which is a bit excessive but will take more invasive refactoring to solve so leaving for now.

### DIFF
--- a/op-challenger/fault/agent.go
+++ b/op-challenger/fault/agent.go
@@ -35,12 +35,11 @@ func (a *Agent) Act(ctx context.Context) error {
 	}
 	game, err := a.newGameFromContracts(ctx)
 	if err != nil {
-		a.log.Error("Failed to create new game", "err", err)
-		return err
+		return fmt.Errorf("create game from contracts: %w", err)
 	}
 	// Create counter claims
 	for _, claim := range game.Claims() {
-		if err := a.move(ctx, claim, game); err != nil {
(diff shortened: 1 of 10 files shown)


op-challenger/fault/agent.go
```diff
@@ -57,11 +56,13 @@ func (a *Agent) Act(ctx context.Context) error {
...
 		err := a.responder.Resolve(ctx)
 		if err != nil {
-			return true
+			a.log.Error("Failed to resolve the game", "err", err)
+			return false
 		}
-		a.log.Error("failed to resolve the game", "err", err)
+		return true
 	}
 	return false
```
refcell, line 67: I think we can do something like

```suggestion
        if err == nil {
            return true
        }
        a.log.Error("Failed to resolve the game", "err", err)
    }
    return false
```

refcell (approved):
Looks great! Just one small suggestion that can be discarded if I misunderstood :)

op-challenger/fault/agent.go
refcell: Will open a new pr so we can get these merged.

PR #6396, opened 2023-07-21
cannon: Update pre-image input for local data

Inphi:
Use the new PreimageOracle interface to load local pre-image data.

### DIFF
--- a/cannon/mipsevm/witness.go
+++ b/cannon/mipsevm/witness.go
@@ -55,10 +55,13 @@ func (wit *StepWitness) EncodePreimageOracleInput() ([]byte, error) {
 		var input []byte
 		input = append(input, LoadLocalDataBytes4...)
 		input = append(input, wit.PreimageKey[:]...)
+
+		preimagePart := wit.PreimageValue[wit.PreimageOffset:]
 		var tmp [32]byte
-		copy(tmp[:], wit.PreimageValue[8:])
+		copy(tmp[:], preimagePart)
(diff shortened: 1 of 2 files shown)


cannon/mipsevm/witness.go
```diff
@@ -49,16 +49,14 @@ func (wit *StepWitness) EncodePreimageOracleInput() ([]byte, error) {
...
+			return nil, fmt.Errorf("local pre-image exceeds maximum size of 32 bytes with key 0x%x", wit.PreimageKey)
+		}
 		var input []byte
-		input = append(input, CheatBytes4...)
-		input = append(input, uint32ToBytes32(wit.PreimageOffset)...)
+		input = append(input, LoadLocalDataBytes4...)
 		input = append(input, wit.PreimageKey[:]...)
 		var tmp [32]byte
-		copy(tmp[:], wit.PreimageValue[wit.PreimageOffset:])
+		copy(tmp[:], wit.PreimageValue[8:])
```
Inphi: This doesn't work right now because the `PreimageOracle` loads only 1 or 2 parts at offsets 0 and 32 respectively, depending on the pre-image length.
Since the VM can load at most 4 bytes of the pre-image from memory, the op-program may request pre-image parts at any offset greater than zero.

Inphi: Should be working now.

clabby (approved):
ty :)

PR #6399, opened 2023-07-21
vscode: delete newline at end of file

maurelian:
rinse and repeat :)


tynes:
fwiw, its considered standard to end files with newlines
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

tynes (approved):

PR #6406, opened 2023-07-21
feat(op-challenger): Trace Type Flags for Cannon TraceProvider

refcell:
**Description**

This PR adds CLI Flags for the op-challenger TraceProvider type: `--trace-type`.

It then checks that the correct trace source is specified by the respective _optional_ flag.

So,
- For the Alphabet TraceProvider, we require that the `--alphabet` flag is specified when `--trace-type=alphabet` is provided.
- For the Cannon TraceProvider, we require that the `--cannon-datadir` flag is specified when `--trace-type=cannon` is passed into the `op-challenger` binary.

For example:
```bash
$ ./op-challenger --trace-type=cannon --cannon-datadir=./data
$ ./op-challenger --trace-type=alphabet --alphabet=abcdefgh
```

Note (as @ajsutton pointed out), this will change significantly down the road when multi-games are introduced so it's more important that this just works and is syntactically easy to use.

**Metadata**

Fixes CLI-4225


refcell:
* develop
  * **PR #6406** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6406" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6406?utm_source=stack-comment).

op-challenger/challenger.go
```diff
@@ -36,7 +38,13 @@ func Main(ctx context.Context, logger log.Logger, cfg *config.Config) error {
...
 		return fmt.Errorf("failed to create the responder: %w", err)
 	}
-	trace := fault.NewAlphabetProvider(cfg.AlphabetTrace, uint64(cfg.GameDepth))
+
+	var trace fault.TraceProvider
+	if cfg.TraceType == flags.CannonTraceType {
+		trace = cannon.NewCannonTraceProvider(cfg.CannonDatadir)
+	} else if cfg.TraceType == flags.AlphabetTraceType {
+		trace = fault.NewAlphabetProvider(cfg.AlphabetTrace, uint64(cfg.GameDepth))
+	}
```
ajsutton: We probably should use a switch here with a default that returns an error. It shouldn't happen but worth being defensive to ensure we get a good error message if we add a new type in the future and forget to add it here.

op-challenger/config/config.go
```diff
@@ -75,11 +94,16 @@ func NewConfigFromCLI(ctx *cli.Context) (*Config, error) {
 
 	txMgrConfig := txmgr.ReadCLIConfig(ctx)
 
+	traceTypeFlag := strings.TrimLeft(ctx.String(flags.TraceTypeFlag.Name), "[")
+	traceTypeFlag = strings.TrimRight(traceTypeFlag, "]")
```
ajsutton: This feels like we're fighting the CLI framework and losing.  I don't think we want a `StringSliceFlag` type since that allows multiple values but rather just a `GenericFlag` with a custom type like is used for rpckind in op-node.

op-challenger/cmd/main_test.go
```diff
@@ -59,15 +61,14 @@ func TestL1ETHRPCAddress(t *testing.T) {
 	})
 }
 
-func TestAlphabetTrace(t *testing.T) {
+func TestTraceType(t *testing.T) {
```
ajsutton, line 64: We should have a test for invalid trace types too, and check that all valid options are accepted.

ajsutton (commented):
I like the approach generally but `StringSliceFlag` is the wrong type.  I'll push up an extra commit that changes it over to `GenericFlag` which tidies things up more. Happy to revisit if you disagree but want to keep things flowing. :)

refcell:
> I like the approach generally but `StringSliceFlag` is the wrong type.  I'll push up an extra commit that changes it over to `GenericFlag` which tidies things up more. Happy to revisit if you disagree but want to keep things flowing. :)

Makes sense to me!

ajsutton (approved):
That was messier than expected because of #6408 but got something work. Will approve this, then have a play with our module layout to see if we can do something that is less cycle-prone. I'm not really a fan of `TraceType` having to be in `flags` module instead of `fault/types.go` for example...

PR #6411, opened 2023-07-23
contracts-bedrock: deploy eas implementations

tynes:
**Description**

Deploy the Ethereum Attestation Service implementations to mainnet. These are meant to be proxied behind predeploy addresses in the future.

The diff in the `.chainId` file is because the forge script seemingly chokes on trailing newline characters when reading that file.

The deployed contracts are verified
- https://optimistic.etherscan.io/address/0x4E0275Ea5a89e7a3c1B58411379D1a0eDdc5b088
- https://optimistic.etherscan.io/address/0x6232208d66bAc2305b46b4Cb6BCB3857B298DF13


refcell (approved):
Reviewed both deployments.
- EAS address, contract code, and bytecode match.
- Schema Registry address, contract code, and bytecode match.

PR #6412, opened 2023-07-23
feat: forge deploy allocs

tynes:
**Description**
Refactor op-e2e to use the same deploy script that is used in
production. There is only 1 way to deploy contracts now and
it goes through foundry. This removes the need to maintain
multiple sets of contracts deployments/devnets. It does
add a new tooling dep to the monorepo which is `geth`.
A top level `make` command is provided as well as some
docs in the `op-e2e` package on its usage.

The `devnetL1.json` deployconfig is the source of truth
for deploy configs in `op-e2e` now. In the long term,
we should delineate which config values are safe to alter
in memory and which are not. Some things are still coupled
together in weird ways, but I tried to add assertions where
I could to ensure that future people will get better error
messages.

To install geth, run the command:

```
make install-geth
```

To create the genesis state, run the command:

```
make devnet-allocs
```

This will run the forge deploy script against an instance of
`geth --dev` and then call `debug_dumpBlock` which will dump
the state into a format that we can inject at runtime into
the genesis of both `op-e2e` as well as the L1 of the docker
compose setup.

I tried to use `anvil` instead of `geth` but it does not
support this endpoint. It supports a different endpoint
that uses a compressed binary serialization of state,
it is easier to work with the JSON.

There are no more deterministic addresses on L1 because it is too much maintenance to maintain 2 deployment systems. Now the production deployment system is used in `op-e2e`

The addresses are stored in `.devnet/addresses.json` and I added an nginx server to serve them at `localhost:8080/addresses.json`. This is because they will be different every time and gives an easy way for the front end to access them.

Can confirm that the devnet blocktime + proposal time results in a usable local devnet, the L2 doesn't run away from the proposals

Alternative to https://github.com/ethereum-optimism/optimism/pull/6123

Closes CLI-4141


op-e2e/actions/l2_proposer_test.go
```diff
@@ -67,13 +78,32 @@ func TestProposer(gt *testing.T) {
...
 	require.NoError(t, err)
+
+	ival, err := outputOracleContract.SUBMISSIONINTERVAL(&bind.CallOpts{})
+	require.NoError(t, err)
+	require.Equal(t, ival.Uint64(), dp.DeployConfig.L2OutputOracleSubmissionInterval, "submission interval must match")
+
+	idx, err := outputOracleContract.LatestBlockNumber(&bind.CallOpts{})
+	require.NoError(t, err)
+	require.Greater(t, int64(idx.Uint64()), int64(0), "output index must be greater than 0")
+	fmt.Println("==============", idx)
```
tynes: wip stuff here

op-e2e/e2eutils/secrets.go
```diff
@@ -157,11 +148,40 @@ func (s *Secrets) Addresses() *Addresses {
 	}
 }
 
+func (s *Secrets) String() string {
```
tynes: This could prob be deleted

op-e2e/geth.go
```diff
@@ -124,7 +124,7 @@ func initL1Geth(cfg *SystemConfig, genesis *core.Genesis, c clock.Clock, opts ..
 		HTTPModules: []string{"debug", "admin", "eth", "txpool", "net", "rpc", "web3", "personal", "engine"},
 	}
 
-	l1Node, l1Eth, err := createGethNode(false, nodeConfig, ethConfig, []*ecdsa.PrivateKey{cfg.Secrets.CliqueSigner}, opts...)
```
tynes: Was somehow L1 never mining blocks because this was not passed in?

tynes:
L1 genesis has pre merge config:

```
WARN [07-26|19:49:38.636] Served eth_getBlockByNumber              reqid=9 duration="161.419µs" err="'finalized' tag not supported on pre-merge network"
WARN [07-26|19:49:38.636] Served eth_getBlockByNumber              reqid=10 duration="166.15µs"  err="'safe' tag not supported on pre-merge network"
```

op-e2e/bridge_test.go
```diff
@@ -8,20 +8,24 @@ import (
...
 	"github.com/stretchr/testify/require"
+	//"runtime"
 )
 
 // TestERC20BridgeDeposits tests the the L1StandardBridge bridge ERC20
 // functionality.
 func TestERC20BridgeDeposits(t *testing.T) {
 	InitParallel(t)
 
+	//runtime.Breakpoint()
```
tynes: dead code

op-e2e/setup.go
```diff
@@ -454,6 +388,7 @@ func (cfg SystemConfig) Start(_opts ...SystemConfigOption) (*System, error) {
 		if name == "l1" {
 			continue
 		}
+		// In here someplace?
```
tynes: dead code

op-node/cmd/genesis/cmd.go
```diff
@@ -55,7 +55,8 @@ var Subcommands = cli.Commands{
 				return err
 			}
 
-			l1Genesis, err := genesis.BuildL1DeveloperGenesis(config)
+			// TODO: this needs new inputs now to be able to create an L1 genesis
```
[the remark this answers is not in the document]
tynes: Add an extra CLI arg for the path to the dump file

op-e2e/config/init.go
```diff
@@ -0,0 +1,164 @@
...
+	if !ok {
+		return common.Address{}, fmt.Errorf("%s not found", name)
+	}
+	return addr, nil
+}
+
+// ReadAllocs
+func ReadAllocs(filename string) (*state.Dump, error) {
+	_, err := os.Stat(filename)
+	if os.IsNotExist(err) {
```
tynes: `errors.Is(err, fs.ErrNotExist)`

tynes:
In the devnet with genesis contracts:

```
timeout 10m npx hardhat deposit-erc20 --network devnetL1
Using signer 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
An unexpected error occurred:

TypeError: Cannot read properties of undefined (reading 'l1')
    at SimpleTaskDefinition.action (/home/circleci/project/packages/sdk/tasks/deposit-erc20.ts:154:21)
```

tynes:
```
       	Error Trace:	/root/project/op-e2e/system_fpp_test.go:108
        	            				/root/project/op-e2e/system_fpp_test.go:31
        	Error:      	Received unexpected error:
        	            	Post "http://127.0.0.1:37781": context deadline exceeded
        	Test:       	TestVerifyL2OutputRootEmptyBlock
```

packages/sdk/src/interfaces/types.ts
```diff
@@ -27,7 +27,7 @@ export enum L2ChainID {
   OPTIMISM_GOERLI = 420,
   OPTIMISM_HARDHAT_LOCAL = 31337,
   OPTIMISM_HARDHAT_DEVNET = 17,
-  OPTIMISM_BEDROCK_LOCAL_DEVNET = 901,
+  OPTIMISM_BEDROCK_LOCAL_DEVNET = 902,
```
tynes: see if we can change these back without any issues

packages/contracts-bedrock/scripts/Deployer.sol
```diff
@@ -441,7 +441,7 @@ abstract contract Deployer is Script {
             return "optimism-goerli";
         } else if (chainid == 10) {
             return "optimism-mainnet";
-        } else if (chainid == 900) {
+        } else if (chainid == 900 || chainid == 1337) {
```
tynes: don't let this get out of sync with the values in the config

packages/contracts-bedrock/deploy-config/devnetL1.json
```diff
@@ -1,47 +1,47 @@
 {
-  "l1ChainID": 900,
-  "l2ChainID": 901,
+  "l1ChainID": 901,
```
tynes: todo: revert this diff of the chain id changes

tynes:
```
    system_config_test.go:103: 
        	Error Trace:	/root/project/op-e2e/actions/system_config_test.go:103
        	Error:      	Not equal: 
        	            	expected: 0x3
        	            	actual  : 0x4
        	Test:       	TestBatcherKeyRotation
        	Messages:   	L2 block with L1 origin that included config change
```

op-e2e/e2eutils/secrets.go
```diff
@@ -97,8 +97,8 @@ func (m *MnemonicConfig) Secrets() (*Secrets, error) {
 
 	return &Secrets{
 		Deployer:     deployer,
-		CliqueSigner: cliqueSigner,
 		SysCfgOwner:  sysCfgOwner,
+		CliqueSigner: cliqueSigner,
```
tynes, line 103: reduce unnecessary diff in this file

op-e2e/setup.go
```diff
@@ -78,74 +79,8 @@ func newTxMgrConfig(l1Addr string, privKey *ecdsa.PrivateKey) txmgr.CLIConfig {
...
-	addresses := secrets.Addresses()
-
-	deployConfig := &genesis.DeployConfig{
-		L1ChainID:   900,
-		L2ChainID:   901,
-		L2BlockTime: 1,
-
-		FinalizationPeriodSeconds: 60 * 60 * 24,
-		MaxSequencerDrift:         10,
-		SequencerWindowSize:       30,
```
tynes, line 90: double check sequencer window size

tynes:
```
        	Error Trace:	/root/project/op-e2e/actions/blocktime_test.go:53
        	            				/root/project/op-e2e/actions/blocktime_test.go:81
        	Error:      	Not equal: 
        	            	expected: 0x11
        	            	actual  : 0x15
        	Test:       	TestBatchInLastPossibleBlocks
```

tynes:
For some reason it seems like there are p2p issues between a verifier and sequencer in the tests? Not sure why, still debugging. Asserting that the unsafe block signer is set correctly, seems like the only thing that would cause this

op-e2e/actions/system_config_test.go
```diff
@@ -92,15 +101,25 @@ func TestBatcherKeyRotation(gt *testing.T) {
...
-	payload, err := engCl.PayloadByNumber(t.Ctx(), sequencer.L2Safe().Number+12) // 12 new L2 blocks: 5 with origin before L1 block with batch, 6 with origin of L1 block with batch, 1 with new origin that changed the batcher
-	require.NoError(t, err)
-	ref, err := derive.PayloadToBlockRef(payload, &sd.RollupCfg.Genesis)
-	require.NoError(t, err)
-	require.Equal(t, ref.L1Origin.Number, cfgChangeL1BlockNum, "L2 block with L1 origin that included config change")
-	require.Equal(t, ref.SequenceNumber, uint64(0), "first L2 block with this origin")
-	sysCfg, err := derive.PayloadToSystemConfig(payload, sd.RollupCfg)
-	require.NoError(t, err)
-	require.Equal(t, dp.Addresses.Bob, sysCfg.BatcherAddr, "bob should be batcher now")
+	for i := 0; i <= 24; i++ {
```
tynes: Does setting a L2 blocktime of 2 remove the need for this diff?

bedrock-devnet/devnet/genesis.py
```diff
@@ -8,7 +8,7 @@
 
 GENESIS_TMPL = {
     'config': {
```
tynes: Ideally we can remove this template in favor of a command in OP Node that builds the L1 genesis file, then we have 1 source of truth for genesis files which is the op node `genesis` command

packages/contracts-bedrock/deploy-config/devnetL1.json
```diff
@@ -1,47 +1,47 @@
...
   "p2pSequencerAddress": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
-  "batchInboxAddress": "0xff00000000000000000000000000000000000000",
+  "batchInboxAddress": "0xff00000000000000000000000000000000000901",
   "batchSenderAddress": "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC",
-  "l2OutputOracleSubmissionInterval": 20,
-  "l2OutputOracleStartingTimestamp": -1,
+  "cliqueSignerAddress": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
+  "l1UseClique": true,
+  "l1StartingBlockTag": "earliest",
+  "l2OutputOracleSubmissionInterval": 6,
```
tynes, line 14: see if we can increase this safely

packages/contracts-bedrock/deploy-config/devnetL1.json
```diff
@@ -1,47 +1,47 @@
 {
-  "l1ChainID": 900,
-  "l2ChainID": 901,
-  "l2BlockTime": 2,
+  "l1ChainID": 901,
+  "l2ChainID": 902,
+  "l2BlockTime": 1,
```
tynes: having this as 1 speeds up ci

op-e2e/config/init.go
```diff
@@ -0,0 +1,135 @@
...
+
+	layout, err := bindings.GetStorageLayout("OptimismPortal")
+	if err != nil {
+		panic(err)
+	}
+	entry, err := layout.GetStorageLayoutEntry("params")
+	if err != nil {
+		panic(err)
+	}
+	slot := common.BigToHash(big.NewInt(int64(entry.Slot)))
```
tynes: not ideal that we have to do this, it also means we will have to do it in the python code. ideally we could do it if we were using anvil but it doesn't support the rpc we need. it has an alt state dumping rpc that i'm not exactly sure how to work with to be honest, its binary + compressed, would need to look into it more

op-e2e/setup.go
```diff
@@ -78,81 +80,27 @@ func newTxMgrConfig(l1Addr string, privKey *ecdsa.PrivateKey) txmgr.CLIConfig {
...
 
-	return SystemConfig{
-		Secrets: secrets,
+	require.Equal(t, secrets.Addresses().Batcher, deployConfig.BatchSenderAddress)
+	require.Equal(t, secrets.Addresses().SequencerP2P, deployConfig.P2PSequencerAddress)
+	require.Equal(t, secrets.Addresses().Proposer, deployConfig.L2OutputOracleProposer)
 
-		Premine: make(map[common.Address]*big.Int),
+	// TODO: maybe don't need this?
+	premine := make(map[common.Address]*big.Int)
```
tynes, line 94: i don't believe we need this anymore

op-node/sources/eth_client.go
```diff
@@ -403,6 +403,11 @@ func (s *EthClient) GetStorageAt(ctx context.Context, address common.Address, st
 // ReadStorageAt is a convenience method to read a single storage value at the given slot in the given account.
 // The storage slot value is verified against the state-root of the given block if we do not trust the RPC provider, or directly retrieved without proof if we do trust the RPC.
 func (s *EthClient) ReadStorageAt(ctx context.Context, address common.Address, storageSlot common.Hash, blockHash common.Hash) (common.Hash, error) {
+	// TODO: temp
```
tynes: not sure why this was an issue, archival mode must be off?

op-e2e/config/init.go
tynes: fixed this issue

op-e2e/config/init.go
```diff
@@ -0,0 +1,124 @@
...
+	// Set the L1 genesis block timestamp to now
+	DeployConfig.L1GenesisBlockTimestamp = hexutil.Uint64(time.Now().Unix())
+	DeployConfig.FundDevAccounts = true
+
+	if L1Deployments != nil {
+		DeployConfig.SetDeployments(L1Deployments)
+	}
+
+	/*
+		L1Allocs, err = genesis.PostProcessL1DeveloperGenesis(L1Allocs, L1Deployments)
```
tynes: todo: dead code

Makefile
```diff
@@ -77,8 +77,9 @@ devnet-up:
 	PYTHONPATH=./bedrock-devnet python3 ./bedrock-devnet/main.py --monorepo-dir=.
 .PHONY: devnet-up
 
+# alias for devnet-up
 devnet-up-deploy:
-	PYTHONPATH=./bedrock-devnet python3 ./bedrock-devnet/main.py --monorepo-dir=. --deploy
+	PYTHONPATH=./bedrock-devnet python3 ./bedrock-devnet/main.py --monorepo-dir=.
 .PHONY: devnet-up-deploy
```
refcell: If `devnet-up-deploy` is now the same as `devnet-up`, can we just remove it or if not point `devnet-up-deploy` to `devnet-up`.

ie.
```suggestion
devnet-up-deploy: devnet-up
```

.circleci/config.yml
```diff
@@ -886,9 +923,9 @@ jobs:
       - run:
           name: Install latest golang
           command: |
-            wget https://go.dev/dl/go1.19.linux-amd64.tar.gz
+            wget https://go.dev/dl/go1.20.linux-amd64.tar.gz
             sudo rm -rf /usr/local/go
-            sudo tar -C /usr/local -xzf go1.19.linux-amd64.tar.gz
+            sudo tar -C /usr/local -xzf go1.20.linux-amd64.tar.gz
             export PATH=$PATH:/usr/local/go/bin
             go version
```
refcell: go versioning should also be more tightly controlled.
this is error-prone and at times not an obvious toolchain fix.

note, this should just be ticketed and fixed in a separate pr.

packages/contracts-bedrock/deploy-config/devnetL1.json
tynes: added adee57d17a7708a566108cc0d17ae7dd433f5ea2 to get logging on the rate of output proposals vs L2 blocks. If L2 is running away from L1 it makes the docker devnet setup unusable for dev. Will fix based on the results of the logging

tynes: Can confirm that the devnet will not have issues with an interval of 6 with the current L1 and L2 blocktimes

bedrock-devnet/devnet/__init__.py
```diff
@@ -9,22 +9,23 @@
...
-parser.add_argument('--deploy', help='Whether the contracts should be predeployed or deployed', type=bool, action=argparse.BooleanOptionalAction)
+parser.add_argument('--allocs', help='Path to the allocs file', type=bool, action=argparse.BooleanOptionalAction)
 
 log = logging.getLogger()
 
 class Bunch:
     def __init__(self, **kwds):
         self.__dict__.update(kwds)
 
+
```
refcell, line 28: nit
```suggestion
```

bedrock-devnet/devnet/__init__.py
```diff
@@ -9,22 +9,23 @@
...
 
 import devnet.log_setup
-from devnet.genesis import GENESIS_TMPL
 
 pjoin = os.path.join
 
 parser = argparse.ArgumentParser(description='Bedrock devnet launcher')
 parser.add_argument('--monorepo-dir', help='Directory of the monorepo', default=os.getcwd())
-parser.add_argument('--deploy', help='Whether the contracts should be predeployed or deployed', type=bool, action=argparse.BooleanOptionalAction)
+parser.add_argument('--allocs', help='Path to the allocs file', type=bool, action=argparse.BooleanOptionalAction)
```
tynes: This comment is wrong

packages/sdk/tasks/deposit-erc20.ts
```diff
@@ -144,30 +145,51 @@ task('deposit-erc20', 'Deposits WETH9 onto L2.')
...
+        l1: {
+          AddressManager: json.AddressManager,
+          L1CrossDomainMessenger: json.L1CrossDomainMessengerProxy,
+          L1StandardBridge: json.L1StandardBridgeProxy,
+          StateCommitmentChain: ethers.constants.AddressZero,
+          CanonicalTransactionChain: ethers.constants.AddressZero,
+          BondManager: ethers.constants.AddressZero,
+          OptimismPortal: json.OptimismPortalProxy,
+          L2OutputOracle: json.L2OutputOracleProxy,
+        },
```
refcell, line 159: This adds another point of maintenance if we add contracts. Unless there's a good reason, can we revert to passing through the contracts json?

packages/sdk/tasks/deposit-eth.ts
```diff
@@ -88,8 +89,18 @@ task('deposit-eth', 'Deposits ether to L2.')
...
+        l1: {
+          AddressManager: json.AddressManager,
+          L1CrossDomainMessenger: json.L1CrossDomainMessengerProxy,
+          L1StandardBridge: json.L1StandardBridgeProxy,
+          StateCommitmentChain: ethers.constants.AddressZero,
+          CanonicalTransactionChain: ethers.constants.AddressZero,
+          BondManager: ethers.constants.AddressZero,
+          OptimismPortal: json.OptimismPortalProxy,
+          L2OutputOracle: json.L2OutputOracleProxy,
+        },
```
refcell, line 103: ibid

packages/sdk/tasks/deposit-eth.ts
```diff
@@ -187,11 +210,11 @@ task('deposit-eth', 'Deposits ether to L2.')
       contracts: contractAddrs,
     })
 
-    const opBalanceBefore = await signer.provider.getBalance(
+    const opBalanceBefore = await signer!.provider!.getBalance(
```
refcell, line 213: I don't think we need the non-null operator here and below since its address is immediately fetched after instantiation. ie, the script will error when calling `signer.getAddress()` on the null `signer = signers[0]`

refcell:
* develop
  * **PR #6412** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6412" target="_blank">Graphite</a>  👈
    * **PR #6552** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6552" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6412?utm_source=stack-comment).

op-node/cmd/genesis/cmd.go
```diff
@@ -49,39 +51,36 @@ var Subcommands = cli.Commands{
...
+			if deployments != nil {
+				config.SetDeployments(deployments)
 			}
 
-			l1Genesis, err := genesis.BuildL1DeveloperGenesis(config)
-			if err != nil {
-				return err
+			if err := config.Check(); err != nil {
+				return fmt.Errorf("deploy config at %s invalid: %w", deployConfig, err)
 			}
```
refcell, line 69: This can be moved up to immediately after the config is created right?

refcell (changes requested):
Small comments and nits mostly. Amazing PR

Makefile
trianglesphere: +1 to removing `devnet-up-deploy`

bedrock-devnet/devnet/__init__.py
trianglesphere: Are you planning on fixing it? What should it be instead?

bedrock-devnet/devnet/__init__.py
```diff
@@ -33,88 +34,112 @@ def main():
...
       genesis_l2_path=pjoin(devnet_dir, 'genesis-l2.json'),
+      allocs_path=pjoin(devnet_dir, 'allocs-l1.json'),
       addresses_json_path=pjoin(devnet_dir, 'addresses.json'),
       sdk_addresses_json_path=pjoin(devnet_dir, 'sdk-addresses.json'),
       rollup_config_path=pjoin(devnet_dir, 'rollup.json')
     )
 
     os.makedirs(devnet_dir, exist_ok=True)
 
+    if args.allocs:
```
trianglesphere, line 61: So `--allocs` means building the genesis?

bedrock-devnet/devnet/__init__.py
```diff
@@ -33,88 +34,112 @@ def main():
...
 
 # Bring up the devnet where the contracts are deployed to L1
 def devnet_deploy(paths):
     if os.path.exists(paths.genesis_l1_path):
         log.info('L1 genesis already generated.')
     else:
         log.info('Generating L1 genesis.')
-        write_json(paths.genesis_l1_path, GENESIS_TMPL)
+        if os.path.exists(paths.allocs_path) == False:
+            devnet_l1_genesis(paths)
```
trianglesphere, line 127: what is the difference between `allocs_path` and `genesis_l1_path`? Why do we appear to be generating `allocs_path` with the function `devnet_l1_genesis`?

op-e2e/README.md
```diff
@@ -0,0 +1,20 @@
+# op-e2e
+
+The end to end tests in this repo depend on genesis state that is
+created with the `bedrock-devnet` package. To create this state,
+run the following commands from the root of the repository:
+
+```bash
+make install-geth
+make devnet-allocs
```
trianglesphere, line 9: Maybe we don't always run `make install-geth`, but we should set the makefile rules up such that `make devnet-allocs` is ran when `devnetL1.json` is newer than anything in `.devnet` & should be ran when running `make test` inside op-e2e. I don't like that we manually have to run several steps to run tests & I think it's likely that we will run into issues where `devnet-allocs` gets out of date.

trianglesphere (commented):
First pass seems solid, but there's a lot to review.

packages/sdk/tasks/deposit-erc20.ts
tynes: We cannot pass thru the json directly unfortunately, these tasks will be deprecated along with the sdk soon enough

tynes: The keys changed

packages/sdk/tasks/deposit-eth.ts
tynes: My LSP was yelling at me about this

bedrock-devnet/devnet/__init__.py
tynes: It technically isn't building the genesis file, its building the allocs field in the genesis file

tynes:
A point of confusion is the difference between allocs and L1 genesis. Right now the implementation separates the creation of generating the L1 state and the L1 genesis block itself. It was slightly easier to implement it this way but I do believe that I could refactor things so that the L1 genesis is created and there isn't a concept of allocs if that is preferred. The L1 genesis has a timetamp on it which could go stale if its not created dynamically, but that is something that we could always modify at time of loading the file

.circleci/config.yml
tynes: Ok yes I agree

tynes: Reverted this, broke out into its own PR

op-node/cmd/genesis/cmd.go
tynes: It needs to happen after the deployments are set, otherwise it will fail

op-e2e/README.md
tynes: I def see what you mean. I should clarify in the readme that you only need to run `make install-geth` once and then again if we update the version

We need to run `make devnet-allocs` if anything in `contracts-bedrock/{src,deploy-config,scripts}` changes

bedrock-devnet/devnet/__init__.py
tynes: The allocs are the result of `debug_dumpBlock` so its the state dump, its called allocs in the genesis struct

Makefile
tynes: this is now fixed, good call

bedrock-devnet/devnet/__init__.py
tynes: the rest of the file uses 2 newlines between things

bedrock-devnet/devnet/__init__.py
tynes: this has been fixed

op-e2e/README.md
tynes: This doesn't work very well with make. 9ed901f58f15e82480c372b2ca8b726de0fefa37 was an attempt at solving this but it wasn't portable to `sh` and making bash the shell broke ci

tynes:
Can we follow up in a future PR for the auto detect/rebuild functionality? cc @trianglesphere

op-chain-ops/genesis/config.go
```diff
@@ -74,10 +77,15 @@ type DeployConfig struct {
 	// L2OutputOracleChallenger is the address of the account that challenges L2 outputs.
 	L2OutputOracleChallenger common.Address `json:"l2OutputOracleChallenger"`
 
+	// CliqueSignerAddress represents the signer address for the clique consensus engine.
+	// It is used in the multi-process devnet to sign blocks.
+	CliqueSignerAddress common.Address `json:"cliqueSignerAddress"`
+	// L1UseClique represents whether or not to use the clique consensus engine.
+	L1UseClique bool `json:"l1UseClique"`
+
```
protolambda, line 85: Fine to make it optional in this PR, but would like to get rid of clique soon, since it's not going to be around for very long upstream.

op-e2e/e2eutils/setup_test.go
```diff
@@ -22,6 +23,7 @@ func TestSetup(t *testing.T) {
 		MaxSequencerDrift:   40,
 		SequencerWindowSize: 120,
 		ChannelTimeout:      120,
+		L1BlockTime:         15,
```
protolambda, line 26: Why 15 and not 12? I think it previously defaulted to 12.

op-e2e/actions/l2_verifier_test.go
```diff
@@ -34,6 +34,7 @@ func TestL2Verifier_SequenceWindow(gt *testing.T) {
 		MaxSequencerDrift:   10,
 		SequencerWindowSize: 24,
 		ChannelTimeout:      10,
+		L1BlockTime:         15,
```
protolambda, line 37: Same here, why 15 and not 12?

protolambda (approved):
generally looks good

op-node/cmd/genesis/cmd.go
refcell: Oops missed that, thanks!

refcell (approved):
Looks good to me!

op-chain-ops/genesis/config.go
tynes: I definitely would like to get rid of clique too, the problem is that the devnet still relies on it, I made a linear issue on this. https://linear.app/optimism/issue/CLI-4309/update-l1-node-in-docker-compose-devnet

op-e2e/e2eutils/setup_test.go
tynes: I thought that I made this exactly what the value was previously on purpose, maybe I made a mistake. The test works either way

op-e2e/actions/l2_verifier_test.go
tynes: The default is 15 so I am going to leave as is

https://github.com/ethereum-optimism/optimism/blob/00e7095ce39a60f74de27189c031c384405a4f95/op-e2e/actions/l1_replica_test.go#L21-L26

PR #6413, opened 2023-07-24
op-challenger: Tidy up modules

ajsutton:
**Description**

The module structure in `op-challenger` wound up with key shared types and the wiring code that creates everything in the same `fault` module. That made it very difficult to split code out into other modules without creating cyclic dependencies as they either needed to avoid using the core types or avoid being created.

To resolve this, there's two key changes in this PR:
1. The `config` module no longer depends on `flags` - the ability to create a config from the CLI context is now in the `flags` module. This makes the `flags` module entirely responsible for defining flags and mapping them to the programmatic config. While the `config` module is now focussed on just defining the programmatic interface for configuring the challenger.
2. The core types of `Game`, `Position`, `TraceProvider` and things like `GameState` are moved to a `fault/types` module. The wiring code is left in `fault` which makes it possible to split code out of the main `fault` module without introducing cycles.

Next steps will be to move more of the code in `fault` to more focussed submodules. In particular the alphabet trace provider should be its own module so we have separate modules for each trace provider implementation, and the game solver should be its own module with a clear separation between it and the logic for tracking game state and sending transactions etc.  The solver tests are currently tightly coupled to internals of the alphabet provider though so they need to be improved first.

### DIFF
--- a/op-challenger/cmd/main.go
+++ b/op-challenger/cmd/main.go
@@ -60,7 +60,7 @@ func run(args []string, action ConfigAction) error {
 		}
 		logger.Info("Starting op-challenger", "version", VersionWithMeta)
 
-		cfg, err := config.NewConfigFromCLI(ctx)
+		cfg, err := flags.NewConfigFromCLI(ctx)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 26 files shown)


refcell (approved):
Generally looks good, will follow up with a minimal touchup pr to prevent blocking as discussed :)

PR #6414, opened 2023-07-24
op-challenger: Modify solver tests to not be dependent on alphabet provider details

ajsutton:
**Description**

Modify solver tests to not be dependent on the implementation details of alphabet provider. Also makes the tests more comprehensive.

Introduces the ClaimBuilder test utility to make creating claims in various situations easier and reduce duplication. 

Tweaked solver implementation so that it uses a `TraceProvider` rather than actually _being` a `TraceProvider` and exposing it as part of its API.

**Additional context**

Builds on #6413  

**Metadata**

Fixes CLI-4274

### DIFF
--- a/op-challenger/cmd/main.go
+++ b/op-challenger/cmd/main.go
@@ -60,7 +60,7 @@ func run(args []string, action ConfigAction) error {
 		}
 		logger.Info("Starting op-challenger", "version", VersionWithMeta)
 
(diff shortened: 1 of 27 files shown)


refcell (approved):
Looks good as well! I love the `ClaimBuilder` testing utility for building table tests, very clean abstraction.

PR #6424, opened 2023-07-24
fix(op-challenger): Refactor Solver Component into its own Module

refcell:
**Description**

Following #6413, this PR implements the suggested change to refactor
the `Solver` component into its own "submodule" within the `fault` module
of the `op-challenger`.

The core `Solver` logic is placed inside `fault/solver/solver.go`, with
alphabet tests inside `solver_test.go`.

In a pr stacked ontop of this, and still as part of CLI-4275, the solver
will be further refactored to separate out tightly coupled internal logic.

Once this logic is decoupled, the solver tests can be focussed onto unit
functions and a generic solver testing utility can be built that can be
typed with various `TraceProvider` instances (eg `AlphabetTraceProvider`
and `CannonTraceProvider`) and run in those fault module testbeds.

**Metadata**

Fixes CLI-4275.

### DIFF
--- a/op-challenger/fault/agent.go
+++ b/op-challenger/fault/agent.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/ethereum-optimism/optimism/op-challenger/fault/solver"
 	"github.com/ethereum-optimism/optimism/op-challenger/fault/types"
 	"github.com/ethereum/go-ethereum/log"
 )
@@ -19,7 +20,7 @@ type Responder interface {
 }
 
 type Agent struct {
-	solver                  *Solver
+	solver                  *solver.Solver
 	loader                  Loader
 	responder               Responder
 	maxDepth                int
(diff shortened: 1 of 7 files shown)


refcell:
* develop
  * **PR #6424** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6424" target="_blank">Graphite</a>  👈
    * **PR #6426** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6426" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6424?utm_source=stack-comment).

ajsutton (approved):
LGTM. I don't think we need to test the solver with different trace providers though - it doesn't actually care what the claims are as long as it can get the data from the trace provider.  We just don't want it to be tied to the internal implementations of the trace provider it uses (prior to some recent clean up it was manually calculating hashes that had to match what the alphabet provider would generate).

PR #6425, opened 2023-07-24
op-program: Boot program using output root

Inphi:
**Description**

The output root replaces the L2 head hash at boot as the agreed upon point to run the fault proof program. This makes it easy to step the fault proof program on-chain as the output root is readily available in the output oracle contract.
The L2 block hash at the output root is still necessary for derivation but is now redundant as it's already contained in the output.

However, the host program still needs the L2 head to drive the client off-chain. While it's possible to retrieve the l2 head associated with the output root from the output oracle, we do not do this in the host to keep the implementation minimal.

**Tests**

The existing `make verify-goerli` test in op-program.


Inphi:
cc: @clabby

ajsutton (approved):
LGTM.

op-node/eth/output.go
```diff
@@ -12,3 +15,70 @@ type OutputResponse struct {
...
+	switch ver {
+	case OutputVersionV0:
+		return unmarshalOutputV0(data)
+	default:
+		return nil, ErrInvalidOutputVersion
+	}
+}
+
+func unmarshalOutputV0(data []byte) (*OutputV0, error) {
+	if len(data) < 128 {
```
protolambda:
```suggestion
	if len(data) != 128 {
```
Shouldn't we be more strict about the data size?

protolambda (approved):

op-node/eth/output.go
Inphi: I wanted to accommodate usecases where the caller has a buffer containing both OutputRootV0 and other data. But this is premature. Will fix.

PR #6427, opened 2023-07-24
feat(chain-mon): Default wd-mon scan from start of FPW

maurelian:
**Description**

Simply modifies the start block that the withdrawal sync from, to take the one from the start of the current fault proof window on startup.

**Additional context**

I considered taking a similar approach to how the Fault Detector identifies [the start batch](https://github.com/ethereum-optimism/optimism/blob/36ae8c60b5ca3ed279f920acad1ac418e931c04d/packages/chain-mon/src/fault-mon/service.ts#L205) (or rather an L2Output since Bedrock) using a binary search. that turned out to be overkill in this situation. However, because the `wd-mon` looks for an L1 block, 

 would mean doing the following: 
1. identifying the last unfinalized output based on the fault proof window. 
2. then we'd use that output to get the L1 timestamp at which it was posted
3. then we'd need to do a binary search over L1 blocks to find the one closest to that timestamp, which is at most going to be off by one from just going back by `faultProofWindow / 12` blocks.


packages/chain-mon/src/wd-mon/service.ts
```diff
@@ -109,10 +109,18 @@ export class WithdrawalMonitor extends BaseServiceV2<Options, Metrics, State> {
...
+      `fault proof window is ${this.state.faultProofWindow} seconds`
+    )
+
+    // Set the start block number.
     if (this.options.startBlockNumber === -1) {
+      // We default to starting from the earliest block still in the fault proof window.
+      const l1BlockNumber = await this.options.l1RpcProvider.getBlockNumber()
       this.state.highestUncheckedBlockNumber =
-        await this.options.l1RpcProvider.getBlockNumber()
+        l1BlockNumber - this.state.faultProofWindow / 12
```
hamdiallam: Just for sanity purposes, can we walk backwards from this estimated block height and terminate when reaching a block outside of the fpw? Would be great to ensure the right height

hamdiallam:
> then we'd use that output to get the L1 timestamp at which it was posted

Unfortunate that the L1 block number isn't included onchain

packages/chain-mon/src/wd-mon/service.ts
maurelian: OK, just to make sure I'm following that'd mean:
1. get the timestamp from the latest block. 
2. query for a block number based on the math here
3. check if that block has a timestamp from before the start of the FPW
4. If not, try the previous block, etc.

hamdiallam: Yep exactly, the estimate should be pretty good so it should be a very short walk. Maybe add a small comment about the estimate given the 12s slot time

hamdiallam (approved):

PR #6428, opened 2023-07-24
contracts-bedrock: deploy config for sepolia

tynes:
**Description**

Adds the deploy config for sepolia. Also clean up a little bit of dead code and parse sepolia L1 and L2 in the deployer script. Optimistic Kovan had a chain id of 69 and is now deprecated, so re-using 69 for Optimistic Sepolia so that 420 can be used by goerli still while both goerli and sepolia run in parallel.

To review the deploy config, the values should almost all match goerli except for 
- the eip 1559 params which should now match mainnet
- The chainids should also be different
- The gas oracle params should match mainnet


packages/contracts-bedrock/deploy-config/sepolia.json
```diff
@@ -0,0 +1,41 @@
...
+  "proxyAdminOwner": "0xf80267194936da1E98dB10bcE06F3147D580a62e",
+  "baseFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "l1FeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "sequencerFeeVaultRecipient": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "baseFeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
+  "l1FeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
+  "sequencerFeeVaultMinimumWithdrawalAmount": "0x8ac7230489e80000",
+  "baseFeeVaultWithdrawalNetwork": 0,
+  "l1FeeVaultWithdrawalNetwork": 0,
+  "sequencerFeeVaultWithdrawalNetwork": 0,
```
protolambda, line 29: 0 = L1, 1 = L2 is kind of confusing. Seems good otherwise though.

packages/contracts-bedrock/deploy-config/sepolia.json
```diff
@@ -0,0 +1,41 @@
+{
+  "finalSystemOwner": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "portalGuardian": "0xBc1233d0C3e6B5d53Ab455cF65A6623F6dCd7e4f",
+  "l1StartingBlockTag": "",
```
protolambda: Are we picking this explicitly at a later point, or do we just leave it empty?

tynes: We need to explicitly pick this at a later point, after contract deployment
https://github.com/ethereum-optimism/optimism/blob/9af277e9d0ac7cf89d8409234241869241c1b0a3/op-node/cmd/genesis/cmd.go#L136-L140

packages/contracts-bedrock/deploy-config/sepolia.json
tynes: We could write a go type for handling this, probably useful in the longer term, human friendly string -> solidity type in `UnmarshalJSON`

tynes:
The last thing is the L2 chain id, decided to not reuse OP Kovan's

packages/contracts-bedrock/deploy-config/sepolia.json
tynes: @refcell is going to handle writing a go type that marshals `"local"` and `"remote"`

packages/contracts-bedrock/deploy-config/sepolia.json
```diff
@@ -0,0 +1,41 @@
...
+  "channelTimeout": 300,
+  "p2pSequencerAddress": "0x715b7219D986641DF9eFd9C7Ef01218D528e19ec",
+  "batchInboxAddress": "0xff00000000000000000000000000000011155420",
+  "batchSenderAddress": "0x7431310e026B69BFC676C0013E12A1A11411EEc9",
+  "l2OutputOracleSubmissionInterval": 120,
+  "l2OutputOracleStartingBlockNumber": 0,
+  "l2OutputOracleStartingTimestamp": 0,
+  "l2OutputOracleProposer": "0x02b1786A85Ec3f71fBbBa46507780dB7cF9014f6",
+  "l2OutputOracleChallenger": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301",
+  "finalizationPeriodSeconds": 12,
```
K-Ho, line 19: I like this value for ease of testing withdrawals,but curious if this needs to be a higher value for the testing we want to do internally on FPs

packages/contracts-bedrock/scripts/Deployer.sol
```diff
@@ -445,6 +445,10 @@ abstract contract Deployer is Script {
             return "devnetL1";
         } else if (chainid == 31337) {
             return "hardhat";
+        } else if (chainid == 11155111) {
+            return "sepolia";
+        } else if (chainid == 69) {
```
K-Ho: This isn't the same as the chain id above

packages/contracts-bedrock/deploy-config/sepolia.json
```diff
@@ -0,0 +1,41 @@
...
+  "baseFeeVaultWithdrawalNetwork": 0,
+  "l1FeeVaultWithdrawalNetwork": 0,
+  "sequencerFeeVaultWithdrawalNetwork": 0,
+  "gasPriceOracleOverhead": 188,
+  "gasPriceOracleScalar": 684000,
+  "enableGovernance": true,
+  "governanceTokenSymbol": "OP",
+  "governanceTokenName": "Optimism",
+  "governanceTokenOwner": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301",
+  "l2GenesisBlockGasLimit": "0x17D7840",
```
K-Ho: Should be `0x1C9C380` - 30m max gas limit, elasticity of 6, which gives us a target of 5m gas.

packages/contracts-bedrock/deploy-config/sepolia.json
```diff
@@ -0,0 +1,41 @@
...
+  "finalSystemOwner": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301",
+  "portalGuardian": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301",
+  "l1StartingBlockTag": "0x7c83e929deffe1b6a97cc9f9ef99845e6970a512b9ce18432d95d6f7f851ea53",
+  "l1ChainID": 11155111,
+  "l2ChainID": 11155420,
+  "l2BlockTime": 2,
+  "maxSequencerDrift": 600,
+  "sequencerWindowSize": 3600,
+  "channelTimeout": 300,
+  "p2pSequencerAddress": "0x715b7219D986641DF9eFd9C7Ef01218D528e19ec",
```
K-Ho, line 11: where does this value come from?

packages/contracts-bedrock/deploy-config/sepolia.json
```diff
@@ -0,0 +1,41 @@
...
+  "batchInboxAddress": "0xff00000000000000000000000000000011155420",
+  "batchSenderAddress": "0x7431310e026B69BFC676C0013E12A1A11411EEc9",
```
K-Ho, line 13: where does this value come from?

K-Ho (changes requested):

K-Ho (approved):

protolambda (approved):

PR #6429, opened 2023-07-25
contracts-bedrock: add network specific config tags

tynes:
**Description**

This commit adds network specific tags to the contract config that should be specific to individual networks. Right now there is config that is `immutable` but network specific, meaning that the same implementations cannot be used for all proxies across the superchain.

Closes CLI-4261


refcell (approved):
LGTM

tynes:
Going to close this PR and get the changes in with each individual contract's PR

PR #6431, opened 2023-07-25
op-e2e: enable shanghai on L1

tynes:
**Description**

Changes the genesis generation in `op-chain-ops` so that the L1 genesis includes shanghai. This impacts `op-e2e` that uses the libraries in `op-chain-ops` such that now `op-e2e` will run with shanghai enabled for L1. Note that shanghai is not yet enabled on L2. This commit updates the `fakepos` utility to use the post shanghai execution api.

This commit also deletes the `fork_test` in which a test existed for the upgrade to shapella but has already happened. Keeping that test around adds extra burden to the maintainers and isn't particularly necessary to keep around anymore now that the network has upgraded successfully.


op-chain-ops/deployer/deployer.go
```diff
@@ -40,11 +40,19 @@ type Deployment struct {
 
 type Deployer func(*backends.SimulatedBackend, *bind.TransactOpts, Constructor) (*types.Transaction, error)
 
+// NewBackend returns a SimulatedBackend suitable for L1. It has
+// the latest L1 hardforks enabled.
 func NewBackend() *backends.SimulatedBackend {
```
ajsutton: nit: Should we rename this to `NewL1Backend` to avoid people potentially using it for L2 by accident?

tynes: Good call, addressed this

ajsutton (approved):

Inphi (approved):

PR #6432, opened 2023-07-25
op-chain-ops: remove clique signer address from config

tynes:
**Description**

Having an alternate code path for using clique on L1 instead of fake pos only adds complexity. This commit removes the concept of the clique signer from the deploy config so that it cannot be misconfigured accidentally by setting the value.

### DIFF
--- a/ops-bedrock/entrypoint-l1.sh
+++ b/ops-bedrock/entrypoint-l1.sh
@@ -27,9 +27,15 @@ fi
 if [ ! -d "$GETH_CHAINDATA_DIR" ]; then
 	echo "$GETH_CHAINDATA_DIR missing, running init"
 	echo "Initializing genesis."
-	# Handle setting the clique config in the extra data
-	EXTRA_DATA="$BLOCK_SIGNER_ADDRESS""0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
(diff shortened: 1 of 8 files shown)


tynes:
Looks like we still use clique in the docker-compose based devnet and this change breaks that

tynes:
Converting to draft because this needs more work to get working e2e

ops-bedrock/entrypoint-l1.sh
```diff
@@ -27,6 +27,10 @@ fi
 if [ ! -d "$GETH_CHAINDATA_DIR" ]; then
 	echo "$GETH_CHAINDATA_DIR missing, running init"
 	echo "Initializing genesis."
+	# Handle setting the clique config in the extra data
+	EXTRA_DATA="$BLOCK_SIGNER_ADDRESS""0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
tynes: This needs the clique config to be set as well

refcell: Should there then be a check here that the "0x" prefix is _not_ prepended to the extra data? 

The removed go code would do that through its use of common.Address as I understand it.

tynes: you are right that its a bug there is no `0x` prefix

tynes:
Still broken

tynes:
We will not be able to do this until we have L1 setup without needing clique. It should be possible to do so if we used `op-wheel`

PR #6434, opened 2023-07-25
op-challenger: Add config and CLI args to support actually executing cannon

ajsutton:
**Description**

Adds `--cannon-bin`, `--cannon-prestate` and `--cannon-l2` args plus config to op-challenger. These will be required to actually execute cannon.  The L1 can be taken from the existing option.

Also tweaked the way tests and config creation work to better fit the fact that some options are required only when using a specific trace type, rather than just always specifying them.

**Tests**

Added unit tests.

**Metadata**

- https://linear.app/optimism/issue/CLI-4241/cli-for-cannon-inputs


refcell (approved):
LGTM!

PR #6435, opened 2023-07-25
op-bindings: storage layout entry getter

tynes:
**Description**

Adds a getter to the storage layout type for getting the layout entry by label. This is generally a useful operation when manipulating state manually, so writing a function for it here instead of rewriting the same code many times.

Part of https://github.com/ethereum-optimism/optimism/pull/6412 to make it easier to contribute to.


clabby (approved):

refcell (approved):
LGTM

PR #6436, opened 2023-07-25
op-chain-ops: deploy config copy

tynes:
**Description**

Add a method for deeply copying a `DeployConfig`.
This is useful for using a single source of truth
for deploy configs in `op-e2e`.

Part of https://github.com/ethereum-optimism/optimism/pull/6412

### DIFF
--- a/op-chain-ops/genesis/config.go
+++ b/op-chain-ops/genesis/config.go
@@ -7,6 +7,7 @@ import (
 	"math/big"
 	"os"
 	"path/filepath"
+	"reflect"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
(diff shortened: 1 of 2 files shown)


op-chain-ops/genesis/config.go
```diff
@@ -169,6 +169,89 @@ type DeployConfig struct {
 	FundDevAccounts bool `json:"fundDevAccounts"`
 }
 
+// Copy will deeply copy the DeployConfig
+func (d *DeployConfig) Copy() *DeployConfig {
```
ajsutton, line 174: Q: Would it be reasonable and lower-maintence to implement `Copy` by just serialising to JSON and back? It's probably not the end of the world to have to list every field here but if we can do the copy automatically that would be nicer and it doesn't seem like something that needs to be hugely optimised.

tynes: I definitely see your point, it would be far less code. It could probably also be implemented in a generic way using some `reflect` metaprogramming stuff. fwiw the test would catch the case where there are new values added, assuming they are not set to their zero value. I suppose i could change the implementation here

PR #6444, opened 2023-07-25
feat(op-challenger): Preimage Oracle Data Loading

refcell:
**Description**

This PR is the first in a stack to load the preimage oracle with
key-data pairs for cannon's required local and global preimages.

This defines a new method on the `TraceProvider` interface to query
this data to keep the `Solver` and `TraceProvider` implementations
decoupled.

A PR stacked ontop of this will build out the onchain calls made
before a step is attempted to load in this preimage oracle data.

These onchain calls are currently blocked by #6373 and are
stubbed out since the generated abis are not yet up to date.

**Metadata**

Fixes CLI-4243


refcell:
* develop
  * **PR #6444** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6444" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6444?utm_source=stack-comment).

op-challenger/fault/cannon/cannon_provider.go
```diff
@@ -52,6 +52,21 @@ func (p *CannonTraceProvider) Get(i uint64) (common.Hash, error) {
 	return value, nil
 }
 
+func (p *CannonTraceProvider) GetOracleData(i uint64) ([]byte, []byte, []byte, []byte, error) {
+	// TODO: Get the correct preimage oracle values out of the cannon trace
```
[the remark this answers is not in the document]
refcell: image

op-challenger/fault/agent.go
```diff
@@ -10,25 +10,16 @@ import (
...
 )
 
-// Responder takes a response action & executes.
-// For full op-challenger this means executing the transaction on chain.
-type Responder interface {
-	CanResolve(ctx context.Context) bool
-	Resolve(ctx context.Context) error
-	Respond(ctx context.Context, response types.Claim) error
-	Step(ctx context.Context, stepData types.StepCallData) error
-}
```
ajsutton, line 20: hmm, I suspect this should stay here given the go convention of declaring interfaces where they're used (the class implementing them doesn't need to reference the interface at all).

op-challenger/fault/solver/solver_test.go
```diff
@@ -100,6 +100,64 @@ func TestNextMove(t *testing.T) {
...
+			name:        "OracleData_ProviderError",
+			claim:       builder.Seq(false).Attack(false).Get(),
+			isLocal:     false,
+			key:         []byte{0x01},
+			data:        []byte{0x01},
+			expectedErr: fmt.Errorf("provider error"),
+		},
+	}
+	for _, tableTest := range tests {
+		t.Run(tableTest.name, func(t *testing.T) {
```
ajsutton, line 141: nit: To be defensive we should take a local reference to the tableTest value. It doesn't actually matter here but if the tests are made parallel in the future they wind up running with the wrong value in `tableTest` as it gets updated as the loop moves forward.
```suggestion
	for _, tableTest := range tests {
	    tableTest := tableTest
		t.Run(tableTest.name, func(t *testing.T) {
```

op-challenger/fault/solver/solver_test.go
```diff
@@ -100,6 +100,64 @@ func TestNextMove(t *testing.T) {
...
+			data:        []byte{0x07},
+			expectedErr: nil,
+		},
+		{
+			name:        "OracleData_ProviderError",
+			claim:       builder.Seq(false).Attack(false).Get(),
+			isLocal:     false,
+			key:         []byte{0x01},
+			data:        []byte{0x01},
+			expectedErr: fmt.Errorf("provider error"),
```
ajsutton, line 137: We probably should be expecting both an error and key/data.

op-challenger/fault/cannon/provider.go
```diff
@@ -52,6 +54,14 @@ func (p *CannonTraceProvider) Get(i uint64) (common.Hash, error) {
 	return value, nil
 }
 
+func (p *CannonTraceProvider) GetOracleData(i uint64) ([]byte, []byte, error) {
+	proof, err := p.loadProof(i)
+	if err != nil {
+		return nil, nil, err
+	}
+	return proof.OracleKey, proof.OracleValue, nil
+}
```
ajsutton, line 63: We should add some tests for this.

op-challenger/fault/solver/solver.go
```diff
@@ -46,6 +46,17 @@ func (s *Solver) NextMove(claim types.Claim, agreeWithClaimLevel bool) (*types.C
 	}
 }
 
+// OracleData returns the pre-image data to populate the oracle with
+// if any exists.
+func (s *Solver) OracleData(claim types.Claim) (types.PreimageOracleData, error) {
```
ajsutton, line 51: I suspect this should just be provided as part of `StepData` returned from `AttemptStep`. You only ever need it when you need to perform a step.

And then the trace provider could probably return it as part of `GetPreimage` (it may be worth introducing a struct there rather than just adding more and more return values mind you.)

op-challenger/fault/types/types.go
```diff
@@ -26,12 +27,44 @@ type StepCallData struct {
 	Proof      []byte
 }
 
+// PreimageOracleData encapsulates the preimage oracle data
+// to load into the onchain oracle.
+type PreimageOracleData struct {
```
ajsutton, line 32: Seems like maybe we should use this struct everywhere we're tracking the preimage key and data? It's usually better to be consistent with the type we use for something rather than sometimes having it wrapped and sometimes not.

op-challenger/fault/responder.go
```diff
@@ -79,6 +80,53 @@ func (r *faultResponder) BuildTx(ctx context.Context, response types.Claim) ([]b
 	}
 }
 
+// LoadOracleData loads the preimage oracle data into the onchain preimage
+// oracle contract.
+func (r *faultResponder) LoadOracleData(ctx context.Context, data types.PreimageOracleData) error {
```
ajsutton, line 85: I'd probably call this `PopulateOracleData` or `PrepareOracleData` - `Load` makes it sound like it will fetch and return the data.

ajsutton (commented):
I'd suggest trimming this PR back to just what we can have working and tested now. The steps I can see:

1. Add pre-image data to TraceProvider
2. Add pre-image data to Solver `AttemptStep`
3. Update agent to call the appropriate method on `Responder`
4. Actually implement the call in `Responder`.

They don't necessarily have to be all separate PRs, but given item 4 is blocked on having the bindings updated leave it for a separate PR.  I'd do 1&2 together and then may or may not include 3 in that first PR depending on how long it took me to get there.

refcell:
> I'd suggest trimming this PR back to just what we can have working and tested now. The steps I can see:
> 
> 
> 
> 1. Add pre-image data to TraceProvider
> 
> 2. Add pre-image data to Solver `AttemptStep`
> 
> 3. Update agent to call the appropriate method on `Responder`
> 
> 4. Actually implement the call in `Responder`.
> 
> 
> 
> They don't necessarily have to be all separate PRs, but given item 4 is blocked on having the bindings updated leave it for a separate PR.  I'd do 1&2 together and then may or may not include 3 in that first PR depending on how long it took me to get there.

Appreciate the suggestion, I think I'll build this out in a fresh stack tomorrow to keep the implementation clean.

refcell:
Closing in favor of #6458

PR #6445, opened 2023-07-25
op-chain-ops: add L1Deployments helper

tynes:
**Description**

Adds a type that represents the set of L1Deployments. Each contract, including proxies, are in the struct. This is meant to be used as part of https://github.com/ethereum-optimism/optimism/pull/6412

This L1Deployments struct represents the intermediate JSON file that is produced from the foundry scripts. The first deploy script produces this JSON file and then the second turns this JSON file into the set of hardhat-deploy style artifacts in the `deployments` directory.


refcell (approved):
LGTM

PR #6446, opened 2023-07-25
contracts-bedrock: migrate OptimismPortal single impl

tynes:
**Description**

Migrate the network specific config in the `OptimismPortal` from being `immutable` values into storage. This allows for a single implementation to be used for all proxies in the network.

As seen in the storage layout, no storage values are modified. Only new storage values are added. This is very important for ensuring the integrity of storage in the contract.

The deployment complexity is reduced by removing all constructor values and setting all config in the `initialize` function. The sentinel contract should use all null values for its fields so that it is not confused as the contract for an actual network.

The bindings are regnerated. A follow up commit will bump the minor semver version in the `OptimismPortal`

Closes CLI-4310


packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -105,25 +105,30 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
 
     /// @custom:semver 1.7.2
     /// @notice Constructs the OptimismPortal contract.
+    constructor() Semver(1, 7, 2) {
```
maurelian: Semver should be incremented.

refcell: Bumped a minor since it's still backwards-compatible.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
```diff
@@ -60,6 +51,15 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
...
     bool public paused;
 
+    /// @notice Address of the L2OutputOracle contract.
+    L2OutputOracle public L2_ORACLE;
+
+    /// @notice Address of the SystemConfig contract.
+    SystemConfig public SYSTEM_CONFIG;
+
+    /// @notice Address that has the ability to pause and unpause withdrawals.
+    address public GUARDIAN;
```
maurelian: These values are mostly used internally within the Bedrock system. Is there no way that we can use the proper casing on them now?

If not, at some point I think we'll need to do a major release that resets this.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: Was aware of this as I left a note in the top level comment: `The bindings are regnerated. A follow up commit will bump the minor semver version in the OptimismPortal`

maurelian: Sorry, need to get better at reading the descriptions.

packages/contracts-bedrock/contracts/L1/OptimismPortal.sol
tynes: This is a good question, I don't love the idea of adding a bunch of extra getters and being stuck with both sets always, then if we change the calls that our contracts make we need to be sure that we upgrade in the correct order or in a single tx so that it always works. Adding them now and then removing the old ones in the future seems like it could work, wondering if we will always be doing this dance of adding and removing. I'm sure that offchain apps may call these values

mslipper (approved):
LGTM from a Go perspective.

maurelian (approved):

packages/contracts-bedrock/src/L1/OptimismPortal.sol
```diff
@@ -105,39 +108,77 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
 
     /// @custom:semver 1.7.2
     /// @notice Constructs the OptimismPortal contract.
+    constructor() Semver(1, 7, 2) {
```
tynes: This needs a semver bump to `1.8.0`

packages/contracts-bedrock/src/L1/OptimismPortal.sol
```diff
@@ -60,6 +51,18 @@ contract OptimismPortal is Initializable, ResourceMetering, Semver {
     ///         This may be removed in the future.
     bool public paused;
 
+    /// @notice Address of the L2OutputOracle contract.
+    /// @custom:network-specific
+    L2OutputOracle internal _L2_ORACLE;
```
tynes: These are not immutables so they should not be snake case

refcell (approved):
Very nice 💯

tynes:
@Mergifyio refresh

tynes:
Thank you for review!

PR #6447, opened 2023-07-25
feat: migrate to forge fmt

tynes:
**Description**

WIP, will update

Running into issues where because the deps are installed via npm, its making the Go tests depend on the JS deps install
https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/22408/workflows/ced361ac-6d08-450c-9fa9-199cb0334392/jobs/974793

We don't want to be in a situation where the Go tests have to wait until JS deps install, that will slow down CI. Moving to `forge fmt` will help to decouple CI

This should be done in 2 commits, one updates the style and the other updates the patch version. It is important to see no bytecode diff in the first commit


tynes:
Closing this, will reopen

PR #6449, opened 2023-07-26
feat: gitmodules for contracts deps

tynes:
**Description**

Migrates away from using npm for contracts deps to git submodules. As much as people like to complain about git submodules, it is the best that we have and the industry standard for foundry projects. The main problem this is solving is not needing to have the Go tests depend on the installation of the JS deps to work


packages/contracts-bedrock/package.json
```diff
@@ -39,17 +39,10 @@
     "lint:fix": "pnpm lint:contracts:fix && pnpm lint:ts:fix",
     "lint": "pnpm lint:fix && pnpm lint:check"
   },
-  "dependencies": {
-    "@openzeppelin/contracts": "4.7.3",
-    "@openzeppelin/contracts-upgradeable": "4.7.3",
-    "@rari-capital/solmate": "github:transmissions11/solmate#8f9b23f8838670afda0fd8983f2c41e8037ae6bc",
-    "clones-with-immutable-args": "github:Saw-mon-and-Natalie/clones-with-immutable-args#105efee1b9127ed7f6fedf139e1fc796ce8791f2"
-  },
+  "dependencies": {},
```
refcell: Can we just remove the dependencies kv pair?

refcell (approved):
I believe this looks good to me now

tynes:
Need to figure out the issue with the docker build

tynes:
This unblocks https://github.com/ethereum-optimism/optimism/pull/6412 and speeds up contracts CI as it no longer depends on the typescript install for running tests/coverage/slither. The typescript deps still are present for the checks, that is something that we could remove in the future so that the entire contracts CI has no steps that it depends on

.circleci/config.yml
```diff
@@ -502,34 +456,12 @@ jobs:
...
-      - run:
-          name: Install yarn package manager
-          command: |
-            npm i yarn@1 --global
       - run:
           name: slither
           command: |
-            slither --version && pnpm slither || echo "export LINT_STATUS=1" >> "$BASH_ENV"
-      - run:
-          name: invariant docs
```
tynes, line 530: This is done twice, no need to do it twice, its also in `contracts-bedrock-checks`

.circleci/config.yml
```diff
@@ -502,34 +456,12 @@ jobs:
...
-      # packages installed via npm via github automatically run postpack scripts
-      # their postpack script happens to use yarn so we need it here
-      - run:
-          name: Install yarn package manager
-          command: |
-            npm i yarn@1 --global
       - run:
           name: slither
           command: |
-            slither --version && pnpm slither || echo "export LINT_STATUS=1" >> "$BASH_ENV"
```
tynes, line 528: The or case here can fail silently, preserving that behavior but no need to write something to the env in this case

refcell (approved):
Re-reviewed, looks good to me!

Notice for others, the invariant-docs ci check removal is a de-dup (see 2nd instance [here](https://github.com/ethereum-optimism/optimism/blob/b0e3ed19441df4e0d272fbda1b7f9df8b056d9b1/.circleci/config.yml#L420-L425))

.circleci/config.yml
refcell: See: https://github.com/ethereum-optimism/optimism/blob/b0e3ed19441df4e0d272fbda1b7f9df8b056d9b1/.circleci/config.yml#L420-L425

maurelian (approved):
LGTM

Can we also just bite the bullet and move `contracts -> src` now?

tynes:
> LGTM
> 
> Can we also just bite the bullet and move `contracts -> src` now?

Can follow up with another PR for this soon

PR #6451, opened 2023-07-26
op-challenger: Find snapshot to start execution at

ajsutton:
**Description**

Add logic to find the best available starting snapshot to use when executing cannon.

**Tests**

Unit tests around the search logic.

**Metadata**


refcell (approved):
I think this looks good - very nice testing 🎉 

Could possibly refactor out the io reads but maybe those optimizations are premature and the datadir could instead just be a fast db anyways.

PR #6452, opened 2023-07-26
op-program: Add L2 Chain ID to client boot

Inphi:
### Description
Replace rollup and chain configs with the l2 chain ID. The chain ID is used to read the appropriate config from a hardcoded preset.

### Dynamic Chain Configuration
The program behavior has also been changed to accommodate dynamic rollup and chain configs generated during e2e tests. If, and only if, the L2 chain ID is set to a specific _test_ ID, then the program will load rollup and chain configs from the local pre-image store rather than from the hardcoded preset.

### DIFF
--- a/op-program/client/boot.go
+++ b/op-program/client/boot.go
@@ -3,9 +3,11 @@ package client
 import (
 	"encoding/binary"
 	"encoding/json"
+	"math"
 
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
 	preimage "github.com/ethereum-optimism/optimism/op-preimage"
+	"github.com/ethereum-optimism/optimism/op-program/chainconfig"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/params"
 )
@@ -15,17 +17,25 @@ const (
 	L2OutputRootLocalIndex
 	L2ClaimLocalIndex
 	L2ClaimBlockNumberLocalIndex
+	L2ChainIDLocalIndex
+
+	// These local keys are only used for custom chains
 	L2ChainConfigLocalIndex
 	RollupConfigLocalIndex
 )
+
+// CustomChainIDIndicator is used to detect when the program should load custom chain configuration
+const CustomChainIDIndicator = uint64(math.MaxUint64)
 
 type BootInfo struct {
 	L1Head             common.Hash
 	L2OutputRoot       common.Hash
 	L2Claim            common.Hash
 	L2ClaimBlockNumber uint64
-	L2ChainConfig      *params.ChainConfig
-	RollupConfig       *rollup.Config
+	L2ChainID          uint64
+
+	L2ChainConfig *params.ChainConfig
+	RollupConfig  *rollup.Config
 }
 
 type oracleClient interface {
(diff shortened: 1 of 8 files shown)


op-program/chainconfig/testing.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+var (
+	TestChainID = uint64(math.MaxUint64)
+
+	testRollupConfigPath = "fpp_rollup_config.json"
+	testL2GenesisPath    = "fpp_l2_genesis.json"
```
ajsutton: Using fixed file paths is likely to cause conflicts between different tests.  Couple of options I can see:

1. Create a temp dir per test and write these files to the test specific temp dir (using these fixed names).  Then when executing `op-program` make sure its working directory is set to the temp dir.  The client program can then just read the hard coded file name and it will get the right file for that test.
2. Rather than reading from a file, if the chain ID is unknown, request the JSON network config from the pre-image oracle using another local key (ie fallback to the way we're doing it today).  On-chain only the chain ID would work and so maybe we require an additional flag to enable serving these JSON versions of the chain config or something like that.

ajsutton (commented):
We definitely need to preserve the ability to use custom network configs somehow for testing (and need to work out how to make it work for things like the local devnet on-chain too). I'm not sure I'd limit it to a single chain ID because we want people to be able to test this out on a bunch of different chains without having to merge the chain config into the op-program first. It may be better to just have an additional CLI option to enable loading chains that aren't built in - so you still default to safe but it can be used with any chain.

We might need to come up with a way to have cannon's `load-elf` command patch in a custom chain config to make things work on chain with other networks as we'll need to have something that works for the devnet.  We don't need to solve that in this PR though, just need something that keeps the existing e2e tests working.

Inphi:
* develop
  * **PR #6544** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6544" target="_blank">Graphite</a> 
    * **PR #6452** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6452" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6452?utm_source=stack-comment).

op-program/chainconfig/testing.go
Inphi: went with option 2. The diff turned out to be smaller than I expected.

op-program/client/boot.go
```diff
@@ -45,22 +55,39 @@ func (br *BootstrapClient) BootInfo() *BootInfo {
...
-	}
-	rollupConfig := new(rollup.Config)
-	err = json.Unmarshal(br.r.Get(RollupConfigLocalIndex), rollupConfig)
-	if err != nil {
-		panic("failed to bootstrap rollup config")
+	l2ChainID := binary.BigEndian.Uint64(br.r.Get(L2ChainIDLocalIndex))
+
+	var l2ChainConfig *params.ChainConfig
+	var rollupConfig *rollup.Config
+	if l2ChainID == CustomChainIDIndicator {
```
Inphi, line 62: `CustomChainIDIndicator` allows users to configure custom chains that have the same chain ID as a preset. Users that have such configs should feel bad but honestly I don't see a compelling reason to prohibit that usecase.

op-program/client/boot_test.go
```diff
@@ -40,10 +70,18 @@ func (o *mockBoostrapOracle) Get(key preimage.Key) []byte {
 		return o.b.L2Claim[:]
 	case L2ClaimBlockNumberLocalIndex.PreimageKey():
 		return binary.BigEndian.AppendUint64(nil, o.b.L2ClaimBlockNumber)
+	case L2ChainIDLocalIndex.PreimageKey():
+		return binary.BigEndian.AppendUint64(nil, o.b.L2ChainID)
 	case L2ChainConfigLocalIndex.PreimageKey():
+		if !o.custom {
+			panic("unexpected key")
```
ajsutton: Might be worth including the key value in the panic message or at least make the message say it was the chain config index so we can tell which case was hit. Line numbers might be enough but can easily wind up out of sync.

ajsutton (approved):
LGTM. Just the one small nit to make the panic messages more informative.

PR #6453, opened 2023-07-26
op-challenger: Add cannon-server and cannon-snapshot-freq flag

ajsutton:
**Description**

Add a `--cannon-server` flag to specify the binary to use as the pre-image oracle server when executing cannon.  Typically this will be `op-program`.
Add an optional `--canon-snapshot-freq` flag to control how frequently to generate snapshots when executing cannon.

**Tests**

Updated unit tests

**Metadata**

- https://linear.app/optimism/issue/CLI-4241/cli-for-cannon-inputs


op-challenger/flags/flags.go
```diff
@@ -78,6 +83,12 @@ var (
 		Usage:   "L2 Address of L2 JSON-RPC endpoint to use (eth and debug namespace required)  (cannon trace type only)",
 		EnvVars: prefixEnvVars("CANNON_L2"),
 	}
+	CannonSnapshotFreqFlag = &cli.UintFlag{
+		Name:    "cannon-snapshot-freq",
+		Usage:   "Frequency of cannon snapshots to generate in VM steps (cannon trace type only)",
+		EnvVars: prefixEnvVars("CANNON_SNAPSHOT_FREQ"),
+		Value:   config.DefaultCannonSnapshotFreq,
+	}
```
refcell, line 91: Should this be frequency (the number of runs per unit of time) or instead expressed as an interval?

Frequency has a unit of cycles/unit time which isn't very expressive and confusing as opposed to interval which is just a unit of time.

refcell (commented):
Small verbiage question, otherwise looks great!

op-challenger/flags/flags.go
ajsutton: No it definitely has to be number of VM steps, not time as that's what cannon uses. Plus the time varies depending on what else is running on the machine so it's a lot easier to reason about how many VM steps between snapshots instead.

refcell (approved):

op-challenger/flags/flags.go
refcell: Got it thanks for clarifying. All makes sense to me then.

PR #6456, opened 2023-07-26
op-node: safer genesis script

tynes:
**Description**

Updates the `op-node` genesis gen script to prevent `nil` pointer access accidentally. Run the command like so:

```go
go run op-node/cmd/main.go genesis l2 \
  --deploy-config $PWD/packages/contracts-bedrock/deploy-config/mainnet.json \
  --outfile.l2 l2.json \
  --outfile.rollup rollup.json \
  --l1-rpc $ETH_RPC_URL \
  --deployment-dir $PWD/packages/contracts-bedrock/deployments/mainnet
```

When generating the genesis file, it is ok to leave the `l1StartingBlockTag` value as undefined, this script will select an appropriate value and then the value will end up inside of the rollup config under `.genesis.l1.hash` which can then be backported to the deploy config to make the script deterministic.


clabby (approved):
nice!

Inphi (approved):

PR #6469, opened 2023-07-27
chore(c-mon): rename state batch to output

maurelian:
**Description**

Updates some outdated language in fault-mon


tynes:
@maurelian This looks like it needs some linting

hamdiallam (approved):

PR #6470, opened 2023-07-27
contracts-bedrock: foundry-ify

tynes:
**Description**

Move the `contracts` directory to `src` and then move the tests into a top level `test` dir so that its
standard foundry style for a repo.

This is going to break any in flight PRs so worth trying to get in relatively quickly, will be no fun to rebase a bunch of PRs.

No semver bumps required because no production source code is changing


refcell (commented):
King, looks good to me, but will await other's review before approving

tynes:
@maurelian asked for this this morning

tynes:
The last thing is the artifacts output path but not changing that as part of this PR as it will require a diff in typescript code

packages/contracts-bedrock/foundry.toml
```diff
@@ -1,8 +1,9 @@
 [profile.default]
-src = 'contracts'
+src = 'src'
```
maurelian, line 2: probably can delete this?

maurelian (approved):
No bindings changed, so I checked the diff in the storage layout, everything LGTM.
Thanks for this!

PR #6471, opened 2023-07-27
op-challenger: Pass a context through to TraceProvider

ajsutton:
**Description**

Pass a `context.Context` through to trace provider methods. Will enable trace generation processes to terminate cleanly when op-challenger is shut down.

**Metadata**


refcell (approved):
Makes sense to me!

PR #6476, opened 2023-07-27
op-challenger: Execute cannon

ajsutton:
**Description**

Adds the ability to execute cannon to generate proofs.  Will always error out currently because we don't yet have the local game inputs loaded so can't pass them to the pre-image oracle server.

Cannon output is sent to the op-challenger logger.

**Tests**

Added unit tests.

**Metadata**


refcell (approved):
LGTM

PR #6538, opened 2023-07-30
ci: move lint command to makefile

tynes:
**Description**

CI had the command for running linting hardcoded in the CI config. This commit migrates to using `make lint` across all of the services to abstract away from CI what sort of linting is happening, also makes it possible to run the linter locally.

Ideally there is a single place where the lint command is defined, we could have makefiles that import makefiles


refcell (approved):

ajsutton (approved):
🎉

PR #6545, opened 2023-07-31
feat(ctb): `BlockOracle` + L1 head validation

clabby:
## Overview

Adds a new contract, `BlockOracle`, that allows for persisting blockhashes further back than the `BLOCKHASH` opcode for use in the `FaultDisputeGame` and the verification logic for the L1 head hash input.

**Rationale**
Cannon takes a while to run, and it may be unreasonable to expect that all challenge agents have hardware that can crunch out a state hash committing to the full trace in $256 \text{ blocks} \times 13 \text{ seconds/block} \times \frac{1 \text{ minute}}{60 \text{ seconds}} = 55.46 \text{ minutes}$ given that they'll be participating in multiple games. Instead of relying entirely on the `BLOCKHASH` opcode, we can add a block hash oracle that allows for the challenge agent to store the block hash they'll start Cannon from via `BlockHashOracle.store(<block_number>)` and reference its hash > 256 blocks after the block has been processed. 

In addition to having a verified L1 head hash, we also need to ensure that it contains the disputed output root. This verification is done via comparing the disputed output root's timestamp with the L1 head block's child timestamp.

| Condition | Description |
| --------- | ------------|
| `childTimestamp < disputedTimestamp` | The checkpoint was made before the proposal, invalid input. |
| `childTimestamp == disputedTimestamp` | The checkpoint was made in the same block as the proposal, invalid input. |
| `childTimestamp > disputedTimestamp` | The checkpoint was made at least 1 block after the proposal, valid input. |

**Metadata**
Closes CLI-4308


clabby:
* develop
  * **PR #6545** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6545" target="_blank">Graphite</a>  👈
    * **PR #6585** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6585" target="_blank">Graphite</a> 
      * **PR #6586** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6586" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6545?utm_source=stack-comment).

packages/contracts-bedrock/src/dispute/BlockHashOracle.sol
```diff
@@ -0,0 +1,30 @@
...
+pragma solidity 0.8.15;
+
+import "../libraries/DisputeTypes.sol";
+import "../libraries/DisputeErrors.sol";
+
+/// @title BlockHashOracle
+/// @notice Stores a map of block numbers => block hashes for use in dispute resolution
+contract BlockHashOracle {
+    /// @notice Maps block numbers to block hashes
+    mapping(uint256 => Hash) internal blockHashes;
```
tynes: internal prefixed with `_`?

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -425,8 +438,8 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
             })
         );
 
-        // Set the L1 head hash at the time of the game's creation.
-        l1Head = Hash.wrap(blockhash(block.number - 1));
+        // Set the L1 head hash to the block hash of the L1 block number provided.
+        l1Head = BLOCK_HASH_ORACLE.load(_getArgUint256(0x40));
```
tynes: What happens if this reverts?

clabby: Purposefully don't handle reverts gracefully here, a game needs the L1 block hash. This revert will bubble up to the `DisputeGameFactory` and the game won't be created.

packages/contracts-bedrock/src/dispute/BlockHashOracle.sol
clabby: Yeah is that in our style guide? Not opposed, but don't think this style is defined.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -294,6 +301,11 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
         l2BlockNumber_ = _getArgUint256(0x20);
     }
 
+    /// @inheritdoc IFaultDisputeGame
+    function l1BlockNumber() public pure returns (uint256 l1BlockNumber_) {
```
tynes, line 326: What is the set of valid l1 block numbers given a particular dispute game? It just impacts how the game is played correct?

packages/contracts-bedrock/src/dispute/BlockHashOracle.sol
tynes: Not sure

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
tynes: Do you mind adding this as a comment?

packages/contracts-bedrock/src/dispute/BlockHashOracle.sol
tynes: maybe its for internal functions, i think its underspecified in this case. not super opinionated

packages/contracts-bedrock/src/dispute/BlockHashOracle.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import "../libraries/DisputeTypes.sol";
+import "../libraries/DisputeErrors.sol";
+
+/// @title BlockHashOracle
```
tynes: do you think its worth emitting an event with blocknum + msg.sender? Would make it easier to track who is providing this service, because providing this service is def a prime candidate for retro funding

clabby: Yeah this could be neat. We don't need the mapping to be complete, sparse is ok - the idea is to have people that are creating dispute games first store the block hash that they'll be running cannon from here, run cannon, and then create the game with the l1 block number set to the number associated with the block hash.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
clabby: This L1 block number represents the block that the creator of the game begins running the program from (see `op-program`'s `--l1.head` flag), so there's no verification to perform on it. If a game is created with a L1 block number that is invalid for the dispute, another game can be created in its place due to the `l1BlockNumber` being a part of `extraData`, which is included in the construction of the UUID hash.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -425,8 +438,11 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
             })
         );
 
-        // Set the L1 head hash at the time of the game's creation.
-        l1Head = Hash.wrap(blockhash(block.number - 1));
+        // Set the L1 head hash to the block hash of the L1 block number provided.
+        // This call can revert if the block hash oracle does not have information
+        // about the block number provided to it. This revert will bubble up to the
+        // DisputeGameFactory and prevent the game from being created.
+        l1Head = BLOCK_HASH_ORACLE.load(_getArgUint256(0x40));
```
Inphi: nit: Unless this is much more gas efficient, dedup code here by calling `l1BlockNumber()`

op-e2e/e2eutils/disputegame/deploy.go
```diff
@@ -50,18 +53,68 @@ func deployDisputeGameContracts(require *require.Assertions, ctx context.Context
...
 	tx, err = factory.SetImplementation(opts, faultGameType, faultDisputeGameAddr)
 	require.NoError(err)
-
 	_, err = utils.WaitReceiptOK(ctx, client, tx.Hash())
 	require.NoError(err, "wait for final transaction to be included and OK")
 
-	return factory
+	// Wait for the block hash oracle to be ready to receive the checkpoint.
+	// The block time is hard coded at 13 seconds, should parameterize.
+	time.Sleep(15 * time.Second)
```
refcell: Waiting in tests is an anti-pattern, can we parametrize this a callback to allow the caller to roll blocks instead of sleeping.

eg:
```golang
func deployDisputeGameContracts(require *require.Assertions, ctx context.Context, client *ethclient.Client, opts *bind.TransactOpts, gameDuration uint64, oracleDelay func()) (*bindings.DisputeGameFactory, *big.Int) {
	// ...

	/ Wait for the block hash oracle to be ready to receive the checkpoint.
	// The block time is hard coded at 13 seconds, should parameterize.
	oracleDelay()

	// ...
}
```

refcell: There may be a more ergonomic pattern used in our e2e codebase, just don't remember off the top of my head.

packages/contracts-bedrock/src/dispute/BlockOracle.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import "../libraries/DisputeTypes.sol";
+import "../libraries/DisputeErrors.sol";
```
refcell: Let's start to style imports consistently as globals like you did in the Deploy script.

```suggestion
import "src/libraries/DisputeTypes.sol";
import "src/libraries/DisputeErrors.sol";
```

packages/contracts-bedrock/src/dispute/BlockOracle.sol
```diff
@@ -0,0 +1,46 @@
...
+
+        // Estimate the timestamp of the block assuming an average block time of 13 seconds.
+        Timestamp estimatedTimestamp = Timestamp.wrap(uint64(block.timestamp - 13));
+
+        // Persist the block information.
+        blocks[blockNumber_] = BlockInfo({
+            hash: Hash.wrap(blockHash),
+            timestamp: estimatedTimestamp
+        });
+    }
```
refcell: Can we not cache the current block timestamp here and then just use that as a backwards reference instead of using this 13 second block time estimation?

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -68,19 +78,24 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
     /// @param _vm An onchain VM that performs single instruction steps on a fault proof program
     ///            trace.
     /// @param _l2oo The trusted L2OutputOracle contract.
-    /// @custom:semver 0.0.4
+    /// @param _blockOracle The block oracle, used for loading block hashes further back
+    ///                     than the `BLOCKHASH` opcode allows as well as their estimated
+    ///                     timestamps.
+    /// @custom:semver 0.0.6
```
refcell, line 84: Why did we bump this two patch versions?

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
refcell: Can an invalid l1 block number challenge agents to respond and subsequently result in this invalid dispute game to resolve incorrectly since the "honest" actors clocks run out?

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -425,8 +459,46 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
             })
         );
 
-        // Set the L1 head hash at the time of the game's creation.
-        l1Head = Hash.wrap(blockhash(block.number - 1));
+        // Grab the index of the output proposal that commits to the starting L2 head.
+        // All outputs after this one are disputed.
+        // TODO(clabby): This is 2 calls too many for the information we need. Maybe
+        //               add a function to the L2OO?
```
refcell, line 468: This should be tracked with a linear ticket where discussion can be had (in the future github issues - very important to not use close sourced linear software etc etc)

My 2c: it's not important enough to add logic to the L2OO - keep it plain and simple - minimal diff > small gas costs.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -425,8 +459,46 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
...
+        // TODO(clabby): The block hash bisection game will allow us to dispute the first output
+        //               root by using genesis as the starting point. For now, it is critical that
+        //               the first proposed output root of an OP stack chain is done so by an
+        //               honest party.
+        uint256 proposalIdx = L2_OUTPUT_ORACLE.getL2OutputIndexAfter(l2BlockNumber());
+        Types.OutputProposal memory starting = L2_OUTPUT_ORACLE.getL2Output(proposalIdx - 1);
+        Types.OutputProposal memory disputed = L2_OUTPUT_ORACLE.getL2Output(proposalIdx);
+        // SAFETY: This call can revert if the block hash oracle does not have information
```
refcell, line 477: nit - hard to read without nice spacing
```suggestion

        // SAFETY: This call can revert if the block hash oracle does not have information
```

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -425,8 +459,46 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
...
+            index: uint128(proposalIdx - 1),
+            l2BlockNumber: starting.l2BlockNumber,
+            outputRoot: Hash.wrap(starting.outputRoot)
+        });
+        proposals[1] = OutputProposal({
+            index: uint128(proposalIdx),
+            l2BlockNumber: disputed.l2BlockNumber,
+            outputRoot: Hash.wrap(disputed.outputRoot)
+        });
+        // Persist the L1 head hash of the L1 block number provided.
```
refcell, line 515: ibid nit
```suggestion

        // Persist the L1 head hash of the L1 block number provided.
```

packages/contracts-bedrock/src/dispute/interfaces/IFaultDisputeGame.sol
```diff
@@ -63,6 +71,8 @@ interface IFaultDisputeGame is IDisputeGame {
 
     /// @notice The l2BlockNumber that the `rootClaim` commits to. The trace being bisected within
     ///         the game is from `l2BlockNumber - 1` -> `l2BlockNumber`.
-    /// @return l2BlockNumber_ The l2BlockNumber that the `rootClaim` commits to.
     function l2BlockNumber() external view returns (uint256 l2BlockNumber_);
+
+    /// @notice The l1BlockNumber that Cannon was ran from to generate the root claim.
+    function l1BlockNumber() external view returns (uint256 l1BlockNumber_);
```
refcell, line 83: We should be natspeccing where we can including return values to be consistent with our own style guides, but not a big deal

packages/contracts-bedrock/src/libraries/DisputeErrors.sol
```diff
@@ -16,7 +16,7 @@ error NoImplementation(GameType gameType);
 error GameAlreadyExists(Hash uuid);
 
 ////////////////////////////////////////////////////////////////
-//               `DisputeGame_Fault.sol` Errors               //
+//               `FaultDisputeGame.sol` Errors                //
```
refcell: I think we should remove the `.sol` here - i think it looks ugly and illegible

packages/contracts-bedrock/src/libraries/DisputeErrors.sol
```diff
@@ -72,3 +76,14 @@ error AlreadyChallenged();
 /// @notice Thrown when a function that is protected by the `onlyOwner` modifier
 ///          is called from an account other than the owner.
 error NotOwner();
+
+////////////////////////////////////////////////////////////////
+//                  `BlockOracle.sol` Errors                  //
```
refcell: ibid

packages/contracts-bedrock/test/BlockOracle.t.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Test } from "forge-std/Test.sol";
+import { BlockOracle } from "src/dispute/BlockOracle.sol";
+import "src/libraries/DisputeTypes.sol";
+import "src/libraries/DisputeErrors.sol";
```
refcell, line 7: nice use of global import paths

refcell (changes requested):
Fairly thorough review - will message offline to keep discussions efficient

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
clabby: This PR was stacked, the other one was merged into this due to overlap.

packages/contracts-bedrock/src/libraries/DisputeErrors.sol
clabby: Let's do this in another PR.

packages/contracts-bedrock/src/dispute/interfaces/IFaultDisputeGame.sol
clabby: In a few other PR's, you nitted that the `/// @return` for these self explanatory getters was unnecessary - do we want to put something in the style guide for this?

op-e2e/e2eutils/disputegame/deploy.go
Inphi: The idiom is to use the deterministic system clock. There's one in [System](https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/setup.go#L219) that you can use.

op-e2e/e2eutils/disputegame/deploy.go
```diff
@@ -50,18 +53,68 @@ func deployDisputeGameContracts(require *require.Assertions, ctx context.Context
 	alphaVMAddr, err := bind.WaitDeployed(ctx, client, tx)
 	require.NoError(err)
 
+	l2OutputOracle, err := bindings.NewL2OutputOracle(config.L1Deployments.L2OutputOracle, client)
```
Inphi: nit
```suggestion
	l2OutputOracle, err := bindings.NewL2OutputOracle(config.L1Deployments.L2OutputOracleProxy, client)
```

packages/contracts-bedrock/src/dispute/BlockOracle.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    /// @notice Stores a block hash for the previous block number.
+    /// @return blockNumber_ The block number that was checkpointed, which is always
+    ///                      `block.number - 1`.
+    function checkpoint() external returns (uint256 blockNumber_) {
+        // SAFETY: This block hash will always be accessible by the `BLOCKHASH` opcode,
+        //         and in the case of `block.number = 0`, we'll underflow.
+        // Persist the block information.
+        blocks[blockNumber_] = BlockInfo({
+            hash: Hash.wrap(blockhash(blockNumber_ = block.number - 1)),
```
Inphi: nit: much easier to read if the `blockNumber_` assignment stmt were on a separate line.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -59,6 +64,11 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
     /// @notice An append-only array of all claims made during the dispute game.
     ClaimData[] public claimData;
 
+    /// @notice The starting and disputed output proposal for the game. Includes information about
+    ///         the output indexes in the `L2OutputOracle` and the output roots at the time of
+    ///         game creation.
+    OutputProposal[2] public proposals;
```
Inphi: nit: Much clearer to use separate fields wit descriptive names (`agreedProposal`, `disputedProposal`) rather than a fixed array.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -266,21 +281,31 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
...
-            // Load the earliest output root that commits to the passed L2 block number
-            // into the game's local context in the preimage oracle.
-            Types.OutputProposal memory proposal = L2_OUTPUT_ORACLE.getL2OutputAfter(
-                l2BlockNumber()
+        } else if (_ident < 4) {
+            // Grab the index of the proposal to load at the given ident.
+            // Ident 2 loads the starting output root, and ident 3 loads the disputed output root.
+            uint256 proposal;
+            assembly {
+                proposal := iszero(mod(_ident, 2))
```
Inphi: nit: We can avoid using assembly here if the proposals were separated out rather than stored in an array

packages/contracts-bedrock/src/dispute/BlockHashOracle.sol
tynes: Got it, makes sense

packages/contracts-bedrock/src/dispute/BlockOracle.sol
```diff
@@ -0,0 +1,40 @@
...
+        blockInfo_ = blocks[_blockNumber];
+        if (Hash.unwrap(blockInfo_.hash) == 0) revert BlockHashNotPresent();
+    }
+
+    /// @notice Stores a block hash for the previous block number.
+    /// @return blockNumber_ The block number that was checkpointed, which is always
+    ///                      `block.number - 1`.
+    function checkpoint() external returns (uint256 blockNumber_) {
+        // SAFETY: This block hash will always be accessible by the `BLOCKHASH` opcode,
+        //         and in the case of `block.number = 0`, we'll underflow.
```
tynes, line 40: To know which blocknumber was stored, the caller will need to know the implementation details and look at the receipt. Easier to emit a log here?

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
tynes: its the same amount of storage, i suppose just wanting to do it in a single call to get both? More descriptive does make sense

Inphi: Yeah, and you still need to specify the array index to access it anyways.

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -719,10 +720,10 @@ contract Deploy is Deployer {
                 _maxGameDepth: cfg.faultGameMaxDepth(),
                 _gameDuration: Duration.wrap(uint64(cfg.faultGameMaxDuration())),
                 _vm: faultVm,
-                _l2oo: L2OutputOracle(mustGetAddress("L2OutputOracleProxy"))
+                _l2oo: L2OutputOracle(mustGetAddress("L2OutputOracleProxy")),
+                _blockOracle: new BlockOracle()
```
tynes, line 724: We should have a step `deployBlockOracle`?

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -41,6 +42,10 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
     /// @notice The trusted L2OutputOracle contract.
     L2OutputOracle public immutable L2_OUTPUT_ORACLE;
 
+    /// @notice The block hash oracle, used for loading block hashes further back
+    ///         than the `BLOCKHASH` opcode allows as well as their estimated timestamps.
+    BlockOracle public immutable BLOCK_ORACLE;
```
tynes, line 47: Note that this doesn't follow our updated ABI guidelines where we don't leak the impl details into the ABI

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -68,19 +78,24 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
...
     constructor(
         Claim _absolutePrestate,
         uint256 _maxGameDepth,
         Duration _gameDuration,
         IBigStepper _vm,
-        L2OutputOracle _l2oo
-    ) Semver(0, 0, 4) {
+        L2OutputOracle _l2oo,
+        BlockOracle _blockOracle
+    ) Semver(0, 0, 6) {
```
tynes, line 92: This bumps 2 versions but isn't a big deal

tynes: Since its pre major v1

packages/contracts-bedrock/scripts/Deploy.s.sol
clabby: Doing in another PR.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
clabby: This was intentional - squashed the stack due to overlap, but there were 2 versions worth of changes here.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
clabby: Will update all dispute immutables to be compliant in a separate PR - there are others as well.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
refcell: it feels to me like versions should be bumped only when the changes are merged into trunk, otherwise you get this weird version gap in trunk.

But again not a big deal, just shouldn't do in the future imo.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
refcell: Answered offline

packages/contracts-bedrock/src/dispute/interfaces/IFaultDisputeGame.sol
refcell: Oops, totally my fault. Yes, in a linear ticket, let's decide on whether to document in a linear ticket and then add to the style guide.

Created ticket in gardening: https://linear.app/optimism/issue/CLI-4323/docs-decide-and-add-to-style-guide-when-returns-are-worth-documenting

packages/contracts-bedrock/test/BlockOracle.t.sol
refcell: fwiw, created a ticket to track: https://linear.app/optimism/issue/CLI-4324/ctb-use-global-import-paths-and-document-in-style-guide cc @clabby

refcell (approved):

clabby:
@mergifyio requeue

tynes (approved):

clabby:
@mergifyio requeue

Inphi (approved):

PR #6547, opened 2023-07-31
feat(op-challenger): Load the absolute prestate from cannon

refcell:
**Description**

Loads the absolute prestate for a fault dispute game
into the op-challenger using the `state.json` output
from cannon.

**Tests**

Cannon `AbsolutePreState` unit tests.

**Metadata**

Fixes CLI-4238

### DIFF
--- a/op-challenger/fault/cannon/provider.go
+++ b/op-challenger/fault/cannon/provider.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/ethereum-optimism/optimism/cannon/mipsevm"
 	"github.com/ethereum-optimism/optimism/op-challenger/config"
 	"github.com/ethereum-optimism/optimism/op-challenger/fault/types"
 	"github.com/ethereum/go-ethereum/common"
@@ -34,12 +35,14 @@ type ProofGenerator interface {
 
 type CannonTraceProvider struct {
 	dir       string
+	prestate  string
 	generator ProofGenerator
 }
 
 func NewTraceProvider(logger log.Logger, cfg *config.Config) *CannonTraceProvider {
 	return &CannonTraceProvider{
 		dir:       cfg.CannonDatadir,
+		prestate:  cfg.CannonAbsolutePreState,
 		generator: NewExecutor(logger, cfg),
 	}
 }
@@ -82,8 +85,19 @@ func (p *CannonTraceProvider) GetPreimage(ctx context.Context, i uint64) ([]byte
 	return value, data, nil
 }
 
-func (p *CannonTraceProvider) AbsolutePreState(ctx context.Context) []byte {
-	panic("absolute prestate not yet supported")
+func (p *CannonTraceProvider) AbsolutePreState(ctx context.Context) ([]byte, error) {
+	path := filepath.Join(p.dir, p.prestate)
+	file, err := os.Open(path)
+	if err != nil {
+		return []byte{}, fmt.Errorf("cannot open state file (%v): %w", path, err)
+	}
+	defer file.Close()
+	var state mipsevm.State
+	err = json.NewDecoder(file).Decode(&state)
+	if err != nil {
+		return []byte{}, fmt.Errorf("invalid mipsevm state (%v): %w", path, err)
+	}
+	return state.EncodeWitness(), nil
 }
 
 func (p *CannonTraceProvider) loadProof(ctx context.Context, i uint64) (*proofData, error) {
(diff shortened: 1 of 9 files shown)


refcell:
* develop
  * **PR #6547** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6547" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6547?utm_source=stack-comment).

ajsutton:
You've got the wrong issue link. :)

op-challenger/fault/types/types.go
```diff
@@ -60,7 +60,7 @@ type TraceProvider interface {
 	GetPreimage(ctx context.Context, i uint64) (preimage []byte, proofData []byte, err error)
 
 	// AbsolutePreState is the pre-image value of the trace that transitions to the trace value at index 0
-	AbsolutePreState(ctx context.Context) []byte
+	AbsolutePreState(ctx context.Context) (common.Hash, error)
```
trianglesphere: Is it the pre-image (as it was previously) or is it the hash? It might be that we should provide the hash, but if so, we need to update the comment on this function.

op-challenger/fault/cannon/provider.go
```diff
@@ -82,8 +84,19 @@ func (p *CannonTraceProvider) GetPreimage(ctx context.Context, i uint64) ([]byte
...
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("cannot open state file (%v): %w", path, err)
+	}
+	defer file.Close()
+	var state mipsevm.State
+	err = json.NewDecoder(file).Decode(&state)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("invalid mipsevm state (%v): %w", path, err)
+	}
+	return state.PreimageKey, nil
```
trianglesphere: what is `state.Preimagekey`?

trianglesphere (commented):
I left a couple comments, I could just be out of the loop - I haven't tracked what cannon has been doing.

op-challenger/fault/types/types.go
ajsutton: We need to be providing the pre-image of the absolute pre-state here.  It should be the output of cannon's `state.EncodeWitness` function (https://github.com/ethereum-optimism/optimism/blob/ba1c236f94e3254fee4d39a488147346d92ef16d/cannon/mipsevm/state.go#L40).  It's the pre-image provided to the step function and the contract will verify that its hash matches the absolute pre-image _claim_ specified when setting up the dispute game type.

The complication here is that the pre-state pre-image we return here is the binary encoded witness but the `--input` state passed when executing cannon needs to be the JSON file which also includes the full memory state (the memory is accessed via the proof data on-chain since it's too large to post it all).

The question is whether we get Cannon to add the binary witness to the state it generates from `load-elf` or if we just add the ability to parse the state JSON and compute the witness to challenger (essentially pull in that state.json either by just copying it or via a module reference). If we're going to verify the pre-state actually matches we probably need the logic in challenger, otherwise we can verify the binary witness matches the pre-state hash but can't verify the state in the JSON actually matches the witness. It kind of sucks to pull so much of cannon into challenger though and is quite possibly going to cause some problematic dependency cycles.

So maybe the best option is to add a new subcommand to Cannon to convert a JSON state to the encoded witness. Then the challenger can be given the state as JSON to use with `--input` and run this subcommand to convert it to the encoded witness, guaranteeing it matches the JSON content.

refcell: Took the most minimal-diff approach to this whereby the cannon trace provider reads the state.json file and unmarshals it into a cannon `mipsevm.State` which we can just call `EncodeWitness()` on.

ajsutton:
@refcell Looks like the docker build doesn't make the cannon code available.

op-challenger/fault/cannon/provider.go
```diff
@@ -82,8 +84,19 @@ func (p *CannonTraceProvider) GetPreimage(ctx context.Context, i uint64) ([]byte
 	return value, data, nil
 }
 
-func (p *CannonTraceProvider) AbsolutePreState(ctx context.Context) []byte {
-	panic("absolute prestate not yet supported")
+func (p *CannonTraceProvider) AbsolutePreState(ctx context.Context) ([]byte, error) {
+	path := filepath.Join(p.dir, elfState)
```
ajsutton: This should be coming from the config options - `config.CannonAbsolutePreState`. We have the config in `NewCannonTraceProvider` so just need to grab the value on the way through (currently it just passes everything down to the executor).

ajsutton (commented):
Looks good, just need to use the configured pre-state rather than the hard coded path.

ajsutton (approved):
LGTM.

PR #6553, opened 2023-08-01
feat(ctb): L1 head verification in game initialization

clabby:
## Overview

Adds verification for the L1 head input to the `FaultDisputeGame`.

**Metadata**
Closes CLI-4307


clabby:
* clabby/ctb/local-data-correctness
  * **PR #6553** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6553" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6553?utm_source=stack-comment).

packages/contracts-bedrock/src/dispute/BlockOracle.sol
```diff
@@ -0,0 +1,48 @@
...
+    function store(uint256 _blockNumber) external {
+        // Fetch the block hash for the given block number and revert if it is out of
+        // the `BLOCKHASH` opcode's range.
+        bytes32 blockHash = blockhash(_blockNumber);
+        if (blockHash == 0) revert BlockNumberOOB();
+
+        // Estimate the timestamp of the block assuming an average block time of 13 seconds.
+        Timestamp estimatedTimestamp = Timestamp.wrap(
+            uint64(block.timestamp - ((block.number - _blockNumber) * 13))
+        );
```
ajsutton: I don't think this is sufficient validation. Blocks will always be some multiple of 12 seconds apart and the gap could be very large when there are many missed blocks. If we have:

```
Slot     ->    Block Timestamp
Slot     ->   Block Number  ->  Block Timestamp
1        ->   25            ->  50
2        ->   <empty>       ->  (would have been 62)
3        ->   26            ->  74
```
Then calling `store(1)` in a tx in slot 3/block 26 would set the timestamp to `74 - 13 = 61` which is incorrect. By making the block look to have a later timestamp than it really did you could potentially use a L1 head that actually occurred prior to the output root being published.

clabby: Great points. I think the way around doing the estimation is by adding the L1 block number to the `Types.OutputProposal` struct in the L2OO. For now, I've restricted the block checkpointing to 1 block back from the current block, which should reduce the margin of error for the time being.

ajsutton: Bringing this to the public record:
The requirement for an output proposal is that it is submitted on chain _after_ the required batches are all submitted to L1. The L1 head needs to be, at the earliest, the last block that _batch data_ is included in. So the earliest valid L1 head to use when disputing a proposal is actually the block before the proposal was submitted.

Based on that, the block hash oracle `store` method should record the parent block hash and the current timestamp, and we validate that the timestamp in the block hash oracle is equal to or greater than the timestamp in the output oracle. The earliest valid call to the `store` method would then be one made in the same block that the proposal is made, and it will record the block hash of the block prior to the proposal which, as above, must still contain all the required data to derive the output root.

The nice thing about that, is that it means when we do come to modify the L2OO, we can get it to store the block hash which is the exact input we need rather than the block number which should work out but creates extra complexity in the challenger around dealing with reorgs between when it gets the block number from the oracle and when it finds the referenced root.

clabby:
Merging into #6545 because there's so much overlap.

PR #6555, opened 2023-08-01
contracts-bedrock: system config single implementation

tynes:
**Description**

Solve the address discoverability by adding getters on the system config for the L1 proxies. This is very easy to accomplish, now the user only needs to know the address of the systemconfig with a particular op stack chain to know all of the L1 contract addresses. It makes tracking all of the chains in the superchain much easier.

Closes CLI-4313


packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -34,6 +34,12 @@ contract SystemConfig is OwnableUpgradeable, Semver {
     ///         proof to fetch this value.
     bytes32 public constant UNSAFE_BLOCK_SIGNER_SLOT = keccak256("systemconfig.unsafeblocksigner");
 
+    bytes32 public constant L1_CROSS_DOMAIN_MESSENGER_SLOT = keccak256("systemconfig.l1crossdomainmessenger");
+    bytes32 public constant L1_ERC_721_BRIDGE_SLOT = keccak256("systemconfig.l1erc721bridge");
+    bytes32 public constant L1_STANDARD_BRIDGE_SLOT = keccak256("systemconfig.l1standardbridge");
+    bytes32 public constant L2_OUTPUT_ORACLE_SLOT = keccak256("systemconfig.l2outputoracle");
+    bytes32 public constant OPTIMISM_PORTAL_SLOT = keccak256("systemconfig.optimismportal");
+
```
maurelian, line 71: Should these be [offset by 1](https://github.com/ethereum-optimism/optimism/blob/a5cdb4d081d4e91f2ff7fbb1cf50d2bf00fc6a52/packages/contracts-bedrock/contracts/universal/Proxy.sol#L10) like in [erc1967](https://eips.ethereum.org/EIPS/eip-1967)?

maurelian:
It's worth noting that most of these addresses are also stored in the `ProxyAdmin` contract. Perhaps the `ProxyAdmin` should just point to the `SystemConfig` to look up these addresses.

packages/contracts-bedrock/src/L1/SystemConfig.sol
clabby: Should be safe to not do this given that the contract has no possibility of arbitrary storage slot writes other than `_setAddress`, which is internal and only called in `initialize`. Might be nice to add so that the preimage of the slot value is not known in the case of a second preimage attack being discovered on sha3 + future compat with new features, though.

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,6 +184,14 @@ contract SystemConfig is OwnableUpgradeable, Semver {
         return addr;
     }
 
+    /// @notice
```
clabby:
```suggestion
    /// @notice Stores an address in an arbitrary storage slot, `_slot`.
    /// @param _addr The address to store
    /// @param _slot The storage slot to store the address in.
    /// @dev WARNING! This function must be used cautiously, as it allows for overwriting values
    ///      in arbitrary storage slots. Solc will add checks that the data passed as `_addr`
    ///      is 20 bytes or less. 
```

tynes:
> It's worth noting that most of these addresses are also stored in the `ProxyAdmin` contract. Perhaps the `ProxyAdmin` should just point to the `SystemConfig` to look up these addresses.

What do you mean by this? I'm confused - the proxy admin doesn't have a list of addresses. The name mapping is only set for the single legacy resolved delegate proxy that is used

packages/contracts-bedrock/src/L1/SystemConfig.sol
tynes: It may eventually be possible for users to define inputs to things that are placed in storage, would only be an issue if a second preimage attack is found

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -34,6 +44,24 @@ contract SystemConfig is OwnableUpgradeable, Semver {
     ///         proof to fetch this value.
     bytes32 public constant UNSAFE_BLOCK_SIGNER_SLOT = keccak256("systemconfig.unsafeblocksigner");
 
+    /// @notice Storage slot that the L1CrossDomainMessenger address is stored at.
+    bytes32 public constant L1_CROSS_DOMAIN_MESSENGER_SLOT = keccak256("systemconfig.l1crossdomainmessenger");
```
tynes: offset by 1 to prevent 2nd preimage attack

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -34,6 +44,24 @@ contract SystemConfig is OwnableUpgradeable, Semver {
     ///         proof to fetch this value.
     bytes32 public constant UNSAFE_BLOCK_SIGNER_SLOT = keccak256("systemconfig.unsafeblocksigner");
```
tynes, line 50: todo: come up with migration plan for this

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+    ///         solc's automation.
+    /// @param _slot The storage slot to retrieve the address from.
+    function _getAddress(bytes32 _slot) internal view returns (address) {
+        address addr;
+        bytes32 slot = _slot;
+        assembly {
+            addr := sload(slot)
+        }
+        return addr;
+    }
```
refcell: nit: named return args
```suggestion
    function _getAddress(bytes32 _slot) internal view returns (address addr_) {
        bytes32 slot = _slot;
        assembly {
            addr_ := sload(slot)
        }
    }
```

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+    /// @notice Getter for the L1CrossDomainMessenger address.
+    function l1CrossDomainMessenger() external view returns (address) {
+        return _getAddress(L1_CROSS_DOMAIN_MESSENGER_SLOT);
+    }
```
refcell: nit: named return args
```suggestion
    function l1CrossDomainMessenger() external view returns (address addr_) {
        addr_ = _getAddress(L1_CROSS_DOMAIN_MESSENGER_SLOT);
    }
```

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+    /// @notice Getter for the L1ERC721Bridge address.
+    function l1ERC721Bridge() external view returns (address) {
+        return _getAddress(L1_ERC_721_BRIDGE_SLOT);
+    }
```
refcell: nit: named return args
```suggestion
    function l1ERC721Bridge() external view returns (address addr_) {
        addr_ = _getAddress(L1_ERC_721_BRIDGE_SLOT);
    }
```

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+    /// @notice Getter for the L1StandardBridge address.
+    function l1StandardBridge() external view returns (address) {
+        return _getAddress(L1_STANDARD_BRIDGE_SLOT);
+    }
```
refcell: nit: named return args
```suggestion
    function l1StandardBridge() external view returns (address addr_) {
       addr_ = _getAddress(L1_STANDARD_BRIDGE_SLOT);
    }
```

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+    /// @notice Getter for the L2OutputOracle address.
+    function l2OutputOracle() external view returns (address) {
+        return _getAddress(L2_OUTPUT_ORACLE_SLOT);
+    }
```
refcell: nit: named return args
```suggestion
    function l2OutputOracle() external view returns (address addr_) {
        addr_ = _getAddress(L2_OUTPUT_ORACLE_SLOT);
    }
```

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+    /// @notice Getter for the OptimismPortal address.
+    function optimismPortal() external view returns (address) {
+        return _getAddress(OPTIMISM_PORTAL_SLOT);
+    }
```
refcell: nit: named return args
```suggestion
    function optimismPortal() external view returns (address addr_) {
        addr_ = _getAddress(OPTIMISM_PORTAL_SLOT);
    }
```

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+    /// @notice Getter for the OptimismMintableERC20Factory address.
+    function optimismMintableERC20Factory() external view returns (address) {
+        return _getAddress(OPTIMISM_MINTABLE_ERC20_FACTORY_SLOT);
+    }
```
refcell: nit: named return args
```suggestion
    function optimismMintableERC20Factory() external view returns (address addr_) {
        addr_ = _getAddress(OPTIMISM_MINTABLE_ERC20_FACTORY_SLOT);
    }
```

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
+    /// @notice Getter for the BatchInbox address.
+    function batchInbox() external view returns (address) {
+        return _getAddress(BATCH_INBOX_SLOT);
+    }
```
refcell: nit: named return args
```suggestion
    function batchInbox() external view returns (address addr_) {
        addr_ = _getAddress(BATCH_INBOX_SLOT);
    }
```

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -25,15 +25,58 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+
     /// @notice Version identifier, used for upgrades.
     uint256 public constant VERSION = 0;
 
     /// @notice Storage slot that the unsafe block signer is stored at.
     ///         Storing it at this deterministic storage slot allows for decoupling the storage
     ///         layout from the way that `solc` lays out storage. The `op-node` uses a storage
     ///         proof to fetch this value.
+    ///         NOTE: this value will be migrated to another storage slot in a future version.
```
clabby: nit: maybe use an `@dev` natspec tag here so that the docs are clean when we deploy `forge doc` soon?

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -141,10 +213,85 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
+    function _setStartBlock(uint256 _startBlock) internal {
+        if (_startBlock != 0) {
+            // There is an override, it cannot already be set.
+            require(startBlock == 0, "SystemConfig: cannot override an already set start block");
+            startBlock = _startBlock;
+        } else if (startBlock == 0) {
+            // There is no override and it is not set in storage. Set it to the block number.
+            startBlock = block.number;
+        }
+    }
```
refcell: Can we change this into the following? The current dispatch logic is a little wonky to reason about. Note, the following changes one codepath whereby currently if the `_startBlock == 0` and `startBlock != 0`, nothing will happen, in the following suggestion, the top-level require statement will cause a revert if the `startBlock != 0` (even if `_startBlock == 0`)
```suggestion
    /// @notice Sets the start block in a backwards compatible way. Proxies
    ///         that were initialized before this existed will not have it
    ///         be set by `block.number`.
    function _setStartBlock(uint256 _startBlock) internal {
        require(startBlock == 0, "SystemConfig: cannot override an already set start block");
        if (_startBlock != 0) {
            // There is an override, it cannot already be set.
            startBlock = _startBlock;
        } else {
            // There is no override and it is not set in storage. Set it to the block number.
            startBlock = block.number;
        }
    }
```

refcell (changes requested):

packages/contracts-bedrock/src/L1/SystemConfig.sol
clabby: Discussed in slack - I'm a fan of this change. @tynes, does pulling the revert out of the first branch affect the reinitializer? That's the only reason I could think of why this logic was wonky.

tynes: I do think that this is better, great suggestion @refcell - we will need to update the source anyways to reinit, once all contracts have this storage set then we can remove the concept of the override. There is def some storage layout magic that we need to do

clabby (approved):
LGTM

refcell (approved):

PR #6565, opened 2023-08-02
fix(ctb): revert if fee vault recipient reverts.

maurelian:
Fixes an issue in the account of FeeVault.sol where insufficient gas could result in the value of totalProcessed being incremented, even if the RECIPIENT reverts.

**Tests**

A test was added which works just by having the RECIPIENT revert (rather than by sending insufficient gas). Since these conditions are indistinguishable from the CALLER contract, this should be fine.


packages/contracts-bedrock/src/universal/FeeVault.sol
```diff
@@ -76,7 +76,10 @@ abstract contract FeeVault {
         emit Withdrawal(value, RECIPIENT, msg.sender, WITHDRAWAL_NETWORK);
 
         if (WITHDRAWAL_NETWORK == WithdrawalNetwork.L2) {
-            SafeCall.send(RECIPIENT, gasleft(), value);
+            require(
+                SafeCall.send(RECIPIENT, gasleft(), value),
```
maurelian: I kept this to minimize the diff, but I think there's a good argument for just checking the status of `(bool success, ) = RECIPIENT.call(hex"")`. 

The `SafeCall` lib arguably led to this bug being introduced by hiding the silent failure.

maurelian: And that we definitely won't be setting the `RECIPIENT` address to a contract that returns so much memory that we OOG here.

packages/contracts-bedrock/test/SequencerFeeVault.t.sol
```diff
@@ -1,8 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
+import { console } from "forge-std/console.sol";
```
tynes: nit: leftover console

packages/contracts-bedrock/src/universal/FeeVault.sol
tynes: This is true, i suppose there was no invariant defined here

tynes: But you are correct that the bug here results in the total processed variable being incorrect which is not ideal

tynes: depending on what the invariant is, you could also do:

```
bool success = safecall();
if (success) { totalProcessed + balance }
```

maurelian: Do you mean `SafeCall.send()` where you use `safecall()`?

packages/contracts-bedrock/src/universal/FeeVault.sol
```diff
@@ -76,7 +76,10 @@ abstract contract FeeVault {
...
+                "FeeVault: failed to send ETH to L2"
```
dapperscene6: nit: could we change this to `FeeVault: failed to send ETH to L2 fee recipient` to make what failed clearer?

dapperscene6:
@maurelian should we also bump the three FeeVaults `Semver` versions in this PR?

packages/contracts-bedrock/src/universal/FeeVault.sol
tynes: yes, was just typing out pseudocode

tynes:
> @maurelian should we also bump the three FeeVaults `Semver` versions in this PR?

Yes we definitely should

packages/contracts-bedrock/src/universal/FeeVault.sol
maurelian: So I ended up doing the following, because: 
1. we can remove an import
2. reverting on failure is consistent with how solidity handles a failure if the withdrawal is to L1 via `L2StandardBridge.bridgeETHTo()`. 

```
  (bool success, ) = RECIPIENT.call{ value: value }(hex"");
  require(success, "FeeVault: failed to send ETH to L2 fee recipient");
```

anikaraghu (commented):
thanks for doing this 👍

dapperscene6:
LGTM

packages/contracts-bedrock/src/universal/FeeVault.sol
tynes: I don't think removing an import is a particularly strong argument but not super opinionated, I do think using safecall is generally better because it will be cheaper

packages/contracts-bedrock/src/L2/L1FeeVault.sol
```diff
@@ -9,7 +9,7 @@ import { FeeVault } from "../universal/FeeVault.sol";
 /// @title L1FeeVault
 /// @notice The L1FeeVault accumulates the L1 portion of the transaction fees.
 contract L1FeeVault is FeeVault, Semver {
-    /// @custom:semver 1.2.1
```
tynes, line 12: Semver should be `1.3.0`, patch is reset on minor bump

maurelian: whoops

maurelian: got it.

packages/contracts-bedrock/src/universal/FeeVault.sol
maurelian: I tested it and `SafeCall.send()` save 31 gas vs the current implementation. I would prefer to stick with this more explicit approach.

tynes (approved):

PR #6566, opened 2023-08-02
feat: enable L2OutputOracle impl usage by multiple proxies

tynes:
**Description**

Allows for the `L2OutputOracle` to be usable by multiple proxies and update the ABI in a backwards compatible way.

- The storage layout is not altered, only updated
- `reinitializer(2)` allows for the same proxy to be initialized again, this number must increment

Closes https://linear.app/optimism/issue/CLI-4311/l2outputoracle


packages/contracts-bedrock/test/L2OutputOracle.t.sol
```diff
@@ -61,13 +48,41 @@ contract L2OutputOracle_constructor_Test is L2OutputOracle_Initializer {
...
             _l2BlockTime: l2BlockTime,
-            _startingBlockNumber: startingBlockNumber,
-            _startingTimestamp: block.timestamp,
-            _proposer: proposer,
-            _challenger: owner,
             _finalizationPeriodSeconds: 7 days
         });
     }
+
+    /* For some reason, this test causes foundry to stack overflow
```
tynes: Will come back to this

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
```diff
@@ -37,6 +31,14 @@ contract L2OutputOracle is Initializable, Semver {
...
     Types.OutputProposal[] internal l2Outputs;
 
+    /// @notice The address of the challenger. Can be updated via upgrade.
+    /// @custom:network-specific
+    address internal _CHALLENGER;
+
+    /// @notice The address of the proposer. Can be updated via upgrade.
+    /// @custom:network-specific
+    address internal _PROPOSER;
+
```
tynes, line 50: These can technically have their style changed now

op-proposer/proposer/abi_test.go
```diff
@@ -34,10 +34,7 @@ func setupL2OutputOracle() (common.Address, *bind.TransactOpts, *backends.Simula
 		big.NewInt(10),
 		big.NewInt(2),
 		big.NewInt(0),
-		big.NewInt(0),
-		from,
-		common.Address{0xdd},
-		big.NewInt(100))
+	)
```
refcell: I don't think this is correct. The constructor of the `L2OutputOracle` changed to accept args like so:
```diff
        uint256 _submissionInterval,
        uint256 _l2BlockTime,
-       uint256 _startingBlockNumber,
-       uint256 _startingTimestamp,
-       address _proposer,
-       address _challenger,
        uint256 _finalizationPeriodSeconds
```

So, I believe this should be:
```suggestion
		big.NewInt(10),
		big.NewInt(2),
		big.NewInt(100))
	)
```

packages/contracts-bedrock/test/L2OutputOracle.t.sol
```diff
@@ -61,13 +48,23 @@ contract L2OutputOracle_constructor_Test is L2OutputOracle_Initializer {
...
-            _startingBlockNumber: startingBlockNumber,
-            _startingTimestamp: block.timestamp,
-            _proposer: proposer,
-            _challenger: owner,
             _finalizationPeriodSeconds: 7 days
         });
     }
+
+    /// @dev Tests that the constructor reverts if the starting timestamp is invalid.
+    function test_constructor_badTimestamp_reverts() external {
```
refcell: nit
```suggestion
    /// @dev Tests that the initialize function reverts if the starting timestamp is invalid.
    function test_initialize_badTimestamp_reverts() external {
```

refcell (changes requested):
Couple of comments and nits, otherwise looks good. Nice catch on the chain-mon helper update.

op-proposer/proposer/abi_test.go
tynes: Great catch here

refcell (approved):

mslipper (approved):

PR #6460, opened 2023-07-26
feat(op-challenger): Responder Preimage Oracle Data Population

refcell:
**Description**

This PR implements the methods to populate onchain preimage oracle with
local and global key-value preimage data.

Right now, data packing methods are stubbed out since this PR is blocked
by having the updated `FaultDisputeGame` and `PreimageOracle` contract 
abis in order to pack the transaction data. This is implemented in PR
#6373 which is a blocker for this to be implemented.

**Tests**

Missing tests right now since methods are stubs.

**Metadata**

Fixes CLI-4243

PR #6571, opened 2023-08-02
feat(op-challenger): Cannon OracleUpdater Tx Crafting

refcell:
**Description**

This PR continues the stack to replace #6460.

It updates the cannon `OracleUpdater` component
to dispatch on local vs global preimage oracle
data, build the tx, and send it through the tx
manager.

**Tests**

Unit tests around the cannon `OracleUpdater`
implementation.

**Metadata**

Fixes CLI-4243.

PR #6572, opened 2023-08-02
feat(op-challenger): Wire the Oracle Data Updating into the Fault Agent

refcell:
**Description**

This PR is the final stacked branch in the replacement of #6460.

It wires the oracle data call and onchain tx sending into the fault
agent component.

**Metadata**

Fixes CLI-4243.


refcell:
* develop
  * **PR #6588** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6588" target="_blank">Graphite</a> 
    * **PR #6571** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6571" target="_blank">Graphite</a> 
      * **PR #6572** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6572" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6572?utm_source=stack-comment).

op-challenger/fault/agent.go
```diff
@@ -132,6 +134,17 @@ func (a *Agent) step(ctx context.Context, claim types.Claim, game types.Game) er
...
 
+	oracleData, err := a.solver.GetOracleData(ctx, claim)
+	if err != nil {
+		a.log.Debug("Failed to get oracle data", "err", err)
+		return nil
+	}
+
+	a.log.Info("Updating oracle data", "oracleKey", oracleData.OracleKey, "oracleData", oracleData.OracleData)
+	if a.updater.UpdateOracle(ctx, *oracleData) != nil {
+		return fmt.Errorf("load oracle data: %w", err)
```
trianglesphere:
```suggestion
		return fmt.Errorf("failed to load oracle data: %w", err)
```

op-challenger/scripts/alphabet/charlie.sh
```diff
@@ -12,11 +12,12 @@ CHARLIE_KEY="74feb147d72bfae943e6b4e483410933d9e447d5dc47d52432dcc2c1454dabb7"
 MALLORY_ADDRESS="0x4641c704a6c743f73ee1f36C7568Fbf4b80681e4"
 MALLORY_KEY="28d7045146193f5f4eeb151c4843544b1b0d30a7ac1680c845a416fac65a7715"
 
-FAULT_GAME_ADDRESS="0x8daf17a20c9dba35f005b6324f493785d239719d"
+FAULT_GAME_ADDRESS=$(cat ../../.fault-game-address)
```
trianglesphere: maybe this could be passed in, but I'm not a huge fan of having a new hidden file with this configuration. It makes it a lot harder for people not very familiar with this code to run the demo.

trianglesphere:
Also, this is two big unrelated changes: loading the oracle data & modifying the runner scripts.

op-challenger/scripts/alphabet/charlie.sh
refcell: You should literally just be able to run "make alphabet" since the file is read by subsequent scripts.

refcell:
> Also, this is two big unrelated changes: loading the oracle data & modifying the runner scripts.

Yes, I am getting it to a state ready for next weeks demo - didn't have time to split up and would like to still put adrian in a good place on monday...

Changes are minimal and easy to intuit about though. Also diff is very much in the acceptable range so I think it's not a big deal to stress over this.

op-challenger/fault/agent.go
ajsutton: This isn't assigning the returned error to `err`.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,117 @@
...
+echo "Verifying that there are at least 2 outputs in the L2 Output Oracle..."
+INDEX=$(cast call $L2_OUTPUT_ORACLE_PROXY "getL2OutputIndexAfter(uint256)" $BLOCK_NUMBER)
+INDEX=$(cast to-dec $INDEX)
+echo "Index: $INDEX"
+
+# We will use the l2 block number of 1 for the dispute game.
+# We need to check that the block oracle contains the corresponding l1 block number.
+echo "Checkpointing the block oracle..."
+L1_BLOCK_NUMBER=$(cast send --private-key $DEVNET_SPONSOR $BLOCK_ORACLE_PROXY "checkpoint()" --json | jq .blockNumber)
+L1_BLOCK_NUMBER=$(echo $L1_BLOCK_NUMBER | tr -d '"')
```
ajsutton: I think you can just use:
```suggestion
L1_BLOCK_NUMBER=$(cast send --private-key $DEVNET_SPONSOR $BLOCK_ORACLE_PROXY "checkpoint()" --json | jq -r .blockNumber)
```

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,117 @@
...
+L1_BLOCK_NUMBER=$(cast to-dec $L1_BLOCK_NUMBER)
+echo "L1 Block Number (dec): $L1_BLOCK_NUMBER"
+((EXTRA_DATA_ARG=L1_BLOCK_NUMBER-1))
+
+echo "Getting the l2 output at index 0"
+((PRIOR_INDEX=INDEX-1))
+cast call $L2_OUTPUT_ORACLE_PROXY "getL2Output(uint256)" $PRIOR_INDEX
+
+echo "Getting the l2 output at index $INDEX"
+cast call $L2_OUTPUT_ORACLE_PROXY "getL2Output(uint256)" $INDEX
```
ajsutton: Why are we doing this? We seem to be discarding the results anyway.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,117 @@
...
+ROOT_CLAIM=$(cast keccak $(cast abi-encode "f(uint256,uint256)" 15 122))
+
+# fault dispute game extra data:
+# abi.encode(uint256(l2_block_number), uint256(l1_block_number))
+EXTRA_DATA=$(cast abi-encode "f(uint256,uint256)" $BLOCK_NUMBER $EXTRA_DATA_ARG)
+
+echo "Initializing the game"
+FAULT_GAME_ADDRESS=$(cast call --private-key $MALLORY_KEY $DISPUTE_GAME_PROXY "create(uint8,bytes32,bytes)" $GAME_TYPE $ROOT_CLAIM $EXTRA_DATA)
+echo "Creating game at address $FAULT_GAME_ADDRESS"
+cast send --private-key $MALLORY_KEY $DISPUTE_GAME_PROXY "create(uint8,bytes32,bytes)" $GAME_TYPE $ROOT_CLAIM $EXTRA_DATA
```
ajsutton: We seem to be creating two games here.

ajsutton (commented):
I'd agree that we should split this up into two separate PRs. The pre-image loading just needs a small tweak to the error handling and I think it's good to go, but the game playing scripts probably need a bit more thought to get something maintainable.

We may have reached the point where it's too complex to maintain a bash script to init games and we'd be better off just writing that part as a standalone go file that creates the game and prints just the game address.  Then I'd pass the address as arguments to the mallory and charlie scripts rather than reading a hidden file.  It's cool that we can just run `make alphabet` and have it all work but I think these scripts are useful because they're closer to real world so it's kind of nice to execute the two challenger instances separately. Creating the game is currently out of scope for the challenger so we need to have some temporary tooling for that.

We have e2e tests that run the whole process automatically so can use those if we want a single command to execute it all.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,117 @@
...
+echo "----------------------------------------------------------------"
+echo " - Starting a new devnet"
+echo "----------------------------------------------------------------"
+
+make cannon-prestate
+make devnet-up
+
+DEVNET_SPONSOR="ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
+DISPUTE_GAME_PROXY=$(jq .DisputeGameFactoryProxy .devnet/addresses.json)
+DISPUTE_GAME_PROXY=$(echo $DISPUTE_GAME_PROXY | tr -d '"')
```
ajsutton:
```suggestion
DISPUTE_GAME_PROXY=$(jq -r .DisputeGameFactoryProxy .devnet/addresses.json)
```

should work here too

ajsutton:
Ok this kind of sucks to do in go as well, so the bash init script may still be the way to go, unless we can reuse the e2e utilities to deploy the game.

op-challenger/scripts/alphabet/init_game.sh
refcell: The cast call returns the game address and doesn't actually send the transaction- I just used what was previously in the init_game script which was a cast call then a cast send

ajsutton (approved):

PR #6573, opened 2023-08-02
contracts-bedrock: single implementation L1StandardBridge

tynes:
**Description**

Closes CLI-4315


tynes:
This is going to require changes to the L2 genesis gen script, need to initialize the L2 contract

packages/contracts-bedrock/src/L1/L1StandardBridge.sol
```diff
@@ -76,13 +77,36 @@ contract L1StandardBridge is StandardBridge, Semver {
...
+    modifier clearLegacySlot() {
+        assembly {
+            sstore(0, 0)
+        }
+        _;
+    }
+
+    /// @notice Initializer
+    ///         The fix modifier should be removed during the next contract upgrade.
+    function initialize(CrossDomainMessenger _messenger) public clearLegacySlot() reinitializer(2) {
```
tynes: this should be the l1xdm for better type safety

op-chain-ops/cmd/check-l2/main.go
```diff
@@ -793,3 +797,31 @@ func checkPredeploy(client *ethclient.Client, i uint64) error {
...
+	}
+	slot := common.BigToHash(big.NewInt(int64(entry.Slot)))
+	value, err := client.StorageAt(context.Background(), addr, slot, nil)
+	if err != nil {
+		return nil, err
+	}
+	if entry.Offset+typ.NumberOfBytes > uint(len(value)) {
+		return nil, fmt.Errorf("value length is too short")
+	}
+	initialized := new(big.Int).SetBytes(value[entry.Offset : entry.Offset+typ.NumberOfBytes])
```
tynes: This calculation is off

tynes: i think the endianness needs to be swapped

op-bindings/solc/types.go
```diff
@@ -64,6 +64,17 @@ func (s *StorageLayout) GetStorageLayoutEntry(name string) (StorageLayoutEntry,
...
 
+// GetStorageLayoutType returns the StorageLayoutType where the label matches
+// the provided name.
+func (s *StorageLayout) GetStorageLayoutType(name string) (StorageLayoutType, error) {
+	for typeName, typ := range s.Types {
+		if typeName == name {
+			return typ, nil
+		}
+	}
+	return StorageLayoutType{}, fmt.Errorf("%s not found", name)
```
refcell: It looks like `s.Types` is a `map[string]StorageLayoutType`.
https://github.com/ethereum-optimism/optimism/blob/1403c7d7d66d11ceb517d2d05505c5b5a46e2882/op-bindings/solc/types.go#L53
Can we instead of looping over the map and checking equality of keys, just explicitly grab the item from the map as suggested below?
```suggestion
    if ty, ok := s.Types[name]; ok {
          return ty, nil
    }
	return StorageLayoutType{}, fmt.Errorf("%s not found", name)
```

packages/contracts-bedrock/src/L1/L1StandardBridge.sol
```diff
@@ -76,13 +77,32 @@ contract L1StandardBridge is StandardBridge, Semver {
...
+    /// @custom:semver 1.2.0
     /// @notice Constructs the L1StandardBridge contract.
-    /// @param _messenger Address of the L1CrossDomainMessenger.
-    constructor(address payable _messenger)
-        Semver(1, 1, 1)
-        StandardBridge(_messenger, payable(Predeploys.L2_STANDARD_BRIDGE))
-    {}
+    constructor()
+        Semver(1, 2, 0)
+        StandardBridge(StandardBridge(payable(Predeploys.L2_STANDARD_BRIDGE)))
```
refcell, line 84: Should this not just be the following?
```suggestion
        StandardBridge(payable(Predeploys.L2_STANDARD_BRIDGE))
```

packages/contracts-bedrock/src/universal/StandardBridge.sol
```diff
@@ -40,10 +38,13 @@ abstract contract StandardBridge {
     /// @notice Mapping that stores deposits for a given pair of local and remote tokens.
     mapping(address => mapping(address => uint256)) public deposits;
 
+    /// @notice Messenger contract on this domain.
+    CrossDomainMessenger internal _MESSENGER;
```
refcell: Should this have a `@custom:network-specific` natspec tag?

packages/contracts-bedrock/src/universal/StandardBridge.sol
```diff
@@ -9,28 +9,26 @@ import { SafeCall } from "../libraries/SafeCall.sol";
...
 
     /// @notice The L2 gas limit set when eth is depoisited using the receive() function.
     uint32 internal constant RECEIVE_DEFAULT_GAS_LIMIT = 200_000;
 
-    /// @notice Messenger contract on this domain.
-    CrossDomainMessenger public immutable MESSENGER;
-
     /// @notice Corresponding bridge on the other domain.
-    StandardBridge public immutable OTHER_BRIDGE;
+    StandardBridge internal immutable _OTHER_BRIDGE;
```
refcell: Should this have a `@custom:network-specific` natspec tag?

packages/contracts-bedrock/test/StandardBridge.t.sol
```diff
@@ -14,7 +14,7 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 ///         internal functions so they can be more easily tested directly.
 contract StandardBridgeTester is StandardBridge {
     constructor(address payable _messenger, address payable _otherBridge)
-        StandardBridge(_messenger, _otherBridge)
+        StandardBridge(StandardBridge(_otherBridge))
```
refcell, line 17: ibid, can we just do
```suggestion
        StandardBridge(_otherBridge)
```

refcell (commented):
Couple comments and questions 👍

op-bindings/solc/types.go
tynes: Yes that is definitely better, good catch

packages/contracts-bedrock/src/universal/StandardBridge.sol
tynes: Yes, good catch

tynes:
Fixed up this based on review, will push another commit addressing the final stuff

packages/contracts-bedrock/test/StandardBridge.t.sol
tynes: The constructor was updated to accept a `StandardBridge`? This type conversion is required right? `otherBridge` is an address

packages/contracts-bedrock/src/universal/StandardBridge.sol
tynes: Addressed

tynes:
Getting a race condition in `op-chain-ops` tests again

refcell (approved):
Looks good 💯

PR #6581, opened 2023-08-03
op-node,op-program: source rollup-configs from superchain-registry

protolambda:
**Description**

To make OP-Stack chain configurations available across OP-Stack client implementations, we are experimenting with a "superchain-registry": one place that hosts all superchain configurations, which can then be embedded into the software for convenience & OP-Stack wide rollout of config changes.

This depends on the [op-geth superchain-registry integration](https://github.com/ethereum-optimism/op-geth/pull/113)

This will enable op-node users to pick any chain from the registered Superchains as chain to run the op-node with: no more manual `rollup.json` required. This simplifies the configuration-management: network upgrades can be specified in superchain-wide configuration.

`rollup.json` might change, but we plan to keep supporting manual configuration, for testing / bootstrap / customization purposes, the OP-Stack is open after all.

The `op-program` is not in production yet, and thus gets full access to the superchain-registry data.
The `op-node` is in production, and while the superchain-registry is experimental, does not expose the full selection of chains.

To access a configuration from the superchain-registry, add `--beta.extra-networks` to your selection of `op-node` flags, and then use `--network` to select a chain like normal.

The default-network configurations are tested to match the expected configurations, to ensure the op-node does not behave differently, even with the superchain-registry dependency.

**Metadata**

Fix CLI-4207


op-node/chaincfg/chains.go
```diff
@@ -2,135 +2,100 @@ package chaincfg
...
+	return []string{"op-mainnet", "op-goerli", "op-sepolia"}
+}
+
+// BetaAvailableNetworks returns all available network configurations in the superchain-registry.
+// This set of configurations is experimental, and may change at any time.
+func BetaAvailableNetworks() []string {
 	var networks []string
-	for name := range NetworksByName {
-		networks = append(networks, name)
+	for _, cfg := range superchain.OPChains {
```
trianglesphere, line 45: Is `OPChains` a strict superset of `{op-mainnet, op-goerli, op-sepolia}`?

op-node/chaincfg/chains.go
```diff
@@ -2,135 +2,100 @@ package chaincfg
...
+}
 
-	return network, nil
+// ChainByName returns a chain, from known available configurations, by name.
+// ChainByName returns nil when the chain name is unknown.
+func ChainByName(name string) *superchain.ChainConfig {
+	// Handle legacy name aliases
+	name = handleLegacyName(name)
+	for _, chainCfg := range superchain.OPChains {
+		if strings.EqualFold(chainCfg.Chain+"-"+chainCfg.Superchain, name) {
```
trianglesphere, line 84: What is `chainCfg.Chain` & `chainCfg.Superchain`?

op-node/rollup/superchain.go
```diff
@@ -0,0 +1,96 @@
+package rollup
```
trianglesphere, line 1: Why is this in the rollup package & not chain-config?

op-program/chainconfig/chaincfg.go
```diff
@@ -2,123 +2,40 @@ package chainconfig
...
-	},
+func init() {
+	mustLoadConfig := func(chainID uint64) *params.ChainConfig {
+		cfg, err := params.LoadOPStackChainConfig(chainID)
+		if err != nil {
+			panic(err)
+		}
+		return cfg
+	}
+	OPGoerliChainConfig = mustLoadConfig(420)
```
trianglesphere, line 20: can use use the constants for Goerli/Sepolia/Mainnet chain IDs here?

trianglesphere (commented):
The chains test does give me high confidence that this isn't breaking anything & I verified that the chain configs in the test are the same as the chain configs on develop.

I'm having a hard time reviewing this because so much is pulled in from the superchain-registry repo & there is a lot of magic in terms of how stuff is expected to be rather than simple data.

op-program/chainconfig/chaincfg.go
protolambda: The intention is to remove those global configs once the superchain-registry is stable/approved. So I don't want the globals in the critical path. The test will ensure that the registry behaves as expected for these important configs.

op-node/rollup/superchain.go
protolambda: Either place would work. I think I can move it. It's one of the primary places to instantiate rollup configs, so hence the original choice to put it next to the rollup config type.

op-node/chaincfg/chains.go
protolambda: `Chain` is the simple lowercase identifier of the chain (without superchain part, e.g. `op`, `base`, `zora`, etc.) and `chainCfg.Superchain` is the superchain identifier (`mainnet`, `sepolia`, `goerli`, etc.). And there's `chainCfg.Name` for the prettified display name (e.g. `OP-Mainnet`).

op-node/chaincfg/chains.go
protolambda: Yes, `superchain.OPChains` contains all chains that are in the superchain-registry, which includes some others than the OP chains.

op-node/rollup/superchain.go
trianglesphere: hmm, I see that `chaincfg` depends on `rollup/types.go`. Package organization wise, I think it does make more sense here. My two small concerns are around the rollup package getting large and codeowners, but I don't think that that's enough to move them back.

op-node/chaincfg/chains.go
trianglesphere: One callout here is that I wasn't able to figure this out by reading the chainconfig comments. https://github.com/ethereum-optimism/superchain-registry/blob/5d3ca1966435827ac8c4216bddcb541ec64c1d6c/superchain/superchain.go#L33-L51

trianglesphere (approved):
LGTM

PR #6584, opened 2023-08-03
contracts-bedrock: cleanup deploy script

tynes:
**Description**

Use a standard modifier to skip deploying contracts to non devnet networks. Simplifies the code, removes branching, ensures that the contracts are always deployed as expected.

### DIFF
--- a/packages/contracts-bedrock/scripts/Deploy.s.sol
+++ b/packages/contracts-bedrock/scripts/Deploy.s.sol
@@ -112,7 +112,7 @@ contract Deploy is Deployer {
     /// @notice Modifier that will only allow a function to be called on devnet.
     modifier onlyDevnet() {
         uint256 chainid = block.chainid;
-        if (chainid != Chains.LocalDevnet && chainid != Chains.GethDevnet) {
+        if (chainid == Chains.LocalDevnet || chainid == Chains.GethDevnet) {
             _;
         }
     }
--- a/packages/contracts-bedrock/scripts/Deploy.s.sol
+++ b/packages/contracts-bedrock/scripts/Deploy.s.sol
@@ -110,9 +110,9 @@ contract Deploy is Deployer {
     }
 
     /// @notice Modifier that will only allow a function to be called on devnet.
-    modifier onlyDevnet() internal view returns (bool) {
+    modifier onlyDevnet() {
         uint256 chainid = block.chainid;
-        if chaind != Chains.LocalDevnet && chainid != Chains.GethDevnet {
+        if (chainid != Chains.LocalDevnet && chainid != Chains.GethDevnet) {
             _;
         }
     }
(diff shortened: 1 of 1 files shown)


packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -109,6 +109,14 @@ contract Deploy is Deployer {
         vm.stopBroadcast();
     }
 
+    /// @notice Modifier that will only allow a function to be called on devnet.
+    modifier onlyDevnet() {
+        uint256 chainid = block.chainid;
+        if (chainid != Chains.LocalDevnet && chainid != Chains.GethDevnet) {
```
Inphi: Shouldn't this be `chainid == Chains.LocalDevnet || chainid == Chains.GethDevnet`?

also, btw, is `chains.OPLocalDevnet` still useful?

tynes: Good catch, can confirm from the logs

```
Syncing deployment AddressManager: contract AddressManager
  Syncing deployment L1CrossDomainMessenger: contract L1CrossDomainMessenger
  Syncing deployment L1CrossDomainMessengerProxy: contract ResolvedDelegateProxy
  Syncing deployment L1ERC721Bridge: contract L1ERC721Bridge
  Syncing deployment L1ERC721BridgeProxy: contract Proxy
  Syncing deployment L1StandardBridge: contract L1StandardBridge
  Syncing deployment L1StandardBridgeProxy: contract L1ChugSplashProxy
  Syncing deployment L2OutputOracle: contract L2OutputOracle
  Syncing deployment L2OutputOracleProxy: contract Proxy
  Syncing deployment OptimismMintableERC20Factory: contract OptimismMintableERC20Factory
  Syncing deployment OptimismMintableERC20FactoryProxy: contract Proxy
  Syncing deployment OptimismPortal: contract OptimismPortal
  Syncing deployment OptimismPortalProxy: contract Proxy
  Syncing deployment ProxyAdmin: contract ProxyAdmin
  Syncing deployment SystemConfig: contract SystemConfig
  Syncing deployment SystemConfigProxy: contract Prox
```

tynes: > also, btw, is chains.OPLocalDevnet still useful?

It is useful if you manually deploy against the local devnet. The docker compose setup uses `geth --dev` to deploy against which uses 1337 as its chainid and it cannot be changed without being a bit more invasive. Without both chainids in here, its all or nothing with the docker setup, either bring the whole thing down and bring the whole thing back up

If anvil supported `debug_dumpBlock` then we could remove `geth` as a dep and just start anvil with chainid 900 and then dump the state and always use the same chainid

Inphi (approved):

clabby (approved):
Looks great :)

PR #6593, opened 2023-08-04
feat(ctb): Deploy `FaultDisputeGame` with `MIPS` backend on devnet

clabby:
## Overview

Adds a second `DisputeGameFactory` deployment to the devnet for a separate implementation of the `FaultDisputeGame` with a `MIPS` backend.

**Metadata**
Closes CLI-4326


clabby:
* develop
  * **PR #6593** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6593" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6593?utm_source=stack-comment).

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -710,30 +724,37 @@ contract Deploy is Deployer {
...
-                _vm: faultVm,
-                _l2oo: L2OutputOracle(mustGetAddress("L2OutputOracleProxy")),
-                _blockOracle: BlockOracle(mustGetAddress("BlockOracle"))
-            }));
-            console.log("DisputeGameFactory: set `FaultDisputeGame` implementation");
+        string[2] memory contractNames = ["DisputeGameFactoryProxy", "MIPSDisputeGameFactoryProxy"];
+
+        for (uint256 i; i < contractNames.length; i++) {
+            DisputeGameFactory factory = DisputeGameFactory(mustGetAddress(contractNames[i]));
+            Claim absolutePrestate = Claim.wrap(bytes32(cfg.faultGameAbsolutePrestate()));
```
ajsutton: Not necessarily a blocker for this PR, but we will need to have different absolute pre-states for the alphabet and cannon games.

clabby: Was just about to ping slack - do we know the absolute prestate for Cannon at the moment?

refcell (approved):
LGTM

clabby:
Waiting to merge until @ajsutton adds an absolute prestate generation script that we can consume for the `FaultDisputeGame` deployment on the devnet.

ajsutton:
https://github.com/ethereum-optimism/optimism/pull/6594 should do the trick.  make cannon-prestate in the top directory and contract deployment will want op-program/bin/prestate-proof.json and the pre hash from there I think (the post is after trace step 0 executed).

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -262,17 +262,21 @@ contract Deploy is Deployer {
     /// @notice Deploy the DisputeGameFactoryProxy
     function deployDisputeGameFactoryProxy() onlyDevnet broadcast() public returns (address) {
         address proxyAdmin = mustGetAddress("ProxyAdmin");
-        Proxy proxy = new Proxy({
-            _admin: proxyAdmin
-        });
```
tynes: Do we need both of these in the long term?

clabby: Not necessarily, if we'd like to no longer test the alphabet game, we can remove it. We can also have different game type mappings for devnet - this was just to make it easy with `GameType.FAULT = 0`

ajsutton: I would say we want to keep alphabet and cannon long term.  It is _way_ cheaper to test the game and challenger with alphabet than with cannon. Plus it's much easier to create understandable differences in an alphabet trace than a cannon trace.

Inphi (approved):
LGTM!
If we're keeping the alphabet game, then let's rename the deploy configs to unambiguously refer to a game type. For example, the `faultGameMaxDepth` should be two configs; one for alphabet and MIPS.

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -262,17 +262,21 @@ contract Deploy is Deployer {
...
+            Proxy proxy = new Proxy({
+                _admin: proxyAdmin
+            });
 
-        save("DisputeGameFactoryProxy", address(proxy));
-        console.log("DisputeGameFactoryProxy deployed at %s", address(proxy));
+            address admin = address(uint160(uint256(vm.load(address(proxy), OWNER_KEY))));
+            require(admin == proxyAdmin);
 
-        return address(proxy);
```
tynes: These functions have a baked in assumption that they deploy a single contract based on the return signature, it will only impact `forge script --sig 'deployDisputeGameFactoryProxy()' | jq .`

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -710,30 +724,51 @@ contract Deploy is Deployer {
...
-                _blockOracle: BlockOracle(mustGetAddress("BlockOracle"))
-            }));
-            console.log("DisputeGameFactory: set `FaultDisputeGame` implementation");
+        // Create the absolute prestate dump
+        string memory filePath = string.concat(vm.projectRoot(), "/../../op-program/bin/prestate-proof.json");
+        bytes32 mipsAbsolutePrestate;
+        string[] memory commands = new string[](3);
+        commands[0] = "bash";
+        commands[1] = "-c";
+        commands[2] = "[[ -f ../../op-program/bin/prestate-proof.json ]] && echo \"present\"";
```
tynes: not using `filePath` variable?

tynes (approved):

PR #6597, opened 2023-08-04
feat(ctb): Proper contract ABI for MIPS.sol

Inphi:
**Description**

This is necessary so the `FaultDisputeGame` can easily interact with the MIPS VM. `MIPS.step` previously required a non-standard calldata format where the packed state and proofs did not contain any padding. As such, interacting with the VM contract in Solidity would require careful and expensive construction of calldata.

With this patch, the FDG can interface with MIPS.sol in the usual way. The increased calldata cost (up to 31 bytes) of adhering to the ABI is negligible compared to the proof size.

**Tests**

Added a simple test asserting MIPS ABI.


refcell (approved):
sweet

PR #6607, opened 2023-08-05
feat(indexer): Index OptimismPortal & L2ToL1MP Txs

hamdiallam:
The standard bridge table marks the `cross_domain_messenger_nonce` as optional to allow entries to be created for ETH sends directly to the OptimismPortal/L2ToL1MP contracts. This allows our bridge API to serve both cases for bridging ETH.

Changes:
- Process optimism portal and l2tol2mp transactions
- Reuse the `withdrawal_hash` and the `source_hash` as primary keys for the StandardBridge deposits table
- E2E tests

Part of ethereum-optimism/devx#26


hamdiallam:
* develop
  * **PR #6615** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6615" target="_blank">Graphite</a> 
    * **PR #6607** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6607" target="_blank">Graphite</a>  👈
      * **PR #6608** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6608" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6607?utm_source=stack-comment).

indexer/processor/encoding.go
```diff
@@ -0,0 +1,24 @@
+package processor
+
+import (
+	"encoding/binary"
+	"math/big"
+)
+
+// DecodeVersionNonce is an re-implementation of Encoding.sol#decodeVersionedNonce.
+// If the nonce is greater than 32 bytes (solidity uint256), bytes [32:] are ignored
```
roninjin10, line 9: Not sure how hard the test is to write and don't feel super strongly but a test that passes inputs into both this and the solidity version and asserts the same output would be nice

indexer/database/bridge_transactions.go
```diff
@@ -0,0 +1,152 @@
...
+	return result.Error
+}
+
+func (db *bridgeTransactionsDB) L2TransactionWithdrawal(withdrawalHash common.Hash) (*L2TransactionWithdrawal, error) {
+	var withdrawal L2TransactionWithdrawal
+	result := db.gorm.Where(&L2TransactionWithdrawal{WithdrawalHash: withdrawalHash}).Take(&withdrawal)
+	if result.Error != nil {
+		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+			return nil, nil
+		}
```
roninjin10, line 117:
```suggestion
    if errors.Is(result.Error, gorm.ErrRecordNotFound) {
			return nil, nil
	}
	if result.Error != nil {
```
Nittiest of all nits: Don't waste time changing this and not even sure if the compiler will allow this but I always prefer this way of writing stuff

indexer/database/bridge_transactions.go
```diff
@@ -0,0 +1,152 @@
...
+		return err
+	} else if withdrawal == nil {
+		return fmt.Errorf("transaction withdrawal hash %s not found", withdrawalHash)
+	} else if withdrawal.ProvenL1EventGUID == nil {
+		return fmt.Errorf("cannot mark unproven withdrawal hash %s as finalized", withdrawal.WithdrawalHash)
+	}
+
+	withdrawal.FinalizedL1EventGUID = &finalizedL1EventGuid
+	result := db.gorm.Save(&withdrawal)
+	return result.Error
```
roninjin10: Nittiest of all nits:
DOn't waste time on style nits but feels like we are mixing an else if pattern with a don't use else pattern.  e.g. could both move the last 3 lines into an else block or alternatively replace the else if with just if blocks

indexer/e2e_tests/bridge_transactions_e2e_test.go
```diff
@@ -0,0 +1,170 @@
+package e2e_tests
```
roninjin10, line 1: wow these tests are great and comprehensive great job!

roninjin10 (commented):
pr makes sense and lgtm looks like tests still failing though

indexer/database/bridge_transactions.go
hamdiallam: I dont think the compiler would complain about this. Happy to revert to using the suggestion but we should make the change in a separate PR since there are many places this pattern exists in the existing models

roninjin10: Nah not a good use of time IMO

indexer/database/bridge_transactions.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/**
+ * Transactions deposited from L1
+ */
+
+func (db *bridgeTransactionsDB) StoreL1TransactionDeposits(deposits []*L1TransactionDeposit) error {
+	result := db.gorm.Create(&deposits)
+	return result.Error
+}
```
roninjin10, line 87: Curious if idiomatic go code ever returns

```go
if result.Error == nil {
  return result.Error, nil
}
return &deposits, nil
```

indexer/database/bridge_transactions.go
```diff
@@ -0,0 +1,152 @@
...
+			return nil, nil
+		}
+		return nil, result.Error
+	}
+
+	return &withdrawal, nil
+}
+
+// MarkL2TransactionWithdrawalProvenEvent links a withdrawn transaction with associated Prove action on L1.
+func (db *bridgeTransactionsDB) MarkL2TransactionWithdrawalProvenEvent(withdrawalHash common.Hash, provenL1EventGuid uuid.UUID) error {
```
roninjin10, line 125: Would it make sense in the future to offer a way of kinda opening a websocket to get notified when a specific withdrawal is proven or claimed? It's common to want to wait on this in UIs and usually we poll

indexer/processor/optimism_portal.go
```diff
@@ -24,14 +37,46 @@ type OptimismPortalProvenWithdrawal struct {
...
+		return nil, err
+	}
+
+	eventName := "TransactionDeposited"
+	if optimismPortalAbi.Events[eventName].ID != derive.DepositEventABIHash {
+		return nil, errors.New("op-node deposit event abi hash & optimism portal tx deposit mismatch")
+	}
+
+	processedTxDepositedEvents := events.eventsBySignature[derive.DepositEventABIHash]
+	txDeposits := make([]OptimismPortalTransactionDepositEvent, len(processedTxDepositedEvents))
```
roninjin10, line 52: noob question: what does make do?

roninjin10 (approved):
lgtm have some questions about the processor but they are kinda unrelated to this pr and easier to review the processor in the repo than this pr

indexer/processor/optimism_portal.go
hamdiallam: if you know the size ahead of time, make will allocate a contiguous amount of memory which makes things like lookups and general operations faster.

Without it, the runtime may have to reallocate and copy arrays when appending new entries

indexer/processor/encoding.go
hamdiallam: yea unfortunately, the bindings dont expose an eth_call to the Encoding.sol smart contract. I made sure the E2E test decodes the nonce and checks it with I think is good enough

indexer/e2e_tests/bridge_transactions_e2e_test.go
hamdiallam: thanks!

indexer/database/bridge_transactions.go
hamdiallam: we dont return the deposits here, just the error

Assuming we return deposits `return result.Error, nil` would fail since `result.Error` is typed as an error versus a slice

indexer/database/bridge_transactions.go
hamdiallam: yea i think we can add these kind of features in the API server

indexer/api/api_test.go
```diff
@@ -8,32 +8,25 @@ import (
...
 // MockBridgeTransfersView mocks the BridgeTransfersView interface
 type MockBridgeTransfersView struct{}
 
-const (
-	guid1 = "8408b6d2-7c90-4cfc-8604-b2204116cb6a"
-	guid2 = "8408b6d2-7c90-4cfc-8604-b2204116cb6b"
-)
-
 var (
 	deposit = database.L1BridgeDeposit{
```
hamdiallam, line 18: opening comment so mergebot doesn't merge

PR #6616, opened 2023-08-06
cannon: Add subcommand to convert from json state to binary witness

ajsutton:
**Description**

Adds a `witness` subcommand to cannon that converts from a JSON state to the binary witness. The witness hash is written to stdout and the pre-image optionally written to a file.


refcell (commented):
LGTM but probably should await @protolambda's review

cannon/cmd/witness.go
```diff
@@ -0,0 +1,53 @@
...
+var (
+	WitnessInputFlag = &cli.PathFlag{
+		Name:      "input",
+		Usage:     "path of input JSON state.",
+		TakesFile: true,
+		Required:  true,
+	}
+	WitnessOutputFlag = &cli.PathFlag{
+		Name:      "output",
+		Usage:     "path to write binary witness.",
```
protolambda, line 21: A binary witness is fine with me, but is something like a json-wrapped hex string not more readable?

protolambda (approved):
Code looks good, but the output-format may be something you may want to change, depending on the usage with the challenger or other tools.

cannon/cmd/witness.go
ajsutton: My thinking is that we want the exact binary output since we're going to hash it and for the go side this is definitely easiest since we can just read the file and use the result directly.  I assume we can just read a binary file as part of the contract deployments scripts but its possible it will come unstuck there and we'll have to add a json wrapper.

PR #6618, opened 2023-08-07
feat(ctb): `MIPS` constructor

clabby:
## Overview

Adds a constructor to the `MIPS` contract that initializes the `PreimageOracle` address. This allows for the `MIPS` contract to communicate with the `PreimageOracle` to read local data and global keccak preimages.

**Metadata**
Closes CLI-4329


clabby:
* develop
  * **PR #6618** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6618" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/monorepo/6618?utm_source=stack-comment).

packages/contracts-bedrock/src/cannon/MIPS.sol
```diff
@@ -53,9 +53,14 @@ contract MIPS {
     uint32 constant EBADF = 0x9;
     uint32 constant EINVAL = 0x16;
 
-    /// @notice The pre-image oracle.
+    /// @notice The preimage oracle contract.
```
Inphi, line 56: nit: We can now make this immutable. Doing this will require changes to the cannon evm tests so it doesn't have to be done on this PR.

Inphi (approved):

packages/contracts-bedrock/src/cannon/MIPS.sol
clabby: Yeah this is why I didn't in this PR: https://github.com/ethereum-optimism/optimism/blob/4ef51e13805ac8ca5cc58b014552df6066bee453/cannon/mipsevm/evm.go#L115-L117

If we could change that up, would be great :)

clabby: p.s. @tynes recently changed the style guide for immutables so that our ABIs don't leak information about the variable, if you do pick this up, we'll need to make the immutable internal and add an external getter that maintains the external ABI as-is (`function oracle() external view returns (IPreimageOracle oracle_)`).

Inphi: Thanks. Ill keep that in mind when implementing the feat.

tynes: Will every network use the same preimage oracle and mips contract?

Inphi: @tynes Yes. The cannon contracts don't depend on the network/chain.

PR #6619, opened 2023-08-07
Web3.js Plugin for Fee Estimation

Spacesai1or:
**Description**

Adding a web3.js plugin that offers the same API as the [@eth-optimism/fee-estimation](https://github.com/ethereum-optimism/optimism/tree/develop/packages/fee-estimation) package

**Tests**

I've added tests for each user facing method, but they mostly compare the returned value to the expected type of the value, but not the value its self e.g.

```typescript
expect(typeof (await web3.op.getBaseFee())).toBe('bigint'))
```

Unlike the tests for the `@eth-optimism/fee-estimation` package, most of these RPC calls cannot be made against the state of the chain at a specific block (because web3.js doesn't support this unlike Viem), so the return values are not deterministic. The plugin methods, `estimateFees` and `getL2Fee` should support specifying a block, but in my testing specifying the block didn't return the same value for each request - I have not looked into this further to determine why this doesn't work, but I'm assuming it's a bug in web3.js

**Additional context**

There's an on-going discussion between @roninjin10 and I about better testing for fee estimation, including testing submitting a transaction to Goerli to verify fee estimation is returning a sensible value. These enhanced tests are likely to come in a later PR

**Metadata**

- [This](https://github.com/ethereum-optimism/devx/issues/19#issuecomment-1652591932) comment


packages/web3.js-plugin-fee-estimation/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "types": "src/plugin.ts",
+  "files": [
+    "dist/",
+    "src/"
+  ],
+  "scripts": {
+    "build": "tsup",
+    "lint": "prettier --check .",
+    "lint:fix": "prettier --write .",
+    "test": "vitest",
```
roninjin10:
```suggestion
    "test": "vitest --coverage",
```
Not opinionated here but I started recently doing this. Up to you if you prefer to not show or show coverage

packages/web3.js-plugin-fee-estimation/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ethereum-optimism/optimism.git",
+    "directory": "packages/web3js-plugin-fee-estimation"
+  },
+  "homepage": "https://optimism.io",
+  "type": "module",
+  "main": "dist/plugin.js",
+  "module": "dist/plugin.mjs",
+  "types": "src/plugin.ts",
```
roninjin10: I'm the one you copied this shortcut from and I think it's fine to start but worth noting this generally should be building d.ts files rather than pointing at TS. It's fine to start here though just calling out that this is generally a shortcut and not a good pattern

packages/web3.js-plugin-fee-estimation/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../../tsconfig.json",
```
roninjin10: Unrelated to PR. We should move the tsconfig.json to a package sometime. It's a tiny antipattern to require files outside of a package without depending on them. I haven't gotten around to doing this yet though because it's super super low priority though

packages/web3.js-plugin-fee-estimation/src/plugin.ts
```diff
@@ -0,0 +1,337 @@
...
+        // with empty uint8Array. Because L1 fee calculation
+        // is dependent on the number of bytes, we are removing
+        // the zero values bytes
+        RLP.encode(ethereumjsTransaction.raw().slice(0, -3))
+      )
+    )
+  }
+}
+
+// Module Augmentation to add op namespace to root {web3} instance
```
roninjin10: Cool pattern

packages/web3.js-plugin-fee-estimation/.prettierrc.cjs
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ...require('../../.prettierrc.js'),
```
roninjin10: Unrelated to PR. We should move the prettierrc to a package sometime. It's a tiny antipattern to require files outside of a package without depending on them. I haven't gotten around to doing this yet though because it's super super low priority though

packages/web3.js-plugin-fee-estimation/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# @eth-optimism/web3.js-plugin-fee-estimation
```
roninjin10: Delete this file for now changesets will generate it when it needs to

packages/web3.js-plugin-fee-estimation/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 Optimism
```
roninjin10:
```suggestion
Copyright (c) 2023 Optimism
```

packages/web3.js-plugin-fee-estimation/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@eth-optimism/web3.js-plugin-fee-estimation",
+  "version": "0.1.0",
+  "description": "A Web3.js plugin for doing OP-Chain gas estimation",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ethereum-optimism/optimism.git",
+    "directory": "packages/web3js-plugin-fee-estimation"
```
roninjin10:
```suggestion
    "directory": "packages/web3.js-plugin-fee-estimation"
```
Missing a '.'

FWIW I would slightly prefer removing the '.' from the file path but that's just a nit up to you

packages/web3.js-plugin-fee-estimation/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "compilerOptions": {
+    "outDir": "./dist",
+    "baseUrl": "./src",
+    "strict": true,
```
roninjin10:
```suggestion
    "strict": true,
    "verbatimModuleSyntax": true,
    "noUncheckedIndexedAccess": true,
```
VerbatimModuleSyntax forces us to add `type` to imports
noUncheckedIndexAccess makes ts scrictor when it comes to index types.

These are the 2 most important missing config options but more optional strict rules can be found in this file https://github.com/evmts/evmts-monorepo/blob/main/tsconfig/base.json

packages/web3.js-plugin-fee-estimation/src/plugin.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+import {
+  l2StandardBridgeABI,
+  l2StandardBridgeAddress,
+  optimistABI,
+  optimistAddress,
+} from '@eth-optimism/contracts-ts'
+
+import OptimismFeeEstimationPlugin from './plugin'
+
+const provider = process.env.VITE_L2_RPC_URL ?? 'https://mainnet.optimism.io'
```
roninjin10:
```suggestion
import {z} from 'zod'
const defaultProvider = 'https://mainnet.optimism.io'
const provider = z.string().url(process.env.VITE_L2_RPC_URL).default(defaultProvider)
if (provider === defaultProvider) console.warn('warning: using default public provider. Tests may face rate limits. To pass in a different provider set VITE_L2_RPC_URL in env')
```
Nit would prefer strict validation like this

packages/web3.js-plugin-fee-estimation/src/plugin.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+describe('OptimismFeeEstimationPlugin', () => {
+  let web3: Web3
+
+  beforeAll(() => {
+    web3 = new Web3(provider)
+    web3.registerPlugin(new OptimismFeeEstimationPlugin())
+  })
+
+  test('should be registered under .op namespace', () =>
+    expect(web3.op).toBeDefined())
```
roninjin10:
```suggestion
    expect(web3.op).toMatchInlineSnapshot())
```
Nit: maybe do this so we can at a glance see every method registered on web3.op for anyone using this test as documentation

packages/web3.js-plugin-fee-estimation/src/plugin.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  describe('should return a bigint by default', () => {
+    test('getBaseFee', async () =>
+      expect(typeof (await web3.op.getBaseFee())).toBe('bigint'))
```
roninjin10:
```suggestion
      expect(await web3.op.getBaseFee()).toBeTypeOf('bigint'))
```
Nit: this is more idiomatic

packages/web3.js-plugin-fee-estimation/src/plugin.ts
```diff
@@ -0,0 +1,337 @@
...
+   * @returns {Promise<bigint>} - The L2 base fee as a BigInt by default, but {returnFormat} determines type
+   * @example
+   * const baseFeeValue: bigint = await web3.op.getBaseFee();
+   * @example
+   * const numberFormat = { number: FMT_NUMBER.NUMBER, bytes: FMT_BYTES.HEX }
+   * const baseFeeValue: number = await web3.op.getBaseFee(numberFormat);
+   */
+  public async getBaseFee<
+    ReturnFormat extends DataFormat = typeof DEFAULT_RETURN_FORMAT
+  >(returnFormat?: ReturnFormat) {
```
roninjin10: nit: for libraries prefer being explicit about the return type. This helps guarantee we don't accidentally make a breaking change.

roninjin10:
```suggestion
  >(returnFormat?: ReturnFormat): Promise<bigint> {
```

packages/web3.js-plugin-fee-estimation/src/plugin.ts
```diff
@@ -0,0 +1,337 @@
...
+   * @returns {Promise<Numbers>} - The number of decimals as a BigInt by default, but {returnFormat} determines type
+   * @example
+   * const decimalsValue: bigint = await web3.op.getDecimals();
+   * @example
+   * const numberFormat = { number: FMT_NUMBER.NUMBER, bytes: FMT_BYTES.HEX }
+   * const decimalsValue: number = await web3.op.getDecimals(numberFormat);
+   */
+  public async getDecimals<
+    ReturnFormat extends DataFormat = typeof DEFAULT_RETURN_FORMAT
+  >(returnFormat?: ReturnFormat) {
```
roninjin10:
```suggestion
  >(returnFormat?: ReturnFormat): Promise<bigint> {
```

packages/web3.js-plugin-fee-estimation/src/plugin.ts
```diff
@@ -0,0 +1,337 @@
...
+    ReturnFormat extends DataFormat = typeof DEFAULT_RETURN_FORMAT
+  >(
+    transaction: Transaction,
+    options?: {
+      blockNumber?: BlockNumberOrTag
+      returnFormat?: ReturnFormat
+    }
+  ) {
+    const [l1Fee, l2Fee] = await Promise.all([
+      this.getL1Fee(transaction),
```
roninjin10: Should this be passing in the options.blockNumber too?

packages/web3.js-plugin-fee-estimation/src/plugin.ts
roninjin10: Ah this is actually pretty complicated though since the return type depends on the return format. What do you think about being more modular and only returning the result of the call and expecting the user to call Web3.utils.format themselves? Simplifies the API, simplifies the type infererence, more modular, more of a 1 to 1 match with the underlying api.

Alternatively force BigInt

roninjin10: Actually nvm I see that estimateGas from web3.js does teh same pattern. We should match what web3.js does 1to1 so ignore the above suggestion. And the return type is complicated enough where we are likely best off just inferring it rather than being explicit

roninjin10 (changes requested):
What do you think about just naming this package `web3-plugin` so we can add additional functionality in future. for example, we could later add a method that lets you send a l2 tx from l1.

Great stuff! Most comments are nits

packages/web3.js-plugin-fee-estimation/src/plugin.ts
```diff
@@ -0,0 +1,337 @@
...
+} from 'web3'
+import { TransactionFactory, TxData } from 'web3-eth-accounts'
+import { estimateGas, formatTransaction } from 'web3-eth'
+import {
+  gasPriceOracleABI,
+  gasPriceOracleAddress,
+} from '@eth-optimism/contracts-ts'
+import { RLP } from '@ethereumjs/rlp'
+
+export default class OptimismFeeEstimationPlugin extends Web3PluginBase {
```
roninjin10:
```suggestion
export class OptimismFeeEstimationPlugin extends Web3PluginBase {
```
Let's not export as default so later we can export more stuff without it being a breaking change

packages/web3.js-plugin-fee-estimation/src/plugin.spec.ts
```diff
@@ -0,0 +1,241 @@
+import { beforeAll, describe, expect, test } from 'vitest'
+import Web3, { Contract, FMT_BYTES, FMT_NUMBER } from 'web3'
+import {
+  l2StandardBridgeABI,
+  l2StandardBridgeAddress,
+  optimistABI,
+  optimistAddress,
+} from '@eth-optimism/contracts-ts'
+
+import OptimismFeeEstimationPlugin from './plugin'
```
roninjin10:
```suggestion
import {OptimismFeeEstimationPlugin} from './plugin'
```
Let's not import it as a default import so later if we add more plugins to this package we don't need to make a breaking change

packages/web3.js-plugin-fee-estimation/package.json
```diff
@@ -0,0 +1,45 @@
+{
```
tynes: Do we need a dot in the package name?

tynes:
Def have a preference to call this package name something else that doesnt have a dot in it so its not in the filepath

packages/web3.js-plugin-fee-estimation/package.json
Spacesai1or: It was a preference for web3.js plugins to have the dot, but I think it'll be fine to remove it

packages/web3.js-plugin-fee-estimation/tsconfig.json
Spacesai1or: I agree, I could fix this since I have a limited amount of work on my plate while ramping up

packages/web3.js-plugin-fee-estimation/LICENSE
Spacesai1or: Copied from [fee-estimation](https://github.com/ethereum-optimism/optimism/blob/4ef1634c71b55b4e821750272e3ad49de8eba3d7/packages/fee-estimation/LICENSE#L3) package, do we need to update there as well?

packages/web3.js-plugin-fee-estimation/src/plugin.ts
Spacesai1or: @roninjin10 Ideally yes, but specifying the `blockNumber` for contract calls is not supported by web3.js

packages/web3.js-plugin-fee-estimation/package.json
Spacesai1or: Added `dts: true` to [tsup.config.ts](https://github.com/ethereum-optimism/optimism/pull/6619/files#diff-1e8b4b032cda3302800f90e8377c75a685f1af5eb3cc605850d49944d9429404R13) and updated [package.json](https://github.com/ethereum-optimism/optimism/pull/6619/files#diff-4dd0e96c9dc5cb5cd01b42690052167fb26fef68952dee0fee7a730c0acdcce1R15) to: `"types": "dist/plugin.d.ts",`

packages/web3.js-plugin-fee-estimation/package.json
roninjin10: unrelated, but ethers.js calls their src folder `src.ts` and it bothers me everytime I see it

packages/web3.js-plugin-fee-estimation/package.json
roninjin10: This is fine (especially for a package with 1 file) but you might not know this but this is also an antipattern. You generally want to avoid building types in the same pipeline as building your javascript (or building your tests).  Typescript is such a performance problem at scale that you almost always want your typescript pipeline to be seperate from everything else.

As an example, in the https://github.com/latticexyz/mud repo I made their rebuilds go from 20 seconds to 200ms simply by removing a dts in their file.

Definitely fine here but a thing I try to tell everybody since it isn't common knowledge is to just be aware of this. If you can run your tests and build your javascript completely seperate from type generation and typechecking your builds will always be maximally performant.

As an example my [EVMts repo](https://github.com/evmts/evmts-monorepo/blob/main/package.json#L15) exposes a `build:dist` and `build:types` in every package and builds them seperately. The tests use vitest which uses esbuild under the hood (doesn't typecheck). This keeps everything peformant and the types build in parallel with the only thorn being I haven't optimized next.js example project to do the same thing yet

roninjin10: I always tell folks to think of typescript as being a linter. For apps this works literally but for libraries the metaphor doesn't work as well since you do have to actually output types at the end of the day so it is a build step

packages/web3.js-plugin-fee-estimation/tsconfig.json
roninjin10: Just opened an issue for this today might be worth giving it a week or 2 to see if an outside contributor looking for a first issue wants to pick it up https://github.com/ethereum-optimism/optimism/issues/6629

packages/web3.js-plugin-fee-estimation/LICENSE
roninjin10: yea copy pasta. Funny because @nickbalestra left the same comment I think in that pr too but a force push or bad rebase likely ruined it

packages/web3.js-plugin-fee-estimation/src/plugin.ts
roninjin10: hmm I think that should likely bubble up to making this entire method not support blockNumber since it halfway supporting blockNumber makes this a bit of a leaky abstraction in that folks using it would need to be aware of that nuance

packages/web3.js-plugin-fee-estimation/LICENSE
roninjin10: No need to fix it in this pr though it's fine tbh

packages/web3.js-plugin-fee-estimation/src/plugin.ts
Spacesai1or: Oh yea, that totally makes sense. Will remove it

packages/web3.js-plugin-fee-estimation/package.json
tynes: Apologies, I meant in the package name in the repo, i'm not super opinionated about npm package names, the dot in the directory name is what is throwing me off

roninjin10: yea let's keep the package name consistent with web3.js standards but the folder name throws me off as well

roninjin10: also remove fee-estimation from the foldername

roninjin10 (approved):

felipe-op (approved):

PR #6620, opened 2023-08-07
op-e2e: Use pre-deployed dispute game contracts

ajsutton:
**Description**

Updates the fault proof e2e tests to use the pre-deployed contracts rather than deploying new versions for each test.

Makes the game duration 5 minutes instead of 2 to ensure that e2e tests have long enough to complete.
Fixes the alphabet vm absolute pre-state. It needs to be the hash of the pre-state passed to the constructor. The pre-image is only supplied when calling `step` to attack trace index 0.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4325/use-pre-deployed-dispute-game-contracts-in-e2e


op-chain-ops/genesis/config.go
```diff
@@ -470,6 +470,7 @@ func NewDeployConfigWithNetwork(network, path string) (*DeployConfig, error) {
 // L1Deployments represents a set of L1 contracts that are deployed.
 type L1Deployments struct {
 	AddressManager                    common.Address `json:"AddressManager"`
```
tynes, line 472: This should also have mips and alphabet?

tynes:
The current setup is prone to silent failures which is not ideal

op-chain-ops/genesis/config.go
ajsutton: There's no reason to refer to them directly - they should just be accessed via the dispute game factory and the games it creates.

ajsutton:
> The current setup is prone to silent failures which is not ideal

I'm not sure what you mean by this.

op-chain-ops/genesis/config.go
refcell: Even though they can be considered "L1 Deployments", we should follow our practice of only including items that are used for testability and maintainability. As such, we shouldn't need to include MIPS and Alphabet themselves

refcell (approved):
Will clean up 1 nit in a follow-on pr to prevent blockage.

tynes:
> > The current setup is prone to silent failures which is not ideal
> 
> I'm not sure what you mean by this

If building the genesis fails, then CI continues and the allocs file and addresses file dont exist. Recently @clabby added the cannon prestate as a dep to the deployment script, which if that build fails then it's not immediately obvious what failed because it causes the deploy script to fail but CI continues, that's what I meant by silent failure. @protolambda noticed this

ajsutton:
> If building the genesis fails, then CI continues and the allocs file and addresses file dont exist.

Ah ok, that should be resolvable - just need to ensure the genesis build exits with a non-zero status if it fails.

PR #6621, opened 2023-08-07
op-e2e: Add initial e2e test for cannon dispute game

ajsutton:
**Description**

Builds on #6620 and adds an additional e2e test that uses cannon in dispute games.  The test is currently skipped as op-challenger doesn't yet handle trace extension correctly so fails to generate the required proof file. 

With https://github.com/ethereum-optimism/optimism/pull/6622 it passes if you tell op-challenger that cannon games have a max depth of 4.

Refactors the dispute game helper to split into separate files and have separate game helpers for alphabet and cannon games.

Ultimately will want to extend the test to make further attack/defend steps manually and confirm the challenger continues to counter them all the way through to performing an actual `step` call. Worth merging this now for the refactoring and to make it easier for people to test the challenger running against cannon dispute games.

**Tests**

All about the tests.

**Metadata**

### DIFF
--- a/op-e2e/e2eutils/disputegame/alphabet_helper.go
+++ b/op-e2e/e2eutils/disputegame/alphabet_helper.go
@@ -1,0 +1,33 @@
+package disputegame
+
+import (
+	"context"
+
+	"github.com/ethereum-optimism/optimism/op-challenger/config"
+	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils/challenger"
+)
+
+type AlphabetGameHelper struct {
+	FaultGameHelper
+	claimedAlphabet string
+}
+
+func (g *AlphabetGameHelper) StartChallenger(ctx context.Context, l1Endpoint string, name string, options ...challenger.Option) *challenger.Helper {
+	opts := []challenger.Option{
+		func(c *config.Config) {
+			c.GameAddress = g.addr
+			c.GameDepth = alphabetGameDepth
+			c.TraceType = config.TraceTypeAlphabet
+			// By default the challenger agrees with the root claim (thus disagrees with the proposed output)
+			// This can be overridden by passing in options
+			c.AlphabetTrace = g.claimedAlphabet
+			c.AgreeWithProposedOutput = false
+		},
+	}
+	opts = append(opts, options...)
+	c := challenger.NewChallenger(g.t, ctx, l1Endpoint, name, opts...)
+	g.t.Cleanup(func() {
+		_ = c.Close()
+	})
+	return c
+}
--- a/op-e2e/e2eutils/disputegame/cannon_helper.go
+++ b/op-e2e/e2eutils/disputegame/cannon_helper.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/ethereum-optimism/optimism/op-challenger/config"
 	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils/challenger"
-	"github.com/ethereum/go-ethereum/common"
 )
 
 type CannonGameHelper struct {
@@ -24,7 +23,6 @@ func (g *CannonGameHelper) StartChallenger(ctx context.Context, l1Endpoint strin
 			c.CannonDatadir = g.t.TempDir()
 			c.CannonServer = "../op-program/bin/op-program"
 			c.CannonAbsolutePreState = "../op-program/bin/prestate.json"
-			c.PreimageOracleAddress = common.Address{0xbb}
 			c.CannonSnapshotFreq = config.DefaultCannonSnapshotFreq
 		},
 	}
--- a/op-e2e/e2eutils/disputegame/cannon_helper.go
+++ b/op-e2e/e2eutils/disputegame/cannon_helper.go
@@ -1,0 +1,38 @@
+package disputegame
+
+import (
+	"context"
+
+	"github.com/ethereum-optimism/optimism/op-challenger/config"
+	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils/challenger"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type CannonGameHelper struct {
+	FaultGameHelper
+}
+
+func (g *CannonGameHelper) StartChallenger(ctx context.Context, l1Endpoint string, l2Endpoint string, name string, options ...challenger.Option) *challenger.Helper {
+	opts := []challenger.Option{
+		func(c *config.Config) {
+			c.GameAddress = g.addr
+			c.GameDepth = cannonGameDepth
+			c.TraceType = config.TraceTypeCannon
+			// By default, agree with the proposed output root
+			c.AgreeWithProposedOutput = true
+			c.CannonL2 = l2Endpoint
+			c.CannonBin = "../cannon/bin/cannon"
+			c.CannonDatadir = g.t.TempDir()
+			c.CannonServer = "../op-program/bin/op-program"
+			c.CannonAbsolutePreState = "../op-program/bin/prestate.json"
+			c.PreimageOracleAddress = common.Address{0xbb}
+			c.CannonSnapshotFreq = config.DefaultCannonSnapshotFreq
+		},
+	}
+	opts = append(opts, options...)
+	c := challenger.NewChallenger(g.t, ctx, l1Endpoint, name, opts...)
+	g.t.Cleanup(func() {
+		_ = c.Close()
+	})
+	return c
+}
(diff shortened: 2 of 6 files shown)


op-e2e/e2eutils/disputegame/alphabet_helper.go
```diff
@@ -0,0 +1,33 @@
...
+func (g *AlphabetGameHelper) StartChallenger(ctx context.Context, l1Endpoint string, name string, options ...challenger.Option) *challenger.Helper {
+	opts := []challenger.Option{
+		func(c *config.Config) {
+			c.GameAddress = g.addr
+			c.GameDepth = alphabetGameDepth
+			c.TraceType = config.TraceTypeAlphabet
+			// By default the challenger agrees with the root claim (thus disagrees with the proposed output)
+			// This can be overridden by passing in options
+			c.AlphabetTrace = g.claimedAlphabet
+			c.AgreeWithProposedOutput = false
```
refcell, line 24: This is the opposite of what is done in the parent `NewChallenger` function inside `op-e2e/e2eutils/challenger/helper.go` - i.e. `AgreeWithProposedOutput = true`.
I think it would make more sense to be consistent here.

op-e2e/e2eutils/disputegame/cannon_helper.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (g *CannonGameHelper) StartChallenger(ctx context.Context, l1Endpoint string, l2Endpoint string, name string, options ...challenger.Option) *challenger.Helper {
+	opts := []challenger.Option{
+		func(c *config.Config) {
+			c.GameAddress = g.addr
+			c.GameDepth = cannonGameDepth
+			c.TraceType = config.TraceTypeCannon
+			// By default, agree with the proposed output root
+			c.AgreeWithProposedOutput = true
```
refcell: Unless we're trying to be explicit here, this is redundant since the default `c *config.Config` will have `AgreeWithProposedOutput = true` by default in `op-e2e/e2eutils/challenger/helper.go#NewChallenger`.

op-e2e/faultproof_test.go
```diff
@@ -143,6 +144,32 @@ func TestChallengerCompleteDisputeGame(t *testing.T) {
...
+	ctx := context.Background()
+	sys, l1Client := startFaultDisputeSystem(t)
+	t.Cleanup(sys.Close)
+
+	disputeGameFactory := disputegame.NewFactoryHelper(t, ctx, sys.cfg.L1Deployments, l1Client)
+	game := disputeGameFactory.StartCannonGame(ctx, common.Hash{0xaa})
+	require.NotNil(t, game)
+
+	game.StartChallenger(ctx, sys.NodeEndpoint("l1"), sys.NodeEndpoint("sequencer"), "Challenger", func(c *config.Config) {
+		c.AgreeWithProposedOutput = true // Agree with the proposed output, so disagree with the root claim
```
refcell, line 160: If I am understanding correctly, this is already set by both the `cannon_helper` and the `challenger/helper.go`. Unless we want to be explicit here, we shouldn't need to do this.

op-e2e/e2eutils/disputegame/game_helper.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		return false, nil
+	})
+	g.require.NoError(err)
+}
+
+func (g *FaultGameHelper) WaitForClaimAtMaxDepth(ctx context.Context, countered bool) {
+	g.WaitForClaim(ctx, func(claim ContractClaim) bool {
+		pos := types.NewPositionFromGIndex(claim.Position.Uint64())
+		return pos.Depth() == g.maxDepth && claim.Countered == countered
```
refcell, line 83: I think we should be able to just do the following since Position is embedded in `Claim`.
```suggestion
		return claim.Depth() == g.maxDepth && claim.Countered == countered
```

op-e2e/e2eutils/disputegame/game_helper.go
```diff
@@ -0,0 +1,111 @@
...
+	g.require.NoError(err)
+	_, err = utils.WaitReceiptOK(ctx, g.client, tx.Hash())
+	g.require.NoError(err)
+}
+
+func (g *FaultGameHelper) WaitForGameStatus(ctx context.Context, expected Status) {
+	g.t.Logf("Waiting for game %v to have status %v", g.addr, expected)
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+	err := utils.WaitFor(ctx, 1*time.Second, func() (bool, error) {
```
refcell: nit
```suggestion
	err := utils.WaitFor(ctx, time.Second, func() (bool, error) {
```

op-e2e/e2eutils/disputegame/game_helper.go
```diff
@@ -0,0 +1,111 @@
...
+func (g *FaultGameHelper) GameDuration(ctx context.Context) time.Duration {
+	duration, err := g.game.GAMEDURATION(&bind.CallOpts{Context: ctx})
+	g.require.NoError(err, "failed to get game duration")
+	return time.Duration(duration) * time.Second
+}
+
+func (g *FaultGameHelper) WaitForClaimCount(ctx context.Context, count int64) {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+	err := utils.WaitFor(ctx, 1*time.Second, func() (bool, error) {
```
refcell: nit
```suggestion
	err := utils.WaitFor(ctx, time.Second, func() (bool, error) {
```

op-e2e/e2eutils/disputegame/helper.go
```diff
@@ -126,129 +107,79 @@ func (h *FactoryHelper) StartAlphabetGame(ctx context.Context, claimedAlphabet s
...
-	_, err = utils.WaitReceiptOK(ctx, g.client, tx.Hash())
-	g.require.NoError(err)
-}
-
-func (g *FaultGameHelper) WaitForGameStatus(ctx context.Context, expected Status) {
-	g.t.Logf("Waiting for game %v to have status %v", g.addr, expected)
+// checkpointL1Block stores the current L1 block in the oracle
+// Returns the L1 block number that was stored as the checkpoint
+func (h *FactoryHelper) checkpointL1Block(ctx context.Context) *big.Int {
 	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
refcell, line 177: nit
```suggestion
	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```

op-e2e/e2eutils/disputegame/helper.go
```diff
@@ -86,38 +85,20 @@ func NewFactoryHelper(t *testing.T, ctx context.Context, deployments *genesis.L1
...
-		h.t.Logf("Latest output index: %v", index)
-		return index.Cmp(big.NewInt(1)) >= 0, nil
-	})
-	h.require.NoError(err, "Did not get two output roots")
+func (h *FactoryHelper) StartAlphabetGame(ctx context.Context, claimedAlphabet string) *AlphabetGameHelper {
+	h.waitForProposals(ctx)
+	l1Head := h.checkpointL1Block(ctx)
 
-	ctx, cancel = context.WithTimeout(ctx, 1*time.Minute)
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
refcell, line 92: nit
```suggestion
	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```

refcell (changes requested):
Few nits and questions - otherwise nice refactoring 💯

op-e2e/e2eutils/disputegame/game_helper.go
ajsutton: Different types in e2e to what you're thinking - this is `ContractClaim` and `Position` is just a `*big.Int` - it's what comes directly from the contract bindings.

op-e2e/e2eutils/disputegame/cannon_helper.go
ajsutton: Yeah being deliberately explicit because it makes the test much easier to read and understand.

op-e2e/e2eutils/disputegame/alphabet_helper.go
ajsutton: There's different semantics between the two - alphabet helper knows the trace that matches the root claim, so it makes sense to inherit that and thus disagree with the proposed output.  Cannon helper doesn't know the trace that matches the root and what will be generated by default is almost certainly going to agree with the proposed output.  But I don't like the inconsistency much either and since we want to be explicit anyway, I'll switch cannon to default to false like the alphabet provider.

op-e2e/faultproof_test.go
ajsutton: Yep being deliberately explicit.  The rule I use is that tests should be explicit about things that are important to what they're testing and accept defaults for things that are just required to make it work.  So we'd be explicit about this but can accept the default for things like l1 endpoint and what network is configured etc.

op-e2e/e2eutils/disputegame/game_helper.go
refcell: Ah makes sense, thanks for the clarification!

refcell (approved):

PR #6628, opened 2023-08-07
feat(op-challenger): Update Alphabet Game Scripts

refcell:
**Description**

Ripped out of #6572, this PR fixes the broken alphabet
game challenger scripts and refactors them in preparation
for the cannon fault dispute game.

The full alphabet dispute game with two challenge agents
can now be run in one bash context.

The game should resolve successfully in favor of charlie.

To run it:
```sh
cd op-challenger
make alphabet
```


refcell:
* develop
  * **PR #6628** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6628" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6628?utm_source=stack-comment).

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,114 @@
...
+  echo "[BLOCK: $(cast block-number)] Waiting for output proposals..."
+  sleep 2
+done
+
+BLOCK_NUMBER=12
+echo "At block number: $BLOCK_NUMBER"
+echo "Verifying that there are at least 2 outputs in the L2 Output Oracle..."
+INDEX=$(cast call $L2_OUTPUT_ORACLE_PROXY "getL2OutputIndexAfter(uint256)" $BLOCK_NUMBER)
+INDEX=$(cast to-dec $INDEX)
+echo "Index: $INDEX"
```
clabby: Should probably block on the condition that there are at least 2 outputs in the oracle. Shouldn't fail as is, but just in case :)

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,114 @@
...
+# (Alphabet) Fault game type = 0
+GAME_TYPE=0
+
+# Root claim commits to the entire trace.
+# Alphabet game claim construction: keccak256(abi.encode(trace_index, trace[trace_index]))
+ROOT_CLAIM=$(cast keccak $(cast abi-encode "f(uint256,uint256)" 15 122))
+
+# fault dispute game extra data:
+# abi.encode(uint256(l2_block_number), uint256(l1_block_number))
+EXTRA_DATA=$(cast abi-encode "f(uint256,uint256)" $BLOCK_NUMBER $EXTRA_DATA_ARG)
```
clabby: nit: For the reader's clarity, can we rename these vars to `L2_BLOCK_NUMBER` and `L1_BLOCK_NUMBER`? The current `L1_BLOCK_NUMBER` var could be called `L1_CHECKPOINT_BLOCK` or something to that effect.

op-challenger/scripts/visualize.sh
```diff
@@ -0,0 +1,24 @@
...
+echo "----------------------------------------------------------------"
+
+FAULT_GAME_ADDRESS=$1
+
+dir=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+cd "$dir"
+cd ../packages/contracts-bedrock
+
+forge script scripts/FaultDisputeGameViz.s.sol --sig "remote(address)" $FAULT_GAME_ADDRESS --fork-url http://localhost:8545
+mv dispute_game.svg "$dir"
```
clabby: Maybe we can link to the more modern visualizer in this script which hooks into the local devnet? URL schema is:
```
https://dispute.clab.by/game?addr=<address>
```
Optional, though, as this viz script will work for the Cannon game with a much lower depth. Could switch depending on the game type, maybe.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,114 @@
...
+# We need to check that the block oracle contains the corresponding l1 block number.
+echo "Checkpointing the block oracle..."
+L1_BLOCK_NUMBER=$(cast send --private-key $DEVNET_SPONSOR $BLOCK_ORACLE_PROXY "checkpoint()" --json | jq -r .blockNumber)
+echo "L1 Block Number: $L1_BLOCK_NUMBER"
+L1_BLOCK_NUMBER=$(cast to-dec $L1_BLOCK_NUMBER)
+echo "L1 Block Number (dec): $L1_BLOCK_NUMBER"
+((EXTRA_DATA_ARG=L1_BLOCK_NUMBER-1))
+
+echo "Getting the l2 output at index 0"
+((PRIOR_INDEX=INDEX-1))
```
clabby:
```suggestion
((PRIOR_INDEX=INDEX-1))
echo "Getting the l2 output at index $PRIOR_INDEX"
```

clabby (commented):
Great work, thanks for updating these 🥳

op-challenger/scripts/alphabet/charlie.sh
```diff
@@ -1,22 +1,21 @@
...
 MALLORY_ADDRESS="0x4641c704a6c743f73ee1f36C7568Fbf4b80681e4"
 MALLORY_KEY="28d7045146193f5f4eeb151c4843544b1b0d30a7ac1680c845a416fac65a7715"
 
 FAULT_GAME_ADDRESS="0x8daf17a20c9dba35f005b6324f493785d239719d"
+FAULT_GAME_ADDRESS=$(cat ../../.fault-game-address)
+echo "Fault dispute game address: $FAULT_GAME_ADDRESS"
 
 PREIMAGE_ORACLE_ADDRESS="0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9"
 
 ./bin/op-challenger \
```
tynes: i am confused by these 2 lines, something doesn't seem right with two calls to op-challenger

also prefer no `cd $DIR` and prefixing paths with `$DIR`

op-challenger/scripts/alphabet/mallory.sh
```diff
@@ -1,22 +1,21 @@
...
 MALLORY_ADDRESS="0x4641c704a6c743f73ee1f36C7568Fbf4b80681e4"
 MALLORY_KEY="28d7045146193f5f4eeb151c4843544b1b0d30a7ac1680c845a416fac65a7715"
 
 FAULT_GAME_ADDRESS="0x8daf17a20c9dba35f005b6324f493785d239719d"
+FAULT_GAME_ADDRESS=$(cat ../../.fault-game-address)
+echo "Fault dispute game address: $FAULT_GAME_ADDRESS"
 
 PREIMAGE_ORACLE_ADDRESS="0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9"
 
 ./bin/op-challenger \
```
clabby:
```suggestion
```

op-challenger/scripts/alphabet/charlie.sh
clabby:
```suggestion
```

op-challenger/scripts/alphabet/mallory.sh
clabby: ^ The above line (in both `charlie.sh` and `mallory.sh`) points towards the wrong path for the binary.

op-challenger/scripts/alphabet/charlie.sh
clabby: ibid

ajsutton: Why no `cd $DIR`? It makes everything so much simpler and avoids a lot of bugs when you forget to prefix `$DIR` to a path. The working directory isn't changed for the shell that executes this - it's only within the subshell actually running the script (unless you `source` the script but that's going to pollute your env vars in other ways anyway)

ajsutton: I probably would have gone with cd `$DIR/../../` though so the working dir for the script is the `op-challenger` dir.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
```
ajsutton, line 1: This file and charlie.sh and mallory.sh needs to be marked executable.

op-challenger/scripts/visualize.sh
ajsutton: We probably shouldn't embed links to external websites in the monorepo. Plus the hosted version won't work without making some pretty insecure changes to allow insecure requests from the https hosted site.

op-challenger/scripts/parallel.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
ajsutton, line 1: Also needs to be made executable.

op-challenger/scripts/alphabet/mallory.sh
```diff
@@ -1,22 +1,28 @@
...
-cd "$DIR"
+SOURCE_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+CHALLENGER_DIR=$(echo ${SOURCE_DIR%/*/*})
 
-DISPUTE_GAME_PROXY="0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"
+# Check that the fault game address file exists
+FAULT_GAME_ADDR_FILE="$CHALLENGER_DIR/.fault-game-address"
+if [ ! -f "$FAULT_GAME_ADDR_FILE"]; then
+    echo "Game not initialized, initializing..."
+    $SOURCE_DIR/init_game.sh
```
trianglesphere: I think this is trying to be too clever by running the initialization here. Because it is ran in both charlie & mallory, the likely case is that you attempt to run the init at the same time by both actors & I'm not sure what would happen at that point. It's a good idea to just exit at this point. Typically you'd be running this with `make alphabet` so the init will be taken care of.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -1,43 +1,32 @@
...
-cd "$DIR"
+set -euo pipefail
 
-cd ../../
+SOURCE_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+CHALLENGER_DIR=$(echo ${SOURCE_DIR%/*/*})
+MONOREPO_DIR=$(echo ${SOURCE_DIR%/*/*/*})
 
+cd $MONOREPO_DIR
 make
```
trianglesphere, line 10: This no longer builds a fresh instance of the challenger every time that you are running the game because this `make` is being ran from the top-level makefile, not from the `op-challenger` makefile.

refcell: The top-level `make` does re-build the challenger binary as well.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,99 @@
...
+make devnet-up
+
+DEVNET_SPONSOR="ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
+DISPUTE_GAME_PROXY=$(jq -r .DisputeGameFactoryProxy $MONOREPO_DIR/.devnet/addresses.json)
+
+echo "----------------------------------------------------------------"
+echo " Dispute Game Factory at $DISPUTE_GAME_PROXY"
+echo "----------------------------------------------------------------"
+
+L2_OUTPUT_ORACLE_PROXY=$(jq .L2OutputOracleProxy $MONOREPO_DIR/.devnet/addresses.json)
```
tynes: `jq -r` will remove the need for `tr -d '"'`

op-challenger/scripts/alphabet/init_game.sh
ajsutton: Building the entire monorepo feels like overkill though unless we really need to. Don't want to spend much time on it but if we can trim this down to just building the things we actually need that would be a win.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -0,0 +1,99 @@
...
+
+CHARLIE_ADDRESS="0xF45B7537828CB2fffBC69996B054c2Aaf36DC778"
+CHARLIE_KEY="74feb147d72bfae943e6b4e483410933d9e447d5dc47d52432dcc2c1454dabb7"
+
+MALLORY_ADDRESS="0x4641c704a6c743f73ee1f36C7568Fbf4b80681e4"
+MALLORY_KEY="28d7045146193f5f4eeb151c4843544b1b0d30a7ac1680c845a416fac65a7715"
+
+echo "----------------------------------------------------------------"
+echo " - Fetching balance of the sponsor"
+echo " - Balance: $(cast balance 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266)"
```
tynes, line 44: Slight nit: this could get out of sync if the `$DEVNET_SPONSOR` changes, could do `cast wallet address --private-key` altho it doesn't output a nice format that is easy to parse, would need to string parse with `awk` or something else

op-challenger/scripts/alphabet/mallory.sh
```diff
@@ -1,22 +1,28 @@
...
 
-DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
-cd "$DIR"
+SOURCE_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+CHALLENGER_DIR=$(echo ${SOURCE_DIR%/*/*})
 
-DISPUTE_GAME_PROXY="0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"
+# Check that the fault game address file exists
+FAULT_GAME_ADDR_FILE="$CHALLENGER_DIR/.fault-game-address"
+if [[ ! -f "$FAULT_GAME_ADDR_FILE"]]; then
```
tynes: nit: missing space before `]]`

op-challenger/scripts/alphabet/mallory.sh
```diff
@@ -1,22 +1,28 @@
...
+    echo "Game not initialized, exiting..."
+    exit 1
+fi
 
+DISPUTE_GAME_PROXY="0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"
```
tynes: This depends on the order of the deploy script sending transactions? Could you get this from `.devnet/addresses.json`? Perhaps could check that `cast code` isn't `0x` to catch errors faster?

op-challenger/scripts/alphabet/charlie.sh
```diff
@@ -1,22 +1,28 @@
...
 
-DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
-cd "$DIR"
+SOURCE_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
+CHALLENGER_DIR=$(echo ${SOURCE_DIR%/*/*})
 
-DISPUTE_GAME_PROXY="0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"
+# Check that the fault game address file exists
+FAULT_GAME_ADDR_FILE="$CHALLENGER_DIR/.fault-game-address"
+if [[ ! -f "$FAULT_GAME_ADDR_FILE"]]; then
```
ajsutton: Getting:
```
./scripts/alphabet/mallory.sh: line 9: syntax error in conditional expression: unexpected token `;'
./scripts/alphabet/charlie.sh: line 9: syntax error in conditional expression: unexpected token `;'
```
I think you need:
```suggestion
if [[ ! -f "$FAULT_GAME_ADDR_FILE" ]]; then
```
here and in mallory.

ajsutton (commented):
Ok, I think we're bike shedding and over engineering this way too much at this point. Let's get it working and merged. It's meant to just be a simple little script to help run a demo not actually something we need to rely on.

op-challenger/scripts/alphabet/init_game.sh
refcell: Since this has been used and likely will for a while, let's move this out of this PR since it's quite out of scope and I don't see a good minimal solution.

op-challenger/scripts/alphabet/mallory.sh
refcell: Actually not used, removed. Will follow up in another pr to add the preimageoracle address in there.

clabby (approved):

ajsutton (approved):
LGTM.

PR #6630, opened 2023-08-07
cannon: delete extra go.mod

tynes:
**Description**

Makes dependabot work more easily and keep the
deps in sync.


protolambda (approved):

trianglesphere:
Does this need a rebase? Error in tests is `open ../example/bin/claim.elf`. Maybe CI go something wrong? This doesn't appear to change it.

trianglesphere:
bump @tynes, would a merge in from develop fix this?

tynes:
Maybe, not sure what it's failing

protolambda:
The `claim.elf` that the testing depends on is not getting build, likely due to the effective change in `go.mod` position

tynes:
I rebased this, not sure exactly what the solution is, will close if nothing obvious

PR #6634, opened 2023-08-08
op-challenger: Pass through network config to cannon

ajsutton:
**Description**

Adds CLI flags to specify either a named network or custom rollup/genesis configs to op-challenger which are passed through to `op-program` when running cannon.

**Tests**

Unit tests added.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4331/support-custom-networks-in-cannon-trace-provider


op-challenger/cmd/main_test.go
```diff
@@ -226,6 +227,67 @@ func TestCannonSnapshotFreq(t *testing.T) {
...
+		configForArgs(t, addRequiredArgsExcept(config.TraceTypeAlphabet, "--cannon-network"))
+	})
+
+	t.Run("NotRequiredWhenRollupAndGenesIsSpecified", func(t *testing.T) {
+		configForArgs(t, addRequiredArgsExcept(config.TraceTypeCannon, "--cannon-network",
+			"--cannon-rollup-config=rollup.json", "--cannon-l2-genesis=genesis.json"))
+	})
+
+	t.Run("Valid", func(t *testing.T) {
+		cfg := configForArgs(t, addRequiredArgsExcept(config.TraceTypeCannon, "--cannon-network", "--cannon-network=goerli"))
```
refcell: Instead of hardcoding `goerli` in here, I think we should pull in the chaincfg available networks and just use the first one perhaps?

op-challenger/cmd/main_test.go
```diff
@@ -226,6 +227,67 @@ func TestCannonSnapshotFreq(t *testing.T) {
...
+		"flag cannon-network or cannon-rollup-config and cannon-l2-genesis is required",
+		addRequiredArgsExcept(config.TraceTypeCannon, "--cannon-network", "--cannon-l2-genesis=gensis.json"))
+}
+
+func TestMustNotSpecifyNetworkAndRollup(t *testing.T) {
+	verifyArgsInvalid(
+		t,
+		"flag cannon-network can not be used with cannon-rollup-config and cannon-l2-genesis",
+		addRequiredArgsExcept(config.TraceTypeCannon, "--cannon-network",
+			"--cannon-network=mainnet", "--cannon-rollup-config=rollup.json"))
```
refcell: Same thing as my other comment about hardcoding `mainnet` here. Can we just pull in the first available network name from the chaincfg available networks?

op-challenger/config/config.go
```diff
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/ethereum-optimism/optimism/op-node/chaincfg"
```
refcell, line 7: It feels weird to add the op-node as a dependency here, but I see the tradeoff between added dependency and duplicated code. It also makes sense that a challenge agent should only play on the networks available with derivation clients.

refcell (approved):
Few comments but none are blockers and can likely be dismissed for this pr. Approved to prevent timezone blocking.

op-challenger/config/config.go
ajsutton: Yeah good point. I suspect chain configs should move to their own module in op-service.  There's a weird split in op-program too where it uses the rollup config from op-node and has its own list of l2 chain configs. Would make more sense to have them both together.

Not going to try and tackle that in this PR but will have a play with it in a follow up PR and see if it works.

PR #6635, opened 2023-08-08
op-program: Set l1.rpckind in the goerli verification script

ajsutton:
**Description**

Optimise fetching data in the goerli verification job by setting rpckind to alchemy which is what it's currently using. Fallback from alchemy to basic works well if it ever needs to (whereas `any` doesn't seem to fallback as well sadly).


Inphi (approved):
This is OK. It'll be a bit better to introduce a new env option that sets the rpckind. I sometimes run the verify script locally on a non-alchemy node, so I'll have to tweak the script to avoid using alchemy methods.

ajsutton:
Modified it to take the rpc kind as an optional third arg, defaulting to alchemy so should still be easy enough to run locally with something else (and it should fallback to the basic approach its using now anyway).

PR #6636, opened 2023-08-08
feat(ctb): make MIPS oracle addrs immutable

Inphi:
PreimageOracle address in MIPS.sol is qualified as immutable for extra safety. An explicit getter function is provided to maintain external ABI compat.


packages/contracts-bedrock/src/cannon/MIPS.sol
```diff
@@ -54,11 +54,15 @@ contract MIPS {
     uint32 constant EINVAL = 0x16;
 
     /// @notice The preimage oracle contract.
-    IPreimageOracle public oracle;
+    IPreimageOracle internal immutable ORACLE;
```
tynes, line 56: This would imply that all chains would use the same mips contract

Inphi: The consequence, IIUC, is mips contracts across all chains have the same codehash? This is fine though.

Inphi: The MIPS contract should be completely stateless, operating as a _pure_ function. It works the same way across all networks. Having an immutable oracle baked into the contract gets us there.

packages/contracts-bedrock/src/cannon/MIPS.sol
```diff
@@ -54,11 +54,15 @@ contract MIPS {
...
 
     /// @param _oracle The address of the preimage oracle contract.
     constructor(IPreimageOracle _oracle) {
-        oracle = _oracle;
+        ORACLE = _oracle;
+    }
+
```
tynes, line 62: needs natspec

tynes:
Looks good to me once the natspec is added!

packages/contracts-bedrock/src/cannon/MIPS.sol
Inphi: Yup. Natspec added.

refcell:
Just needs a bindings fix and rebase on develop for flaky tests (?)

tynes (approved):

tynes:
@mergifyio refresh

PR #6638, opened 2023-08-08
feat(ctb): Cannon contracts VM tests

Inphi:
**Description**

Improves observable test coverage of cannon contracts. These tests are for the most part redundant because the Go VM tests are more comprehensive. However, having forge tests allow us to observe code coverage.

**Metadata**

- Fixes CLI-4155


clabby:
Rebased, pesky chain mon docker publish job is failing.

clabby (approved):
Monster PR, tests LGTM. Thank you for doing this!

One small nit is that we use `assertEq(L, R)` rather than `assertTrue(L == R)`, but this can be addressed in another PR.

Inphi:
I initially tried `assertEq` but that doesn't work on bytes32. But now I've realized that `assertEq32` exists. I'll switch over to that.

PR #6643, opened 2023-08-08
RFP/DoNotMerge(StandardBridge): ERC20 Specific StandardBridge Extension

hamdiallam:
We want our thridparty bridge implementations to follow the StandardBridge abstraction. The StandardBridge
being an abstract class versus an interface is great since it enforces execution flow for bridging.

Here's an idea to how we can get our current third party bridges upgraded (DAI, SNX). An extension to the
StandardBridge limited to a single ERC20 token contract.

Some Extra Thoughts:
- via #6555, the StandardBridge can extract the right messenger from the SystemConfig rather than explicitly set on construction. This also allows us to route ETH sends directly to the optimism portal or deployed StandardBridge, regardless of the bridge contract it originates
     - For L1 Implementations. The L2 implementations would make use of the predeploy addresses

### DIFF
--- a/packages/contracts-bedrock/src/universal/StandardBridgeExtension.sol
+++ b/packages/contracts-bedrock/src/universal/StandardBridgeExtension.sol
@@ -1,0 +1,82 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { StandardBridge } from "./StandardBridge.sol";
+
+abstract contract StandardBridgeExtension is StandardBridge {
+    address public immutable LOCAL_TOKEN;
+    address public immutable REMOTE_TOKEN;
+
+    constructor(
+        address payable _messenger,
+        address payable _otherBridge,
+        address _localToken,
+        address _remoteToken
+    ) StandardBridge(_messenger, _otherBridge) {
+        LOCAL_TOKEN = _localToken;
+        REMOTE_TOKEN = _remoteToken;
+    }
+
+    modifier onlyERC20(address _localToken, address remoteToken) {
+        require(
+            LOCAL_TOKEN == _localToken && REMOTE_TOKEN == remoteToken,
+            "ERC20StandardBridge: can only bridge specified ERC20"
+        );
+        _;
+    }
+
+    /**
+     * Disable ETH Bridging
+     *   - (ideally we could re-route this to the official StandardBridge via the right address from SytemConfig)
+     */
+
+    function _initiateBridgeETH(
+        address _from,
+        address _to,
+        uint256 _amount,
+        uint32 _minGasLimit,
+        bytes memory _extraData
+    ) internal pure override {
+        revert("StandardBridgeExtensions: cannot bridge ETH using a standard bridge extension");
+    }
+
+    receive() external override payable {
+        revert("StandardBridgeExtensions: cannot bridge ETH using a standard bridge extension");
+    }
+
+    /**
+     * Bridge only the registered ERC20
+     */
+
+    /// @notice Bridge the registered ERC20 tokeen to the sender's address
+    function bridge(uint256 _amount, uint32 _minGasLimit, bytes calldata _extraData) public virtual {
+        bridgeERC20(LOCAL_TOKEN, REMOTE_TOKEN, _amount, _minGasLimit, _extraData);
+    }
+
+    /// @notice Bridge the registered ERC20 tokeen to the specified receiver's address.
+    function bridgeTo(address _to, uint256 _amount, uint32 _minGasLimit, bytes calldata _extraData) public virtual {
+        bridgeERC20To(LOCAL_TOKEN, REMOTE_TOKEN, _to, _amount, _minGasLimit, _extraData);
+    }
+
+    function bridgeERC20(
+        address _localToken,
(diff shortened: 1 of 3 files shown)


hamdiallam:
* develop
  * **PR #6643** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6643" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6643?utm_source=stack-comment).

packages/contracts-bedrock/src/universal/StandardBridgeExtension.sol
```diff
@@ -0,0 +1,82 @@
...
+        revert("StandardBridgeExtensions: cannot bridge ETH using a standard bridge extension");
+    }
+
+    /**
+     * Bridge only the registered ERC20
+     */
+
+    /// @notice Bridge the registered ERC20 tokeen to the sender's address
+    function bridge(uint256 _amount, uint32 _minGasLimit, bytes calldata _extraData) public virtual {
+        bridgeERC20(LOCAL_TOKEN, REMOTE_TOKEN, _amount, _minGasLimit, _extraData);
```
maurelian, line 54: This means that both `bridge()` and `bridgeERC20()` will exist, which IMO increases the attack surface and number of things an implementer needs to get right.

packages/contracts-bedrock/src/universal/StandardBridgeExtension.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { StandardBridge } from "./StandardBridge.sol";
+
+abstract contract StandardBridgeExtension is StandardBridge {
```
maurelian, line 7: Can you add some natspec like this to explain why this contract exists? 

https://github.com/ethereum-optimism/optimism/blob/c7c8930b9b37e5937057f8c34a9385c74a263aec/packages/contracts-bedrock/src/universal/StandardBridge.sol#L15

packages/contracts-bedrock/src/universal/StandardBridgeExtension.sol
```diff
@@ -0,0 +1,82 @@
...
+        address payable _messenger,
+        address payable _otherBridge,
+        address _localToken,
+        address _remoteToken
+    ) StandardBridge(_messenger, _otherBridge) {
+        LOCAL_TOKEN = _localToken;
+        REMOTE_TOKEN = _remoteToken;
+    }
+
+    modifier onlyERC20(address _localToken, address remoteToken) {
```
tynes, line 21: This is very particular for only a single token per bridge which may be good for SNX and DAI but unclear to me if other projects would want something else. If this is meant just for them then i understand that

packages/contracts-bedrock/src/universal/StandardBridgeExtension.sol
```diff
@@ -0,0 +1,82 @@
...
+        revert("StandardBridgeExtensions: cannot bridge ETH using a standard bridge extension");
+    }
+
+    receive() external override payable {
+        revert("StandardBridgeExtensions: cannot bridge ETH using a standard bridge extension");
+    }
+
+    /**
+     * Bridge only the registered ERC20
+     */
```
tynes, line 50: We have a particular style for the contracts, you should look at them and read the [style guide](https://github.com/ethereum-optimism/optimism/blob/a4f2261b5b12b01009dcd8d6a605eea7fa68b2e1/packages/contracts-bedrock/STYLE_GUIDE.md?plain=1#L2)

packages/contracts-bedrock/test/StandardBridgeExtension.t.sol
```diff
@@ -0,0 +1,104 @@
...
+        super.setUp();
+        bridge = new StandardBridgeExtensionTester ({
+            _messenger: payable(address(0)),
+            _otherBridge: payable(address(0)),
+            _localToken: address(0),
+            _remoteToken: address(1)
+        });
+    }
+
+    function test_revertIfWrongERC20() external {
```
maurelian, line 32: I think the style guide will suggest a different test naming convention with `reverts` at the end.

hamdiallam:
The goal of the extension contract
- Disable ETH bridging to prevent a footgun
- Single-token Bridge use case

After an offline conversation with @tynes, I'm going to try a different strategy which is to split the `StandardBridge.sol` contract into two abstract classes for ETH bridging and ERC20 Bridging (mixed into the final StandardBridge.sol base contract).

That way thirdparty bridges can simply inherit from only the ERC20 part of the standardbridge making this extension contract not really needed as it's trivial to add the additional logic to scope it to a single (or even n > 1) tokens

Lets see if the split can be done in a backwards compatible manner cc @maurelian

maurelian:
> After an offline conversation with @tynes, I'm going to try a different strategy which is to split the StandardBridge.sol contract into two abstract classes for ETH bridging and ERC20 Bridging (mixed into the final StandardBridge.sol base contract).

Nice, I like this direction and was thinking about it myself. I appreciate that it will also remove the need for a deeper inheritance tree in those bridges. 
GL with that refactor :)

hamdiallam:
Created the split in #6673

PR #6644, opened 2023-08-08
Use make clean to clear op-e2e state

trianglesphere:
**Description**

This PR does two things:
1. Modify the op-e2e makefile to use a `make` / `make clean` approach to updating it's dependencies. It will generate all of it's dependencies only if they do not exist. This is because it's slightly complicated to tell when they should be regenerated.
2. Fix the `new-file.sh` script to work on `darwin` and `linux`. Darwin ships with `bsd` `stat` which is not flag compatible with `gnu` stat.


refcell:
This just needs a rebase now that #6645 is merged i believe

op-e2e/Makefile
```diff
@@ -1,14 +1,29 @@
...
+
+cannon-prestate:
+	make -C .. cannon-prestate
+
 devnet-allocs:
-	PYTHONPATH=../bedrock-devnet python3 ../bedrock-devnet/main.py --monorepo-dir=.. --allocs
+	make -C .. devnet-allocs
+
+
+pre-test: pre-test-cannon pre-test-allocs
```
ajsutton, line 11: I'd suggest we also update the circleci config to run this rather than each individual component. That way circleci will enforce that this stays up to date as dependencies change in the future.

op-e2e/Makefile
```diff
@@ -1,14 +1,29 @@
+test: pre-test
+	go test -v ./...
+
+cannon-prestate:
+	make -C .. cannon-prestate
+
 devnet-allocs:
-	PYTHONPATH=../bedrock-devnet python3 ../bedrock-devnet/main.py --monorepo-dir=.. --allocs
+	make -C .. devnet-allocs
```
ajsutton: I think this actually depends on cannon-prestate existing since the prestate is required when deploying the fault dispute contract so probably worth stating that in the makefile.

ajsutton (commented):
It's kind of frustrating that we're not getting any help from our build tool in terms of only rebuilding things that changed. Fixing that is a pretty major project though...

I think this is probably the best approach given where we're at, but will hold off approving so others who are more familiar with this genesis test setup have a chance to review.

op-e2e/Makefile
trianglesphere: oh it does, I should probably leave a comment, I had to switch the order of `pre-test` so it depends on `cannon` before `allocs`.

op-e2e/Makefile
trianglesphere: just run `make test`? I'd assume that nothing would exist yet, but we could also do `make pre-test` to be extra sure.

trianglesphere: wait actually, `make pre-test` always gets ran, the only question is if `make cannon-prestate` or `make devnet-allocs` gets ran (via if files exist).

trianglesphere:
> It's kind of frustrating that we're not getting any help from our build tool in terms of only rebuilding things that changed. Fixing that is a pretty major project though...

Make can do it on a file by file basis, but then you have to put everything into make which we may not want to do.

I'd asked mark to use file timestamps, but based adding the cannon-prestate, I lean to this. (Plus I'm not 100% sure the file timestamps thing is working because IDK if `atime` propagates up. There's the also fun thing where `stat` on linux/mac have completely different flags).

op-e2e/Makefile
trianglesphere: ok, I did the right thing & now this depends on `pre-test-cannon`

op-e2e/Makefile
trianglesphere: ok, so we use a custom runner rather than `make test` so I updated this.

tynes (approved):

PR #6645, opened 2023-08-08
op-chain-ops: Fix flaky TestCode

trianglesphere:
**Description**

If getting an empty code value in the state DB, geth returns `nil`, not an empty array. This is funcationally equivalent if you are checking the length, but not the same if you expect to get the same object out that you put in. This now checks for this special case & prevents the test from flaking.

This can be tested by running `go test ./state -count=10 -run=TestCode` in the op-chain-ops folder. Without this change the test will fail, with this change it will now pass.


ajsutton (approved):
Out of interest, would using `require.EqualValues` also work here?

trianglesphere:
> would using require.EqualValues also work here

I just tested, did not.

PR #6646, opened 2023-08-09
op-chain-ops: initialized + initializing in check-l2

tynes:
**Description**

Logs the values of initialized and initializing in `check-l2`. They are not public so we read them directly from storage using the storage layout.


refcell (approved):
Very Clean, Much Wow

PR #6647, opened 2023-08-09
contracts-bedrock: remove compiler warnings in deploy script

tynes:
**Description**

The only devnet modifier causes branching logic in a way where sometimes nothing is returned, triggering a compiler warning. This compiler warning can be removed by using named return args. Refactor the deploy script to use named return args to remove the compiler warnings.


refcell (commented):
Looks good but can we use our underscore suffix style for return vars?

refcell (approved):
Should be gtg now

PR #6670, opened 2023-08-09
feat(ctb): Consolidate lock files into a single directory

maurelian:
**Description**

Does the following in the contracts package: 

1. moves all lock/snapshot files to a single `locks` dir. This will ensure that these files appear together during PR reviews. 
2. uses the term 'lock' more consistently as a means of codifying this pattern.


.circleci/config.yml
```diff
@@ -382,10 +382,10 @@ jobs:
             pnpm lint:check || echo "export LINT_STATUS=1" >> "$BASH_ENV"
           working_directory: packages/contracts-bedrock
       - run:
-          name: gas snapshot
+          name: gas lock
           command: |
             forge --version
-            pnpm gas-snapshot --check || echo "export GAS_SNAPSHOT_STATUS=1" >> "$BASH_ENV"
+            pnpm gas-lock --check || echo "export GAS_LOCK_STATUS=1" >> "$BASH_ENV"
```
maurelian, line 388: this kind of breaks the standard `gas-snapshot` terminology, so I'd be OK with reverting this.

clabby: Not opinionated here tbh. Both mean the same thing in my head 😄

packages/contracts-bedrock/scripts/storage-lock.sh
```diff
@@ -38,15 +38,15 @@ dir=$(dirname "$0")
 
 echo "Creating storage layout diagrams.."
 
-echo "=======================" > $dir/../.storage-layout
-echo "👁👁 STORAGE LAYOUT snapshot 👁👁" >> $dir/../.storage-layout
-echo "=======================" >> $dir/../.storage-layout
+echo "=======================" > $dir/../.storage-lock
+echo "👁👁 STORAGE LAYOUT LOCK 👁👁" >> $dir/../.storage-lock
+echo "=======================" >> $dir/../.storage-lock
```
clabby:
```suggestion
echo "=======================" > $dir/../locks/storage-lock
echo "👁👁 STORAGE LAYOUT LOCK 👁👁" >> $dir/../locks/storage-lock
echo "=======================" >> $dir/../locks/storage-lock
```

packages/contracts-bedrock/scripts/storage-lock.sh
```diff
@@ -38,15 +38,15 @@ dir=$(dirname "$0")
...
 for contract in ${contracts[@]}
 do
-  echo -e "\n=======================" >> $dir/../.storage-layout
-  echo "➡ $contract">> $dir/../.storage-layout
-  echo -e "=======================\n" >> $dir/../.storage-layout
-  forge inspect --pretty $contract storage-layout >> $dir/../.storage-layout
+  echo -e "\n=======================" >> $dir/../.storage-lock
+  echo "➡ $contract">> $dir/../.storage-lock
+  echo -e "=======================\n" >> $dir/../.storage-lock
+  forge inspect --pretty $contract storageLayout >> $dir/../.storage-lock
```
clabby:
```suggestion
  echo -e "\n=======================" >> $dir/../locks/storage-lock
  echo "➡ $contract">> $dir/../locks/storage-lock
  echo -e "=======================\n" >> $dir/../locks/storage-lock
  forge inspect --pretty $contract storageLayout >> $dir/../locks/storage-lock
```

packages/contracts-bedrock/scripts/storage-lock.sh
```diff
@@ -38,15 +38,15 @@ dir=$(dirname "$0")
...
 done
-echo "Storage layout snapshot stored at $dir/../.storage-layout"
+echo "Storage layout lock stored at $dir/../.storage-lock"
```
clabby:
```suggestion
echo "Storage layout lock stored at $dir/../locks/storage-lock"
```

clabby (changes requested):
Looks great, nice cleanup. Needs a couple edits to `storage-lock.sh`, will approve after :)

maurelian:
Good catch on those broken paths, thanks @clabby.

clabby:
Can't leave an inline comment because it's out of scope of the changes, but in the circleci config, the result of the gas snapshot check is currently ignored. The result is now stored in an env var called `GAS_LOCK_STATUS`, but the conditional in the `contracts-bedrock-checks` job checks for the `GAS_SNAPSHOT_STATUS` (which doesn't exist anymore)

(L425 of the circleci config)

clabby (approved):
LGTM!

twoshark (approved):

refcell (approved):
This is awesome

refcell:
Merged my small fix since CI passed 😄

PR #6671, opened 2023-08-09
feat(indexer): Store RLP bytes of logs/headers when indexing them

hamdiallam:
Consumers of the database will likely expect to get information of logs/headers
by decoding the RLP bytes. To avoid having to make extra RPC calls to a node
node for this data, lets store those bytes alongside the indexable fields for
headers and logs

Whenever topics are missing from logs, we'll use bytes32(0x0) as the event_signature. It's not worth supporting tooling for anon events as it's impossible to determine without the contract ABI

### DIFF
--- a/indexer/database/contract_events.go
+++ b/indexer/database/contract_events.go
@@ -23,19 +23,28 @@ type ContractEvent struct {
 	EventSignature common.Hash `gorm:"serializer:json"`
 	LogIndex       uint64
 	Timestamp      uint64
+
+	GethLog *types.Log `gorm:"serializer:rlp;column:rlp_bytes"`
 }
 
 func ContractEventFromGethLog(log *types.Log, timestamp uint64) ContractEvent {
+	eventSig := common.Hash{}
+	if len(log.Topics) > 0 {
+		eventSig = log.Topics[0]
+	}
+
 	return ContractEvent{
 		GUID: uuid.New(),
 
 		BlockHash:       log.BlockHash,
 		TransactionHash: log.TxHash,
 
-		EventSignature: log.Topics[0],
+		EventSignature: eventSig,
 		LogIndex:       uint64(log.Index),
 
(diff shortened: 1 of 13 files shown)


hamdiallam:
* develop
  * **PR #6671** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6671" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6671?utm_source=stack-comment).

indexer/database/contract_events.go
```diff
@@ -23,19 +23,28 @@ type ContractEvent struct {
...
 	LogIndex       uint64
 	Timestamp      uint64
+
+	GethLog *types.Log `gorm:"serializer:rlp;column:rlp_bytes"`
 }
 
 func ContractEventFromGethLog(log *types.Log, timestamp uint64) ContractEvent {
+	eventSig := common.Hash{}
+	if len(log.Topics) > 0 {
+		eventSig = log.Topics[0]
```
tynes, line 34: There is technically a difference between `common.Hash{}` and empty, there exists `LOG0` which people call the anonymous log which doesn't have any topic at all. You wouldn't be empty to tell the diff here between `LOG0` and `LOG1 null-hash`

Don't think its a huge deal

hamdiallam: ohh interesting, wasn't aware of this. I wanted to avoid having to deal with pointers hence the 0x0 sentinel value.

Will document this edge case

hamdiallam: also realize anon events with indexed fields will still have topics > 0. Just going to note that anon-events wont be well-supported in the indexer

roninjin10: Wait what's the difference?

roninjin10 (approved):

PR #6676, opened 2023-08-09
feat(op-challenger): Absolute Pre-state Validation

### LINKED ISSUE #6674: feat(op-challenger): Verify Absolute Pre-state
**Description**

The `op-challenger` needs the absolute pre-state to attempt a "step" via its `Agent` component's `AttemptStep` method. In turn, the `Agent` queries it's composed `TraceProvider` for the absolute pre-state, implemented concretely by the cannon trace provider and the alphabet trace provider.

On startup (which is for now single-game spawned), the `op-challenger` should fetch the absolute pre-state from the on-chain  fault dispute game contract and compare this state with that of it's trace provider. If there is a mismatch, an error should be thrown.

### Tasks
- [x] Implement a `FetchAbsolutePrestate` method on the `Loader` component.
- [x] Unit test this new `FetchAbsolutePrestate` loader component method.
- [x] On service creation, validate the onchain absolute pre-state with the absolute pre-state from the local trace provider.

**Considered Alternatives**

Initially the thought was to pass the absolute-prestate in via a cli argument or "embed" this network-specific value in a mapping. But, since it is stored in the fault dispute game contract, it is much less error-prone and efficient to fetch this value from onchain.

**Tests**

Unit tests should be implemented for the `Loader` component.

**Linked Linear Ticket**

CLI-4286

refcell:
**Description**

Implements absolute pre-state validation inside the fault service
component once the loader and trace provider are created.

**Tests**

The newly added loader call to the onchain fault dispute game
contract that queries for the absolute pre-state is tested
via unit tests.

The newly added service function that compares the absolute
prestates fetched from the provided trace provider and loader
is also unit tested.

**Metadata**

Github: Fixes #6674
Linear: Fixes CLI-4286

### DIFF
--- a/op-challenger/fault/loader.go
+++ b/op-challenger/fault/loader.go
@@ -8,9 +8,9 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 )
 
-// ClaimFetcher is a minimal interface around [bindings.FaultDisputeGameCaller].
+// MinimalFaultDisputeGameCaller is a minimal interface around [bindings.FaultDisputeGameCaller].
 // This needs to be updated if the [bindings.FaultDisputeGameCaller] interface changes.
-type ClaimFetcher interface {
+type MinimalFaultDisputeGameCaller interface {
 	ClaimData(opts *bind.CallOpts, arg0 *big.Int) (struct {
 		ParentIndex uint32
 		Countered   bool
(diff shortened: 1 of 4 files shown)


refcell:
* develop
  * **PR #6676** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6676" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6676?utm_source=stack-comment).

op-challenger/fault/loader.go
```diff
@@ -19,22 +19,24 @@ type ClaimFetcher interface {
...
 // Loader is a minimal interface for loading onchain [Claim] data.
 type Loader interface {
 	FetchClaims(ctx context.Context) ([]types.Claim, error)
+	FetchAbsolutePrestate(ctx context.Context) ([]byte, error)
 }
 
 // loader pulls in fault dispute game claim data periodically and over subscriptions.
 type loader struct {
-	claimFetcher ClaimFetcher
+	MinimalFaultDisputeGameCaller
```
ajsutton: I don't think it makes sense that the loader _is_ the caller, it should have a reference to it and not directly expose all of its methods.  So I think we need to go back to assigning this field a name.

op-challenger/fault/service.go
```diff
@@ -82,16 +88,30 @@ func newTypedService(ctx context.Context, logger log.Logger, cfg *config.Config,
...
+	if err != nil {
+		return fmt.Errorf("failed to get the trace provider's absolute prestate: %w", err)
+	}
+	onchainPrestate, err := loader.FetchAbsolutePrestate(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get the onchain absolute prestate: %w", err)
+	}
+	if !bytes.Equal(providerPrestate, onchainPrestate) {
+		return fmt.Errorf("trace provider's absolute prestate does not match onchain absolute prestate")
+	}
```
ajsutton: This will always fail beacause you're comparing the absolute pre state pre-image (from the trace provider) to the absolute pre state hash (from the contract).  Which I think is why all the e2e tests are failing. :)

ajsutton (approved):
LGTM assuming e2e tests now pass as we expect. :)

ajsutton (approved):

PR #6679, opened 2023-08-09
ops: fix dockerfile

tynes:
**Description**

Add in openssh-client to the dockerfile since some of the deps are installed via git/ssh and not having an ssh client in the image prevents it from building.

I don't think that this is a particularly secure thing to do, we should rewrite this dockerfile in a way that the dep install happens in a different image and
then the final built typescript is copied over. None of these services have private keys so I'm not super worried about this.


roninjin10:
Can close for https://github.com/ethereum-optimism/optimism/pull/6682

PR #6681, opened 2023-08-09
op-e2e: Update cannon e2e test to execute moves down to the max depth

ajsutton:
**Description**

The e2e test will now call attack on every claim the challenger makes to drive the game down to the max depth.  The final step isn't called because the depth is such that the invalid claim from the e2e would have to call step and it can't. Will follow up with tests that exercise that.

Game helper pulls the game depth from the contract instead of hard coding it.

**Tests**

Its all about the tests.

**Metadata**


refcell (approved):
LGTM

PR #6683, opened 2023-08-10
op-challenger: Skip oracle updates when no data to publish

ajsutton:
**Description**

Skip publishing pre-image data to the oracle when there is no pre-image data to publish. Resolves a panic when a `nil` array is attempted to be used.

Fix cannon pre-state loading to treat the prestate path from the CLI as relative to the working directory, not cannon's data directory same as for the cannon and op-program binaries.


refcell (approved):
Great, LGTM

PR #6685, opened 2023-08-10
contracts-bedrock: modularize deploy script

tynes:
**Description**

The way to deploy a new OP Stack chain that is part of the superchain will be through
the `deployProxies` function in this script.

Some ordering of the deploy script was shifted around, it should be safe to do so. It will
be obvious by failing devnet deployment if its not safe to do this.

Closes CLI-4263


packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -120,6 +97,38 @@ contract Deploy is Deployer {
         }
     }
 
+    /// @notice Deploy all of the proxies
+    function deployProxies() public {
+        deployAddressManager();
+        deployProxyAdmin();
```
refcell, line 103: Slight preference to rip these out of the proxy-specific function and keep them top level.
Maybe not so much the proxy admin, but the address manager at least.

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -120,6 +97,38 @@ contract Deploy is Deployer {
...
+        deployOptimismPortalProxy();
+        deployL2OutputOracleProxy();
+        deploySystemConfigProxy();
+        deployL1StandardBridgeProxy();
+        deployL1CrossDomainMessengerProxy();
+        deployOptimismMintableERC20FactoryProxy();
+        deployL1ERC721BridgeProxy();
+        deployDisputeGameFactoryProxy();
+
+        transferAddressManagerOwnership();
```
refcell, line 114: ibid - preference to keep this outside the deployProxies function and call at the top-level.

tynes: We need a single function that can be used to deploy the proxies when adding a chain to the superchain

packages/contracts-bedrock/scripts/Deploy.s.sol
tynes: We need the address manager ownership to be transferred as part of the set up proxy step, otherwise the proxies wont be set up

packages/contracts-bedrock/scripts/Deploy.s.sol
refcell: Ah ok, this intuition makes sense. Thanks for the clarification!

refcell (approved):

PR #6686, opened 2023-08-10
proxyd: add support for limiting max block range

mdehoog:
**Description**

We've noticed some folks calling eth_getLogs and other RPC methods with large block ranges, which can cause our nodes to OOM. This PR introduces the ability for proxyd to block requests that have a large block range, similar to other node providers.

**Tests**

Added tests for the rewriter.


proxyd/server.go
```diff
@@ -530,6 +575,25 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
...
+		return true
+	}
+	if fc.BlockHash != nil {
+		return true
+	}
+	if fc.ToBlock == nil || fc.ToBlock.Int64() < 0 {
+		fc.ToBlock = s.latestBlockNumber
+	}
+	if fc.FromBlock == nil || fc.FromBlock.Int64() < 0 {
+		fc.FromBlock = s.latestBlockNumber
```
mdehoog: TODO: guard concurrent reads (or use atomic package)

proxyd/server.go
```diff
@@ -411,6 +458,10 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 			continue
 		}
 
+		if (parsedReq.Method == "eth_newFilter" || parsedReq.Method == "eth_getLogs") && !s.allowedBlockRange(parsedReq) {
+			responses[i] = NewRPCErrorRes(parsedReq.ID, errors.New("block range too large"))
+		}
```
refcell: Looks like this is missing a `continue` here unless I'm misunderstanding
```suggestion
		    continue
		}
```

proxyd/server.go
```diff
@@ -175,7 +186,42 @@ func NewServer(
...
+	for {
+		select {
+		case <-ticker.C:
+			s.srvMu.Lock()
+			if s.shutdown {
+				s.srvMu.Unlock()
+				return
+			}
+			s.updateLatestBlockNumber()
+			s.srvMu.Unlock()
```
refcell: It looks like we're holding this mutex across the entire function. It might make more sense to split the `updateLatestBlockNumber` function into a `fetchLatestBlockNumber() (uint64, error)` and then a `updateLatestBlockNumber(uint64)` that contains the locking logic. Something like:

```golang
func (s *Server) pollLatestBlockNumber() {
	ticker := time.NewTicker(2 * time.Second)
	for {
		select {
		case <-ticker.C:
			bn, err := s.fetchLatestBlockNumber()
            if err != nil {
                s.updateLatestBlockNumber(bn)
            }
		}
	}
}

func (s *Server) updateLatestBlockNumber(bn uint64) {
    s.srvMu.Lock()
    defer s.srvMu.Unlock()
	s.latestBlockNumber.Store(bn)
}

func (s *Server) fetchLatestBlockNumber() (uint64, error) {
	req := json.RawMessage("{\"id\":0,\"jsonrpc\":\"2.0\",\"method\":\"eth_blockNumber\"}")
	res, _, err := s.handleBatchRPC(context.Background(), []json.RawMessage{req}, func(method string) bool { return false }, false)
	if err != nil {
		log.Error("error requesting latest block number", "err", err)
		return uint64(0), err
	}
	bns := res[0].Result.(string)
	bn, err := hexutil.DecodeUint64(bns)
	if err != nil {
		log.Error("error decoding hex block number", "err", err)
		return uint64(0), err
	}
	return bn, nil
}
```

mdehoog: The purpose of using the mutex is to guard against calling `Shutdown` during a call to `updateLatestBlockNumber()`, which would cause the backend groups to shutdown and therefore the `handleBatchRPC` could panic. It also guards the check of the `shutdown` boolean so the goroutine can exit. But if we don't care about this during shutdown, we can remove

proxyd/server.go
mdehoog: good catch! thanks

proxyd/server.go
refcell: SGTM, will dismiss this, still works the current way so no concern here :)

proxyd/latest_block_poller.go
```diff
@@ -0,0 +1,78 @@
...
+	for range ticker.C {
+		p.mutex.Lock()
+		if p.shutdown {
+			ticker.Stop()
+			p.mutex.Unlock()
+			return
+		}
+		p.poll()
+		p.mutex.Unlock()
+	}
```
refcell: Note, a `for range` loop over a receiving channel will _not_ execute until the first interval is passed. In other words, it will first wait 2 seconds and then perform the first poll.

To instead have this execute immediately, we could do something like the following, which will run the loop immediately and wait on receive for each subsequent iteration.
```suggestion
    for ; true; <-ticker.C {
		p.mutex.Lock()
		if p.shutdown {
			ticker.Stop()
			p.mutex.Unlock()
			return
		}
		p.poll()
		p.mutex.Unlock()
	}
```

proxyd/latest_block_poller.go
```diff
@@ -0,0 +1,78 @@
...
+type RoundTripper func(ctx context.Context, req json.RawMessage) (*RPCRes, error)
+
+// NewLatestBlockPoller creates a new LatestBlockPoller and starts polling
+// for the latest block number in a separate goroutine.
+func NewLatestBlockPoller(rt RoundTripper) *LatestBlockPoller {
+	p := &LatestBlockPoller{
+		rt: rt,
+	}
+	p.poll()
+	go p.start()
```
refcell: The comment below addressing the `start` method would allow you to remove this additional `p.poll()` call.

proxyd/latest_block_poller.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Shutdown stops the poller.
+func (p *LatestBlockPoller) Shutdown() {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	p.shutdown = true
+}
+
+func (p *LatestBlockPoller) start() {
+	ticker := time.NewTicker(2 * time.Second)
```
refcell: Might make sense to have this `2 * time.Second` duration be configurable as a `LatestBlockPoller` field and argument to the `NewLatestBlockPoller` constructor.

refcell (commented):
Looking great!

proxyd/latest_block_poller.go
mdehoog: It's a minor thing, but i made the `poll` call in the original goroutine (and allowed the interval to pass in the separate one) so that the latest block has a chance to be populated before returning. But given the handling around 0 blocks, it probably doesn't matter too much. Let me know if you think we should switch to your suggestion.

proxyd/latest_block_poller.go
mdehoog: done!

proxyd/latest_block_poller.go
trianglesphere: I lean to this being fine. The majority of the time proxyd is running, it will have this value populated, & if it doesn't it's not the end of the world.

trianglesphere (commented):
The poller & `IsAllowedBlockRange` look good to me. Because I'm not super familiar with this code, I'd like eyes from someone else to confirm that we're running `IsAllowedBlockRange` in all of the places that we need to be running it.

proxyd/latest_block_poller.go
mdehoog: sgtm, will update

proxyd/latest_block_poller.go
mdehoog: done

felipe-op:
Please take a look at the consensus poller: https://github.com/ethereum-optimism/optimism/blob/develop/proxyd/consensus_poller.go
It is already updating the latest block in a configurable interval. This change seems to duplicate the same behaviour.

You can enable consensus awareness in the backend group configs:
https://github.com/ethereum-optimism/optimism/blob/edafea3e21251e4efbc3075471e6e4a0e088cdd7/proxyd/example.config.toml#L94

Once enabled, block tags such as `latest` will be rewritten by their correspondent block number: https://github.com/ethereum-optimism/optimism/blob/edafea3e21251e4efbc3075471e6e4a0e088cdd7/proxyd/rewriter.go#L117

And you then can just validate the numerical block range interval without worrying with tracking `latest` or converting the block tags to block numbers.

felipe-op (changes requested):
This change should integrate with consensus awareness

proxyd/backend.go
```diff
@@ -220,6 +221,12 @@ func WithConsensusSkipPeerCountCheck(skipPeerCountCheck bool) BackendOpt {
 	}
 }
 
+func WithConsensusForcedCandidate(forcedCandidate bool) BackendOpt {
+	return func(b *Backend) {
+		b.forcedCandidate = forcedCandidate
+	}
+}
```
mdehoog, line 228: Allows a backend to opt-out of all the other consensus health checks (sync, peers, healthy, banned, etc). This is important for us as we only have a single backend and we never want it to not be a candidate.

proxyd/rewriter.go
```diff
@@ -9,9 +9,10 @@ import (
 )
 
 type RewriteContext struct {
-	latest    hexutil.Uint64
-	safe      hexutil.Uint64
-	finalized hexutil.Uint64
+	latest        hexutil.Uint64
+	safe          hexutil.Uint64
+	finalized     hexutil.Uint64
+	maxBlockRange uint64
```
mdehoog, line 15: should the `maxBlockRange` be a field in the `RewriteContext` struct, or passed in as a parameter to the `RewriteTags` / `RewriteRequest` / `rewriteRange` methods?

this way seems cleaner to me

proxyd/rewriter.go
```diff
@@ -121,6 +123,15 @@ func rewriteRange(rctx RewriteContext, req *RPCReq, res *RPCRes, pos int) (Rewri
...
 	}
 
+	// if either fromBlock or toBlock is defined, default the other to "latest" if unset
+	_, hasFrom := p[pos]["fromBlock"]
+	_, hasTo := p[pos]["toBlock"]
+	if hasFrom && !hasTo {
+		p[pos]["toBlock"] = "latest"
+	} else if hasTo && !hasFrom {
+		p[pos]["fromBlock"] = "latest"
+	}
```
mdehoog, line 133: this avoids the situation where somebody uses `fromBlock = 0` but leaves `toBlock` blank, which might be ignored by the proxyd range check, but `toBlock` defaults to `latest` in geth

mdehoog:
@felipe-op great suggestion, reimplemented 👍

proxyd/backend.go
```diff
@@ -692,6 +700,11 @@ func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch
 				})
 				if errors.Is(err, ErrRewriteBlockOutOfRange) {
 					res.Error = ErrBlockOutOfRange
+				} else if errors.Is(err, ErrRewriteRangeTooLarge) {
+					res.Error = &RPCErr{
+						Code:    JSONRPCErrorInternal,
```
mslipper: [Nit] Should be an invalid params error - can use `ErrInvalidParams()` to generate this.

proxyd/rewriter_test.go
```diff
@@ -148,6 +148,16 @@ func TestRewriteRequest(t *testing.T) {
 			expected:    RewriteOverrideError,
 			expectedErr: ErrRewriteBlockOutOfRange,
 		},
+		{
```
mslipper, line 151: Recommend some additional tests

- Range from `earliest` to `latest`
- No `toBlock` defined
- No `fromBlock` defined

mslipper (approved):
LGTM once testing/nits are addressed.

proxyd/rewriter_test.go
mdehoog: good suggestion on `earliest`, picked up a bug where `rewriteTags` doesn't rewrite `earliest` or `pending`

felipe-op (approved):

felipe-op:
Please squash before merging

mdehoog:
thanks... squashed 👍

mdehoog:
(pushed a no-op commit to re-run tests, seems flaky)

mslipper:
Merged in upstream - that should fix the tests.

PR #6687, opened 2023-08-10
op-challenger: Fix provision of pre-state, proof and oracle data from cannon proofs

ajsutton:
Fix provision of pre-state, proof and oracle data from cannon. This gets `step` execution working with cannon for the absolute pre-state and steps in the middle of the trace. 

Combines GetPreimage and GetOracleData into GetStepData. GetStepData returns the data needed to execute the step at the specified trace index, so the state is the pre-state rather than matching the claim at the specified index. The proof data contains proofs for data accessed while executing the step and oracle data the pre-images required to execute the step. This matches up correctly with the data in the cannon proof files and what is required when calling the step function.

For trace indices beyond the end of the actual cannon trace, rather than reusing the last step, we now generate a proof from the final, exited state. Each extended step is effectively a no-op that makes no change to the state so the pre and post-states are identical. This also avoids the need to re-execute cannon to generate the final proof.

Also sets the game depth to 30 so cannon actually has to be executed.


refcell (approved):
LGTM

trianglesphere:
Is the alphabet game creation known to be flaky?

ajsutton:
> Is the alphabet game creation known to be flaky?

Yeah I've seen it flake a bit recently but haven't been able to track down why.  I suspect we need to grab a debug trace of the tx when it fails so we can dig into it more.

Inphi (approved):

PR #6702, opened 2023-08-10
fix(ctb): Use bundled ds-test in forge remap

Inphi:
The other ds-test is an older version that's missing features like `assertNotEq`


refcell:
LGTM, should just need a rebase on develop?

Inphi:
> LGTM, should just need a rebase on develop?

already did this afternoon. It's the accursed chain-mon docker build that's flakey. Need to investigate

tynes (approved):

PR #6703, opened 2023-08-10
spike(ctb): `L2OutputOracleV2`

clabby:
## Overview

> **Note**
> This is a PoC and is incomplete without further changes to the `FaultDisputeGame` contract. Do not merge.

Implements a draft V2 of the `L2OutputOracle` as defined in the PoPs + Output Bisection proposal. This version makes a few fundamental changes:
* The oracle now knows about the `DisputeGameFactory`, and uses it to forward the legacy `proposeL2Output` function to `DisputeGameFactory.create` (which is the new way to propose L2 outputs). The factory is also consulted to authenticate the caller of the new `finalizeProposal()` function.
* `proposeL2Output` is now a legacy function, and it forwards the inputs to `DisputeGameFactory.create`. It's optional to retain this function - we were the only caller of this function before due to the honest proposer assumption, so this should not be an issue if we do remove it.
* Adds a new function, `finalizeProposal()`. This function first checks if the caller is a dispute game that was created by the `DisputeGameFactory` and that it has resolved to `DEFENDER_WINS`. If these conditions are met and there isn't a newer finalized proposal in the oracle already, the game's proposal is finalized.
* `deleteL2Outputs` has been deprecated, but retained for the time being. During the alpha, we will still want to be able to manually delete outputs if there are bugs in the system. In the future, the `CHALLENGER` key and this function will no longer be necessary.

Possible changes:
* This contract can be further simplified if we're comfortable with breaking the API further. We can remove a significant amount of the logic around computing an L2 block's timestamp, some of the immutables like `FINALIZATION_PERIOD_SECONDS`, `SUBMISSION_INTERVAL`, and `L2_BLOCK_TIME`, etc. etc.

**Metadata**
Closes CLI-4338

### DIFF
--- a/packages/contracts-bedrock/semver-lock.json
+++ b/packages/contracts-bedrock/semver-lock.json
@@ -3,6 +3,7 @@
   "src/L1/L1ERC721Bridge.sol": "0xbb10b777d1cd36ef98b53df6675f37a20b14a9a82b174f0d8f8872eedca65f17",
   "src/L1/L1StandardBridge.sol": "0xbd7b303cefe46bc14bf1a2b81e5702ff45ce9c5257524e59778e11c75f7f5bdc",
   "src/L1/L2OutputOracle.sol": "0x05ea17a834563ffa50cade81189b120b6f0805ba316d6a9893c8cf8b231e57e3",
+  "src/L1/L2OutputOracleV2.sol": "0x12f2eb26de3e3a265656776c525f79cf7a084e0b8c6b2b03ab189759df60ebbe",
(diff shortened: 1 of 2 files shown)


clabby:
* develop
  * **PR #6703** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6703" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6703?utm_source=stack-comment).

refcell:
image

clabby:
Closing; Experiment.

PR #6709, opened 2023-08-11
op-e2e: Add a cannon e2e test that executes a defend step

ajsutton:
**Description**

Add a cannon e2e test that executes a defend step.
Establishes infrastructure for e2e tests to create correct cannon trace values.

**Metadata**


refcell (approved):
LGTM

PR #6711, opened 2023-08-11
op-e2e: Specify correct l2 block number when create dispute games

ajsutton:
**Description**

Should fix flakiness when creating games as the previous hard coded block number may not have had suitable outputs published.

**Metadata**


refcell (approved):
Ah that makes sense, great find

PR #6713, opened 2023-08-11
ci: Fix Dockerfile.packages build

Inphi:
A recent change (56e29726511fb86bb2b92ed6c88241fbee469904) replaced git HTTP references in the pnpm lock file with `git@github.com`. This forces git to clone repositories via ssh rather than https. Thus, an ssh client is now needed for the Dockerfile.packages.


Inphi:
Obsoleted by https://github.com/ethereum-optimism/optimism/pull/6729

PR #6729, opened 2023-08-11
build: Fix pnpm lockfile

Inphi:
Regenerates the lockfile so it no longer contains SSH git repositories.
A pnpm hook is introduced to assert during pnpm install that the lockfile does not contain SSH git repos.

We avoid using git-ssh to download packages to accommodate environments that do not have ssh keys setup for github access (ex: docker).

Note: This change will prevent dependabot from submitting PRs that include a pnpm-lock.yaml containing SSH git repos.


roninjin10 (approved):

Inphi:
Merging this manually since the merge queue is currently blocked by this problem.

PR #6737, opened 2023-08-11
bedrock-devnet: Handle exceptions in child processes

Inphi:
This ensures that the bedrock-devnet script exits unsuccessfully when an exception has occurred during contract deployment.


refcell (approved):
LGTM

PR #6743, opened 2023-08-11
op-node: move eth package into op-service

protolambda:
**Description**

Since so many packages depend on the `eth` utils package, including `op-service` itself (especially when we move more RPC utils into it) I think this should be part of `op-service`, not `op-node`. In the long-run I want to remove all circular / oddly structured dependencies, to make addition of new modules cleaner, and avoid unnecessary dependencies on the op-node.

Note: added small op-service fuzzing CI task, split out from the op-node fuzzing CI task.

**Metadata**

Fix CLI-4188


tynes (approved):

refcell (approved):
Changes look good to me 👑

trianglesphere:
This seems to have broken the stateviz build inside the devnet. Indexer tests are maybe flakey.

protolambda:
Stateviz dockerfile wasn't copying in op-service into the build, fixed now.
Not sure what the indexer failure is about, could be a flake maybe.

trianglesphere:
> Not sure what the indexer failure is about, could be a flake maybe.

Yea, it looks more like a flake b/c the tests are building & running. If a change like this causes problems, I'd expect it at that point, not at a later point.

protolambda:
rebased onto develop, hope that fixes the flake

ajsutton (approved):
LGTM.

trianglesphere (approved):

PR #6744, opened 2023-08-11
Only Resolve Won Dispute Games

refcell:
**Description**

Only resolves dispute games which the challenger won.

In order to allow the `Agent` fault component to retrieve
the game status, the `Caller` has to be passed in during
construction by the `service.go` initialization function.

**Tests**

Introduces minimal agent unit tests for a newly exported
`ShouldResolve` method.

**Metadata**

Fixes CLI-4285

### DIFF
--- a/op-challenger/fault/abi_test.go
+++ b/op-challenger/fault/abi_test.go
@@ -4,8 +4,6 @@ import (
 	"math/big"
 	"testing"
 
-	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
-	"github.com/ethereum-optimism/optimism/op-challenger/fault/types"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind/backends"
 	"github.com/ethereum/go-ethereum/common"
@@ -13,6 +11,9 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/stretchr/testify/require"
+
+	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
+	"github.com/ethereum-optimism/optimism/op-challenger/fault/types"
 )
 
 // setupFaultDisputeGame deploys the FaultDisputeGame contract to a simulated backend
@@ -26,7 +27,10 @@ func setupFaultDisputeGame() (common.Address, *bind.TransactOpts, *backends.Simu
 	if err != nil {
 		return common.Address{}, nil, nil, nil, err
 	}
-	backend := backends.NewSimulatedBackend(core.GenesisAlloc{from: {Balance: big.NewInt(params.Ether)}}, 50_000_000)
+	backend := backends.NewSimulatedBackend(
+		core.GenesisAlloc{from: {Balance: big.NewInt(params.Ether)}},
+		50_000_000,
+	)
 
 	blockHashOracle, _, _, err := bindings.DeployBlockOracle(opts, backend)
 	if err != nil {
@@ -55,7 +59,7 @@ func TestBuildFaultDefendData(t *testing.T) {
 	_, opts, _, contract, err := setupFaultDisputeGame()
 	require.NoError(t, err)
 
-	responder, _ := newTestFaultResponder(t, false)
+	responder, _ := newTestFaultResponder(t)
 
 	data, err := responder.buildFaultDefendData(1, [32]byte{0x02, 0x03})
 	require.NoError(t, err)
@@ -72,7 +76,7 @@ func TestBuildFaultAttackData(t *testing.T) {
 	_, opts, _, contract, err := setupFaultDisputeGame()
 	require.NoError(t, err)
 
-	responder, _ := newTestFaultResponder(t, false)
+	responder, _ := newTestFaultResponder(t)
 
 	data, err := responder.buildFaultAttackData(1, [32]byte{0x02, 0x03})
 	require.NoError(t, err)
@@ -89,7 +93,7 @@ func TestBuildFaultStepData(t *testing.T) {
 	_, opts, _, contract, err := setupFaultDisputeGame()
 	require.NoError(t, err)
 
-	responder, _ := newTestFaultResponder(t, false)
+	responder, _ := newTestFaultResponder(t)
 
 	data, err := responder.buildStepTxData(types.StepCallData{
 		ClaimIndex: 2,
--- a/op-challenger/fault/agent.go
+++ b/op-challenger/fault/agent.go
@@ -5,15 +5,16 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/ethereum/go-ethereum/log"
+
 	"github.com/ethereum-optimism/optimism/op-challenger/fault/solver"
 	"github.com/ethereum-optimism/optimism/op-challenger/fault/types"
-	"github.com/ethereum/go-ethereum/log"
 )
 
 // Responder takes a response action & executes.
 // For full op-challenger this means executing the transaction on chain.
 type Responder interface {
-	CanResolve(ctx context.Context) bool
+	CallResolve(ctx context.Context) (uint8, error)
 	Resolve(ctx context.Context) error
 	Respond(ctx context.Context, response types.Claim) error
 	Step(ctx context.Context, stepData types.StepCallData) error
@@ -29,7 +30,15 @@ type Agent struct {
 	log                     log.Logger
 }
 
-func NewAgent(loader Loader, maxDepth int, trace types.TraceProvider, responder Responder, updater types.OracleUpdater, agreeWithProposedOutput bool, log log.Logger) *Agent {
+func NewAgent(
+	loader Loader,
+	maxDepth int,
+	trace types.TraceProvider,
+	responder Responder,
+	updater types.OracleUpdater,
+	agreeWithProposedOutput bool,
+	log log.Logger,
+) *Agent {
 	return &Agent{
 		solver:                  solver.NewSolver(maxDepth, trace),
 		loader:                  loader,
@@ -52,7 +61,8 @@ func (a *Agent) Act(ctx context.Context) error {
 	}
 	// Create counter claims
 	for _, claim := range game.Claims() {
-		if err := a.move(ctx, claim, game); err != nil && !errors.Is(err, types.ErrGameDepthReached) {
+		if err := a.move(ctx, claim, game); err != nil &&
+			!errors.Is(err, types.ErrGameDepthReached) {
 			log.Error("Failed to move", "err", err)
 		}
 	}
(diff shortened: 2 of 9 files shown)


refcell:
* develop
  * **PR #6744** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6744" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6744?utm_source=stack-comment).

op-e2e/e2eutils/disputegame/game_helper.go
```diff
@@ -113,7 +113,7 @@ func (g *FaultGameHelper) WaitForGameStatus(ctx context.Context, expected Status
 	g.t.Logf("Waiting for game %v to have status %v", g.addr, expected)
 	ctx, cancel := context.WithTimeout(ctx, time.Minute)
 	defer cancel()
-	err := utils.WaitFor(ctx, time.Second, func() (bool, error) {
+	err := utils.WaitFor(ctx, 3*time.Second, func() (bool, error) {
```
ajsutton: We shouldn't need to check less often which is what the second option is controlling here - it will just make the test slower.  The timeout is set on the context.

ajsutton:
The contract will always return the game status as `In Progress` because that only changes when resolve is called. I don't think this change is a priority since we expect to run an honest challenger that should always win anyway, but if we're going to implement it we'll probably need to have `resolve` return the resulting status and then check that when we do the test resolve via `eth_Call`.

op-challenger/fault/agent.go
```diff
@@ -65,12 +71,32 @@ func (a *Agent) Act(ctx context.Context) error {
 	return nil
 }
 
+// ShouldResolve returns true if the agent should resolve the game.
+func (a *Agent) ShouldResolve(ctx context.Context) bool {
+	status, err := a.statusFetcher.GetGameStatus(ctx)
```
trianglesphere: Why would the status be `types.GameStatusDefenderWon` or `types.GameStatusChallengerWon`  before the resolve function has been called?

refcell: Another challenger could have resolved the game?

refcell: Ah I see why you're asking - addressing...

refcell:
> The contract will always return the game status as `In Progress` because that only changes when resolve is called. I don't think this change is a priority since we expect to run an honest challenger that should always win anyway, but if we're going to implement it we'll probably need to have `resolve` return the resulting status and then check that when we do the test resolve via `eth_Call`.

Ah my mistake, I goofed how I thought about the game status call. Fixed this to actually simulate the resolve call. If we're more comfortable opportunistically resolving games (concerned that we won't call resolve if this added logic keeps failing) - I can stack a pr ontop of this with a cli flag like `--check-resolve` or something similar that enables this logic.

op-challenger/fault/abi_test.go
```diff
@@ -26,7 +27,10 @@ func setupFaultDisputeGame() (common.Address, *bind.TransactOpts, *backends.Simu
 	if err != nil {
 		return common.Address{}, nil, nil, nil, err
 	}
-	backend := backends.NewSimulatedBackend(core.GenesisAlloc{from: {Balance: big.NewInt(params.Ether)}}, 50_000_000)
+	backend := backends.NewSimulatedBackend(
```
trianglesphere, line 30: Why a diff here?

op-challenger/fault/agent.go
```diff
@@ -29,7 +30,15 @@ type Agent struct {
 	log                     log.Logger
 }
 
-func NewAgent(loader Loader, maxDepth int, trace types.TraceProvider, responder Responder, updater types.OracleUpdater, agreeWithProposedOutput bool, log log.Logger) *Agent {
+func NewAgent(
+	loader Loader,
+	maxDepth int,
```
trianglesphere, line 35: ditto why a diff?

op-challenger/fault/agent.go
```diff
@@ -52,7 +61,8 @@ func (a *Agent) Act(ctx context.Context) error {
 	}
 	// Create counter claims
 	for _, claim := range game.Claims() {
-		if err := a.move(ctx, claim, game); err != nil && !errors.Is(err, types.ErrGameDepthReached) {
+		if err := a.move(ctx, claim, game); err != nil &&
```
trianglesphere, line 64: ditto

op-challenger/fault/agent.go
```diff
@@ -65,10 +75,32 @@ func (a *Agent) Act(ctx context.Context) error {
...
 }
 
+// ShouldResolve returns true if the agent should resolve the game.
+func (a *Agent) ShouldResolve(ctx context.Context, status uint8) bool {
+	expectedStatus := types.GameStatusDefenderWon
+	if a.agreeWithProposedOutput {
+		expectedStatus = types.GameStatusChallengerWon
+	}
+	if expectedStatus != types.GameStatus(status) {
+		a.log.Warn(
```
trianglesphere: this can be a single line or if it needs to be multline could be a max of 2 lines

op-challenger/fault/agent.go
```diff
@@ -65,10 +75,32 @@ func (a *Agent) Act(ctx context.Context) error {
...
+
 // tryResolve resolves the game if it is in a terminal state
 // and returns true if the game resolves successfully.
 func (a *Agent) tryResolve(ctx context.Context) bool {
-	if !a.responder.CanResolve(ctx) {
+	status, err := a.responder.CallResolve(ctx)
+	if err != nil {
+		return false
+	}
+	if status != 0 && !a.ShouldResolve(ctx, status) {
```
trianglesphere: Why isn't this `status != 0` check inside `ShouldResolve`? It seems like `ShouldResovlve(tyep.GameStatusInProgress` should return false.

op-challenger/fault/agent.go
```diff
@@ -106,9 +138,22 @@ func (a *Agent) move(ctx context.Context, claim types.Claim, game types.Game) er
 		return nil
 	}
 	move := *nextMove
-	log := a.log.New("is_defend", move.DefendsParent(), "depth", move.Depth(), "index_at_depth", move.IndexAtDepth(),
-		"value", move.Value, "trace_index", move.TraceIndex(a.maxDepth),
-		"parent_value", claim.Value, "parent_trace_index", claim.TraceIndex(a.maxDepth))
+	log := a.log.New(
```
trianglesphere: ditto. IMO this is a much harder log to read.

op-challenger/fault/agent_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ethereum-optimism/optimism/op-node/testlog"
+)
+
+// TestAgent_ShouldResolve tests the [Agent] resolution logic.
+func TestAgent_ShouldResolve(t *testing.T) {
```
trianglesphere, line 14: Nice tests

op-challenger/fault/responder.go
```diff
@@ -79,18 +89,31 @@ func (r *faultResponder) BuildTx(ctx context.Context, response types.Claim) ([]b
 	}
 }
 
-// CanResolve determines if the resolve function on the fault dispute game contract
-// would succeed. Returns true if the game can be resolved, otherwise false.
-func (r *faultResponder) CanResolve(ctx context.Context) bool {
+// unpackResolveResult unpacks the result of the resolve function.
+func (r *faultResponder) unpackResolveResult(result []byte) (uint8, error) {
```
trianglesphere: Where is this function tested?

op-e2e/faultproof_test.go
```diff
@@ -231,6 +231,10 @@ func TestCannonDisputeGame(t *testing.T) {
 			sys.TimeTravelClock.AdvanceTime(game.GameDuration(ctx))
 			require.NoError(t, wait.ForNextBlock(ctx, l1Client))
 
+			// Manually resolve the game since the "honest actor" is the defender
+			// and won't call resolve.
+			game.Resolve(ctx)
```
trianglesphere, line 236: Nice that we have to manually resolve here. One callout is that I don't think we are testing doing the resolve now (though I could miss where we are doing it).

trianglesphere (commented):
Core functionality looks decent, but to be honest, it's hard to determine what diffs are important and which are not.

op-challenger/fault/abi_test.go
refcell: line-length formatter - if it's important i can revert back

refcell: Actually, I have it setup to format on save which is generally really nice to have. Can we accept these formatting changes?

op-challenger/fault/agent.go
refcell: Agreed, good callout.

op-challenger/fault/abi_test.go
trianglesphere: What length is it set to? Running your own linter that's relatively different than the style of the codebase causes a huge amount of churn.

op-challenger/fault/utils.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type LogBuilder struct {
+	ctx []interface{}
+}
+
+// NewLogBuilder creates a new [logBuilder] instance.
+func NewLogBuilder() *LogBuilder {
```
trianglesphere, line 12: I'm not sure we need this functionality. We can also get a lot of this with `log.New`

op-challenger/fault/responder.go
```diff
@@ -79,18 +89,25 @@ func (r *faultResponder) BuildTx(ctx context.Context, response types.Claim) ([]b
 	}
 }
 
-// CanResolve determines if the resolve function on the fault dispute game contract
-// would succeed. Returns true if the game can be resolved, otherwise false.
-func (r *faultResponder) CanResolve(ctx context.Context) bool {
+// CallResolve determines if the resolve function on the fault dispute game contract
+// would succeed. Returns the game status if the call would succeed, errors otherwise.
+func (r *faultResponder) CallResolve(ctx context.Context) (uint8, error) {
```
ajsutton, line 94: Can we return `GameStatus` here instead of `uint8`? That way the details of the contract representation are better encapsulated and we get more specify typing for callers.

op-challenger/fault/responder.go
```diff
@@ -79,18 +89,25 @@ func (r *faultResponder) BuildTx(ctx context.Context, response types.Claim) ([]b
...
+	res, err := r.txMgr.Call(ctx, ethereum.CallMsg{
 		To:   &r.fdgAddr,
 		Data: txData,
 	}, nil)
-	return err == nil
+	if err != nil {
+		return 0, err
+	}
+	var status uint8
+	if err = r.fdgAbi.UnpackIntoInterface(&status, "resolve", res); err != nil {
```
trianglesphere, line 107: Ok, this is being hit with a cover profile, but where is the test?

trianglesphere (changes requested):

op-challenger/fault/utils.go
```diff
@@ -0,0 +1,24 @@
+package fault
+
+import (
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type LogBuilder struct {
+	ctx []interface{}
+}
```
ajsutton, line 9: This doesn't belong as part of the fault module. I suspect needing it is a sign that we're adding too many fields to a logger which will just create noise in the logs.  If it really needs to exist it should be somewhere in op-service.

op-e2e/faultproof_test.go
ajsutton: This seems wrong to me - I'd have expected the challenger to resolve this game as it should be winning.  The defender is the actor who starts the game and provides the root claim (they are defending the root claim).

ajsutton (commented):
Would definitely be nice to separate the actual changes from the formatting changes. And I feel like many of the new line wrappings are forcing a too restrictive line length.

op-challenger/fault/responder.go
refcell: How would you like me to test it?

trianglesphere: I'm asking the location of the test code which tests this. I'm open to lots of ways of testing this, but I'm not seeing an obvious test which tests this. It could be that I'm just missing where the test is.

refcell:
Closing this out with fixes in PR #6956

PR #6701, opened 2023-08-10
feat: Add EIP1271 support to EAS

lbeder:
This PR adds EIP271 support to EAS

PR #6747, opened 2023-08-11
feat: eas with eip1271 support

tynes:
**Description**

Replacement of https://github.com/ethereum-optimism/optimism/pull/6701
Also adds in the EAS contracts to the semver lock and then rebuilds the bindings


packages/contracts-bedrock/src/EAS/SchemaRegistry.sol
```diff
@@ -20,6 +20,7 @@ contract SchemaRegistry is ISchemaRegistry, Semver {
     uint256[MAX_GAP - 1] private __gap;
 
     /// @dev Creates a new SchemaRegistry instance.
+    /// @custom:semver 1.0.2
     constructor() Semver(1, 0, 2) { }
```
refcell: Should we bump the semver patch version here technically?

packages/contracts-bedrock/src/EAS/eip1271/EIP1271Verifier.sol
```diff
@@ -98,26 +101,42 @@ abstract contract EIP712Verifier is EIP712 {
...
 
         bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(REVOKE_TYPEHASH, request.schema, data.uid, nonce)));
+        _verifySignature(digest, signature, request.revoker);
+    }
+
+    /// @notice Verifies EIP712 signatures (with EIP1271 support).
+    /// @param digest The typed-data digest to verify.
+    /// @param signature The signature to verify (either a "real" ECDSA signature or an EIP1271-aware signature).
+    /// @param signer The signer to verify the signature against.
+    function _verifySignature(bytes32 digest, Signature memory signature, address signer) private view {
```
refcell: If we're going to be using openzeppelin's libraries, can we use their signature validation libraries as well as opposed to rolling our own?
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/utils/cryptography/SignatureChecker.sol#L14

packages/contracts-bedrock/src/EAS/SchemaRegistry.sol
tynes: Yes you are right about the semver patch version, good catch

packages/contracts-bedrock/src/EAS/eip1271/EIP1271Verifier.sol
tynes: cc @lbeder

lbeder: Yeah, we can use OZ's `SignatureChecker`, but it will require to pack the signature into `bytes`. Please let me know if this is preferred and I'll update it.

tynes: I want to keep this in sync with your upstream as much as possible

lbeder: Sure thing. Let me address this in our upstream and then I'll open a PR to this PR 👍

tynes: Ok thank you! Super appreciate it :D

lbeder: I've opened https://github.com/ethereum-optimism/optimism/pull/6750 (which was ported and tested from https://github.com/ethereum-attestation-service/eas-contracts/pull/81)

refcell: Thank you so much!

lbeder: 🤝

packages/contracts-bedrock/src/EAS/EAS.sol
```diff
@@ -80,7 +80,8 @@ contract EAS is IEAS, Semver, EIP712Verifier {
     uint256[MAX_GAP - 3] private __gap;
 
     /// @dev Creates a new EAS instance.
-    constructor() Semver(1, 0, 2) EIP712Verifier("EAS", "1.0.1") { }
+    /// @custom:semver 1.2.0
+    constructor() Semver(1, 2, 0) EIP1271Verifier("EAS", "1.0.1") { }
```
tynes: This should be 1.1.0

lbeder: Bumping this version will affect the EIP712 domain, thus invalidate pending EIP712 delegated attestations. IMHO, it's ok (there should not be many of those + they are easily invalidated/expire anyway). I'll bump it regardless, but FYI.

lbeder: Please see https://github.com/ethereum-optimism/optimism/pull/6764

tynes: What is your reasoning with bumping the EIP712 version? The semver and the EIP712 version could be decoupled. Our policy is that bytecode changes are minor bumps and different security models are major bumps. I was thinking that we just leave the EIP712 version as is for simplicity and then only bump it in the future if we explicitly want to invalidate signatures

lbeder: Oh, my bad. I thought that you meant to bump the EIP712 version, since the `1.2.0` change was made later by you directly. In this case, please ignore my PR and let's change the version to `1.1.0` here, while considering leaving the EIP712 version as `1.0.1` (even though, bumping it would be de-facto pretty harmless at this stage as well).

tynes: Ah apologies, i was not clear in my comment, I just left it for myself because i accidentally bumped it twice, will push a fix

lbeder: Ah, gotcha. Better safe than sorry! :D

PR #6748, opened 2023-08-11
deploy-config: use recent L1 timestamp for L2OO

tynes:
**Description**

By setting the value to `-1`, the deploy script
will fetch a recent timestamp from L1 to use as
the L2 Output Oracle's starting timestamp.

Also updates the batcher and proposer keys


trianglesphere (approved):

twoshark (approved):
The new addresses match what I generated

trianglesphere (approved):

PR #6751, opened 2023-08-11
contracts-bedrock: slight cleanup + docs

tynes:
**Description**

Makes it easier to do deployments with a better abstraction. The first step of a deployment is always to do the deployment, then run `sync()` to parse the forge deploy artifact into a hardhat style artifact. This commit improves the UX of `sync()`


refcell (approved):

tynes:
ty for fix @refcell <3

PR #6752, opened 2023-08-11
Restructure Challenger

refcell:
**Description**

Restructures the `op-challenger` in preparation for
the tower defense milestone.

This modularizes fault components into legible
and isolated packages.

Further work and cleaning will be needed, but this
is a small step towards structuring the challenger
in an extensible, digestible way.


refcell:
* develop
  * **PR #6752** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6752" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6752?utm_source=stack-comment).

op-challenger/challenger.go
```diff
@@ -5,13 +5,13 @@ import (
 	"fmt"
 
 	"github.com/ethereum-optimism/optimism/op-challenger/config"
-	"github.com/ethereum-optimism/optimism/op-challenger/fault"
+	"github.com/ethereum-optimism/optimism/op-challenger/fault/utils"
```
ajsutton, line 8: `utils` is generally a bad module name because its too generic and winds up being a dumping ground for stuff that just didn't get categorised properly.

ajsutton (commented):
I'm not sure this has really moved us towards having a clear entry point for a single game and separating out the single-game vs shared multi-game code. The new modules out necessarily bad, but they are feeling a bit too fine grained for my preference.

I think what we need is more to focus on separating things that are dependent on `cfg.GameAddress` which will be game-specific and things that aren't dependent on that which are then our shared components. We won't have many shared components yet but things like the L1 client should be shared and the top level service.

We also currently access `config.GameAddress` in quite a wide range of places but want to move away from that and have the game address specifically passed in to each of the game-specific components. `config.GameAddress` should only then be accessed by the shared initial setup code and ultimately we'll replace it with the game loader component.

The aim isn't just refactoring to tidy up, it's to refactor the code into a shape that makes it easy to add multi-game support (and obvious where that functionality should be plugged in). Might be more than one PR to get there obviously.

refcell:
Closing for now - this should come after tower defense anyways. Premature optimization on my part

PR #6753, opened 2023-08-11
contracts-bedrock: fix system config initialize

tynes:
**Description**

Every sentinel contract should have null values in their storage to make them unusable. Sometimes there are constraints on what these values can be in the initializer and they cannot be exactly 0 but must be close enough to make the contract useless. This commit updates the startBlock number so that it is not 0 because if it is 0 then it will be set to `block.number` which is not what we want.


clabby (dismissed):
LGTM - needs rebase on @maurelian's lockfile changes.

refcell (approved):

tynes:
> LGTM - needs rebase on @maurelian's lockfile changes.

That PR is broken, this is already rebased on it

tynes:
this needs a minor version bump

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -118,7 +118,7 @@ contract SystemConfig is OwnableUpgradeable, Semver {
                 systemTxMaxGas: 0,
                 maximumBaseFee: 0
             }),
-            _startBlock: 0,
+            _startBlock: 1,
```
Inphi: maybe something like `MAX_UINT` would be a better placeholder. Since the system will eventually reach block 1?

tynes: Thats a great idea

Inphi (approved):

PR #6754, opened 2023-08-11
contracts-bedrock: fix pnpm clean

tynes:
**Description**

Accidentally was deleting the source instead
of the binary. Fixes it so it deletes the binary
instead.


refcell (approved):
Good patch, thanks

PR #6755, opened 2023-08-11
GameLoader Fault Component

refcell:
**Description**

In order for the challenger to automatically detect
and engage with fault dispute games, it needs to be
able to fetch all dispute games from the
DisputeGameFactory contract.

This PR introduces a new fault component game loader
that calls the DisputeGameFactory contract for each
of it's games.

Note: This is a naive component that can be
enhanced to be much more efficient.

It currently:
- queries games iteratively to the caller. (ie
it does not use rpc batching).
- it does not return games by their status (in
progress), or keep track of completed games and
their respective indices.
- it does not filter games by their game type (
eg, for the fault component, we'd only want
to query for games of the type Fault Dispute
Game)

**Tests**

Table tests are performed using a stubbed caller.

**Metadata**

Fixes CLI-4340.


refcell:
* develop
  * **PR #6755** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6755" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6755?utm_source=stack-comment).

op-challenger/fault/factory.go
```diff
@@ -0,0 +1,64 @@
...
+func NewGameLoader(caller MinimalDisputeGameFactoryCaller) *gameLoader {
+	return &gameLoader{
+		caller: caller,
+	}
+}
+
+// FetchAllGames fetches all dispute games from the factory.
+// todo(refcell): batch call this to give mr. rpc a break.
+func (l *gameLoader) FetchAllGames(ctx context.Context) ([]FaultDisputeGame, error) {
+	gameCount, err := l.caller.GameCount(&bind.CallOpts{Context: ctx})
```
ajsutton: We need to have a consistent view of the chain when performing multiple calls so shouldn't just use the latest head in every call.

op-challenger/fault/factory.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	games := make([]FaultDisputeGame, gameCount.Uint64())
+	for i := uint64(0); i < gameCount.Uint64(); i++ {
+		game, err := l.caller.GameAtIndex(&bind.CallOpts{Context: ctx}, big.NewInt(int64(i)))
+		if err != nil {
+			return nil, err
```
ajsutton: We should add context to this error before returning it - otherwise we'll just get an obscure error from JSON-RPC logged and not know where it came from.

op-challenger/fault/factory_test.go
```diff
@@ -0,0 +1,189 @@
...
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			loader := NewGameLoader(test.caller)
+
+			games, err := loader.FetchAllGames(context.Background())
+			require.Equal(t, test.expectedErr, err)
```
ajsutton: `require.ErrorIs` would work better here so it can handle the error being wrapped with extra context.

op-challenger/fault/factory.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		games[i] = FaultDisputeGame{
+			Proxy:     game.Proxy,
+			Timestamp: game.Timestamp,
+		}
```
ajsutton: There's no need to copy here, the two structs have the same data so you can just use:
```suggestion
		games[i] = game
```

The trade off is that if there are new fields added to the bindings we'll need to add them to our struct, but I think it's better to get a clear compiler error in that case rather than just silently dropping the field here.

op-challenger/fault/factory_test.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+}
+
+func generateMockGames(count uint64) []struct {
+	proxy     common.Address
+	timestamp *big.Int
+} {
+	games := make([]struct {
+		proxy     common.Address
+		timestamp *big.Int
```
ajsutton: You could simplify this test a lot by using `FaultDisputeGame` type instead of redeclaring the `struct` everywhere except the return type for the mock's `GameAtIndex` return value. Avoids a heap of duplication and copying etc.

op-challenger/fault/factory.go
```diff
@@ -0,0 +1,87 @@
...
+		BlockNumber: l.blockNumber,
+	}
+}
+
+// FetchAllGames fetches all dispute games from the factory.
+// todo(refcell): batch call this to give mr. rpc a break.
+func (l *gameLoader) FetchAllGames(ctx context.Context) ([]FaultDisputeGame, error) {
+	if l.blockNumber == nil {
+		return nil, ErrMissingBlockNumber
+	}
```
ajsutton: Why wouldn't this just be passed in as a method param?

ajsutton (approved):

PR #6762, opened 2023-08-12
contracts-bedrock: goerli deploy config

tynes:
**Description**

Deploy the new implementations to Goerli.

Update the deploy config to be ready to be used to initialize the contracts. Set the system config start block to the height of the initial L1 origin. This value can be used by the `op-node` to know where to start searching for logs from. This value comes from the `op-node` goerli chain config. It allows us to remove information from the rollup config, simplifying management of chains.


Inphi (commented):
The goerli deployment contract addresses have changed! Adding new fields to the deployment config shouldn't cause this.

Inphi:
Oh these are new implementations? Why are we doing this again?

refcell:
+1 to @Inphi's Q. Also, how can we best verify this?

Inphi (approved):

PR #6763, opened 2023-08-12
cannon: Fix j/jal VM emulation

Inphi:
Fixes two bugs:
- Jump offsets should be zero-extended
- Jump offsets are relative to the PC-region aligned to 256 MB. The PC-region is encoded in the topmost bits of the destination effective address.


cannon/mipsevm/mips.go
```diff
@@ -285,13 +285,13 @@ func (m *InstrumentedState) mipsStep() error {
...
-		// TODO likely bug in original code: MIPS spec says this should be in the "current" region;
-		// a 256 MB aligned region (i.e. use top 4 bits of branch delay slot (pc+4))
 		linkReg := uint32(0)
 		if opcode == 3 {
 			linkReg = 31
 		}
-		return m.handleJump(linkReg, SE(insn&0x03FFFFFF, 26)<<2)
+		// Take top 4 bits of current PC (its 256 MB region), and concatenate with 26-bit offset
+		target := (m.state.NextPC & 0xF0000000) | ((insn & 0x03FFFFFF) << 2)
+		return m.handleJump(linkReg, target)
```
protolambda: Most emulators actually get this wrong, or completely ignore the top 4 bits, but I believe we got it correct here now. It shouldn't be the 4 bits of the current PC, but the 4 bits of the next PC (delay slot), or else the "programming notes" in the spec wouldn't hold true. The spec expression of this operation is quite unclear about it though.

> This definition creates the following boundary case: When the branch instruction is in the last word of a 256MB
region, it can branch only to the following 256MB region containing the branch delay slot.

So because it's using next-PC, the top 4 bits will match that of the next 256MB (`2**28`, the remaining bits) region.

And it's also correctly not doing sign-extension (`SE`) anymore, as the spec says "PC-region branch (not PC-relative);".

packages/contracts-bedrock/src/cannon/MIPS.sol
```diff
@@ -667,9 +667,9 @@ contract MIPS {
 
             // j-type j/jal
             if (opcode == 2 || opcode == 3) {
-                // TODO(CLI-4136): likely bug in original code: MIPS spec says this should be in the "current" region;
-                // a 256 MB aligned region (i.e. use top 4 bits of branch delay slot (pc+4))
-                return handleJump(opcode == 2 ? 0 : 31, SE(insn & 0x03FFFFFF, 26) << 2);
+                // Take top 4 bits of current PC (its 256 MB region), and concatenate with 26-bit offset
+                uint32 target = (state.nextPC & 0xF0000000) | (insn & 0x03FFFFFF) << 2;
+                return handleJump(opcode == 2 ? 0 : 31, target);
```
protolambda: Thank you for fixing the TODO I left here :+1:

protolambda (approved):

cannon/mipsevm/mips.go
Inphi: yup, meant to say nextPC in the comments based on that bit in the spec. Will fix the comments so it's clearer.

clabby (approved):

PR #6766, opened 2023-08-12
feat: use deterministic deployment for token factories

ncitron:
**Description**

Use CREATE2 for mintable token deployments.

**Tests**

I needed to change the test for double deployments. Since it is deterministic now we will expect a second creation for the same remote token, symbol, and name combination to fail.

**Context**
https://warpcast.com/tynes/0x5b3dd3

@tynes


roninjin10 (approved):
Very nice change! I wish we would have planted this tree a long time ago instead of just now. LGTM but will need contract folks to review.

roninjin10:
@ncitron need to fix linter btw

roninjin10:
Need to commit new gas snapshots and also update the semver modifier on the contracts

tynes:
This should not create any issues with address collisions, there are 160 bits of security for somebody trying to create a collision with a legacy token address with the create2 style address. Its a feature that only 1 token can be created per input. I cannot think of anything unsafe about this PR, will just need to make sure that offchain tooling is aware of this functionality. Its going to be near impossible to get tokens to migrate, but at least this is good for new chains

packages/contracts-bedrock/src/universal/OptimismMintableERC20Factory.sol
```diff
@@ -70,7 +70,8 @@ contract OptimismMintableERC20Factory is Semver {
     {
```
tynes, line 70: We are going to want to bump the semver to `1.2.0`

packages/contracts-bedrock/src/universal/OptimismMintableERC721Factory.sol
```diff
@@ -47,7 +47,9 @@ contract OptimismMintableERC721Factory is Semver {
     {
         require(_remoteToken != address(0), "OptimismMintableERC721Factory: L1 token address cannot be address(0)");
```
tynes, line 48: We are going to want to bump the semver to `1.3.0`

tynes:
We should also bump the semver on the token implementations themselves, so that offchain tooling can know they were deployed with create2

tynes:
@refcell will help get this merged

refcell:
@ncitron Opened up a pr into your branch in your repo: https://github.com/ncitron/optimism/pull/1

This should fix the semver bumps and gas snapshots. Then we can very the bindings :)

refcell:
Hey @ncitron - sorry I forgot to update the semver lock. Addressed this and bindings in a follow-on pr https://github.com/ncitron/optimism/pull/2

Hopefully this should fix the ci errors

refcell:
Ok - one more @ncitron, I hope. https://github.com/ncitron/optimism/pull/3

refcell:
That's odd - regened the semver lock and removed the trailing newline in https://github.com/ncitron/optimism/pull/4. Thank you for bearing with me here @ncitron

tynes (approved):

tynes:
Looks good to me!

refcell (approved):

PR #6770, opened 2023-08-13
Fix minor logging bug in E2E test

mdehoog:
**Description**

Noticed a bug in Base's `withdrawer` utility, fixed here: https://github.com/base-org/withdrawer/commit/72d9fe94c317010f27257ba31e494c378c3f5eef.

The number reported for the finalization period was using the incorrect variable.

This code was copied from Optimism's E2E tests; this PR upstreams the fix.

**Tests**

It is tests.


protolambda (approved):
nice find

PR #6768, opened 2023-08-12
feat(indexer): Implement cursor based pagination

roninjin10:
- Implement pagination in API

PR #6767, opened 2023-08-12
feat(indexer): Preallocate space for a slice

roninjin10:
Optimize mapping of db data to api response by preallocating a slice since we know it's length ahead of time

PR #6738, opened 2023-08-11
feat: Implement API with dummy data

roninjin10:
- implement api with dummy data

PR #6726, opened 2023-08-11
feat(indexer): Hook up api to cli app

roninjin10:
This is a really old pr that never landed

PR #6771, opened 2023-08-13
fix(indexer): Fix config usage make config deps more explicit

roninjin10:
- remove literal logger from config replace with logger config
  - Inferred that this was not how it worked before because we actually need a logger before we even read the config
  - Handle this by initializing a logger with reasonable defaults first
- dependency inject the db into indexer to make the dependencies more clear
- replace using the entire config object with only the pieces of the config a service needs, this makes it more explicit
- Move l1Contracts to the config out of the indexer package
  - this blocked a future pr because it became too easy to make a circular dependency
  - the config package should rely on no other indexer code as it's the code everything else consumes to "configure it"

### DIFF
--- a/indexer/api/routes/deposits.go
+++ b/indexer/api/routes/deposits.go
@@ -30,7 +30,7 @@ type DepositResponse struct {
 // TODO this is original spec but maybe include the l2 block info too for the relayed tx
 // FIXME make a pure function that returns a struct instead of newWithdrawalResponse
 func newDepositResponse(deposits []*database.L1BridgeDepositWithTransactionHashes) DepositResponse {
-	var items []DepositItem
+	items := make([]DepositItem, len(deposits))
 	for _, deposit := range deposits {
 		item := DepositItem{
 			Guid: deposit.L1BridgeDeposit.TransactionSourceHash.String(),
--- a/indexer/api/routes/withdrawals.go
+++ b/indexer/api/routes/withdrawals.go
@@ -43,7 +43,7 @@ type WithdrawalResponse struct {
 
 // FIXME make a pure function that returns a struct instead of newWithdrawalResponse
 func newWithdrawalResponse(withdrawals []*database.L2BridgeWithdrawalWithTransactionHashes) WithdrawalResponse {
-	var items []WithdrawalItem
+	items := make([]WithdrawalItem, len(withdrawals))
 	for _, withdrawal := range withdrawals {
 		item := WithdrawalItem{
 			Guid: withdrawal.L2BridgeWithdrawal.TransactionWithdrawalHash.String(),


roninjin10:
* develop
  * **PR #6738** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6738" target="_blank">Graphite</a> 
    * **PR #6767** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6767" target="_blank">Graphite</a> 
      * **PR #6773** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6773" target="_blank">Graphite</a> 
        * **PR #6768** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6768" target="_blank">Graphite</a> 
          * **PR #6726** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6726" target="_blank">Graphite</a> 
            * **PR #6799** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6799" target="_blank">Graphite</a> 
      * **PR #6771** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6771" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6771?utm_source=stack-comment).

indexer/cli/cli.go
```diff
@@ -23,7 +24,11 @@ type Cli struct {
 }
 
 func runIndexer(ctx *cli.Context) error {
-	logger := log.NewLogger(log.ReadCLIConfig(ctx))
+	logger := log.NewLogger(log.CLIConfig{
+		Level:  "warn",
+		Color:  false,
+		Format: "terminal",
+	})
```
roninjin10: We need to log if loading the config (which has the logger config in it) fails so provide a logger with normal defaults. Is terminal the correct choice or should I use TEXT?

indexer/cli/cli.go
```diff
@@ -32,8 +37,15 @@ func runIndexer(ctx *cli.Context) error {
 		return err
 	}
 
-	cfg.Logger = logger
-	indexer, err := indexer.NewIndexer(cfg)
+	logger = log.NewLogger(cfg.Logger)
+
+	db, err := database.NewDB(cfg.DB)
```
roninjin10: Dependency injecting the DB. Doesn't make a difference but is slightly cleaner and makes the configs more explicit imo (obvious that only DB is using this config)

indexer/cli/cli.go
```diff
@@ -32,8 +37,15 @@ func runIndexer(ctx *cli.Context) error {
...
+
+	if err != nil {
+		return err
+	}
+
+	indexer, err := indexer.NewIndexer(cfg.Chain, cfg.RPCs, db, logger)
```
roninjin10: Being explicit with config instead of passing in the entire thing

indexer/config/config.go
```diff
@@ -16,14 +17,41 @@ type Config struct {
...
+	L1CrossDomainMessenger common.Address
+	L1StandardBridge       common.Address
+	L1ERC721Bridge         common.Address
+
+	// Some more contracts -- ProxyAdmin, SystemConfig, etcc
+	// Ignore the auxiliary contracts?
+
+	// Legacy contracts? We'll add this in to index the legacy chain.
+	// Remove afterwards?
+}
```
roninjin10: Moved this from processor to avoid any circular imports that might get ran into

indexer/config/config.go
```diff
@@ -16,14 +17,41 @@ type Config struct {
...
+
+func (c L1Contracts) ToSlice() []common.Address {
+	fields := reflect.VisibleFields(reflect.TypeOf(c))
+	v := reflect.ValueOf(c)
+
+	contracts := make([]common.Address, len(fields))
+	for i, field := range fields {
+		contracts[i] = (v.FieldByName(field.Name).Interface()).(common.Address)
```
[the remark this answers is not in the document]
roninjin10: Indexer takes no user input so this is fine

indexer/database/db.go
```diff
@@ -17,7 +20,14 @@ type DB struct {
...
 
-func NewDB(dsn string) (*DB, error) {
+func NewDB(dbConfig config.DBConfig) (*DB, error) {
+	dsn := fmt.Sprintf("host=%s port=%d dbname=%s sslmode=disable", dbConfig.Host, dbConfig.Port, dbConfig.Name)
+	if dbConfig.User != "" {
+		dsn += fmt.Sprintf(" user=%s", dbConfig.User)
+	}
+	if dbConfig.Password != "" {
+		dsn += fmt.Sprintf(" password=%s", dbConfig.Password)
+	}
```
roninjin10: Seems slightly cleaner to make the dsn here but think it's reasonable to do it in CLI too

indexer/cli/cli.go
hamdiallam: why are we removing `log.ReadCLIConfig`? imo log configuration is the one thing that we should be reading from flags and configuration files

hamdiallam: Initially was thinking of not having it be flags when i was implementing this but realized log configuration via flags is the most idiomatic way to starting a process

indexer/cli/cli.go
hamdiallam: nit: logger being the first argument

hamdiallam (approved):
LGTM. Although I think we should revert the logger change as I think it's easier to have changes to log settings simply be flags to the process

roninjin10:
Will update logger in future pr (cleanup phase)

indexer/cli/cli.go
roninjin10:
```suggestion
    // revert me to only reading logger from cli flags and remove from toml configuration
	logger := log.NewLogger(log.CLIConfig{
		Level:  "warn",
		Color:  false,
		Format: "terminal",
	})
```

roninjin10: Makes sense agree

indexer/cli/cli.go
roninjin10:
```suggestion
    // fixme logger should be first to stay consistent
	indexer, err := indexer.NewIndexer(cfg.Chain, cfg.RPCs, db, logger)
```
Leaving cleanup phase comment to fix since I bet I made this mistake elsewhere too

roninjin10:
Depends on other prs enough to not want to rebase

PR #6769, opened 2023-08-12
feat(indexer): Add tokens to the db schema

roninjin10:
- Indexer needs to index tokens too

PR #6772, opened 2023-08-13
feat(indexer): Start adding ui to indexer docker-compose

roninjin10:
- Currently we use go directly to run indexer not docker-compose
- We spin up postgres in circleci rather than using docker-compose
- Use the docker-compose as a way to run the UI and backend
  - Will be useful if ecopod reports a bug we can instantly spin up the entire stack and investigate
- Docker compose is currently broken will fix on future pr this pr only adds UI and UI's backend services


roninjin10:
* develop
  * **PR #6775** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6775" target="_blank">Graphite</a> 
    * **PR #6769** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6769" target="_blank">Graphite</a> 
      * **PR #6772** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6772" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6772?utm_source=stack-comment).

indexer/docker-compose.yml
```diff
@@ -75,5 +75,121 @@ services:
       postgres:
         condition: service_healthy
 
+  gateway-frontend:
```
roninjin10, line 101: note: this docker compose has never worked (prs to make it work never landed and we don't use docker-compose locally or in e2e tests atm).

Side note: would be good to use docker-compose on circleci instead of manually setting up postgres since it makes running the tests way easier and 1:1 with CI. Not important will do this change if it ever becomes useful enough to do

In future I will make this entire thing work e2e via building the gateway locally making all the changes the gateway needs and here to get everything running. This just sets up being able to start that process

hamdiallam, line 101: Since the gateway isn't open source yet, it is appropriate it add it to the compose file? I'd imagine this docker configuration would exist in the gateway repo vs the indexer

roninjin10: The docker images are published publicly. THe docker configuration needs to exist in both ideally for our own convenience. We could consider removing it once everything is running but in meantime seperating them needlessly slows us down

hamdiallam (approved):

PR #6775, opened 2023-08-14
feat(indexer): Implement presets for toml file

roninjin10:
- Implement presets for toml file for known chains like base base goerli op op goerli zora and zora goerli
- Add full test suite since this code is ready to go outside of possibly adding more contracts

### DIFF
--- a/indexer/config/config.go
+++ b/indexer/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"os"
 	"reflect"
 
@@ -11,6 +12,8 @@ import (
 	geth_log "github.com/ethereum/go-ethereum/log"
 	"github.com/joho/godotenv"
 )
+
+// in future presets can just be onchain config and fetched on initialization
 
 // Config represents the `indexer.toml` file used to configure the indexer
 type Config struct {
@@ -52,7 +55,8 @@ func (c L1Contracts) ToSlice() []common.Address {
 // ChainConfig configures of the chain being indexed
 type ChainConfig struct {
 	// Configure known chains with the l2 chain id
-	Preset      int
+	Preset int
+	// Configure custom chains via providing the L1Contract addresses
 	L1Contracts L1Contracts
 }
 
@@ -94,19 +98,25 @@ func LoadConfig(logger geth_log.Logger, path string) (Config, error) {
 
 	var conf Config
 
-	// Read the config file.
 	data, err := os.ReadFile(path)
 	if err != nil {
 		return conf, err
 	}
 
-	// Replace environment variables.
 	data = []byte(os.ExpandEnv(string(data)))
 
-	// Decode the TOML data.
 	if _, err := toml.Decode(string(data), &conf); err != nil {
 		logger.Info("Failed to decode config file", "message", err)
 		return conf, err
+	}
+
+	if conf.Chain.Preset != 0 {
+		knownContracts, ok := presetL1Contracts[conf.Chain.Preset]
+		if ok {
+			conf.Chain.L1Contracts = knownContracts
+		} else {
+			return conf, fmt.Errorf("unknown preset: %d", conf.Chain.Preset)
+		}
 	}
 
 	logger.Debug("Loaded config file", conf)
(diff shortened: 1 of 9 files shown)


roninjin10:
* develop
  * **PR #6775** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6775" target="_blank">Graphite</a>  👈
    * **PR #6769** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6769" target="_blank">Graphite</a> 
      * **PR #6772** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6772" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6775?utm_source=stack-comment).

indexer/config/presets.go
```diff
@@ -0,0 +1,63 @@
...
+		L1ERC721Bridge:         common.HexToAddress("0x8DD330DdE8D9898d43b4dc840Da27A07dF91b3c9"),
+	},
+	// Base Mainnet
+	8453: {
+		OptimismPortal:         common.HexToAddress("0x49048044D57e1C92A77f79988d21Fa8fAF74E97e"),
+		L2OutputOracle:         common.HexToAddress("0x56315b90c40730925ec5485cf004d835058518A0"),
+		L1CrossDomainMessenger: common.HexToAddress("0x866E82a600A1414e583f7F13623F1aC5d58b0Afa"),
+		L1StandardBridge:       common.HexToAddress("0x3154Cf16ccdb4C6d922629664174b904d80F2C35"),
+		// FIXME update this to the correct address
+		L1ERC721Bridge: common.HexToAddress("0x0000000000000000000000000000000000000000"),
```
roninjin10, line 34: Gonna have to find these via this other_bridge property. Since we aren't using anything other than OP atm I skipped for now 
image

roninjin10: Could consider not specifying it and just index it but seems pointless to add that complexity when onchain/onrpc config is coming

indexer/config/config.go
```diff
@@ -85,19 +89,25 @@ type MetricsConfig struct {
...
 
-	// Replace environment variables.
 	data = []byte(os.ExpandEnv(string(data)))
 
-	// Decode the TOML data.
 	if _, err := toml.Decode(string(data), &conf); err != nil {
 		return conf, err
 	}
 
+	if conf.Chain.Preset != 0 {
```
roninjin10, line 113: The toml library seems to set everything to 0 if it isn't provided. I looked in the docs for a way to make it explicitly nil but couldn't (quickly) find one. This passes all the tests this implementation though.

Would be nice if we could throw an error when required stuff is not provided fully off of the type. This will throw an error if a unknown config option is passed in though. Maybe worth noting in a github issue that the toml is defaulting a lot of stuff we would prefer to not default

roninjin10: Maybe we should make a ticket to provide extra validation to fail early here? We could 
- check all the contract addresses exist on the chain on startup
- - Check that the rpc is up
- - Check that every field is uo

indexer/config/presets.go
hamdiallam: I'm also open to simply removing this contract address from the list until we get around to supporting the 721 bridge

indexer/config/config.go
hamdiallam: what happens if you make the field a pointer to an int? Will it initialize it to `nil` if it's not present?

indexer/indexer.toml
```diff
@@ -20,3 +22,11 @@ port = 8080
 host = "127.0.0.1"
 port = 7300
 
+[logger]
```
hamdiallam, line 25: I think we're going with removing this from config? ffti if you're going to make undo change in a different PR

hamdiallam (approved):
LGTM! Left a comment about how you might get around toml deserializattion for values not present

indexer/indexer.toml
roninjin10: Yea will remove

indexer/config/config.go
roninjin10: will give it a try

PR #6776, opened 2023-08-14
op-challenger: Add e2e test that disputes an output root by agreeing 

ajsutton:
**Description**

Tests the corner case where both sides have the same trace that says the output root is valid but the malicious user has disputed the original output root anyway.

This could potentially be considered out of scope for the alpha given that adding the output root bisection game changes the options here, but worth capturing this test so we have the infrastructure code it provides and can easily reproduce the situation. The test is marked as skipped since it won't pass until we agree on and implement a fix for this case.

**Metadata**

- https://linear.app/optimism/issue/CLI-4350/handle-output-roots-being-disputed-by-providing-a-cannon-trace-that
- Fixes https://linear.app/optimism/issue/CLI-4306/cannon-e2e-tests


refcell (approved):
LGTM

trianglesphere:
Needs updating b/c of conflict

op-e2e/faultproof_test.go
```diff
@@ -290,6 +290,54 @@ func TestCannonDefendStep(t *testing.T) {
...
+
+	maxDepth := game.MaxDepth(ctx)
+	for claimCount := int64(1); claimCount < maxDepth; {
+		game.LogGameData(ctx)
+		claimCount++
+		// Wait for the challenger to counter
+		game.WaitForClaimCount(ctx, claimCount)
+
+		// Defend everything because we have the same trace as the honest proposer
+		correctTrace.Defend(ctx, claimCount-1)
```
trianglesphere, line 324: When running this test locally, it seems to be stuck on waiting for claim count 2. It is failing with a timeout in that step which as far as I can tell, not what should be happening.

ajsutton: That's exactly what we expect would happen at this point. The challenger agrees with the root claim so doesn't want to attack, but you can't defend the root claim either so it does nothing.

trianglesphere (approved):

PR #6784, opened 2023-08-14
op-chain-ops: tx-builder safe types

tynes:
**Description**

Implement the types for working with the Safe
tx-builder application. These types should be able to be round trip serialized into JSON. These types will be used to create Safe tx-builder bundles
that can be dropped into the UI or can be combined with additional tooling to send the transactions
to the Safe backend directly or directly interact
with the safe contracts.


op-chain-ops/safe/batch_file.go
```diff
@@ -0,0 +1,132 @@
...
+type BatchFile struct {
+	Version      string             `json:"version"`
+	ChainID      *big.Int           `json:"chainId"`
+	CreatedAt    uint64             `json:"createdAt"`
+	Meta         BatchFileMeta      `json:"meta"`
+	Transactions []BatchTransaction `json:"transactions"`
+}
+
+// bathcFileMarshaling is a helper type used for JSON marshaling.
+type batchFileMarshaling struct {
```
trianglesphere, line 25: why do you need the `Marshaling` types? they appear to be exactly the same as the primary type.

trianglesphere (commented):
One duplication question. The rest of the code seems fine, but I'm not familiar with Safe smart contract wallets.

op-chain-ops/safe/batch_file.go
tynes: The types are not exactly the same, if you see the string vs big.int. We could get rid of the types and use anon structs in the marshal methods but opted to not do that, geth uses this pattern but they autogen the marshal types

trianglesphere: Why does marshalling the big int not work natively?

trianglesphere: heh, so you do need a different type to marshall it - could be custom BigInt or just do the string. Make sense.

trianglesphere (approved):

PR #6793, opened 2023-08-15
Updated Indexer Documentation & Added Minor Code Comments

ethenotethan:
**Description**
Updated portions of the `/indexer` documentation to ensure better readability and usability by other OP Stack chain operators. More specifically:
* Updated `indexer/README.md` with relevant testing and env config information
* Added minor code comments around L1 system config representation
* Updated `L1Contracts ` struct with `toml` flags
* Updated `indexer.toml` file to store blank L1 system contract addresses

 

**Tests**
No real logical changes were made to application software that require testing.

**Additional context**

**Metadata**

- Fixes N/A


hamdiallam (approved):
Hell yea, this looks great! Conflicting with some changes @roninjin10 recently got in that pulls out the contract definitions from the processor package and also checks for presets

indexer/processor/l1_processor.go
```diff
@@ -23,14 +23,17 @@ import (
...
-	OptimismPortal         common.Address
-	L2OutputOracle         common.Address
-	L1CrossDomainMessenger common.Address
-	L1StandardBridge       common.Address
-	L1ERC721Bridge         common.Address
+	OptimismPortal         common.Address `toml:"optimism-portal"`
+	L2OutputOracle         common.Address `toml:"l2-output-oracle"`
+	L1CrossDomainMessenger common.Address `toml:"l1-cross-domain-messenger"`
+	L1StandardBridge       common.Address `toml:"l1-standard-bridge"`
+	L1ERC721Bridge         common.Address `toml:"l1-erc721-bridge"`
```
roninjin10: The definition for this L1Contracts struct is now in config/config.go

roninjin10 (changes requested):
Just have to do same change to the contracts def in config/config.go. We should add these tags to all the config structs in a future pr as well

indexer/indexer.toml
```diff
@@ -3,6 +3,14 @@
...
 # OP Goerli
 preset = 420
+
+[chain.l1-contracts]
+optimism-portal = ""
+l2-output-oracle = ""
+l1-cross-domain-messenger = ""
+l1-standard-bridge = ""
+l1-erc721-bridge = ""
+
```
roninjin10:
```suggestion
```
Will document this better but a Preset essentially means you don't provide the l1-contracts because they are already provided by the preset. Presets exist for chains like OP, base, base-goerli, and zora. The l1-contracts is simply an escape hatch for when a chain doesn't have a preset

roninjin10 (changes requested):
lgtm just remove the contract addresses from the toml file

indexer/indexer.toml
ethenotethan: That makes sense, just updated the config accordingly!

roninjin10 (approved):

roninjin10:
tyty so much for the contribution 🔥

PR #6807, opened 2023-08-15
contracts-bedrock: upgrade to multichain script

tynes:
**Description**

Implements a `forge script` to upgrade to the new L1 contracts. This assumes that the proxy admin owner is a gnosis safe.

The test can be ran with the following command:

```bash
forge test -vv --contracts scripts/upgrades --mc Multichain --rpc-url $ETH_RPC_URL
```

Where `ETH_RPC_URL` is set to a goerli endpoint. Add `-vvvvv` instead of `-vv` to see the execution that happens.


refcell (approved):
Checked through a few times - don't see anything that points out at me. I think any small style changes can be split out and addressed in a separate effort to refactor and clean up scripts.

PR #6818, opened 2023-08-16
contracts-bedrock: system config fix

tynes:
**Description**

Be sure to emit an event in the initialize function for the changes to the properties that the op-node cares about. This will ensure that the op-node's view of the system config is never corrupted. Without this change, an accidental reinitialize where the values are changed could result in this happening. Add events to prevent this from happening. This is done by making internal functions that are not `onlyOwner` and using those internal functions in `initialize`.

Also allows for the op-node to no longer need the hardcoded initial values of the system config because it could look at the start block and then it could fetch logs at the start block for the initial values of the system config, altho chains before this change will not have these these logs present so the values will need to be hardcoded still for the older chains.

Bump the version of the system config a minor version.


clabby (approved):
LGTM.

clabby:
Needs lint + semver lock.

refcell (approved):
Looks good 👑

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -165,10 +165,10 @@ contract SystemConfig is OwnableUpgradeable, Semver {
...
-        overhead = _overhead;
-        scalar = _scalar;
-        batcherHash = _batcherHash;
-        gasLimit = _gasLimit;
+        _setGasConfig({ _overhead: _overhead, _scalar: _scalar });
+
+        _setBatcherHash(_batcherHash);
+        _setGasLimit(_gasLimit);
 
         _setAddress(_unsafeBlockSigner, UNSAFE_BLOCK_SIGNER_SLOT);
```
tynes: double check if this is emitting an event

clabby: Add tests for `expectEmit` ?

tynes: Tests are added

clabby (approved):

PR #6827, opened 2023-08-16
Cast GameStatusString onto Status Type

refcell:
**Description**

Removes the `GameStatusString` function that accepts a game status type and adds
s `String()` method onto the game status type itself.

### DIFF
--- a/op-challenger/fault/caller.go
+++ b/op-challenger/fault/caller.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"math/big"
 
-	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
-	"github.com/ethereum-optimism/optimism/op-challenger/fault/types"
-
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/log"
+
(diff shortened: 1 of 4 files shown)


refcell:
* develop
  * **PR #6827** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6827" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6827?utm_source=stack-comment).

op-challenger/fault/caller.go
```diff
@@ -53,7 +57,7 @@ func (fc *FaultCaller) LogGameInfo(ctx context.Context) {
 		fc.log.Error("failed to get claim count", "err", err)
 		return
 	}
-	fc.log.Info("Game info", "claims", claimLen, "status", GameStatusString(status))
+	fc.log.Info("Game info", "claims", claimLen, "status", status.String())
```
ajsutton: We don't actually need to call `.String()` here right - it should just happen automatically?

ajsutton: Same for the other places its used in logging.

ajsutton (approved):
Looks good but should be able to avoid the explicit calls to `.String()`

op-challenger/fault/caller.go
trianglesphere: It should just work.

PR #6838, opened 2023-08-16
feat(indexer) decouple the bridge and general etl to eliminate circular dependencies in bridge processing

hamdiallam:
In it's current implementation, the bridge is coupled with the ETL phase of the chain
which introduces a circular depedency if concurrent finalization events on L2/L1 exists
in the same ETL batch that requires them waiting on each other. 

This PR splits the ETL/Bridge into independent go-routines.

1. The ETL for both L1 and L2 indexes the respective chains and contract events always at HEAD. 
This package will also responsible for eventual reorg handling

2. The BridgeProcessor simply walks indexed contracts events on both L1/L2 when it is certain that it
can process all initiated and finalization events on l1/l2 without conflict. It does this by only indexing
up till the latest Epoch indexed for both L1 & L2 (every ~6 L1 blocks)

This PR is ready for review but has some TODOs that are trivial to address. Dind't want to prolong
publishing this even longer since it's ready to generally get look at.
- Retry semantics in L1ETL & L2ETL
- Index OutputProposals on L1
- Feed in the contracts from config to the L1/L2 ETL than relying on devnet address


hamdiallam:
* develop
  * **PR #6838** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6838" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6838?utm_source=stack-comment).

indexer/database/blocks.go
```diff
@@ -198,10 +200,37 @@ func (db *blocksDB) LatestL2BlockHeader() (*L2BlockHeader, error) {
...
+func (db *blocksDB) LatestEpoch() (*Epoch, error) {
+	// Since L1 blocks occur less frequently than L2, we do a INNER JOIN from L1 on
+	// L2 for a faster query. Per the protocol, the L2 block that starts a new epoch
+	// will have a matching timestamp with the L1 origin.
+	query := db.gorm.Table("l1_block_headers").Order("l1_block_headers.timestamp DESC")
+	query = query.Joins("INNER JOIN l2_block_headers ON l1_block_headers.timestamp = l2_block_headers.timestamp")
+	query = query.Select("*")
+
+	var epoch Epoch
+	result := query.Take(&epoch)
```
roninjin10, line 240: Gorms API somewhat confuses me I get how it works but I don't get why it works this way (passing in a pointer)

indexer/database/bridge_transfers.go
```diff
@@ -214,3 +212,24 @@ finalized_l1_contract_events.transaction_hash AS finalized_l1_transaction_hash`)
...
+	}
+
+	depositsQuery := db.gorm.Table("l1_bridge_deposits").Select("MAX(l1_bridge_deposits.timestamp) as latest_deposit_time")
+	depositsQuery = depositsQuery.Joins("l1_block_headers").Where("l1_block_headers.timestamp <= latest_deposit_time")
+	depositsQuery = depositsQuery.Joins("INNER JOIN l2_block_headers ON l1_block_headers.timestamp = l2_block_headers.timestamp")
+	depositsQuery = depositsQuery.Select("l1_block_headers.*")
+
+	return nil, nil
+}
+*/
```
roninjin10:
```suggestion
```
This got moved elsewhere it looks like

indexer/database/contract_events.go
```diff
@@ -218,3 +199,34 @@ func (db *contractEventsDB) L2ContractEventsWithFilter(filter ContractEvent, fro
 
 	return events, nil
 }
+
+// Auxilliary methods for both L1 and L2
+
+// ContractEventsWithFilter will retrieve contract events within the specified range according ot `chain`.
+// NOTE: Not the fondest of the `chain` arg but it unblocks everything else
```
roninjin10: Seems fine but if we wanted to do something here easiest thing is to Create l1 and l2 implementations of a struct that have the same interface and start using that instead of naming things l1 and l2 and switching on them.

I think it's fine as is though doing everything in a switch like this is the exact same pattern expressed in a different way that some people prefer

indexer/etl/etl.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+type ETL struct {
+	log log.Logger
+
+	headerTraversal *node.HeaderTraversal
+	ethClient       *ethclient.Client
+	contracts       []common.Address
+
+	etlBatches chan ETLBatch
```
roninjin10, line 28: Huge go noob. Can you eli5 what a channel is specific this use case?

indexer/etl/etl.go
```diff
@@ -0,0 +1,112 @@
...
+	Logger log.Logger
+
+	Headers   []types.Header
+	HeaderMap map[common.Hash]*types.Header
+
+	Logs           []types.Log
+	HeadersWithLog map[common.Hash]bool
+}
+
+func (etl *ETL) Start(ctx context.Context) error {
```
roninjin10, line 41: perfect!  Lots of complexity removed out of rest of code and stuck here

indexer/etl/l1_etl.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type L1ETL struct {
+	ETL
+
+	db *database.DB
+}
+
+func NewL1ETL(log log.Logger, db *database.DB, client node.EthClient) (*L1ETL, error) {
```
roninjin10: Here you are doing literally the pattern i recommended above so same page on how to make hte l1/l2 stuff cleaner

roninjin10: Generally everything is cleaner the more the code looks like
chain.pairchain
chain.l1
chain.l2
chain.origin
chain.destination
Basically the more you are treating chain objects as generic and the relationship between l1 and l2 as declarative and generic all the switching logic and room for bugs there melts away. I need to finish this library which tries to [export the pattern](https://github.com/roninjin10/rollupChains)

indexer/etl/l1_etl.go
```diff
@@ -0,0 +1,106 @@
...
+	log = log.New("etl", "l1")
+
+	// work only with devnet deployments for now
```
roninjin10:
```suggestion
```
This isn't true anymore right?

indexer/etl/etl.go
```diff
@@ -0,0 +1,112 @@
...
+	defaultLoopInterval     = 5 * time.Second
+	defaultHeaderBufferSize = 500
+)
+
+type ETL struct {
+	log log.Logger
+
+	headerTraversal *node.HeaderTraversal
+	ethClient       *ethclient.Client
+	contracts       []common.Address
```
roninjin10, line 26: since everyone ends up at the same place here eventually we could actually abstract right away this to 

```typescript
Array<{
name: string
abi: string
address: string
startBlock: number
}>
```
It's a safe abstraction to do since everyone always ends up here from what I've seen

indexer/etl/etl.go
```diff
@@ -0,0 +1,112 @@
...
+	var headers []types.Header
+	for {
+		select {
+		case <-done:
+			etl.log.Info("stopping etl")
+			return nil
+
+		case <-pollTicker.C:
+			if len(headers) == 0 {
+				newHeaders, err := etl.headerTraversal.NextFinalizedHeaders(defaultHeaderBufferSize)
```
roninjin10, line 56: Scale this range up and down dynamically based on RPC failure is the easiest way to speed up the indexer and lots of hardened reference implementations out there

indexer/processors/bridge/l1_bridge_processor.go
```diff
@@ -0,0 +1,249 @@
...
+			GasLimit:             depositTx.GasLimit,
+			Tx:                   depositTx.Tx,
+		}
+
+		// catch ETH transfers to the portal contract.
+		if len(depositTx.DepositTx.Data) == 0 && depositTx.DepositTx.Value.BitLen() > 0 {
+			ethDeposits = append(ethDeposits, &database.L1BridgeDeposit{
+				TransactionSourceHash: depositTx.DepositTx.SourceHash,
+				BridgeTransfer:        database.BridgeTransfer{Tx: transactionDeposits[i].Tx, TokenPair: database.ETHTokenPair},
+			})
```
roninjin10, line 48: nit: seems unnecessary to handle ethDeposits while processing we can figure out if something is an eth deposit or not purely via reading

indexer/processors/bridge/l1_bridge_processor.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+		if len(ethDeposits) > 0 {
+			log.Info("detected portal ETH transfers", "size", len(ethDeposits))
+			if err := db.BridgeTransfers.StoreL1BridgeDeposits(ethDeposits); err != nil {
+				return err
+			}
+		}
+	}
+
+	// (2) L1CrossDomainMessenger
```
roninjin10, line 65: Not easy to tell at a glance if these are 3 isolated steps or if there is some cross step dependencies like OptimismPortal is a dependency to L1CrossDomainmessenger being able to process

indexer/processors/bridge.go
```diff
@@ -0,0 +1,119 @@
...
+
+func NewBridgeProcessor(log log.Logger, db *database.DB) (*BridgeProcessor, error) {
+	log = log.New("processor", "bridge")
+
+	// TODO: Just start at genesis. This processor wont be restartable until
+	// we can fill in the starting point
+	return &BridgeProcessor{log: log, db: db, LastEpoch: nil}, nil
+}
+
+func (b *BridgeProcessor) Start(ctx context.Context) error {
```
roninjin10, line 62: One thing that I think is missing from the processor in general is the idea of "catching up" while also processing at head. Ideally when you run the indexer it both is always indexing new events right away while also catching up from it's latest fully synced state. Not only does this make the indexer 'good enough' for users right away even in a full resync in many cases but it also makes the UI more testable e2e tests vs live networks easier and hardens the code by guaranteeing it always gracefully handles processing blocks out of order

roninjin10 (commented):
Got about halfway through reviewing it looks great so far

indexer/database/blocks.go
roninjin10: This isn't gorms api btw just passing a pointer into anything to fill up the pointer doesn't click with me sure it's a perfectly normal idiomatic thing though

hamdiallam: Lol it's been fun to learn Gorms API. Pretty intuitive but also a bit quirky in some situations

Also the reason the pointer is required is so that after the function returns, it's the same block of memory (struct backing the pointer) that's being updated as what the `epoch` variable is pointing to outside of the function call. If it didn't take a pointer, then the function would merely be updating a copy of the struct on the stack. The changes would not be seen in the intended `epoch` variable. Hope this makes some sense

indexer/database/bridge_transfers.go
hamdiallam: Ah no this is intended, just didn't get around to finishing it.

1. `BridgeTransfers.LastestEpoch`: detects the last epoch indexed for BridgeTransfers as this is where the bridge processor should start from. This will allow the bridge processor to pick up where it left on from on when calling `NewBridgeProcessor`.

2. `Blocks.LastestEpoch` is the last epoch indexed between the L1 & L2 ETLs. This is used by the bridge processor to understand how much forward it progress relative to the L1/L2 ETL.

Will definitely be adding some comments to these functions before merging

hamdiallam: Since the E2E tests pass as it always starts from genesis. This is one of the TODOs i was leaving to quickly implement tomorrow

indexer/database/contract_events.go
hamdiallam: Yea it's a bit quirky but was the least-hackiest solution for universal contracts like `contracts_bedrock/cross_domain_messenger.go` to be cleaner. Almost was gonna explore generics and such but figured I was getting into over-engineering territory when this is a super simple solution.

We can re-think this is it starts popping up on more places. I'm confident though that this is the only place it's needed so that `cross_domain_messenger.go` & `standard_bridge.go` could remain universal and not switch on l1/l2

indexer/etl/etl.go
hamdiallam: yea ofc! So basically `etl.go`'s job is to just retrieve the next set of headers and logs. This channel represents where that information gets dumped to. Also someone has to read from the channel before it can dump another batch. So it also acts as a buffer to wait until the pending batches sent in the channel are consumed -- (nit but this is a property of "unbuffered channels". "buffered channels work slightly differently)

`l1_etl` and `l2_etl` listen for batches in their respective channels and indexes them into the database. The indexing part is separate because on L2 we index everything and on L1 we only index blocks with logs. This file helps DRY the extraction process of new `ETLBatch`s

indexer/etl/etl.go
hamdiallam: yea i definitely agree. I dont like how the abis are always extracted on every call in `processors/contracts_bedrock`. I figured we could clean this up a bit and share some contract metadata. Lets create a ticket in the backlog for this

indexer/etl/etl.go
hamdiallam: +1. Even more powerful is if the range is beyond some threshold (i.e 256 blocks), we simply partition the range across workers and wait for them all to finish

indexer/processors/bridge/l1_bridge_processor.go
hamdiallam: So the reason I had this is because the deposits handler reads from the bridge deposits table and not the transactions deposits table. By adding this here, it allows an entry to be created in the bridge deposits table making it show in the API.

hamdiallam: Alternatively, we can have the deposits/withdrawals return **all** unique deposits/withdrawals like arbitrary messages and our API can simply hydrate a deposit/withdrawal with optional properties
- transaction deposit X is a StandardBridge message and here's the token information 
- transaction deposit Y is a L1CrossDomain message and here's the target contract and message it'll execute 
- transaction deposit Z is solely a transaction deposit via the portal. here's the raw info

And in this case we can also have a optional query param to only return results of a given type. "i only want standardbridge deposits/withdrawals"

indexer/processors/bridge/l1_bridge_processor.go
hamdiallam: I can add some more to the comment above but they are dependencies such that the bridge transfers can refer down to the messenger and the messenger can refer to the portal, etc, etc

Hence code blocks like this:
```go
	if len(crossDomainSentMessages) > len(transactionDeposits) {
		return fmt.Errorf("missing transaction deposit for each cross-domain message. deposits: %d, messages: %d", len(transactionDeposits), len(crossDomainSentMessages))
	}
```

indexer/processors/bridge.go
hamdiallam: ooh yea totally this is a good point and requires some design. Let's file an issue for this as a good followup / nice to have after launch

Kinda like the my reply about speeding up the ETL, I can imagine you start indexing at HEAD then parallelize back-filling everything in the past split across a set of workers

indexer/database/blocks.go
hamdiallam: Also do be a bit more illustrative. It's not a requirement to take a reference of the first argument. It depends on the variable. For example this would work too

```go
epoch := new(Epoch)
result := query.Take(epoch)
```
This works because `new` allocates memory for the struct but returns a pointer. So `epoch` is already a point. Doing `Take(&epoch)` in this example would result in runtime error since &epoch is a pointer to a pointer

However this would **not** work
```go
var epoch *Epoch
result := query.Take(epoch)
```
This is because no memory has actually been allocated for an `Epoch` struct. In this state, epoch is simply a nil pointer typed to `*Epoch` but doesn't point anywhere. This would compile fine but result in a panic at runtime

indexer/database/bridge_transfers.go
hamdiallam: Dont need this anymore

indexer/config/config.go
```diff
@@ -23,31 +22,18 @@ type Config struct {
 
 // fetch this via onchain config from RPCsConfig and remove from config in future
 type L1Contracts struct {
-	OptimismPortal         common.Address `toml:"optimism-portal"`
-	L2OutputOracle         common.Address `toml:"l2-output-oracle"`
-	L1CrossDomainMessenger common.Address `toml:"l1-cross-domain-messenger"`
-	L1StandardBridge       common.Address `toml:"l1-standard-bridge"`
-	L1ERC721Bridge         common.Address `toml:"l1-erc721-bridge"`
```
roninjin10, line 30: Removing this intentionally

indexer/processors/contracts_bedrock/utils.go
```diff
@@ -0,0 +1,63 @@
+package contracts_bedrock
```
roninjin10: Is there a better place this code could live later? This is golang sdk utils
```suggestion
// Consider moving me to a general utils package for bedrock
package contracts_bedrock
```

indexer/processors/contracts_bedrock/cross_domain_messenger.go
```diff
@@ -0,0 +1,167 @@
+package contracts_bedrock
```
roninjin10:
```suggestion
package contracts
```
Can just name them contracts because contracts_bedrock is suposed to be contracts. It's only contracts_bedrock to avoid name collision and they plan on changing it eventually

roninjin10 (approved):
lgtm

indexer/processors/contracts_bedrock/utils.go
hamdiallam: yea we probably move the nonce decoding logic elsewhere

indexer/processors/contracts_bedrock/cross_domain_messenger.go
hamdiallam: +1 to just naming them contracts. I'll update

PR #6841, opened 2023-08-16
op-chain-ops: always fund precompile balances in L2 genesis

tynes:
**Description**

Previously, the precompile balances were not always funded with a single wei and it was behind a developer flag. It is important to ensure that predeploys have a single wei so they are not marked as empty when interacting with them, otherwise calls may revert.


trianglesphere (commented):
Test failing, but once that is fixed, we should bring in this change.

trianglesphere (approved):

PR #6845, opened 2023-08-16
op-e2e: Print tx trace info

ajsutton:
**Description**

When a failure message includes a long line, the long line and everything after it gets trucated by testify. So print the tx trace rather than adding it to the error. The truncation happens because testify uses a `bufio.Scanner` which applies a maximum token size but testify ignores errors returned by `Scanner` so the content is just silently lost.


trianglesphere (approved):

ajsutton:
https://github.com/stretchr/testify/issues/746

PR #6854, opened 2023-08-17
Make shebangs more portable

sveitser:
**Description**

The `/bin/bash` binary/symlink is not available on all systems.

The new shebang works if `/usr/bin/env` exists and `bash` is in the path which should be the case for most reasonably modern systems.

### DIFF
--- a/op-bindings/Makefile
+++ b/op-bindings/Makefile
@@ -1,4 +1,4 @@
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
 
 pkg := bindings
 monorepo-base := $(shell dirname $(realpath .))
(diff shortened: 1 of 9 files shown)


maurelian:
We haven't had anyone report this as being an issue for them, have you had a problem with it yourself?

sveitser:
Yes. On NixOS there's no `/bin/bash`.

tynes (approved):

sveitser:
Do I need to do anything to trigger the CI? It's been saying "Expected — Waiting for status to be reported" ever since I opened the PR.

trianglesphere (approved):
changes lgtm. The CI thing is weird, maybe do a force push to see if it will re-trigger? A change like this should trigger CI.

PR #6865, opened 2023-08-17
contracts-bedrock: add deploy config validator

tynes:
**Description**

Adds a validator for the deploy configs so that we can be automatically sure that deploy configs are valid. The problem is that `getting-started.json` will never pass deploy config validation because it is invalid JSON and it is the one that we want to be sure is up to date. Will leave solving that to future work because it requires a change to the getting started guide for the op stack.


tynes:
We will want to rebase this on top of https://github.com/ethereum-optimism/optimism/pull/6867 once it is merged

tynes:
To prevent regressions in the future where the `getting-started` guide isn't broken, we will need to make it be valid JSON. Since it is not valid JSON, it will always fail these tests. This will involve altering the guide. I recommend that we use easy to read but valid values instead of things like `ADMIN` but instead an address like `0xad000..` for admin

cc @sbvegan

op-chain-ops/genesis/config.go
```diff
@@ -172,6 +174,14 @@ type DeployConfig struct {
...
 	// EIP1559Denominator is the denominator of EIP1559 base fee market.
 	EIP1559Denominator uint64 `json:"eip1559Denominator"`
+	// FaultGameAbsolutePrestate
+	FaultGameAbsolutePrestate common.Hash `json:"faultGameAbsolutePrestate"`
+	// FaultGameMaxDepth
+	FaultGameMaxDepth uint64 `json:"faultGameMaxDepth"`
+	// FaultGameMaxDuration
+	FaultGameMaxDuration uint64 `json:"faultGameMaxDuration"`
+	// SystemConfigStartBlock
+	SystemConfigStartBlock uint64 `json:"systemConfigStartBlock"`
```
clabby, line 180: Will need to be rebased on #6867

clabby (commented):
Looks good! Changes in #6867 add the fault proof params in, will need a rebase.

trianglesphere (approved):

packages/contracts-bedrock/deploy-config/hardhat.json
```diff
@@ -5,7 +5,8 @@
   "l1StartingBlockTag": "earliest",
   "l1ChainID": 900,
   "l2ChainID": 901,
-  "l2BlockTime": 2,
+  "l2BlockTime": 1,
+  "l1BlockTime": 2,
```
refcell, line 9: Was this intentional, or should this be
```suggestion
  "l2BlockTime": 2,
  "l1BlockTime": 1,
```

refcell: Also looks like everywhere else is set to `12`?

packages/contracts-bedrock/deploy-config/mainnet.json
```diff
@@ -32,7 +33,6 @@
   "governanceTokenSymbol": "OP",
   "governanceTokenOwner": "0x5C4e7Ba1E219E47948e6e3F55019A647bA501005",
   "l2GenesisBlockGasLimit": "0x1c9c380",
-  "l2GenesisBlockCoinbase": "0x4200000000000000000000000000000000000011",
```
refcell, line 35: Why'd this get removed?

refcell (approved):

packages/contracts-bedrock/deploy-config/mainnet.json
tynes: It is always hardcoded to that value, its legacy from when it was configurable

PR #6868, opened 2023-08-17
cannon: Refactor ALU execution

Inphi:
Reduce cyclomatic complexity of Go and Solidity MIPS VMs for more readable code. This has minimal effect on runtime. Though better gas savings in Solidity.

Fixes CLI-4136


protolambda (changes requested):
With this change the onchain `MIPS.sol` and offchain `mips.go` read much differently than before. Can we structure it to read more like the same? Finding discrepancies between the solidity and Go was much easier in the older version. Readability and cross-reference checking especially is more important than any performance gain.

Inphi:
@protolambda fwiw, MIPS.sol would be less readable if it were to be more similar to the Go VM. There are some things we can do in Go but can't in Solidity. Though the overall control flow between both VMs are still identical in this PR.

The most significant change here is the use of Yul assembly to implement the ALU operations. I adopted Yul because it's surprisingly easier to read than Solidity (and having access to switch statements is nice). There are several existing Yul instructions that map cleanly with MIPS. For example, implementing MIPS `slt` in Solidity involves branches. Whereas in Yul, it's a simple `slt` yul expression. The main issue with Yul is remembering to explicitly sign-extend integers since we're operating in 32-bits, since we can't simply cast to `int32`. While debugging discrepancies between VMs, this was the most common issue I ran into.

Inphi:
Both VMs should be a bit more similar now. Yul is still used, but I hope that's not too dissimilar from the Go counterparts.

protolambda:
IMO the readability is still not ideal, but I was able to verify it's correct with side-by-side sources now.
As of commit a1137e878deb10064c3d6b391bebd9337c563d04

Besides the large amount of yul for the store/load instructions (which all checked out, but not easily compared), I did find:
```go
		case 0x03: // sra
			shamt := (insn >> 6) & 0x1F
			return SE(rt>>shamt, 32-shamt)
```
```yul
                    // sra
                    case 0x03 {
                        let shamt := and(shr(6, insn), 0x1F)
                        out := sar(signextend(3, shamt), signextend(3, rt))
```
To be an interesting diff.


The shift-amount `shamt` matches, but then:
- in Go it shifts first, and then sign-extends (with a Go helper that has a different API than the yul `signextend`)
- in Yul it sign-extends the shift-amount and rt at the `3*8+7=31` bit (to full 256 bit range, because EVM...). The sign-extension on `shamt` is unnecessary, as the 31th bit is never set (because of the `and(..., 0x1F)`). `sar` then preserves the sign-extension, but also preserves the full 256 bits I think, the write to a `uint32` output is unchecked. And only because this happened with `swr` before, the caller masks it back to `uint32`.

The Yul is nice for the switch-statement, but IMO we gain little, since the Go does implementation is still kind of different to the reader. In my risc-v fault-proof I tried to work-around that by introducing yul functions that operate on 64-bits, and Go functions that look like yul, but that isn't as clean either.

I'm ok merging this as-is, maybe with the fix that we remove that unnecessary `signextend` and do the 32-bit-size masking of the output at the end of the`execute` function, rather than relying on the caller doing it.

protolambda:
Can someone else give this a second look? I think it's ok as-is, but it's a noisy diff on important code.

Inphi:
Good catch on the unnecessary `signextend`. It does illustrate the case for almost identical implementations as I'd be less likely to miss that otherwise. 
I'll see what I can do to reduce assembly here.

clabby (approved):

protolambda (approved):

PR #6871, opened 2023-08-17
op-service: jsonutil sorted-map

protolambda:
**Description**

This is a small util to get consistent JSON-encoding order of map-data. This is stacked between the `superchain-config` PR and the `registry-data-tool`, since the tool needs both the superchain registry types and the new JSON-util.

**Tests**

Tests that the encoding/decoding works as expected.

PR #6872, opened 2023-08-17
op-chain-ops: registry-data tool

protolambda:
**Description**

This is a small tool to convert a `genesis.json` or genesis block header into the assets to be added to the superchain-registry.
See https://github.com/ethereum-optimism/superchain-registry/pull/2

This PR depends on #6581 and #6871 to be merged first.

Fix CLI-4318


protolambda:
rebased + fixed lint

op-chain-ops/cmd/registry-data/main.go
```diff
@@ -0,0 +1,261 @@
...
+			GasUsed:    genesisHeader.GasUsed,
+			ParentHash: genesisHeader.ParentHash,
+			BaseFee:    (*hexutil.Big)(genesisHeader.BaseFee),
+			Alloc:      make(jsonutil.LazySortedJsonMap[common.Address, GenesisAccount]),
+			StateHash:  &genesisHeader.Root,
+		}
+		if err := writeGzipJSON(ctx.Path(OutputFlag.Name), out); err != nil {
+			return fmt.Errorf("failed to write output: %w", err)
+		}
+		return nil
```
ajsutton, line 176: nit: Is it worth making this a separate method?

ajsutton (approved):
Looks good. Should we have at least a basic test for this?

op-chain-ops/cmd/registry-data/main.go
protolambda: It's only really this tool that does the conversion this way. All the op-node/op-getg code depends onthe data to objects conversion the other way. Maybe when we add new tooling for a superchain-registry CI flow we can consider splitting it out.

PR #6093, opened 2023-06-21
fix(batcher): solve race condition of BatchSubmitter

bnoieh:
**Description**

op-batcher crushed due to error:
```
fatal error: concurrent map writes
​
goroutine 863705 [running]:
github.com/ethereum-optimism/optimism/op-batcher/batcher.(*channelManager).nextTxData(0xc0002e21a0)
	/app/op-batcher/batcher/channel_manager.go:177 +0x287
github.com/ethereum-optimism/optimism/op-batcher/batcher.(*channelManager).TxData(0xc0002e21a0, {{0x3d, 0x8d, 0x6e, 0xbf, 0xf, 0x34, 0x79, 0xb9, 0x67, ...}, ...})
	/app/op-batcher/batcher/channel_manager.go:224 +0x42a
github.com/ethereum-optimism/optimism/op-batcher/batcher.(*BatchSubmitter).publishTxToL1(0xc000602340, {0x12df0f0?, 0xc00030e3c0?}, 0xc001be23c8?, 0xc0025b17a0?)
	/app/op-batcher/batcher/driver.go:366 +0x309
github.com/ethereum-optimism/optimism/op-batcher/batcher.(*BatchSubmitter).publishStateToL1.func1()
	/app/op-batcher/batcher/driver.go:335 +0xc5
created by github.com/ethereum-optimism/optimism/op-batcher/batcher.(*BatchSubmitter).publishStateToL1
	/app/op-batcher/batcher/driver.go:326 +0xfb
​
... ...
```

I config batcher `target-num-frames=20` and `max-pending-tx=30`, then load test for 24 hours, above error show up.

The `BatchSubmitter` could concurrently access [state](https://github.com/ethereum-optimism/optimism/blob/48e0b9fcb58e1de430f4763546654adcd2ea56e9/op-batcher/batcher/driver.go#L47)  ([here](https://github.com/ethereum-optimism/optimism/blob/48e0b9fcb58e1de430f4763546654adcd2ea56e9/op-batcher/batcher/driver.go#L339) and [here](https://github.com/ethereum-optimism/optimism/blob/48e0b9fcb58e1de430f4763546654adcd2ea56e9/op-batcher/batcher/driver.go#L352)), which result in batcher panic

**Tests**

I have ran another 24 hours load test for this fix-pr, this error did not show up

**Additional context**

Add any other context about the problem you're solving.

**Metadata**

- Fixes #[Link to Issue]

PR #6878, opened 2023-08-17
op-batcher: Make channel manager safe for concurrent use

trianglesphere:
**Description**

Because the batcher is concurrent with respect to sending transactions, the main loop would fetch transaction data & record transaction status updates in different go routines. This would lead to concurrent access of the txChannels map which causes a panic.

This commit adds locks to the public methods for the channel manager to make it safe for concurrent use.

Reported by bnoeih in https://github.com/ethereum-optimism/optimism/pull/6093

**Tests**

None.

**Metadata**

- Fixes CLI-4347


ajsutton (approved):
LGTM.

tynes (approved):

PR #6889, opened 2023-08-17
op-challenger: Add --game-factory-address CLI option.

ajsutton:
**Description**

The factory address isn't actually used yet, but all the tests and scripts have been updated to provide it.  Currently only the single game specified with `--game-address` is actually played. This is a slightly odd place to leave things, but moves towards the model of specifying the factory with the ability to filter to specific games to play via `--game-address`, just with the filter being a lot less optional than we ultimately want.

**Tests**

Tests updated.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4341/add-option-to-specify-the-game-factory-address


packages/contracts-bedrock/deploy-config/devnetL1.json
```diff
@@ -41,10 +41,10 @@
   "governanceTokenOwner": "0xa0Ee7A142d267C1f36714E4a8F75612F20a79720",
   "eip1559Denominator": 50,
   "eip1559Elasticity": 6,
-  "l1GenesisBlockTimestamp": "0x64c811bf",
+  "l1GenesisBlockTimestamp": "0x64dea08d",
```
clabby: nit: unstage these changes

clabby:
changes lgtm; will defer to @refcell or @trianglesphere for approval.

packages/contracts-bedrock/deploy-config/devnetL1.json
ajsutton: ugh, I hate that the devnet changes checked in files these days.  Should be fixed now.

trianglesphere (approved):

op-challenger/flags/flags.go
```diff
@@ -105,14 +111,15 @@ var (
...
-	DGFAddressFlag,
+	FactoryAddressFlag,
 	TraceTypeFlag,
 	AgreeWithProposedOutputFlag,
 }
 
 // optionalFlags is a list of unchecked cli flags
 var optionalFlags = []cli.Flag{
 	AlphabetFlag,
+	DGFAddressFlag,
```
refcell: The `DGFAddressFlag` here should be renamed. The `DGF` stands for `dispute game factory` which doesn't make sense since it is a **single** dispute game.

op-e2e/e2eutils/disputegame/cannon_helper.go
```diff
@@ -48,9 +48,17 @@ func (g *CannonGameHelper) CreateHonestActor(ctx context.Context, rollupCfg *rol
...
 
-func createConfigOption(t *testing.T, rollupCfg *rollup.Config, l2Genesis *core.Genesis, gameAddr common.Address, l2Endpoint string) challenger.Option {
+func createConfigOption(
+	t *testing.T,
+	rollupCfg *rollup.Config,
+	l2Genesis *core.Genesis,
+	factoryAddr common.Address,
+	gameAddr common.Address,
+	l2Endpoint string,
+) challenger.Option {
```
refcell, line 58: Is this from a gofmt from a golang lint or did we settle on some config?

ajsutton: This method definition got another option which made it even longer and so I manually wrapped it. So following the existing "use common sense" policy.

op-challenger/flags/flags.go
ajsutton: oh wow, that was always a bad name for it. Fixed.

PR #6894, opened 2023-08-18
op-challenger: Initial support for playing multiple games

ajsutton:
**Description**

Wire in initial support for playing multiple games.

* Only works with alphabet games - cannon games all use the same data dir and will interfere with each other
* Resources are not cleaned up when games complete (including both files on disk and the game agent in memory)
* Loads all games from the factory contract which isn't sustainable long term as the number of games grows
* Games are progressed entirely sequentially (each update period it iterates through each game and attempts one move).

The `--game-address` option restricts the challenger to playing a single game but it must exist in the specified dispute game factory.

**Tests**

e2e test for playing multiple alphabet game is now passing.

**Metadata**

- https://linear.app/optimism/issue/CLI-4130/automatically-detect-new-games

### DIFF
--- a/op-challenger/fault/game.go
+++ b/op-challenger/fault/game.go
@@ -39,6 +39,7 @@ func NewGame(
 	txMgr txmgr.TxManager,
 	client *ethclient.Client,
 ) (*Game, error) {
+	logger = logger.New("game", addr)
 	contract, err := bindings.NewFaultDisputeGameCaller(addr, client)
 	if err != nil {
 		return nil, fmt.Errorf("failed to bind the fault dispute game contract: %w", err)
@@ -74,22 +75,21 @@ func NewGame(
 		return nil, fmt.Errorf("failed to validate absolute prestate: %w", err)
 	}
 
-	gameLogger := logger.New("game", addr)
-	responder, err := NewFaultResponder(gameLogger, txMgr, addr)
+	responder, err := NewFaultResponder(logger, txMgr, addr)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create the responder: %w", err)
 	}
 
-	caller, err := NewFaultCallerFromBindings(addr, client, gameLogger)
+	caller, err := NewFaultCallerFromBindings(addr, client, logger)
 	if err != nil {
 		return nil, fmt.Errorf("failed to bind the fault contract: %w", err)
 	}
 
 	return &Game{
-		agent:                   NewAgent(loader, int(gameDepth), provider, responder, updater, cfg.AgreeWithProposedOutput, gameLogger),
+		agent:                   NewAgent(loader, int(gameDepth), provider, responder, updater, cfg.AgreeWithProposedOutput, logger),
 		agreeWithProposedOutput: cfg.AgreeWithProposedOutput,
 		caller:                  caller,
-		logger:                  gameLogger,
+		logger:                  logger,
 	}, nil
 }
 
--- a/op-challenger/fault/game.go
+++ b/op-challenger/fault/game.go
@@ -24,21 +24,21 @@ type GameInfo interface {
 	LogGameInfo(ctx context.Context)
 }
 
-type Game struct {
+type GamePlayer struct {
 	agent                   Actor
 	agreeWithProposedOutput bool
 	caller                  GameInfo
 	logger                  log.Logger
 }
 
-func NewGame(
+func NewGamePlayer(
 	ctx context.Context,
 	logger log.Logger,
 	cfg *config.Config,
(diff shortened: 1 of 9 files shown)


op-challenger/fault/monitor.go
```diff
@@ -2,63 +2,94 @@ package fault
...
+func (m *gameMonitor) progressGame(ctx context.Context, gameData FaultDisputeGame) error {
+	game, ok := m.games[gameData.Proxy]
+	if !ok {
+		newGame, err := m.createGame(gameData.Proxy)
+		if err != nil {
+			return fmt.Errorf("failed to progress game %v: %w", gameData.Proxy, err)
+		}
+		m.games[gameData.Proxy] = newGame
+		game = newGame
 	}
```
refcell: todo(refcell): stack ontop of this pr: refactor out and unit test a new `FetchOrCreateGame()` function

op-challenger/fault/game.go
```diff
@@ -0,0 +1,120 @@
...
+	GetGameStatus(context.Context) (types.GameStatus, error)
+	LogGameInfo(ctx context.Context)
+}
+
+type Game struct {
+	agent                   Actor
+	agreeWithProposedOutput bool
+	caller                  GameInfo
+	logger                  log.Logger
+}
```
refcell: This is a bit confusing since we already have the notion of a `Game` inside `fault/types/game.go`. Once is representative of the challenger components involved in responding to a dispute game, and the other (types/game.go) is a state representation of the fault dispute game.

A lot is being done in this PR. I think it would behove us to talk a step back and put together concrete, small and specific tickets and steps to progress the challenger instead of large(r >600 diff) prs.

That said, this does bring us further toward our goal and i quite strongly dislike blocking progress so I think we can work with what is here.

op-challenger/fault/game.go
```diff
@@ -0,0 +1,120 @@
...
+		updater, err = cannon.NewOracleUpdater(ctx, logger, txMgr, addr, client)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create the cannon updater: %w", err)
+		}
+	case config.TraceTypeAlphabet:
+		provider = alphabet.NewTraceProvider(cfg.AlphabetTrace, gameDepth)
+		updater = alphabet.NewOracleUpdater(logger)
+	default:
+		return nil, fmt.Errorf("unsupported trace type: %v", cfg.TraceType)
+	}
```
refcell, line 72: I really wish there was a more idiomatic way to do this in golang. Perhaps we can push construction onto the types themselves?

Something where we can get the provider or oracle updater off the `cfg.TraceType`. IDK - will play around with this to learn.

op-challenger/fault/game.go
```diff
@@ -0,0 +1,120 @@
...
+		if g.agreeWithProposedOutput {
+			expectedStatus = types.GameStatusChallengerWon
+		} else {
+			expectedStatus = types.GameStatusDefenderWon
+		}
+		if expectedStatus == status {
+			g.logger.Info("Game won", "status", status)
+		} else {
+			g.logger.Error("Game lost", "status", status)
+		}
```
refcell, line 114: Nested branching indicates to me this is a good candidate for splitting out into another function.

refcell (approved):

op-challenger/fault/game.go
ajsutton: Yeah the naming is getting really hard - we already have a `GameAgent` which is potentially what I'd have called this. It could be `GamePlayer` I guess, or I'm open to any suggestions you have for better naming.

In terms of the size of the PR I don't think it's as big as you think - there's a bunch of code that got moved but is otherwise unchanged. That makes the diff look big but the functional change is fairly small - just wiring in the existing dispute game loader and having it create an instance of `Game` for each game it finds (what is now `Game` was just the single `service` instance previously).

op-challenger/fault/game.go
ajsutton: Yeah, happy to look at refactoring this in a follow up but this code hasn't changed. Longer term we likely drop the config option for trace type and determine it based on the game type but that's not in the scope of FPA.

op-challenger/fault/game.go
ajsutton: Also happy to look at refactoring this in a follow up but this also hasn't changed - just moved.

op-challenger/fault/monitor.go
ajsutton: I've done this refactoring. Once you pull out the fetch or create logic, the `progressGame` method isn't worth having so I've inlined it into `progressGames`.

PR #6895, opened 2023-08-18
chore(op-service) - cleanup logging service utils

### LINKED ISSUE #6734: op-service: cleanup logging service utils
- [ ] We should refactor the log flags to use `cli.GenericFlag`, which can then validate the flag value during flag-parsing time, instead of requiring `config.Check()`
- [ ] The log writes to `os.Stdout` but should be writing to `app.Writer`, so we can test the CLI better, by running the urfave app commands and outputting into a buffer/checker.
- [ ] Remove side-effects: The local logger config affects the global logger: this is undocumented/bad side-effect behavior. `NewLogger` should not touch `log.Root()`.
- [ ] Log-level should be runtime-configurable. We can have an `admin_setLogLevel` RPC endpoint to change it.
- [ ] Move testlog package from op-node to op-service

dhanusaputra:
**Description**

- [x] We should refactor the log flags to use cli.GenericFlag, which can then validate the flag value during flag-parsing time, instead of requiring config.Check() -> use Action for each flag, instead of using cli.GenericFlag
- [x] The log writes to os.Stdout but should be writing to app.Writer, so we can test the CLI better, by running the urfave app commands and outputting into a buffer/checker.
- [x] Remove side-effects: The local logger config affects the global logger: this is undocumented/bad side-effect behavior. NewLogger should not touch log.Root().
- [x] Log-level should be runtime-configurable. We can have an admin_setLogLevel RPC endpoint to change it.
- [x] Move testlog package from op-node to op-service

**Tests**

added unit test TestLogFormat

**Metadata**

- Fixes #6734


protolambda:
> Log-level should be runtime-configurable. We can have an admin_setLogLevel RPC endpoint to change it.

I recommend you implement the log-handler interface like the geth `log.LvlFilterHandler` but make the `maxLvl` changeable with a method, and then replace the `LvlFilterHandler` usage in the op-service `log.NewLogger` function. By then retaining a reference to the handler after creation of the logger (maybe add a `LogControls` interface` as second return argument, and return the handler), you can make the log level changeable from within the op-node, and add an RPC method to the op-node admin API to let the user change it.

protolambda:
Re-opening, will fix up this PR with merge-conflict resolution and the missing features once work backlog frees up.

op-node/rollup/driver/state.go
```diff
@@ -388,6 +391,16 @@ func (s *Driver) eventLoop() {
 			}
 		case respCh := <-s.sequencerActive:
 			respCh <- !s.driverConfig.SequencerStopped
+		case respCh := <-s.seLogLevel:
```
protolambda: Only chain/sync/engine changes should be routed through the driver. The log level handler does not have to be synchronized like this, and is unrelated to the driver. Adding a new interface to the admin API to set the log-level through would be better. And the handler should not be changed like this, or every new call ends up creating a handler wrapping the old one, and won't allow the log-level to become more verbose than the previous log level.

protolambda (commented):
Thanks for updating this PR, but the handler update isn't quite right, see review comment. I can help fix it later today if you like

op-node/rollup/driver/state.go
dhanusaputra: sorry for delay, have updated based on your comment
but not sure with this part

> and won't allow the log-level to become more verbose than the previous log level

the logger itself doesn't store the state of previous log level so not sure how to compare it

op-service/log/cli.go
```diff
@@ -46,28 +63,14 @@ type CLIConfig struct {
...
-}
-
-func NewLogger(cfg CLIConfig) log.Logger {
-	handler := log.StreamHandler(os.Stdout, Format(cfg.Format, cfg.Color))
+	handler := log.StreamHandler(wr, Format(cfg.Format, cfg.Color))
 	handler = log.SyncHandler(handler)
 	handler = log.LvlFilterHandler(Level(cfg.Level), handler)
-	// Set the root handle to what we have configured. Some components like go-ethereum's RPC
-	// server use log.Root() instead of being able to pass in a log.
-	log.Root().SetHandler(handler)
```
trianglesphere, line 70: This was pretty explicitly set. Because we use geth as a fully fledged library here, we need to adapt to their logging patterns. I'm not aware of a way to provide  a logger to the RPC component. See https://github.com/ethereum-optimism/optimism/pull/5086

Setting the root handler is also beneficial in the case where we accidentally call `log.Info` instead of `logger.Info`. I do acknowledge that it has a tradeoff when running multiple in memory instances and is not nice to have a global logger.

op-node/node/api.go
```diff
@@ -71,6 +74,22 @@ func (n *adminAPI) SequencerActive(ctx context.Context) (bool, error) {
 	return n.dr.SequencerActive(ctx)
 }
 
+func (n *adminAPI) SetLogLevel(ctx context.Context, lvlStr string) error {
```
trianglesphere, line 77: Probably out of scope for this PR, but this would be nice to also have in op-service so the batcher/proposer/challenger would automatically inherit this capability.

protolambda: Good suggestion, but we do not have a standard admin-RPC in the op-service yet. I'll open an issue for it, definitely nice to have.

protolambda: #7417

op-service/log/cli.go
protolambda: I changed it so that the global-log-handler is explicitly set to a copy of the created log handler during the CLI setup phase, this way the logging setup is otherwise side-effect free, while we still preserve the global log output in regular program execution. And then during op-e2e testing we can set the global log handler to the testlog handler, without it being overwritten, cleaning up the test logs nicely.

protolambda:
Ok, applied some changes to make it merge-ready:
- Admin log level API now doesn't do the handler-wrapping, but changes the level of the existing handler
- Cleaned up global log handler; preserve side-effect when desired by making it explicit, while allowing test setup to set it to the test logger too.
- Cleaned up remaining bits to not change driver interface unnecessarily
- Implemented the log attributes with CLI generic flags, to make validation part of the flag parsing as originally intended in the issue, as well as making the configuration more type-safe.
- Misc cleanup with imports, log util package API, and log-handler-writer usage.

trianglesphere (approved):
LGTM

hamdiallam (approved):

protolambda:
@Mergifyio refresh

tynes:
Any idea what is going on with CI?

```
ERROR: gcloud crashed (OAuthError): ('Error code invalid_request: Invalid value for "audience". This value should be the full resource name of the Identity Provider. See https://cloud.google.com/iam/docs/reference/sts/rest/v1/TopLevel/token for the list of possible formats.', '{"error":"invalid_request","error_description":"Invalid value for \\"audience\\". This value should be the full resource name of the Identity Provider. See https://cloud.google.com/iam/docs/reference/sts/rest/v1/TopLevel/token for the list of possible formats."}')

If you would like to report this issue, please run the following command:
  gcloud feedback

To check gcloud for common problems, please run the following command:
  gcloud info --run-diagnostics

Exited with code exit status 1
```

protolambda:
@tynes maybe some CI job that does not handle the case of the branch being sourced from outside of the repository? Is that log from one of the docker publish jobs?

PR #6896, opened 2023-08-18
feat: API example of viem decorators

roninjin10:
This is a POC of what viem based sdk would look like

## Current sdk

- bespoke API with it's own complexity and documentation
- ethers v5 based

## New sdk

- viem based
- easy to use without docs because it's api matches viem exactly
- extends viem natively
- rollup functionality that is more general than OP perhaps we upstream to viem

## Example

In this example I create a function `bridgeContractCall` which calls a contract on l2 from l1

- a reference implementation that more or less matches what a real implementation would look like
- A test that shows what using this library looks like for the user

bridgeWriteContract has the exact same API as viem's [writeContract](https://viem.sh/docs/contract/writeContract.html#writecontract). It ergonomically feels no different than calling contractCall on l2 to the developer

## What would need to happen to productionize this

- lots of TODOs in the code
- we would want to decorate a `bridgeCreateContract` method that is like creating a bridged version of `createContract`
- The action should be exported both as a standalone action and also as a client decorator matching how viem works. Users can use actions without clients (usually better code splitting this way)
- would generalize to all OP chains. Needs to be able to override options such as crossChainMessenger address
- Ideally we would add extended viem `chain` objects that include these config options for most chains
  - Ideally these are generated from on chain configs

### DIFF
--- a/packages/viem/package.json
+++ b/packages/viem/package.json
@@ -1,0 +1,54 @@
+{
+  "name": "@eth-optimism/fee-estimation",
+  "version": "0.15.2",
+  "description": "Lightweight library for doing OP-Chain gas estimation",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ethereum-optimism/optimism.git",
+    "directory": "packages/fee-estimation"
+  },
+  "homepage": "https://optimism.io",
+  "type": "module",
+  "main": "dist/estimateFees.cjs",
+  "module": "dist/estimateFees.js",
+  "exports": {
+    ".": {
+      "import": "./dist/estimateFees.js",
+      "require": "./dist/estimateFees.cjs",
+      "default": "./dist/estimateFees.js",
+      "types": "./src/estimateFees.ts"
+    }
+  },
+  "types": "src/estimateFees.ts",
+  "files": [
+    "dist/",
+    "src/"
+  ],
+  "scripts": {
+    "build": "tsup",
+    "lint": "prettier --check .",
+    "lint:fix": "prettier --write .",
+    "test": "vitest",
+    "test:coverage": "vitest run --coverage",
+    "typecheck": "tsc --noEmit"
+  },
+  "devDependencies": {
+    "@eth-optimism/contracts-ts": "workspace:^",
+    "@testing-library/jest-dom": "^5.17.0",
+    "@testing-library/react-hooks": "^8.0.1",
+    "@vitest/coverage-istanbul": "^0.34.1",
+    "abitype": "^0.9.3",
+    "isomorphic-fetch": "^3.0.0",
+    "jest-dom": "link:@types/@testing-library/jest-dom",
+    "jsdom": "^22.1.0",
+    "tsup": "^7.1.0",
+    "typescript": "^5.1.6",
+    "viem": "^1.3.1",
+    "vite": "^4.4.6",
+    "vitest": "^0.33.0"
+  },
+  "peerDependencies": {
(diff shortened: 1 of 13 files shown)


roninjin10:
* develop
  * **PR #6896** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6896" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6896?utm_source=stack-comment).

tynes:
this is sick

packages/viem/src/index.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  // this api is the exact same api as the normal viem function `writeContract`
+  const l2TxHash = await extendedClient.bridgeWriteContract({
+    // Interacting with the optimist abi
+    abi: optimistABI,
+    // It's an SBT but pretend it isn't
+    functionName: 'transferFrom',
+    args: [myAddress, theirAddress, BigInt(1)],
+    address: optimistAddress[10],
+    account: myAddress,
+    chain: mainnet,
```
roninjin10, line 35:
```suggestion
    // TODO modify api here make it so the chain is the destination chain instead of origin chain
    chain: mainnet,
```

packages/viem/src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+type OptimismExtended = {
+  bridgeWriteContract: WalletClient['writeContract']
+}
+
+export const opViemWalletExtension = <
+  transport extends Transport = Transport,
+  chain extends Chain | undefined = Chain | undefined,
+  account extends Account | undefined = Account | undefined,
+>(client: WalletClient<transport, chain, account>) => ({
+  bridgeWriteContract: async (args) => {
```
roninjin10, line 22: make this typesafe such that only opChains have this method

roninjin10:
```suggestion
  // TODO move me to an action 
  // @see https://viem.sh/docs/actions/wallet/introduction.html
  bridgeWriteContract: async (args) => {
```

packages/viem/src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    // TODO don't hardcode this
+    const minGasLimit = 200_000
+    const message = encodeFunctionData({
+      abi: args.abi,
+      functionName: args.functionName,
+      args: args.args,
+    } as unknown as EncodeFunctionDataParameters<typeof args.abi, typeof args.functionName>)
+    const l1TxHash = await client.writeContract({
```
roninjin10, line 31:
```suggestion
    // TODO internal implementations should only rely on actions
    const l1TxHash = await client.writeContract({
```

packages/viem/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@eth-optimism/fee-estimation",
```
Spacesai1or, line 2: Would this be `@eth-optimism/viem` or something different?

spacesailor24:
👍 This is much nicer than the web3js or ethers implementations

packages/viem/package.json
roninjin10:
```suggestion
  "name": "@eth-optimism/viem",
```
I never have opinionhs on naming things but viem would be my first guess. Could just call it `@eth-optimism/sdk` too as it is the next generation of the sdk and more than other implementations our official one

roninjin10:
This merged into another repo here https://github.com/wilsoncusack/op-viem/pull/3

PR #6898, opened 2023-08-18
feat: tx bundle builder

tynes:
**Description**

Transaction builder type that can be used to automatically build transactions and serialize them as JSON so that they can be easily inspected and:
1) submitted to the network via a forge script
2) imported into the safe web ui

Includes various test coverage. It currently cannot validate tuples, I think we need to update upstream to be able to work on slices of any instead of just the bindings generated structs to be able to do this in a generic way.


op-chain-ops/safe/batch_file.go
```diff
@@ -6,44 +6,123 @@ package safe
...
+		batch.ChainID = b.ChainID.String()
+	}
+	return json.Marshal(batch)
 }
 
-// UnmarshalJSON will unmarshal a BatchFile from JSON.
-func (b *BatchFile) UnmarshalJSON(data []byte) error {
-	var bf batchFileMarshaling
+// UnmarshalJSON will unmarshal a Batch from JSON.
+func (b *Batch) UnmarshalJSON(data []byte) error {
```
maurelian: For each transaction, could we add some sanity checks here to ensure that the JSON is well formatted, such as: 
1. the function signature matches the 4byte prefix of the data
2. the data abi decodes correctly
3. the decoded data matches the `contractInputsValues`

tynes: I added a `Check` method for this instead of coupling to the JSON logic

ajsutton (commented):
Rebasing _should_ fix the e2e test failure.

This looks fine to me, but I'm not particularly familiar with what's being done here and all the details of solidity method encoding etc so I'm not confident approving. I'll request an alternative reviewer.

tynes:
I have rebased this

mslipper (approved):
LGTM. Do you know how the checksum is calculated? I think that'd be a beneficial thing to have, since the SAFE UI will flag mismatched checksums.

tynes:
> LGTM. Do you know how the checksum is calculated? I think that'd be a beneficial thing to have, since the SAFE UI will flag mismatched checksums.

I could reimplement the checksum but avoided it because its over JSON and the canonical impl is in JS and i worried that the checksum would be tied to the JS impl of JSON

PR #6900, opened 2023-08-18
feat(ctb): Pack `GameType` in DGF's `GameId` type

clabby:
## Overview

Packs the `GameType` within the `GameId` so that off chain agents that don't rely on event subscriptions can quickly determine
the types of detected games without performing an `eth_call` on each one.


clabby:
* develop
  * **PR #6900** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6900" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6900?utm_source=stack-comment).

packages/contracts-bedrock/src/libraries/DisputeTypes.sol
```diff
@@ -31,12 +33,13 @@ type Duration is uint64;
 
 /// @notice A `GameId` represents a packed 12 byte timestamp and a 20 byte address.
```
trianglesphere: Comment needs update.

op-challenger/fault/factory.go
```diff
@@ -19,14 +19,16 @@ var (
 type MinimalDisputeGameFactoryCaller interface {
 	GameCount(opts *bind.CallOpts) (*big.Int, error)
 	GameAtIndex(opts *bind.CallOpts, _index *big.Int) (struct {
+		GameType  uint8
```
trianglesphere, line 22: Where is the mapping from `GameType` to string? It'd also be really nice to already have the golang version of it.

clabby: Don't believe y'all have one yet - glad to add it!

clabby: Done, lmk if that's what you were looking for.

op-challenger/config/config.go
```diff
@@ -32,10 +32,30 @@ type TraceType string
...
+	DevnetGameIDCannon   = uint8(1)
+
+	// Mainnet game IDs
+	MainnetGameIDFault = uint8(0)
 )
 
 var TraceTypes = []TraceType{TraceTypeAlphabet, TraceTypeCannon}
 
+// GameIdToString maps game IDs to their string representation on a per-network basis.
+var GameIdToString = map[uint64]map[uint8]string{
```
trianglesphere, line 47: I didn't realize that we these IDs meant different things based on the network - I know it's weird to put in the alphabet game on mainnet, but it'd save a bit of complexity in configuration.

op-challenger/fault/factory.go
trianglesphere: that was it

op-challenger/config/config.go
clabby: Yep - It's a bit odd, but would be nice if we can support different game type mappings for the various networks (we only have 255 available types, so best not to pollute mainnet's imo). Glad to think of a way to standardize them across the networks if we want this, though.

refcell (approved):

op-challenger/config/config.go
clabby: One temporary solution is to deploy devnet-specific games with game types descending from 255, and it'll probably be a long time (if ever) before we'd need to worry about them colliding. Still would be more game types on devnet than mainnet, but the game types that exist on mainnet would have the same values on devnet.

i.e. the alphabet game moves to game type `255` on devnet, and the cannon game moves to the same value on mainnet (`0`)

trianglesphere: yea, that's not a bad solution. Let's merge this PR as is, but I do want to follow up on this.

op-challenger/config/config.go
```diff
@@ -32,10 +32,30 @@ type TraceType string
 const (
 	TraceTypeAlphabet TraceType = "alphabet"
 	TraceTypeCannon   TraceType = "cannon"
+
+	// Devnet game IDs
+	DevnetGameIDAlphabet = uint8(0)
```
tynes, line 37: Feels kind of weird that different networks have different ids

clabby: See comment above from @trianglesphere, ticket created @ https://linear.app/optimism/issue/CLI-4369/consider-having-game-types-be-the-same-across-all-networks

refcell:
@mergifyio requeue

PR #6906, opened 2023-08-18
fix(aj/cannon-multigame): Fix leaky timer

refcell:
**Description**

Fixes the leaky timer used in the game monitor component.

Does not block on draining the timer so the error can bubble up without hindrance.

### DIFF
--- a/op-challenger/fault/cannon/provider.go
+++ b/op-challenger/fault/cannon/provider.go
@@ -50,17 +50,17 @@ type CannonTraceProvider struct {
 	lastProof *proofData
 }
 
-func NewTraceProvider(ctx context.Context, logger log.Logger, cfg *config.Config, l1Client bind.ContractCaller) (*CannonTraceProvider, error) {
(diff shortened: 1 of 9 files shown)


trianglesphere (changes requested):
This PR breaks the challenger. A timer only fires once and you're creating outside the for loop while it used to be inside the for loop. If you want to do this, you need to use a ticker instead. This code will leak a single timer when the context is cancelled. That should only happen on exit though it isn't ideal.

ajsutton:
The ticker version isn't the same thing either - it is trying to run every 300ms rather than pausing 300ms after the previous actions complete. Not the end of the world because tickers will drop events that don't get read in time to avoid them piling up but this is likely to wind up being equivalent to a tight loop in a lot of cases now.

Personally I wouldn't be concerned by an unused timer only being cleaned up 300ms after the function ends, especially since it only happens once when the component shuts down (which is also typically when the entire program exits). But if we want to fix it we'd want something like:
```

		t := time.NewTimer(300 * time.Millisecond)
		select {
		case <-t.C:
		// Continue
		case <-ctx.Done():
			t.Stop()
			return ctx.Err()
		}
```

Possibly a better option would be to remove this loop entirely and only check games for updates when there's a new L1 head block because the game state can only change when there's a new block anyway.

PR #6909, opened 2023-08-18
feat(ctb): Preserve prod game type space

clabby:
## Overview

Moves the alphabet game to game ID `255` on the devnet and the cannon game to `0`, making the game impls that will be deployed to prod have the same game IDs on devnet. Going forward, the devnet will maintain parity with prod game type values and any devnet-specific games will descend from `255`. 

**Metadata**
Closes CLI-4369

### DIFF
--- a/packages/contracts-bedrock/scripts/Deploy.s.sol
+++ b/packages/contracts-bedrock/scripts/Deploy.s.sol
@@ -76,7 +76,8 @@ contract Deploy is Deployer {
         initializeL2OutputOracle();
         initializeOptimismPortal();
 
-        setFaultGameImplementation();
+        setAlphabetFaultGameImplementation();
+        setCannonFaultGameImplementation();
 
         transferProxyAdminOwnership();
         transferDisputeGameFactoryOwnership();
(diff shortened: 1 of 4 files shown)


packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -775,31 +775,69 @@ contract Deploy is Deployer {
...
+                })
+            );
+
+            uint8 rawGameType = GameType.unwrap(_gameType);
+            console.log(
+                "DisputeGameFactoryProxy: set `FaultDisputeGame` implementation (Backend: %s | GameType: %s)",
+                rawGameType == 0 ? "Cannon" : "Alphabet",
+                vm.toString(rawGameType)
+            );
         }
```
refcell, line 846: Would be nice devx to warn on the else branch here. So, if this function is used differently in the future and there _is_ a game implementation (non address(0)), it'll log a warning.

Not needed for this pr but just thinking as we may start to refactor the scripts.

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -775,31 +775,69 @@ contract Deploy is Deployer {
...
+
+        // Set the Alphabet FaultDisputeGame implementation in the factory.
+        Claim alphabetAbsolutePrestate = Claim.wrap(bytes32(cfg.faultGameAbsolutePrestate()));
+        _setFaultGameImplementation(
+            factory,
+            GameType.wrap(255),
+            alphabetAbsolutePrestate,
+            IBigStepper(new AlphabetVM(alphabetAbsolutePrestate)),
+            4 // The max game depth of the alphabet game is always 4.
+        );
```
refcell: These are now good candidates for their own functions. Especially since the parameters are entirely mutually exclusive less the factory.

The devnet function can then also verify that the chain id is indeed the devnet.

clabby: I think we should keep it how it is now due to us not wanting the deploy script to deploy the fault proof contracts to live networks yet, but agreed that these should be split out soon.

After tower defense, will be nice to re work the deployment of these contracts :)

refcell: I think we said the same after bedrock migration 🤣  But sure, nbd

refcell (approved):

ajsutton (approved):
Looks good but probably worth updating the `op-challenger/scripts/alphabet/init_script.sh` to have the new alphabet game type as well.

PR #6911, opened 2023-08-18
op-challenger: Add basic metrics & pprof setup

trianglesphere:
**Description**

This adds metrics & pprof to the op-challenger. I had to add a default config option to the metrics & pprof in order to make the tests work (because it verifies that the result of the CLI parsing is as expected when compared to an config struct which is constructed).

This also had to modify how the challenger stores its version information because we include that information in the metrics when we record that it is up. Unlike the proposer & batcher, the challenger has a different initialization flow which does not easily make the version information accessible.

**Tests**

Tested manually by modifying the alphabet game scripts.

**Metadata**

- Fixes CLI-4134


ajsutton (commented):
LGTM. Will see if I can resolve the conflicts in service.

> I had to add a default config option to the metrics & pprof in order to make the tests work (because it verifies that the result of the CLI parsing is as expected when compared to an config struct which is constructed)

Yep that's why those tests exist. We want the programmatic entry point for the challenger to be equivalent to the CLI entry point. Providing the same, sensible defaults means we don't have to go update every programmatic integration when we add a new config option.

ajsutton (approved):

PR #6912, opened 2023-08-18
chore: Rebase PR 5329

refcell:
**Description**

Rebased `fgimenez:feat/eip1967` onto develop. Commit history is preserved so the pr is ready for merge and we can close out #5329.


tynes:
Looks like this needs a lint, also there are tests where the `bytes32` values are being used instead of the defined constants

tynes:
Thank you for helping out here!

tynes (approved):

PR #6913, opened 2023-08-18
cannon,op-challenger,dispute-game: VM Status prefix byte in claim hash

protolambda:
**Description**

Introduce a prefix-byte in the state-hash, independent of the FP-VM, to indicate the status of the VM as part of the VM state commitment.
This way claim-hashes can be enforced to have certain properties, like in the case of the root-claim which always has to disagree with the validation, or else a valid program execution can cause deletion of a valid output-root.

Changes:
- Extend trace-provider interface with `StateHash` function: each trace-provider may determine the hash and VM status differently. In the case of Cannon we look at the `Exited` and `ExitCode` attributes in the state-witness.

TODO:
- [x] Document the VM Status property, and VM state-witness -> VM state-hash in general. What's the right place for this?
- [x] Update bindings
- [x] Update Alphabet game solidity tests to use the pre-image oracle to determine the claimed final alphabet letter, and implement state-hashing to include the valid/invalid/panic/unfinished status.

**Tests**

Added a test to verify that a root-claim with a bad VM-status indeed results in an error when trying to create a new fault-game.

**Metadata**

Fix CLI-4355


packages/contracts-bedrock/src/dispute/DisputeGameFactory.sol
```diff
@@ -91,6 +91,10 @@ contract DisputeGameFactory is OwnableUpgradeable, IDisputeGameFactory, Semver {
         // If there is no implementation to clone for the given `GameType`, revert.
         if (address(impl) == address(0)) revert NoImplementation(_gameType);
 
+        // The VMStatus must indicate (1) 'invalid', to argue that disputed thing is invalid.
+        // Games that agree with the existing outcome are not allowed.
+        if (uint8(Claim.unwrap(_rootClaim)[0]) != 1) revert UnexpectedRootClaim(_rootClaim);
```
clabby: This kind of defeats the purpose of the abstraction - can we move this check into the `FaultDisputeGame`'s initialization function?

If the `initialize` function reverts, it'll bubble up here, and we can still create different game types that don't require this check.

clabby (dismissed):

cannon/mipsevm/state.go
```diff
@@ -60,3 +65,37 @@ func (s *State) EncodeWitness() []byte {
...
+	offset := 32*2 + 4*6
+	exited := sw[offset]
+	exitCode := sw[offset+1]
+	status := vmStatus(exited == 1, exitCode)
+	hash[0] = status
+	return hash
+}
+
+func vmStatus(exited bool, exitCode uint8) uint8 {
+	if exited {
```
ajsutton: nit: Invert this condition and return early?

op-challenger/fault/alphabet/provider.go
```diff
@@ -58,19 +59,39 @@ func (ap *AlphabetTraceProvider) Get(ctx context.Context, i uint64) (common.Hash
...
+func (ap *AlphabetTraceProvider) StateHash(ctx context.Context, state []byte) (common.Hash, error) {
+	return alphabetStateHash(state), nil
+}
+
 // BuildAlphabetPreimage constructs the claim bytes for the index and state item.
 func BuildAlphabetPreimage(i uint64, letter string) []byte {
 	return append(IndexToBytes(i), LetterToBytes(letter)...)
 }
 
+const maxAlphabet = 26
```
ajsutton: note that the alphabet game has a max depth of 4, so it's only 16 letters long.

op-challenger/game/fault/trace/alphabet/provider.go
```diff
@@ -58,19 +59,48 @@ func (ap *AlphabetTraceProvider) Get(ctx context.Context, i uint64) (common.Hash
...
+	i := new(big.Int).SetBytes(state[:32])
+	finalStateIdx := uint64(len(ap.state) - 1)
+	if !i.IsUint64() || i.Uint64() > finalStateIdx {
+		h[0] = mipsevm.VMStatusPanic // this state should never be reached, if we increment by 1 per step
+	} else if i.Uint64() == finalStateIdx {
+		h[0] = mipsevm.VMStatusInvalid
+	} else {
+		h[0] = mipsevm.VMStatusUnfinished
+	}
+	return h
```
ajsutton: This never sets the `VMStatusValid` case which seems suspicious.

op-challenger/game/fault/types/types.go
```diff
@@ -105,6 +105,12 @@ type TraceProvider interface {
 
 	// AbsolutePreState is the pre-image value of the trace that transitions to the trace value at index 0
 	AbsolutePreState(ctx context.Context) (preimage []byte, err error)
+
+	// AbsolutePreStateCommitment is the commitment of the pre-image value of the trace that transitions to the trace value at index 0
+	AbsolutePreStateCommitment(ctx context.Context) (hash common.Hash, err error)
+
+	// StateHash computes the state-hash of the given state, or returns an error if the state is invalid.
+	StateHash(ctx context.Context, state []byte) (common.Hash, error)
```
ajsutton: There doesn't appear to be anything that needs this to be part of the `TraceProvider` interface.

op-challenger/game/fault/trace/alphabet/provider.go
ajsutton: I think what actually needs to happen here is that the status is always invalid if we disagree with the output root and always valid if we agree with it.  It's a bit fake but mirrors the way a real cannon trace would work better where the person creating the game must have a cannon trace that says the root is invalid and the other side must think its valid.

On-chain should always have unfinished or Invalid since it's not valid to dispute a game with the correct trace.  However, that will cause `ChallengerIncorrectAtLastTraceIndex` to fail as it and a couple of other alphabet tests in that method are using a root claim with the correct alphabet. I'm not quite sure how we adjust those tests with this new restriction....

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -163,7 +167,11 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
         // SAFETY:    While the `attack` path does not need an extra check for the post
         //            state's depth in relation to the parent, we don't need another
         //            branch because (n - n) % 2 == 0.
-        bool validStep = VM.step(_stateData, _proof) == Claim.unwrap(postState.claim);
+        //            We ignore the highest order byte of the digest because it is used to
+        //            indicate the VM Status and is added after the digest is computed. The
+        //            hash commits to the exit code and the exit status, so the VM status does
+        //            not need to be checked for equivalence.
+        bool validStep = (VM.step(_stateData, _proof)) << 8 == (Claim.unwrap(postState.claim) << 8);
```
ajsutton: Ignoring the vm state byte here means that an attacker can provide the correct trace (that says the output is valid) but simply change the first byte of the root claim to have vm status invalid.  When the honest actor tries to call `step` it will revert because the post state matches what the dishonest actor claimed in everything except the vm status byte (which is being ignored).

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -149,7 +149,11 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
 
         // INVARIANT: The prestate is always invalid if the passed `_stateData` is not the
         //            preimage of the prestate claim hash.
-        if (keccak256(_stateData) != Claim.unwrap(preStateClaim)) revert InvalidPrestate();
+        //            We ignore the highest order byte of the digest because it is used to
+        //            indicate the VM Status and is added after the digest is computed.
+        if (keccak256(_stateData) << 8 != Claim.unwrap(preStateClaim) << 8) {
```
ajsutton, line 154: I think ignoring the vm status byte here is ok, because even if the two sides disagree on the vm state byte, they still agree on what the pre-state is. Even a noop step on that pre-state will switch the vm status byte to the correct value if needed.

ajsutton (dismissed):
I think the `ChallengerIncorrectAtLastTraceIndex` test has mislead us here.  It's actually an invalid test since the root claim has the correct trace (as do a few others in that method but they don't have an honest challenger running so don't wind up at the end of the trace).  We'll need to rethink those tests - I suspect the AlphabetVM will need to somehow have a concept of whether the trace is valid or not.  Probably skip those tests for now and then we can discuss updating the AlphabetVM separately.

The key issue though is that ignoring the vm status byte like we're doing doesn't work correctly.  I've updated the e2e tests to demonstrate this - particularly the cannon one which is what we need to get passing with this change: https://github.com/ethereum-optimism/optimism/compare/fpvm-status...aj/fpvm-status-e2e

cannon/mipsevm/state.go
refcell: ++

```golang
func vmStatus(exited bool, exitCode uint8) uint8 {
	if !exited {
		return VMStatusUnfinished
    }

    switch exitCode {
	case 0:
		return VMStatusValid
	case 1:
		return VMStatusInvalid
	default:
		return VMStatusPanic
	}
}
```

op-challenger/game/fault/player.go
```diff
@@ -159,22 +158,21 @@ func (g *GamePlayer) logGameStatus(ctx context.Context, status types.GameStatus)
...
 	}
-	providerPrestateHash := crypto.Keccak256(providerPrestate)
 	onchainPrestate, err := loader.FetchAbsolutePrestateHash(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to get the onchain absolute prestate: %w", err)
 	}
-	if !bytes.Equal(providerPrestateHash, onchainPrestate) {
-		return fmt.Errorf("trace provider's absolute prestate does not match onchain absolute prestate")
+	if !bytes.Equal(providerPrestateHash[:], onchainPrestate[:]) {
+		return fmt.Errorf("trace provider's absolute prestate does not match onchain absolute prestate: Provider: %s | Chain %s", providerPrestateHash.Hex(), onchainPrestate.Hex())
```
refcell, line 175: Should this assertion also ignore the prepended vm status byte?
ie
```suggestion
	if !bytes.Equal(providerPrestateHash[1:], onchainPrestate[1:]) {
		return fmt.Errorf("trace provider's absolute prestate does not match onchain absolute prestate: Provider: %s | Chain %s", providerPrestateHash.Hex(), onchainPrestate.Hex())
```

op-challenger/game/fault/trace/alphabet/provider.go
refcell: I think @ajsutton is correct here. The zeroth byte should be set to `VMStatusValid` if agreed.

op-challenger/game/fault/trace/cannon/provider.go
```diff
@@ -122,6 +123,14 @@ func (p *CannonTraceProvider) AbsolutePreState(ctx context.Context) ([]byte, err
 	return state.EncodeWitness(), nil
 }
 
+func (p *CannonTraceProvider) AbsolutePreStateCommitment(ctx context.Context) (common.Hash, error) {
+	state, err := p.AbsolutePreState(ctx)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("cannot load absolute pre-state: %w", err)
+	}
+	return mipsevm.StateWitness(state).StateHash(), nil
```
refcell:
```suggestion
	return p.StateHash(ctx, state)
```

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -434,9 +442,17 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
...
         // SAFETY: Any revert in this function will bubble up to the DisputeGameFactory and
         // prevent the game from being created.
+        //
         // Implicit assumptions:
         // - The `gameStatus` state variable defaults to 0, which is `GameStatus.IN_PROGRESS`
 
+        // The VMStatus must indicate (1) 'invalid', to argue that disputed thing is invalid.
+        // Games that agree with the existing outcome are not allowed.
+        // NOTE(clabby): This assumption will change in Alpha Chad.
+        if (uint8(Claim.unwrap(rootClaim())[0]) != VMStatus.unwrap(VMStatuses.INVALID)) {
```
Inphi: I'm thinking about the case where an output is proposed, but no one can validate it because the block transition results in a panic (`VMStatuses.Panic`). Don't we want to prevent such outputs? Safety vs Liveness tradeoff.

op-challenger/game/fault/types/types.go
clabby: This function is included in the `TraceProvider` due to the fact that there may be different post-processing tasks (i.e. the VM status byte) for different `TraceProviders` in the future. The thought was that having each `TraceProvider` define how a given state should be hashed + [pre/post]processed will likely save us some headaches in the future when there are more games.

op-challenger/game/fault/trace/alphabet/provider.go
ajsutton: The real challenge here is that the alphabet VM has no concept of valid or invalid - it's not validating anything. So I think this code is actually right and the alphabet VM always has to resolve that to invalid so games can be played at all, but that means we can't use the alphabet vm with the correct trace as the root claim.

My take is that the right way forward here is to just accept that and move the tests that no longer work over to use cannon.

op-challenger/game/fault/types/types.go
ajsutton: It may be useful in the future but right now it's largely leaking an internal detail and not adding any value.  I'm not strongly opinionated but generally I'd apply YAGNI (you ain't gonna need it) and only add it when we actually have a use for it.

op-challenger/game/fault/player.go
ajsutton: No. The local prestate should have the same vm status as the on-chain commitment did.  Otherwise something is very wrong.

In general we should just treat the vm status as part of the hash and not trim it.  When we do trim it we need to have a very good reason and be very careful that we aren't breaking any security properties.  Much safer to keep it when possible.

specs/cannon-fault-proof-vm.md
```diff
@@ -53,6 +54,34 @@ It consists of the following fields:
 
 The state is represented by packing the above fields, in order, into a 226-byte buffer.
 
+### State Hash
+
+The state hash is computed by hashing the 226-byte state buffer with the Keccak256 hash function
```
trianglesphere, line 59: Where did 226 come from? Should it be 256bits?

cannon/mipsevm/state.go
```diff
@@ -60,3 +65,37 @@ func (s *State) EncodeWitness() []byte {
...
+	VMStatusValid      = 0
+	VMStatusInvalid    = 1
+	VMStatusPanic      = 2
+	VMStatusUnfinished = 3
+)
+
+func (sw StateWitness) StateHash() common.Hash {
+	hash := crypto.Keccak256Hash(sw)
+	offset := 32*2 + 4*6
+	exitCode := sw[offset]
```
trianglesphere, line 89: `StateWitness` is a byte array - how do we know that it is as large as `offset`?

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
clabby: Good point. For now, we'll probably want games with `INVALID` or `PANIC` to be created. In alpha chad, we'll be able to change this to just be `VALID`.

specs/cannon-fault-proof-vm.md
clabby: 226 is the packed size of the state witness encoding - can be referenced in `EncodeWitness`: https://github.com/ethereum-optimism/optimism/blob/develop/cannon/mipsevm/state.go#L40-L62

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
Inphi: how would it change with Alpha Chad?

clabby: Alpha chad will change the purpose of the game to proving an output root correct rather than incorrect - in addition we'll have output bisection, so the root claim won't have a status byte (since it'll be the output root).

The first state hash claim (if it's by an honest actor, which we can guarantee by deliberately setting the split level), should have the VM status as valid always in that world. Not set in stone, though.

cannon/mipsevm/state.go
```diff
@@ -2,11 +2,16 @@ package mipsevm
...
 	"encoding/binary"
+	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/crypto"
 )
 
+// StateWitnessSize is the size of the state witness encoding in bytes.
+var StateWitnessSize = 226
```
ajsutton, line 13: Can we add a test to verify this is still accurate? Basically just encode a witness and assert the length == StateWitnessSize. Even if other tests would fail a specific test will make it a lot easier to understand why.

cannon/mipsevm/state.go
```diff
@@ -60,3 +69,41 @@ func (s *State) EncodeWitness() []byte {
...
+type StateWitness []byte
+
+const (
+	VMStatusValid      = 0
+	VMStatusInvalid    = 1
+	VMStatusPanic      = 2
+	VMStatusUnfinished = 3
+)
+
+func (sw StateWitness) StateHash() (common.Hash, error) {
```
ajsutton, line 82: We should add some unit tests to cover this.

ajsutton (commented):
Looks good just worth adding some unit tests for `StateHash`.

For the e2e tests I think we need to remove the cases in `TestChallengerCompleteDisputeGame` that are creating a game with the correct alphabet root as they will now be intermittent (they expect `step` to fail but it will actually succeed, the test just doesn't wait for it so sometimes passes incorrectly).

We should also pull in the changes from https://github.com/ethereum-optimism/optimism/compare/fpvm-status...aj/fpvm-status-e2e so that we can un-skip `TestCannonChallengeWithCorrectRoot` which is the key thing we're trying to fix with this PR.

ajsutton:
I could be convinced to un-skip `TestCannonChallengeWithCorrectRoot` in a follow up PR if you prefer. Have removed my request changes so it doesn't block merging given timezones. :)

ajsutton (approved):
After some further discussion it makes more sense to un-skip the e2e tests in a follow up.  So this is good to go I think.

ajsutton:
The alphabet e2e tests are actually fine and won't be intermittent. The opposing challenger (deliberately) has an incorrect trace so will fail to call step because its claims are wrong.  The case the alphabet trace doesn't handle is when you create a game with the correct trace you can't defeat it but that's not required for the alphabet trace (it's not inherently disputing a published output root like cannon is).

clabby (approved):

PR #6914, opened 2023-08-19
Optional L1 Starting Height for Indexer Syncing

### LINKED ISSUE #6794: Optional `L1StartingHeight` Param for OP Indexer
**Is your feature request related to a problem? Please describe.**
Currently, running the OP Indexer with a blank database results in the application indexing all layer-1 blocks from genesis. This is very time consuming and blocks the `L2Processor` from traversing to head until correlated L1 system contract bridge events can be detected. In some circumstances this could be unnecessary since all OP Stack chains have (L2->L1) genesis commitments much later in the Ethereum ledger. Ideally, an OP Indexer user should be able to specify some L1 starting height that's indicative of some L2 genesis on L1 and the point where the `L1Processor` begins traversal. 

**Describe the solution you'd like**
Add support for an optional indexer config value (`L1StartingHeight`) that specifies a starting l1 block height for which to begin indexing chain data from.

**Describe alternatives you've considered**
* You could bypass having to reindex from L1 genesis by providing a dummy row into the `l1_block_header` psql table with the L1 (block_number - 1) height that you'd like to start sequentially indexing from. This works but is rather counterintuitive and requires a few calculated steps to properly execute. Furthermore, this assumes that a user has arbitrary write access to the indexer managed database. 

**Additional context**

ethenotethan:
**Description**
Added support for a `l1-starting-height` indexer config value that allows a user to specify a L1 starting height to sync the indexer from. Once the L1 `SystemConfig` contract is upgraded, this value can be dynamically fetched from L1 using the preset address. 

More specifically, the following changes were applied:
* Moved mocks from test files to specific `mock.go` files per sub-packages
* Added mocking support for the `DB` struct that allows for database operations to be mocked
* Migrated conversion logic that transforms a `L1Contracts` struct to a slice into an independent receiver method with tests
* Added support for a `l1-starting-height` config value
* Added table tests for testing `L1ETL` construction logic
* Modified some mock function implementations to support nil returns
* Updated `README.md` with information about starting height feature

**Tests**
Added unit tests to validate the construction logic of the `L1ETL` struct. These tests assert that the proper starting height is fed to the `HeaderTraversal` provided specific parameter combinations to the constructor. 

**Additional context**

**Metadata**

- Fixes #6794


indexer/config/config.go
```diff
@@ -34,15 +36,38 @@ type L1Contracts struct {
 	// Remove afterwards?
 }
 
+// converts struct of to a slice of addresses for easy iteration
+// also validates that all fields are addresses
+func (c *L1Contracts) AsSlice() ([]common.Address, error) {
```
hamdiallam, line 41: Rather than `AsSlice` how about a `ForEach` that takes a func? That way we can retain the logs on startup that spits out the configured (name, contract_address) pairs?

indexer/README.md
```diff
@@ -4,7 +4,7 @@
...
+The `indexer.toml` stores a set of preset environmental variables that can be used to run the indexer with the exception of the network specific `l1-rpc` and `l2-rpc` variables. The `indexer.toml` file can be ran as a default config, otherwise a custom `.toml` config can provided via the `--config` flag when running the application. An optional `l1-starting-height` value can be provided to the indexer to specify the L1 starting block height to begin indexing from. This should be ideally be an L1 block that holds a correlated L2 genesis transaction. If no starting height value is provided and the database is empty, the indexer will begin sequentially processing from L1 genesis.
```
hamdiallam: Maybe specify that it must be <= this expected height: "ideally be an L1 block that holds a correlated L2 genesis transaction"

Some integrity check will probably bork if the person supplies a height too large

indexer/indexer.toml
```diff
@@ -4,6 +4,8 @@
 # OP Goerli
 preset = 420
```
hamdiallam: rm this new line

indexer/etl/l1_etl.go
```diff
@@ -20,43 +19,49 @@ type L1ETL struct {
 	db *database.DB
 }
 
-func NewL1ETL(log log.Logger, db *database.DB, client node.EthClient, contracts config.L1Contracts) (*L1ETL, error) {
+// NewL1ETL creates a new L1ETL instance that will start indexing from different starting points
```
hamdiallam, line 22: /s/start/startingHeight

indexer/etl/l1_etl.go
```diff
@@ -20,43 +19,49 @@ type L1ETL struct {
...
+
+	} else { // latestHeader == nil
+		log.Info("no indexed state in storage, starting from supplied L1 height", "height", start.String())
+		header, err := client.BlockHeaderByNumber(start)
+		if err != nil {
+			return nil, fmt.Errorf("could not fetch starting block header: %w", err)
+		}
+
+		fromHeader = header
 	}
```
hamdiallam: The first part of this `else if` condition isn't needed since we know that `latestHeader == nil` after the first `if` check. How about this change:


```suggestion
	} else if start.Bitlen() > 0 {
	log.Info("no indexed state in storage, starting from supplied L1 height", "height", start.String())
		header, err := client.BlockHeaderByNumber(start)
		if err != nil {
			return nil, fmt.Errorf("could not fetch starting block header: %w", err)
		}

		fromHeader = header
	} else {
		log.Info("no indexed state in storage or starting height, starting from L1 genesis")
    }
```

Having the last case be the simplest case (genesis) reads a bit better imo

hamdiallam (approved):
LGTM! Left a few comments, 🎉 on the first external contribution!

indexer/etl/l1_etl.go
ethenotethan: Deferring to `startHeight` if that's cool

ethenotethan: [dd4f561](https://github.com/ethereum-optimism/optimism/pull/6914/commits/dd4f561552f16d8de1e09cc46f39079d2bdcb77d)

indexer/etl/l1_etl.go
ethenotethan: [dd4f561](https://github.com/ethereum-optimism/optimism/pull/6914/commits/dd4f561552f16d8de1e09cc46f39079d2bdcb77d)

indexer/indexer.toml
ethenotethan: cbed843

PR #6918, opened 2023-08-21
op-challenger: Use game specific data dir for cannon trace provider

ajsutton:
**Description**

Modifies the cannon trace provider to use a different subdirectory per-game so the different games don't interfere with each other.

**Tests**

Added unit tests. The multiple cannon games e2e test now passes.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4130/automatically-detect-new-games


refcell (approved):
LGTM

PR #6919, opened 2023-08-21
op-e2e: Capture return value when tracing failed tx

ajsutton:
**Description**

When capturing a tx trace because a receipt had an unexpected value, capture the return data as well.


ajsutton:
@clabby Should we capture memory here too or will a revert reason turn up in the return value?

Inphi (approved):

tynes (approved):

PR #6920, opened 2023-08-21
Cleanup stringer types with ~string generic

### LINKED ISSUE #6732: op-service: cleanup stringer types with `~string` generic
`Stringered` / `StringeredList` / `EnumString` are hacky. We can get rid of the `Stringer` stuff, and use `~string` (generic type, captures types of underlying string type) and use that to support `EnumString`. These helper funcs are used in things like the CLI flag construction, of which the usage will have to be updated.

CarterBloop:
**Description**

Cleanup stringer types with ~string generic. Replaced Stringered and StringeredList with ~string.

**Tests**

Removed Stringered and StringeredList tests because they are not needed.

**Additional context**

**Metadata**

- Fixes #6732


op-service/enum/enum.go
```diff
@@ -1,36 +1,18 @@
...
 	for i, v := range values {
-		out.WriteString(v.String())
+		out.WriteString(string(v))
 		if i+1 < len(values) {
 			out.WriteString(", ")
 		}
 	}
 	return out.String()
-}
+}
```
ajsutton: The linter seems to want this line ending back.  If you run:
```
gofmt -w op-service/enum/enum.go
```
it should fix up the formatting for you.

ajsutton:
Thanks - this looks good to me but will give a chance for @protolambda to review it to make sure it covers the full scope of what he was thinking for that ticket.

protolambda (approved):
Nice cleanup :+1:

PR #6924, opened 2023-08-21
op-challenger: Combine FaultDisputeGameCaller and Loader into a single wrapper

ajsutton:
**Description**

Putting this up for discussion.  The idea is to make a clearer anti-corruption layer to translate between the op-challenger and contracts bounded contexts. The translation layer makes the two contexts more decoupled so they can be more easily changed independently but also allows the challenger to use a cleaner set of APIs to interact with the contracts.

If we go all the way with this we should have a similar wrapper for the dispute game factory contract and replace any direct usage of the `bindings` module so they all go through this wrappers (which should make testing a lot easier too if we create one good stub for the wrapper).


ajsutton:
This did not pan out so well..

PR #6925, opened 2023-08-21
simplify: remove unreachable code

bryanchriswhite:
**Description**

Removes redundant error check conditional and its unreachable body.

**Tests**

N/A / un-coverable


refcell (approved):
LGTM

PR #6927, opened 2023-08-21
feat(op-challenger): Only Load Recent Games

refcell:
**Description**

Adds an optional `--game-window` flag (default value of 11 days in seconds) that
specifies the maximum amount of time in the past for games to interact with.

So, for example, if a game's timestamp is 12 days in the past, this is further
back than the default 11 day window, and the game will not be played by the challenger.
Any game created within than `<game window>` time interval in the past will be
returned by the game loader component so the challenger can play it.

The default 11 days is set so we provide a resolution time buffer of 4 days
(half the game duration rounded up to an even number) plus the full 7 day game duration.

**Metadata**

Fixes CLI-4358


refcell:
* develop
  * **PR #6927** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6927" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6927?utm_source=stack-comment).

op-challenger/flags/flags.go
```diff
@@ -108,6 +108,12 @@ var (
 		EnvVars: prefixEnvVars("CANNON_SNAPSHOT_FREQ"),
 		Value:   config.DefaultCannonSnapshotFreq,
 	}
+	GameWindowFlag = &cli.UintFlag{
```
trianglesphere: urfave CLI has a specific time duration flag which does nice parsing for different ways of specifying time

refcell: Awesome, good callout

op-challenger/cmd/main_test.go
```diff
@@ -233,6 +234,18 @@ func TestCannonSnapshotFreq(t *testing.T) {
...
 
+func TestGameWindow(t *testing.T) {
+	t.Run("UsesDefault", func(t *testing.T) {
+		cfg := configForArgs(t, addRequiredArgs(config.TraceTypeAlphabet))
+		require.Equal(t, config.DefaultGameWindow, cfg.GameWindow)
+	})
+
+	t.Run("Valid", func(t *testing.T) {
+		cfg := configForArgs(t, addRequiredArgs(config.TraceTypeAlphabet, "--game-window=60ns"))
+		require.Equal(t, time.Duration(int64(60)), cfg.GameWindow)
```
trianglesphere:
```suggestion
		require.Equal(t, time.Duration(60 * time.Nanosecond), cfg.GameWindow)
```

op-challenger/config/config.go
```diff
@@ -72,7 +73,14 @@ func ValidTraceType(value TraceType) bool {
...
 }
 
-const DefaultCannonSnapshotFreq = uint(1_000_000_000)
+const (
+	DefaultCannonSnapshotFreq = uint(1_000_000_000)
+	// DefaultGameWindow is the default maximum time duration in the past
+	// that the challenger will look for games to progress.
+	// The default value is 10.5 days, which is a 3.5 day resolution buffer
+	// plus the 7 day game finalization window.
+	DefaultGameWindow = time.Duration(int64(907_200))
```
trianglesphere:
```suggestion
	DefaultGameWindow = time.Duration(252 * time.Hour)
```

Side note: if we did a full number of days, we could do `11 * 24 * time.Hour`. Golang doesn't provide a duration greater than 1 hour. (Scroll to the bottom of this section: https://pkg.go.dev/time#pkg-constants)

op-challenger/fault/factory.go
```diff
@@ -56,14 +56,19 @@ func (l *gameLoader) FetchAllGamesAtBlock(ctx context.Context, blockNumber *big.
...
 
-	games := make([]FaultDisputeGame, gameCount.Uint64())
-	for i := uint64(0); i < gameCount.Uint64(); i++ {
-		game, err := l.caller.GameAtIndex(callOpts, big.NewInt(int64(i)))
+	games := make([]FaultDisputeGame, 0)
+	if gameCount.Uint64() == 0 {
+		return games, nil
+	}
+	for i := gameCount.Uint64(); i > 0; i-- {
+		game, err := l.caller.GameAtIndex(callOpts, big.NewInt(int64(i-1)))
```
trianglesphere, line 64: Very minor nit: I tend to prefer to do the off by one adjustment in the for loop when counting down, but you don't have to.
```suggestion
	for i := gameCount.Uint64() - 1; i >= 0; i-- {
		game, err := l.caller.GameAtIndex(callOpts, big.NewInt(int64(i)))
```

The trick of counting backwards here & relying on the sorted array is really nice.

op-challenger/fault/monitor.go
```diff
@@ -47,7 +49,8 @@ func (m *gameMonitor) progressGames(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to load current block number: %w", err)
 	}
-	games, err := m.source.FetchAllGamesAtBlock(ctx, new(big.Int).SetUint64(blockNum))
+	earliest := time.Now().Unix() - int64(m.gameWindow)
```
trianglesphere: Very important to not just cast `time.Duration`. Golang may let you, but it's a structured type & has accessor methods for lots of components on it.
```suggestion
	earliest := time.Now().Unix() - int64(m.gameWindow.Seconds())
```

op-challenger/fault/monitor_test.go
```diff
@@ -86,15 +87,15 @@ func setupMonitorTest(t *testing.T, allowedGame common.Address) (*gameMonitor, *
 	fetchBlockNum := func(ctx context.Context) (uint64, error) {
 		return 1234, nil
 	}
-	monitor := newGameMonitor(logger, fetchBlockNum, allowedGame, source, games.CreateGame)
+	monitor := newGameMonitor(logger, time.Duration(int64(0)), fetchBlockNum, allowedGame, source, games.CreateGame)
```
trianglesphere: Golang automatically casts untyped integer constants to the right type.
```suggestion
	monitor := newGameMonitor(logger, time.Duration(0), fetchBlockNum, allowedGame, source, games.CreateGame)
```

trianglesphere (changes requested):
High level logic looks good, but there are some bugs with the handling of the `time.Duration` type

op-challenger/config/config.go
refcell: Changed to 11 to keep the round number. For now the extra 0.5 days in the naive approach most likely won't make a difference and just leans toward being conservative in approach.

op-challenger/fault/factory.go
refcell: The problem of doing this is that in the last iteration of the loop, the `i` value will decrement from 0 and underflow, resulting in the loop continuing at the max uint64 value. I tried this out initially.

op-challenger/fault/monitor.go
refcell: That makes sense, thanks.

op-challenger/fault/monitor.go
```diff
@@ -50,7 +52,8 @@ func (m *gameMonitor) progressGames(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to load current block number: %w", err)
 	}
-	games, err := m.source.FetchAllGamesAtBlock(ctx, new(big.Int).SetUint64(blockNum))
+	earliest := time.Now().Unix() - int64(m.gameWindow.Seconds())
```
ajsutton: Personally, I'd have written this as:
```suggestion
	earliest := time.Now().Sub(m.gameWindow).Unix()
```

But feel free to ignore that.

ajsutton: Actually, this definitely should use a `Clock` rather than calling `time.Now()` directly. Then you can control time in unit tests as well.

ajsutton (commented):
For some reason the tests are very unhappy with this change - the challenger doesn't seem to be responding to games anymore.

refcell:
> For some reason the tests are very unhappy with this change - the challenger doesn't seem to be responding to games anymore.

Yep looks like it. Will debug and try to fix what's up.

op-challenger/fault/factory.go
trianglesphere: yea, that won't work with a uint.

ajsutton (approved):
LGTM.

trianglesphere (approved):
time.Duration handling looks correct

PR #6941, opened 2023-08-21
op-e2e: Increase gas limit when creating dispute games.

ajsutton:
**Description**

Manually double the gas limit when creating fault dispute games. Tests were flaky because this tx sometimes ran out of gas. We believe this occurred when a new output root was posted between estimating the gas and actually executing the transaction on chain, causing the search to use more gas.


mslipper (approved):

tynes (approved):

trianglesphere (approved):
op-e2e seems to have passed first try

PR #6948, opened 2023-08-22
op-challenger: Don't progress completed games

ajsutton:
**Description**

Don't attempt to perform any actions on games that are already completed. Avoids quite a few calls to the game contract and in particular avoids recreating cannon trace outputs to check if we agree with claims or not.

Still doesn't actually clean up resources on disk for completed games but does remove them from memory. Cleaning up resources on disk is under a separate ticket.

Since there's a not-insignificant cost to creating a `GamePlayer` (it has to load the local game inputs, checks the absolute prestate etc), the players are kept in memory until they are considered too old and not returned from the game factory loader (https://github.com/ethereum-optimism/optimism/pull/6927 will exclude old games from that).

**Tests**

Updated unit tests.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4370/stop-attempting-to-progress-completed-games

### DIFF
--- a/op-challenger/fault/monitor.go
+++ b/op-challenger/fault/monitor.go
@@ -66,17 +66,27 @@ func (m *gameMonitor) progressGames(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to load games: %w", err)
 	}
+	requiredGames := make(map[common.Address]bool)
 	for _, game := range games {
 		if !m.allowedGame(game.Proxy) {
 			m.logger.Debug("Skipping game not on allow list", "game", game.Proxy)
 			continue
 		}
+		requiredGames[game.Proxy] = true
 		player, err := m.fetchOrCreateGamePlayer(game)
 		if err != nil {
 			m.logger.Error("Error while progressing game", "game", game.Proxy, "err", err)
 			continue
 		}
 		player.ProgressGame(ctx)
+	}
+	// Remove the player for any game that's no longer being returned from the list of active games
+	for addr := range m.players {
+		if _, ok := requiredGames[addr]; ok {
+			// Game still required
+			continue
+		}
+		delete(m.players, addr)
 	}
 	return nil
 }
--- a/op-challenger/fault/player.go
+++ b/op-challenger/fault/player.go
@@ -30,6 +30,8 @@ type GamePlayer struct {
 	caller                  GameInfo
 	logger                  log.Logger
 	cleanup                 func() error
+
+	completed bool
 }
 
 func NewGamePlayer(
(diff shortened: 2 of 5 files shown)


op-challenger/fault/monitor.go
```diff
@@ -66,18 +66,28 @@ func (m *gameMonitor) progressGames(ctx context.Context) error {
...
 			m.logger.Debug("Skipping game not on allow list", "game", game.Proxy)
 			continue
 		}
+		requiredGames[game.Proxy] = true
 		player, err := m.fetchOrCreateGamePlayer(game)
 		if err != nil {
 			m.logger.Error("Error while progressing game", "game", game.Proxy, "err", err)
 			continue
 		}
 		player.ProgressGame(ctx)
```
refcell, line 95: I think this currently will only delete games that are not in the `allowedGame` list. Instead, we should be checking the bool return value of `ProgressGame` which will return true if the game is no longer in progress. We can then only add the game proxy address to the `requiredGames` mapping when the `ProgressGame` return value is false.

A way to do this would look be the following.
```suggestion
		if !m.allowedGame(game.Proxy) {
			m.logger.Debug("Skipping game not on allow list", "game", game.Proxy)
			continue
		}
		player, err := m.fetchOrCreateGamePlayer(game)
		if err != nil {
			m.logger.Error("Error while progressing game", "game", game.Proxy, "err", err)
			continue
		}
		requiredGames[game.Proxy] = !player.ProgressGame(ctx)
```

op-challenger/fault/player.go
```diff
@@ -94,6 +94,15 @@ func NewGamePlayer(
 }
 
 func (g *GamePlayer) ProgressGame(ctx context.Context) bool {
+	status, err := g.caller.GetGameStatus(ctx)
+	if err != nil {
+		g.logger.Warn("Unable to retrieve game status", "err", err)
+		return false
+	}
```
refcell: We are essentially doing this twice in this function (before and after `Act`ing) so maybe we can refactor it out into a `getGameStatus` method on the `GamePlayer` that can dedupe the logging at least.

refcell: Actually, looking at the logic closer, maybe we can just perform the status check before acting. Since it exits early, it will only add an iteration to a loop, but remove an rpc call which should be a worthwhile tradeoff.

op-challenger/fault/monitor.go
ajsutton: If we do it that way the next update will create the player again because it's still in the list of games to progress. Since creating the player involves a bunch of calls but it only uses a small amount of memory I think we want to keep the player around in memory for as long as its present in the list of games we care about, rather than drop it from memory when `ProgressGame` returns true.

We can still delete the data from disk when `ProgressGame` returns true, we just need something to remember the details of the game to save us loading them all on every update.

op-challenger/fault/player.go
ajsutton: I'm not entirely sure what you mean here - in my mind this _is_ the code that performs the status check before acting and exits early.

op-challenger/fault/monitor.go
refcell: Ah got it - I didn't intuit the on disk / in memory resource split here until our sync call. That makes sense to me!

op-challenger/fault/player.go
ajsutton: Discussed this separately. There's a couple of factors here:
1. We would ideally like to avoid extra RPC calls
2. We want to log the final game result once and then stop logging anything about the game

So going to change this to cache the completed status rather than requesting the status before and after acting.

refcell (approved):
Looks good!

PR #6951, opened 2023-08-22
feat(op-challenger): Game Player and Cannon Cleanup

refcell:
**Description**

Adds a cleanup method to cannon that deletes the game-specific cannon directory.

Also exposes a cleanup method on the game player which calls down to the cannon
trace provider's cleanup method. This is to expose cleanup to the monitor component.

Monitor component wiring should follow #6948 since that pr implements game player
deletion inside the monitor component. A followup pr should ensure a game player is
cleaned up before it's "deletion" to avoid resource leakage.

**Metadata**

Fixes CLI-4373


refcell:
* develop
  * **PR #6951** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6951" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6951?utm_source=stack-comment).

ajsutton (approved):
Looks good.  As discussed elsewhere, we probably want to refactor a bit so that the trace provider isn't responsible for using a unique data dir, but worth getting this in and things working first since we have most of the code already and then we can refactor and see what it looks like.

PR #6952, opened 2023-08-22
op-wheel: add engine-api utils

protolambda:
**Description**

This adds two new utils to `op-wheel`:
- set forkchoice by block number
- interact with engine API through raw JSON RPC call, like `cast`, but with engine-API JWT support

Usage:
```
# build
go build -o ~/home/go/bin/op-wheel ./op-wheel/cmd

op-wheel engine set-forkchoice --engine=http://localhost:8551 --engine.jwt-secret=ops-bedrock/test-jwt-secret.txt  --finalized=1000 --safe=2000 --unsafe=3000

# cast-like usage
op-wheel engine json --engine=http://localhost:8551 --engine.jwt-secret=ops-bedrock/test-jwt-secret.txt eth_getBlockByNumber latest false

# or use std-in (useful for large payloads, e.g. block building)
echo \"latest\" false | op-wheel engine json --engine=http://localhost:8551 --engine.jwt-secret=ops-bedrock/test-jwt-secret.txt --stdin eth_getBlockByNumber

# and use the env-vars for repeat usage of the same engine API endpoint and JWT secret file
```

These are tools for debugging / testing / ops.


ajsutton (approved):
LGTM.

PR #6953, opened 2023-08-22
fix(op-challenger): Touchup Responder Test Suite

refcell:
**Description**

In preparation to redo #6744, this pr rips out some unaffiliated changes
that make its implementation cleaner and a smaller diff.

PR #6956, opened 2023-08-22
fix(op-challenger): Only Resolve Won Dispute Games

refcell:
**Description**

Replaces PR #6744.

Only resolves dispute games which the challenger won.

In order to allow the `Agent` fault component to retrieve
the game status, the `Caller` has to be passed in during
construction by the `service.go` initialization function.

**Tests**

Introduces minimal agent unit tests for a newly exported
`ShouldResolve` method.

**Metadata**

Fixes CLI-4285

### DIFF
--- a/op-challenger/fault/agent.go
+++ b/op-challenger/fault/agent.go
@@ -13,7 +13,7 @@ import (
 // Responder takes a response action & executes.
 // For full op-challenger this means executing the transaction on chain.
 type Responder interface {
-	CanResolve(ctx context.Context) bool
+	CallResolve(ctx context.Context) (types.GameStatus, error)
 	Resolve(ctx context.Context) error
 	Respond(ctx context.Context, response types.Claim) error
 	Step(ctx context.Context, stepData types.StepCallData) error
@@ -65,10 +65,27 @@ func (a *Agent) Act(ctx context.Context) error {
 	return nil
 }
 
+// shouldResolve returns true if the agent should resolve the game.
+// This method will return false if the game is still in progress.
+func (a *Agent) shouldResolve(ctx context.Context, status types.GameStatus) bool {
+	expected := types.GameStatusDefenderWon
+	if a.agreeWithProposedOutput {
+		expected = types.GameStatusChallengerWon
+	}
+	if expected != status {
+		a.log.Warn("Game will be lost", "expected", expected, "actual", status)
+	}
+	return expected == status
+}
+
 // tryResolve resolves the game if it is in a terminal state
 // and returns true if the game resolves successfully.
 func (a *Agent) tryResolve(ctx context.Context) bool {
-	if !a.responder.CanResolve(ctx) {
+	status, err := a.responder.CallResolve(ctx)
+	if err != nil {
+		return false
+	}
+	if !a.shouldResolve(ctx, status) {
 		return false
 	}
 	a.log.Info("Resolving game")
--- a/op-challenger/fault/agent_test.go
+++ b/op-challenger/fault/agent_test.go
@@ -1,0 +1,31 @@
+package fault
+
+import (
+	"context"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ethereum-optimism/optimism/op-challenger/fault/types"
+	"github.com/ethereum-optimism/optimism/op-node/testlog"
+)
+
+// TestShouldResolve tests the resolution logic.
+func TestShouldResolve(t *testing.T) {
+	log := testlog.Logger(t, log.LvlCrit)
+
+	t.Run("AgreeWithProposedOutput", func(t *testing.T) {
+		agent := NewAgent(nil, 0, nil, nil, nil, true, log)
+		require.False(t, agent.shouldResolve(context.Background(), types.GameStatusDefenderWon))
+		require.True(t, agent.shouldResolve(context.Background(), types.GameStatusChallengerWon))
+		require.False(t, agent.shouldResolve(context.Background(), types.GameStatusInProgress))
+	})
+
+	t.Run("DisagreeWithProposedOutput", func(t *testing.T) {
+		agent := NewAgent(nil, 0, nil, nil, nil, false, log)
+		require.True(t, agent.shouldResolve(context.Background(), types.GameStatusDefenderWon))
+		require.False(t, agent.shouldResolve(context.Background(), types.GameStatusChallengerWon))
+		require.False(t, agent.shouldResolve(context.Background(), types.GameStatusInProgress))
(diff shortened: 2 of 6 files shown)


refcell:
* develop
  * **PR #6953** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6953" target="_blank">Graphite</a> 
    * **PR #6956** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6956" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6956?utm_source=stack-comment).

op-challenger/fault/agent.go
```diff
@@ -65,10 +65,27 @@ func (a *Agent) Act(ctx context.Context) error {
 	return nil
 }
 
+// ShouldResolve returns true if the agent should resolve the game.
+// This method will return false if the game is still in progress.
+func (a *Agent) ShouldResolve(ctx context.Context, status types.GameStatus) bool {
```
ajsutton: nit: This looks like it should be a private function rather than exported.

op-challenger/fault/agent.go
```diff
@@ -65,10 +65,27 @@ func (a *Agent) Act(ctx context.Context) error {
...
+	expected := types.GameStatusDefenderWon
+	if a.agreeWithProposedOutput {
+		expected = types.GameStatusChallengerWon
+	}
+	if expected != status {
+		a.log.Warn("game status mismatch", "expected", expected, "actual", status)
```
ajsutton:
```suggestion
		a.log.Warn("Game will be lost", "expected", expected, "actual", status)
```

Mostly to start with a capital but I think we also want the message to make it clear that the game is effectively lost.

op-challenger/fault/agent_test.go
```diff
@@ -0,0 +1,31 @@
...
+		require.False(t, agent.ShouldResolve(context.Background(), types.GameStatus(2)))
+		require.True(t, agent.ShouldResolve(context.Background(), types.GameStatus(1)))
+		require.False(t, agent.ShouldResolve(context.Background(), types.GameStatus(0)))
+	})
+
+	t.Run("DisagreeWithProposedOutput", func(t *testing.T) {
+		agent := NewAgent(nil, 0, nil, nil, nil, false, log)
+		require.True(t, agent.ShouldResolve(context.Background(), types.GameStatus(2)))
+		require.False(t, agent.ShouldResolve(context.Background(), types.GameStatus(1)))
+		require.False(t, agent.ShouldResolve(context.Background(), types.GameStatus(0)))
```
ajsutton: We should use the `GameStatusInProgress` etc constants instead of casting ints.

op-challenger/fault/responder.go
```diff
@@ -79,18 +79,25 @@ func (r *faultResponder) BuildTx(ctx context.Context, response types.Claim) ([]b
...
-// CanResolve determines if the resolve function on the fault dispute game contract
-// would succeed. Returns true if the game can be resolved, otherwise false.
-func (r *faultResponder) CanResolve(ctx context.Context) bool {
+// CallResolve determines if the resolve function on the fault dispute game contract
+// would succeed. Returns the game status if the call would succeed, errors otherwise.
+func (r *faultResponder) CallResolve(ctx context.Context) (types.GameStatus, error) {
 	txData, err := r.buildResolveData()
 	if err != nil {
-		return false
+		return types.GameStatus(0), err
```
ajsutton:
```suggestion
		return types.GameStatusInProgress, err
```

op-challenger/fault/responder.go
```diff
@@ -79,18 +79,25 @@ func (r *faultResponder) BuildTx(ctx context.Context, response types.Claim) ([]b
...
 	}
-	_, err = r.txMgr.Call(ctx, ethereum.CallMsg{
+	res, err := r.txMgr.Call(ctx, ethereum.CallMsg{
 		To:   &r.fdgAddr,
 		Data: txData,
 	}, nil)
-	return err == nil
+	if err != nil {
+		return types.GameStatus(0), err
```
ajsutton:
```suggestion
		return types.GameStatusInProgress, err
```

op-challenger/fault/responder.go
```diff
@@ -79,18 +79,25 @@ func (r *faultResponder) BuildTx(ctx context.Context, response types.Claim) ([]b
...
+	}
+	var status uint8
+	if err = r.fdgAbi.UnpackIntoInterface(&status, "resolve", res); err != nil {
+		return types.GameStatus(0), err
```
ajsutton:
```suggestion
		return types.GameStatusInProgress, err
```

op-challenger/fault/responder_test.go
```diff
@@ -25,18 +25,31 @@ var (
...
-func TestCanResolve(t *testing.T) {
+// TestCallResolve tests the [Responder.CallResolve].
+func TestCallResolve(t *testing.T) {
 	t.Run("SendFails", func(t *testing.T) {
 		responder, mockTxMgr := newTestFaultResponder(t)
 		mockTxMgr.callFails = true
-		require.False(t, responder.CanResolve(context.Background()))
+		status, err := responder.CallResolve(context.Background())
+		require.ErrorIs(t, err, mockCallError)
+		require.Equal(t, types.GameStatus(0), status)
```
ajsutton: I'd either stick to using the GameStatus constants through this test or just not assert the returned value (since the caller shouldn't use the result if an error is being returned).

ajsutton (commented):
Looks good generally just little details.

op-challenger/fault/agent.go
refcell: This reads nicely to me

op-challenger/fault/agent_test.go
refcell: Good point, thanks!

op-challenger/fault/types/types.go
```diff
@@ -34,6 +34,20 @@ func (s GameStatus) String() string {
...
+	case 0:
+		return GameStatusInProgress, nil
+	case 1:
+		return GameStatusChallengerWon, nil
+	case 2:
+		return GameStatusDefenderWon, nil
+	default:
+		return GameStatus(0), errors.New("invalid game status")
+	}
+}
```
ajsutton: I think this is where a cast is fine since the values used for the constants should match up with what the contract uses. Having a dedicated method to do the conversion is probably a good idea, but it can just do the cast - it still provides a single point to handle any difference between the contract values and go ones if needed in the future.

ajsutton (approved):
Looks good, just the one suggestion for simplification.

ajsutton (approved):

PR #6965, opened 2023-08-22
op-challenger: Delete game files when complete

ajsutton:
**Description**

Once a dispute game is resolved, cleanup its files from disk to save disk space.

**Tests**

Updated monitor unit tests.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4374/wire-cleanup-into-the-monitor-component

### DIFF
--- a/op-challenger/fault/monitor.go
+++ b/op-challenger/fault/monitor.go
@@ -13,6 +13,7 @@ import (
 
 type gamePlayer interface {
 	ProgressGame(ctx context.Context) bool
+	Cleanup() error
 }
 
 type playerCreator func(address common.Address) (gamePlayer, error)
@@ -78,7 +79,15 @@ func (m *gameMonitor) progressGames(ctx context.Context) error {
 			m.logger.Error("Error while progressing game", "game", game.Proxy, "err", err)
 			continue
 		}
-		player.ProgressGame(ctx)
+		done := player.ProgressGame(ctx)
+		if done {
+			// Remove resources on disk as soon as the game is complete to save disk space.
+			// We keep the player in memory to avoid recreating it on every update but will no longer
+			// need the resources on disk because there are no further actions required on the game.
+			if err := player.Cleanup(); err != nil {
+				m.logger.Error("Unable to cleanup player data", "err", err)
+			}
+		}
 	}
 	// Remove the player for any game that's no longer being returned from the list of active games
 	for addr := range m.players {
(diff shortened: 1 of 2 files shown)


op-challenger/fault/monitor.go
```diff
@@ -78,7 +79,15 @@ func (m *gameMonitor) progressGames(ctx context.Context) error {
 			m.logger.Error("Error while progressing game", "game", game.Proxy, "err", err)
 			continue
 		}
-		player.ProgressGame(ctx)
+		done := player.ProgressGame(ctx)
+		if done {
+			// Remove resources on disk as soon as the game is complete to save disk space.
+			// We keep the player in memory to avoid recreating it on every update but will no longer
+			// need the resources on disk because there are no further actions required on the game.
```
refcell, line 100: Good idea to add a comment, very helpful

op-challenger/fault/monitor.go
```diff
@@ -78,7 +79,15 @@ func (m *gameMonitor) progressGames(ctx context.Context) error {
 			m.logger.Error("Error while progressing game", "game", game.Proxy, "err", err)
 			continue
 		}
-		player.ProgressGame(ctx)
+		done := player.ProgressGame(ctx)
+		if done {
```
refcell, line 97: Is this done to make it explicit what the return value is indicating here (the game being "done")?

If it's not important, could probably just make this `if player.ProgressGame(ctx) { ...`

op-challenger/fault/monitor.go
```diff
@@ -78,7 +79,15 @@ func (m *gameMonitor) progressGames(ctx context.Context) error {
...
+			// Remove resources on disk as soon as the game is complete to save disk space.
+			// We keep the player in memory to avoid recreating it on every update but will no longer
+			// need the resources on disk because there are no further actions required on the game.
+			if err := player.Cleanup(); err != nil {
+				m.logger.Error("Unable to cleanup player data", "err", err)
```
refcell, line 102: Interesting, although this shouldn't fail, if there's a bug, it's going to keep printing this log every time the game is attempted to progress (concretely, every time the `progressGames` method on the monitor is called) because the game player will always return `true` from the `done := player.ProgressGame(ctx)` call.

op-challenger/fault/monitor.go
ajsutton: Yeah I think the named variable is useful for clarity.  `ProgressGame` doesn't tell you anything about what the boolean it returns means so it seems useful to name it.

op-challenger/fault/monitor.go
ajsutton: Yeah that will be messy but I think spamming the logs is probably better than failing to clean up once and giving up and just leaking the disk space.

refcell: Agree - it's explicitly an error as well, not like we're cluttering logs with dismiss-able warnings

refcell (approved):

PR #6966, opened 2023-08-22
op-challenger: Make datadir required for all trace types

ajsutton:
**Description**

Change the `--canon-datadir` argument to `--datadir` and make it required for all trace types.  This is the first step to doing the top level data directory management in the "multi-game" code rather than requiring each trace provider implementation to do its own isolation between games.

The downside is that a datadir is now required for alphabet games, but that seems like a reasonable trade-off given it's just a test trace type and every "real" trace provider will almost certainly need to cache results somewhere.

**Tests**

Updated unit and e2e tests.

**Metadata**

- https://linear.app/optimism/issue/CLI-4343/remove-cannon-datadirs-of-completed-games


op-challenger/fault/cannon/executor_test.go
```diff
@@ -21,10 +21,9 @@ const execTestCannonPrestate = "/foo/pre.json"
 
 func TestGenerateProof(t *testing.T) {
 	input := "starting.json"
-	cfg := config.NewConfig(common.Address{0xbb}, "http://localhost:8888", config.TraceTypeCannon, true)
 	tempDir := t.TempDir()
 	dir := filepath.Join(tempDir, "gameDir")
-	cfg.CannonDatadir = tempDir
+	cfg := config.NewConfig(common.Address{0xbb}, "http://localhost:8888", config.TraceTypeCannon, true, dir)
```
refcell: In the previous code, we were effectively setting the cannon data directory to the temp directory.

But here it looks like we now pass in the `dir` which is `filepath.Join(tempDir, "gameDir")` as opposed to just the `tempDir` value. Is this correct?

op-challenger/config/config_test.go
```diff
@@ -99,10 +98,10 @@ func TestCannonAbsolutePreStateRequired(t *testing.T) {
 	require.ErrorIs(t, config.Check(), ErrMissingCannonAbsolutePreState)
 }
 
-func TestCannonDatadirRequired(t *testing.T) {
-	config := validConfig(TraceTypeCannon)
-	config.CannonDatadir = ""
-	require.ErrorIs(t, config.Check(), ErrMissingCannonDatadir)
+func TestDatadirRequired(t *testing.T) {
+	config := validConfig(TraceTypeAlphabet)
```
refcell, line 102: For the other required config values, we use the `TraceTypeCannon`, is there a reason to use the alphabet trace type instead here?

refcell:
Just a couple small comments/questions. Otherwise, looks great. Very nice that this removes an additional optional cannon flag validation since it is checked in the required loop.

op-challenger/config/config_test.go
ajsutton: It's pretty arbitrary, but the old behaviour was to require it for cannon and not alphabet, so went with alphabet here to prove it was required where it previously wasn't.

op-challenger/fault/cannon/executor_test.go
ajsutton: Good spot - we should be passing `tempDir` here.  The point is to have a different directory in the config to what we pass in to the `GenerateProof` call to ensure the executor doesn't just use the value from the config. Will fix.

refcell (approved):
LGTM 👑

PR #6967, opened 2023-08-23
fix(style guide): document custom directives

sambacha:
Adds additional custom directive that is used.

**Description**

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/packages/contracts-bedrock/STYLE_GUIDE.md
+++ b/packages/contracts-bedrock/STYLE_GUIDE.md
@@ -20,6 +20,7 @@ with additional rules. These are:
 
 We also have the following custom tags:
 
+- `@custom:predeploy`: Add to a contract whenever its been pre-deployed.
 - `@custom:proxied`: Add to a contract whenever it's meant to live behind a proxy.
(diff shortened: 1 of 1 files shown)


packages/contracts-bedrock/STYLE_GUIDE.md
```diff
@@ -20,6 +20,7 @@ with additional rules. These are:
 
 We also have the following custom tags:
 
+- `@custom:predeploy`: Add to a contract whenever its been pre-deployed.
```
refcell, line 23: nit
```suggestion
- `@custom:predeploy`: Add to a contract whenever it's been pre-deployed.
```

tynes: > The difference between its and it's
Here's the rub: its (without an apostrophe) is a possessive pronoun, like his or her, for nouns that don't have a defined gender. In contrast, it's (with an apostrophe) is the shortened form, or contraction, of it is or it has

Seems right?

PR #6969, opened 2023-08-23
Update to golang 1.20.7 and alpine 3.18

ajsutton:
**Description**

Updates docker images to use go 1.20.7 and use alpine 3.18 as a base.

### DIFF
--- a/endpoint-monitor/Dockerfile
+++ b/endpoint-monitor/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.19.9-alpine3.16 as builder
+FROM golang:1.20.7-alpine3.18 as builder
 
 RUN apk --no-cache add make jq bash git alpine-sdk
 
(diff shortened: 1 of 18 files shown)


tynes:
I don't see a diff to the `go.mod` which is the source of truth for what version is used. It will still use an older version if a newer compiler version is installed because the `go.mod` is the source of truth

ajsutton:
> I don't see a diff to the `go.mod` which is the source of truth for what version is used. It will still use an older version if a newer compiler version is installed because the `go.mod` is the source of truth

Today I learned... Will go change all the go.mod's I can find too.

tynes:
Looks like some static analysis issues now

tynes:
Also any reason to not do the latest version? 1.21.0?

ajsutton:
> Also any reason to not do the latest version? 1.21.0?

Because go-libp2p doesn't work with 1.21 yet.

refcell (approved):
LGTM

tynes (approved):

roninjin10 (approved):

mslipper (approved):
Proto/Josh review only needed for a trivial test. Approving.

PR #6710, opened 2023-08-11
Migrate bedrock-devnet to go

refcell:
**Description**

Migrates and refactors the `bedrock-devnet/` to golang
in a new `ops-devnet` package.

PR #6972, opened 2023-08-23
make usage of docker compose v2 explicit

trajan0x:
**Description**

Since docker-compose v2, the command has changed from `docker-compose` to `docker compose` [see docs migration guide](https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose)

I've switched the various references to docker-compose to make this more explicit. This also makes the package more usable since attempting to install or use docker-compose on some systems `apt`,`deb`, etc will lead to the installation of the compose v1 api which will not be compatible with this docker file (see notice  [here](https://docs.docker.com/compose/install/linux/))

If this pr is merged, these changes should be made in ethereum-optimism/optimism#6710 as well

**Tests**

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,7 +119,7 @@ Note that these environment variables significantly speed up build time.
 cd ops-bedrock
 export COMPOSE_DOCKER_CLI_BUILD=1
 export DOCKER_BUILDKIT=1
-docker-compose build
+docker compose build
 ```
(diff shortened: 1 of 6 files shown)


tynes (approved):
Big fan of this, thank you

roninjin10 (approved):
til didn't even know docker compose existed yet alone was the preferred way to run docker-compose now

PR #6974, opened 2023-08-23
feat(op-challenger): Progress Games on New L1 Block

refcell:
**Description**

Updates the monitor component to only progress games on new l1 block.

**Metadata**

 CLI-4371.


refcell:
* develop
  * **PR #6974** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6974" target="_blank">Graphite</a>  👈
    * **PR #6993** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6993" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6974?utm_source=stack-comment).

op-challenger/fault/monitor.go
```diff
@@ -147,13 +143,23 @@ func (m *gameMonitor) fetchOrCreateGamePlayer(gameData FaultDisputeGame) (gamePl
...
+			if err != nil {
+				m.logger.Error("Failed to load current block number", "err", err)
+				continue
+			}
+			if nextBlockNum > blockNum {
+				blockNum = nextBlockNum
+				if err := m.progressGames(ctx, nextBlockNum); err != nil {
+					m.logger.Error("Failed to progress games", "err", err)
+				}
+			}
```
ajsutton, line 162: This is now a tight loop constantly polling the L1 head which isn't great. We should put the sleep back in (and probably make it sleep for say 1 second rather than 300ms).

With the sleep back in we can probably merge this as-is, but I'd probably keep the ticket open as we should switch to using a subscription to track the head (can reuse the same code as op-node uses to track L1 head) to avoid constantly polling. We've avoided subscriptions for log events because of the risk of missing one, but that's not an issue for L1 head because we'll just get the next head update and automatically catch everything up again.

refcell: Sounds good to me. I'm happy to follow up with a subscription based pr.

ajsutton (approved):

PR #6993, opened 2023-08-23
feat(op-challenger): Subscription Monitoring

refcell:
**Description**

Updates the game monitoring to use a new head subscription.

**Tests**

Unit tests are performed around the tight channel loop in `MonitorGames`
using a mock subscription.

**Metadata**

Fixes CLI-4371.

### DIFF
--- a/op-challenger/fault/monitor.go
+++ b/op-challenger/fault/monitor.go
@@ -7,7 +7,11 @@ import (
 	"time"
 
 	"github.com/ethereum-optimism/optimism/op-service/clock"
+	"github.com/ethereum-optimism/optimism/op-service/eth"
+
+	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
 )
 
@@ -38,6 +42,8 @@ type gameMonitor struct {
 	fetchBlockNumber blockNumberFetcher
 	allowedGames     []common.Address
 	players          map[common.Address]gamePlayer
+	l1HeadsSub       ethereum.Subscription
+	l1Source         eth.NewHeadSource
 }
 
 func newGameMonitor(
@@ -49,6 +55,7 @@ func newGameMonitor(
 	allowedGames []common.Address,
 	source gameSource,
 	createGame playerCreator,
+	l1Source eth.NewHeadSource,
 ) *gameMonitor {
 	return &gameMonitor{
 		logger:           logger,
@@ -60,6 +67,7 @@ func newGameMonitor(
 		fetchBlockNumber: fetchBlockNumber,
 		allowedGames:     allowedGames,
 		players:          make(map[common.Address]gamePlayer),
+		l1Source:         l1Source,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


refcell:
* develop
  * **PR #6974** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6974" target="_blank">Graphite</a> 
    * **PR #6993** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6993" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6993?utm_source=stack-comment).

op-challenger/fault/monitor.go
```diff
@@ -143,25 +150,20 @@ func (m *gameMonitor) fetchOrCreateGamePlayer(gameData FaultDisputeGame) (gamePl
...
-	blockNum := uint64(0)
+	headers, sub, err := m.headSubscriber()
+	if err != nil {
+		return fmt.Errorf("failed to subscribe to new headers: %w", err)
+	}
+
 	for {
 		select {
+		case err := <-sub.Err():
+			return fmt.Errorf("subscription error: %w", err)
```
ajsutton: I think we need to log this and reconnect the subscription. Its pretty common for a subscription to get disconnected and we don't want to crash in that situation.

op-challenger/fault/monitor_test.go
```diff
@@ -103,10 +105,76 @@ func TestMonitorOnlyCreateSpecifiedGame(t *testing.T) {
...
+				Timestamp: 9999,
+			},
+		}
+
+		var err error
+		go func() {
+			err = monitor.MonitorGames(context.Background())
+		}()
+		mockSub.errChan <- ethereum.NotFound
+		require.ErrorIs(t, err, ethereum.NotFound)
```
ajsutton: Also, this is a race condition depending on how quickly the go func winds up assigning to err.  I think you need a WaitGroup to allow the go func to flag when all its work is done.

op-challenger/fault/monitor.go
refcell: Agreed, just wanted to elicit your review for the overall setup first before going too deep in the implementation.

op-challenger/fault/monitor_test.go
refcell: Good callout hadn't paid attention to that.

op-challenger/fault/monitor_test.go
```diff
@@ -103,10 +105,76 @@ func TestMonitorOnlyCreateSpecifiedGame(t *testing.T) {
...
+			{
+				Proxy:     addr2,
+				Timestamp: 9999,
+			},
+		}
+
+		var err error
+		go func() {
+			err = monitor.MonitorGames(context.Background())
+		}()
```
ajsutton: Something will need to trigger this function to exit and then wait for it to complete or we'll leak resources.

op-service/retry/wrapper.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"context"
+)
+
+// Wrap will continue to retry the provided function unless the
+// boolean return value is false, indicating that it should not
+// be retried. For example, if the context inside the wrapped
+// function is cancelled, it should return false so Wrap will
+// not retry it with its cancelled context.
+func Wrap(ctx context.Context, f func() (bool, error)) error {
```
ajsutton: This doesn't seem to use the ctx arg.

ajsutton (commented):
This looks ok but the e2e tests are all failing which suggests there's a problem somewhere.  I'm a bit nervous about reinventing the persistent subscription logic that op-node already has and wonder if there's some subtle detail we've missed that the existing code would handle if we reused it.

refcell:
Reimplementing in #7073

PR #6992, opened 2023-08-23
chore(indexer): Rename (l1/l2)_tokens table

roninjin10:
- rename tables to avoid collision with old indexer

PR #6994, opened 2023-08-24
feat(indexer): add indexer all command

roninjin10:
Add indexer all command
- Convenient to run in dev mode quickly without having to spin up 2 processes
- Also convenient for initial deployment to make it look more like the old indexer as we migrate


roninjin10:
* develop
  * **PR #6992** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6992" target="_blank">Graphite</a> 
    * **PR #6994** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6994" target="_blank">Graphite</a>  👈
      * **PR #6995** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6995" target="_blank">Graphite</a> 
        * **PR #6996** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6996" target="_blank">Graphite</a> 
          * **PR #7001** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7001" target="_blank">Graphite</a> 
            * **PR #7002** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7002" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/6994?utm_source=stack-comment).

indexer/cmd/indexer/cli.go
```diff
@@ -76,6 +76,18 @@ func runApi(ctx *cli.Context) error {
 	return api.Listen(apiCtx, cfg.API.Port)
 }
 
+func runAll(ctx *cli.Context) error {
```
hamdiallam, line 79: should we follow the pattern in `Indexer.Run()`. Where if either process errors out, it shuts down the other (by sharing a parent context that's cancelled).

Right now if the indexer stops running, the api server never halts. And if the API server crashes, then the indexer process isn't gracefully closed (which technically is okay)

hamdiallam (approved):
Since both process will clutter the logs, we should add a default "role" param to each logger.

For example
```go
func runIndexer(ctx *cli.Context) error {
	logger := log.NewLogger(log.ReadCLIConfig(ctx)).New("role", "indexer")
	cfg, err := config.LoadConfig(logger, ctx.String(ConfigFlag.Name))
```

That way kinda like how the unit tests show the role for the batcher/sequencer/verifier nodes, we can see indexer/api logs seperately

indexer/cmd/indexer/cli.go
roninjin10: Nice! Yea definitely I wanted to do that and only didn't here because I didn't know how but see Indexer.Run() has an example of how

roninjin10: Doing in a different pr since it will need a review that I did it right

PR #6998, opened 2023-08-24
op-challenger: Spike for how to support parallel game playing

ajsutton:
**Description**

Putting this up for visibility but it's definitely not ready to merge. I think the e2e tests will pass (actually looks like there's either a bug or a race condition with deleting the files) so it seems to work but unit tests haven't been updated and the new game scheduler is still completely untested. The wiring isn't all in place either - most notably the goroutines spawned aren't ever stopped.

The design will likely have to change to make things more testable but this is a useful spike to work through how parallel game execution can work.

There's a few properties we want to preserve when operating in parallel:

* The game-specific code shouldn't need to deal with concurrency. The multi-game coordination should ensure that only one thread is ever progressing a game at a time
* There needs to be a limit on the number of concurrent games played
* Delete files on disk once a game is resolved
* Avoid recreating the `gamePlayer` for games (as it requires a bunch of calls to load game inputs even if the game is resolved)
* It shouldn't wait for all games to finishes progressing before moving on to handle later updates. For example game A might need to generate the initial trace and so could take an hour to complete its update. If at the same time game B can perform its next move quickly because the snapshots are available, game B might be able to perform multiple moves by the time game A generates its trace.

Will get this documented in comments for the final implementation, but the high level design here is:

* The monitor component periodically loads the list of games and filters them with the allow list like it does today.
* The monitor then submits the games to play as a batch to the `GameScheduler`
* The `GameScheduler` routes those games to its single "coordinator" thread which tracks the game state and controls when files are deleted.
* The coordinator updates state to remove any games it was tracking that are not in the current list (as they've expired and are no longer relevant). It filters out any games that already have an update job in flight, then queues the others to be updated in a channel
* A configurable number of "worker" threads read from the channel of games to update and they call the existing `ProgressGame` method that goes off into the single-game area of code.
* The worker thread sends the result of the update (whether the game is resolved or not) back to the coordinator thread via a channel
* The coordinator processes those results and updates the game state. It then deletes the files for any games that are resolved and not currently scheduled for an update.

**Tests**

**Additional context**

**Metadata**

- https://linear.app/optimism/issue/CLI-4342/support-parallel-games

### DIFF
--- a/op-challenger/fault/disk.go
+++ b/op-challenger/fault/disk.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
 	"golang.org/x/exp/slices"
 )
 
@@ -15,11 +16,15 @@ const gameDirPrefix = "game-"
 
 // diskManager coordinates the storage of game data on disk.
 type diskManager struct {
+	logger  log.Logger
 	datadir string
 }
 
-func newDiskManager(dir string) *diskManager {
-	return &diskManager{datadir: dir}
+func newDiskManager(logger log.Logger, dir string) *diskManager {
+	return &diskManager{
+		logger:  logger,
+		datadir: dir,
+	}
 }
 
 func (d *diskManager) DirForGame(addr common.Address) string {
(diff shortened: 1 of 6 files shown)


refcell:
Definitely going to need to think about this a little bit and all the (non-fearless) concurrency implications here.

Some specs and architectural/design mockups would be very helpful as well. I think we should aim to produce some high level view of the challenger that shouldn't be affect by changes in the implementation details. That way we can have better intuition around the implementation goalposts.

PR #6999, opened 2023-08-24
op-e2e: Wait for game data to be deleted

ajsutton:
**Description**

Fixes a race condition which could cause flaky e2e tests. The challenger should delete game data _after_ the game is resolved but that takes time and happens asynchronously so it's important for the e2e test to wait rather than just doing a one-off check if the data has been deleted.


Inphi (approved):

Inphi:
Would be better to fix the asynchronous cleanup so it occurs explicitly. But that can be done later.

PR #7015, opened 2023-08-25
Configurable Polling Variables for Indexer

### LINKED ISSUE #6964: Configurable Polling Intervals for Indexer
**Is your feature request related to a problem? Please describe.**
Currently the op-indexer ETL [polls](https://github.com/ethereum-optimism/optimism/blob/0200fa460f03382bb4d3ec4c80c165ebf6bf485b/indexer/etl/etl.go#L16-L22) both L1 and L2 for new block events every 5 seconds. These values could be irrespective or non-proportional to actual block production rates for some L1/L2 chains. Additionally a default interval of 5 seconds may be too slow for reading L2 blocks provided the current rate of L2 block production on OP Stack chains (2 seconds per block). 

**Describe the solution you'd like**
Add support for configurable millisecond poll interval values (`L1_POLL_INTERVAL`, `L2_POLL_INTERVAL`) that allow a user to configure the respective polling rates per each chain. 

**Describe alternatives you've considered**
N/A

**Additional context**

ethenotethan:
**Description**
Added support for four config variables (i.e, `l1-polling-interval`, `l2-polling-interval`, `l1-header-buffer-size`, `l2-header-buffer-size`) that provide a user fine grained control over the indexer's polling frequency and batch sizes. If these values are unset, the existing repo defaults are embedded instead. 

More specifically:
* Extended the `ChainConfig` struct to hold polling configurations
* Added an etl `Config` type that stores parameters for a L1/L2 ETL instance
* Updated dependency injection logic to build L1/L2 etl configs 
* Updated config tests to ensure polling values are properly loaded

**Tests**
Updated existing config tests to ensure that default values set when expected and don't set when unexpected. 

**Additional context**

**Metadata**

- Fixes #6964

### DIFF
--- a/indexer/config/config.go
+++ b/indexer/config/config.go
@@ -11,6 +11,12 @@ import (
 	geth_log "github.com/ethereum/go-ethereum/log"
 )
 
+const (
+	// default to 5 seconds
+	defaultLoopInterval     = 5000
+	defaultHeaderBufferSize = 500
+)
+
 // in future presets can just be onchain config and fetched on initialization
 
 // Config represents the `indexer.toml` file used to configure the indexer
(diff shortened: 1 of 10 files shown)


indexer/config/config.go
```diff
@@ -11,6 +11,12 @@ import (
 	geth_log "github.com/ethereum/go-ethereum/log"
 )
 
+const (
+	// default to 5 seconds
+	defaultLoopInterval     = 5000
```
hamdiallam, line 16: We should have a better default here, and instead override the indexer config in `e2e_tests/setup.go` to something like 1 or 2 seconds.

Just spitballing: with a ~15s second block time, how about ~3 minutes so we elapse at least 10 L1 blocks?

hamdiallam: And probably a tighter loop interval on L2 since the block time is ~2s. Maybe like 1 minute on L2?

hamdiallam (commented):
LGTM! 1 comment about using a better default and instead overriding these values for tests. The original default loop interval was super tight just so it'd pick up changes super fast on testnet.

indexer/config/config.go
ethenotethan: Hmm I'm curious as to why we'd want a longer interval that allows us to not know about blocks for awhile. Would it not make more sense to do polling significantly (i.e, in-between new blocks) more frequently so that we can capture and process new block events as they happen in almost real-time?

hamdiallam: hmm the primary tradeoff is # RPCs vs efficiency via batching. But i can see this being pretty lightweight enough to run on each block.

ffti and lets maybe re-evaluate how much overhead the indexer running on this interval is adding to the replica it's connected with

hamdiallam (approved):

PR #7017, opened 2023-08-25
op-challenger: Play games in parallel

ajsutton:
**Description**

Implements a `GameScheduler` component that can progress games in parallel.  It looks big, but more than half the changes are tests. 😄 

The number of worker threads is currently hard coded - will follow up with adding a CLI option to control it.

**Tests**

Lots of them.

**Additional context**

This is a much improved version of https://github.com/ethereum-optimism/optimism/pull/6998. Same idea but much more readable code.

**Metadata**

- https://linear.app/optimism/issue/CLI-4342/support-parallel-games

### DIFF
--- a/op-challenger/fault/monitor.go
+++ b/op-challenger/fault/monitor.go
@@ -2,20 +2,17 @@ package fault
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 	"time"
 
+	"github.com/ethereum-optimism/optimism/op-challenger/fault/scheduler"
 	"github.com/ethereum-optimism/optimism/op-service/clock"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 )
 
-type gamePlayer interface {
-	ProgressGame(ctx context.Context) bool
-}
-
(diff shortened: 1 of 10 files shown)


refcell (approved):
LGTM.

Love the package layout of scheduling components in the new `scheduler` dir, especially the splitting out of external types - very easily approachable and intuitive.

I think we may eventually want to revisit how we think about the delineation between the actual challenger and the game solver. If we made this a clean intuitive interface, it could be quite easy to "plug-and-play" different types of games in the challenger.

Right now it seems we are placing too much of this into the fault package - much of this scheduler code can _probably_ be pulled out into a core challenge package. This could make the delineation of challenger vs game solver at a lower level, possibly the Agent.

ajsutton:
@refcell Yeah we absolutely need to do something about the package structure - the naming conflicts are getting out of hand.  I might take a first step on Monday to split the multigame and game-specific code into separate packages. I've wanted to do that for a while as its a clear boundary but it would cause a bunch of conflicts so have held off.  You have a good point about there also being a bunch of stuff that's generic to any game type as welll Let's do some pair refactoring on it when you're back from the conference - then we can go nuts moving and renaming everything and merge it in quickly to avoid conflicts.

PR #7022, opened 2023-08-26
Allow passing Config and Value to TxMgr

demon-xxi:
**Description**

This PR includes two changes:

1. Adds constructor of TxMgr that accepts Config type directly instead of CLIConfig
2. Allows specifying Value field in transaction

**Tests**

Added test around config validation similar to CLIConfig. 
For Value field no tests as change is minimal but running in prod successfully for few month now.

**Additional context**

We wanted to reuse `txmgr` logic to improve reliability of balance transfer transactions. 
One of the requirements we have is using KMS based signer so existing CLIConfig logic with passing PK would not work.
Also using txmgs as a package does not make much sense with CLI config as that one is specific to txmgs cli.
Passing `Config` type directly allows to use custom `Signer` implementation and generally simplify config creation. 
Adding `Value` field is obviously needed for balance transfer transactions.

### DIFF
--- a/op-service/txmgr/cli.go
+++ b/op-service/txmgr/cli.go
@@ -290,3 +290,34 @@ type Config struct {
 	Signer opcrypto.SignerFn
 	From   common.Address
 }
+
+func (m Config) Check() error {
+	if m.Backend == nil {
+		return errors.New("must provide the Backend")
+	}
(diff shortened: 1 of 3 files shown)


ajsutton (approved):
LGTM. Definitely nice to not have to use the CLIConfig. Thanks.

PR #7028, opened 2023-08-27
Add note about arbitrary message sending from OptimismPortal

mdehoog:
**Description**

Add a small note to the spec about arbitrary message sending from OptimismPortal as an expected behavior of the portal contract.


refcell (approved):

maurelian (commented):
Most of this also applies to the L1 and L2 xDMs. Ie. They can also send arbitrary messages and should not hold tokens. 

We should include them here as well.

PR #7030, opened 2023-08-28
op-challenger: Separate multigame and single game code into different…

ajsutton:
**Description**

Rearranges op-challenger code to slightly better packages. There's potential to go further but the key aim with this change is to have separation between the code that handles multiple games (top level of the `game` package) and the code that's focussed on a single game (the `fault` package).  I think this also lines up quite well with the `fault` package being specific to fault dispute games and the `game` package being generic.


op-challenger/fault/service_test.go
```diff
@@ -1,102 +0,0 @@
-package fault
```
refcell, line 1: Looks like this file was deleted. Was that intentional?

refcell (approved):
LGTM

op-challenger/fault/service_test.go
refcell: NVM - noticed that you moved the validate absolute prestate function. LGTM

PR #7031, opened 2023-08-28
op-challenger: Make scripts more flexible

ajsutton:
**Description**

Updates the op-challenger scripts to make them useful for cannon games on goerli rather than just the devnet.

* Splits a out a script to start a new game from the `init_game.sh` that also starts the devnet.
* Adds a `move.sh` script to perform moves against a game

**Metadata**

- https://linear.app/optimism/issue/CLI-4366/[fpa-usability]-create-a-guidescript-to-call-attackdefend-to-break-the

### DIFF
--- a/op-challenger/scripts/alphabet/init_game.sh
+++ b/op-challenger/scripts/alphabet/init_game.sh
@@ -59,43 +59,8 @@ do
   sleep 2
 done
 
-# Fetch the latest block number
-L2_BLOCK_NUMBER=$(cast call $L2_OUTPUT_ORACLE_PROXY "latestBlockNumber()")
-echo "Using the latest L2OO block number: $L2_BLOCK_NUMBER"
-
-# We will use the l2 block number of 1 for the dispute game.
-# We need to check that the block oracle contains the corresponding l1 block number.
-echo "Checkpointing the block oracle..."
-L1_CHECKPOINT=$(cast send --private-key $DEVNET_SPONSOR $BLOCK_ORACLE_PROXY "checkpoint()" --json | jq -r .blockNumber | cast to-dec)
-((L1_CHECKPOINT=L1_CHECKPOINT-1))
-echo "L1 Checkpoint: $L1_CHECKPOINT"
-
-INDEX=$(cast call $L2_OUTPUT_ORACLE_PROXY "getL2OutputIndexAfter(uint256)" $L2_BLOCK_NUMBER | cast to-dec)
-((PRIOR_INDEX=INDEX-1))
-echo "Getting the l2 output at index $PRIOR_INDEX"
-cast call $L2_OUTPUT_ORACLE_PROXY "getL2Output(uint256)" $PRIOR_INDEX
-
-echo "Getting the l2 output at index $INDEX"
-cast call $L2_OUTPUT_ORACLE_PROXY "getL2Output(uint256)" $INDEX
-
-# (Alphabet) Fault game type = 255
-GAME_TYPE=255
-
 # Root claim commits to the entire trace.
 # Alphabet game claim construction: keccak256(abi.encode(trace_index, trace[trace_index]))
 ROOT_CLAIM=$(cast keccak $(cast abi-encode "f(uint256,uint256)" 15 122))
 
-# Fault dispute game extra data is calculated as follows.
-# abi.encode(uint256(l2_block_number), uint256(l1 checkpoint))
-EXTRA_DATA=$(cast abi-encode "f(uint256,uint256)" $L2_BLOCK_NUMBER $L1_CHECKPOINT)
-
-echo "Initializing the game"
-FAULT_GAME_ADDRESS=$(cast call --private-key $MALLORY_KEY $DISPUTE_GAME_PROXY "create(uint8,bytes32,bytes)" $GAME_TYPE $ROOT_CLAIM $EXTRA_DATA)
-
-echo "Creating game at address $FAULT_GAME_ADDRESS"
-cast send --private-key $MALLORY_KEY $DISPUTE_GAME_PROXY "create(uint8,bytes32,bytes)" $GAME_TYPE $ROOT_CLAIM $EXTRA_DATA
-
-FORMATTED_ADDRESS=$(cast parse-bytes32-address $FAULT_GAME_ADDRESS)
-echo "Formatted Address: $FORMATTED_ADDRESS"
-
-echo $FORMATTED_ADDRESS > $CHALLENGER_DIR/.fault-game-address
+GAME_TYPE=255 ${SOURCE_DIR}/../create_game.sh http://localhost:8545 "${DISPUTE_GAME_PROXY}" "${ROOT_CLAIM}" --private-key "${DEVNET_SPONSOR}"
--- a/op-challenger/scripts/alphabet/init_game.sh
+++ b/op-challenger/scripts/alphabet/init_game.sh
@@ -15,7 +15,7 @@ make cannon-prestate
 make devnet-up
 
 DEVNET_SPONSOR="ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
-DISPUTE_GAME_PROXY=$(jq -r .DisputeGameFactoryProxy $MONOREPO_DIR/.devnet/addresses.json)
+DISPUTE_GAME_FACTORY=$(jq -r .DisputeGameFactoryProxy $MONOREPO_DIR/.devnet/addresses.json)
 
 echo "----------------------------------------------------------------"
 echo " Dispute Game Factory at $DISPUTE_GAME_PROXY"
@@ -63,4 +63,4 @@ done
 # Alphabet game claim construction: keccak256(abi.encode(trace_index, trace[trace_index]))
 ROOT_CLAIM=$(cast keccak $(cast abi-encode "f(uint256,uint256)" 15 122))
 
-GAME_TYPE=255 ${SOURCE_DIR}/../create_game.sh http://localhost:8545 "${DISPUTE_GAME_PROXY}" "${ROOT_CLAIM}" --private-key "${DEVNET_SPONSOR}"
+GAME_TYPE=255 ${SOURCE_DIR}/../create_game.sh http://localhost:8545 "${DISPUTE_GAME_FACTORY}" "${ROOT_CLAIM}" --private-key "${DEVNET_SPONSOR}"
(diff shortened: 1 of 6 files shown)


op-challenger/scripts/visualize.sh
```diff
@@ -2,30 +2,19 @@
...
+for i in $(seq 0 "${COUNT}")
+do
+  CLAIM=$(cast call --rpc-url "${RPC}" "${GAME_ADDR}" 'claimData(uint256) returns(uint32 parentIndex, bool countered, bytes32 claim, uint128 position, uint128 clock)' "${i}")
+  SAVEIFS=$IFS   # Save current IFS (Internal Field Separator)
+  IFS=$'\n'      # Change IFS to newline char
+  CLAIM=($CLAIM) # split the string into an array by the same name
+  IFS=$SAVEIFS   # Restore original IFS
+
+  echo "${i}  Parent: ${CLAIM[0]} Countered: ${CLAIM[1]} Claim: ${CLAIM[2]} Position: ${CLAIM[3]} Clock ${CLAIM[4]}"
+done
```
refcell: This seems like the wrong script to make these changes?

The visualize script was useful specifically for generating the game visualization. These changes seem to just print the game information. I don't think it makes sense to delete the visualize script logic and replace it with non-"visualizing" logic.

op-challenger/scripts/move.sh
```diff
@@ -0,0 +1,14 @@
...
+
+RPC=${1:?Must specify RPC URL}
+GAME_ADDR=${2:?Must specify game address}
+ACTION=${3:?Must specify attack or defend}
+CLAIM=${4:?Must specify claim hash}
+SIGNER_ARGS="${@:5}"
+
+# Respond to the last claim that was made
+CLAIM_IDX=$(cast call --rpc-url "${RPC}" "${GAME_ADDR}" 'claimDataLen() returns(uint256)')
+((CLAIM_IDX=CLAIM_IDX-1))
```
refcell: FWIW This will be undefined behavior if the claim index is 0.

op-challenger/scripts/alphabet/init_game.sh
```diff
@@ -59,43 +59,8 @@ do
...
-FAULT_GAME_ADDRESS=$(cast call --private-key $MALLORY_KEY $DISPUTE_GAME_PROXY "create(uint8,bytes32,bytes)" $GAME_TYPE $ROOT_CLAIM $EXTRA_DATA)
-
-echo "Creating game at address $FAULT_GAME_ADDRESS"
-cast send --private-key $MALLORY_KEY $DISPUTE_GAME_PROXY "create(uint8,bytes32,bytes)" $GAME_TYPE $ROOT_CLAIM $EXTRA_DATA
-
-FORMATTED_ADDRESS=$(cast parse-bytes32-address $FAULT_GAME_ADDRESS)
-echo "Formatted Address: $FORMATTED_ADDRESS"
-
-echo $FORMATTED_ADDRESS > $CHALLENGER_DIR/.fault-game-address
+GAME_TYPE=255 ${SOURCE_DIR}/../create_game.sh http://localhost:8545 "${DISPUTE_GAME_PROXY}" "${ROOT_CLAIM}" --private-key "${DEVNET_SPONSOR}"
```
refcell: Can we rename `DISPUTE_GAME_PROXY` to something that indicates it is the factory and not the proxy address of a single dispute game? Maybe just `DISPUTE_GAME_FACTORY_PROXY`.

op-challenger/scripts/create_game.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "Checkpointing the block oracle..."
+L1_CHECKPOINT=$(cast send --rpc-url "${RPC}" ${SIGNER_ARGS} "${BLOCK_ORACLE_ADDR}" "checkpoint()" --json | jq -r .blockNumber | cast to-dec)
+((L1_CHECKPOINT=L1_CHECKPOINT-1))
+echo "L1 Checkpoint: $L1_CHECKPOINT"
+
+# Fault dispute game extra data is calculated as follows.
+# abi.encode(uint256(l2_block_number), uint256(l1 checkpoint))
+EXTRA_DATA=$(cast abi-encode "f(uint256,uint256)" "${L2_BLOCK_NUM}" "${L1_CHECKPOINT}")
+
+
```
refcell: nit
```suggestion

```

op-challenger/scripts/create_game.sh
```diff
@@ -0,0 +1,44 @@
...
+MONOREPO_DIR=$(echo ${CHALLENGER_DIR%/*})
+
+# ./create_game.sh <rpc-addr> <dispute-game-factory-addr> <cast signing args>
+RPC=${1:?Must specify RPC address}
+FACTORY_ADDR=${2:?Must specify factory address}
+ROOT_CLAIM=${3:?Must specify root claim}
+SIGNER_ARGS="${@:4}"
+
+# Default to Cannon Fault game type
+GAME_TYPE=${GAME_TYPE:-0}
```
refcell, line 15: Feels like this should be parameterized as well

op-challenger/scripts/create_game.sh
```diff
@@ -0,0 +1,44 @@
...
+# Fault dispute game extra data is calculated as follows.
+# abi.encode(uint256(l2_block_number), uint256(l1 checkpoint))
+EXTRA_DATA=$(cast abi-encode "f(uint256,uint256)" "${L2_BLOCK_NUM}" "${L1_CHECKPOINT}")
+
+
+echo "Initializing the game"
+FAULT_GAME_DATA=$(cast send --rpc-url "${RPC}" ${SIGNER_ARGS} "${FACTORY_ADDR}" "create(uint8,bytes32,bytes) returns(address)" "${GAME_TYPE}" "${ROOT_CLAIM}" "${EXTRA_DATA}" --json)
+FAULT_GAME_ADDRESS=$(echo "${FAULT_GAME_DATA}" | jq -r '.logs[0].topics[1]' | cast parse-bytes32-address)
+echo "Fault game address: ${FAULT_GAME_ADDRESS}"
+echo "${FAULT_GAME_ADDRESS}" > $CHALLENGER_DIR/.fault-game-address
```
refcell, line 49: Creating multiple games via this script will overwrite this file. I wonder if it'd be worth creating a temp dir and naming the file according to the claim index. That way game players could grab the fdg address via the claim index. Might also be an overoptimization if these scripts are to be deleted.

ajsutton: Writing to this file was really only added for compatibility with the `make alphabet` approach. In normal use users would just copy the address that's printed to the console and not try to use this file so its fine if it gets overwritten each time.

op-challenger/scripts/create_game.sh
ajsutton: If it's added as a param it becomes required for all users which isn't really needed. It just needs to be possible to override it so that this script can be used from `init_game.sh` for `make alphabet`. Seemed worth it to avoid duplication and means that the create_game script gets tested every time we run `make alphabet`.

op-challenger/scripts/move.sh
ajsutton: You can't have a game with no claims - the root claim is always added with contract creation.

op-challenger/scripts/visualize.sh
ajsutton: I replaced the old script because it doesn't work at all with cannon games:
```
  [29849] FaultDisputeGameViz::remote(0x78A3f18f29fc7f1B92924C0Fd1ec4b628BE6aCca)
    ├─ [0] VM::load(0x78A3f18f29fc7f1B92924C0Fd1ec4b628BE6aCca, 0x0000000000000000000000000000000000000000000000000000000000000001) [staticcall]
    │   └─ ← 0x883ab54109c90c676cdac086acec1641312ad5135fc05d999e2197b3d55ee2f5
    └─ ← "Memory allocation overflow"
```

I've brought it back with a couple of tweaks and made this new script `list_claims.sh` but I worry that the visualize script will just cause confusion given it doesn't actually work.

op-challenger/scripts/list_claims.sh
```diff
@@ -0,0 +1,20 @@
...
+COUNT=$(cast call --rpc-url "${RPC}" "${GAME_ADDR}" 'claimDataLen() returns(uint256)')
+echo "Claim count: ${COUNT}"
+((COUNT=COUNT-1))
+for i in $(seq 0 "${COUNT}")
+do
+  CLAIM=$(cast call --rpc-url "${RPC}" "${GAME_ADDR}" 'claimData(uint256) returns(uint32 parentIndex, bool countered, bytes32 claim, uint128 position, uint128 clock)' "${i}")
+  SAVEIFS=$IFS   # Save current IFS (Internal Field Separator)
+  IFS=$'\n'      # Change IFS to newline char
+  CLAIM=($CLAIM) # split the string into an array by the same name
+  IFS=$SAVEIFS   # Restore original IFS
```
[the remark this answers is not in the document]
ajsutton: Can't use an inline change to IFS in this case because we're just setting a var, not running an actual command.

op-challenger/scripts/list_claims.sh
```diff
@@ -0,0 +1,20 @@
...
+GAME_ADDR=${2:?Must specify fault dispute game address}
+
+COUNT=$(cast call --rpc-url "${RPC}" "${GAME_ADDR}" 'claimDataLen() returns(uint256)')
+echo "Claim count: ${COUNT}"
+((COUNT=COUNT-1))
+for i in $(seq 0 "${COUNT}")
+do
+  CLAIM=$(cast call --rpc-url "${RPC}" "${GAME_ADDR}" 'claimData(uint256) returns(uint32 parentIndex, bool countered, bytes32 claim, uint128 position, uint128 clock)' "${i}")
+  SAVEIFS=$IFS   # Save current IFS (Internal Field Separator)
+  IFS=$'\n'      # Change IFS to newline char
```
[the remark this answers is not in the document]
ajsutton: Can't use an inline change to IFS in this case because we're just setting a var, not running an actual command.

ajsutton:
I've added some basic documentation on the scripts to the challenger README so there's something close by to explain them. We'll write up more complete docs on the process of creating games, performing moves etc as part of FPA launch and can add a link to them once they exist for more detailed mechanics.

refcell:
> I've added some basic documentation on the scripts to the challenger README so there's something close by to explain them. We'll write up more complete docs on the process of creating games, performing moves etc as part of FPA launch and can add a link to them once they exist for more detailed mechanics.

Thank you for doing this 👑!

refcell (approved):

PR #7033, opened 2023-08-28
ci: Convert go-e2e-tests CI to make targets

jyellick:
**Description**

This PR removes the specific test commands from the go-e2e-test CI definition and pushes them into the Makefile.  It also tries to clean up the Makefile somewhat by adding `.PHONY` where appropriate.

This change makes it much easier to make build changes to the op-e2e without having to reflect the changes from the Makefile into the CI definition.  It also makes easier for developers to mirror the execution of what's going on in CI locally.

This change was discussed with @ajsutton in PR #6280 but was outside the scope and is instead coming in via this PR.


ajsutton (approved):
LGTM. Not entirely sure what happened with the indexer tests but they shouldn't be affected by this change.

ajsutton:
@jyellick Can you try merging in the latest develop changes to trigger a rebuild please? I suspect the indexer-tests just flaked but I can't retry the build in circleci as it's not building in the optimism-ethereum org.

jyellick:
I tried re-running via ssh and it passed, but then the ssh session seemed to hang (seemed like it was waiting for postgres to exit).  Regardless, just rebased it onto the tip of develop and re-pushed.  If you know of any way to force CircleCI to execute on your instances so that the publishing works and you can re-run tests, I'd love to hear it!

ajsutton:
> I tried re-running via ssh and it passed, but then the ssh session seemed to hang (seemed like it was waiting for postgres to exit). Regardless, just rebased it onto the tip of develop and re-pushed. If you know of any way to force CircleCI to execute on your instances so that the publishing works and you can re-run tests, I'd love to hear it!

Not entirely sure if this will work, but in the past where jobs have executed in unexpected orgs, the answer was for you to unfollow your fork of the repo in circleci. It's a weird and obscure thing that I've never really understood...

I think your manual re-run was successful so hopefully with it rebased it will pass.  There's some things that fail because of permissions but that shouldn't stop it getting merged.  If it doesn't work out this time, I'll go track someone down who can help. :)

PR #7043, opened 2023-08-29
op-challenger: Use gzip encoded proofs and snapshots

ajsutton:
**Description**

Reduces the disk space requirements for op-challenger by gzip compressing snapshots and proofs.

Introduces a `ioutil` package to op-service with methods to open reader/writers that optionally support gzip compression based on the file extension.

Increases the max duration for devnet fault proof games. The previous 5 minute limit with a game depth of 30 only gave about 10 seconds per move.

**Tests**

Add unit tests, e2e tests confirm overall operation still works.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4382/use-gzipped-proofssnapshots-in-op-challenger


packages/contracts-bedrock/deploy-config/devnetL1.json
```diff
@@ -45,6 +45,6 @@
   "l2GenesisRegolithTimeOffset": "0x0",
   "faultGameAbsolutePrestate": "0x41c7ae758795765c6664a5d39bf63841c71ff191e9189522bad8ebff5d4eca98",
   "faultGameMaxDepth": 30,
-  "faultGameMaxDuration": 300,
```
refcell, line 48: Why this bump?

refcell (approved):

packages/contracts-bedrock/deploy-config/devnetL1.json
ajsutton: The previous 5 minute limit with a game depth of 30 only gave about 10 seconds per move, so e2e tests could fail because the clock expired.  Adding gzip saves disk space but uses more CPU so makes things a bit slower when under the load of e2e tests.

PR #7047, opened 2023-08-29
cannon: Prevent deadlock on pre-image reads

Inphi:
Reading a FileChannel that does not have a writer blocks indefinitely. This occurs in cannon whenever the pre-image oracle server shuts down prematurely. To deal, we introduce a new io.ReadWriter implementation that can detect when a FileChannel would otherwise block indefinitely.

### DIFF
--- a/cannon/cmd/run.go
+++ b/cannon/cmd/run.go
@@ -188,6 +188,8 @@ func (p *ProcessPreimageOracle) Close() error {
 	if p.cmd == nil {
 		return nil
 	}
+	// Give the pre-image server time to exit cleanly before killing it.
+	time.Sleep(time.Second * 1)
 	_ = p.cmd.Process.Signal(os.Interrupt)
 	return <-p.waitErr
 }
--- a/cannon/cmd/run.go
+++ b/cannon/cmd/run.go
@@ -188,15 +188,6 @@ func (p *ProcessPreimageOracle) Close() error {
 	if p.cmd == nil {
 		return nil
 	}
-	// We first check if the process has already exited before signaling.
-	// Note: This is a teeny bit racy since the process could exit after the check
-	// above and another process is assigned the same pid.
-	select {
-	case err := <-p.waitErr:
-		return err
-	case <-time.After(time.Second * 1):
-		// give the wait goroutine time to reap process
-	}
 	_ = p.cmd.Process.Signal(os.Interrupt)
 	return <-p.waitErr
 }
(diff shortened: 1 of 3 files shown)


cannon/cmd/run.go
```diff
@@ -166,23 +179,37 @@ func (p *ProcessPreimageOracle) Start() error {
...
+	select {
+	case err := <-p.waitErr:
+		return err
+	case <-time.After(time.Second * 1):
+		// give the wait goroutine time to reap process
+	}
 	_ = p.cmd.Process.Signal(os.Interrupt)
-	// Go 1.20 feature, to introduce later
-	//p.cmd.WaitDelay = time.Second * 10
+	return <-p.waitErr
```
ajsutton: Why do we need to wait before interrupting here? Wouldn't an interrupt to an exited process just be a no-op (especially if we ignore the error)?

And I would have expected that go would hold a reference to the actual child process it spawned rather than having to repeatedly look it up by pid which should avoid the race condition. Is it not doing that for some reason?

ajsutton (commented):
So when doing IO it sets a timeout for each read/write operation and if that timeout is reached it then checks if the context is done and bails out if not - otherwise it does the read again. Have I understood that right?

I'm unclear on the changes in run.go though - commented inline.

cannon/cmd/run.go
Inphi: you're right, interrupting an exited process is a no-op. I misread the stdlib to be doing otherwise.

Inphi:
> So when doing IO it sets a timeout for each read/write operation and if that timeout is reached it then checks if the context is done and bails out if not - otherwise it does the read again. Have I understood that right?

Correct. ReadWriter operations will effectively block indefinitely as before, _until_ either the context is done or all requested data has been read/written.

Inphi:
The main change to run.go is to asynchronously wait out the pre-image server. This has to be done off the main goroutine to detect when the pre-image server crashes so we can unblock ongoing client I/O operations.

ajsutton:
The test failure btw is because `ClockTimeExceeded()`.  If you haven't got the latest develop changes it would be worth pulling those in - I recently increased the game duration for e2e because it was too tight.

ajsutton:
That's error code `0x3381d114` btw.  Clabby was kind enough to upload the symbols so `cast 4byte 0x3381d114` now works. 🎉

Inphi:
Rebasing did the trick. thanks

ajsutton (approved):

PR #7049, opened 2023-08-29
Add cannon-prestate as prereq to devnet-up

jyellick:
The devnet-up ultimately requires the cannont-prestate target to have already executed, or the script errors out.  Consequently, for someone checking out the repo if they simply do a:

```
make && make devnet-up
```

Things still fail.  This change simply mirrors a similar check added to the e2e Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,9 @@ nuke: clean devnet-clean
 .PHONY: nuke
 
 devnet-up:
+	@if [ ! -e op-program/bin ]; then \
+		make cannon-prestate; \
+	fi
 	$(shell ./ops/scripts/newer-file.sh .devnet/allocs-l1.json ./packages/contracts-bedrock)
 	if [ $(.SHELLSTATUS) -ne 0 ]; then \
 		make devnet-allocs; \


ajsutton (approved):
Thank you. :)

ajsutton:
I don't really know how `contracts-bedrock-slither` didn't get marked as passing - circleci shows it passed fine.  @jyellick Can you pull in the latest changes from develop and see if the rebuild gets it sorted please?

Sorry for all the trouble CI seems to cause your PRs...

jyellick:
Yes, sorry for your cycles spent with the flaking CI as well.

Per your other suggestion to 'unfollow' the project in our Circle instance, unfortunately I don't think that will work, as we want to be able to run CI against changes that either aren't ready or aren't appropriate to upstream.

I am working to get the publish targets going by linking in a GCP instance, but I think there are other caching components of the CI that maybe should be enabled as well?  Is there any infra-as-code type repo (or even simple docs) where it's established how to appropriately bootstrap the 3rd party components of the CI config?

ajsutton:
> Per your other suggestion to 'unfollow' the project in our Circle instance, unfortunately I don't think that will work, as we want to be able to run CI against changes that either aren't ready or aren't appropriate to upstream.

Ah yeah, I was worried that would be the case. It's not a big deal - just means you need to manage rerunning if necessary rather than me.
 
> I am working to get the publish targets going by linking in a GCP instance, but I think there are other caching components of the CI that maybe should be enabled as well? Is there any infra-as-code type repo (or even simple docs) where it's established how to appropriately bootstrap the 3rd party components of the CI config?

No I don't think we've written anything up there, probably worth raising an issue about it so the request is tracked. Most of the CI setup predates me so I'm not too familiar with it either to be honest.

PR #7051, opened 2023-08-29
Fix dangling modified file on make devnet-up

jyellick:
When starting from a fresh clone of the repo and running

```
make
make cannon-prestate
make devnet-up
```

We are left with a file changed in the tree at

```
packages/contracts-bedrock/deploy-config/devnetL1.json
```

This file is a bit annoying as new changes to it should generally not be checked in but shows as an outstanding change.  The python script reads the original file, backs it up, then rewrites that file with a new timestamp.

I'm not entirely certain why the timestamp needs to be rewritten, so I've left that modification in place.  Instead, this change takes the approach of moving the original file to a 'template' path, and ignoring the modified file.  In this way there's no longer any need to constantly create a backup and have the changes show up in a git diff.

If we don't like this approach, I'm happy to implement something else, just wanting to stop accidentally capturing this file in my changesets.


Inphi (approved):
Thanks!

Inphi:
@jyellick could you rerun the failing circle ci check - https://app.circleci.com/pipelines/github/bobanetwork/v3-anchorage/446/workflows/f89a47a6-4e1a-4870-af05-383b7d775d47/jobs/18765. The PR passes for me locally so this could just be a flake.

jyellick:
This change seems to make the devnet CI a bit flaky, going to investigate, will undraft once ready.

jyellick:
Modified the change a bit to ensure that a version with the original timestamp is written during the initial alloc, while an updated timestamp is written during the devnet-up.  I'm not entirely sure why this behavior is necessary, but, it seems to make CI happy.

jyellick:
@mergifyio requeue

PR #7055, opened 2023-08-29
op-chain-ops: Remove old TODO

ajsutton:
**Description**

Remove an unattributed TODO that's been hanging around for quite a while and doesn't seem to be important to act on.  Semgrep has been failing because of it.


op-chain-ops/safe/batch_helpers.go
```diff
@@ -141,7 +141,6 @@ func createContractInput(input abi.Argument, inputs []ContractInput) ([]Contract
 		return nil, err
 	}
 
-	// TODO: could probably do better than string comparison?
```
Inphi, line 144: I guess a better method is to check `input.Type.T == abi.TupleTy`, like we do [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-chain-ops/safe/batch_helpers.go#L178). Either is fine imo.

Inphi (approved):

op-chain-ops/safe/batch_helpers.go
ajsutton: Going to merge this to get the build passing and then will take a stab at using that better method as you suggest.

PR #7059, opened 2023-08-30
op-program: Retrying fetching if preimage disappears between fetching and retrying read

ajsutton:
**Description**

Keep retrying the prefetch if a preimage still isn't available.  This better handles the case where the pre-image is downloaded by the prefetcher but then deleted by some other process before it is read by the retried `Get` call.

This is a first step towards being able to handle files being deleted unexpectedly and generally better handling IO errors.

**Tests**

Added unit tests.

**Metadata**

- https://linear.app/optimism/issue/CLI-4379/op-program-handle-pre-images-being-deleted-unexpectedly


op-program/host/prefetcher/prefetcher.go
```diff
@@ -58,14 +58,18 @@ func (p *Prefetcher) Hint(hint string) error {
...
 		hint := p.lastHint
-		p.lastHint = ""
 		if err := p.prefetch(ctx, hint); err != nil {
 			return nil, fmt.Errorf("prefetch failed: %w", err)
 		}
-		// Should now be available
-		return p.kvStore.Get(key)
+		pre, err = p.kvStore.Get(key)
+		if err != nil {
+			p.logger.Error("Fetched pre-images for last hint but did not find required key", "hint", hint, "key", key)
```
Inphi, line 71: This could generate a lot of logging as it's in a loop. How often are you observing deleted pre-images?

ajsutton: Only when I write really really buggy code that keeps deleting files from under it.

I wouldn't currently expect to see it in production at all, but we might see it sometimes if we move to using a shared Preimage cache and implement some form of pruning - it shouldn't be frequent though.

There is a chance that we introduce a bug somewhere which means the prefetching doesn't ever populate the required key but at that point we're dead in the water and I'm ok with spamming the crap out of the logs so we know there's a problem.

refcell (approved):

PR #7061, opened 2023-08-30
feat(ctb): Rework FaultDisputeGame resolution

Inphi:
Presenting an alternative resolution mechanism based on _subgames_.

## Subgame Resolution

A subgame is a subgraph in the FDG DAG with a max depth of 1. Resolving the subgame determines the validity of the root claim of its subgraph. If the subgame of the absolute root claim in the FDG is valid, then Defenders win. Otherwise, Challengers win.
Uncontested claims are initially considered valid. Once contested, it gets a little bit fuzzy. Their validity depends on that of its immediate children. Depending on a claim's depth, there are two rules to determine its validity:
- At max game depth, a claim's validity is determined by the step function. If step succeeds, then the claim at max game depth becomes _invalid_.
- At other depths, a contested claim becomes valid if all of its children are invalid. If the claim is uncontested, then it remains valid.

## Resolution Order

Resolving the Fault Dispute Game now involves every claim to be individually resolved, via `resolveClaim`. Once every claim has been resolved, the entire game can finally be resolved as usual.
In order to resolve a given claim, all of its children must also be resolved. Claims are max game depth are considered already resolved. But every other claim must be resolved before its dispute can be considered eligible against the parent claim.

Once the root claim, and thus extension every other claim in the FDG, has been resolved then the FDG can be resolved as usual via the `resolve` function.

## Honest Challenger

This changes the strategy of game players. The op-challenger should dispute every claim they disagree with, provided they're _not_ on a disagreeable path. A path to the root claim is considered disagreeable, if every other claim along the path are not all honest. Consider following graph:
`A -> B -> C -> D -> E`
The honest challenger will dispute `E` iff `D` and `B` (and the output proposal) are all honest.

This is still effectively restricts responses to claims on "levels we disagree with", while avoiding lending support to dishonest claims on such "levels we agree with"


packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -75,6 +75,12 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
     /// @notice An internal mapping to allow for constant-time lookups of existing claims.
     mapping(ClaimHash => bool) internal claims;
 
+    /// @notice An internal mapping of subgames rooted at a claim index to other claim indices in the subgame.
+    mapping(uint256 => uint256[]) internal subgames;
```
Inphi, line 79: Subgames have a max depth of 1. So this maps claim indices to an array of other claims that _directly_ dispute it.
The total number of subgames and their edges are both $O(N)$.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -232,9 +238,10 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, Semver {
         // Construct the next clock with the new duration and the current block timestamp.
         Clock nextClock = LibClock.wrap(nextDuration, Timestamp.wrap(uint64(block.timestamp)));
 
-        // INVARIANT: A claim may only exist at a given position once. Multiple claims may exist
-        //            at the same position, however they must have different values.
-        ClaimHash claimHash = _claim.hashClaimPos(nextPosition);
+        // INVARIANT: There cannot be multiple identical claims with identical moves on the same challengeIndex. Multiple
+        // claims
+        //            at the same position may dispute the same challengeIndex. However, the must have different values.
```
clabby, line 243:
```suggestion
        // claims at the same position may dispute the same challengeIndex. However, the must have different values.
```

clabby (approved):
Implementation LGTM. There's some stuff we can clean up here to make it more efficient, but let's worry about that in another PR.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
Inphi: Linting fails with this suggestion. Weirdly, the forge formatter prefers the current change.

refcell (approved):

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
refcell: Should be fixed later. Small formatting issues should not block

refcell (approved):

refcell:
@maurelian addressed the ci flake in #7308

Inphi:
@mergifyio requeue

PR #7064, opened 2023-08-30
feat(indexer): index service /healthz endpoint

hamdiallam:
The indexer is a seperate service from the api. It needs to spin up it's own HTTP server
in order to serve a /healthz endpoint

Renamed `APIConfig` to `ServerConfig` such that both `MetricsServer` and `HTTPServer`
share the same structural configuration (different toml keys, `metrics` and `http`)

### DIFF
--- a/indexer/cmd/indexer/cli.go
+++ b/indexer/cmd/indexer/cli.go
@@ -38,7 +38,7 @@ func runIndexer(ctx *cli.Context) error {
 	}
 	defer db.Close()
 
-	indexer, err := indexer.NewIndexer(log, db, cfg.Chain, cfg.RPCs, cfg.Metrics)
+	indexer, err := indexer.NewIndexer(log, db, cfg.Chain, cfg.RPCs, cfg.HTTPServer, cfg.MetricsServer)
 	if err != nil {
 		log.Error("failed to create indexer", "err", err)
 		return err
@@ -63,7 +63,7 @@ func runApi(ctx *cli.Context) error {
 	defer db.Close()
 
 	api := api.NewApi(log, db.BridgeTransfers)
-	return api.Listen(ctx.Context, cfg.API.Port)
+	return api.Listen(ctx.Context, cfg.HTTPServer.Port)
 }
 
 func runAll(ctx *cli.Context) error {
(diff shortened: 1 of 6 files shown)


hamdiallam:
* develop
  * **PR #7064** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7064" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7064?utm_source=stack-comment).

indexer/cmd/indexer/cli.go
```diff
@@ -38,7 +38,7 @@ func runIndexer(ctx *cli.Context) error {
 	}
 	defer db.Close()
 
-	indexer, err := indexer.NewIndexer(log, db, cfg.Chain, cfg.RPCs, cfg.Metrics)
+	indexer, err := indexer.NewIndexer(log, db, cfg)
```
roninjin10: Nothing should ever have access to the entire config object as was done intentionally in [https://github.com/ethereum-optimism/optimism/pull/6798](https://github.com/ethereum-optimism/optimism/pull/6798)

roninjin10 (changes requested):
Looks good only blocking on the indexer taking entire cfg as a dep

indexer/cmd/indexer/cli.go
hamdiallam: this makes sense but with this addition, the indexer makes use of all parts of the config so I thought it'd be simpler to take in the entire thing config.

If config grows again & creates new things parts that the index service doesn't use, makes sense to take things in individually? wdyt?

Or I can create an `IndexerConfig` struct in this same file for brevity

roninjin10: Gonna need to either do one of
1. `api.config.toml` `indexer.config.toml` thing
2. Same idea as 1 but keep `indexer.config.toml` as one config for convenience
3. Use seperate CLI apps for both
I don't think overloading a config option like configuring a server should be overloaded but instead be explicit. Going straight to one of these options works.

So all 3 of these options involve the config growing

hamdiallam: forsure, i'll just pass in each config struct explicitly

roninjin10 (approved):

PR #7068, opened 2023-08-30
contracts,op-node: SuperchainConfig L1 contract and protocol version signal

protolambda:
**Description**

This is *draft*, that introduces opt-in beta functionality to check protocol version compatibility.

- [x] Define `ProtocolVersions` contract, with all the patterns we established in `SystemConfig`, to host configuration that is relevant to a superchain as a whole.
  - TBD: some chains in a superchain may need individual upgrade signal overrides, if the follow a more experimental/fast protoocl upgrade schedule. Perhaps we should host the same `recommended`/`required` vars in the `SystemConfig`?
- [x] Upade op-node runtime-config loader to load the `recommended`/`required` protocol versions.
- [x] Update op-node to periodically reload the runtime config (Note: overlaps with backlog p2p-signer runtime config loading fix)
- [x] Add logging, metrics, and opt-in mitigation actions, when op-node detects it's outdated.
- [x] Call new engine-API method with loaded values, to signal the protocol version to op-geth.

I'll split it up in a PR stack for review purposes once the inter-dependencies are separated better.


protolambda:
implemented in #7227 and #7226

PR #7077, opened 2023-08-31
feat(op-challenger): Move Counter Metrics

refcell:
**Description**

Propagates the metricer down to the fault component agent so it can record a game move when calling the responder component.

Tests for the metricer call inside the agent will be added in a followup pr since we are missing Agent unit tests on the `Act` method.

**Metadata**

Fixes CLI-4391
Fixes CLI-4392

### DIFF
--- a/op-challenger/game/fault/agent_test.go
+++ b/op-challenger/game/fault/agent_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/ethereum-optimism/optimism/op-challenger/game/fault/types"
+	"github.com/ethereum-optimism/optimism/op-challenger/metrics"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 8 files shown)


refcell:
* develop
  * **PR #7077** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7077" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7077?utm_source=stack-comment).

op-challenger/game/fault/agent_test.go
```diff
@@ -16,14 +16,14 @@ func TestShouldResolve(t *testing.T) {
 	log := testlog.Logger(t, log.LvlCrit)
 
 	t.Run("AgreeWithProposedOutput", func(t *testing.T) {
-		agent := NewAgent(nil, 0, nil, nil, nil, true, log)
+		agent := NewAgent(nil, nil, 0, nil, nil, nil, true, log)
```
ajsutton: We should pass in `NoopMetrics` here instead of nil.

op-challenger/game/fault/agent_test.go
```diff
@@ -16,14 +16,14 @@ func TestShouldResolve(t *testing.T) {
...
 		require.False(t, agent.shouldResolve(context.Background(), types.GameStatusDefenderWon))
 		require.True(t, agent.shouldResolve(context.Background(), types.GameStatusChallengerWon))
 		require.False(t, agent.shouldResolve(context.Background(), types.GameStatusInProgress))
 	})
 
 	t.Run("DisagreeWithProposedOutput", func(t *testing.T) {
-		agent := NewAgent(nil, 0, nil, nil, nil, false, log)
+		agent := NewAgent(nil, nil, 0, nil, nil, nil, false, log)
```
ajsutton: We should pass in `NoopMetrics` here instead of nil.

op-challenger/game/monitor_test.go
```diff
@@ -99,11 +100,21 @@ func setupMonitorTest(t *testing.T, allowedGames []common.Address) (*gameMonitor
...
 		return i, nil
 	}
+	m := &mockMetricer{}
 	sched := &stubScheduler{}
-	monitor := newGameMonitor(logger, clock.SystemClock, source, sched, time.Duration(0), fetchBlockNum, allowedGames)
+	monitor := newGameMonitor(logger, m, clock.SystemClock, source, sched, time.Duration(0), fetchBlockNum, allowedGames)
 	return monitor, source, sched
 }
 
+type mockMetricer struct {
```
ajsutton: This is going to get annoying to have to maintain as it will require updating every time we add a new metric.  Probably worth generating a mock with mockery to make it more maintainable, or at least put the mock somewhere reusable so we don't wind up maintaining multiple versions of it.

Actually I see it's not actually doing anything so currently could just be replaced with `NoopMetrics`.

ajsutton (commented):
Looks good generally just a couple of tweaks to the test side to use the NoopMetrics.

ajsutton (approved):
LGTM.

PR #7078, opened 2023-08-31
feat(op-challenger): Record Cannon Execution Time

refcell:
**Description**

Adds a prometheus histogram to record cannon execution time by propagating
the Metricer down through the challenger to the cannon trace provider.

The executor accepts a minimal metrics interface containing the single
cannon execution time record method.

**Metadata**

Fixes CLI-4393

### DIFF
--- a/op-challenger/game/fault/trace/cannon/executor.go
+++ b/op-challenger/game/fault/trace/cannon/executor.go
@@ -11,6 +11,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/ethereum-optimism/optimism/op-challenger/config"
 	oplog "github.com/ethereum-optimism/optimism/op-service/log"
@@ -30,6 +31,7 @@ type cmdExecutor func(ctx context.Context, l log.Logger, binary string, args ...
 
 type Executor struct {
 	logger           log.Logger
+	metrics          CannonMetricer
 	l1               string
 	l2               string
 	inputs           LocalGameInputs
@@ -45,9 +47,10 @@ type Executor struct {
 	cmdExecutor      cmdExecutor
 }
 
-func NewExecutor(logger log.Logger, cfg *config.Config, inputs LocalGameInputs) *Executor {
+func NewExecutor(logger log.Logger, m CannonMetricer, cfg *config.Config, inputs LocalGameInputs) *Executor {
 	return &Executor{
 		logger:           logger,
+		metrics:          m,
 		l1:               cfg.L1EthRpc,
 		l2:               cfg.CannonL2,
 		inputs:           inputs,
@@ -119,7 +122,13 @@ func (e *Executor) GenerateProof(ctx context.Context, dir string, i uint64) erro
 		return fmt.Errorf("could not create proofs directory %v: %w", proofDir, err)
 	}
 	e.logger.Info("Generating trace", "proof", i, "cmd", e.cannon, "args", strings.Join(args, ", "))
-	return e.cmdExecutor(ctx, e.logger.New("proof", i), e.cannon, args...)
+	execStart := time.Now()
+	err = e.cmdExecutor(ctx, e.logger.New("proof", i), e.cannon, args...)
+	if err != nil {
+		execDuration := time.Since(execStart).Seconds()
+		e.metrics.RecordCannonExecutionTime(execDuration)
+	}
+	return err
 }
 
 func runCmd(ctx context.Context, l log.Logger, binary string, args ...string) error {
--- a/op-challenger/game/fault/trace/cannon/executor_test.go
+++ b/op-challenger/game/fault/trace/cannon/executor_test.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/ethereum-optimism/optimism/op-challenger/config"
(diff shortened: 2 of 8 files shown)


refcell:
* develop
  * **PR #7078** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7078" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7078?utm_source=stack-comment).

op-challenger/game/fault/trace/cannon/executor.go
```diff
@@ -131,7 +134,13 @@ func runCmd(ctx context.Context, l log.Logger, binary string, args ...string) er
...
 	cmd.Stdout = stdOut
 	cmd.Stderr = stdErr
-	return cmd.Run()
+	start := time.Now()
+	err := cmd.Run()
+	if err != nil {
+		execTime := time.Since(start).Seconds()
+		m.RecordCannonExecutionTime(execTime)
+	}
+	return err
```
ajsutton: I'd suggest tracking this metric outside this `runCmd` so we can keep this really minimally focussed on the difficult-to-test actual execution of a process.  Effectively the time it takes for `runCmd` to execute is the cannon execution time as far as op-challenger is concerned.

op-challenger/game/fault/trace/cannon/executor_test.go
```diff
@@ -210,3 +212,7 @@ func TestFindStartingSnapshot(t *testing.T) {
 		require.Equal(t, filepath.Join(dir, "100.json.gz"), snapshot)
 	})
 }
+
+type mockMetricer struct{}
```
ajsutton: Should be able to just use `NoopMetrics` instead.

op-challenger/game/monitor_test.go
```diff
@@ -99,11 +100,20 @@ func setupMonitorTest(t *testing.T, allowedGames []common.Address) (*gameMonitor
...
 	return monitor, source, sched
 }
 
+type mockMetricer struct {
+	txmetrics.NoopTxMetrics
+}
+
+func (m *mockMetricer) RecordInfo(version string)           {}
+func (m *mockMetricer) RecordUp()                           {}
+func (m *mockMetricer) RecordCannonExecutionTime(t float64) {}
```
ajsutton: NoopMetrics. :)

op-challenger/metrics/metrics.go
```diff
@@ -58,6 +62,12 @@ func NewMetrics() *Metrics {
 			Name:      "up",
 			Help:      "1 if the op-challenger has finished starting up",
 		}),
+		cannonExecutionTime: factory.NewHistogram(prometheus.HistogramOpts{
+			Namespace: Namespace,
+			Name:      "cannon_execution_time",
+			Help:      "Time (in seconds) to execute cannon",
+			Buckets:   append([]float64{1.0, 10.0}, prometheus.LinearBuckets(30.0, 30.0, 14)...),
```
ajsutton: If I understand this correctly it's only tracking up to 7 minutes but the initial run of cannon is expected to take between 1 and 2 hours. I suspect we want to use exponential buckets.

op-e2e/e2eutils/disputegame/metricer.go
```diff
@@ -0,0 +1,13 @@
+package disputegame
+
+import (
+	"github.com/ethereum-optimism/optimism/op-service/txmgr/metrics"
+)
+
+type mockMetricer struct {
```
ajsutton: This is just NoopMetrics. :)

op-challenger/metrics/metrics.go
refcell: Updated to use exponential buckets which should cover a few hours if i did my math right

op-challenger/metrics/metrics.go
```diff
@@ -58,6 +62,12 @@ func NewMetrics() *Metrics {
 			Name:      "up",
 			Help:      "1 if the op-challenger has finished starting up",
 		}),
+		cannonExecutionTime: factory.NewHistogram(prometheus.HistogramOpts{
+			Namespace: Namespace,
+			Name:      "cannon_execution_time",
+			Help:      "Time (in seconds) to execute cannon",
+			Buckets:   append([]float64{1.0, 10.0}, prometheus.ExponentialBuckets(30.0, 2.0, 14)...),
```
ajsutton, line 83: What about just using:
```suggestion
			Buckets:   append([]float64{1.0, 10.0}, prometheus.ExponentialBucketsRange(30.0, 60.0 * 3.0, 14)...),
```

Then we're defining the range we want and the number of buckets and letting prometheus do the math. :)

ajsutton (approved):
LGTM. Just the one suggestion for how to define the buckets.

op-challenger/metrics/metrics.go
refcell: That looks good to me - can always lower or raise the max when we start testing metrics :)

ajsutton: Yeah 3 hours is pretty arbitrary but I think we'd want to find a faster CPU if cannon is taking longer than that...

PR #7080, opened 2023-08-31
feat(op-challenger): Game GaugeVec

refcell:
**Description**

Adds a game status tracking gauge vec.

Required a hoist of the internal fault `GameStatus` type to the top-level game package.

**Metadata**

Fixes CLI-4394.

### DIFF
--- a/op-challenger/game/scheduler/coordinator.go
+++ b/op-challenger/game/scheduler/coordinator.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/ethereum-optimism/optimism/op-challenger/game/types"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 	"golang.org/x/exp/slices"
@@ -17,7 +19,7 @@ type PlayerCreator func(address common.Address, dir string) (GamePlayer, error)
 type gameState struct {
 	player   GamePlayer
 	inflight bool
-	resolved bool
+	status   types.GameStatus
 }
 
 // coordinator manages the set of current games, queues games to be played (on separate worker threads) and
@@ -31,6 +33,7 @@ type coordinator struct {
 	resultQueue <-chan job
 
 	logger       log.Logger
+	m            SchedulerMetricer
 	createPlayer PlayerCreator
 	states       map[common.Address]*gameState
 	disk         DiskManager
@@ -49,18 +52,35 @@ func (c *coordinator) schedule(ctx context.Context, games []common.Address) erro
 		}
 	}
 
+	var gamesInProgress int
+	var gamesChallengerWon int
+	var gamesDefenderWon int
 	var errs []error
+	var jobs []job
 	// Next collect all the jobs to schedule and ensure all games are recorded in the states map.
 	// Otherwise, results may start being processed before all games are recorded, resulting in existing
 	// data directories potentially being deleted for games that are required.
-	var jobs []job
 	for _, addr := range games {
 		if j, err := c.createJob(addr); err != nil {
 			errs = append(errs, err)
 		} else if j != nil {
 			jobs = append(jobs, *j)
 		}
+		state, ok := c.states[addr]
+		if ok {
+			switch state.status {
+			case types.GameStatusInProgress:
+				gamesInProgress++
+			case types.GameStatusDefenderWon:
+				gamesDefenderWon++
+			case types.GameStatusChallengerWon:
+				gamesChallengerWon++
+			}
+		} else {
+			c.logger.Warn("Game not found in states map", "game", addr)
+		}
 	}
+	c.m.RecordGamesStatus(gamesInProgress, gamesChallengerWon, gamesDefenderWon)
 
 	// Finally, enqueue the jobs
 	for _, j := range jobs {
@@ -114,7 +134,7 @@ func (c *coordinator) processResult(j job) error {
 		return fmt.Errorf("game %v received unexpected result: %w", j.addr, errUnknownGame)
 	}
 	state.inflight = false
-	state.resolved = j.resolved
+	state.status = j.status
 	c.deleteResolvedGameFiles()
 	return nil
 }
@@ -122,7 +142,7 @@ func (c *coordinator) processResult(j job) error {
 func (c *coordinator) deleteResolvedGameFiles() {
 	var keepGames []common.Address
 	for addr, state := range c.states {
-		if !state.resolved || state.inflight {
+		if state.status == types.GameStatusInProgress || state.inflight {
 			keepGames = append(keepGames, addr)
 		}
 	}
@@ -131,9 +151,10 @@ func (c *coordinator) deleteResolvedGameFiles() {
 	}
 }
 
-func newCoordinator(logger log.Logger, jobQueue chan<- job, resultQueue <-chan job, createPlayer PlayerCreator, disk DiskManager) *coordinator {
+func newCoordinator(logger log.Logger, m SchedulerMetricer, jobQueue chan<- job, resultQueue <-chan job, createPlayer PlayerCreator, disk DiskManager) *coordinator {
 	return &coordinator{
 		logger:       logger,
+		m:            m,
 		jobQueue:     jobQueue,
 		resultQueue:  resultQueue,
 		createPlayer: createPlayer,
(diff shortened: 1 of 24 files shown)


refcell:
* develop
  * **PR #7080** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7080" target="_blank">Graphite</a>  👈
    * **PR #7086** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7086" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7080?utm_source=stack-comment).

op-challenger/game/factory.go
```diff
@@ -56,6 +74,7 @@ func (l *gameLoader) FetchAllGamesAtBlock(ctx context.Context, earliestTimestamp
 		return nil, fmt.Errorf("failed to fetch game count: %w", err)
 	}
 
+	l.m.ClearTrackedGames()
```
ajsutton: We need to do the update more atomically than this, otherwise the metrics may get sampled in between this clearing and the games being recounted which will mess with the stats.

I suspect this loader isn't the right place to try and add this metric given it doesn't know the game status either.  The scheduler coordinator might be a better place. It tracks the `gameState` already and while that currently is just tracking status as resolved or not, it would actually be quite reasonable to move that to tracking the actual status. It would mean the `GamePlayer.ProgressGame` returns the status instead of just a `bool` but I think that's probably clearer anyway.

op-challenger/metrics/metrics.go
```diff
@@ -20,6 +20,11 @@ type Metricer interface {
 
 	// Record Tx metrics
 	txmetrics.TxMetricer
+
+	ClearTrackedGames()
+	RecordGameInProgress()
+	RecordGameDefenderWon()
+	RecordGameChallengerWon()
```
ajsutton: To make this atomic I think it will need to be something more like `RecordTrackedGames(inProgressCount int, defenderWon int, challengerWon int)` or possibly passing a `map[GameStatus]int` makes sense but then you need to ensure you treat anything not present as 0 so it doesn't fail to update when a particular game status isn't present.

op-challenger/metrics/metrics.go
```diff
@@ -58,9 +65,25 @@ func NewMetrics() *Metrics {
...
 }
 
+const (
+	// GameStatusInProgress is the status of a game that is in progress.
+	GameStatusInProgress = "in_progress"
+	// GameStatusChallengerWon is the status of a game that the challenger won.
+	GameStatusChallengerWon = "challenger_won"
+	// GameStatusDefenderWon is the status of a game that the defender won.
+	GameStatusDefenderWon = "defender_won"
+)
```
ajsutton: Label values can be pretty free form so I think we should just use `GameStatus.String()` to get the label value. Saves having to manually maintain an additional mapping of states to strings.

op-challenger/metrics/metrics.go
```diff
@@ -84,3 +107,27 @@ func (m *Metrics) RecordUp() {
...
+func (m *Metrics) ClearTrackedGames() {
+	m.trackedGames.Delete(prometheus.Labels{
+		"status": GameStatusInProgress,
+	})
+	m.trackedGames.Delete(prometheus.Labels{
+		"status": GameStatusChallengerWon,
+	})
+	m.trackedGames.Delete(prometheus.Labels{
+		"status": GameStatusDefenderWon,
+	})
```
ajsutton: We don't want to delete - we want to set to 0.  The game statuses still exist, we just have no games in that status.

op-challenger/metrics/metrics.go
refcell: I think the precedent set in the batcher's metrics are to make metricer functions explicit in what the do. Allowing any string to be passed in as a label seems like an antipattern to me since there's no input bounds. If you feel strongly that we should accept a string argument, I can change the functions to one that just accepts a string though.

ajsutton: I definitely wouldn't accept an arbitrary string. I was expecting that `Metricer` would accept a `GameState` which is an enum with a fixed set of values, then it would convert that to a string using the `GameState.String()` function internally.

refcell: I updated the game progression to return a `types.GameStatus` and pushed metrics recording into the coordinator. The issue is this tightly couples the scheduler package with internal fault types (ie the GameStatus type). What if we hoisted the GameStatus type to be a `game/` package level type as opposed to an internal fault type?

ajsutton: Ah yeah, `GameStatus` is not fault dispute specific - it applies to all game types.  So makes sense to move it up to the game level.  Good spot.

op-challenger/metrics/metrics.go
```diff
@@ -120,3 +134,17 @@ func (m *Metrics) RecordGameStep() {
...
+
+func (m *Metrics) RecordGameDefenderWon() {
+	m.trackedGames.WithLabelValues("in_progress").Dec()
+	m.trackedGames.WithLabelValues("defender_won").Inc()
+}
+
+func (m *Metrics) RecordGameChallengerWon() {
+	m.trackedGames.WithLabelValues("in_progress").Dec()
+	m.trackedGames.WithLabelValues("challenger_won").Inc()
+}
```
ajsutton: This is going to count games multiple times on every update.  We need a single method that accepts the total counts for each game type and then just set those values on the gauge.

refcell: Right, good callout - I'm not really seeing where this atomic update would happen. The coordinator just routes jobs individually and never has a complete view of games with updated game types as far as I understand.

ajsutton: The coordinator always gets a full list of games - that's how it can produce the list of still-in-progress games that it shouldn't delete files for.  So the coordinator `schedule` method could compute the totals and make the call to update metrics at the end.  Its `createJob` method should probably just be updated to return the job to execute (possibly null as it is now) plus also return the current status as its recorded in the `gameState`. I wouldn't make another call to fetch the game status, just depend on it being updated whenever the game progresses.

ajsutton: ah yeah admittedly the delete code didn't actually wind up in the `schedule` method which makes it harder to see that - it does remove game states for any games not in the list which also depends on the fact that its given the full list of active games.

It does have the allowlist already applied so there are some games it doesn't "see" but I think that's probably good - then the metric reports only the games that the challenger is actually tracking.

refcell: Ok attempted to change it up to this - let me know if this is along the lines of what you're thinking

op-challenger/game/scheduler/coordinator.go
```diff
@@ -49,18 +52,31 @@ func (c *coordinator) schedule(ctx context.Context, games []common.Address) erro
...
 			jobs = append(jobs, *j)
+
+			switch j.status {
+			case types.GameStatusInProgress:
+				gamesInProgress++
+			case types.GameStatusDefenderWon:
+				gamesDefenderWon++
+			case types.GameStatusChallengerWon:
+				gamesChallengerWon++
+			}
```
ajsutton: This is the right idea, but we should also count games that we didn't create a job for. Otherwise we would miss counting games that are already in flight and so we're not scheduling a second job for. 

Either `createJob` will need to return the game status or once `createJob` has run its safe to assume the `gameState` exists in the tracking map so you could just grab the status from there - not sure which is cleaner/safer...

refcell: Attempted to fix this - read the state on the non-matching case with an ok check.

op-challenger/game/scheduler/coordinator.go
```diff
@@ -49,18 +52,43 @@ func (c *coordinator) schedule(ctx context.Context, games []common.Address) erro
...
+			statusSet = true
+		} else {
+			// Read from the game state
+			state, ok := c.states[addr]
+			if ok {
+				status = state.status
+				statusSet = true
+			}
+		}
+		if statusSet {
```
ajsutton: If a game is executing a job it _must_ have a `gameState` in the map so I don't think we need these two separate paths to get the data.  If we're going to get the status from the map, just do that all the time.

Also since the status is expected to be present, we should log an error if its not there rather than silently ignoring.

refcell: Fixed.

ajsutton (approved):

PR #7082, opened 2023-09-01
op-challenger: Don't try to perform moves on resolvable games.

ajsutton:
**Description**

Once a game becomes resolvable, there's no point trying to perform moves or steps because any move that would affect the game outcome will be rejected because of expired clocks. We already skipped performing moves if we successfully resolved the game, but not if we were in a losing position or there was an error resolving the game.

**Tests**

Added unit tests.

**Metadata**

- Fixes https://linear.app/optimism/issue/CLI-4396/op-challenger-dont-attempt-to-perform-moves-after-clock-has-expired


refcell (approved):
LGTM

PR #7086, opened 2023-09-01
feat(op-challenger): Inflight game metrics

refcell:
**Description**

Introduces a gauge to track inflight games in the scheduler.

**Metadata**

Fixes CLI-4395.

### DIFF
--- a/op-challenger/game/scheduler/scheduler.go
+++ b/op-challenger/game/scheduler/scheduler.go
@@ -13,6 +13,8 @@ var ErrBusy = errors.New("busy scheduling previous update")
 
 type SchedulerMetricer interface {
 	RecordGamesStatus(inProgress, defenderWon, challengerWon int)
+	RecordGameUpdateScheduled()
+	RecordGameUpdateCompleted()
 }
 
 type Scheduler struct {
--- a/op-challenger/metrics/noop.go
+++ b/op-challenger/metrics/noop.go
@@ -19,3 +19,6 @@ func (*noopMetrics) RecordGameStep() {}
 func (*noopMetrics) RecordCannonExecutionTime(t float64) {}
 
 func (*noopMetrics) RecordGamesStatus(inProgress, defenderWon, challengerWon int) {}
+
+func (*noopMetrics) RecordGameUpdateScheduled() {}
+func (*noopMetrics) RecordGameUpdateCompleted() {}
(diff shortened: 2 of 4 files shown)


refcell:
* develop
  * **PR #7080** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7080" target="_blank">Graphite</a> 
    * **PR #7086** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7086" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7086?utm_source=stack-comment).

op-challenger/game/scheduler/scheduler.go
```diff
@@ -84,10 +88,12 @@ func (s *Scheduler) loop(ctx context.Context) {
 			if err := s.coordinator.schedule(ctx, games); err != nil {
 				s.logger.Error("Failed to schedule game updates", "games", games, "err", err)
 			}
+			s.m.RecordGamesInflight(len(games))
```
ajsutton: This is the wrong place for this metric. It will increment by the total number of games being tracked all the time even if some of them don't actually schedule an update.  The increment needs to be inside the coordinator component and only apply to games where a new `job` is actually being scheduled.

op-challenger/metrics/noop.go
```diff
@@ -19,3 +19,6 @@ func (*noopMetrics) RecordGameStep() {}
 func (*noopMetrics) RecordCannonExecutionTime(t float64) {}
 
 func (*noopMetrics) RecordGamesStatus(inProgress, defenderWon, challengerWon int) {}
+
+func (*noopMetrics) RecordGamesInflight(n int)  {}
+func (*noopMetrics) RecordGamesProcessed(n int) {}
```
ajsutton: I'd make these `RecordGameUpdateScheduled()` and `RecordGameUpdateComplete()` so its clear that they're related to each other and incrementing/decrementing the value rather than replacing it.

ajsutton (commented):
If you run `make alphabet` and check this metric (`curl http://localhost:7304/metrics  | grep inflight`) it shows the problems quite clearly. For me it's reporting 10 inflight games when actually there's only 1 game that exists at all and its now mostly idle so the metric should be reporting 0 or 1.  It's really important to actually run and verify the changes being proposed - tests are powerful but only verify the code does what you expect, not if that results in the right result. :)

ajsutton (approved):

PR #7105, opened 2023-09-04
Dockerfiles: Run go mod download with just the go.mod and go.sum files

ajsutton:
**Description**

Avoids the need to re-run mod download when any source file changes and means most dockerfiles can reuse the cached layer with download already done.


Inphi (approved):

roninjin10 (approved):
Nice

PR #7111, opened 2023-09-05
op-node: implement runtime config reloading

protolambda:
**Description**

- Add runtime-config (the non-derivation related part of the SystemConfig contents) reload background loop
- Add flag to make reload interval configurable (can also be disabled)
- Apply verifier confirmation-depth to runtime config reload
- Update initial config loading to use the same reload code, and the retry util
- Add getter to op-node to get read-only runtime config

**Tests**

Added op-e2e test that checks if the verifier picks up on a runtime-config change through reloading, by changing the p2p sequencer key (also known as the unsafe-block signer addresss) in the L1 SystemConfig.

**Additional context**

This same config reloading will be extended with the protocol-version pulling, in #7068 

TBD: whether or not we use the same polling loop for both runtime config data contracts, or separate them (can still reuse code).

Ideally we also add a separate event-log based updating mechanism to the runtime-config, by piping out the System-config updates from the derivation, to get instant event-based updates. This is currently low-priority for the current scope of config data however.

**Metadata**

Fix CLI-4398


op-e2e/system_test.go
```diff
@@ -1388,3 +1388,48 @@ func TestPendingBlockIsLatest(t *testing.T) {
...
+	opts, err := bind.NewKeyedTransactorWithChainID(cfg.Secrets.SysCfgOwner, cfg.L1ChainIDBig())
+	require.Nil(t, err)
+	// the unsafe signer address is part of the runtime config
+	tx, err := sysCfgContract.SetUnsafeBlockSigner(opts, newUnsafeBlocksSigner)
+	require.NoError(t, err)
+
+	// wait for the change to confirm
+	receipt, err := bind.WaitMined(context.Background(), l1, tx)
+	require.NoError(t, err)
+	require.Equal(t, types.ReceiptStatusSuccessful, receipt.Status)
```
ajsutton: I think we can just use `wait.ForReceiptOK` here.

op-node/node/node.go
```diff
@@ -159,27 +159,70 @@ func (n *OpNode) initRuntimeConfig(ctx context.Context, cfg *Config) error {
...
-		return nil
+	// initialize the runtime config before unblocking
+	if _, err := retry.Do(ctx, 5, retry.Fixed(time.Second*10), func() (eth.L1BlockRef, error) {
+		return reload(ctx)
+	}); err != nil {
+		return fmt.Errorf("failed to load runtime configuration repeatedly, last error: %w", err)
 	}
 
-	return errors.New("failed to load runtime configuration repeatedly")
+	// start a background loop, to keep reloading it every 10 minutes
```
ajsutton: nit: It's not a fixed 10 minute duration now.
```suggestion
	// start a background loop, to keep reloading it at the configured reload interval
```

ajsutton (approved):
LGTM.

PR #7114, opened 2023-09-06
Create an e2e test showing cannon proving an output root is invalid

ajsutton:
**Description**

Creates a new fault proof e2e test which is disputing an invalid output proposal. This gives us a cannon game where we expect the defender to win instead of the challenger.

We don't yet have e2e helpers to try and call the `step` function which is required to really proof the invalid trace can't be validated but this is a useful example even as it is.

**Additional context**
We'll likely need to expand these kind of "cannon with invalid output root" tests to replace the alphabet based tests we had which become invalid in #6913.

**Metadata**

- https://github.com/ethereum-optimism/client-pod/issues/21

### DIFF
--- a/op-e2e/e2eutils/l2oo/helper.go
+++ b/op-e2e/e2eutils/l2oo/helper.go
@@ -1,0 +1,130 @@
+package l2oo
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"math/big"
+	"testing"
+	"time"
+
+	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
+	"github.com/ethereum-optimism/optimism/op-chain-ops/genesis"
+	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils/wait"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/stretchr/testify/require"
+)
+
+type L2OOHelper struct {
+	t       *testing.T
+	require *require.Assertions
+	client  *ethclient.Client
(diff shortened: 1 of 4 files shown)


op-e2e/e2eutils/l2oo/helper.go
```diff
@@ -0,0 +1,130 @@
...
+type L2OOHelper struct {
+	t       *testing.T
+	require *require.Assertions
+	client  *ethclient.Client
+	l2oo    *bindings.L2OutputOracle
+
+	// Nil when read-only
+	transactOpts *bind.TransactOpts
+	rollupCfg    *rollup.Config
+}
```
roninjin10, line 29: Just for my own understanding is it normal to set some properties nil sometimes in golang? My intuition here is you would need to make L200Helper and l2HelperReadOnly seperate structs repeating t, require, client, and l2oo props between the 2. I never would have thought of making pointers nullable

roninjin10 (approved):
This is cool!

op-e2e/e2eutils/l2oo/helper.go
refcell: So since they're non or unexported fields (the first letter is not upper cased), they cannot be accessed outside the golang package. As such, it is intended to not reference the internal variables of the `L2OOHelper` struct directly, but rather to use the public-facing, exposed api accessor and mutator methods. If these are implemented correctly, the internal behavior of making internal fields nullable is abstracted away from external users and IMO it's acceptable behavior. I think monadic type representations are just better all around, though.

refcell (approved):

op-e2e/e2eutils/l2oo/helper.go
Inphi: @roninjin10 yes it's quite normal and idiomatic in Go. All types have well-defined defaults in Go, including pointers, and programmers are encouraged to utilize this fact for simplicity.
I agree with refcell that monads would be better, but then the language is no longer Go 😄

PR #7121, opened 2023-09-06
Add the DelayedVetoable contract

maurelian:
**Description**

This PR adds the Delayed Vetoable contract. This contract implements a transparent pattern, so that the initiator of a call can queue it up, and any account can forward it.

Some notable functionality has not yet been added:
- A nonce to ensure that transactions are executed in order, as this is likely to be ensured simply by a delay between calls being initiated, and the fact that finalizing a call after the delay is permissionless.
- The ability to modify the authorized accounts and the delay time. These values will soon be moved to the new `SuperChainConfig` contract.

Note also that this contract is currently _not_ a singleton, there is one per chain. We may choose to change that in the future as the design evolves, but this arch is fine for now and meets the current requirements.


packages/contracts-bedrock/src/universal/DelayedVetoable.sol
```diff
@@ -0,0 +1,61 @@
...
+        address target = _target;
+        emit Forwarded(msg.data);
+        assembly {
+            // Copy calldata into memory at 0x0....calldatasize.
+            calldatacopy(0x0, 0x0, calldatasize())
+
+            // TODO(maurelian): can we emit this in the assembly block to deduplicate
+            //   getting the calldata? I think I was doing it correctly, but the forge
+            //   test doesn't recognize this
+            // log1(0x0, calldatasize(), topic)
```
clabby:
```suggestion
    function _handleCall() internal {
        address target = _target;
        require(target != address(0), "DelayedVetoable: target not initialized");
        bytes32 forwarded = Forwarded.selector;

        assembly {
            // Copy calldata into memory at 0x0....calldatasize.
            calldatacopy(0x0, 0x0, calldatasize())

            // Log the `Forwarded` event with the passed calldata.
            log1(0x0, calldatasize(), forwarded)
```

2 things here:
1. We can save an extra `sload` by caching `_target` on the stack prior to the check.
2. `log1` can be understood by forge, but we'll have to also update the test to use a raw log.

```solidity
contract DelayedVetoable_HandleCall_Test is DelayedVetoable_Init {
    function testFuzz_handleCall_succeeds(bytes memory data) external {
        bytes32 topic = Forwarded.selector;

        vm.expectCall(target, data);
        vm.expectEmit(true, false, false, true, address(delayedVetoable));
        assembly {
            log1(add(data, 0x20), mload(data), topic)
        }

        (bool success,) = address(delayedVetoable).call(data);
        assert(success);
    }
}
```

packages/contracts-bedrock/test/invariants/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,33 @@
...
+
+    // The address that delayedVetoable will call to
+    address dvTarget;
+
+    function setUp() public {
+        delayedVetoable = new DelayedVetoable({
+            target: address(dvTarget)
+        });
+
+        targetContract(address(delayedVetoable));
```
clabby: As it stands this won't work. Proxies / call forwarders with invariant tests are kind of annoying - the fuzzer uses the contract's abi to crawl it, and the `DelayedVotable` contract only has a fallback. You'll probably need to target an actor a la:

```solidity
contract DelayedVetoableCaller {
    Vm internal immutable VM;

    DelayedVetoable public dv;
    bool public failed;

    constructor(Vm _vm) {
        VM = _vm;
    }

    function setDelayedVetoable(address _target) public {
        if (_target == address(0)) {
            _target = VM.addr(uint256(keccak256(abi.encode(_target))));
        }

        dv = new DelayedVetoable({ target: _target });
    }

    function doCall(bytes memory _toForward) external {
        if (address(dv) == address(0)) {
            setDelayedVetoable(VM.addr(uint256(keccak256(_toForward))));
        }

        bytes32 forwardHash = keccak256("Forwarded(bytes)");
        VM.expectCall(address(uint160(uint256(VM.load(address(dv), bytes32(0))))), _toForward);
        VM.expectEmit(true, false, false, true, address(dv));
        assembly {
            log1(add(_toForward, 0x20), mload(_toForward), forwardHash)
        }

        (bool success,) = address(dv).call(_toForward);
        if (!success) {
            failed = true;
        }
    }
}
```

and adjust the condition of the invariant test to:

```solidity
contract DelayedVetoable_Invariants is DelayedVetoable_Invariant_Harness {
    /// @custom:invariant Calls are always forwarded to the target
    function invariant_callIsForwarded_succeeds() public {
        assertFalse(actor.failed() || this.failed());
    }
}
```

clabby (dismissed):
Looks good so far - couple nits on the invariant test as well as a suggestion for the event emission in inline asm.

packages/contracts-bedrock/src/universal/DelayedVetoable.sol
maurelian: Thanks, very helfpful suggestions. For now I've decided to remove the assembly log emission to keep things simple.
Can re-introduce in the future if the cost of an upgrade becomes excessive.

packages/contracts-bedrock/test/invariants/DelayedVetoable.t.sol
maurelian: As we discussed offline, I've decided to remove these tests for now.
They may be added back later on, and I've saved a copy of them.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+    modifier handleCallIfNotVetoer() {
+        if (msg.sender == _vetoer || msg.sender == address(0)) {
+            _;
+        } else {
+            // This WILL halt the call frame on completion.
+            _handleCall();
+        }
+    }
+
+    /// @custom:semver 0.0.1
```
tynes: Comment doesn't match inline version

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+    event Vetoed(bytes32 indexed callHash, bytes data);
+
+    /// @notice The address that all calls are forwarded to after the delay.
+    address internal _target;
+
+    // TODO(maurelian): move this to the new SuperChainConfig contract
+    // @notice The address that can veto a call.
+    address internal _vetoer;
+
+    // TODO(maurelian): move this to the new SuperChainConfig contract
```
tynes: Do we want to couple this to another contract? Feels like this should work outside of the context of op stack

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
```
tynes: Should this be exactly 15?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+        } else {
+            // The ability to finalize the call after sufficient time has passed does not require
+            // authorization.
+
+            // Delete the call to prevent replays
+            delete _queuedAt[callHash];
+
+            // Forward the call
+            emit Forwarded(callHash, msg.data);
+            (bool success,) = _target.call(msg.data);
```
tynes: Should forward value too

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /// @custom:semver 0.0.1
+    /// @notice Sets the target admin during contract deployment.
+    /// @param vetoer_ Address of the vetoer.
+    /// @param initiator_ Address of the initiator.
+    /// @param target_ Address of the target.
+    /// @param delay_ Address of the delay.
+    constructor(address vetoer_, address initiator_, address target_, uint256 delay_) Semver(0, 1, 1) {
+        _vetoer = vetoer_;
```
tynes: Do we need vetoer and initializer to be different? Can't this just be ownable?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+
+        delete _queuedAt[callHash];
+        emit Vetoed(callHash, data);
+    }
+
+    /// @notice Receives all calls other than those made by the vetoer.
+    ///         This enables transparent initiation and forwarding of calls to the target and avoids
+    ///         the need for additional layers of abi encoding.
+    function _handleCall() internal {
+        if (_target == address(0)) {
```
tynes: This makes no sense to happen at runtime, should happen at deploy time since target is not mutable

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /// @notice Used for all calls that pass data to the contract.
+    fallback() external payable {
+        _handleCall();
+    }
+
+    /// @notice Vetoes a call. This method can only be called by the vetoer. If called by another
+    ///         address, execution will be redirected to _handleCall()
+    function veto(bytes memory data) external handleCallIfNotVetoer {
```
tynes: Do you have strong preference for this being in the abi? I suppose it does simplify things when interacting with the contract but I imagined it being completely transparent

tynes:
Left some comments

tynes (commented):
Left some comments

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: I think putting these values in the superchain config is necessary since they are global across OP chains. 

But I could maybe make these addresses be returned by virtual functions so that this is more reusable.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: Initializer is the security council, vetoer is the foundation.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: True

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: That breaks the current mechanism of forwarding by resubmitting the data. See the comment on line 97, if you're ok with the trade off we can add value.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: This made the handleCall func a bit cleaner, but I'm not strongly opinionated. 

It's also already not perfectly transparent due to the getters.

maurelian: I took out this method.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: Thinking about this more, I do think that we're going to need to allow forwarding value.

In order to achieve that, I will make it so that only the `initiator` passes in unencoded data. All other parties (`vetoer` or any account attempting to forward) will identify the transaction they want to interact with by submitting a tuple of `(nonce, callhash, value)`.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+    /// @param data The data forwarded to the target.
+    event Forwarded(bytes32 indexed callHash, bytes data);
+
+    /// @notice An event that is emitted each time a call is vetoed.
+    /// @param callHash The hash of the call data.
+    /// @param data The data forwarded to the target.
+    event Vetoed(bytes32 indexed callHash, bytes data);
+
+    /// @notice The address that all calls are forwarded to after the delay.
+    address internal _target;
```
tynes: The addresses can be immutable? any reason to be in storage?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    /// @custom:semver 0.0.1
+    /// @notice Sets the target admin during contract deployment.
+    /// @param vetoer_ Address of the vetoer.
+    /// @param initiator_ Address of the initiator.
+    /// @param target_ Address of the target.
+    /// @param delay_ Address of the delay.
+    constructor(address vetoer_, address initiator_, address target_, uint256 delay_) Semver(0, 0, 1) {
```
tynes: If you rebase then you can use `ISemver` instead of `Semver`

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+
+        // Case 1: The initiator is calling the contract to initiate a call.
+        if (msg.sender == _initiator && _queuedAt[callHash] == 0) {
+            _queuedAt[callHash] = block.timestamp;
+            emit Initiated(callHash, msg.data);
+            return;
+        }
+
+        // Case 2: The vetoer is calling the contract to veto a call.
+        if (msg.sender == _vetoer && _queuedAt[callHash] != 0 && block.timestamp <= _queuedAt[callHash] + _delay) {
```
tynes: The vetoer has the ability to lose its ability to veto on maturation?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: no, good point.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    // TODO(maurelian): move this to the new SuperChainConfig contract
+    // @notice The address that can initiate a call.
+    address internal _initiator;
+
+    /// @notice The time that a call was initiated.
+    mapping(bytes32 => uint256) internal _queuedAt;
+
+    // @notice The time to wait before forwarding a call.
+    uint256 internal _delay;
```
tynes, line 58: Ability to set `_delay` by the vetoer or initiator if delay is 0

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: yeah, I kind of think that's the right approach. 
IMO this basically just makes the correct behavior (of the vetoer) explicit.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Semver } from "../universal/Semver.sol";
```
tynes: `src/universal/Semver.sol`

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: In order to keep the contract fully transparent to the initiator, the delay can be activated by either party by submitting `null` data. Feedback welcome on that approach.

https://github.com/ethereum-optimism/optimism/blob/af3aa7be5ce3446718dc177c4e382f53190c18c6/packages/contracts-bedrock/src/L1/DelayedVetoable.sol#L126-L134

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: Hmm.... For short delay periods this does make censorship attacks against the vetoer a bit easier.
I now lean towards removing this to allow for a bidding war to persist past the delay, wdyt?

maurelian: I ended up making that change and including a comment to clarify the rationale.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,195 @@
...
+    /// @param data The data forwarded to the target.
+    event Forwarded(bytes32 indexed callHash, bytes data);
+
+    /// @notice An event that is emitted each time a call is vetoed.
+    /// @param callHash The hash of the call data.
+    /// @param data The data forwarded to the target.
+    event Vetoed(bytes32 indexed callHash, bytes data);
+
+    /// @notice The address that all calls are forwarded to after the delay.
+    address internal immutable _target;
```
tynes: immutables should be camel cased right?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,195 @@
...
+    /// @notice Sets the target admin during contract deployment.
+    /// @param vetoer_ Address of the vetoer.
+    /// @param initiator_ Address of the initiator.
+    /// @param target_ Address of the target.
+    /// @param operatingDelay_ Time to delay when the system is operational.
+    constructor(address vetoer_, address initiator_, address target_, uint256 operatingDelay_) {
+        _vetoer = vetoer_;
+        _initiator = initiator_;
+        _target = target_;
+        _delay = 0;
```
tynes: We don't need to explicitly set delay here to 0, it will be by default. Prefer a comment explaining why its initially 0

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,195 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { ISemver } from "src/universal/ISemver.sol";
+
+contract DelayedVetoable is ISemver {
```
tynes, line 13: Could you add natspec here?

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,177 @@
...
+        assertEq(delayedVetoable.target(), target);
+        assertEq(delayedVetoable.delay(), operatingDelay);
+    }
+}
+
+contract DelayedVetoable_Getters_TestFail is DelayedVetoable_Init {
+    function test_getters_notVetoer() external {
+        // getter calls from addresses other than the zero address will revert in the
+        // initiation branch of the proxy.
+        vm.assume(msg.sender != address(0) && msg.sender != initiator && msg.sender != vetoer);
```
tynes: What does `vm.assume` do in a non fuzz test?

maurelian: oh nothing, that's a bug good call out.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: OK. The var was documented in its own natspec, but made this change.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,199 @@
...
+    address internal immutable TARGET;
+
+    /// @notice The address that can veto a call.
+    address internal immutable VETOER;
+
+    /// @notice The address that can initiate a call.
+    address internal immutable INITIATOR;
+
+    /// @notice The current amount of time to wait before forwarding a call.
+    uint256 internal _delay;
```
tynes: nit: group immutables together so storage layout is easier to deduce

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /// @notice Receives all calls other than those made by the vetoer.
+    ///         This enables transparent initiation and forwarding of calls to the target and avoids
+    ///         the need for additional layers of abi encoding.
+    function _handleCall() internal {
+        // The initiator and vetoer activate the delay by passing in null data.
+        if (msg.data.length == 0) {
+            if (msg.sender != INITIATOR && msg.sender != VETOER) {
+                // todo(maurelian): make this error have an expected array.
```
tynes: What do you mean by this todo?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,199 @@
...
+        _forwardAndHalt(callHash);
+    }
+
+    /// @notice Forwards the call to the target and halts the call frame.
+    function _forwardAndHalt(bytes32 callHash) internal {
+        // Forward the call
+        emit Forwarded(callHash, msg.data);
+        (bool success,) = TARGET.call(msg.data);
+        assembly {
+            // Success == 0 means a revert. We'll revert too and pass the data up.
```
tynes: This passes the data up? The call will copy the memory back but not at the zero'th offset?

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,175 @@
...
+        assert(success);
+    }
+
+    /// @dev The delay can be activated by the vetoer or initiator, and are not forwarded until the delay has passed
+    ///      once activated.
+    function testFuzz_handleCall_forwardingWithDelay_succeeds(bytes memory data) external {
+        assumeNonzeroData(data);
+
+        vm.prank(initiator);
+        // it's immediately forwarding for some reason.
```
tynes: what do you mean by this comment?

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,175 @@
...
+        (bool success,) = address(delayedVetoable).call(data);
+
+        vm.warp(block.timestamp + operatingDelay);
+        vm.expectEmit(true, false, false, true, address(delayedVetoable));
+        emit Forwarded(keccak256(data), data);
+
+        // Forward the call
+        vm.expectCall({ callee: target, data: data });
+        (success,) = address(delayedVetoable).call(data);
+        assert(success);
```
tynes: why assert? usually we use `assertTrue`?

packages/contracts-bedrock/test/DelayedVetoable.t.sol
maurelian: Debugging comment. Removed.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: The `Unauthorized(expected, actual)` error lists a single expected address, but it could have been either the `INITIATOR` or the `VETOER`. But I think this is close enough, and just removed the comment. Setting up an array here adds a lot of noise for not much benefit.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: Oh, good catch.  
I fixed this and made sure the tests assert on the returndata:
https://github.com/ethereum-optimism/optimism/blob/e2c0949d0496edb7937e6ca5446c7713884d0a90/packages/contracts-bedrock/test/DelayedVetoable.t.sol#L101
https://github.com/ethereum-optimism/optimism/blob/e2c0949d0496edb7937e6ca5446c7713884d0a90/packages/contracts-bedrock/test/DelayedVetoable.t.sol#L177-L179

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,200 @@
...
+    fallback() external {
+        _handleCall();
+    }
+
+    /// @notice Receives all calls other than those made by the vetoer.
+    ///         This enables transparent initiation and forwarding of calls to the target and avoids
+    ///         the need for additional layers of abi encoding.
+    function _handleCall() internal {
+        // The initiator and vetoer activate the delay by passing in null data.
+        if (msg.data.length == 0) {
```
tynes: This means that we will never be able to forward a call with no data

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,181 @@
...
+contract DelayedVetoable_HandleCall_Test is DelayedVetoable_Init {
+    /// @dev A call can be initiated by the initiator.
+    function testFuzz_handleCall_initiation_succeeds(bytes memory data) external {
+        assumeNonzeroData(data);
+
+        vm.expectEmit(true, false, false, true, address(delayedVetoable));
+        emit Initiated(keccak256(data), data);
+
+        vm.prank(initiator);
+        (bool success,) = address(delayedVetoable).call(data);
```
tynes, line 78: Can we also check `_queuedAt` here? It is an important part of initiation and will capture the state diff of this code path

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,181 @@
...
+        assertEq(returnData, outData);
+    }
+
+    /// @dev The delay can be activated by the vetoer or initiator, and are not forwarded until the delay has passed
+    ///      once activated.
+    function testFuzz_handleCall_forwardingWithDelay_succeeds(bytes memory data) external {
+        assumeNonzeroData(data);
+
+        vm.prank(initiator);
+        (bool success,) = address(delayedVetoable).call(data);
```
tynes, line 113: Can we check the mapping for presence and then it being deleted?

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,181 @@
...
+        (success,) = address(delayedVetoable).call(data);
+        assertTrue(success);
+    }
+}
+
+contract DelayedVetoable_HandleCall_TestFail is DelayedVetoable_Init {
+    /// @dev The delay is inititially set to zero and the call is immediately forwarded.
+    function test_handleCall_unauthorizedInitiation_reverts() external {
+        vm.expectRevert(abi.encodeWithSelector(Unauthorized.selector, initiator, address(this)));
+        (bool success,) = address(delayedVetoable).call(NON_ZERO_DATA);
```
tynes, line 134: I am confused here, expect revert but success is true?

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,181 @@
...
+
+        vm.warp(block.timestamp + operatingDelay);
+        vm.expectEmit(true, false, false, true, address(delayedVetoable));
+        emit Forwarded(keccak256(inData), inData);
+
+        vm.mockCallRevert(target, inData, outData);
+
+        // Forward the call
+        vm.expectRevert(outData);
+        (bool success2,) = address(delayedVetoable).call(inData);
```
tynes, line 185: Could you also assert eq on the `bytes memory` of the `call`?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,200 @@
...
+            _;
+        } else {
+            // This WILL halt the call frame on completion.
+            _handleCall();
+        }
+    }
+
+    /// @notice Semantic version.
+    /// @custom:semver 0.0.1
+    string public constant version = "0.0.1";
```
tynes: We will need to remember to bump this to `1.0.0`, to be honest I think this contract is basically "done" pending any small bugfixes so perhaps just setting it to `1.0.0` as part of this PR makes sense

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,200 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { ISemver } from "src/universal/ISemver.sol";
+
+/// @title DelayedVetoable
+/// @notice This contract enables a delay before a call is forwarded to a target contract, and during the delay period
+///         the call can be vetoed by the authorized vetoer.
```
tynes, line 8: Perhaps add comment about explicitly not supporting value transfers

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
tynes: If we wanted to make this branch be `msg.data.length == 0 && _delay == 0` it would relax this constraint. Slightly more gas on each call but makes the state machine more explicit in that branch only happens once

tynes:
I like this contract a lot, great work. I wish there was an easy way to generate a state machine diagram for it, its very simple and would be a good resource for "how to think about smart contracts"

tynes:
Thoughts on moving this to `universal` instead of `L1`?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,200 @@
...
+
+    /// @notice Gets the initiator
+    /// @return initiator_ Initiator address.
+    function initiator() external virtual readOrHandle returns (address initiator_) {
+        initiator_ = INITIATOR;
+    }
+
+    //// @notice Queries the vetoer address.
+    /// @return vetoer_ Vetoer address.
+    function vetoer() external virtual readOrHandle returns (address vetoer_) {
```
tynes, line 102: I am curious if etherscan will work correctly with this

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,200 @@
...
+    /// @notice Semantic version.
+    /// @custom:semver 0.0.1
+    string public constant version = "0.0.1";
+
+    /// @notice Sets the target admin during contract deployment.
+    /// @param vetoer_ Address of the vetoer.
+    /// @param initiator_ Address of the initiator.
+    /// @param target_ Address of the target.
+    /// @param operatingDelay_ Time to delay when the system is operational.
+    constructor(address vetoer_, address initiator_, address target_, uint256 operatingDelay_) {
```
tynes, line 85: I don't think you can make the constructor virtual? There will be a few unnecessary variables on a child contract but i think that will be ok, they will just end up in the bytecode and be useless

packages/contracts-bedrock/test/DelayedVetoable.t.sol
maurelian: It super weird see the "gotcha" [here](https://book.getfoundry.sh/cheatcodes/expect-revert?highlight=expectrevert#description). 

I'm going to delete the `assertTrue` statement, it's just confusing and redundant with `expectRevert`.

I think that the only reason I added the `assertTrue` was to verify the gotcha.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: > This means that we will never be able to forward a call with no data

I don't really see why we'd ever want to make a call with no data and no value. The internal function it calls does revert if `_delay != 0`. But this does add functionality, it's a good suggestion.

I'm going to:
1. delete `activateDelay` and inline that logic
2. remove the revert and just skip this if delay is zero.

packages/contracts-bedrock/test/DelayedVetoable.t.sol
maurelian: According to that same [gotcha](https://book.getfoundry.sh/cheatcodes/expect-revert?highlight=expectrevert#description) above: 

> Apart from this, `expectRevert` also mangles return data on low level calls, and is not usable.

This caused me much confusion

maurelian: also, `vm.expectRevert()` should be handling that assertion

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: I tried, no. 
> `Constructors cannot be virtual. solidity(7001)`

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,193 @@
...
+    /// @dev The delay is inititially set to zero and the call is immediately forwarded.
+    function testFuzz_handleCall_initialForwardingImmediately_succeeds(
+        // bytes memory inData,
+        bytes memory outData
+    )
+        external
+    {
+        // Clashes with initiator() (no problem)
+        // bytes memory inData = hex"5c39fcc100000000000000000000000000000000000000000000000000000000";
+        // Clashes with queuedAt(bytes32)
```
tynes:
```
vm.assume(inData.length >= 4 && inData[0:4] == keccak256("queuedAt(bytes32)")
```

packages/contracts-bedrock/test/DelayedVetoable.t.sol
```diff
@@ -0,0 +1,193 @@
...
+
+        vm.prank(initiator);
+        (bool success,) = address(delayedVetoable).call(data);
+        assertTrue(success);
+    }
+
+    /// @dev The delay is inititially set to zero and the call is immediately forwarded.
+    function testFuzz_handleCall_initialForwardingImmediately_succeeds(
+        // bytes memory inData,
+        bytes memory outData
```
tynes: Can you turn this into calldata?

tynes:
This looks like it needs a rebase

.gitmodules
```diff
@@ -13,3 +13,10 @@
 [submodule "packages/contracts-bedrock/lib/forge-std"]
 	path = packages/contracts-bedrock/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
+[submodule "packages/contracts-bedrock/lib/safe-contracts"]
+	path = packages/contracts-bedrock/lib/safe-contracts
+	url = https://github.com/safe-global/safe-contracts
+	branch = v1.4.0
+
+
```
tynes: why extra whitespace here?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,198 @@
...
+        bytes32 callHash = keccak256(msg.data);
+
+        // Case 1: The initiator is calling the contract to initiate a call.
+        if (msg.sender == INITIATOR && _queuedAt[callHash] == 0) {
+            if (_delay == 0) {
+                // This forward function will halt the call frame on completion.
+                _forwardAndHalt(callHash);
+            }
+            _queuedAt[callHash] = block.timestamp;
+            emit Initiated(callHash, msg.data);
```
tynes, line 153: Curious if you put thought into why the digest and data here, you can derive the hash from the data pretty easily, but making the digest indexed makes sense, then you can offchain find the receipt easily for a given callhash to know the timestamp

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -0,0 +1,198 @@
...
+        target_ = TARGET;
+    }
+
+    /// @notice Gets the delay
+    /// @return delay_ Delay address.
+    function delay() external readOrHandle returns (uint256 delay_) {
+        delay_ = _delay;
+    }
+
+    /// @notice Gets entries in the _queuedAt mapping.
```
tynes, line 118: I am not suggesting we do this but what if a future design pattern for transparent contracts involved a read contract and a write contract. Not sure if this makes sense, but it would solve the problem that we have here with `queuedAt(bytes32)`

tynes (approved):

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: That's an awesome idea, makes a ton of sense. 
The core contract could route calls from the reader contract to a transparent version of the [extsload](https://eips.ethereum.org/EIPS/eip-2330).

.gitmodules
maurelian: Manual editing plus being reliant on the IDE auto-formatting which it doesn't give me for this file.

But the correct changes are already in `develop`, so I'm gonna remove this diff.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: I have thought about it. 
IMO it's easy to get the hash from the logs, but this will save effort for off-chain devs and facilitate filtering/indexing.

We've felt a lack of DX on many of the events in the bedrock contracts, so I felt it made sense to risk logging too much rather than too little.

PR #7122, opened 2023-09-06
specs: Update Dispute Game Specifications

Inphi:
**Description**

Updates the Dispute Game spec interfaces and adds new specs for the FDG.

**Metadata**

- Fixes https://github.com/ethereum-optimism/client-pod/issues/27


specs/dispute-game-interface.md
```diff
@@ -146,41 +189,52 @@ interface IDisputeGame {
...
+    ///         game UUID's preimage provided in the cwia payload. The preimage of the UUID is
+    ///         constructed as `keccak256(gameType . rootClaim . extraData)` where `.` denotes
+    ///         concatenation.
+    /// @return gameType_ The type of proof system being used.
+    /// @return rootClaim_ The root claim of the DisputeGame.
+    /// @return extraData_ Any extra data supplied to the dispute game contract by the creator.
+    function gameData() external view returns (GameType gameType_, Claim rootClaim_, bytes memory extraData_);
 }
 
 ////////////////////////////////////////////////////////////////
```
trianglesphere: QQ: Why is the output attestation dispute game part of this interface by the fault dispute game isn't?

Inphi: The Fault Dispute Game is part of the interface, we just don't document it here in the spec. The interface is [quite larger](https://github.com/ethereum-optimism/optimism/blob/55dc82b802865996ef83db1fa5802bd9e45e542c/packages/contracts-bedrock/src/dispute/interfaces/IFaultDisputeGame.sol) and I feel adding it in here detracts users from understanding the FDG.

ajsutton: Given we're not doing the attestation dispute game anymore should we remove it from the specs?

Inphi: Yeah, it should be removed. I'll do so in this pr

Inphi:
I'm changing my mind on adding a section on honest behavior. It's like telling readers how to behave in the FDG, rather than reasoning about this from the game rules What do y'all think?

ajsutton:
> I'm changing my mind on adding a section on honest behavior. It's like telling readers how to behave in the FDG, rather than reasoning about this from the game rules What do y'all think?

I think the argument for having it is that the key security property for the game is that so long as there's at least one honest player, the game will always resolve correctly.  If we don't define what an honest player is then it's harder to verify that.

It's like the honest validator spec for the beacon chain - you _can_ reason it all out from the state transition spec, but it's way easier to understand when it's written out explicitly.  And given we took a few goes to get the right algorithm it clearly isn't obvious from the spec.

Inphi:
I guess I'm trying to be careful to not create the impression that all self-interested, but honest actors must behave a certain way. Especially if there's some strategy we haven't considered that results in a correct resolution.

But the validator spec is a good analogy. I'll create a separate doc for this section so it's a bit clearer that it's not the Way The Game Must Be Played.

refcell (approved):

PR #7155, opened 2023-09-08
op-challenger: Updated for the revised resolution

Inphi:
Follow up to #7061 to update op-challenger behavior.

A couple notes:
- The resolver actor is built into the game player actor. This will be useful later when the op-challenger keeps track of its bonded claims. Such that it can avoid resolving subgames it did not participate in. In any case, the op-e2e contains helper functions for claim resolution if we decide to separate resolver actions from the challenger.
- Multiple resolve actors may create the issue highlighted [here](https://github.com/ethereum-optimism/client-pod/issues/61). This will be fixed in a later PR.

### DIFF
--- a/op-challenger/game/fault/agent.go
+++ b/op-challenger/game/fault/agent.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/ethereum-optimism/optimism/op-challenger/game/fault/solver"
 	"github.com/ethereum-optimism/optimism/op-challenger/game/fault/types"
@@ -18,6 +19,8 @@ import (
 type Responder interface {
 	CallResolve(ctx context.Context) (gameTypes.GameStatus, error)
 	Resolve(ctx context.Context) error
+	CallResolveClaim(ctx context.Context, claimIdx int64) error
+	ResolveClaim(ctx context.Context, claimIdx int64) error
 	PerformAction(ctx context.Context, action types.Action) error
 }
 
@@ -112,6 +115,10 @@ func (a *Agent) shouldResolve(status gameTypes.GameStatus) bool {
 // tryResolve resolves the game if it is in a winning state
 // Returns true if the game is resolvable (regardless of whether it was actually resolved)
 func (a *Agent) tryResolve(ctx context.Context) bool {
+	if err := a.resolveClaims(ctx); err != nil {
+		a.log.Error("Failed to resolve claims", "err", err)
+		return false
+	}
 	status, err := a.responder.CallResolve(ctx)
 	if err != nil || status == gameTypes.GameStatusInProgress {
 		return false
@@ -124,6 +131,41 @@ func (a *Agent) tryResolve(ctx context.Context) bool {
 		a.log.Error("Failed to resolve the game", "err", err)
 	}
 	return true
+}
+
+func (a *Agent) resolveClaims(ctx context.Context) error {
+	claims, err := a.loader.FetchClaims(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to fetch claims: %w", err)
+	}
+	if len(claims) == 0 {
+		return nil
+	}
+
+	var resolvableClaims []int64
+	for _, claim := range claims {
+		a.log.Debug("checking if claim is resolvable", "claimIdx", claim.ContractIndex)
+		if err := a.responder.CallResolveClaim(ctx, int64(claim.ContractIndex)); err == nil {
+			a.log.Info("Resolving claim", "claimIdx", claim.ContractIndex)
+			resolvableClaims = append(resolvableClaims, int64(claim.ContractIndex))
+		}
+	}
+	a.log.Info("Resolving claims", "numClaims", len(resolvableClaims))
+
+	var wg sync.WaitGroup
+	wg.Add(len(resolvableClaims))
+	for _, claimIdx := range resolvableClaims {
+		claimIdx := claimIdx
+		go func() {
+			defer wg.Done()
+			err := a.responder.ResolveClaim(ctx, claimIdx)
+			if err != nil {
+				a.log.Error("Failed to resolve claim", "err", err)
+			}
+		}()
+	}
+	wg.Wait()
+	return nil
 }
 
 // newGameFromContracts initializes a new game state from the state in the contract
--- a/op-challenger/game/fault/agent.go
+++ b/op-challenger/game/fault/agent.go
@@ -177,12 +177,12 @@ func (a *Agent) resolveClaims(ctx context.Context) error {
 	for {
 		err := a.tryResolveClaims(ctx)
 		switch err {
-		case err:
-			return err
 		case errNoResolvableClaims:
 			return nil
-		default:
+		case nil:
 			continue
+		default:
+			return err
 		}
 	}
 }
--- a/op-challenger/game/fault/agent.go
+++ b/op-challenger/game/fault/agent.go
@@ -141,7 +141,7 @@ func (a *Agent) tryResolveClaims(ctx context.Context) error {
 		return fmt.Errorf("failed to fetch claims: %w", err)
 	}
 	if len(claims) == 0 {
-		return nil
+		return errNoResolvableClaims
 	}
 
 	var resolvableClaims []int64
--- a/op-challenger/game/fault/agent.go
+++ b/op-challenger/game/fault/agent.go
@@ -19,8 +19,8 @@ import (
 type Responder interface {
 	CallResolve(ctx context.Context) (gameTypes.GameStatus, error)
 	Resolve(ctx context.Context) error
-	CallResolveClaim(ctx context.Context, claimIdx int64) error
-	ResolveClaim(ctx context.Context, claimIdx int64) error
+	CallResolveClaim(ctx context.Context, claimIdx uint64) error
+	ResolveClaim(ctx context.Context, claimIdx uint64) error
 	PerformAction(ctx context.Context, action types.Action) error
 }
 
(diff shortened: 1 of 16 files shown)


op-challenger/game/fault/agent.go
```diff
@@ -126,6 +133,41 @@ func (a *Agent) tryResolve(ctx context.Context) bool {
...
+	for _, claim := range claims {
+		a.log.Debug("checking if claim is resolvable", "claimIdx", claim.ContractIndex)
+		if err := a.responder.CallResolveClaim(ctx, int64(claim.ContractIndex)); err == nil {
+			a.log.Info("Resolving claim", "claimIdx", claim.ContractIndex)
+			resolvableClaims = append(resolvableClaims, int64(claim.ContractIndex))
+		}
+	}
+	a.log.Info("Resolving claims", "numClaims", len(resolvableClaims))
+
+	var wg sync.WaitGroup
```
Inphi, line 160: Resolving claims is done asynchronously to speed up resolution. Otherwise, in large games containing N claims, it'll take at least N blocks for the entire game to be resolved.
Instead, this lets us resolve the game in roughly M blocks, where M is the max game depth.

refcell: We should be able to batch these actions in a single multicall3 call if we order them correctly

refcell: But that improvement can be ticketed and done later

op-challenger/game/fault/agent.go
```diff
@@ -18,6 +19,8 @@ import (
 type Responder interface {
 	CallResolve(ctx context.Context) (gameTypes.GameStatus, error)
 	Resolve(ctx context.Context) error
+	CallResolveClaim(ctx context.Context, claimIdx int64) error
+	ResolveClaim(ctx context.Context, claimIdx int64) error
```
refcell: Indices should be unsigned `u256` numbers onchain, can we make these `uint64` values?

op-challenger/game/fault/agent.go
```diff
@@ -112,6 +115,10 @@ func (a *Agent) shouldResolve(status gameTypes.GameStatus) bool {
 // tryResolve resolves the game if it is in a winning state
 // Returns true if the game is resolvable (regardless of whether it was actually resolved)
 func (a *Agent) tryResolve(ctx context.Context) bool {
+	if err := a.resolveClaims(ctx); err != nil {
+		a.log.Error("Failed to resolve claims", "err", err)
+		return false
+	}
```
refcell, line 121: I don't believe this function is working as intended.

It fetches all claims, and performs `resolveClaim` for all calls that will succeed.

But the idea is to resolve all the claims so that `tryResolve` may proceed to call the final `resolve()` function right?

Shouldn't `resolveClaim` keep looping over it's current action until there are no `resolveClaim` calls that would succeed (as part of the naive heuristic)?

Currently any `resolveClaim` calls where there are still subgames will fail so each time `tryResolve` is called it will essentially only perform the lowest game depth claim resolves and then fail at the next step trying to call `resolve()` below this section.

op-challenger/game/fault/types/game.go
```diff
@@ -23,40 +23,50 @@ type Game interface {
...
 }
 
 // gameState is a struct that represents the state of a dispute game.
 // The game state implements the [Game] interface.
 type gameState struct {
 	agreeWithProposedOutput bool
-	root                    ClaimData
-	claims                  map[ClaimData]*extendedClaim
+	root                    claimEntry
+	claims                  map[claimEntry]*extendedClaim
```
refcell, line 54: Can we use a reference `*claimEntry` as a key so we can avoid copying data?

op-challenger/game/fault/agent.go
Inphi: Yeah this can be improved. Tho we probably shouldn't use multicall3 due to https://github.com/ethereum-optimism/client-pod/issues/61. Given multiple players, it's likely that a single call fails - which short-circuits the entire transaction.

op-challenger/game/fault/types/game.go
```diff
@@ -124,17 +134,34 @@ func (g *gameState) MaxDepth() uint64 {
...
+	}
+	return nil
+}
+
+func makeClaimEntry(claim Claim) claimEntry {
+	return claimEntry{
+		ClaimData:           claim.ClaimData,
+		ParentContractIndex: claim.ParentContractIndex,
 	}
 }
```
refcell, line 167: Would prefer to place this onto the `Claim` itself and call it a `ClaimKey` since it represents the claim data that's used as a key.
```golang
func (c *Claim) Key() ClaimKey {
    return ClaimKey{
        ClaimData: c.ClaimData,
        ParentContractIndex: c.ParentContractIndex,
    }
}
```

op-challenger/game/fault/types/game.go
```diff
@@ -124,17 +134,34 @@ func (g *gameState) MaxDepth() uint64 {
...
+		return nil
+	}
+	// TODO(inphi): refactor gameState for faster parent lookups
+	for _, c := range g.claims {
+		if c.self.ContractIndex == claim.ParentContractIndex {
+			return c
+		}
+	}
+	return nil
+}
```
refcell, line 160: If we use `ClaimEntry` (or `ClaimKey` as suggested below), the `Claim` `Parent` field should be changed from `ClaimData` to this `ClaimEntry`/`ClaimKey` so we can then just continue to do the following:
```golang
func (g *gameState) GetParent(claim Claim) (Claim, error) {
	if claim.IsRoot() {
		return Claim{}, ErrClaimNotFound
	}
	if parent, ok := g.claims[claim.Parent]; !ok {
		return Claim{}, ErrClaimNotFound
	} else {
		return parent.self, nil
        }
}
```

op-e2e/e2eutils/challenger/helper.go
```diff
@@ -68,6 +68,12 @@ func WithAlphabet(alphabet string) Option {
 	}
 }
 
+func WithPollInterval(pollInterval time.Duration) Option {
+	return func(c *config.Config) {
+		c.PollInterval = pollInterval
+	}
+}
```
refcell, line 75: Great addition, thank you

op-e2e/e2eutils/disputegame/game_helper.go
```diff
@@ -239,9 +292,27 @@ func (g *FaultGameHelper) ChallengeRootClaim(ctx context.Context, performMove Mo
...
+		newClaimLen = actual.Int64()
+		return actual.Cmp(big.NewInt(checkPoint)) > 0, nil
+	})
+	return newClaimLen, err
+}
+
 func (g *FaultGameHelper) Attack(ctx context.Context, claimIdx int64, claim common.Hash) {
 	tx, err := g.game.Attack(g.opts, big.NewInt(claimIdx), claim)
+	g.LogGameData(ctx)
+	g.t.Logf("Attack. claim: %v. claimIdx: %d", claim, claimIdx)
```
refcell: Think we can remove these
```suggestion
```

op-e2e/e2eutils/disputegame/game_helper.go
```diff
@@ -239,9 +292,27 @@ func (g *FaultGameHelper) ChallengeRootClaim(ctx context.Context, performMove Mo
...
 	g.require.NoError(err, "Attack transaction did not send")
+	g.t.Logf("Attack tx: %x. claim: %v. claimIdx: %d", tx.Hash(), claim, claimIdx)
```
refcell: ibid
```suggestion
```

op-challenger/game/fault/agent.go
Inphi: Good point. It only technically works because the game actor periodically checks if the game is resolvable. So `resolveClaims` also gets called periodically.
But it'll be better to do as you've described here rather than rely on the game scheduler to wake up the actor. I'll make the changes.

op-challenger/game/fault/types/game.go
Inphi: Can't use pointers as map keys as the address will be different each time we create the key.

op-challenger/game/fault/types/game.go
Inphi: I think it's better to keep those concerns separated. So the `Claim` isn't concerned with how it's referenced in the game state.
But I like the idea of a typed `ClaimKey`. I can add that to game.go.

Inphi: ahh nvmind. The `ClaimKey` would have to be a `Claim` anyways which defeats the purpose of keying using only those two fields.

op-challenger/game/fault/types/game.go
Inphi: `Claim` doesn't contain its parent `Claim`, but only contains a reference index to it. To do this you'd need a self-reference to the `Claim` type for its parent. Which complicates contracts usage, as it now requires two calls to properly construct a `types.Claim`.

Inphi: Also generally when building graph structures, you wanna avoid using direct references. As that complicates lifetimes and ownership, even in GC languages.

The refactor I had in mind is to store the entire solidity claims array in the game state. Then simply use indices to reference them in the graph. This lets us lookup parent claims in constant time. I can iterate on that in another PR.

refcell: I agree that makes the most sense. The original design of this game structure was not my decision.

op-challenger/game/fault/types/game.go
refcell: Good point, we could use a hash though and implement the hashing function on the object pointer type.

refcell (approved):
This is at a solid state now where we can iterate on these changes once merged.

*Once ci passes 😅

op-challenger/game/fault/types/game.go
Inphi: ehhh, `claimEntry` is already a Hash and three words. Not too terrible storage wise. But I don't mind changing this tho.

Inphi:
@refcell  CI is finally passing. Gonna merge this in! Thanks for the reviews.

refcell:
> @refcell  CI is finally passing. Gonna merge this in! Thanks for the reviews.

Sweet, let's ship it! Nice work!

PR #7164, opened 2023-09-10
contracts-bedrock: use create2 to deploy implementations

tynes:
**Description**

Use create2 to deploy L1 contract implementations. This is useful to deduplicate config, making the contracts have canoncial addresses per version across networks. Every L1 contract implementation should use the same create2 salt over time.


clabby (approved):
LGTM, nice 😄

trianglesphere (approved):

PR #7165, opened 2023-09-10
contracts-bedrock: migrate to ISemver

tynes:
**Description**

Migrate away from using `Semver` where the semantic version is immutable to using `ISemver` where the semantic version is a constant. Only update the L1 contracts, this change should be followed up with additional PRs that update the other contracts. This change simplifies the code and reduces the bytecode size of each of the contracts.


maurelian (approved):
Nice. This feels cleaner and easier to maintain.

tynes:
Just realized this should be minor version bumps and not patch, will update

PR #7176, opened 2023-09-11
deps: remove dead js deps

tynes:
**Description**

The following deps are no longer used in the monorepo:

- lerna
- copyfiles
- mkdirp

This was confirmed using `git grep -rin` for the package name in the monorepo. Deleting these from the root `package.json` will reduce some noise in the monorepo by preventing dependabot update PRs as well as reducing the dependency tree and install time.


refcell:
leaving this to the king of typescript for review @roninjin10

roninjin10 (approved):

PR #7178, opened 2023-09-11
deps: use node lts

tynes:
**Description**

Upgrade from node.js v16 to v18. v16 starts end of life on 9/11/2023, meaning it will no longer get security updates. v18 is currently LTS, https://github.com/nodejs/release#release-schedule for additional information.

We previously were not able to upgrade node.js versions because we were using an older version of hardhat that did not support newer versions of node.js but that is no longer the case.


refcell (approved):

.nvmrc
```diff
@@ -1 +1 @@
-v16.16.0
+v18.17.1
```
roninjin10, line 1:
```suggestion
v18.17.1
```
It might be worth seeing if node 20 works. Node 18 has been relatively unstable with memory leaks, bugs, and regressions in minor releases

ops/docker/Dockerfile.packages
```diff
@@ -35,7 +35,7 @@ FROM us-docker.pkg.dev/oplabs-tools-artifacts/images/ci-builder:latest as foundr
 # we use it rather than alpien because it's not much
```
roninjin10:
```suggestion
# we use it rather than alpine because it's not much
```
This mispelling was me

roninjin10 (approved):

.nvmrc
tynes: I have a preference for LTS in production but if you feel like v20 is preferable then I'm open to it

roninjin10: It's pretty close. Might be best to use LTS and then consider upgrading if we happen to run into problems.

trianglesphere (approved):

PR #7183, opened 2023-09-11
feat(op-challenger): Last Step Disk Cache Store

refcell:
**Description**

Stores the last step on disk to avoid re-generating a proof via the executor
when the cannon trace provider has access to this data.

**Metadata**

Fixes https://github.com/ethereum-optimism/client-pod/issues/17

### DIFF
--- a/op-challenger/game/fault/trace/cannon/provider.go
+++ b/op-challenger/game/fault/trace/cannon/provider.go
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	proofsDir = "proofs"
+	proofsDir      = "proofs"
+	diskStateCache = "state.json.gz"
 )
 
 type proofData struct {
@@ -34,6 +35,9 @@ type proofData struct {
 	OracleOffset uint32        `json:"oracle-offset,omitempty"`
 }
 
+type LastStepWriter func(dir string, step uint64) error
+type LastStepReader func(dir string) (uint64, error)
+
 type CannonMetricer interface {
 	RecordCannonExecutionTime(t float64)
 }
@@ -44,10 +48,12 @@ type ProofGenerator interface {
 }
 
 type CannonTraceProvider struct {
-	logger    log.Logger
-	dir       string
-	prestate  string
-	generator ProofGenerator
+	logger         log.Logger
+	dir            string
+	prestate       string
+	generator      ProofGenerator
+	lastStepWriter LastStepWriter
+	lastStepReader LastStepReader
 
 	// lastStep stores the last step in the actual trace if known. 0 indicates unknown.
 	// Cached as an optimisation to avoid repeatedly attempting to execute beyond the end of the trace.
@@ -75,10 +81,12 @@ func NewTraceProvider(ctx context.Context, logger log.Logger, m CannonMetricer, 
 
 func NewTraceProviderFromInputs(logger log.Logger, m CannonMetricer, cfg *config.Config, localInputs LocalGameInputs, dir string) *CannonTraceProvider {
 	return &CannonTraceProvider{
-		logger:    logger,
-		dir:       dir,
-		prestate:  cfg.CannonAbsolutePreState,
-		generator: NewExecutor(logger, m, cfg, localInputs),
+		logger:         logger,
+		dir:            dir,
+		prestate:       cfg.CannonAbsolutePreState,
+		generator:      NewExecutor(logger, m, cfg, localInputs),
+		lastStepWriter: WriteLastStep,
+		lastStepReader: LoadLastStep,
 	}
 }
 
@@ -142,6 +150,16 @@ func (p *CannonTraceProvider) loadProof(ctx context.Context, i uint64) (*proofDa
 		// If the requested index is after the last step in the actual trace, extend the final no-op step
 		return p.lastProof, nil
 	}
+	// Attempt to read the last step from disk cache
+	if p.lastProof == nil && p.lastStep == 0 {
+		step, err := p.lastStepReader(p.dir)
+		if err != nil {
+			p.logger.Warn("Failed to read last step from disk cache", "err", err)
+		} else {
+			p.lastStep = step
+			i = step
+		}
+	}
 	path := filepath.Join(p.dir, proofsDir, fmt.Sprintf("%d.json.gz", i))
 	file, err := ioutil.OpenDecompressed(path)
 	if errors.Is(err, os.ErrNotExist) {
@@ -168,6 +186,9 @@ func (p *CannonTraceProvider) loadProof(ctx context.Context, i uint64) (*proofDa
 				if err != nil {
 					return nil, fmt.Errorf("cannot hash witness: %w", err)
 				}
+				if err := p.lastStepWriter(p.dir, p.lastStep); err != nil {
+					p.logger.Warn("Failed to write last step to disk cache", "step", p.lastStep)
+				}
 				proof := &proofData{
 					ClaimValue:   witnessHash,
 					StateData:    hexutil.Bytes(witness),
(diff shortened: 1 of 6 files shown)


refcell:
* develop
  * **PR #7183** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7183" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7183?utm_source=stack-comment).

op-service/ioutil/gzip.go
```diff
@@ -38,6 +41,27 @@ func OpenCompressed(file string, flag int, perm os.FileMode) (io.WriteCloser, er
...
+// if the filename ends with .gz.
+// If no flags are provided, [os.O_CREATE|os.O_WRONLY] is used.
+func WriteCompressedJson(file string, obj any, flags ...int) error {
+	if !IsGzip(file) {
+		return fmt.Errorf("file %v does not have .gz extension", file)
+	}
+	if len(flags) == 0 {
+		flags = []int{os.O_CREATE | os.O_WRONLY}
+	}
+	flag := lo.Reduce(flags, func(agg int, item int, _ int) int {
```
ajsutton: This would actually be simpler and easier to read as a for loop and we could avoid introducing the new `lo` dependency.

op-challenger/game/fault/trace/cannon/provider_test.go
```diff
@@ -215,14 +246,37 @@ func setupTestData(t *testing.T) (string, string) {
...
+		logger:         testlog.Logger(t, log.LvlInfo),
+		dir:            dataDir,
+		generator:      generator,
+		prestate:       filepath.Join(dataDir, prestate),
+		lastStepReader: stepper.loadLastStep,
+		lastStepWriter: stepper.storeLastStep,
+	}, generator, stepper
+}
+
+type stubStepper struct {
```
ajsutton: I don't think I'd use the stub stepper here.  The other tests in this file read and write to real disk already because the majority of the code in the provider is disk access.  I think the production code would be slightly simpler without supporting a stub and then we'd actually test the read/write code for the final state as well which is currently missed.

The test would basically be to create a trace provider and request data after the end of the trace - it should then write the final state file.  Then create a new trace provider from the same dir and it should read that state file without having to re-generate.

op-challenger/game/fault/trace/cannon/provider.go
```diff
@@ -44,10 +48,12 @@ type ProofGenerator interface {
...
-	logger    log.Logger
-	dir       string
-	prestate  string
-	generator ProofGenerator
+	logger         log.Logger
+	dir            string
+	prestate       string
+	generator      ProofGenerator
+	lastStepWriter LastStepWriter
+	lastStepReader LastStepReader
```
ajsutton: We don't need these to be stored as function references anymore - the production code can just directly call the methods since the tests don't need to stub them anymore.

op-challenger/game/fault/trace/cannon/provider_test.go
```diff
@@ -129,6 +129,49 @@ func TestGetStepData(t *testing.T) {
...
+			ClaimValue:   common.Hash{0xaa},
+			StateData:    []byte{0xbb},
+			ProofData:    []byte{0xcc},
+			OracleKey:    common.Hash{0xdd}.Bytes(),
+			OracleValue:  []byte{0xdd},
+			OracleOffset: 10,
+		}
+		_, _, _, err := provider.GetStepData(context.Background(), 7000)
+		require.NoError(t, err)
+		require.Contains(t, initGenerator.generated, 10, "should have tried to generate the proof")
```
ajsutton: This fails for me locally - it actually generates at step 7000 since it doesn't know what the final step is yet.  I'm not really sure how this isn't failing in CI...

ajsutton: Oh, it fails when running just this test case individually but passes otherwise.  We shouldn't be reusing the call to `dataDir, prestate := setupTestData(t)` across all these subtests.

refcell: It doesn't have the `test_data` in ci, nor do I have it locally which is why it was passing

op-service/ioutil/gzip.go
```diff
@@ -38,6 +39,28 @@ func OpenCompressed(file string, flag int, perm os.FileMode) (io.WriteCloser, er
 	return out, nil
 }
 
+// WriteCompressedJson writes the object to the specified file as a compressed json object
+// if the filename ends with .gz.
+// If no flags are provided, [os.O_CREATE|os.O_WRONLY] is used.
+func WriteCompressedJson(file string, obj any, flags ...int) error {
```
trianglesphere: Do we ever override these flags? If not, I'd remove the option to specify them.

op-challenger/game/fault/trace/cannon/provider.go
```diff
@@ -168,6 +184,9 @@ func (p *CannonTraceProvider) loadProof(ctx context.Context, i uint64) (*proofDa
 				if err != nil {
 					return nil, fmt.Errorf("cannot hash witness: %w", err)
 				}
+				if err := WriteLastStep(p.dir, p.lastStep+1); err != nil {
```
trianglesphere: Could also use `state.Step` - I had to dig a little to figure out why 1 was being added to `p.lastStep`

op-challenger/game/fault/trace/cannon/provider.go
```diff
@@ -142,6 +146,18 @@ func (p *CannonTraceProvider) loadProof(ctx context.Context, i uint64) (*proofDa
...
 	}
+	// Attempt to read the last step from disk cache
+	if p.lastProof == nil && p.lastStep == 0 {
+		step, err := ReadLastStep(p.dir)
+		if err != nil {
+			p.logger.Warn("Failed to read last step from disk cache", "err", err)
+		} else {
+			p.lastStep = step
+			if i > p.lastStep {
+				i = step
```
trianglesphere, line 156: Why do we lower `i` here?

op-challenger/game/fault/trace/cannon/provider.go
```diff
@@ -194,3 +213,26 @@ func (p *CannonTraceProvider) loadProof(ctx context.Context, i uint64) (*proofDa
 	}
 	return &proof, nil
 }
+
+type diskStateCacheObj struct {
+	Step uint64 `json:"step"`
+}
+
+func ReadLastStep(dir string) (uint64, error) {
```
trianglesphere, line 221: Would be nice to have some comments on what these functions do / how the last step value relates to what snapshots exists.

op-challenger/game/fault/trace/cannon/provider.go
refcell: To avoid having to generate the proof again since the last step's proof should be present in this case.

ajsutton: Specifically, the post state of the final step in the actual execution is extended to all remaining trace indices in the game.  So if we request a step after the end of the trace, we actually need to generate the proof at the final step cannon will execute and use that instead.

op-challenger/game/fault/trace/cannon/provider.go
```diff
@@ -34,6 +35,9 @@ type proofData struct {
 	OracleOffset uint32        `json:"oracle-offset,omitempty"`
 }
 
+type LastStepWriter func(dir string, step uint64) error
+type LastStepReader func(dir string) (uint64, error)
```
ajsutton: We should be able to remove these types now.

ajsutton (approved):

PR #7184, opened 2023-09-11
specs: Honest Challenger specs

Inphi:
Specifies the behavior of the Honest Challenger.

fixes https://github.com/ethereum-optimism/client-pod/issues/45


specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+An honest challenger strives to ensure a correct, truthful, game resolution.
+The honest challenger is also _rational_ as any deviation from its behavior will result in
+negative outcomes.
+This document specifies the expected behavior of an honest challenger.
+
+## Overview
+
+The Honest Challenger has two primary duties:
+
+1. Challenge invalid outputs by either creating new FDGs or supporting existing ones that
```
ajsutton: Not sure if we need to note that in the alpha op-challenger won't create new FDGs, and in the final design for fault proofs a new FDG is created to propose an output root so that won't be a responsibility of the honest challenger.

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+against these claims (valid defined by the response in [Counter Claims](#counter-claims)),
+the only option for an honest challenger is to execute a VM step on-chain to disprove the claim
+at `MAX_GAME_DEPTH`.
+If the VM step proves this claim correct, the claim will be left uncountered.
+
+The same rules for determining whether a move is an attack, defense, or noop from the above
+ section apply to claims at the bottom level of the tree.
+ Instead of calling `attack` or `defend`, the challenger issues a [step](./fault-dispute-game.md#step).
+ An honest challenger will issue an attack step if it disagrees with the claim,
+ otherwise a defense step is issued.
```
ajsutton: Might be worth mentioning the need to populate the pre-image oracle with any required data prior to calling step.

ajsutton (commented):
Looks good generally. Just a couple of comments.

specs/honest-challenger-fdg.md
refcell: Agree with this. I think the honest challenger has a primary, overarching goal to engage in fault dispute games, challenging invalid root claims and defending valid ones (or some variant of this wording). Keep it simple where possible

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+## Overview
+
+The Honest Challenger has two primary duties:
+
+1. Challenge invalid outputs by either creating new FDGs or supporting existing ones that
+aim to delete such outputs.
+2. Contest root claims in FDGs that seek to remove valid outputs.
+
+The honest challenger monitors the `L2OutputOracle` to identify invalid proposals and scans
+the `DisputeGameFactory` contract to find on-going FDGs.
```
refcell: If the former is corrected this should change to something like the following
```suggestion
The honest challenger polls the `DisputeGameFactory` contract for new and on-going Fault Dispute Games.
```

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+1. If the trusted node agrees with the output, take no action. A `FaultDisputeGame` is
+designed to prove that a proposed output root is incorrect and consequently, to delete it.
+Therefore, an honest challenger will not create a dispute game that challenges an output
+root that its trusted node agrees with.
+2. If the trusted node disagrees, create a new `FaultDisputeGame` via the `DisputeGameFactory`.
+In contrast to the above, an honest challenger aims to delete any output roots that its trusted
+node disagrees with in order to claim the bond attached to it.
+The honest challenger assumes that their rollup node is synced to the canonical state and that
+the fault proof program is correct, so it is willing to put its money on the line to counter
+any faults.
```
refcell: This can probably all be removed if the game creation is to happen in another module or called something different than the "honest challenger".

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+1. [**Attack**](./fault-dispute-game.md#attack) if they disagree with the root claim.
+When an honest challenger disagrees with a root claim of a game,
+it is akin to them agreeing with the output proposal that the game is attempting to delete.
+The root claim commits to the entire execution trace, so the first move by a defender of the
+output root is to attack with the [ClaimHash](./fault-dispute-game.md#claims) at the midpoint
+instruction within their execution trace.
+2. **Do Nothing** if they agree with the root claim. If an honest challenger agrees with a
+root claim of a game, it means that they disagree with the output root it is trying to delete.
+They do nothing because if the root claim is left un-countered, the game will delete the output
+root they disagree with.
```
refcell: I would just add here something to the effect of:
```suggestion
root they disagree with. NOTE: The honest challenger will still track this game in order to defend any subsequent claims made against the root claim - in effect, "playing the game".
```

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+When a new claim is made in a dispute game with a [game tree](./fault-dispute-game.md#game-tree)
+depth in the range of `[1, MAX_DEPTH]`, the honest challenger processes it and performs
+a response. If multiple claims are observed since the last sync,
+the honest challenger responds to them in chronological order.
+
+To determine the appropriate response, the challenger first needs to know which
+[_team_](./fault-dispute-game.md#team-dynamics) it belongs to.
+This determines the set of claims it should respond to in the FDG.
+If the agent determines itself to be a Defender, aiming to delete an output root,
+ then it must dispute claims positioned at odd depths in the game tree.
```
refcell: nit
```suggestion
then it must dispute claims positioned at odd depths in the game tree.
```

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+    if depth(claim.position) % 2 != chal.value:
+        if chal_trace[trace_index(claim.position)] == claim.claim_hash:
+            defend()
+        else:
+            attack()
+    else: pass # no response
+```
+
+In attack or defense, the honest challenger MUST submit a `ClaimHash` corresponding to the state
+identified by the trace index of their response position.
```
refcell: I think we don't need to stress this here since we state initially that the doc details the expected behavior of the challenger. So we can just write what the honest challenger (a concept) does.
```suggestion
In attack or defense, the honest challenger submits a `ClaimHash` corresponding to the state
identified by the trace index of their response position.
```

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+
+The honest challenger SHOULD respond to claims as soon as possible to avoid the clock of its
+counter-claim from expiring.
```
refcell: ibid
```suggestion
The honest challenger responds to claims as soon as possible to avoid the clock of its
counter-claim from expiring.
```

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Steps
+
+At the max depth of the game, claims represent commitments to the state of the fault proof VM
+at a single instruction step interval.
+Because the game can no longer bisect further, when the honest challenger has a valid move
+against these claims (valid defined by the response in [Counter Claims](#counter-claims)),
+the only option for an honest challenger is to execute a VM step on-chain to disprove the claim
+at `MAX_GAME_DEPTH`.
```
refcell: nit
```suggestion
the only option is to execute a VM step on-chain to disprove the claim at `MAX_GAME_DEPTH`.
```

specs/honest-challenger-fdg.md
refcell: nit
```suggestion
section apply to claims at the bottom level of the tree.
Instead of calling `attack` or `defend`, the challenger issues a [step](./fault-dispute-game.md#step).
An honest challenger will issue an attack step if it disagrees with the claim,
otherwise a defense step is issued.
```

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+If the VM step proves this claim correct, the claim will be left uncountered.
```
refcell: I don't think it's very intuitive what this means. What do you think about explicitly write what leaving a claim uncountered means with respect to the state of the fault proof VM since that's how the section started?
For example
> Since the claim commits to a single instruction, it's correct execution and leaving it uncountered means that this bisection failed to disprove the root claim.

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+
+The same rules for determining whether a move is an attack, defense, or noop from the above
```
refcell: I think this section has a fair amount of repetition to the one above - it might make sense to combine them so as to articulate from the outset that calling `attack` or `defend` won't work and the challenger should issue the step (with the hyperlink)

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+ section apply to claims at the bottom level of the tree.
+ Instead of calling `attack` or `defend`, the challenger issues a [step](./fault-dispute-game.md#step).
+ An honest challenger will issue an attack step if it disagrees with the claim,
+ otherwise a defense step is issued.
+
+## Resolution
+
+When one side of a `FaultDisputeGame`'s chess clock runs out, the honest challenger’s responsibility
+is to resolve the game.
+This action entails the challenger calling the `resolve` function on the `FaultDisputeGame` contract.
```
refcell: nit
```suggestion
When the [chess clock](./fault-dispute-game.md#game-clock) of a `FaultDisputeGame` team runs out, the game can be resolved.
The honest challenger does this by calling the `resolve` function on the `FaultDisputeGame` contract.
```

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,145 @@
...
+
+The `FaultDisputeGame` does not put a time cap on resolution - because of the liveness assumption
+on honest challengers and the bonds attached to the claims they’ve countered,
+challengers should resolve the game promptly in order to make their funds
+liquid again and capture their reward(s).
```
refcell: I think we can keep it simple here and just say something like
```suggestion
challengers are economically incentivized to resolve the game promptly to capture the bonds.
```

refcell (changes requested):

specs/honest-challenger-fdg.md
Inphi: Makes sense. I'll rewrite this.

specs/honest-challenger-fdg.md
Inphi: Will reword. I wanted to highlight that determining a challenger's  _stance_ (attack or defense, but not the moves) works the same way for steps and moves. We don't have existing verbiage for this kind of thing. Since we overload `attack` and `defense` in multiple contexts.

specs/honest-challenger-fdg.md
Inphi: This was missing in the original fdg specs. Added it there in its step section.

specs/honest-challenger-fdg.md
Inphi: The meaning of uncountered (or rather `countered` since uncountered isn't a word) is explained in the [FDG specs](https://github.com/ethereum-optimism/optimism/blob/develop/specs/fault-dispute-game.md#step-types). Though I think this sentence is redundant given the context so I've removed it.

refcell (approved):

specs/honest-challenger-fdg.md
refcell: Ah understood, that makes sense to me!

specs/honest-challenger-fdg.md
```diff
@@ -0,0 +1,122 @@
...
+claim is left un-countered, the game resolves to their agreement.
+NOTE: The honest challenger will still track this game in order to defend any subsequent
+claims made against the root claim - in effect, "playing the game".
+
+### Counter Claims
+
+When a new claim is made in a dispute game with a [game tree](./fault-dispute-game.md#game-tree)
+depth in the range of `[1, MAX_DEPTH]`, the honest challenger processes it and performs
+a response. If multiple claims are observed since the last sync,
+the honest challenger responds to them in chronological order.
```
trianglesphere: chronological order appears to be an over specification (unless it is somehow critical to resolution working correctly)

trianglesphere (commented):
This is a good overview of what the challenger does. Probably out of scope for this, but I feel like we are missing the argument for why these movements (combined with resolution) will make sure that the correct challenger always wins.

Inphi:
> This is a good overview of what the challenger does. Probably out of scope for this, but I feel like we are missing the argument for why these movements (combined with resolution) will make sure that the correct challenger always wins.

Yeah this would be out of scope for a spec. This should def be communicated, maybe in a blogpost.

specs/honest-challenger-fdg.md
Inphi: good point. The ordering doesn't matter. Fixed.

ajsutton (approved):

PR #7185, opened 2023-09-11
feat(op-challenger): Trace Provider Position Type Port

refcell:
**Description**

Migrates the trace provider to accept positions.

**Metadata**

Fixes https://github.com/ethereum-optimism/client-pod/issues/42


refcell:
* develop
  * **PR #7185** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7185" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7185?utm_source=stack-comment).

op-challenger/game/fault/solver/solver.go
```diff
@@ -97,22 +97,21 @@ func (s *claimSolver) AttemptStep(ctx context.Context, game types.Game, claim ty
...
+		preState, proofData, oracleData, err = s.trace.GetStepData(ctx, claim.Position)
 		if err != nil {
 			return StepData{}, err
 		}
 	} else {
 		// We agree with the claim so Defend and use this claim as the starting point to execute the step after
 		// Thus we need the pre-state of the next step
-		// Note: This makes our maximum depth 63 because we need to add 1 without overflowing.
-		preState, proofData, oracleData, err = s.trace.GetStepData(ctx, index+1)
+		defendingPosition := types.NewPosition(s.gameDepth, int(claim.TraceIndex(s.gameDepth))+1)
```
ajsutton: I'd push this logic into `Position` then it can just increment the index at depth directly without having to tie it to `TraceIndex`:

```go
func (p Position) StepDefend() Position {
	return Position{
		depth:        p.depth,
		indexAtDepth: new(big.Int).Add(p.IndexAtDepth(), big.NewInt(1)),
	}
}
```

Maybe it should just be called `MoveRight` since step action really only makes sense at the maximum depth and this is agnostic of that.

ajsutton: er, except in this PR indexAtDepth isn't a big.Int yet - I had your other PR still checked out. 😄

op-challenger/game/fault/test/claim_builder.go
```diff
@@ -33,34 +33,38 @@ func (c *ClaimBuilder) CorrectTraceProvider() types.TraceProvider {
 
 // CorrectClaim returns the canonical claim at a specified trace index
 func (c *ClaimBuilder) CorrectClaim(idx uint64) common.Hash {
-	value, err := c.correct.Get(context.Background(), idx)
+	pos := types.NewPosition(c.maxDepth, int(idx))
+	value, err := c.correct.Get(context.Background(), pos)
```
ajsutton: This is actually only used internally to `ClaimBuilder` and ultimately from places that already have a `Position` (except `CreateRootClaim` but it can just use `types.NewPositionFromGIndex(1)` as the root position). So I'd probably just delete it and use `CorrectClaimAtPosition` instead.

op-challenger/game/fault/trace/cannon/provider_test.go
```diff
@@ -121,7 +127,8 @@ func TestGetStepData(t *testing.T) {
 			OracleValue:  []byte{0xdd},
 			OracleOffset: 10,
 		}
-		preimage, proof, data, err := provider.GetStepData(context.Background(), 7000)
+		pos := types.NewPosition(int(provider.gameDepth), 7000)
```
ajsutton: We're repeating this pattern of creating a position for a specific trace index a lot in tests - I wonder if its worth having a helper function for it somewhere.

op-e2e/e2eutils/disputegame/helper.go
```diff
@@ -172,14 +174,26 @@ func (h *FactoryHelper) StartCannonGameWithCorrectRoot(ctx context.Context, roll
 		L2Claim:       challengedOutput.OutputRoot,
 		L2BlockNumber: challengedOutput.L2BlockNumber,
 	}
-	provider := cannon.NewTraceProviderFromInputs(testlog.Logger(h.t, log.LvlInfo).New("role", "CorrectTrace"), metrics.NoopMetrics, cfg, inputs, cfg.Datadir)
-	rootClaim, err := provider.Get(ctx, math.MaxUint64)
+
+	maxDepth := uint64(math.MaxUint64)
```
ajsutton: That definitely doesn't seem right. You should be able to load the max game depth from the cannon game implementation contract even before a game is created:

```suggestion
	cannonTypeAddr, err := h.factory.GameImpls(opts, cannonGameType)
	h.require.NoError(err, "fetch cannon game type impl")
	gameImpl, err := bindings.NewFaultDisputeGameCaller(cannonTypeAddr, h.client)
	h.require.NoError(err, "bind fault dispute game caller")
	maxDepth, err := gameImpl.MAXGAMEDEPTH(opts)
	h.require.NoError(err, "fetch max game depth")
```

op-e2e/e2eutils/disputegame/helper.go
```diff
@@ -172,14 +174,26 @@ func (h *FactoryHelper) StartCannonGameWithCorrectRoot(ctx context.Context, roll
...
+	provider := cannon.NewTraceProviderFromInputs(
+		testlog.Logger(h.t, log.LvlInfo).New("role", "CorrectTrace"),
+		metrics.NoopMetrics,
+		cfg,
+		inputs,
+		cfg.Datadir,
+		maxDepth,
+	)
+	pos := faultTypes.NewPosition(int(maxDepth), int(maxDepth))
```
ajsutton: You actually want `faultTypes.NewPositionFromGIndex(1)` here - it's the root position that we need the claim for. Might be worth defining the root position as a constant actually.

op-e2e/e2eutils/disputegame/honest_helper.go
```diff
@@ -38,7 +38,7 @@ func (h *HonestHelper) Defend(ctx context.Context, claimIdx int64) {
 	pos := types.NewPositionFromGIndex(claim.Position.Uint64())
 	defendPos := pos.Defend()
 	traceIdx := defendPos.TraceIndex(int(h.game.MaxDepth(ctx)))
-	value, err := h.correctTrace.Get(ctx, traceIdx)
+	value, err := h.correctTrace.Get(ctx, defendPos)
 	h.game.require.NoErrorf(err, "Get correct claim at trace index %v", traceIdx)
```
ajsutton: I'd remove the traceIdx reference from these errors and include the position instead. Otherwise you'll get odd error messages when the trace provider is composing two different trace providers into a single game.

op-e2e/e2eutils/disputegame/honest_helper.go
```diff
@@ -48,12 +48,14 @@ func (h *HonestHelper) StepFails(ctx context.Context, claimIdx int64, isAttack b
 	defer cancel()
 	claim := h.game.getClaim(ctx, claimIdx)
 	pos := types.NewPositionFromGIndex(claim.Position.Uint64())
-	traceIdx := pos.TraceIndex(int(h.game.MaxDepth(ctx)))
+	gameDepth := int(h.game.MaxDepth(ctx))
+	traceIdx := pos.TraceIndex(gameDepth)
```
ajsutton: Need to avoid accessing the trace index here as the trace provider might need to translate it for child providers. Should be doing the same thing as the production code and just use a `StepDefend` or `MoveRight` call on Position and pass that through.

ajsutton (commented):
Looks good generally.

op-challenger/game/fault/solver/solver.go
refcell: I don't believe we can just use the same position and move 1 index at that depth right.
Consider the following naive demonstration:

```mermaid
  graph TD;
      1-->2;
      1-->3;
      2-->4;
      2-->5;
      3-->6;
      3-->7;
```

The node at g index `2` would have depth of 1, and index at depth of `0`. Performing a `MoveRight` operation like above, would return the position at gindex `3`, which commits to the trace index at leaf node g index `7`. Whereas, the initial node at g index `2` commits to the trace index at leaf node `5`.

What we actually want is to trace down to the leaf node of the node at g index `2`, so a (max) depth of 2 and index at depth of 1, which is the leaf node with g index `5`. Then we create a new position at g index `6` for the position with the "next" step data. This is the position at max depth with the trace index of the initial node (`2`) + 1.

Thus, our `MoveRight` function should actually be a `NextStepPosition` function that returns the node with g index `6` rather than the node with g index `7`. This is done by using the max game depth and then adding 1 to the trace index of the claim's position. I've updated the implementation in this way.

ajsutton: But you can't ever perform a step against node 2 because it's not a leaf node. If it was, the node you'd want to wind up on would be 3. We don't need to consider moving down the tree here because to perform a step we must already be at the bottom of the tree.  Which is another good reason not to call it `StepDefend` really.

refcell: Ah in this case, it doesn't make a difference since the claim is checked to be a leaf node - I see. Will revert to your suggestion and use a `MoveRight` method using the claim depth 😄

op-challenger/game/fault/trace/cannon/provider_test.go
refcell: Good callout, will do. I think this is a case where we could benefit from being open to using object oriented test suites like the [stretchr testify `suite`](https://github.com/stretchr/testify) since we can move a lot of this redundant unit testing code onto the suite itself as opposed to having one-off setup and builder methods. Happy to explore migrating this to a suite if we're open to it in a follow up pr :)

ajsutton: Not sure if that's useful or not as I'm not familiar with it, but even in Java this would likely just be a helper function somewhere, not a test setup type method.  Either way, starting to use suite is more of a team wide kind of discussion. :)

op-e2e/e2eutils/disputegame/helper.go
refcell: Awesome, great catch.

op-challenger/game/fault/trace/cannon/provider_test.go
refcell: Sure, I'm pretty interested in using suites, so I'll try to write something comprehensive up on this.

op-challenger/game/fault/test/claim_builder.go
```diff
@@ -31,54 +31,50 @@ func (c *ClaimBuilder) CorrectTraceProvider() types.TraceProvider {
...
+		return c.CorrectClaimAtPosition(pos)
 	} else {
-		return c.incorrectClaim(idx)
+		return c.incorrectClaim(pos)
 	}
 }
 
 func (c *ClaimBuilder) CreateRootClaim(correct bool) types.Claim {
-	value := c.claim((1<<c.maxDepth)-1, correct)
+	value := c.claim(types.NewPositionFromGIndex((1<<c.maxDepth)-1), correct)
```
ajsutton:
```suggestion
	value := c.claim(types.NewPositionFromGIndex(1), correct)
```

The root position is always GIndex 1 and its trace index is the right most index.

ajsutton (approved):
LGTM. Just need to fix up the root position as suggested.

op-challenger/game/fault/test/claim_builder.go
refcell: Oh slipped my mind, thanks!

PR #7199, opened 2023-09-12
op-challenger: Use helper to define the expected responses

ajsutton:
**Description**

Use helper to define the expected responses in game solver tests. Avoids the need to track claim indices when writing tests.

Downside is it required moving the `Action` type out of the `solver` package to `types`. `solver` already depends on `types.Game` so seems like a reasonable thing to do.

**Tests**

All about the tests.


op-challenger/game/fault/types/actions.go
```diff
@@ -0,0 +1,28 @@
...
+
+const (
+	ActionTypeMove ActionType = "move"
+	ActionTypeStep ActionType = "step"
+)
+
+type Action struct {
+	Type      ActionType
+	ParentIdx int
+	IsAttack  bool
```
refcell, line 19: Since we're refactoring the `Action` here, what do you think about remove the `IsAttack` field in `Action` and making `ActionType`s more verbose to include whether the action is an attack or defense?

For example
```golang
const (
	AttackMoveAction ActionType = "attack_move"
        DefendMoveAction ActionType = "defend_move"
	AttackStepAction ActionType = "attack_step"
        DefendStepAction ActionType = "defend_step"
)
```

refcell (approved):

op-challenger/game/fault/types/actions.go
ajsutton: I actually started that way and found that it was really painful. You wind up doing a lot of `if Type == AttackMoveAction || Type == DefendMoveAction` type stuff.

PR #7200, opened 2023-09-12
op-challenger: Define basic validation rules for all honest actions.

ajsutton:
**Description**

Defines some basic validation rules that the GameSolver tests apply to every action the solver returns. This ensures that all the defined invariants hold for every situation we define a test for.

**Tests**

Improves unit tests.


refcell (approved):

Inphi (commented):
LGTM. Could also add a rule restricting moves against root to attacks.

trianglesphere (approved):

Inphi (approved):

PR #7201, opened 2023-09-12
cannon: Use atomic write pattern to avoid leaving partially written files

ajsutton:
**Description**

Modifies cannon to write to a temp file and then rename into place. This avoids creating partially written files if cannon crashes while writing the file.

**Tests**

Existing unit tests ensure the write still operates as expected.

**Metadata**

- https://github.com/ethereum-optimism/client-pod/issues/54


cannon/cmd/json.go
```diff
@@ -29,13 +30,20 @@ func loadJSON[X any](inputPath string) (*X, error) {
 
 func writeJSON[X any](outputPath string, value X, outIfEmpty bool) error {
 	var out io.Writer
+	finish := func() error { return nil }
 	if outputPath != "" {
-		f, err := ioutil.OpenCompressed(outputPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0755)
+		// Write to a tmp file but reserve the file extension if present
+		tmpPath := outputPath + "-tmp" + path.Ext(outputPath)
```
protolambda, line 36: This will be `foobar.json-tmp.json` which I assume is fine, but `foobar.json.tmp` may be cleaner. If the tmp file is still around, it may not have been written fully, and thus not quite json. And we may want to add `*.tmp` to the `.gitignore` (assuming it's not already in the global gitignore of the user).

protolambda (approved):
Change looks good, but maybe we should change the temporary the file-extension and add it to the gitignore.

cannon/cmd/json.go
ajsutton: Yeah I wasn't too worried about what the temp file name was, but needed to preserve the extension so the gzip detection works.  `foobar.json-tmp.json` or `foobar.json.tmp` makes no difference, but `foobar.json.gz-tmp.gz` is compressed and `foobar.json.gz.tmp` is not.

We could do something smarter though like use `foobar.json.gz~` like vim and have the gzip detection pick up on either `.gz` or `.gz~`.

PR #7215, opened 2023-09-12
op-node: Add canyon network upgrade override

trianglesphere:
**Description**

Adds flags & config for the meadow network upgrade.


op-node/flags/flags.go
```diff
@@ -244,6 +244,12 @@ var (
 		EnvVars: prefixEnvVars("BETA_EXTRA_NETWORKS"),
 		Hidden:  true,
 	}
+	PostRegolithOverrideFlag = &cli.Uint64Flag{
+		Name:    "override.postregolith",
```
ajsutton: Similar to op-geth PR, it would be really nice if we could just pick a name for this fork and use it from the start.

ajsutton (approved):
Approved but with a preference to just nail someone down and make them pick a final name for the fork. :)

op-node/flags/flags.go
trianglesphere: naming it meadow

protolambda (commented):
Suggested hardfork naming changed again. Change looks good, but we may have to swap the name :sweat:

trianglesphere:
> Suggested hardfork naming changed again

Do we have a specific name we know we want to use? At this point I'd like to merge this so we can start working on it. I'm fine renaming it - that's easy.

trianglesphere:
@protolambda this needs your review b/c your the codeowner

protolambda (approved):

PR #7216, opened 2023-09-12
cannon: Use atomic write pattern to avoid leaving partially written files

ajsutton:
**Description**

 Use atomic write pattern to avoid leaving partially written files (reapply with fixes).

Writing to stdout is now done by setting the output to - and an empty string means do not write. Generating the cannon prestate no longer needs to write the output to `/dev/null`.


ajsutton:
Confirmed e2e tests are actually running this time. :)

protolambda (approved):

PR #7217, opened 2023-09-12
Airdrop 2 Onion Cli

spacesailor24:


packages/contracts-bedrock/.env.example
```diff
@@ -0,0 +1,5 @@
+GITHUB_TOKEN=
+SOURCE_FILE="/absolute/path/to/unclaimed_op_drop.txt"
+SANCTIONED_URL="https://raw.githubusercontent.com/ethereum-optimism/sanctioned-addresses/main/list.json"
+SANCTIONED_FILE="/absolute/path/to/sanctioned-addresses.json"
+SAVE_DESTINATION="/absolute/path/to/non-sanctioned-addresses.txt"
```
tremarkley, line 5: maybe we should prefix all of these with `ONION_` since they are specific to the onion cli?

tremarkley (approved):

tynes:
We are trying to keep minimal js deps in contacts bedrock so please put this in another package or keep it someplace outside of the monorepo as I don't believe it'll be used long term

nickbalestra (changes requested):
Agree with @tynes , Can we move this to its own repo, so that we have the code somewhere for future reference, but no need to have this within the optimism monorepo.

PR #7219, opened 2023-09-13
op-challenger: Add unit test for the poisoned pre-state case

ajsutton:
**Description**

Document the poisoned pre-state test in a unit test.

### DIFF
--- a/op-challenger/game/fault/solver/game_solver_test.go
+++ b/op-challenger/game/fault/solver/game_solver_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"testing"
 
+	"github.com/ethereum-optimism/optimism/cannon/mipsevm"
 	faulttest "github.com/ethereum-optimism/optimism/op-challenger/game/fault/test"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/stretchr/testify/require"
@@ -82,6 +83,32 @@ func TestCalculateNextActions(t *testing.T) {
 					DefendCorrect()
 				lastHonestClaim.AttackCorrect().ExpectStepDefend()
 				lastHonestClaim.Attack(common.Hash{0xdd}).ExpectStepAttack()
+			},
+		},
+
+		{
+			name:                "PoisonedPreState",
+			agreeWithOutputRoot: true,
+			setupGame: func(builder *faulttest.GameBuilder) {
+				// Create a cannon state that is exited and resolved that the output root was invalid
+				maliciousState := mipsevm.State{
+					Memory:   mipsevm.NewMemory(),
+					ExitCode: 1,
+					Exited:   true,
+				}
+				maliciousStateHash, _ := maliciousState.EncodeWitness().StateHash()
+
+				// Dishonest actor counters their own claims to set up a situation with an invalid prestate
+				// The honest actor should attack all claims that support the root claim (disagree with the output root)
+				builder.Seq().ExpectAttack(). // This expected action is the winning move.
+								Attack(maliciousStateHash).
+								Defend(maliciousStateHash).ExpectAttack().
+								Attack(maliciousStateHash).
+								Attack(maliciousStateHash).ExpectStepAttack()
+
+				// The attempt to step against our malicious leaf node will fail because the pre-state won't match our
+				// malicious state hash. However, it is the very first expected action, attacking the root claim with
+				// the correct hash that wins the game since it will be the left-most uncountered claim.
 			},
 		},
 	}
--- a/op-challenger/game/fault/solver/game_solver_test.go
+++ b/op-challenger/game/fault/solver/game_solver_test.go
@@ -5,7 +5,6 @@ import (
 	"encoding/hex"
 	"testing"
 
-	"github.com/ethereum-optimism/optimism/cannon/mipsevm"
 	faulttest "github.com/ethereum-optimism/optimism/op-challenger/game/fault/test"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


op-challenger/game/fault/solver/game_solver_test.go
```diff
@@ -84,6 +85,32 @@ func TestCalculateNextActions(t *testing.T) {
...
+
+				// Dishonest actor counters their own claims to set up a situation with an invalid prestate
+				// The honest actor should attack all claims that support the root claim (disagree with the output root)
+				builder.Seq().ExpectAttack(). // This expected action is the winning move.
+								Attack(maliciousStateHash).
+								Defend(maliciousStateHash).ExpectAttack().
+								Attack(maliciousStateHash).
+								Attack(maliciousStateHash).ExpectStepAttack()
+
+				// The attempt to step against our malicious leaf node will fail because the pre-state won't match our
```
Inphi, line 103: do you mean to say the step will _succeed_, countering the malicious claim?

op-challenger/game/fault/solver/game_solver_test.go
```diff
@@ -84,6 +85,32 @@ func TestCalculateNextActions(t *testing.T) {
...
+				// malicious state hash. However, it is the very first expected action, attacking the root claim with
+				// the correct hash that wins the game since it will be the left-most uncountered claim.
```
Inphi, line 105: Isn't the attack against root also malicious, with an incorrect hash?
Seems to me that this game, the CHALLENGER wins (by default) not because the first action is correct, but because the DEFENDER failed provide an uncontested claim. Which is the correct outcome.

Inphi (commented):
Thanks for writing a test case for this.

op-challenger/game/fault/solver/game_solver_test.go
ajsutton: No, the setup I'm going for is where the honest challenger can't call step because it would have to use the attacker's pre-state and that won't be what it tries to provide.

The claims the honest actor evaluates are:
```
    game_solver_test.go:123: Claim 0: Pos: 1 TraceIdx: 15 ParentIdx: 0, Countered: false, Value: 0x000000000000000000000000000000000000000000000000000000000000000f
    game_solver_test.go:123: Claim 1: Pos: 2 TraceIdx: 7 ParentIdx: 0, Countered: false, Value: 0x018bd1331cba4019004e3d2a2cb9fcb748d97fdfe2972b8b856b71d55e06c2b3
    game_solver_test.go:123: Claim 2: Pos: 6 TraceIdx: 11 ParentIdx: 1, Countered: false, Value: 0x018bd1331cba4019004e3d2a2cb9fcb748d97fdfe2972b8b856b71d55e06c2b3
    game_solver_test.go:123: Claim 3: Pos: 12 TraceIdx: 9 ParentIdx: 2, Countered: false, Value: 0x018bd1331cba4019004e3d2a2cb9fcb748d97fdfe2972b8b856b71d55e06c2b3
    game_solver_test.go:123: Claim 4: Pos: 24 TraceIdx: 8 ParentIdx: 3, Countered: false, Value: 0x018bd1331cba4019004e3d2a2cb9fcb748d97fdfe2972b8b856b71d55e06c2b3
```

So trace index 7, 8 and 9 have all been provided by the attacker and the honest actor won't be able to provide a state that matches them.

Actually I probably should have just used random hash values that there is no pre-image for rather than creating a malicious state - if the honest actor was really smart they could actually apply a step based on that malicious state. I'll push that change up.

op-challenger/game/fault/solver/game_solver_test.go
ajsutton: There will be two attacks against the root - the malicious one from the first `Attack(maliciousStateHash)` call which is already countered by the dishonest actor and we expect the honest actor to also attack the root but with the correct claim.  The honest actor's attack on the root will be what goes on to win the game - the dishonest actor could respond to it but it shows the honest actor can "escape" the poisoning by attacking left.

Expected moves are:
```
    game_solver_test.go:125: Move 0: Type: move, ParentIdx: 0, Attack: true, Value: 0x0157c6880317692d1e241d4e588624849a414af4c7f958811b533447ec42aec8, PreState: , ProofData: 
    game_solver_test.go:125: Move 1: Type: move, ParentIdx: 2, Attack: true, Value: 0x0110caf4eb645e6e135f3034a51a2295a2e216f2fb35fdb47d2c2a2d591db7ff, PreState: , ProofData: 
    game_solver_test.go:125: Move 2: Type: step, ParentIdx: 4, Attack: true, Value: 0x0000000000000000000000000000000000000000000000000000000000000000, PreState: 00000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000068, ProofData: 07
```

ajsutton:
I'm tossing up how we can integrate the actual resolution and step implementations into these tests. I don't really want to reimplement that logic in go but setting up this state in a real block chain would be way slower.  It would be really nice to assert that if the honest actors actions were applied that it would always wind up in a winning position though.

op-challenger/game/fault/solver/game_solver_test.go
Inphi: What I meant is that the dishonest actor calls step (as `ExpectStepAttack` asserts) and succeeds in countering a claim supported by the honest challenger.
The motivation here is to DoS/coerce the honest challenger into supporting untruthful claims. If the poisoned claim remains uncontested after a step, then there's no point to creating poisoned claims.

op-challenger/game/fault/solver/game_solver_test.go
Inphi: gotcha. I misread the setup code as the only moves in the game.

op-challenger/game/fault/solver/game_solver_test.go
Inphi: I think we agree here, but perhaps I'm interpreting a "failed step", as worded here in the comments, to mean the claim being disputed remains uncontested.

ajsutton: The `Expect*()` calls are setting out what we expect the honest challenger to do - or at least try to do. So it's actually the honest challenger that tries to call step but it will supply the wrong pre-state and so the call will fail.  With the original malicious state that had a pre-image the dishonest actor could have called step successfully to counter their own leaf claim, though that would have then been showing their original root claim to be invalid which they aren't trying to do.

I'm definitely keen to work out how to make all of this easier to understand - the aim of these tests is to give us something more concrete to discuss but that's not helpful if it's still ambiguous.

Inphi: Understood. The poison scenario I originally had in mind is where the honest actor moves last, so it's the dishonest's _turn_ to step. But this setup is analogous and illustrates poison similarly.

Inphi (approved):

PR #7226, opened 2023-09-13
contracts-bedrock: ProtocolVersions contract

protolambda:
**Description**

`ProtocolVersions` smart contract to signal superchain protocol upgrades to nodes. See #6388 for specs.

Changes:
- New contract
- Contract testing
- Metadata update (semver, gas snapshot)
- Deploy script
- Deploy config updates
- Go bindings artifacts list update + regenerated bindings
- devnetL1 template update, used in op-e2e testing
- op-chain-ops config test updates

**Tests**

The new contract largely follows the same patterns as established in the `SystemConfig` and is tested the same way as well.

**Metadata**

Fix CLI-4205

Fix CLI-4206


packages/contracts-bedrock/test/ProtocolVersions.t.sol
```diff
@@ -0,0 +1,124 @@
...
+        assertEq(protocolVersions.owner(), alice);
+
+        assertEq(ProtocolVersion.unwrap(protocolVersionsImpl.required()), 0);
+        assertEq(ProtocolVersion.unwrap(protocolVersionsImpl.recommended()), 0);
+        assertEq(protocolVersionsImpl.owner(), address(0xdEad));
+    }
+
+    /// @dev Ensures that the events are emitted during initialization.
+    function test_initialize_events_succeeds() external {
+        //vm.store(address(protocolVersions), bytes32(uint256(106)), bytes32(0));
```
tynes: You can delete this dead code

tynes:
Looks like a build error

```
#78 199.4 ../op-chain-ops/genesis/config.go:203:33: undefined: params.ProtocolVersion
#78 199.4 ../op-chain-ops/genesis/config.go:206:36: undefined: params.ProtocolVersion
```

protolambda:
@tynes Ah, PR-splitting issue, I'll change it to `common.Hash` for now, and then update it back to `params.ProtocolVersion` in the next PR that introduces the geth changes.

op-chain-ops/genesis/config.go
```diff
@@ -522,6 +527,8 @@ type L1Deployments struct {
 	ProxyAdmin                        common.Address `json:"ProxyAdmin"`
 	SystemConfig                      common.Address `json:"SystemConfig"`
 	SystemConfigProxy                 common.Address `json:"SystemConfigProxy"`
+	ProtocolVersions                  common.Address `json:"ProtocolVersions"`
```
tynes, line 530: In the previous PR, I did add a useful comment to the godoc for this struct

op-chain-ops/genesis/config.go
```diff
@@ -198,6 +197,12 @@ type DeployConfig struct {
 	// FundDevAccounts configures whether or not to fund the dev accounts. Should only be used
 	// during devnet deployments.
 	FundDevAccounts bool `json:"fundDevAccounts"`
+	// RequiredProtocolVersion indicates the protocol version that
+	// nodes are required to adopt, to stay in sync with the network.
+	RequiredProtocolVersion common.Hash `json:"requiredProtocolVersion"`
+	// RequiredProtocolVersion indicates the protocol version that
+	// nodes are recommended to adopt, to stay in sync with the network.
+	RecommendedProtocolVersion common.Hash `json:"recommendedProtocolVersion"`
```
tynes, line 205: We should probably add something into the `Check` method. If we return an error when its the empty value, then it will break `pnpm validate-deploy-configs` in `contracts-bedrock` and we will need to update all deploy configs with values. Another option is to log a warning if it is empty, which seems sane to do in the short term. Then when its in production, we can error if its not set

op-chain-ops/genesis/config.go
protolambda: Was it this exact struct? Let me try to find it, I may have missed it when diffing and cherry-picking changes from the PR

protolambda: Can't find it in the old PR, I believe it's the deploy-config struct only that had comments, the L1Deployments one does not. Feel free to suggest some (the other addresses do not have comments here either though).

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,131 @@
...
+    /// @param _owner             Initial owner of the contract.
+    /// @param _required          Required protocol version to operate on this chain.
+    /// @param _recommended       Recommended protocol version to operate on thi chain.
+    function initialize(
+        address _owner,
+        ProtocolVersion _required,
+        ProtocolVersion _recommended
+    )
+        public
+        reinitializer(2)
```
maurelian, line 62: More of a question: 1 would work here since this is an new contract, are we just standardizing on 2 for consistency across the codebase?

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    /// @notice Stores a ProtocolVersion in an arbitrary storage slot, `_slot`.
+    /// @param _version The protocol version to store
+    /// @param _slot The storage slot to store the address in.
+    /// @dev WARNING! This function must be used cautiously, as it allows for overwriting values
+    ///      in arbitrary storage slots.
+    function _setProtocolVersion(ProtocolVersion _version, bytes32 _slot) internal {
+        assembly {
+            sstore(_slot, _version)
+        }
```
maurelian, line 87: Could we just inline this logic to `_setRecommended` and `_setRequired`? Then we don't need this function or the big scary warning.

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,131 @@
...
+        _setRequired(_required);
+        _setRecommended(_recommended);
+    }
+
+    /// @notice Returns a ProtocolVersion stored in an arbitrary storage slot.
+    ///         These storage slots decouple the storage layout from solc's automation.
+    /// @param _slot The storage slot to retrieve the address from.
+    function _getProtocolVersion(bytes32 _slot) internal view returns (ProtocolVersion out_) {
+        assembly {
+            out_ := sload(_slot)
```
maurelian, line 75: same as below, why not just inline this into the other two places its used?

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
refcell: ++

refcell (commented):
Looks good to me aside from existing comments

op-chain-ops/genesis/config.go
protolambda: There is no clear spec how to add optional backwards compatible deploy config values. I think the current approach is correct. I'll open an issue to track it, feel free to tweak the deploy config in a follow-up PR.

protolambda: Tracking in CLI-4401

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
protolambda: @tynes suggested to keep `2`, like the system config, for consistency. We can then increment the value globally across all contracts as necessary, to avoid cross-contract storage issues.

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
protolambda: I appreciate the suggestion, but this was implemented exactly what `_getAddress` does in the `SystemConfig`, which I do not think is a bad reference, as we re-use the version loading here twice, and possibly more in the future.

Do we want slight duplication + removal of helper + inconsistency with SystemConfig, or do we want verbosity but consistency with SystemConfig? Need one design direction here, but I'm getting different ideas from different reviewers.

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
protolambda: See other comment

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
maurelian: `_setAddress()` is called 8 times in SystemConfig, so the duplication feels more justified to me there.

maurelian (approved):

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
maurelian: I have approved, I think inlining the logic is the right move, but I won't block it. You can resolve this as you see fit.

tynes (approved):

PR #7227, opened 2023-09-13
op-node: protocol versions signal handling

protolambda:
**Description**

Implements op-node protocol-version signal handling (see specs PR: https://github.com/ethereum-optimism/optimism/pull/6388 ):
- Metrics
- Logging
- Communication of protocol version to geth
- Protocol version loading as part of runtime config system
- Opt-in functionality to halt node on outdated config
- Update op-geth Go dependency ~~(TODO: this is a placeholder reference to the commit of the open PR for now)~~ Updated to latest op-geth at [`b84ba11915a0aac316172bc27f10c7e2287b922b`](https://github.com/ethereum-optimism/op-geth/commit/b84ba11915a0aac316172bc27f10c7e2287b922b)

Note: this PR builds on two other PRs:
- The `ProtocolVersions` contract: #7226 
- The op-geth version handling (defines types and utils that are reused here in op-node): https://github.com/ethereum-optimism/op-geth/pull/128

This also includes a minor fix to `eth_getProof`-verification: zero-storage is not present in the MPT, and thus the proof shows exclusion, and returns a nil result and a nil error. To verify, we need to not encode the nil as an empty RLP value, but rather check that the claimed storage value is zero.

**before merging this PR we should merge the above two PRs, and update the `go.mod` to reference the canonical op-geth commit**

New optional hidden beta flags:
- `--beta.rollup.halt`: to opt in to halting
- `--beta.rollup.load-protocol-versions`: to opt in to the new configuration. If not set, the protocol versions contract address configuration is dropped on startup, and not used when dropped.

**Tests**

- Version types/utils are unit-tested in op-geth
- Config updates and handling is tested in op-e2e, including halting.
  - Halting is tested more extensively in op-geth.
  - The op-e2e overhead is a bit too large to run a matrix of halting tests, so I left it at a recommended + required version-change test there.

**Metadata**

Fix CLI-4210

Fix CLI-4400


op-e2e/system_test.go
```diff
@@ -1434,3 +1435,120 @@ func TestRuntimeConfigReload(t *testing.T) {
...
+		return struct{}{}, fmt.Errorf("no change yet, seeing %s but looking for %s", v, newRecommendedProtocolVersion)
+	})
+	require.NoError(t, err)
+}
+
+func TestRequiredProtocolVersionChangeAndHalt(t *testing.T) {
+	InitParallel(t)
+
+	cfg := DefaultSystemConfig(t)
+	// activate the functionality by updating the config // TODO may need to just activate it by default in op-e2e?
```
tynes: This seems like a good idea to do before merging?

tynes: Could read from the global deploy config that is populated in the `init` function

op-e2e/system_test.go
```diff
@@ -1434,3 +1435,120 @@ func TestRuntimeConfigReload(t *testing.T) {
...
+	cfg.Nodes["verifier"].Rollup.ProtocolVersionsAddress = cfg.L1Deployments.ProtocolVersions
+	// to speed up the test, make it reload the config more often, and do not impose a long conf depth
+	cfg.Nodes["verifier"].RuntimeConfigReloadInterval = time.Second * 5
+	cfg.Nodes["verifier"].Driver.VerifierConfDepth = 1
+	// configure halt in verifier op-node
+	cfg.Nodes["verifier"].RollupHalt = "major"
+	// configure halt in verifier op-geth node
+	cfg.GethOptions["verifier"] = []geth.GethOption{
```
tynes: Should this append instead of set? In case a global one is applied in the future, it would overwrite it

op-node/node/config.go
```diff
@@ -128,5 +132,8 @@ func (cfg *Config) Check() error {
 			return fmt.Errorf("p2p config error: %w", err)
 		}
 	}
+	if !(cfg.RollupHalt == "" || cfg.RollupHalt == "major" || cfg.RollupHalt == "minor" || cfg.RollupHalt == "patch") {
```
tynes, line 135: Feels like this would be ideal to have broken out into a smaller function that can be unit tested + the rest of the code in this function is behind higher level abstractions

op-node/node/runtime_config.go
```diff
@@ -67,18 +82,46 @@ func (r *RuntimeConfig) P2PSequencerAddress() common.Address {
...
+		recommendedProtoVersion = params.ProtocolVersion(recommendedVal)
+	}
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	r.l1Ref = l1Ref
-	r.p2pBlockSignerAddr = common.BytesToAddress(val[:])
+	r.p2pBlockSignerAddr = common.BytesToAddress(p2pSignerVal[:])
+	r.required = requiredProtVersion
+	r.recommended = recommendedProtoVersion
 	r.log.Info("loaded new runtime config values!", "p2p_seq_address", r.p2pBlockSignerAddr)
```
tynes, line 125: No logging of the startup req and rec versions?

op-node/node/superchain.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// We may need to halt the node, if the user opted in to handling incompatible protocol-version signals
+	n.HaltMaybe()
+}
+
+// HaltMaybe halts the rollup node if the runtime config indicates an incompatible required protocol change
+// and the node is configured to opt-in to halting at this protocol-change level.
+func (n *OpNode) HaltMaybe() {
+	var needLevel int
+	switch n.rollupHalt {
```
tynes: This seems like a function that could be modularized and unit tested, feels pretty critical

op-node/rollup/superchain.go
```diff
@@ -4,12 +4,16 @@ import (
...
 
+	"github.com/ethereum/go-ethereum/params"
+
 	"github.com/ethereum/go-ethereum/common"
 
 	"github.com/ethereum-optimism/optimism/op-service/eth"
 	"github.com/ethereum-optimism/superchain-registry/superchain"
 )
 
+var OPStackSupport = params.ToProtocolVersion([8]byte{}, 3, 1, 0, 1)
```
tynes: Can you add a godoc to make this easier for humans to read or can we have a more human readable interface for creating this global? Will make updating it much easier

op-node/service.go
```diff
@@ -36,6 +36,10 @@ func NewConfig(ctx *cli.Context, log log.Logger) (*node.Config, error) {
 		return nil, err
 	}
 
+	if !ctx.Bool(flags.BetaRollupLoadProtocolVersions.Name) {
+		rollupConfig.ProtocolVersionsAddress = common.Address{}
```
tynes, line 41: Feels like modifying the rollup config like this warrants a logline, a node runner will probably run into this and not understand why the config isn't being picked up

op-service/eth/account_proof.go
```diff
@@ -51,6 +51,9 @@ func (res *AccountResult) Verify(stateRoot common.Hash) error {
 		if err != nil {
 			return fmt.Errorf("failed to verify storage value %d with key %s (path %x) in storage trie %s: %w", i, entry.Key, path, res.StorageHash, err)
 		}
+		if val == nil && entry.Value.ToInt().Cmp(common.Big0) == 0 { // empty storage is zero by default
```
tynes, line 54: This is useful for proving storage against empty slots?

protolambda: yes, zeroed protocol versions were causing issues here, this handles the zero-default-value case. (it's mentioned in the PR description changes too).

op-node/node/runtime_config.go
protolambda: The runtime config struct `RuntimeConfig` does not have its own logger. After the config `Load` call completes the caller can choose the updated config values. This is what the `Node` does.

op-e2e/system_test.go
protolambda: done

op-e2e/system_test.go
protolambda: done, and fixed the other previous other tests that were like this too

op-node/rollup/superchain.go
protolambda: added a V0 struct definition, so inputs can be declared by name, and then encoded with a `.Encode()` call

op-node/service.go
protolambda: added logging

op-node/node/config.go
protolambda: Can you give an example? I refactored out the other code that does the comparison/option check and outputs the halting-bool, and added a vector unit test for that to cover the halting cases, and ensure there are no other cases. Is there anything to add here? It's just checking the inputs here, unit-test will look the same as the code, just a string comparison.

tynes: Ok I am fine with leaving it as is

tynes (approved):

mslipper (approved):

mslipper (approved):

PR #7242, opened 2023-09-13
cannon: Use zlib compressed base64 instead of hex for serialization

trianglesphere:
**Description**

This changes the serialization format of state.json. It compresses the pages memory with zlib and then uses base64 encoding instead of hex encoding because base64 encoding is more compact.

This sees significant reduction in file size of snapshots. It also reduces the inuse memory caused by JSON marshalling. The sync pool of the zlib writer may slightly reduce the inuse allocations, but significantly reduces the lifetime allocations.

**Tests**

Manually tested detect snippet. Have tested the cannon binary in k8s.

### DIFF
--- a/cannon/mipsevm/page.go
+++ b/cannon/mipsevm/page.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"regexp"
 	"sync"
 
 	"github.com/ethereum/go-ethereum/crypto"
@@ -19,19 +18,6 @@ var zlibWriterPool = sync.Pool{
 		var buf bytes.Buffer
 		return zlib.NewWriter(&buf)
 	},
-}
-
-func detectEncoding(data []byte) (string, error) {
-	// Regular expressions to check for base64 and hex patterns
-	base64Pattern := "^[A-Za-z0-9+/]*={0,2}$"
-	hexPattern := "^[0-9a-fA-F]*$"
-
-	if regexp.MustCompile(hexPattern).MatchString(string(data)) {
-		return "hex", nil
-	} else if regexp.MustCompile(base64Pattern).MatchString(string(data)) {
-		return "base64", nil
-	}
-	return "", fmt.Errorf("Unable to determine encoding for data: %s", data)
 }
 
 type Page [PageSize]byte
@@ -53,12 +39,6 @@ func (p *Page) MarshalJSON() ([]byte, error) { // nosemgrep
 func (p *Page) UnmarshalJSON(dat []byte) error {
 	// Strip off the `"` characters at the start & end.
 	dat = dat[1 : len(dat)-1]
-	// Detect hex or bas64 encoding. legacy hex encoding is uncompressed
-	if t, err := detectEncoding(dat); err != nil {
-		return err
-	} else if t == "hex" {
-		return p.UnmarshalText(dat)
-	}
 	// Decode b64 then decompress
 	r, err := zlib.NewReader(base64.NewDecoder(base64.StdEncoding, bytes.NewReader(dat)))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cannon/mipsevm/page.go
```diff
@@ -1,12 +1,39 @@
...
+func detectEncoding(data []byte) (string, error) {
+	// Regular expressions to check for base64 and hex patterns
+	base64Pattern := "^[A-Za-z0-9+/]*={0,2}$"
+	hexPattern := "^[0-9a-fA-F]*$"
+
+	if regexp.MustCompile(hexPattern).MatchString(string(data)) {
+		return "hex", nil
+	} else if regexp.MustCompile(base64Pattern).MatchString(string(data)) {
+		return "base64", nil
+	}
```
ajsutton: There is an issue where were hex chars are a strict subset of base64 so while its unlikely for memory, it's possible that we wrote base64 using only valid hex chars and would then interpret it wrong on reload.  That would cause games to be lost incorrectly so isn't something we can really risk. (Assuming I'm reading the regex right and the trailing `=` chars on base64 are optional).

We might be better off prefixing base64 content with `base64:` so there's no doubt about the formatting of the data.

cannon/mipsevm/page.go
```diff
@@ -1,12 +1,39 @@
...
+	},
+}
+
+func detectEncoding(data []byte) (string, error) {
+	// Regular expressions to check for base64 and hex patterns
+	base64Pattern := "^[A-Za-z0-9+/]*={0,2}$"
+	hexPattern := "^[0-9a-fA-F]*$"
+
+	if regexp.MustCompile(hexPattern).MatchString(string(data)) {
+		return "hex", nil
```
Inphi: nit: use enums for the encodings

cannon/mipsevm/page.go
trianglesphere: yea, it's a concern. It's also a problem in the other direction though - if a page is all 0s, it will be reported as base64 (which is why the hex check is first).

TBH I think we should migrate to this new format rather than supporting both, but for the short term supporting both is nice in terms of `prestate.json`

ajsutton: Yeah agreed we don't want to live with hex for long, but having a clear `base64:` prefix would simplify the check and completely avoid any risk of incorrect detection while preserving compatibility for the existing prestate. I just think it's worth avoiding ever adding code that would potentially lead to us making invalid claims.

cannon/mipsevm/page.go
```diff
@@ -1,12 +1,39 @@
...
+	} else if regexp.MustCompile(base64Pattern).MatchString(string(data)) {
+		return "base64", nil
+	}
+	return "", fmt.Errorf("Unable to determine encoding for data: %s", data)
+}
+
 type Page [PageSize]byte
 
 func (p *Page) MarshalText() ([]byte, error) {
```
Inphi: We should be able to remove this function. It's unused now that `Page` implements `MarshalJSON`

cannon/mipsevm/page.go
```diff
@@ -1,12 +1,39 @@
...
+		return zlib.NewWriter(&buf)
+	},
+}
+
+func detectEncoding(data []byte) (string, error) {
+	// Regular expressions to check for base64 and hex patterns
+	base64Pattern := "^[A-Za-z0-9+/]*={0,2}$"
+	hexPattern := "^[0-9a-fA-F]*$"
+
+	if regexp.MustCompile(hexPattern).MatchString(string(data)) {
```
Inphi: nit: can pre-compile the regex at init time

Inphi (approved):
LGTM. Just a couple nits

ajsutton (approved):
LGTM.

PR #7252, opened 2023-09-14
feat(proxyd): high availability

felipe-op:
**Description**

This change implements high availability for proxyd instances sharing the same consensus.

**Tests**

Deployed to dev

**Metadata**

Project: https://www.notion.so/Proxyd-HA-7bbb92766d8342f8af241bea99fd4a12

This change fixes
- https://linear.app/optimism/issue/INF-426/bump-go-redis-to-v9
- https://linear.app/optimism/issue/INF-429/lock-state-machine
- https://linear.app/optimism/issue/INF-428/ha-configs
- https://linear.app/optimism/issue/INF-427/split-tracker-into-local-and-remote-states


proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,282 @@ type ConsensusTracker interface {
...
+
+		mutex := rs.NewMutex(key,
+			redsync.WithExpiry(ct.lockPeriod),
+			redsync.WithFailFast(true),
+			redsync.WithTries(1))
+
+		if err := mutex.Lock(); err != nil {
+			log.Debug("failed to obtain lock", "err", err)
+			ct.leader = false
+			return
```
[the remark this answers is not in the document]
felipe-op: this lock is hold indefinitely and it is extended until the leader dies

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,284 @@ type ConsensusTracker interface {
...
 type InMemoryConsensusTracker struct {
-	latestBlockNumber    hexutil.Uint64
-	safeBlockNumber      hexutil.Uint64
-	finalizedBlockNumber hexutil.Uint64
-	mutex                sync.Mutex
+	mutex sync.Mutex
+	state *ConsensusTrackerState
 }
 
 func NewInMemoryConsensusTracker() ConsensusTracker {
```
mslipper, line 51: [nit] Could have this return the concrete implementation to avoid having to cast in the `NewRedisConsensusTracker` method below.

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,284 @@ type ConsensusTracker interface {
...
+func (ct *RedisConsensusTracker) Init() {
+	go func() {
+		for {
+			// follow same context as backend group poller
+			ctx := ct.backendGroup.Consensus.ctx
+			timer := time.NewTimer(ct.heartbeatInterval)
+			ct.stateHeartbeat()
+
+			select {
+			case <-timer.C:
```
mslipper, line 179: What is this timer doing?

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,284 @@ type ConsensusTracker interface {
...
+	jsonState, err := json.Marshal(ct.local.state)
+	if err != nil {
+		log.Error("failed to marshal local", "err", err)
+		ct.leader = false
+		return
+	}
+	ct.client.Set(ct.ctx, ct.key(fmt.Sprintf("state:%s", mutexVal)), jsonState, ct.lockPeriod)
+
+	leader, _ := os.LookupEnv("HOSTNAME")
+	if leader == "" {
```
mslipper: Empty `if` statement?

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,284 @@ type ConsensusTracker interface {
...
+	}
+
+	return tracker
+}
+
+func (ct *RedisConsensusTracker) Init() {
+	go func() {
+		for {
+			// follow same context as backend group poller
+			ctx := ct.backendGroup.Consensus.ctx
```
mslipper: We pass in a context to the `NewRedisConsensuTracker` and use it in other methods, but use the one from the backend group's consensus here. Why not use the context in the tracker itself?

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,311 @@ type ConsensusTracker interface {
...
+				return
+			}
+			state := &ConsensusTrackerState{}
+			err = json.Unmarshal([]byte(val), state)
+			if err != nil {
+				log.Error("failed to unmarshal the remote state", "err", err)
+				return
+			}
+			ct.remote.mutex.Lock()
+			defer ct.remote.mutex.Unlock()
```
mslipper: Can we put the locking logic inside the remote state? Seems suboptimal to have to manage a struct's internal lock from outside the struct itself.

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,311 @@ type ConsensusTracker interface {
...
+	return ct.remote.GetFinalizedBlockNumber()
 }
 
 func (ct *RedisConsensusTracker) SetFinalizedBlockNumber(blockNumber hexutil.Uint64) {
-	ct.client.Set(ct.ctx, ct.key("finalized"), blockNumber, 0)
+	ct.local.SetFinalizedBlockNumber(blockNumber)
+}
+
+func (ct *RedisConsensusTracker) postPayload(mutexVal string) {
+	ct.remote.mutex.Lock()
```
mslipper: Ditto here - can we move this mutex management inside the `remote` object?

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,311 @@ type ConsensusTracker interface {
...
+			defer ct.remote.mutex.Unlock()
+			ct.remote.state.update(state)
+			log.Debug("updated state from remote", "state", val, "leader", leaderName)
+		}
+	} else {
+		if !ct.local.Valid() {
+			log.Warn("local state is not valid or behind remote, skipping")
+			return
+		}
+		if ct.remote.Valid() && ct.local.Behind(ct.remote) {
```
mslipper, line 260: Nice 👍

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,311 @@ type ConsensusTracker interface {
...
+			case <-timer.C:
+			case <-ctx.Done():
+				timer.Stop()
+				return
+			}
+		}
+	}()
+}
+
+func (ct *RedisConsensusTracker) stateHeartbeat() {
```
mslipper, line 189: `stateHeartbeat` does a lot of logging but returns void - may be better to have `stateHeartbeat` return an error, and do the logging in the caller.

mslipper (changes requested):

proxyd/consensus_tracker.go
felipe-op: Thats an option, but there is a trade-off.  You would be just moving the cast to the invoker, to keep the impl abstracted from the interface.

The idea here is to make the `local` and `remote` managed by the tracker, so everything is self contained and abstracted from outside.

proxyd/consensus_tracker.go
felipe-op: It's the heartbeat timer. By default it causes the state to update every 2 seconds.

proxyd/consensus_tracker.go
felipe-op: Oops, that was embarassing.  The env var `HOSTNAME`  is the pod name when running inside k8s, but it has no default value in local testing.  This `if` was supposed to coalesce it into something but I never reached a really unique value for it, so it stayed there hanging useless until this review. 

I'll remove it.

proxyd/consensus_tracker.go
felipe-op: tldr is I wanted to be able to cancel the parent context.  But it makes sense to make the called to make that decision instead of causing side effects to the consensus ctx.

proxyd/consensus_tracker.go
felipe-op: Moved

proxyd/consensus_tracker.go
felipe-op: Moved as well

proxyd/consensus_tracker.go
felipe-op: I rather not.  This is a complex logic and it is more readable to have the logging in-place.  They help to understand the story behind the logic in a way the code is communicating to the outside world.

proxyd/consensus_tracker.go
mslipper: Ooh I see, you make a new timer every tick of the loop.

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,308 @@ type ConsensusTracker interface {
...
+}
+
+func (ct *RedisConsensusTracker) Init() {
+	go func() {
+		for {
+			timer := time.NewTimer(ct.heartbeatInterval)
+			ct.stateHeartbeat()
+
+			select {
+			case <-timer.C:
```
mslipper, line 179:
```suggestion
			case <-timer.C:
			    continue
```

Added `continue` to make this clearer

proxyd/consensus_tracker.go
mslipper: I think you may have forgotten this one - since you moved the mutex inside `update()` I'm pretty sure this will deadlock. I think you can remove the locking here altogether.

proxyd/consensus_tracker.go
```diff
@@ -21,101 +26,308 @@ type ConsensusTracker interface {
...
 }
 
 func (ct *RedisConsensusTracker) SetFinalizedBlockNumber(blockNumber hexutil.Uint64) {
-	ct.client.Set(ct.ctx, ct.key("finalized"), blockNumber, 0)
+	ct.local.SetFinalizedBlockNumber(blockNumber)
+}
+
+func (ct *RedisConsensusTracker) postPayload(mutexVal string) {
+	ct.remote.mutex.Lock()
+	defer ct.remote.mutex.Unlock()
```
mslipper:
```suggestion
```

proxyd/consensus_tracker.go
felipe-op: Nice catch

mslipper (approved):

PR #7253, opened 2023-09-14
op-node: Channel ordering fix for canyon

### LINKED ISSUE #7453: Adjust channel ordering in the channel bank
At the moment channels in the channel bank are read in first in first out order. This should be updated so the channels are read in first ready order. First in first out should remain the tie breaker for channels that are ready at the same time.

Internal Notion Doc: https://www.notion.so/oplabs/Adjust-Channel-Ordering-c3a88ff25b7842dd9074ca356f5206a8

trianglesphere:
**Description**

This specifies & implements a new channel ordering which will activate with the canyon hard fork. Post canyon, the channel bank will read the first ready channel, rather than just the channel at the head of the channel bank queue.

**Tests**

Tests for the before & after behavior.

**Metadata**

- Fixes https://github.com/ethereum-optimism/optimism/issues/7453

### DIFF
--- a/op-node/rollup/derive/channel_bank.go
+++ b/op-node/rollup/derive/channel_bank.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
+	"golang.org/x/exp/slices"
 
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
 	"github.com/ethereum-optimism/optimism/op-service/eth"
@@ -116,6 +117,8 @@ func (cb *ChannelBank) IngestFrame(f Frame) {
 // Read the raw data of the first channel, if it's timed-out or closed.
 // Read returns io.EOF if there is nothing new to read.
 func (cb *ChannelBank) Read() (data []byte, err error) {
+	// Common Code. By returning `nil,nil`, we call back into this to make sure that all timed
+	// out channels at the front of the queue will eventually be removed.
 	if len(cb.channelQueue) == 0 {
 		return nil, io.EOF
 	}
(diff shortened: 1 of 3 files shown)


op-node/rollup/derive/channel_bank.go
```diff
@@ -116,6 +117,8 @@ func (cb *ChannelBank) IngestFrame(f Frame) {
 // Read the raw data of the first channel, if it's timed-out or closed.
 // Read returns io.EOF if there is nothing new to read.
 func (cb *ChannelBank) Read() (data []byte, err error) {
+	// Common Code. By returning `nil,nil`, we call back into this to make sure that all timed
+	// out channels will eventually be removed.
```
trianglesphere: TODO: `timed out channels at the front of the queue`

op-node/rollup/derive/channel_bank.go
```diff
@@ -116,6 +117,8 @@ func (cb *ChannelBank) IngestFrame(f Frame) {
 // Read the raw data of the first channel, if it's timed-out or closed.
 // Read returns io.EOF if there is nothing new to read.
 func (cb *ChannelBank) Read() (data []byte, err error) {
+	// Common Code. By returning `nil,nil`, we call back into this to make sure that all timed
+	// out channels at the front of the queue will eventually be removed.
```
ajsutton: This talks about returning `nil, nil` but the next return is actually returning `nil, io.EOF` which is a bit confusing.

Maybe the comment here is about returning `EOF` if there are no channels waiting to be processed.

Then after the `if Len(cb.cahnnelQueue) == 0` condition have this comment about removing the first channel if it's timed out and returning `nil, nil` to ensure all timed out channels at the front of the queue eventually get removed.

op-node/rollup/derive/channel_bank.go
```diff
@@ -127,23 +130,40 @@ func (cb *ChannelBank) Read() (data []byte, err error) {
...
+			return data, nil
+		}
 	}
+	return nil, io.EOF
+}
+
+// readIndex attempts to read the channel at the specified index. If the channel is
+// not ready (or timed out), it will return io.EOF.
+// If the channel read was successful, it will remove the channel from the channelQueue.
+func (cb *ChannelBank) readIndex(i int) (data []byte, err error) {
```
ajsutton: I got confused this naming which makes it seem like we're reading the channel without checking its ready.  Maybe it should be `tryReadChannel`? Or something like that.

op-node/rollup/derive/channel_bank.go
```diff
@@ -127,23 +130,40 @@ func (cb *ChannelBank) Read() (data []byte, err error) {
...
 		return nil, nil // multiple different channels may all be timed out
 	}
-	if !ch.IsReady() {
-		return nil, io.EOF
+
+	// At the point we have removed all timed out channels from the front of the channelQueue.
+	// Pre-Canyon we simply check the first index.
+	// Post-Canyon we read the entire channelQueue for the first ready channel. If no channel is
+	// available, we return `nil, io.EOF`.
+	if !cb.cfg.IsCanyon(cb.Origin().Time) {
```
ajsutton, line 142: This is comparing the L1 block time against the CanyonTime config which is set based on the L2 block:
```
	// CanyonTime  sets the activation time of the next network upgrade.
	// Active if CanyonTime != nil && L2 block timestamp >= *CanyonTime, inactive otherwise.
	CanyonTime *uint64 `json:"canyon_time,omitempty"`
```

The L2 block can't be before the L1 block but could be after it. This is safe because the channel ordering is completely independent of the L2 block content anyway but we should probably document that we are deliberately passing an L1 time in rather than L2 and consider it safe in this case. Don't want to set the expectation that they're always interchangeable.

specs/derivation.md
```diff
@@ -513,11 +513,14 @@ New frames for timed-out channels are dropped instead of buffered.
...
-
 Upon reading, while the first opened channel is timed-out, remove it from the channel-bank.
 
-Once the first opened channel, if any, is not timed-out and is ready, then it is read and removed from the channel-bank.
+Prior to the Canyon network upgrade, once the first opened channel, if any, is not timed-out and is ready,
+then it is read and removed from the channel-bank. After the Canyon network upgrade, the entire channel bank
+is scanned in FIFO order (by open time) & the first ready (i.e. not timed-out) channel will be returned.
+
+The canyon behavior will activate when frames from a L1 block whose timestamp is greater than or equal to the
+canyon time first enter the channel queue.
```
ajsutton, line 523: Good note. :)

ajsutton (commented):
Looks good to me generally.  Just a few comments that I think can aid clarity.

op-node/rollup/derive/channel_bank.go
trianglesphere: Documented. It's a little tricky with the timestamps because I think anything which is from L1 (batches, frames, channel bank / batch queue) should activate on L1 timestamps and anything which is L2 (attributes, engine API calls) should activate based on the L2 timestamp.

ajsutton: Yeah, I think that logic makes sense as long as the features are unrelated which these are and I'm having trouble thinking of a feature that wouldn't be.  Just feels like something we should always be very very clear about because it's easy to forget and cause problems.

ajsutton (approved):
LGTM.

protolambda (approved):
Nice changes

PR #7258, opened 2023-09-15
op-challenger: Add e2e test for the poisoned post-state case

ajsutton:
**Description**

Adds a skipped e2e test to reproduce the poisoned post-state case.

**Metadata**

- https://github.com/ethereum-optimism/client-pod/issues/59

### DIFF
--- a/op-e2e/faultproof_test.go
+++ b/op-e2e/faultproof_test.go
@@ -341,6 +341,66 @@ func TestCannonProposedOutputRootInvalid(t *testing.T) {
 	}
 }
 
+func TestCannonPoisonedPostState(t *testing.T) {
+	t.Skip("Known failure case")
+	InitParallel(t)
+
+	ctx := context.Background()
+	sys, l1Client := startFaultDisputeSystem(t)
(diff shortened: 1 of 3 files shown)


op-e2e/faultproof_test.go
```diff
@@ -341,6 +341,66 @@ func TestCannonProposedOutputRootInvalid(t *testing.T) {
...
+
+	// Honest defense at "dishonest" level
+	correctTrace.Defend(ctx, 1)
+
+	// Dishonest attack at "honest" level - honest move would be to defend
+	game.Attack(ctx, 2, common.Hash{0x03, 0xaa})
+
+	// Start the honest challenger
+	game.StartChallenger(ctx, sys.RollupConfig, sys.L2GenesisCfg, l1Endpoint, l2Endpoint, "Honest",
+		// Disagree with the proposed output, so agree with the (correct) root claim
```
Inphi: should be "Agree with the proposed output, so disagree with the (correct) root claim"

Inphi (approved):

PR #7259, opened 2023-09-15
Update Span batch specs to allow overlapping batches

ImTei:
**Context**
- Current span batch spec does not allow span batches that overlaps current safe chain.
- So the op-node cannot proceed derivation if the pipeline is reset to the middle of span batch.
- It's very hard to guarantee that the op-node will always reset the safe head to the boundary of the span batch.
- Therefore we need to allow overlapping span batches.

**Changes**
- Specs
    - Updated span batch validation rules to allow overlapping batches


specs/span-batches.md
```diff
@@ -220,7 +222,7 @@ The assumption makes upper inequality to hold. Therefore, we decided to manage `
   - If the last block references canonical L1 chain as its origin,
     we can ensure the all other blocks' origins are consistent with the canonical L1 chain.
 - Parent hash
-  - In V0 Batch spec, we need batch's parent hash to validate if batch's parent is consistent with current L2 safe head.
+  - In V0 Batch spec, we need batch's parent hash to validate if batch's parent is consistent with current L2 chain.
```
trianglesphere: Maybe mention "safe portion of the chain". I guess it wouldn't really happen that we would be asking for a span batch that started at some timestamp where it started at some point in the unsafe chain.

ImTei: This seems to be my mistake. Since this sentence is about batch v0 (singular batch), it would be correct to say `safe head`.

ImTei:
@protolambda Please review this PR when you get a chance

specs/span-batches.md
```diff
@@ -247,14 +249,21 @@ Span-batches share the same queue with v0 batches: batches are processed in L1 i
...
 
 Span-batch rules, in validation order:
 
-- `batch.start_timestamp > next_timestamp` -> `future`: i.e. the batch must be ready to process.
-- `batch.start_timestamp < next_timestamp` -> `drop`: i.e. the batch must not be too old.
-- `batch.parent_check != safe_l2_head.hash[:20]` -> `drop`: i.e. the checked part of the parent hash must be equal
-  to the L2 safe head block hash.
+- `span_start.timestamp > next_timestamp` -> `future`: i.e. the batch must be ready to process.
+- `span_end.timestamp < next_timestamp` -> `drop`: i.e. the batch must have at least one new block.
+- If there's no `prev_l2_block` in the current safe chain -> `drop`: i.e. misaligned timestamp
```
trianglesphere: Why did we add this check?

specs/span-batches.md
```diff
@@ -304,6 +312,17 @@ Span-batch rules, in validation order:
     that is invalid or derived by other means exclusively:
     - any transaction that is empty (zero length `tx_data`)
     - any [deposited transactions][g-deposit-tx-type] (identified by the transaction type prefix byte in `tx_data`)
+- Overlapped blocks checks:
```
trianglesphere, line 344: Is this implemented?

trianglesphere (commented):
This overlapped case is pretty weird, but we need to pay very careful attention to it.

Starting from normal operation, my assumption is that when we have a span batch which contains first a string of valid batches & then an invalid batch & then valid batches again, we will process all the valid batches, reject the invalid batch, and then wait for a sequence window timeout before discarding the span batch & processing more.

Looking at the implementation, we actually just pop all of the batches out of the span batch regardless of if a batch was successfully applied or not. It will get to the attributes queue & will reset the entire pipeline.

specs/span-batches.md
ImTei: e.g. If block timestamps of the current safe chain are ...10, 12, 14, and the `span_start.timestamp` is 15, then we should find the `prev_l2_block` at timestamp 13 (15 - block time), but the block does not exist. so we have to drop this batch.

This new rule is to ensure that the timestamp of the span batch is well-aligned with the existing chain and block time.

specs/span-batches.md
ImTei: yes. you can find this code in [the derivation PR](https://github.com/ethereum-optimism/optimism/pull/7621/files#diff-95e3cbbeed2bde6b84fb4510e4735f6f5fde5df97cec98f25b14d413005ce4d8R328)

ImTei:
@trianglesphere Thanks for the comment. It seems to be a very critical point. 
so I understand that we need to consider the atomicity of the span-batch. This is a different topic than overlapping batch. Is it correct?
- As you said, we don't check if the previous singular batch is applied successfully and just pop the next batch in the current [implementation of span batch derivation](https://github.com/ethereum-optimism/optimism/pull/7621/files#diff-70e161ad6625124a29f49468d591c738ddec2e962fcc198d3a9db85b242c1096R91-R97).
- We must check that. This can be done simply by comparing the given safe head's block hash and the cached batch's parent hash.
- And if some singular batch derived from the span batch fails to be applied, we have to discard the entire span batch. It means that we need to reorg the chain to the state before the span batch.
- This reorg does not mean resetting the entire pipeline. we should just reset the safe head and keep the rest states.

@protolambda what do you think?

specs/span-batches.md
trianglesphere: hmm, I think we use `span_start.timestamp > next_timestamp` and `span_end.timestamp < next_timestamp` to half enforce this as well. If neither of those are true, then `span_start.timestamp == next_timestamp`

trianglesphere:
@ImTei, yep, we do need to consider the atomicity. I would also take a step back & think about what behavior we want if there is an invalid batch inside a span batch. The other thing is that we need to think about this in the context of re-orgs & also other singular batches which have been included.

specs/span-batches.md
ImTei: The second condition is comparing `span_end` instead of `span_start`. It might be a bit confusing 😅

ImTei:
@trianglesphere @protolambda 

I have specified the problematic scenario and what to implement. I'll implement those things in the next follow-up PR. let me know if there are any concerns or suggestions! :)

- Even if the span batch passed the batch queue’s validity checks, a singular batch from the span batch might fail to be applied to the safe chain due to the BlockInsertPayloadErr.
- In this case, we suppress the error and retry the pipeline with the next batch from the batch queue.
- If this case occurs in the middle of a span batch, i.e. we already applied some singular batches from the span batch to the engine, we must revert them for the atomicity of span batches.
- Therefore, we need the following behaviors when BlockInsertPayloadErr occurs during processing span batches:
  - Engine queue must trigger safe chain reorg by resetting its safe head
  - Batch queue must reset cached singular batches derived from the reverted span batch.
- And this reorg must be possible, even if we’re deriving the already finalized blocks. So we have to finalize the L2 blocks when an entire span batch is successfully applied to the chain.
  - To achieve this, the pipeline should pass the signal to indicate whether the current processing singular batch is the last one from the span batch.
  - And the engine queue must consider this signal to schedule the L2 finalization, and to use it for the reorg target block.

specs/span-batches.md
```diff
@@ -304,6 +312,17 @@ Span-batch rules, in validation order:
...
+  - Note: If the span batch overlaps the current L2 safe chain, we must validate all overlapped blocks.
+  - Variables:
+    - `block_input`: an L2 block derived from the span-batch.
+    - `safe_block`: an L2 block from the current L2 safe chain, at same timestamp as `block_input`
+  - Rules:
+    - For each `block_input`, whose timestamp is less than `next_timestamp`:
+      - If there's no `safe_block` for the `block_input` -> `drop`: i.e. misaligned timestamp
```
ajsutton: We don't seem to implement this check in https://github.com/ethereum-optimism/optimism/pull/7621/ but if the prev l2 block exists then the first block must have its timestamp aligned and every other block timestamp is derived by adding the block time so they must also be aligned.

ImTei: @ajsutton you're right. I think this rule is redundant and we can remove it from specs. Thank you!

ajsutton:
I commented in https://github.com/ethereum-optimism/optimism/pull/7621 about the atomicity issues being detected here.  If we're delaying the update of the finalized block until the end of the span batch I think we should also consider delaying the update of safe head - that's another step towards actually applying span batches atomically. We'd still have changes to the l1 origin and things like batch queue `l1Blocks` which are hard to roll back though.

Whichever we go technically, I'd encourage having a really strong focus on simplifying things so rather than trying to handle interesting corner cases like span batches being partially applied, we try to find approaches that reduce the number of corner cases and simplify the way we need to reason about the protocol.

protolambda (approved):

PR #7261, opened 2023-09-15
fix(op-challenger): Don't schedule updates for resolved games

refcell:
**Description**

Prevents the `Coordinator` from queueing job updates for resolved games.

Scheduling these games for an update won't do anything since the `ProgressGame` call will just return the resolved status and not do anything.

This makes the inflight games metric more robust since we don't mark these resolved games as inflight.

**Metadata**

Fixes https://github.com/ethereum-optimism/client-pod/issues/58


refcell:
* develop
  * **PR #7261** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7261" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7261?utm_source=stack-comment).

op-challenger/game/scheduler/coordinator.go
```diff
@@ -109,9 +114,12 @@ func (c *coordinator) createJob(game common.Address) (*job, error) {
 			return nil, fmt.Errorf("failed to create game player: %w", err)
 		}
 		state.player = player
+		state.status = player.Status()
 	}
 	state.inflight = true
-	return &job{addr: game, player: state.player}, nil
```
trianglesphere, line 114: Why don't we still return `&job{...}` here? We don't use it or modify it beyond setting fields.

trianglesphere (commented):
Tests & core changes look good

op-challenger/game/scheduler/coordinator.go
```diff
@@ -64,8 +64,12 @@ func (c *coordinator) schedule(ctx context.Context, games []common.Address) erro
 		if j, err := c.createJob(addr); err != nil {
 			errs = append(errs, err)
 		} else if j != nil {
-			jobs = append(jobs, *j)
-			c.m.RecordGameUpdateScheduled()
+			if j.status == types.GameStatusInProgress {
```
ajsutton: We should be doing this filtering in `createJob` and just not returning a job to schedule.

op-challenger/game/scheduler/coordinator.go
```diff
@@ -64,8 +64,12 @@ func (c *coordinator) schedule(ctx context.Context, games []common.Address) erro
...
+				jobs = append(jobs, *j)
+				c.m.RecordGameUpdateScheduled()
+			} else {
+				c.logger.Warn("Resolved game update not scheduled", "game", addr, "status", j.status)
```
ajsutton: This is normal and very common so probably should just be debug level. Definitely not warn.

op-challenger/game/scheduler/coordinator.go
refcell: Ah good catch!

trianglesphere (approved):

PR #7262, opened 2023-09-15
feat(op-challenger): Concurrency Metrics

refcell:
**Description**

Adds concurrency metrics around the worker `progressGames` goroutine'd function.

**Tests**

Unit tests are done via the `worker_test.go` tests to ensure the active and idle callbacks are called correctly by the worker.

**Metadata**

Fixes https://github.com/ethereum-optimism/client-pod/issues/57


refcell:
* develop
  * **PR #7262** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7262" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7262?utm_source=stack-comment).

op-challenger/game/scheduler/scheduler.go
```diff
@@ -48,13 +55,28 @@ func NewScheduler(logger log.Logger, m SchedulerMetricer, disk DiskManager, maxC
 	}
 }
 
+func (s *Scheduler) ThreadActive() {
+	s.executorMutex.Lock()
+	defer s.executorMutex.Unlock()
+	s.m.DecIdleExecutors()
+	s.m.IncActiveExecutors()
```
ajsutton: You shouldn't need a mutex here - metrics are thread-safe for updates, and the lock won't stop the values being sampled between these two calls anyway.  Just do the increment before the decrement in both cases so its impossible to wind up with a negative value in the gauge - although I think prometheus is ok with that anyway.

trianglesphere (approved):

PR #7264, opened 2023-09-15
feat(op-challenger): Claim Count Metrics

refcell:
**Description**

Adds claim count metrics to the `op-challenger` `agent` component.

This tracks the number of claims tracked locally for a given game (labeled in the prometheus gauge vec by game address).

Whenever the Agent is "acted" upon by the player's progress games function, it constructs a new game state from the contracts.

When this new game state is constructed, the claim count needs to be fetched from the contracts. At this point,
the claim count metric is updated with this fetched count.

When the game is no longer in progress, or is ready to be resolved, the claim count metric is zeroed out by the agent
in order to avoid monotonically flooding the metrics with ever increasing game claim counts.

Further, a grafana dashboard can use this metric to get insight into "what the challenger sees" wrt claims by filtering
out the games with claim counts of zero.


refcell:
* develop
  * **PR #7264** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7264" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7264?utm_source=stack-comment).

op-challenger/game/fault/agent.go
```diff
@@ -52,6 +54,8 @@ func NewAgent(m metrics.Metricer, loader ClaimLoader, maxDepth int, trace types.
 // Act iterates the game & performs all of the next actions.
 func (a *Agent) Act(ctx context.Context) error {
 	if a.tryResolve(ctx) {
+		// Zero out the game claim count since the game will no longer be tracked once resolved.
+		a.metrics.RecordGameClaimCount(a.fdgAddr.String(), 0)
```
trianglesphere: TBH there's no big reason to do this.

op-challenger/metrics/metrics.go
```diff
@@ -93,6 +97,13 @@ func NewMetrics() *Metrics {
 				[]float64{1.0, 10.0},
 				prometheus.ExponentialBuckets(30.0, 2.0, 14)...),
 		}),
+		gameClaimCount: *factory.NewGaugeVec(prometheus.GaugeOpts{
```
trianglesphere, line 114: If it weren't for the fact that L1 can reorg & claims could disappear, this would be better as a counter (always increasing) rather than a gauge, but because the number can go down, we need to use it.

trianglesphere (approved):

PR #7272, opened 2023-09-15
feat(op-challenger): Output Trace Provider

refcell:
**Description**

Introduces the Output Trace Provider which uses Output Roots for L2 Blocks as the given trace.

**Tests**

Unit tests mirror other trace providers and verify that `GetStepData` and `AbsolutePreState` Trace Provider methods error.

**Metadata**

Implements part of https://github.com/ethereum-optimism/client-pod/issues/41

### DIFF
--- a/op-challenger/game/fault/trace/outputs/provider.go
+++ b/op-challenger/game/fault/trace/outputs/provider.go
@@ -63,7 +63,6 @@ func (o *OutputTraceProvider) Get(ctx context.Context, traceIndex uint64) (commo
 }
 
 // AbsolutePreStateCommitment returns the absolute prestate at the configured prestateBlock.
-// For Optimism Mainnet, this is the first block after bedrock.
 func (o *OutputTraceProvider) AbsolutePreStateCommitment(ctx context.Context) (hash common.Hash, err error) {
 	output, err := o.rollupClient.OutputAtBlock(ctx, o.prestateBlock)
 	if err != nil {
--- a/op-challenger/game/fault/trace/outputs/provider.go
+++ b/op-challenger/game/fault/trace/outputs/provider.go
@@ -1,0 +1,81 @@
+package outputs
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ethereum-optimism/optimism/op-challenger/game/fault/types"
+	"github.com/ethereum-optimism/optimism/op-service/client"
+	"github.com/ethereum-optimism/optimism/op-service/eth"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+var (
+	GetStepDataErr      = fmt.Errorf("GetStepData not supported")
+	AbsolutePreStateErr = fmt.Errorf("AbsolutePreState not supported")
+	PreStateRequestErr  = fmt.Errorf("Requested trace index is before prestate block")
+)
+
+type OutputRollupClient interface {
+	OutputAtBlock(ctx context.Context, blockNum uint64) (*eth.OutputResponse, error)
+}
+
+// OutputTraceProvider is a [types.TraceProvider] implementation that uses
+// output roots for given L2 Blocks as a trace.
+type OutputTraceProvider struct {
+	logger        log.Logger
+	rollupClient  OutputRollupClient
+	prestateBlock uint64
+}
+
+func NewTraceProvider(ctx context.Context, logger log.Logger, rollupRpc string, prestateBlock uint64) (*OutputTraceProvider, error) {
+	rollupClient, err := client.DialRollupClientWithTimeout(client.DefaultDialTimeout, logger, rollupRpc)
+	if err != nil {
+		return nil, err
+	}
+	return NewTraceProviderFromInputs(logger, rollupClient, prestateBlock), nil
+}
+
+func NewTraceProviderFromInputs(logger log.Logger, rollupClient OutputRollupClient, prestateBlock uint64) *OutputTraceProvider {
+	return &OutputTraceProvider{
+		logger:        logger,
+		rollupClient:  rollupClient,
+		prestateBlock: prestateBlock,
+	}
+}
+
+func (o *OutputTraceProvider) Get(ctx context.Context, traceIndex uint64) (common.Hash, error) {
+	outputBlock := traceIndex + 1
+	if outputBlock < o.prestateBlock {
+		return common.Hash{}, PreStateRequestErr
+	}
+	output, err := o.rollupClient.OutputAtBlock(ctx, outputBlock)
+	if err != nil {
+		o.logger.Error("Failed to fetch output", "blockNumber", outputBlock, "err", err)
+		return common.Hash{}, err
+	}
+	return common.Hash(output.OutputRoot), nil
+}
+
+// AbsolutePreStateCommitment returns the absolute prestate at the configured prestateBlock.
+// For Optimism Mainnet, this is the first block after bedrock.
+func (o *OutputTraceProvider) AbsolutePreStateCommitment(ctx context.Context) (hash common.Hash, err error) {
+	output, err := o.rollupClient.OutputAtBlock(ctx, o.prestateBlock)
+	if err != nil {
+		o.logger.Error("Failed to fetch output", "blockNumber", o.prestateBlock, "err", err)
+		return common.Hash{}, err
+	}
+	return common.Hash(output.OutputRoot), nil
+}
+
+// AbsolutePreState is not supported in the [OutputTraceProvider].
+func (o *OutputTraceProvider) AbsolutePreState(ctx context.Context) (preimage []byte, err error) {
+	return nil, AbsolutePreStateErr
+}
+
+// GetStepData is not supported in the [OutputTraceProvider].
+func (o *OutputTraceProvider) GetStepData(ctx context.Context, i uint64) (prestate []byte, proofData []byte, preimageData *types.PreimageOracleData, err error) {
+	return nil, nil, nil, GetStepDataErr
+}
--- a/op-challenger/game/fault/trace/outputs/provider.go
+++ b/op-challenger/game/fault/trace/outputs/provider.go
@@ -15,7 +15,6 @@ import (
 var (
 	GetStepDataErr      = fmt.Errorf("GetStepData not supported")
 	AbsolutePreStateErr = fmt.Errorf("AbsolutePreState not supported")
-	PreStateRequestErr  = fmt.Errorf("Requested trace index is before prestate block")
 )
 
 var _ types.TraceProvider = (*OutputTraceProvider)(nil)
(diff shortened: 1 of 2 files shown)


refcell:
* develop
  * **PR #7272** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7272" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7272?utm_source=stack-comment).

op-challenger/game/fault/trace/outputs/provider.go
```diff
@@ -0,0 +1,81 @@
...
+
+func NewTraceProviderFromInputs(logger log.Logger, rollupClient OutputRollupClient, prestateBlock uint64) *OutputTraceProvider {
+	return &OutputTraceProvider{
+		logger:        logger,
+		rollupClient:  rollupClient,
+		prestateBlock: prestateBlock,
+	}
+}
+
+func (o *OutputTraceProvider) Get(ctx context.Context, traceIndex uint64) (common.Hash, error) {
```
trianglesphere: I think we should be shifting `traceIndex` by the `prestateBlock`. `traceIndex` 0 corresponds to `prestateBlock` and we also need to track the `postState` block because when `traceIndex > postStateBlockNumber - preStateBlockNumber` we need to return `postStateBlock`, not `traceIndex + preStateBlockNumber`'s block

op-challenger/game/fault/trace/outputs/provider.go
```diff
@@ -0,0 +1,81 @@
...
+	output, err := o.rollupClient.OutputAtBlock(ctx, outputBlock)
+	if err != nil {
+		o.logger.Error("Failed to fetch output", "blockNumber", outputBlock, "err", err)
+		return common.Hash{}, err
+	}
+	return common.Hash(output.OutputRoot), nil
+}
+
+// AbsolutePreStateCommitment returns the absolute prestate at the configured prestateBlock.
+// For Optimism Mainnet, this is the first block after bedrock.
```
trianglesphere: Have we decided on always bisecting to genesis?

op-challenger/game/fault/trace/outputs/provider.go
```diff
@@ -0,0 +1,81 @@
...
+	PreStateRequestErr  = fmt.Errorf("Requested trace index is before prestate block")
+)
+
+type OutputRollupClient interface {
+	OutputAtBlock(ctx context.Context, blockNum uint64) (*eth.OutputResponse, error)
+}
+
+// OutputTraceProvider is a [types.TraceProvider] implementation that uses
+// output roots for given L2 Blocks as a trace.
+type OutputTraceProvider struct {
```
trianglesphere, line 28: One nice to have is the assertion that `OutputTraceProvider` satisfies the `TraceProvider` interface. You can do this by adding the following line: `var _ TraceProvider = (*OutputTraceProvider)(nil)` which will cause a compilation failure if `OutputTraceProvider` doesn't implement `TraveProvider`.

op-challenger/game/fault/trace/outputs/provider.go
```diff
@@ -0,0 +1,81 @@
...
+func (o *OutputTraceProvider) AbsolutePreStateCommitment(ctx context.Context) (hash common.Hash, err error) {
+	output, err := o.rollupClient.OutputAtBlock(ctx, o.prestateBlock)
+	if err != nil {
+		o.logger.Error("Failed to fetch output", "blockNumber", o.prestateBlock, "err", err)
+		return common.Hash{}, err
+	}
+	return common.Hash(output.OutputRoot), nil
+}
+
+// AbsolutePreState is not supported in the [OutputTraceProvider].
```
trianglesphere, line 77: Why not? We can return a serialized output. Do we not have a getter for it?
We will need this.

op-challenger/game/fault/trace/outputs/provider.go
```diff
@@ -0,0 +1,81 @@
...
+func (o *OutputTraceProvider) AbsolutePreState(ctx context.Context) (preimage []byte, err error) {
+	return nil, AbsolutePreStateErr
+}
+
+// GetStepData is not supported in the [OutputTraceProvider].
+func (o *OutputTraceProvider) GetStepData(ctx context.Context, i uint64) (prestate []byte, proofData []byte, preimageData *types.PreimageOracleData, err error) {
```
trianglesphere: This one probably can be left as is, but it would be nice to have more of the design doc on how the switch is being implemented done.

op-challenger/game/fault/trace/outputs/provider.go
refcell: Ah tysm, great callout!

op-challenger/game/fault/trace/outputs/provider.go
refcell: My understanding is that an output bisection game of depth 63 would _by far_ sufficiently cover blocks from genesis. @clabby had told me it should cover blocks until heat death I believe haha.

But anyways, I'm pretty strictly following what Adrian had outlined in https://github.com/ethereum-optimism/client-pod/issues/41. That way I wouldn't surprise him when he gets back :)

op-challenger/game/fault/trace/outputs/provider.go
refcell: I was strictly following Adrian's outline in https://github.com/ethereum-optimism/client-pod/issues/41

> AbsolutePreState will also always return an error since it is only used for step calls.

refcell: To expand, my understanding is that there won't be any step calls performed at the end of the output bisection. cc @clabby

op-challenger/game/fault/trace/outputs/provider.go
refcell: Yep agreed, I'm just following https://github.com/ethereum-optimism/client-pod/issues/41

> GetStepData will always return an error as step can't be called for output roots.

op-challenger/game/fault/trace/outputs/provider.go
refcell: Good point, thanks!

op-challenger/game/fault/trace/outputs/provider.go
clabby: No step calls performed at the end of output bisection in the current design; Though things are still somewhat up in the air.

I imagine that the transition will be a regular move, with the claim hash being the root claim for execution trace bisection. Steps will occur at the very bottom of the composed bisections.

op-challenger/game/fault/trace/outputs/provider.go
clabby: Yeah an output bisection game over a depth of 63 should cover us until heat death, but it's way overkill to have a tree that large for the outputs. The tentative plan is to split the tree of d=63, There is also a question of where exactly we split it.

If we're only bisecting between the previous finalized output root and the current one, we don't need much depth on that portion of the bisection at all, but if we are going all the way from genesis, we'll need a depth with enough padding to support us for a long time going forwards. For the instruction trace bisection, assuming we're only executing a single block, a depth of 40 supporting $2^{40}$ instructions should be sufficient w/ some extra padding, which would imply the tree needs to be deeper than 63 if we're bisecting from genesis, since the top layer of bisection could only support 8,388,608 blocks. 

Will look something like this:
```
                   o
                 /    \
                o      o
               / \    / \
 ---------------- [SPLIT] ----------------
             o     o o    o
.......................................... and so on
```

clabby: Assuming we go from the previous finalized output, the output bisection tree depth also places an implicit max on the proposal submission interval; Should be concious of that while designing and consider OP Stack chains with sparse proposals. Doesn't seem like a reasonable bet to assume that the proposal submission interval is small everywhere.

ajsutton: Yeah I thought we went with bisecting from genesis so that we aren't forced to post output roots at any regular interval. @karlfloersch had pretty strong feelings about it.  It may be ok if we can start from an arbitrarily old previous finalized output but I'm not sure that buys us anything since we'd have to allow for it to be really stupid old anyway and so still have about the same depth as we would from genesis.

Benefit of not starting from genesis though is that you then don't need an archive node.  But we need an archive node to create the cannon traces so it's not necessarily a win.

op-challenger/game/fault/trace/outputs/provider.go
ajsutton: Yeah step call doesn't make sense for the output root bisection part - there's no way we can apply an entire L2 block on chain to compute the correct next output root. If we wind up having to move to separate games for output root and cannon trace that may change but then things will get a fair bit more complex so lets assume it won't happen for now and we can update if need be in the future.

op-challenger/game/fault/trace/outputs/provider.go
ajsutton: Working on the simple assumption that the first X levels are output roots and when you get to X+1 you just switch to cannon traces.  The contracts will have to have the appropriate assertion that first cannon trace has the right status code (either that the claim is or isn't valid depending on the depth the switch happens at) but the challenger shouldn't need to know anything about that because the honest challenger should always have the appropriate status code in its cannon trace.

The actual code to switch from one trace provider to the other is https://github.com/ethereum-optimism/client-pod/issues/43. Regardless of how the switch works, we know we need a trace provider that can give us output roots for claims which is what we're adding here. If it turns out to also need to do something with step, we can add that later when we know its needed (though I don't expect it to be).

op-challenger/game/fault/trace/outputs/provider_test.go
```diff
@@ -0,0 +1,116 @@
...
+var (
+	prestateBlock       = uint64(100)
+	poststateBlock      = uint64(200)
+	prestateOutputRoot  = common.HexToHash("0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
+	firstOutputRoot     = common.HexToHash("0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb")
+	poststateOutputRoot = common.HexToHash("0xcccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc")
+)
+
+func TestGet(t *testing.T) {
+	t.Run("FirstBlockAfterPrestate", func(t *testing.T) {
```
ajsutton, line 39: Probably should test requests for before the prestate block too. Probably just returns an error.

op-challenger/game/fault/trace/outputs/provider.go
ajsutton: I think the code here works the same regardless of where we start the bisection anyway given it's just given a prestate block number, that could either be the bedrock activation block or the block the last finalized output root is for and it just works.

I'd probably just remove the reference to OP Mainnet here and leave the trace provider to be completely agnostic to how the start block is chosen.

ajsutton (approved):
LGTM. Just one suggested test to add.

ajsutton:
Worth noting that this PR does not complete the issue though. Specifically:
> Should be able to write e2e tests at this point which play through the game using a cannon game type on-chain and just verify that the claims posted have the correct output root.

We can't yet do that and don't yet have such an e2e test.  So the description should reference the issue, but not say `fixes` to avoid closing the issue when the PR is merged.

op-challenger/game/fault/trace/outputs/provider_test.go
refcell: Added tests for that by "misconfiguring" the prestate block number to `0`. Otherwise no way to query before the prestate since the parameter is incremented by the prestate block number plus 1.

ajsutton: Oh that's my bad - I totally missed that point.  You don't really need that test then, sorry.

trianglesphere:
> We can't yet do that and don't yet have such an e2e test. So the description should reference the issue, but not say fixes to avoid closing the issue when the PR is merged.


So fun fact, you can directly say "does not fix" and github will still close the issue. Any time you reference an issue in a PR description, github will close it if the PR is merged.

ajsutton:
> So fun fact, you can directly say "does not fix" and github will still close the issue. Any time you reference an issue in a PR description, github will close it if the PR is merged.

Not so - the key phrase is either "fix <number>" or "fixed <number>" so "does not fix <number>" matches and closes the issue.  If you just link the issue without the word "fix" or "fixed" (or "closes" and possibly some other variants) it will not be closed.  I do it all the time.

PR #7275, opened 2023-09-15
proxyd: avoid banning a backend that is a forced candidate

mdehoog:
**Description**

We run a bunch of nodes behind a load balancer. Because of this, sometimes a node that is slightly behind will return a block older than the one the consensus knows about, and proxyd will ban the backend.

This PR avoids banning any backends that have `consensus_forced_candidate` enabled.

Also fixes an unchecked cast that is causing panics.

**Tests**

Tested on our goerli testnet.


refcell (approved):
LGTM

mslipper (approved):

felipe-op:
@mdehoog you may want to use `consensus_max_block_lag` instead
https://github.com/ethereum-optimism/optimism/blob/adb9595842279f4420928ac95cf92cf4fbec8745/proxyd/config.go#L110

felipe-op:
disabling the ban logic is bad, since proxyd will try to fulfill a request from a backend that is known to be degraded/out of service

trianglesphere:
Does proxyd unban nodes after they are back into consensus?

felipe-op:
> Does proxyd unban nodes after they are back into consensus?

yes, automatically

trianglesphere:
> > Does proxyd unban nodes after they are back into consensus?
> 
> yes, automatically

I see the `IsBanned` function does only last a certain period of time: https://github.com/ethereum-optimism/optimism/blob/61871cf471d3adcbabb3218af20660e2869dd8ff/proxyd/consensus_poller.go#L61-L63

It seems the default ban period is 5 minutes: https://github.com/ethereum-optimism/optimism/blob/61871cf471d3adcbabb3218af20660e2869dd8ff/proxyd/consensus_poller.go#L229, but can be overridden: https://github.com/ethereum-optimism/optimism/blob/61871cf471d3adcbabb3218af20660e2869dd8ff/proxyd/consensus_poller.go#L188-L192

Would applying a ban period of 0s do the same as disabling it? Usually if you pass a zero duration to `time` function it's a no-op. It doesn't seem document, but would be nice to reuse the existing functionality. Also @mdehoog, the lag behavior might be closer to what you want, but you could also do a short ban period.

felipe-op:
> Would applying a ban period of 0s do the same as disabling it? Usually if you pass a zero duration to `time` function it's a no-op. It doesn't seem document, but would be nice to reuse the existing functionality. Also @mdehoog, the lag behavior might be closer to what you want, but you could also do a short ban period.

Yes, making it 0s will effectively "unban" right away.

felipe-op:
@mdehoog  I see you've recently introduced also a config to "force to include backend in consensus", which carry a similar bad pattern.

If you force to include the backend in the consensus, you'll maybe be serving traffic from a backend which is not the best option.

I recommend using a backend group with at least 3 backends, distributed in different azs to be resilient against az network/datacenter failure and load balance traffic between them for scalability

mdehoog:
@felipe-op unfortunately both of those options don't work for us.

This is the logic that is banning our nodes: https://github.com/ethereum-optimism/optimism/blob/61871cf471d3adcbabb3218af20660e2869dd8ff/proxyd/consensus_poller.go#L343-L355

This is because sometimes a node will return a safe or finalized before the previous request because it is slightly behind. The `consensus_max_block_lag` option does not appear to be checked in that logic.

If we set the ban period to 0, we still get the `"backend banned - unexpected block tags"` log.

Also banning backends set their cached block numbers to 0: https://github.com/ethereum-optimism/optimism/blob/61871cf471d3adcbabb3218af20660e2869dd8ff/proxyd/consensus_poller.go#L499-L501
This means we start serving incorrect responses until the next polling period, because the tag replacer doesn't know which block numbers to replace with.

felipe-op:
> This is because sometimes a node will return a safe or finalized before the previous request because it is slightly behind.

This logic is comparing the progression for the same backend.  It does not mean it is "behind", it means the backend itself regressed (probably due to a chain re-org).  It will only consider the block tags as unexpected if the same backend goes back in block tag number, or if the ordering is incorrect.


> Also banning backends set their cached block numbers to 0

Yes, we "forget" the last block tag to allow the backend to start from scratch once it is unbanned. Same behaviour when proxyd is just starting and discovering the initial backend state.


___

@mdehoog  Can you provide some log lines showing the backend block tag progression (like 5 last state updates + what triggered the ban) ? Let's try to find a good solution together.

mdehoog:
For us, the "same backend" can return inconsistent results. We run nodes behind an AWS ALB, and each of these nodes sync separately. So safe / finalized can occasionally appear to go backwards.

The logs are simply `"backend banned - unexpected block tags"`, and then proxyd will start serving the genesis block for all tagged block requests until the next tags are polled by the consensus poller (because we use `consensus_forced_candidate` on this backend). Unfortunately I no longer have the logs from the initial rollout where we were seeing the errors due to our retention policies (we've been running this fork in production for a few weeks).

Also worth noting that we only use the consensus mode so we can opt-in to the block range filtering.

mdehoog:
One option might be to remove the `consensus_forced_candidate` option added in https://github.com/ethereum-optimism/optimism/pull/6686 as part of this PR, as the `consensus_never_ban` is essentially meant to solve the same issue: ensuring that certain backends are always treated as candidates.

Or alternatively we could attach the "never ban" functionality to the `consensus_forced_candidate` option.

felipe-op:
Gotcha.  Can you point directly to your backend nodes, instead of the load balancer? Proxyd is supposed to act as the load balancer, adding the consensus layer on top of it exactly to minimize re-org experienced by clients.

If you think this is not appropriate for your situation/infra, I'm totally fine merging the bypass with a single flag extending the behaviour of `consensus_forced_candidate` but please note you'll be losing an important benefit provided by proxyd.

mdehoog:
Thank you for the thoughtful response, aligned that an additional overlapping configuration option doesn't make sense.

I agree that proxyd makes more sense as a load balancer across nodes. We don't currently have that luxury due to the way our infra is set up, so our proxyd is primarily used for rate limiting and bad-request-filtering, rather than load balancing.

Just pushed a change that removes the additional option and instead avoids the ban using the `consensus_forced_candidate` option instead.

felipe-op (approved):

PR #7279, opened 2023-09-17
Deploy Delayed Vetoable - CI test only

maurelian:


packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -105,9 +108,6 @@ contract Deploy is Deployer {
 
     /// @notice Deploy all of the proxies
     function deployProxies() public {
```
tynes, line 110: Ideally we can call a single function and have everything set up for a new network joining the superchain. This is not possible if you move `deployAddressManager` outside of `deployProxies`. We either need to make `deployProxies` the entrypoint to adding a new chain to the superchain or we wrap something around `deployProxies`

maurelian:
Replaced by #7292, which adds the Safe before adding the delayed vetoable to deployments.

PR #7281, opened 2023-09-18
cannon: proof of concept multi-threading support

protolambda:
**Description**

This is a proof of concept that shows how we can implement a minimal version of futexes and thread scheduling to handle concurrent programs.

This enables Cannon to run Go programs with GC and other concurrency like mutexes, channels, waitgroups, etc.

Cannon still proves only a single sequential execution trace, but as part of the syscalls and instruction execution it can now rotate between threads to support async functionality of programs deterministically.

The proof of concept:
- changes the state format to separate a "thread context" from the global VM state
- updates the offchain MIPS emulator code to support several new syscalls:
  - futex: with wait & wake ops, to support the locking functionalities requird by the Go runtime
    - wake prioritizes finding a thread to test unlock that is stuck on the requested memory address
    - wait parks a thread and continues with the next
  - clone: to create threads
  - exit: to shut down threads
  - yield and nanosleep: simply makes the scheduler look at the next thread
  - several no-op syscalls, to cover new runtime behavior that can be ignored (hints / performance / time things)

This does not yet update the witness-encoding or the onchain emulator part.

For onchain functionality we need to add a fixed-size proof for the current thread in the threads-list (as binary merkle tree), and an optional second proof to help append a new thread upon clone operation. And then just mirror the Go logic for state changes.

**Tests**

Updated the "hello" Go program to include some concurrent things in Go:
- waitgroup
- non-blocking and blocking channels
- go routines
- GC

And I had to disable the EVM <> Go tests for now, as I have not updated the solidity implementation or applied the state format chnages to the EVM Go tests.

**Additional context**

For now just an experiment, but for longer heavier programs making this production ready may be worth the benefits of GC.


protolambda:
Keeping this PR open just to preserve the branch at this point, so I can look at it when experimenting with Cannon/Asterisc in new domains, soon hopefully, once higher priority work clears up :sweat_smile: .

refcell:
Hey @protolambda, tacking on the https://github.com/ethereum-optimism/optimism/labels/S-exempt-stale label on here so you don't need to constantly remove https://github.com/ethereum-optimism/optimism/labels/Stale

protolambda:
One key insight that I had after the initial draft, and which I am not sure I put put there on github, is that it can be implemented without merkleizing a list of thread contexts

The trick is to not go 0, 1, 2, 3, 0, 1, 2, 3, 0, 1, 2, 3 etc. through the threads, but instead 0, 1, 2, 3, 2, 1, 0, 1, 2, 3, 2, ....

By going back and forth in the list of thread-contexts, you can instead just maintain two hash-onions, much simpler to prove, just a single bytes32 witness necessary.

For multi-threading, it doesn't really matter all that much which order the thread continuations are attempted in, as there's no guarantee in an actually parallel environment anyway. It just needs to be deterministic

So by maintaining two "stacks" of thread contexts, committed to as hash-onions, you can get a very simple/elegant way of traversing all that thread data.

And, bonus: adding/removing threads is super simple: you navigate to the thread with the given ID, and then not push it back on either the left or right stack, to remove it. Or, if creating a thread, just add it to either stack, by hashing the new thread with the existing value, and then overwriting the existing value with that

cannon/mipsevm/mips.go
```diff
@@ -6,14 +6,64 @@ import (
...
+	sysEpollPwait = 4313
+	sysGetRandom  = 4353
+	sysUname      = 4122
+	sysStat64     = 4213
+	sysGetuid     = 4024
+	sysGetgid     = 4047
+	sysLlseek     = 4140
+)
+
+// TODO:
```
mbaxter, line 55: From: https://github.com/golang/go/blob/fe87b586c036ae7ad25e22896cb90f95421545fe/src/runtime/sys_linux_mipsx.s#L39-L43

tynes:
heroic

PR #7286, opened 2023-09-18
Support old Make version by avoiding SHELLSTATUS

karlb:
SHELLSTATUS is only available on GNU Make 4.2 (released on 2016-05-22) and later, which is unfortunatly not installed on MacOS by default.

While I am highly annoyed that Apple is unwilling to update their Make version and don't want encourage such behaviour, the change in this commit avoids this requirement and seems even simpler than the original code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,10 +89,8 @@ devnet-up:
 	@if [ ! -e op-program/bin ]; then \
 		make cannon-prestate; \
 	fi
-	$(shell ./ops/scripts/newer-file.sh .devnet/allocs-l1.json ./packages/contracts-bedrock)
-	if [ $(.SHELLSTATUS) -ne 0 ]; then \
-		make devnet-allocs; \
-	fi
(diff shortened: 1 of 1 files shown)


refcell (approved):
Thank you for upstreaming this change!

refcell:
Hey @karlb, I believe circleci status github integration flaked and is stuck waiting for an update from the `contracts-bedrock-slither` job. Can you try to retrigger the job on circleci's dashboard or just rebase this branch and push to retrigger ci?

karlb:
I now get a reproducible CI error:

`ERROR: gcloud crashed (OAuthError): ('Error code invalid_request: Invalid value for "audience". This value should be the full resource name of the Identity Provider. See https://cloud.google.com/iam/docs/reference/sts/rest/v1/TopLevel/token for the list of possible formats.', '{"error":"invalid_request","error_description":"Invalid value for \\"audience\\". This value should be the full resource name of the Identity Provider. See https://cloud.google.com/iam/docs/reference/sts/rest/v1/TopLevel/token for the list of possible formats."}')`

PR #7287, opened 2023-09-18
op-node: Span Batch Hardfork Configuration

ImTei:
## Overview

This PR contains the Hardfork configurations for the [Span Batch](https://github.com/ethereum-optimism/optimism/blob/develop/specs/span-batches.md). 

The Span Batch’s hard fork activation follows the same pattern as the Regolith hard fork. We defined a timestamp that triggers the hard fork at the op-node’s rollup.json—spanBatchTime.

Please refer to the `Hard fork` section of Implementation Design Docs for details.

## Changes

- Added `SpanBatchTime` that triggers the Span Batch hard fork.
- Added `OP_E2E_USE_SPAN_BATCH` flag that executes e2e tests with Span Batch.

PR #7288, opened 2023-09-18
op-node: Span Batch Type, Encoding, and Decoding

ImTei:
## Contexts

This PR contains the type, encoding, and decoding code for [Span Batch](https://github.com/ethereum-optimism/optimism/blob/develop/specs/span-batches.md) as defined specs. 

Please refer to the Encoding Optimization section in Design Docs for the rationale behind the current encoding specs.

This PR has a dependency on the[ **Hardfork Configuration** PR](https://github.com/ethereum-optimism/optimism/pull/7287).

## Changes

- Added the following types:
    - Batch interface
    - RawSpanBatch
    - SpanBatchElement
    - SpanBatch
    - SpanBatchBuilder
- Changed the following types:
    - Renamed BatchV1 to SingularBatch.
    - BatchData
- ^ Please refer to the `Type` section of Implementation Design Docs for details.
- Added unit tests.
    - ^ Please refer to the `Data Structure` category in the Test List Sheet for each test’s details.

### DIFF
--- a/op-node/rollup/derive/batch.go
+++ b/op-node/rollup/derive/batch.go
@@ -72,7 +72,7 @@ func (b *BatchData) encodeTyped(buf *bytes.Buffer) error {
 		buf.WriteByte(SingularBatchType)
 		return rlp.Encode(buf, &b.SingularBatch)
 	case SpanBatchType:
-		buf.WriteByte(byte(b.BatchType))
+		buf.WriteByte(SpanBatchType)
 		return b.RawSpanBatch.encode(buf)
 	default:
 		return fmt.Errorf("unrecognized batch type: %d", b.BatchType)
--- a/op-node/rollup/derive/batch.go
+++ b/op-node/rollup/derive/batch.go
@@ -7,48 +7,44 @@ import (
 	"io"
 	"sync"
 
-	"github.com/ethereum-optimism/optimism/op-node/rollup"
-	"github.com/ethereum-optimism/optimism/op-service/eth"
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
 // Batch format
 // first byte is type followed by bytestring.
 //
-// BatchV1Type := 0
-// batchV1 := BatchV1Type ++ RLP([epoch, timestamp, transaction_list]
-//
 // An empty input is not a valid batch.
 //
 // Note: the type system is based on L1 typed transactions.
-
+//
 // encodeBufferPool holds temporary encoder buffers for batch encoding
 var encodeBufferPool = sync.Pool{
 	New: func() any { return new(bytes.Buffer) },
 }
 
 const (
-	BatchV1Type = iota
+	// SingularBatchType is the first version of Batch format, representing a single L2 block.
+	SingularBatchType = iota
+	// SpanBatchType is the Batch version used after SpanBatch hard fork, representing a span of L2 blocks.
+	SpanBatchType
 )
 
-type BatchV1 struct {
-	ParentHash common.Hash  // parent L2 block hash
-	EpochNum   rollup.Epoch // aka l1 num
-	EpochHash  common.Hash  // block hash
-	Timestamp  uint64
-	// no feeRecipient address input, all fees go to a L2 contract
-	Transactions []hexutil.Bytes
+// Batch contains information to build one or multiple L2 blocks.
+// Batcher converts L2 blocks into Batch and writes encoded bytes to Channel.
+// Derivation pipeline decodes Batch from Channel, and converts to one or multiple payload attributes.
+type Batch interface {
+	GetBatchType() int
+	GetTimestamp() uint64
+	LogContext(log.Logger) log.Logger
 }
 
+// BatchData is a composition type that contains raw data of each batch version.
+// It has encoding & decoding methods to implement typed encoding.
 type BatchData struct {
-	BatchV1
-	// batches may contain additional data with new upgrades
-}
-
-func (b *BatchV1) Epoch() eth.BlockID {
-	return eth.BlockID{Hash: b.EpochHash, Number: uint64(b.EpochNum)}
+	BatchType int
+	SingularBatch
+	RawSpanBatch
 }
 
 // EncodeRLP implements rlp.Encoder
@@ -69,9 +65,18 @@ func (b *BatchData) MarshalBinary() ([]byte, error) {
 	return buf.Bytes(), err
 }
 
+// encodeTyped encodes batch type and payload for each batch type.
 func (b *BatchData) encodeTyped(buf *bytes.Buffer) error {
-	buf.WriteByte(BatchV1Type)
-	return rlp.Encode(buf, &b.BatchV1)
+	switch b.BatchType {
+	case SingularBatchType:
+		buf.WriteByte(SingularBatchType)
+		return rlp.Encode(buf, &b.SingularBatch)
+	case SpanBatchType:
+		buf.WriteByte(byte(b.BatchType))
+		return b.RawSpanBatch.encode(buf)
+	default:
+		return fmt.Errorf("unrecognized batch type: %d", b.BatchType)
+	}
 }
 
 // DecodeRLP implements rlp.Decoder
@@ -94,14 +99,35 @@ func (b *BatchData) UnmarshalBinary(data []byte) error {
 	return b.decodeTyped(data)
 }
 
+// decodeTyped decodes batch type and payload for each batch type.
 func (b *BatchData) decodeTyped(data []byte) error {
 	if len(data) == 0 {
 		return fmt.Errorf("batch too short")
 	}
 	switch data[0] {
-	case BatchV1Type:
-		return rlp.DecodeBytes(data[1:], &b.BatchV1)
+	case SingularBatchType:
+		b.BatchType = SingularBatchType
+		return rlp.DecodeBytes(data[1:], &b.SingularBatch)
+	case SpanBatchType:
+		b.BatchType = int(data[0])
+		return b.RawSpanBatch.decodeBytes(data[1:])
 	default:
 		return fmt.Errorf("unrecognized batch type: %d", data[0])
+	}
+}
+
+// NewSingularBatchData creates new BatchData with SingularBatch
+func NewSingularBatchData(singularBatch SingularBatch) *BatchData {
+	return &BatchData{
+		BatchType:     SingularBatchType,
+		SingularBatch: singularBatch,
+	}
+}
+
+// NewSpanBatchData creates new BatchData with SpanBatch
+func NewSpanBatchData(spanBatch RawSpanBatch) *BatchData {
+	return &BatchData{
+		BatchType:    SpanBatchType,
+		RawSpanBatch: spanBatch,
 	}
 }
--- a/op-node/rollup/derive/batch_tob_test.go
+++ b/op-node/rollup/derive/batch_tob_test.go
@@ -10,7 +10,7 @@ import (
 
 // FuzzBatchRoundTrip executes a fuzz test similar to TestBatchRoundTrip, which tests that arbitrary BatchData will be
 // encoded and decoded without loss of its original values.
-// Does not test span batch because fuzzer is not aware structure of span batch.
+// Does not test the span batch type because the fuzzer is not aware of the structure of a span batch.
 func FuzzBatchRoundTrip(f *testing.F) {
 	f.Fuzz(func(t *testing.T, fuzzedData []byte) {
 		// Create our fuzzer wrapper to generate complex values
--- a/op-node/rollup/derive/batch_tob_test.go
+++ b/op-node/rollup/derive/batch_tob_test.go
@@ -10,6 +10,7 @@ import (
 
 // FuzzBatchRoundTrip executes a fuzz test similar to TestBatchRoundTrip, which tests that arbitrary BatchData will be
 // encoded and decoded without loss of its original values.
+// Does not test span batch because fuzzer is not aware structure of span batch.
 func FuzzBatchRoundTrip(f *testing.F) {
 	f.Fuzz(func(t *testing.T, fuzzedData []byte) {
 		// Create our fuzzer wrapper to generate complex values
@@ -19,6 +20,10 @@ func FuzzBatchRoundTrip(f *testing.F) {
 		// Create our batch data from fuzzed data
 		var batchData BatchData
 		typeProvider.Fuzz(&batchData)
+
+		// force batchdata to only contain singular batch
+		batchData.BatchType = SingularBatchType
+		batchData.RawSpanBatch = RawSpanBatch{}
 
 		// Encode our batch data
 		enc, err := batchData.MarshalBinary()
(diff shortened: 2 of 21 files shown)


op-node/rollup/derive/batch.go
```diff
@@ -7,48 +7,44 @@ import (
...
 type BatchData struct {
-	BatchV1
-	// batches may contain additional data with new upgrades
-}
-
-func (b *BatchV1) Epoch() eth.BlockID {
-	return eth.BlockID{Hash: b.EpochHash, Number: uint64(b.EpochNum)}
+	BatchType int
+	SingularBatch
+	RawSpanBatch
```
protolambda, line 47: Adding the type explicitly, and embedding each kind, is probably the less intrusive change we can make to the existing approach with typed BatchData. We can maybe refactor it to be more similar to `types.Transaction` in upstream geth, where a struct is used to represent the typed decoding/encoding and any caching, and wraps around a single interface, rather than embedding each type of possible inner batch. We can keep this for now, to keep the PR stack easier to land into develop.

op-node/rollup/derive/batch_tob_test.go
```diff
@@ -10,6 +10,7 @@ import (
 
 // FuzzBatchRoundTrip executes a fuzz test similar to TestBatchRoundTrip, which tests that arbitrary BatchData will be
 // encoded and decoded without loss of its original values.
+// Does not test span batch because fuzzer is not aware structure of span batch.
```
protolambda:
```suggestion
// Does not test the span batch type because the fuzzer is not aware of the structure of a span batch.
```

op-node/rollup/derive/span_batch.go
```diff
@@ -0,0 +1,670 @@
...
+// payload := payload = block_count ++ origin_bits ++ block_tx_counts ++ txs
+// txs = contract_creation_bits ++ y_parity_bits ++ tx_sigs ++ tx_tos ++ tx_datas ++ tx_nonces ++ tx_gases
+
+var ErrTooBigSpanBatchFieldSize = errors.New("batch would cause field bytes to go over limit")
+
+type spanBatchPrefix struct {
+	relTimestamp  uint64 // Relative timestamp of the first block
+	l1OriginNum   uint64 // L1 origin number
+	parentCheck   []byte // First 20 bytes of the first block's parent hash
+	l1OriginCheck []byte // First 20 bytes of the last block's L1 origin hash
```
protolambda, line 34: Stylistically we can make these fields `[20]byte`, but also fine to leave it the same for now, functionally the same.

op-node/rollup/derive/span_batch.go
```diff
@@ -0,0 +1,670 @@
...
+}
+
+// GetTimestamp returns timestamp of the first block in the span
+func (b *SpanBatch) GetTimestamp() uint64 {
+	return b.batches[0].Timestamp
+}
+
+// LogContext creates a new log context that contains information of the batch
+func (b *SpanBatch) LogContext(log log.Logger) log.Logger {
+	return log.New(
```
protolambda, line 439: Although empty span-batches are invalid, we might still log a batch while we're building it locally using an empty span batch as start + append calls. Adding a check here in case `len(b.batches) == 0` to create a different log context safely, and not cause a panic, would be nice.

op-node/rollup/derive/span_batch_test.go
```diff
@@ -0,0 +1,514 @@
...
+	"math/big"
+	"math/rand"
+	"testing"
+
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum-optimism/optimism/op-node/testutils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/stretchr/testify/assert"
```
protolambda, line 14: In some cases, when asserting if an error is not nil, using the `require` package (same API) is better, to stop the test at that unexpected error rather than continuing. E.g. `require.NoError(err)`. It can be fixed in a follow-up PR though, it's fine to keep as-is for now.

op-node/rollup/derive/span_batch_tx.go
```diff
@@ -0,0 +1,253 @@
...
+			return nil, fmt.Errorf("failed to decode spanBatchDynamicFeeTxData: %w", err)
+		}
+		return &inner, nil
+	default:
+		return nil, types.ErrTxTypeNotSupported
+	}
+}
+
+// DecodeRLP implements rlp.Decoder
+func (tx *spanBatchTx) DecodeRLP(s *rlp.Stream) error {
```
protolambda, line 119: In `func (btx *spanBatchTxs) fullTxs(chainID *big.Int)` you use the `func (tx *spanBatchTx) UnmarshalBinary(b []byte) error {` version. Is the `DecodeRLP` method really required? And `newSpanBatchTxs` uses `MarshalBinary` too.

op-node/rollup/derive/span_batch_txs.go
```diff
@@ -0,0 +1,475 @@
...
+	var txGases []uint64
+	var txDatas []hexutil.Bytes
+	var txTypes []int
+	contractCreationBits := new(big.Int)
+	yParityBits := new(big.Int)
+	for idx := 0; idx < int(totalBlockTxCount); idx++ {
+		var tx types.Transaction
+		if err := tx.UnmarshalBinary(txs[idx]); err != nil {
+			return nil, errors.New("failed to decode tx")
+		}
```
protolambda, line 434:
```suggestion
		}
		if tx.Protected() && tx.ChainId().Cmp(chainID) != 0 {
			return nil, fmt.Errorf("protected tx has chain ID %d, but expected chain ID %d", tx.ChainId(), chainID)
		}
```

The `chainID` argument is unused. We can use it for a sanity check like this. (but only if the tx is protected, i.e. is specific to a single chain).

protolambda:
Adding a do-not-merge label to give others the chance to review before auto-merge by mergify kicks in.

protolambda (approved):
Looks good! There are some stylistic but breaking things I think we can do in a follow-up PR, after the PR stack is wrapped up.

op-node/rollup/derive/span_batch_tx.go
ImTei: `DecodeRLP` was implemented just for convention with other types. It's not required. Let me remove it in a follow-up PR!

op-node/rollup/derive/batch.go
pcw109550: Will do this in a separate PR. Strongly agree that following what `types.Transaction` did is better because current composition pattern will become more complicated when we need another sub field of composition. We need to add additional field for new span batch type when needed if we do not use a single interface.

Let me make the abstraction idea concrete. type `BatchData` will be refactored as below:
```go
type BatchData struct {
	inner BatchInnerData
}
```
where `BatchInnerData` is an interface:
```go
type BatchInnerData interface {
	batchType() int
        encode(w io.Writer) error
        decode(r *bytes.Reader) error
}
```

Each `SingularBatch` and `RawSpanBatch` will implement `BatchInnerData` interface. This will encapsulate all the batch type related conditionals to each implementations: for example below method will be refactored from:
```go
func (b *BatchData) encodeTyped(buf *bytes.Buffer) error {
	switch b.BatchType {
	case SingularBatchType:
		buf.WriteByte(SingularBatchType)
		return rlp.Encode(buf, &b.SingularBatch)
	case SpanBatchType:
		buf.WriteByte(byte(b.BatchType))
		return b.RawSpanBatch.encode(buf)
	default:
		return fmt.Errorf("unrecognized batch type: %d", b.BatchType)
	}
}
```
to
```go
func (b *BatchData) encodeTyped(buf *bytes.Buffer) error {
        if err := buf.WriteByte(b.inner.batchType()); err != nil {
            return err
        }
        return b.inner.encode(buf)
}
```

pcw109550: I just found out that `buf.WriteByte(SingularBatchType)` is not error handled. Will also fix this.

op-node/rollup/derive/batch.go
```diff
@@ -69,9 +65,18 @@ func (b *BatchData) MarshalBinary() ([]byte, error) {
...
+// encodeTyped encodes batch type and payload for each batch type.
 func (b *BatchData) encodeTyped(buf *bytes.Buffer) error {
-	buf.WriteByte(BatchV1Type)
-	return rlp.Encode(buf, &b.BatchV1)
+	switch b.BatchType {
+	case SingularBatchType:
+		buf.WriteByte(SingularBatchType)
+		return rlp.Encode(buf, &b.SingularBatch)
+	case SpanBatchType:
+		buf.WriteByte(byte(b.BatchType))
```
trianglesphere:
```suggestion
		buf.WriteByte(SpanBatchType)
```

op-node/rollup/derive/span_batch.go
```diff
@@ -0,0 +1,670 @@
...
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+// Batch format
+//
+// SpanBatchType := 1
+// spanBatch := SpanBatchType ++ prefix ++ payload
+// prefix := rel_timestamp ++ l1_origin_num ++ parent_check ++ l1_origin_check
+// payload := payload = block_count ++ origin_bits ++ block_tx_counts ++ txs
```
trianglesphere:
```suggestion
// payload := block_count ++ origin_bits ++ block_tx_counts ++ txs
```

op-node/rollup/derive/span_batch.go
```diff
@@ -0,0 +1,670 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read l1 origin num: %w", err)
+	}
+	bp.l1OriginNum = L1OriginNum
+	return nil
+}
+
+// decodeParentCheck parses data into bp.parentCheck
+func (bp *spanBatchPrefix) decodeParentCheck(r *bytes.Reader) error {
+	bp.parentCheck = make([]byte, 20)
```
trianglesphere, line 104: A lot of the decode functions are duplicating basic building blocks like decode20Bytes or decodeUVarInt

pcw109550: Code duplication should be avoided, but these structures are intentionally duplicated and not abstracted. For example, lets say we want to abstract out `binary.ReadUvarint(r)` as a function `decodeUvarint`:
```go
func decodeUvarint(r *bytes.Reader) (uint64, error) {
    return binary.ReadUvarint(r)
}
```
Above implementation is an unncessary abstraction. `decode20Bytes` would be better:
```go
func decode20Bytes(r *bytes.Reader) ([]byte, error) {
	result := make([]byte, 20)
	_, err := io.ReadFull(r, result)
	if err != nil {
		return nil, err
	}
	return result, err
}
```
and implement `decodeParentCheck` like:
```go
func (bp *spanBatchPrefix) decodeParentCheck(r *bytes.Reader) error {
	result, err := decode20Bytes(r)
	if err != nil {
		return fmt.Errorf("failed to read parent check: %w", err)
	}
	bp.parentCheck = result
	return nil
}
```
However I can observe the pattern: call `make` to allocate and call `io.ReadFull` throughout the entire codebase. These patterns are not abstracted, and I think leaving it as is(not encapsulating as function) is better for readability.

pcw109550: Also, we can observer structure of method `decodeParentCheck` and `decodeL1OriginCheck` is identical:
```go
// decodeParentCheck parses data into bp.parentCheck
func (bp *spanBatchPrefix) decodeParentCheck(r *bytes.Reader) error {
	bp.parentCheck = make([]byte, 20)
	_, err := io.ReadFull(r, bp.parentCheck)
	if err != nil {
		return fmt.Errorf("failed to read parent check: %w", err)
	}
	return nil
}

// decodeL1OriginCheck parses data into bp.decodeL1OriginCheck
func (bp *spanBatchPrefix) decodeL1OriginCheck(r *bytes.Reader) error {
	bp.l1OriginCheck = make([]byte, 20)
	_, err := io.ReadFull(r, bp.l1OriginCheck)
	if err != nil {
		return fmt.Errorf("failed to read l1 origin check: %w", err)
	}
	return nil
}

Comment (pcw109550) [reply op-node/rollup/derive/span_batch.go:104]: Also, we can observe structure of method `decodeParentCheck` and `decodeL1OriginCheck` is identical:
```go
// decodeParentCheck parses data into bp.parentCheck
func (bp *spanBatchPrefix) decodeParentCheck(r *bytes.Reader) error {
	bp.parentCheck = make([]byte, 20)
	_, err := io.ReadFull(r, bp.parentCheck)
	if err != nil {
		return fmt.Errorf("failed to read parent check: %w", err)
	}
	return nil
}

// decodeL1OriginCheck parses data into bp.decodeL1OriginCheck
func (bp *spanBatchPrefix) decodeL1OriginCheck(r *bytes.Reader) error {
	bp.l1OriginCheck = make([]byte, 20)
	_, err := io.ReadFull(r, bp.l1OriginCheck)
	if err != nil {
		return fmt.Errorf("failed to read l1 origin check: %w", err)
	}
	return nil
}
```
and I may implement `decode20Bytes` function and call it twice instead of calling different two methods. I thought that even though there is a duplication, breaking the decoding logic into chunks and explicitly naming them(make them into function) would be much better for readability.

ImTei:
applied simple [changes](https://github.com/ethereum-optimism/optimism/pull/7288/commits/23c658d60e03e633a9097e5939d33798f2837170) from code reviews and rebased to the latest develop branch.

Other suggestions will be applied in the follow-up PR!

op-node/rollup/derive/singular_batch.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/ethereum-optimism/optimism/op-service/eth"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// Batch format
+//
+// SingularBatchType := 0
+// singularBatch := SingularBatchType ++ RLP([epoch, timestamp, transaction_list]
```
protolambda:
```suggestion
// singularBatch := SingularBatchType ++ RLP([parent_hash, epoch_number, epoch_hash, timestamp, transaction_list])
```

protolambda:
Follow-ups to implement in later PR:

- `spanBatchTxs` has unused `EncodeRLP` and `DecodeRLP`. The marshal/unmarshal-binary functions are enough.
- Some encoding/decoding tests in `span_batch_tx_test.go` randomize the test data and assert 
  if the tx types all occurred at the end. Instead, we can make sub-tests with `t.Run()` for each tx-type,
  and randomize the same amount of txs per type.
- The `BatchData` wrapper around the typed batch shouldn't have an explicit byte field for the type:
  it can already be inferred from the contained batch data. See `types.Transaction` in geth as example.
- `decodeTyped` can be simplified, and the switch-statement should have be more consistent (not mix constants and `data[0]`)
- `SingularBatchType = iota` is neat, but being explict about protocol constants is better. `iota` is only really meant for internal enumerations that you can flexibly add/remove to.
- Review usage of `require` vs `assert` in batch tests
- `parentCheck   []byte` and `l1OriginCheck []byte` should be `[20]byte` types

And opened #7615 to fix a blocking issue: we must not allow span-batches to be decoded successfully before the HF, or we diverge from the actual mainnet specs.

PR #7289, opened 2023-09-18
op-node: Span Batch Derivation

ImTei:
## Contexts

This PR contains the derivation code for [Span Batch](https://github.com/ethereum-optimism/optimism/blob/develop/specs/span-batches.md).

Please refer to the **Implementations - Derivation** and **Hard Fork Activation** sections of the Design Docs and **op-node** section of Implementation Design Docs for details & rationales.

This PR has a dependency on the [**Span Batch Type, Encoding, and Decoding** PR](https://github.com/ethereum-optimism/optimism/pull/7288).

## Changes

- Modifications:
    - Modify **ChannelReader** to decode both **SingularBatch** and **SpanBatch**.
    - Verify if the processing batch’s timestamp matches the expected batch type by **spanBatchTime** at **BatchQueue**.
    - **BatchQueue** splits a **SpanBatch** to multiple **SingularBatch**-es.
    - Leave the rest of the derivation stages as-is by passing **SingularBatches**.
    - ^ Please refer to the `op-node` section of Implementation Design Docs for details.
- Added unit tests.
    - ^ Please refer to the `Derivation` category in the Test List Sheet for each test’s details.

### DIFF
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -243,15 +243,15 @@ batchLoop:
 	// batch to ensure that we at least have one batch per epoch.
 	if nextTimestamp < nextEpoch.Time || firstOfEpoch {
 		bq.log.Info("Generating next batch", "epoch", epoch, "timestamp", nextTimestamp)
-		return &BatchData{
-			BatchV1{
+		return NewSingularBatchData(
+			SingularBatch{
 				ParentHash:   l2SafeHead.Hash,
 				EpochNum:     rollup.Epoch(epoch.Number),
 				EpochHash:    epoch.Hash,
 				Timestamp:    nextTimestamp,
 				Transactions: nil,
 			},
-		}, nil
+		), nil
 	}
 
 	// At this point we have auto generated every batch for the current epoch
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -78,7 +78,7 @@ func (bq *BatchQueue) popNextBatch(safeL2Head eth.L2BlockRef) *SingularBatch {
 	return nextBatch
 }
 
-func (bq *BatchQueue) advanceEpoch(nextBatch *SingularBatch) {
+func (bq *BatchQueue) advanceEpochMaybe(nextBatch *SingularBatch) {
 	if nextBatch.GetEpochNum() == rollup.Epoch(bq.l1Blocks[0].Number)+1 {
 		// Advance epoch if necessary
 		bq.l1Blocks = bq.l1Blocks[1:]
@@ -89,7 +89,7 @@ func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) 
 	if len(bq.nextSpan) > 0 {
 		// If there are cached singular batches, pop first one and return.
 		nextBatch := bq.popNextBatch(safeL2Head)
-		bq.advanceEpoch(nextBatch)
+		bq.advanceEpochMaybe(nextBatch)
 		return nextBatch, nil
 	}
 
@@ -169,7 +169,7 @@ func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) 
 		return nil, NewCriticalError(fmt.Errorf("unrecognized batch type: %d", batch.GetBatchType()))
 	}
 
-	bq.advanceEpoch(nextBatch)
+	bq.advanceEpochMaybe(nextBatch)
 	return nextBatch, nil
 }
 
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -29,7 +29,12 @@ import (
 
 type NextBatchProvider interface {
 	Origin() eth.L1BlockRef
-	NextBatch(ctx context.Context) (*BatchData, error)
+	NextBatch(ctx context.Context) (Batch, error)
+}
+
+type SafeBlockFetcher interface {
+	L2BlockRefByNumber(context.Context, uint64) (eth.L2BlockRef, error)
+	PayloadByNumber(context.Context, uint64) (*eth.ExecutionPayload, error)
 }
 
 // BatchQueue contains a set of batches for every L1 block.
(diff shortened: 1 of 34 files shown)


op-node/rollup/derive/batch_queue.go
```diff
@@ -42,24 +47,52 @@ type BatchQueue struct {
 
 	l1Blocks []eth.L1BlockRef
 
-	// batches in order of when we've first seen them, grouped by L2 timestamp
-	batches map[uint64][]*BatchWithL1InclusionBlock
+	// batches in order of when we've first seen them
+	batches []*BatchWithL1InclusionBlock
```
protolambda, line 51: Is the change to a single batch queue, rather than ones grouped by timestamp, required for span-batch processing?

Some previous context you may not be aware of: A previous version of the code implemented this with the slice of batches, but then changed it to a map where items are grouped by timestamp, to optimize the iteration if there are lots of batches.

I like the simplification back to a slice however, as with span-batches a lot of the batches will be ordered anyway, and iterating the batches does not seem like such a big deal. But happy to hear the other arguments for the `map`.

op-node/rollup/derive/batch_queue.go
```diff
@@ -42,24 +47,52 @@ type BatchQueue struct {
...
+
+	// nextSpan is cached SingularBatches derived from SpanBatch
+	nextSpan []*SingularBatch
```
protolambda, line 54: I like how the span is expanded into singular batches, and the testing is applied to both the batch types.

op-node/rollup/derive/batch_queue.go
```diff
@@ -42,24 +47,52 @@ type BatchQueue struct {
...
-func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) (*BatchData, error) {
+func (bq *BatchQueue) popNextBatch(safeL2Head eth.L2BlockRef) *SingularBatch {
+	nextBatch := bq.nextSpan[0]
+	bq.nextSpan = bq.nextSpan[1:]
+	// Must set ParentHash before return. we can use safeL2Head because the parentCheck is verified in CheckBatch().
+	nextBatch.ParentHash = safeL2Head.Hash
+	return nextBatch
+}
+
+func (bq *BatchQueue) advanceEpoch(nextBatch *SingularBatch) {
```
protolambda: I would rename this to `advanceEpochMaybe`, since it checks if it has to advance, before it actually advances.

op-node/rollup/derive/batch_queue_test.go
```diff
@@ -367,6 +533,387 @@ func TestBatchQueueMissing(t *testing.T) {
...
+	}
+
+	l2Client := testutils.MockL2Client{}
+	var nilErr error
+	for i, batch := range expectedOutputBatches {
+		if batch != nil {
+			blockRef := singularBatchToBlockRef(t, batch, uint64(i+1))
+			payload := singularBatchToPayload(t, batch, uint64(i+1))
+			l2Client.Mock.On("L2BlockRefByNumber", uint64(i+1)).Times(9999).Return(blockRef, &nilErr)
+			l2Client.Mock.On("PayloadByNumber", uint64(i+1)).Times(9999).Return(&payload, &nilErr)
```
protolambda: When using a mock you it's nice to assert what happens, i.e. `l2Client.Mock.AssertExpectations(t)` after running through the calls somewhere at the end of the test, to see if the calls were all actually made as expected. You can use `.Maybe()` to make a call optional. I see you use it more like a util here, to set up the client in a supporting form, rather than exact call-by-call assertions.

op-node/rollup/derive/batch_queue_test.go
```diff
@@ -367,6 +533,387 @@ func TestBatchQueueMissing(t *testing.T) {
...
```
protolambda: Same here

op-node/rollup/derive/batches.go
```diff
@@ -141,3 +167,204 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
+	}
+
+	// Check the L1 origin of the batch
+	if startEpochNum > parentBlock.L1Origin.Number+1 {
+		log.Warn("batch is for future epoch too far ahead, while it has the next timestamp, so it must be invalid", "current_epoch", epoch.ID())
+		return BatchDrop
+	}
+
+	endEpochNum := batch.GetBlockEpochNum(batch.GetBlockCount() - 1)
+	originChecked := false
```
protolambda, line 238: I think this is safe-ish: it's limited by the `origin_bits` length, and can thus not be infinite. But we might end up waiting for a very large span-batch (in terms of covered L1 origins) if one is crafted maliciously like that. We may need to harden it with additional checks/limits here. We can do so in a follow-up PR however.

op-node/rollup/derive/batches.go
```diff
@@ -28,14 +31,37 @@ const (
...
+	case SpanBatchType:
+		spanBatch, ok := batch.Batch.(*SpanBatch)
+		if !ok {
+			log.Error("failed type assertion to SpanBatch")
+			return BatchDrop
+		}
+		if !cfg.IsSpanBatch(batch.Batch.GetTimestamp()) {
+			log.Warn("received SpanBatch before SpanBatch hard fork")
+			return BatchDrop
+		}
```
protolambda, line 53: I think we may want to add a second check like this based on the L1 block the batch was included in, to check that timestamp as well, to really make sure we do not end up processing span-batches too early by accident.

protolambda (commented):
Looks good, left a few comments about testing and edge-cases to harden against. I will do a second review pass later, since the derivation code is quite critical.

protolambda:
Adding a do-not-merge label to give others the chance to review as well.

op-node/rollup/derive/batch_queue.go
ImTei: @protolambda Changing to a single batch queue is required to allow overlapping span batches. In the original code, overlapping batches may not be in the queue at `nextTimestamp`, so we have to scan every batches from the past.

Additionally, if I understand correctly, there might be a memory leak in the current code since we don't clean up `remaining` batches from the queue after accepting a batch. This code change also resolves this leak.

op-node/rollup/derive/batches.go
ImTei:
```
The span-batch containing L2 blocks after the hard fork, is included in L1 block before the hard fork timestamp.
```
^ Is this the scenario that could be a problem? can you explain why it's a problem? (we already discussed this in the last weekly sync, sorry for the repeated question 😅)

If it's a problem and we have to check the L1 inclusion time, I think we should change the spec for hard fork activation rule described in [superchain-upgrades](https://github.com/ethereum-optimism/optimism/blob/develop/specs/superchain-upgrades.md#l2-block-timestamp-based-activation). what do you think?

protolambda:
Opened https://github.com/ethereum-optimism/optimism/pull/7621 to resolve conflicts with `develop`

Follow-ups to implement after this PR:
- \[in response to span-batch origin-bits checking]:
  I think this is safe-ish: it's limited by the origin_bits length, and can thus not be infinite.
  But we might end up waiting for a very large span-batch (in terms of covered L1 origins) if one is crafted maliciously like that.
  We may need to harden it with additional checks/limits here. We can do so in a follow-up PR however.
- The Batch typing improved a little, but we should look if we can clean this up more:
  ideally there are less different types of intermediate batch forms.
- Proto moved the fix from last PR that filters span-batches upon reading from the batch-reader to the channel-in-reader,
  since the batch-reader no longer has L1 inclusion data (due to the way span-batches work).
  We still want to avoid span-batch code-paths before the hardfork activates however,
  so it's worth checking the L1 data before converting a raw span-batch into a full one.
- In the batch-checking function we need to review the hardfork activation conditions, 
  and update the specs as well, so the span-batches activate at the right moment (w.r.t. L1 inclusion and L2 block time)
  This should be added to the span-batch derivation specs, and then referenced as a feature of Canyon,
  in the superchain-upgrades spec document.

protolambda:
Closing in favor of #7621

PR #7290, opened 2023-09-18
op-batcher: Span Batch Submission

ImTei:
## Contexts

This PR contains the batcher code for [Span Batch](https://github.com/ethereum-optimism/optimism/blob/develop/specs/span-batches.md).

Please refer to the **Implementations - Batcher** and **Hard Fork Activation** sections of the Design Docs and **op-batcher** section of Implementation Design Docs for details & rationales.

This PR has a dependency on the [**Span Batch Derivation** PR](https://github.com/ethereum-optimism/optimism/pull/7289).

## Changes

- Modifications:
    - New flag: `--batch-type` : configure batch type. 0 for `SingularBatch` and 1 for `SpanBatch`. (default: 0)
    - When `--batch-type=1`:
        - For each L2 block accumulated, compress the entire batch and check if the compressed size overflows.
        - Rationale: We cannot use the compression stream for Span Batch since it changes the entire contents as each L2 block is accumulated.
    - ^ Please refer to the `op-batcher` section of Implementation Design Docs for details.
- Added unit tests.
    - ^ Please refer to the `Batch Submission` category in the Test List Sheet for each test’s details.

### DIFF
--- a/op-batcher/batcher/driver.go
+++ b/op-batcher/batcher/driver.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
 	"io"
 	"math/big"
 	_ "net/http/pprof"
@@ -17,6 +16,7 @@ import (
 
 	"github.com/ethereum-optimism/optimism/op-batcher/metrics"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
 	"github.com/ethereum-optimism/optimism/op-service/eth"
 	"github.com/ethereum-optimism/optimism/op-service/txmgr"
 )
--- a/op-batcher/batcher/driver.go
+++ b/op-batcher/batcher/driver.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
 	"io"
 	"math/big"
 	_ "net/http/pprof"
@@ -16,7 +17,6 @@ import (
 
 	"github.com/ethereum-optimism/optimism/op-batcher/metrics"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
-	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
 	"github.com/ethereum-optimism/optimism/op-service/eth"
 	"github.com/ethereum-optimism/optimism/op-service/txmgr"
 )
@@ -74,7 +74,7 @@ type BatchSubmitter struct {
 func NewBatchSubmitter(setup DriverSetup) *BatchSubmitter {
 	return &BatchSubmitter{
 		DriverSetup: setup,
-		state:       NewChannelManager(setup.Log, setup.Metr, setup.Channel),
+		state:       NewChannelManager(setup.Log, setup.Metr, setup.Channel, setup.RollupCfg),
 	}
 }
 
@@ -91,7 +91,11 @@ func (l *BatchSubmitter) StartBatchSubmitting() error {
 
 	l.shutdownCtx, l.cancelShutdownCtx = context.WithCancel(context.Background())
 	l.killCtx, l.cancelKillCtx = context.WithCancel(context.Background())
-	l.state.Clear()
+	syncStatus, err := fetchSyncStatus(l.shutdownCtx, l.RollupClient, l.Cfg.NetworkTimeout)
+	if err != nil {
+		return err
+	}
+	l.state.Clear(&syncStatus.SafeL2)
 	l.lastStoredBlock = eth.BlockID{}
 
 	l.wg.Add(1)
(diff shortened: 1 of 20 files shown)


op-batcher/batcher/driver.go
```diff
@@ -115,10 +117,17 @@ func NewBatchSubmitter(ctx context.Context, cfg Config, l log.Logger, m metrics.
 
 	cfg.metr = m
 
+	syncStatus, err := fetchSyncStatus(ctx, cfg.RollupNode, cfg.NetworkTimeout)
+	if err != nil {
+		return nil, err
+	}
```
protolambda: adding the safe-L2 block ref as constructor argument is probably better, so the caller can manage retries to deal with any network errors, before creating the batch-submitter object.

protolambda (commented):
Changes look ok. Left a comment about the network-fetch in the a constructor, which I think we should clean up. And I would appreciate a 2nd review from @trianglesphere here, who has more context on the batch-submitter.

protolambda:
Added a do-not-merge label so others have the chance to review as well.

op-node/rollup/derive/channel_out.go
```diff
@@ -136,6 +173,70 @@ func (co *ChannelOut) AddBatch(batch *BatchData) (uint64, error) {
 	return uint64(written), err
 }
 
+// writeSpanBatch appends a SingularBatch to the channel's SpanBatch.
+// A channel can have only one SpanBatch. And compressed results should not be accessible until the channel is closed, since the prefix and payload can be changed.
+// So it resets channel contents and rewrites the entire SpanBatch each time, and compressed results are copied to reader after the channel is closed.
```
trianglesphere: How does this integrate with the shadow/non-shadow compressor? I don't love the default behavior of recompressing the channel every time we add a new block, but the shadow compressor does do something similar.

op-node/rollup/derive/channel_out.go
```diff
@@ -57,20 +64,35 @@ type ChannelOut struct {
 
 	// Compressor stage. Write input data to it
 	compress Compressor
-
+	// closed indicates if the channel is closed
 	closed bool
+	// batchType indicates whether this channel uses SingularBatch or SpanBatch
+	batchType uint
+	// spanBatchBuilder contains information requires to build SpanBatch
+	spanBatchBuilder *SpanBatchBuilder
```
trianglesphere: What's the API for building span batches - it kinda seems different enough that we should be using a `SpanChannelOut` rather than having two different sets of logic inside `ChannelOut`

op-node/rollup/derive/channel_out.go
ImTei: I agree that recompressing the entire channel every time may not be that great. But we couldn't think of a better alternative for that part due to the data layout of the span-batch.

As you said, using the shadow compressor makes unnecessary compression. so the ratio compressor will be more suitable for span-batch.

ImTei:
@trianglesphere 
I also agree that the implementation might be confusing or unnatural—we also discussed a lot about this. It may be because the current abstraction structure of the batch-submitter is difficult to handle span batch as it’s designed for singular batches. 

- ChannelOut adds the next block(=singular batch) at the end of the channel.
- Compressor estimates the size of the stream after adding a block.

These are not the case for the span batch due to its data layout. We must rebuild and recompress the entire channel when a block is added because existing contents in the stream are being modified—you cannot change the data that’s already in the stream.

As you said, we can design a common API for singular & span batches, and implement a new ChannelOut for span batches with its own compressor. We may have a better code, but it will have much more code diffs. Our priority was designing the most efficient Span-batch structure with the safest and simplest implementation. 

I think we can plan for a better design structure for the batch submitter that accommodates both singular and span batches after this PR. What do you think?

op-node/rollup/derive/channel_out.go
ImTei: Low estimation accuracy of the ratio compressor is not a problem, because `ChannelOut` does not use the compressor's estimation logic when it's working with span-batch.

trianglesphere:
> These are not the case for the span batch due to its data layout. We must rebuild and recompress the entire channel when a block is added because existing contents in the stream are being modified—you cannot change the data that’s already in the stream.

You can estimate the amount of data which would be included in the stream & use that estimate with the target compressed size method.

> I think we can plan for a better design structure for the batch submitter that accommodates both singular and span batches after this PR. What do you think?

I think we should look into splitting the channel builder into two different ones for this PR rather than trying to disentangle them later.

We can and should look at a different design for the batcher to make it easier to handle the grouping of blocks. I do think that when possible, we should start with the improvements to make it easy to the new features.

ImTei:
@trianglesphere Got it! I'll make a better code design asap :)

trianglesphere:
> @trianglesphere Got it! I'll make a better code design asap :)

Sweet. Just to ensure that there's no confusion, for this PR I'm fine with splitting it up to two builders. Larger changes can be part of a later PR.

ImTei:
suggestions are applied!
- [4cbd78e](https://github.com/ethereum-optimism/optimism/pull/7290/commits/4cbd78e79652b3c47341ea1e0bc93ad0589ef7d4): Moving fetching safe head out of the batch submitter constructor
- [f5813ba](https://github.com/ethereum-optimism/optimism/pull/7290/commits/f5813bad3f7c38149b7ecebce3ba72e20ad8c8ad): Splitting ChannelOut into two implementations for each batch type

op-batcher/batcher/driver.go
protolambda: I revisited this, and think we should remove the notion of `lastProcessedBlock` from the channel-manager entirely. The span-batch builder can tell from the L2 blocks themselves whether or not the L1 epoch changed in the first L2 block of the span. (by checking if the "seqnum" of the L1 block info of the block is non-zero).

protolambda:
Changes:
- Rebased upon the derivation PR, which was rebased on develop, and is close to merge-ready. This rebase includes the batcher-service cleanup, so I had to move the "batch type" config value and make a minor change to some config init code.
- Added a TODO about `lastProcessedBlock`: we can avoid this stateful information in the channel-manager, to make it less error-prone. The span-batch builder should determine the L1-origin changed bit based on the seq-num information in the L2 block itself: then it does not depend on prior state. To do this, we need to change the `AppendSingularBatch` interface to take the L1 block info, which can be parsed from the original L2 block (but not present in the batch, which does not contain the system deposit tx with the L1 info).

protolambda:
The branch of the base was removed accidentally because the PR was merged. Reopened that branch, will update the target of this PR now

protolambda (approved):

PR #7292, opened 2023-09-18
Deploy via Safe

maurelian:
**Description**

Updates the deploy scripts to use a gnosis safe for all initialization calls. 

A new 1 of 1 Safe contract is deployed, and is the owner of the `ProxyAdmin`.

The diff is minimized by introducing utility functions for executing the calls via the Safe.

Generating HH artifacts is complicated by the fact that the `SafeProxy` is deployed in a 
nested CREATE2 call from the `SafeProxyFactory`, which makes some of the necessary data very difficult to retrieve and would require custom code. Since the Safe is not a core part of the OP Stack, the artifacts are not required and we simply skip generating them.

https://github.com/ethereum-optimism/optimism/pull/7292/files#diff-276ae96a873cd59782bbdfe7fb7d37d61157440cd4d3745b7ff8a852076f28cc


packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -818,21 +891,22 @@ contract Deploy is Deployer {
...
     /// @notice Transfer ownership of the DisputeGameFactory contract to the final system owner
     function transferDisputeGameFactoryOwnership() public onlyDevnet broadcast {
         DisputeGameFactory disputeGameFactory = DisputeGameFactory(mustGetAddress("DisputeGameFactoryProxy"));
         address owner = disputeGameFactory.owner();
-        address finalSystemOwner = cfg.finalSystemOwner();
-        if (owner != finalSystemOwner) {
-            disputeGameFactory.transferOwnership(finalSystemOwner);
-            console.log("DisputeGameFactory ownership transferred to: %s", finalSystemOwner);
+
+        address safe = mustGetAddress("Safe");
```
tynes: This will not work in the case we are on a live network, ie we hit the branch where we did not deploy a safe

maurelian: Any suggestions for how that lack of this safe should be handled on a live network?

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -151,6 +160,34 @@ contract Deploy is Deployer {
...
 
+    // @notice Gets the address of the SafeProxyFactory and Safe singleton for use in deploying a new GnosisSafe.
+    function _getSafeFactory() internal returns (SafeProxyFactory safeProxyFactory_, Safe safeSingleton_) {
+        if (_isDevnet() == true) {
+            safeProxyFactory_ = new SafeProxyFactory();
+            safeSingleton_ = new Safe();
+        } else {
+            safeProxyFactory_ = SafeProxyFactory(0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2);
+            safeSingleton_ = Safe(payable(0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552));
+        }
```
tynes: You can solve the problem with `mustGetAddress("Safe");` by calling `save()` here:
```solidity
save(address(safeProxyFactory), "SafeProxyFactory"));
save(address(safeSingleton), "Safe"));
```

packages/contracts-bedrock/scripts/Deploy.s.sol
tynes: I answered the above on how to populate the backend for `mustGetAddress`, it either reads from storage and then falls back to disk if its not found in storage

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -513,16 +550,42 @@ contract Deploy is Deployer {
         require(addressManager.owner() == proxyAdmin);
     }
```
tynes, line 559: nit: natspec on these 2 functions

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -513,16 +550,42 @@ contract Deploy is Deployer {
...
 
+    function _callViaSafe(address _target, bytes memory _data) internal {
+        Safe safe = Safe(mustGetAddress("Safe"));
+
+        bytes memory signature = abi.encodePacked(uint256(uint160(tx.origin)), bytes32(0), uint8(1));
```
tynes: Why `tx.origin`? Is it going to be different than `msg.sender`? I have used `msg.sender` in the past to get the account corresponding to the signing key, ie the `--private-key` argument

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -151,6 +160,34 @@ contract Deploy is Deployer {
...
+    }
+
+    /// @notice Deploy the Safe
+    function deploySafe() public broadcast returns (address addr_) {
+        (SafeProxyFactory safeProxyFactory, Safe safeSingleton) = _getSafeFactory();
+
+        address[] memory signers = new address[](1);
+        signers[0] = msg.sender;
+
+        bytes memory initData = abi.encodeWithSelector(
```
tynes, line 188: Comment for arg names would be great and or a link to the arg names

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -151,6 +160,34 @@ contract Deploy is Deployer {
...
+            Safe.setup.selector, signers, 1, address(0), hex"", address(0), address(0), 0, address(0)
+        );
+        address safe = address(safeProxyFactory.createProxyWithNonce(address(safeSingleton), initData, block.timestamp));
```
tynes, line 191: Using `block.timestamp` here is smart to prevent collisions, otherwise I would recommend to compute the address upfront and see if bytecode already exists and then skip the deployment to make it idempotent. Always deploying a new one is more simple

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "packages/contracts-bedrock/lib/forge-std"]
 	path = packages/contracts-bedrock/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
+[submodule "packages/contracts-bedrock/lib/safe-contracts"]
+	path = packages/contracts-bedrock/lib/safe-contracts
+	url = https://github.com/safe-global/safe-contracts
```
tynes: This is going to need a rebase on the other PR when its merged first

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: I see. 
What I don't get though is how this script could work on a live deployment, unless we update the Safe helper functions to handle a threshold greater than 1.

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -685,24 +756,33 @@ contract Deploy is Deployer {
...
+                _data: abi.encodeCall(ProxyAdmin.setProxyType, (l1CrossDomainMessengerProxy, ProxyAdmin.ProxyType.RESOLVED))
+            });
         }
         require(uint256(proxyAdmin.proxyType(l1CrossDomainMessengerProxy)) == uint256(ProxyAdmin.ProxyType.RESOLVED));
 
         string memory contractName = "OVM_L1CrossDomainMessenger";
         string memory implName = proxyAdmin.implementationName(l1CrossDomainMessenger);
         if (keccak256(bytes(contractName)) != keccak256(bytes(implName))) {
-            proxyAdmin.setImplementationName(l1CrossDomainMessengerProxy, contractName);
+            // proxyAdmin.setImplementationName(l1CrossDomainMessengerProxy, contractName);
```
tynes: This is commented out?

maurelian: deleted

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -101,9 +107,15 @@ contract Deploy is Deployer {
 
     /// @notice Modifier that will only allow a function to be called on devnet.
     modifier onlyDevnet() {
+        if (_isDevnet() == true) {
+            _;
+        }
+    }
+
+    function _isDevnet() internal returns (bool _devnet) {
```
tynes: this confuses me because the underscore is a prefix, i read it incorrectly at first, should be suffix

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -153,6 +165,36 @@ contract Deploy is Deployer {
         deployProtocolVersions();
     }
 
+    // @notice Gets the address of the SafeProxyFactory and Safe singleton for use in deploying a new GnosisSafe.
+    function _getSafeFactory() internal returns (SafeProxyFactory safeProxyFactory_, Safe safeSingleton_) {
+        if (_isDevnet() == true) {
```
tynes: An alternative here would be to check if the code exists and deploy if not but ok with modularizing the is devnet check too

tynes: i feel like we should do the no code check because then otherwise live networks without safe infra will break

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Ack. Need a linter for this.

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: OK, adding that.

packages/contracts-bedrock/scripts/Deploy.s.sol
tynes: I feel like this should be `msg.sender`?

tynes:
This is going to break the "deploy your own op stack chain" guide, we will need to have that guide add a commit hash to checkout before this change. It is broken based on the `create2` change already tbh

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Whoops, I had fixed this on a downstream branch. Fixing here now.

tynes:
This looks good to me! pending CI passing

refcell (approved):

PR #7303, opened 2023-09-19
op-node: Fix race condition while closing OpNode

Inphi:
The `OpNode` closes its data sources during shutdown. If the runtime config reloader goroutine is still using the data sources during shutdown then it will panic. Specifically, when a `limitClient` is used, it will be fail to write an in-flight request to its semaphore after the client is closed.

### DIFF
--- a/op-node/node/node.go
+++ b/op-node/node/node.go
@@ -49,6 +49,9 @@ type OpNode struct {
 	resourcesCtx   context.Context
 	resourcesClose context.CancelFunc
 
+	// Indicates when it's safe to close data sources used by the runtimeConfig bg loader
+	runtimeConfigReloaderDone chan struct{}
+
 	closed atomic.Bool
 }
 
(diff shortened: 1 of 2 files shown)


op-node/node/node.go
```diff
@@ -197,9 +200,8 @@ func (n *OpNode) initRuntimeConfig(ctx context.Context, cfg *Config) error {
 			return l1Head, err
 		}
 
-		n.handleProtocolVersionsUpdate(ctx)
-
-		return l1Head, nil
+		err = n.handleProtocolVersionsUpdate(ctx)
+		return l1Head, err
```
refcell, line 204: Is there a reason to return the `l1Head` on error? It seems more idiomatic (according to golang) to do the following.
```suggestion
		if err := n.handleProtocolVersionsUpdate(ctx); err != nil {
		    return eth.L1BlockRef{}, err
		}

		return l1Head, nil
```

Also asking because above, the `l1Head` is still returned when the `runCfg.Load` call errors on line 200.

Inphi: Yeah. It's not pretty, but the reason is so we can log the last protocol version update l1Head on L230 even/especially when there's a halt error.

refcell: Got it makes sense!

refcell (approved):

PR #7310, opened 2023-09-20
Implement the superchain config contract

maurelian:
**Description**

A rough sketch of the arch of a new `SuperchainConfig` contract.

**TODOs:**

Impl:

- [x] Add getters and setters
- [x] Implement the correct auth for setting each of values in the SuperchainConfig contract
- [x] Ensure storage layout mods in `SystemConfig` are safe

Testing/Docs:

- [x] Get unit tests working
- [x] Update deployment scripts
- [x] Define invariant tests

Open todos: 
- [x] Update config var names with the prefix of the contract they are defined in. 
- [x] add `maxPause` as a config var


packages/contracts-bedrock/src/L1/OptimismPortal.sol
```diff
@@ -83,53 +90,46 @@ contract OptimismPortal is Initializable, ResourceMetering, ISemver {
...
     )
         public
         reinitializer(2)
     {
         l2Sender = Constants.DEFAULT_L2_SENDER;
         l2Oracle = _l2Oracle;
         systemConfig = _systemConfig;
-        guardian = _guardian;
-        paused = _paused;
+        superchainConfig = _superchainConfig;
```
tynes: Is the `guardian()(address)` getter still present?

packages/contracts-bedrock/src/L1/OptimismPortal.sol
```diff
@@ -148,21 +148,7 @@ contract OptimismPortal is Initializable, ResourceMetering, ISemver {
     /// @notice Getter for the Guardian
     /// @custom:legacy
     function GUARDIAN() external view returns (address) {
```
tynes, line 144: We can deprecate `GUARDIAN()(address)` i think

packages/contracts-bedrock/src/L1/OptimismPortal.sol
```diff
@@ -46,10 +47,10 @@ contract OptimismPortal is Initializable, ResourceMetering, ISemver {
     /// @notice A mapping of withdrawal hashes to `ProvenWithdrawal` data.
     mapping(bytes32 => ProvenWithdrawal) public provenWithdrawals;
 
-    /// @notice Determines if cross domain messaging is paused.
-    ///         When set to true, withdrawals are paused.
-    ///         This may be removed in the future.
-    bool public paused;
```
tynes, line 52: Add getter `paused()(bool)` that wraps superchain config

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -70,6 +73,10 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
     /// @notice Storage slot that the batch inbox address is stored at.
     bytes32 public constant BATCH_INBOX_SLOT = bytes32(uint256(keccak256("systemconfig.batchinbox")) - 1);
 
+    /// @notice SuperchainConfig address
+    ///         This can only be set via upgrade.
+    SuperchainConfig public superchainConfig;
+
```
tynes, line 79: needs to go to bottom of list of storage variables

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
```diff
@@ -0,0 +1,122 @@
...
+        reinitializer(2)
+    {
+        __Ownable_init();
+        transferOwnership(_owner);
+
+        initiator = _initiator;
+        vetoer = _vetoer;
+        guardian = _guardian;
+        delay = _delay;
+        paused = false;
```
tynes: This doesn't use the variable passed in but hardcodes `false`?

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
```diff
@@ -0,0 +1,122 @@
...
+        VETOER,
+        GUARDIAN,
+        DELAY,
+        PAUSED,
+        ADD_SEQUENCER,
+        REMOVE_SEQUENCER
+    }
+
+    /// @notice Event version identifier.
+    uint256 public constant VERSION = 0;
```
tynes: do we need this?

maurelian: no will remove.

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
maurelian: fixed.

packages/contracts-bedrock/src/L1/OptimismPortal.sol
maurelian: 'tis.

packages/contracts-bedrock/src/L1/OptimismPortal.sol
```diff
@@ -83,53 +85,47 @@ contract OptimismPortal is Initializable, ResourceMetering, ISemver {
...
-    /// @notice Emitted when the pause is triggered.
-    /// @param account Address of the account triggering the pause.
-    event Paused(address account);
-
-    /// @notice Emitted when the pause is lifted.
-    /// @param account Address of the account triggering the unpause.
-    event Unpaused(address account);
-
+    // TODO: move to a new inheritable contract so that this modifier can be reused in the
+    // xDM and bridge
```
maurelian: This coment is wrong. 
I do still need to expand the pause to the other bridge contracts, but these events don't need to be in them.

packages/contracts-bedrock/src/L1/OptimismPortal.sol
```diff
@@ -147,22 +143,14 @@ contract OptimismPortal is Initializable, ResourceMetering, ISemver {
 
     /// @notice Getter for the Guardian
     /// @custom:legacy
```
tynes, line 143: The lowercase getter is not legacy

tynes: We will want to keep the legacy `GUARDIAN()(address)` getter around unfortunately

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
```diff
@@ -0,0 +1,247 @@
...
+        Types.SequencerKeys[] memory _sequencers
+    )
+        public
+        reinitializer(2)
+    {
+        _setAddress(SYSTEM_OWNER_SLOT, _systemOwner);
+        emit ConfigUpdate(UpdateType.SYSTEM_OWNER, abi.encode(_systemOwner));
+
+        _setAddress(INITIATOR_SLOT, _initiator);
+        emit ConfigUpdate(UpdateType.INITIATOR, abi.encode(_initiator));
```
tynes: The `ConfigUpdate` event in the `ChainConfig` was meant to be consumed by the `op-node`. I don't know how useful the `[type, bytes]` style events are in this context

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
```diff
@@ -0,0 +1,247 @@
...
+        address _vetoer,
+        address _guardian,
+        uint256 _delay,
+        Types.SequencerKeys[] memory _sequencers
+    )
+        public
+        reinitializer(2)
+    {
+        _setAddress(SYSTEM_OWNER_SLOT, _systemOwner);
+        emit ConfigUpdate(UpdateType.SYSTEM_OWNER, abi.encode(_systemOwner));
```
tynes: Prefer internal functions and not emitting the events directly, will help to ensure the events are always emitted if there are changes in the future

```solidity
function _setSystemOwner(address _systemOwner) internal {
  _setAddress(SYSTEM_OWNER_SLOT, _systemOwner);
  emit ConfigUpdate(...);
}
```

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
maurelian: yeah, I'm not attached to this, but I felt like there was a certain "Config Contract" pattern emerging from the `ProtocolVersion` and `SystemConfig` contracts. 

If that's not the case, then IMO we should just get rid of the whole eip1967 layout as well, it brings a ton of boilerplate with it.

packages/contracts-bedrock/src/L1/OptimismPortal.sol
maurelian: OK, I can revert that commit, but are you sure? I was addressing [this comment](https://github.com/ethereum-optimism/optimism/pull/7310#discussion_r1331666133).

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
maurelian: Can do, note however that these variables (with the exception of `paused`) are all updated via upgrade, so we'd be creating 6 new setters that are only used in the initializer.

op-chain-ops/genesis/testdata/l1-deployments.json
```diff
@@ -18,5 +18,7 @@
   "SystemConfig": "0x36bAcDD96F28e1ac0780bB9CbE6e20780840730F",
   "SystemConfigProxy": "0x14065A373936533A0c88b7986CADabDD62d471e6",
   "ProtocolVersions": "0x883C06a27D76B1CEbdf7EB376f5556c355afC8e5",
-  "ProtocolVersionsProxy": "0x3732a4D4Ab006cA4825822baEA1569A107683fa1"
+  "ProtocolVersionsProxy": "0x3732a4D4Ab006cA4825822baEA1569A107683fa1",
+  "SuperchainConfig": "0x883C06a27D76B1CEbdf7EB376f5556c355afC8e5",
+  "SuperchainConfigProxy": "0x3732a4D4Ab006cA4825822baEA1569A107683fa1"
```
maurelian: These are not the correct values. I just put them in to get the correct address from the failing test.

packages/contracts-bedrock/.storage-layout
```diff
@@ -68,28 +68,39 @@
...
 | provenWithdrawals    | mapping(bytes32 => struct OptimismPortal.ProvenWithdrawal) | 52   | 0      | 32    | src/L1/OptimismPortal.sol:OptimismPortal |
-| paused               | bool                                                       | 53   | 0      | 1     | src/L1/OptimismPortal.sol:OptimismPortal |
+| spacer_53_0_1        | bool                                                       | 53   | 0      | 1     | src/L1/OptimismPortal.sol:OptimismPortal |
 | l2Oracle             | contract L2OutputOracle                                    | 53   | 1      | 20    | src/L1/OptimismPortal.sol:OptimismPortal |
 | systemConfig         | contract SystemConfig                                      | 54   | 0      | 20    | src/L1/OptimismPortal.sol:OptimismPortal |
-| guardian             | address                                                    | 55   | 0      | 20    | src/L1/OptimismPortal.sol:OptimismPortal |
+| superchainConfig     | contract SuperchainConfig                                  | 55   | 0      | 20    | src/L1/OptimismPortal.sol:OptimismPortal |
```
maurelian, line 75: I believe this is a safe overwrite to make.

packages/contracts-bedrock/foundry.toml
```diff
@@ -24,6 +24,9 @@ bytecode_hash = 'none'
 build_info = true
 build_info_path = 'artifacts/build-info'
 
+# Accomodate the Deploy script's memory intensive sync() method
+memory_limit = 67108864
```
maurelian, line 28: I was getting a `MemoryLimitOOG` error.

packages/contracts-bedrock/src/L1/L1CrossDomainMessenger.sol
```diff
@@ -53,4 +53,16 @@ contract L1CrossDomainMessenger is CrossDomainMessenger, ISemver {
     function _isUnsafeTarget(address _target) internal view override returns (bool) {
         return _target == address(this) || _target == address(PORTAL);
     }
+
+    /// @inheritdoc CrossDomainMessenger
+    function paused() public view override returns (bool) {
+        return PORTAL.paused();
```
maurelian: PORTAL.paused() is [here](https://github.com/ethereum-optimism/optimism/pull/7310/files#diff-f5f7dc90c748b7d98110fa1cc80f3bedd5aa9f299b691c51b9bb473a20b526caR158), it just calls out to the SuperchainConfig to get the paused status. 

IMO this should be changed to also call to the SystemConfig, but this approach made the refactoring easier so far.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -0,0 +1,270 @@
...
+    /// @notice The delay time in seconds between when an upgrade is initiated and when it can be finalized.
+    ///         It can only be modified by an upgrade.
+    bytes32 public constant DELAY_SLOT = bytes32(uint256(keccak256("superchainConfig.delay")) - 1);
+
+    /// @notice The time until which the system is paused.
+    bytes32 public constant PAUSED_SLOT = bytes32(uint256(keccak256("superchainConfig.paused")) - 1);
+
+    // todo(maurelian): make this time configurable. It was just a lot easier to mock it up by hardcoding it in.
+    /// @notice The maximum time in seconds that the system can be paused for.
+    uint256 public constant maxPause = 1 weeks;
```
maurelian: Need a new config value to make this not constant.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -0,0 +1,270 @@
...
+    /// @notice Emitted when the pause is triggered.
+    event Paused();
+
+    /// @notice Emitted when the pause is lifted.
+    event Unpaused();
+
+    /// @notice Emitted when configuration is updated.
+    /// @param updateType Type of update.
+    /// @param data       Encoded update data.
+    event ConfigUpdate(UpdateType indexed updateType, bytes data);
```
maurelian, line 77: I realize that this event format is used in other contracts because the op-node reads one, but I like it here regardless, and don't think it makes sense to have a different event for each config var being changed.

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -89,25 +92,30 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
...
+    uint256 public startBlock;
+
+    /// @notice SuperchainConfig address
+    ///         This can only be set via upgrade.
+    SuperchainConfig public superchainConfig;
+
+    /// @notice Semantic version.
+    /// @custom:semver 2.0.0
+    string public constant version = "2.0.0";
+
```
maurelian, line 105: This diff moves the ConfigUpdate event below the variable declarations.
It should not cause issues with the storage layout.

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -286,9 +297,32 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
     }
 
     /// @notice Updates the unsafe block signer address. Can only be called by the owner.
+    /// @param _batcherHash New batch hash.
     /// @param _unsafeBlockSigner New unsafe block signer address.
-    function setUnsafeBlockSigner(address _unsafeBlockSigner) external onlyOwner {
+    function setSequencer(bytes32 _batcherHash, address _unsafeBlockSigner) external onlyOwner {
```
maurelian, line 285: Note that there is now only a single `setSequencer` method, rather than separate `setUnsafeBlockSigner` and `setBatcherHash`. 
This is because they are defined in the `SuperchainConfig` allowlist as key pairs, so it does not make sense to separate them out here.

packages/contracts-bedrock/src/libraries/Types.sol
```diff
@@ -67,4 +67,13 @@ library Types {
         uint256 gasLimit;
         bytes data;
     }
+
+    /// @title SequencerKeys
+    /// @notice Struct representing a sequencer.
+    /// @custom:field batcherHash         Identifier for the batcher.
+    /// @custom:field unsafeBlockSigner   Unsafe block signer address
+    struct SequencerKeys {
```
maurelian: I'd like to rename this to SequencerKeyPair

packages/contracts-bedrock/test/CommonTest.t.sol
```diff
@@ -169,58 +170,113 @@ contract L2OutputOracle_Initializer is CommonTest {
     }
 }
 
-contract Portal_Initializer is L2OutputOracle_Initializer {
-    // Test target
-    OptimismPortal internal opImpl;
-    OptimismPortal internal op;
-    SystemConfig systemConfig;
+contract SuperchainConfig_Initializer is L2OutputOracle_Initializer {
```
maurelian, line 173: This diff is hard to read, what I did is:
1. add a new `SuperchainConfig_Initializer` contract
2. move `SystemConfig_Initializer` to this file rather than in `SystemConfig.t.sol`
3. Put both of those before `Portal_Initializer` because it depends on them being deployed already.

packages/contracts-bedrock/test/OptimismPortal.t.sol
```diff
@@ -15,85 +15,30 @@ import { ResourceMetering } from "../src/L1/ResourceMetering.sol";
...
-        op.pause();
-        assertEq(op.paused(), true);
-
-        assertTrue(op.GUARDIAN() != alice);
-        vm.expectRevert("OptimismPortal: only guardian can unpause");
-        vm.prank(alice);
-        op.unpause();
-
-        assertEq(op.paused(), true);
-    }
```
maurelian, line 95: I moved these tests to `SuperchainConfig.t.sol`

packages/contracts-bedrock/test/SystemConfig.t.sol
```diff
@@ -2,76 +2,21 @@
...
-                    })
-                )
-            )
-        );
-
-        sysConf = SystemConfig(address(proxy));
-    }
-}
-
-contract SystemConfig_Initialize_Test is SystemConfig_Init {
```
maurelian, line 74: moved all of this stuff to CommonTest

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -604,11 +644,33 @@ contract Deploy is Deployer {
...
+
+        bytes32 batcherHash = bytes32(uint256(uint160(cfg.batchSenderAddress())));
+
+        Proxy(superchainConfigProxy).upgradeToAndCall({
+            _implementation: superchainConfig,
+            _data: abi.encodeCall(
+                SuperchainConfig.initialize,
+                (
+                    cfg.finalSystemOwner(), // systemOwner
+                    cfg.securityCouncil(), // initiator
```
tynes: It feels like a bit of a leaky abstraction to have security council in the deploy config instead of initiator. Something that I'd like to strive for is consistency in the language. We already have some issues with this, like unsafe block signer vs p2p sequencer key. Since initiator is used in the contract, I'd prefer to keep that language in the deploy config. We can add a comment saying that in practice, it is the security council

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
maurelian: Did this.

op-chain-ops/genesis/config.go
```diff
@@ -204,6 +214,15 @@ type DeployConfig struct {
 	// RequiredProtocolVersion indicates the protocol version that
 	// nodes are recommended to adopt, to stay in sync with the network.
 	RecommendedProtocolVersion params.ProtocolVersion `json:"recommendedProtocolVersion"`
+	// SecurityCouncil represents the address of the Security Council role.
+	SecurityCouncil common.Address `json:"securityCouncil"`
+	// Foundation represents the address of the Founcation role.
```
tynes: nit: typo with `founcation`

op-chain-ops/genesis/testdata/l1-deployments.json
tynes: This test fixture is for json serialization?

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -286,9 +297,32 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
...
         _setUnsafeBlockSigner(_unsafeBlockSigner);
+        _setBatcherHash(_batcherHash);
+
+        Types.SequencerKeys memory _sequencer =
+            Types.SequencerKeys({ unsafeBlockSigner: _unsafeBlockSigner, batcherHash: _batcherHash });
```
maurelian: @tynes Something you may notice is that `setSequencer()` could accept as struct arg, which I think would be cleaner, ie.

```
setSequencer(Types.SequencerKeys memory _sequencer)
```

However when I put it into the interface, the [same type](https://github.com/ethereum-optimism/optimism/pull/7310/files#diff-e463331be751e9c6fa0f44ee6b7534067c0db4728537611f6575b328fe533b26R32-R35) which is currently in `superchainconfig.go` is also added to `systemconfig.go` and I get this error: 

```
./op-bindings/bindings/systemconfig.go:52:6: TypesSequencerKeys redeclared in this block
../op-bindings/bindings/superchainconfig.go:32:6: other declaration of TypesSequencerKeys
```

It's the reason I renamed the import [here](https://github.com/ethereum-optimism/optimism/pull/7310/files#diff-016d692e745a10a098159957d937a3a02c5317f6bf727d24c584f8448e81cfceR6), but that didn't change the name of the struct in the binding. I wonder if you've encountered that before or can think of a workaround?

packages/contracts-bedrock/.storage-layout
tynes: I agree this is safe to overwrite this

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -604,11 +644,33 @@ contract Deploy is Deployer {
         console.log("DisputeGameFactory version: %s", version);
     }
 
+    function initializeSuperchainConfig() public broadcast {
+        address payable superchainConfigProxy = mustGetAddress("SuperchainConfigProxy");
+        address superchainConfig = mustGetAddress("SuperchainConfig");
+
+        bytes32 batcherHash = bytes32(uint256(uint160(cfg.batchSenderAddress())));
+
+        Proxy(superchainConfigProxy).upgradeToAndCall({
```
tynes: This is being done thru the eoa and not thru the safe. If you feel like this is the easiest way to do it then i'm fine but it would be breaking the standard way that other things are initialized

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -62,14 +64,30 @@ contract Deploy is Deployer {
...
+
+    /// @notice Deploy a full system with a new SuperchainConfig
+    function setupSuperchainConfig() public returns (address supConf_) {
+        supConf_ = deploySuperchainConfigProxy();
+        deploySuperchainConfig();
+        initializeSuperchainConfig();
+    }
+
+    /// @notice Deploy a new OP Chain, with an existing SuperchainConfig provided
+    function setupOpChain(address supConf) public {
```
tynes: Technically setting up an op chain doesn't need to deploy the implementations

tynes: We don't need to worry about this now and can eventually follow up with a deploy script that has clean abstractions for:
- creating superchain (deploy superchain config)
- deploying implementations
- deploying proxies
- initializing proxies

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -367,6 +395,19 @@ contract Deploy is Deployer {
         addr_ = address(proxy);
     }
```
tynes, line 403: nit: no natspec here

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -604,11 +644,33 @@ contract Deploy is Deployer {
         console.log("DisputeGameFactory version: %s", version);
     }
 
+    function initializeSuperchainConfig() public broadcast {
```
tynes, line 656: nit: no natspec

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -127,6 +145,16 @@ contract Deploy is Deployer {
         }
     }
 
+    function deploySuperchainConfigProxy() public broadcast returns (address addr_) {
+        Proxy proxy = new Proxy({
+            _admin: msg.sender
```
tynes: It looks like the deployer is the admin of the proxy at the end of the script, we want the proxy admin to be the admin?

packages/contracts-bedrock/src/L1/L1CrossDomainMessenger.sol
```diff
@@ -53,4 +53,16 @@ contract L1CrossDomainMessenger is CrossDomainMessenger, ISemver {
     function _isUnsafeTarget(address _target) internal view override returns (bool) {
         return _target == address(this) || _target == address(PORTAL);
     }
+
```
tynes, line 62: We need minor semver bump here

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -0,0 +1,280 @@
...
+        assembly {
+            sstore(_slot, _address)
+        }
+    }
+
+    /// @notice Returns a uint256 stored in an arbitrary storage slot.
+    ///         These storage slots decouple the storage layout from
+    ///         solc's automation.
+    /// @param _slot The storage slot to retrieve the address from.
+    function _getValue(bytes32 _slot) internal view returns (uint256 value_) {
```
tynes: Perhaps a follow up would be to turn these sorts of getters into a library because we have them defined in multiple contracts

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -0,0 +1,280 @@
...
+
+    /// @notice Getter for the guardian address.
+    function guardian() public view returns (address guardian_) {
+        guardian_ = _getAddress(GUARDIAN_SLOT);
+    }
+
+    /// @notice Getter for the delay address.
+    function delay() public view returns (uint256 delay_) {
+        // We do some casting rather than define a new getter.
+        delay_ = uint256(uint160(_getAddress(DELAY_SLOT)));
```
tynes: This seems bug prone if the delay is more than 160 bits, unlikely in practice but imo these sorts of shortcuts create the habits that lead to bugs

packages/contracts-bedrock/src/libraries/Types.sol
tynes: I am on board with this

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Fixed

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: yeah, totally agree.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: hmmm, looks like I already have a getValue() I can use there anyways.

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Good call out. I think that likely will need to be resolved before merging back into `develop`. Going to leave this comment open for now.

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Fixed. The `superchainConfigProxy` is now deployed with the Safe as it's admin.

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: I've set these back to the more standard names

packages/contracts-bedrock/src/L1/SystemConfig.sol
tynes: I did have a problem similar to this when doing the gas config in the system config/resource metering contracts. I can't remember exactly how I handled it to be honest. It is trying to generate the type twice in the same package, this is a problem with how we namespace our bindings

packages/contracts-bedrock/scripts/DeployConfig.s.sol
```diff
@@ -50,6 +51,10 @@ contract DeployConfig is Script {
     uint256 public systemConfigStartBlock;
     uint256 public requiredProtocolVersion;
     uint256 public recommendedProtocolVersion;
+    address public updateInitiator;
+    address public updateVetoer;
+    uint256 public updateDelay;
+    Types.SequencerKeys[] public sequencerKeys;
```
tynes: Did you still want to rename this to sequencer key pair?

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -0,0 +1,238 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Initializable } from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import { ISemver } from "src/universal/ISemver.sol";
+import { Types as T } from "src/libraries/Types.sol";
```
tynes: With the import causing bindings issues, in the future we can use the `abigen`  flag for excluding types, its `abigen -exc` but that would require a change to the bindings build script/config. If you would like to take it on to get the preferable API, that is cool but also its all good if not

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -0,0 +1,238 @@
...
+        reinitializer(2)
+    {
+        _setSystemOwner(_systemOwner);
+        _setInitiator(_initiator);
+        _setVetoer(_vetoer);
+        _setGuardian(_guardian);
+        _setDelay(_delay);
+
+        for (uint256 i = 0; i < _sequencers.length; i++) {
+            bytes32 sequencerHash = Hashing.hashSequencerKeys(_sequencers[i]);
```
tynes: Curious if you think this should call `_addSequencer` and then make `addSequencer` have the auth check

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /// @notice Getter for the delay address.
+    function delay() public view returns (uint256 delay_) {
+        delay_ = Slot.getUint(DELAY_SLOT);
+    }
+
+    /// @notice Getter for the paused address.
+    function paused() public view returns (bool paused_) {
+        paused_ = Slot.getUint(PAUSED_SLOT) > block.timestamp;
```
tynes, line 161: Should this be `>` or `>=`?

packages/contracts-bedrock/src/L1/SystemConfig.sol
tynes: Ah i understand, its trying to generate the type that is the input to the function and its not smart enough to know that the type is already generated in a different file

tynes: The exclude functionality on abigen could work, there is also a combined json flag on abigen that we don't use and i wonder if that would help, altho idk how to get the combined json out of forge plus it would prob include the test files and stuff

tynes: My sense for now its something that we just leave as is and then we solve it sometime in the future

maurelian: Would it be too brittle to find and delete the duplicate type def from one of the bindings at the end of generation?

packages/contracts-bedrock/src/l1/SuperchainConfig.sol
maurelian: @tynes thinking about it more, I do like this format, whether or not the op-node reads it. 
Config contracts are inevitably going to get more and more values, having to add a new event for each one is onerous, and doesn't help off-chain code that much. 

Please LMK if you still wnat me to change it.

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Resolving now.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: I think we need to aggressively cut scope here, resolving :)

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: Yes, I like that, implemented.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: I think `>` because that way it is paused until the end up `duration`.

packages/contracts-bedrock/src/libraries/Types.sol
maurelian: I have done so

packages/contracts-bedrock/src/L1/SystemConfig.sol
maurelian: Going to resolve this. In a future clean up I might handle this with --exc

op-chain-ops/genesis/config.go
```diff
@@ -31,6 +31,12 @@ var (
 	ErrInvalidImmutablesConfig = errors.New("invalid immutables config")
 )
 
+// SequencerKeys is an auto generated low-level Go binding around an user-defined struct.
```
tynes: Auto generated? I feel like this should say something along the lines of

```
SequencerKeyPair represents the identity of a particular sequencer
```

op-chain-ops/genesis/config.go
```diff
@@ -204,6 +210,17 @@ type DeployConfig struct {
...
+	// SuperchainConfigVetoer represents the address of the SuperchainConfigVetoer role.
+	SuperchainConfigVetoer common.Address `json:"superchainConfigVetoer"`
+	// SuperchainConfigDelay represents the time after which an upgrade can occur ifit
+	SuperchainConfigDelay uint64 `json:"superchainConfigDelay"`
+	// SuperchainConfigMaxPause represents the time after which an upgrade can occur ifit
+	SuperchainConfigMaxPause uint64 `json:"superchainConfigMaxPause"`
+	// SequencerKeys represents the set of allowed Sequencer key pairs (batcher hash and unsafe
+	// block signer).
+	SuperchainConfigSequencerKeyPairs []SequencerKeyPair `json:"superchainConfigSequencerKeyPairs"`
 }
```
tynes, line 224: We should be sure to either log warnings in `Check` or error. If they must be set, then we should error

packages/contracts-bedrock/src/L1/L1CrossDomainMessenger.sol
tynes: I agree that we should not chain the calls of paused thru the portal, do you plan on updating this?

packages/contracts-bedrock/src/L1/OptimismPortal.sol
tynes: Just want to make sure that this todo is removed

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -0,0 +1,260 @@
...
+    function maxPause() public view returns (uint256 maxPause_) {
+        maxPause_ = Slot.getUint(MAX_PAUSE_SLOT);
+    }
+
+    /// @notice Getter for the paused address.
+    function paused() public view returns (bool paused_) {
+        paused_ = Slot.getUint(PAUSED_SLOT) > block.timestamp;
+    }
+
+    // todo(maurelian): we might need a repause() getter which is only callable by the
```
tynes: Following up on this `pause`/`repause` - what is the thinking here?

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -167,20 +178,21 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
     {
         __Ownable_init();
         transferOwnership(_owner);
+        superchainConfig = _superchainConfig;
 
         // These are set in ascending order of their UpdateTypes.
         _setBatcherHash(_batcherHash);
         _setGasConfig({ _overhead: _overhead, _scalar: _scalar });
         _setGasLimit(_gasLimit);
         _setUnsafeBlockSigner(_unsafeBlockSigner);
```
tynes, line 186: The above setters emit events, the contracts being stored do not need to emit events so it is safe to use `Slot` directly altho i thought it was renamed to `Storage` in another PR?

packages/contracts-bedrock/src/universal/CrossDomainMessenger.sol
```diff
@@ -219,6 +219,10 @@ abstract contract CrossDomainMessenger is
         external
         payable
     {
+        // On L1 this function will check the Portal for its paused status.
+        // On L2 this function should be a no-op.
+        _revertWhenPaused();
```
tynes: Is this necessary? Why can't we do:

```solidity
require(paused() == false, "OptimismPortal: paused");
```

maurelian: My thinking was that calling a function the just returns is more likely to be optimized away than this.
I'm not certain either way, might do a little sandbox test, or else just take your suggestion.

packages/contracts-bedrock/src/L1/SystemConfig.sol
maurelian: I put that PR on `develop` to minimize drift between it and this branch. Now I'm not sure it's working out that way.

op-chain-ops/genesis/testdata/l1-deployments.json
maurelian: Yes. All it does is check that the addresses are non-zero. https://github.com/ethereum-optimism/optimism/blob/develop/op-chain-ops/genesis/testdata/l1-deployments.json#L20-L21

packages/contracts-bedrock/src/L1/L1CrossDomainMessenger.sol
maurelian: Yes, I noticed this open item this morning.

packages/contracts-bedrock/src/L1/OptimismPortal.sol
maurelian: yeah, not worth adding inheritance IMO.

op-chain-ops/genesis/config.go
maurelian: adding that

packages/contracts-bedrock/src/L1/SystemConfig.sol
maurelian: I implemented this using the Slot lib.

packages/contracts-bedrock/src/universal/CrossDomainMessenger.sol
maurelian: I verified this.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: We do need this, but I would prefer to include it in a follow up PR just to get this closed and so that that can happen in the context of a smaller diff.

packages/contracts-bedrock/src/L1/OptimismPortal.sol
tynes: why are you resolving these comments and not removing the todo?

tynes:
I'd like to merge https://github.com/ethereum-optimism/optimism/pull/7428 and then rebase on top of it

packages/contracts-bedrock/src/L1/OptimismPortal.sol
maurelian: it's removed now

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -286,26 +280,43 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
...
+    }
+
+    /// @notice Checks the SuperchainConfig's allow list for the current sequencer. If it is not allowed,
+    ///         the sequencer is removed. Anyone may call this function.
+    function checkSequencer() external {
+        Types.SequencerKeyPair memory sequencer =
+            Types.SequencerKeyPair({ unsafeBlockSigner: unsafeBlockSigner(), batcherHash: batcherHash });
+        bytes32 seqHash = Hashing.hashSequencerKeyPair(sequencer);
+        if (SuperchainConfig(superchainConfig()).allowedSequencers(seqHash)) {
+            revert("SystemConfig: cannot remove allowed sequencer.");
```
tynes, line 305: Why a revert here with a string instead of a require?

packages/contracts-bedrock/src/libraries/Slot.sol
```diff
@@ -0,0 +1,69 @@
...
+    ///         These storage slots decouple the storage layout from
+    ///         solc's automation.
+    /// @param _slot The storage slot to retrieve the address from.
+    function getUint(bytes32 _slot) internal view returns (uint256 value_) {
+        assembly {
+            value_ := sload(_slot)
+        }
+    }
+
+    /// @notice Stores a value in an arbitrary storage slot, `_slot`.
```
tynes, line 38: Ideally we can get the changes to this contract in https://github.com/ethereum-optimism/optimism/pull/7428

packages/contracts-bedrock/src/L1/SystemConfig.sol
maurelian: I find this framing to be more readable, ie. saying "if it is allowed, then stop the check", rather than "require that the sequencer is not allowed".
There are other occurences already of the if/revert pattern in our codebase, so IMO readability is sufficient to justify this.

packages/contracts-bedrock/src/libraries/Slot.sol
maurelian: Those changes won't be on this branch.
If 7428 does get merged in, will this PR be blocked until the feature branch is rebased?

packages/contracts-bedrock/.storage-layout
```diff
@@ -6,26 +6,27 @@
...
+| spacer_202_0_32    | mapping(bytes32 => bool)  | 202  | 0      | 32    | src/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| successfulMessages | mapping(bytes32 => bool)  | 203  | 0      | 32    | src/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| xDomainMsgSender   | address                   | 204  | 0      | 20    | src/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| msgNonce           | uint240                   | 205  | 0      | 30    | src/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| failedMessages     | mapping(bytes32 => bool)  | 206  | 0      | 32    | src/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| __gap              | uint256[42]               | 207  | 0      | 1344  | src/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| PORTAL             | contract OptimismPortal   | 249  | 0      | 20    | src/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
+| superchainConfig   | contract SuperchainConfig | 250  | 0      | 20    | src/L1/L1CrossDomainMessenger.sol:L1CrossDomainMessenger |
```
maurelian, line 29: big diff due to the column width change. The only real diff is the addition of the new superchainConfig var at the end

tynes (approved):

refcell:
FYI - adding the https://github.com/ethereum-optimism/optimism/labels/C-urgent label to test if merge queues still isn't working as intend.

refcell:
NVM - base branch isn't develop

PR #7029, opened 2023-08-28
Rpc client cleanup

Rcontre360:
**Description**

RPC client code organization cleanup. Moved package op-node/client to op-service/client. Changed import names on every filed where is used. Moved DialRollupClientWithTimeout to op-node/sources to solve circular imports issue.

**Tests**

none

**Additional context**

none

**Metadata**

- Fixes #[https://github.com/ethereum-optimism/optimism/issues/6936]

PR #7311, opened 2023-09-20
feat(op-node/op-batcher/op-proposer): add L1 fallbackClient

welkin22:
**Description**

The FallbackClient is designed to automatically switch to an alternative L1 endpoint when the current one encounters issues, and revert back once the primary endpoint is functional again.It can reduce the instability of the L2 chain caused by L1 endpoint instability.

The core logic operates as follows: When an error occurs, it is recorded until either the error count surpasses a predetermined threshold or the count is reset by a timer that executes once per minute. If the threshold is exceeded within a one-minute span, it suggests the L1 endpoint has become unreliable, prompting the system to select the next address in the L1 URL. A new RPC client is then created to replace the current one. Simultaneously, a separate Goroutine monitors the initial endpoint's health and, when deemed stable, reverts back to its use.

As the clients employed in op-node and op-batcher/op-proposer differ significantly, two distinct fallbackClients have been implemented. The op-node version incorporates additional features such as subscription management and RPC legality checks, resulting in a more complex implementation.

Example of l1 flag after code modification:
```bash
--l1=https://addr1.com:8545,https://addr2.com:8545,https://addr3.com:8545
```

**Tests**
I have tested it in the local environment. I am going to add an e2e case, but I encountered some problems when executing the  `make devnet-allocs` command.I'm still trying to solve it

### DIFF
--- a/op-batcher/flags/flags.go
+++ b/op-batcher/flags/flags.go
@@ -24,7 +24,7 @@ func prefixEnvVars(name string) []string {
 
 var (
 	// Required flags
-	L1EthRpcFlag = &cli.StringFlag{
+	L1EthRpcFlag = &cli.StringSliceFlag{
 		Name:    "l1-eth-rpc",
 		Usage:   "HTTP provider URL for L1. Multiple alternative addresses are supported, separated by commas, and the first address is used by default",
 		EnvVars: prefixEnvVars("L1_ETH_RPC"),
--- a/op-batcher/flags/flags.go
+++ b/op-batcher/flags/flags.go
@@ -26,7 +26,7 @@ var (
 	// Required flags
 	L1EthRpcFlag = &cli.StringFlag{
 		Name:    "l1-eth-rpc",
-		Usage:   "HTTP provider URL for L1",
+		Usage:   "HTTP provider URL for L1. Multiple alternative addresses are supported, separated by commas, and the first address is used by default",
 		EnvVars: prefixEnvVars("L1_ETH_RPC"),
 	}
 	L2EthRpcFlag = &cli.StringFlag{
--- a/op-batcher/flags/flags.go
+++ b/op-batcher/flags/flags.go
@@ -24,9 +24,9 @@ func prefixEnvVars(name string) []string {
 
 var (
 	// Required flags
-	L1EthRpcFlag = &cli.StringSliceFlag{
+	L1EthRpcFlag = &cli.StringFlag{
 		Name:    "l1-eth-rpc",
-		Usage:   "HTTP provider URL for L1. Multiple alternative addresses are supported, separated by commas, and the first address is used by default",
+		Usage:   "HTTP provider URL for L1",
 		EnvVars: prefixEnvVars("L1_ETH_RPC"),
 	}
 	L2EthRpcFlag = &cli.StringFlag{
--- a/op-node/sources/fallback_client.go
+++ b/op-node/sources/fallback_client.go
@@ -42,22 +42,33 @@ func NewFallbackClientMetrics(ns string, factory opmetrics.Factory) *FallbackCli
 // and automatically switch back after the first endpoint recovers.
 type FallbackClient struct {
 	// firstRpc is created by the first of the urls, it should be used first in a healthy state
-	firstRpc          client.RPC
-	urlList           []string
-	rpcInitFunc       func(url string) (client.RPC, error)
-	lastMinuteFail    atomic.Int64
-	currentRpc        atomic.Pointer[client.RPC]
-	currentIndex      int
-	mx                sync.Mutex
-	log               log.Logger
+	firstRpc    client.RPC
+	urlList     []string
+	rpcInitFunc func(url string) (client.RPC, error)
+	// lastMinuteFail is used to record the number of errors in the last minute
+	lastMinuteFail atomic.Int64
+	// currentRpc always points to the rpc currently being used
+	currentRpc atomic.Pointer[client.RPC]
+	// currentIndex is the index of the current rpc in the urlList
+	currentIndex int
+	mx           sync.Mutex
+	log          log.Logger
+	// isInFallbackState is used to record whether the current rpc is in a fallback state,
+	// Used to ensure that only one recoverIfFirstRpcHealth process is started at the same time
 	isInFallbackState bool
-	subscribeFunc     func() (event.Subscription, error)
-	headsSub          *ethereum.Subscription
-	chainId           *big.Int
-	genesisBlock      eth.BlockID
-	ctx               context.Context
-	isClose           chan struct{}
-	metrics           FallbackClientMetricer
+	// subscribeFunc is used when switching to an alternative rpc, we need to re-subscribe to the content subscribed by
+	//the previous rpc, such as the subscription action in WatchHeadChanges, which needs to be triggered again
+	subscribeFunc func() (event.Subscription, error)
+	// headsSub is used to record the subscription of the previous rpc, so that we can unsubscribe when switching to
+	// the new rpc
+	headsSub *ethereum.Subscription
+	// chainId and genesisBlock are used to check whether the newly switched rpc is legal.
+	chainId      *big.Int
+	genesisBlock eth.BlockID
+	ctx          context.Context
+	// isClose is used to close the goroutine that monitors the number of errors in the last minute
+	isClose chan struct{}
+	metrics FallbackClientMetricer
 }
 
 const threshold int64 = 20
@@ -76,6 +87,7 @@ func NewFallbackClient(ctx context.Context, rpc client.RPC, urlList []string, lo
 		genesisBlock: genesisBlock,
 	}
 	fallbackClient.currentRpc.Store(&rpc)
+	// monitor the number of errors in the last minute
 	go func() {
 		ticker := time.NewTicker(1 * time.Minute)
 		for {
@@ -86,6 +98,7 @@ func NewFallbackClient(ctx context.Context, rpc client.RPC, urlList []string, lo
 			case <-fallbackClient.isClose:
 				return
 			default:
+				// if the number of errors in the last minute exceeds the threshold, switch to the next rpc
 				if fallbackClient.lastMinuteFail.Load() >= threshold {
 					fallbackClient.switchCurrentRpc()
 				}
@@ -142,6 +155,7 @@ func (l *FallbackClient) EthSubscribe(ctx context.Context, channel any, args ...
 	return subscribe, err
 }
 
+// switchCurrentRpc switches to the next rpc
 func (l *FallbackClient) switchCurrentRpc() {
 	if l.currentIndex >= len(l.urlList) {
 		l.log.Error("the fallback client has tried all urls, but all failed")
@@ -149,12 +163,11 @@ func (l *FallbackClient) switchCurrentRpc() {
 	}
 	l.mx.Lock()
 	defer l.mx.Unlock()
+	// double check to avoid switching to the next rpc at the same time
 	if l.lastMinuteFail.Load() <= threshold {
 		return
 	}
-	if l.metrics != nil {
-		l.metrics.RecordUrlSwitchEvt(strconv.Itoa(l.currentIndex))
-	}
+	// iterate through the urlList to find the next available rpc
 	for {
 		l.currentIndex++
 		if l.currentIndex >= len(l.urlList) {
@@ -165,6 +178,9 @@ func (l *FallbackClient) switchCurrentRpc() {
 		if err != nil {
 			l.log.Warn("the fallback client failed to switch the current client", "err", err)
 		} else {
+			if l.metrics != nil {
+				l.metrics.RecordUrlSwitchEvt(strconv.Itoa(l.currentIndex))
+			}
 			break
 		}
 	}
@@ -181,10 +197,13 @@ func (l *FallbackClient) switchCurrentRpcLogic() error {
 		return vErr
 	}
 	lastRpc := *l.currentRpc.Load()
+	// switch to the new rpc
 	l.currentRpc.Store(&newRpc)
+	// we don't close first rpc, the first rpc need to be recovered when it is healthy
 	if lastRpc != l.firstRpc {
 		lastRpc.Close()
 	}
+	// clear the number of errors in the last minute
 	l.lastMinuteFail.Store(0)
 	if l.subscribeFunc != nil {
 		err := l.reSubscribeNewRpc(url)
@@ -193,6 +212,7 @@ func (l *FallbackClient) switchCurrentRpcLogic() error {
 		}
 	}
 	l.log.Info("switched current rpc to new url", "url", url)
+	// start the process of recovering the first rpc if it has not been started
 	if !l.isInFallbackState {
 		l.isInFallbackState = true
 		l.recoverIfFirstRpcHealth()
@@ -200,6 +220,7 @@ func (l *FallbackClient) switchCurrentRpcLogic() error {
 	return nil
 }
 
+// reSubscribeNewRpc re-subscribe to the content subscribed by the previous rpc and unsubscribe the previous rpc
 func (l *FallbackClient) reSubscribeNewRpc(url string) error {
 	(*l.headsSub).Unsubscribe()
 	subscriptionNew, err := l.subscribeFunc()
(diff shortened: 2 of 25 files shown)


op-service/client/dial.go
```diff
@@ -62,3 +66,50 @@ func dialRPCClientWithBackoff(ctx context.Context, log log.Logger, addr string)
...
+			}
+			return ethClientNew, nil
+		})
+		return fallbackClient, nil
+	}
+
+	return DialEthClientWithTimeout(timeout, l, url)
+}
+
+type EthClient interface {
```
tynes, line 99: Doesn't this interface already exist in go-ethereum?

tynes:
This seems like its more lines of code than necessary for this kind of functionality, there are also very few comments. Perhaps aligning on the desire to have a fallback RPC client first is a better way to approach opening pull requests

op-batcher/flags/flags.go
```diff
@@ -26,7 +26,7 @@ var (
 	// Required flags
 	L1EthRpcFlag = &cli.StringFlag{
```
trianglesphere: A string slice flag would be better than manually parsing strings.

op-node/sources/fallback_client.go
```diff
@@ -0,0 +1,280 @@
...
+	"math/big"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// FallbackClient is an RPC client, it can automatically switch to the next l1 endpoint
+// when there is a problem with the current l1 endpoint
+// and automatically switch back after the first l1 endpoint recovers.
+type FallbackClient struct {
```
trianglesphere, line 43: This may be conflicting with https://github.com/ethereum-optimism/optimism/pull/7029, but the fallback client should be layered onto the client & be easy to create.

trianglesphere (commented):
I understand the desire for this feature, but I am incredibly worried about consistency between different RPCs. Different RPC providers will not be in sync & these components are designed to handle re-orgs, but may not handle data inconsistencies other than re-orgs.

In addition, when a re-org is detected, a significant amount of work has to be done in the op-batcher/op-node so it is important than this client does not show something like a re-org occuring.

op-service/client/dial.go
welkin22: Hi, I couldn't find this interface in go-ethereum. I noticed that there is an ETHBackend interface in op-service/txmgr/txmgr.go:63. Also notice that op-node/client/client.go:15 has a Client interface. Neither of these interfaces can be used directly, so I imitated them and wrote my own interface. Could you please tell me the location of the interface in go-ethereum?

welkin22:
> This seems like its more lines of code than necessary for this kind of functionality, there are also very few comments. Perhaps aligning on the desire to have a fallback RPC client first is a better way to approach opening pull requests

@tynes Hi, I believe that all the code is meant to implement a fallback client. The code appears to have more lines than anticipated because it attempts to merge the fallback client code with the existing code. For example, the places where *ethclient.Client is referenced are changed to ethereum.ChainStateReader, as only by replacing it with an interface can different implementations be introduced.
Sorry about the few comments, I will add more comments to help understand the code.

welkin22:
> I understand the desire for this feature, but I am incredibly worried about consistency between different RPCs. Different RPC providers will not be in sync & these components are designed to handle re-orgs, but may not handle data inconsistencies other than re-orgs.
> 
> In addition, when a re-org is detected, a significant amount of work has to be done in the op-batcher/op-node so it is important than this client does not show something like a re-org occuring.

@trianglesphere Hi, I understand your concerns and I also considered this issue at the beginning of the design. Perhaps we can expand on this concern and discuss it.First of all, the l1 endpoint we fill in is most of the time a node we trust. The inconsistency in its data will only be the difference in the latest block height, and other inconsistencies rarely occur. Secondly, I have not found any examples in the code of op-node and op-batcher where data inconsistency may cause problems. It seems that there are necessary checking and retry mechanisms everywhere. Compared with the serious situation where the endpoint of l1 is paralyzed and cannot be accessed, causing the block to be unable to be produced, triggering the re-org logic and retry mechanism seems to be worthwhile.
If you could further point out a specific data inconsistency problem case, I would be very grateful.

op-batcher/flags/flags.go
welkin22: Yes I will try to modify this

op-node/sources/fallback_client.go
welkin22: It seems that #7029 was not merged, but rather closed. Right now, I can't be sure what it will turn out to be like.

> the fallback client should be layered onto the client & be easy to create.

I tried to achieve this goal, but since the fallback client needs to create multiple underlying RPC clients instead of just adding new functions to a single client, the situation is a bit more complicated. Currently I try to implement the client.RPC interface and wrap the original client.NewRPC when multiple addresses are configured. See [fallbackClientWrap ](https://github.com/ethereum-optimism/optimism/pull/7311/files#diff-a27156a67289c4da44b8746353b897cde1e4fb77737af85838ad9ca91923ef12R197) method.
This is the easiest way I can think of to create it. If you have any better ideas, please let me know.

tynes:
> > This seems like its more lines of code than necessary for this kind of functionality, there are also very few comments. Perhaps aligning on the desire to have a fallback RPC client first is a better way to approach opening pull requests
> 
> @tynes Hi, I believe that all the code is meant to implement a fallback client. The code appears to have more lines than anticipated because it attempts to merge the fallback client code with the existing code. For example, the places where *ethclient.Client is referenced are changed to ethereum.ChainStateReader, as only by replacing it with an interface can different implementations be introduced. Sorry about the few comments, I will add more comments to help understand the code.

What I mean here is that I believe your implementation needs to be simplified before considered for merge. For example, there are many fields that can be removed from this struct

```go
type FallbackClient struct {
	firstRpc          client.RPC
	urlList           []string
	rpcInitFunc       func(url string) (client.RPC, error)
	lastMinuteFail    atomic.Int64
	currentRpc        atomic.Pointer[client.RPC]
	currentIndex      int
	mx                sync.Mutex
	log               log.Logger
	isInFallbackState bool
	subscribeFunc     func() (event.Subscription, error)
	l1HeadsSub        *ethereum.Subscription
	l1ChainId         *big.Int
	l1Block           eth.BlockID
	ctx               context.Context
	isClose           chan struct{}
	metrics           metrics.Metricer
}
```

Why does it need `firstRPC` and `currentRPC`? Why does it need `isInFallbackState`? Why does it have a `l1ChainId` and a `l1Block`? Having all of these makes the internals more complex than it needs to be.

I have a preference for the functionality to be as follows:
- upon construction, make an ethclient for each url
- when caller calls the client:
  - for each client, submit the request
  - compare the results
  - return error if mismatch
  - ignore backend errors/timeouts

This ensures that the backends are consistent altho it does use a linear number of requests. This feature is more useful for power users and power users can run their own nodes. What do you think @trianglesphere ?

tynes:
> For example, the places where *ethclient.Client is referenced are changed to ethereum.ChainStateReader, as only by replacing it with an interface can different implementations be introduced.

I am supportive of this kind of change and it is something that could easily be broken out into its own PR, making things overall easier to review/merge

op-node/sources/fallback_client.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rpc"
+	"math/big"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// FallbackClient is an RPC client, it can automatically switch to the next l1 endpoint
```
tynes: nit: it shouldn't care if its L1 or not L1

op-service/client/dial.go
tynes: Are you able to use these interfaces? https://github.com/ethereum/go-ethereum/blob/545f4c5547178bc8bde6af08b3ccaf68ca27f2c0/interfaces.go

op-node/sources/fallback_client.go
```diff
@@ -0,0 +1,280 @@
...
+	ctx               context.Context
+	isClose           chan struct{}
+	metrics           metrics.Metricer
+}
+
+const threshold int64 = 20
+
+// NewFallbackClient returns a new FallbackClient. l1ChainId and l1Block are used to check
+// whether the newly switched rpc is legal.
+func NewFallbackClient(ctx context.Context, rpc client.RPC, urlList []string, log log.Logger, l1ChainId *big.Int, l1Block eth.BlockID, rpcInitFunc func(url string) (client.RPC, error)) client.RPC {
```
tynes: Do we need 2 implementations or is it possible to use a single impl from `op-service`?

op-node/sources/fallback_client.go
```diff
@@ -0,0 +1,280 @@
...
+		if l.subscribeFunc != nil {
+			err := l.reSubscribeNewRpc(l.urlList[0])
+			if err != nil {
+				l.log.Error("can not subscribe new url", "url", l.urlList[0], "err", err)
+			}
+		}
+		l.log.Info("recover the current rpc to the first rpc", "url", l.urlList[0])
+	}()
+}
+
```
tynes, line 279: Can you add godocs to the functions?

welkin22:
> I have a preference for the functionality to be as follows:
> 
> * upon construction, make an ethclient for each url
> * when caller calls the client:
>   
>   * for each client, submit the request
>   * compare the results
>   * return error if mismatch
>   * ignore backend errors/timeouts
> 
> This ensures that the backends are consistent altho it does use a linear number of requests. This feature is more useful for power users and power users can run their own nodes. What do you think @trianglesphere ?

@tynes Hi, I thought about the method you mentioned. It is indeed simple enough, but there are some problems that need to be solved:
1. It is very difficult to compare the return value of rpc. After all, result is an any type. Is there any way to quickly compare the return values? Will the process of comparing return values slow down the overall request speed?
2. Multiple endpoints are requested at the same time. If the performance of one of them is poor (but less than the timeout period), it will affect the performance of the entire request.
3. If all endpoints fail, but the returned errors are different, which one should I use?
If all of the above issues can be resolved, I would be willing to re-implement a new version following this approach.

> Why does it need `firstRPC` and `currentRPC`? Why does it need `isInFallbackState`? Why does it have a `l1ChainId` and a `l1Block`? Having all of these makes the internals more complex than it needs to be.
> 
firstRPC is the rpc that is used daily. Both `currentRPC` and `isInFallbackState` are used to manage the state when temporarily switching to an alternative rpc. The current design idea is that firstRPC will not fail all the time, but will only fail temporarily for a period of time. It usually works normally. We only need to switch to the alternative RPC during the failure. When firstRPC recovers, it will automatically switch back.
a `l1ChainId` and a `l1Block` are used for checking the L1 endpoint. I noticed that `ValidateL1Config` has this logic, so when switching to a new RPC, this check should also be performed.

op-node/sources/fallback_client.go
welkin22: ok, I would add them

op-batcher/flags/flags.go
welkin22: resolved

op-node/sources/fallback_client.go
welkin22: Tried it before but had circular dependency issues here, I'll try it again

op-node/sources/fallback_client.go
welkin22: removed

welkin22:
> > For example, the places where *ethclient.Client is referenced are changed to ethereum.ChainStateReader, as only by replacing it with an interface can different implementations be introduced.
> 
> I am supportive of this kind of change and it is something that could easily be broken out into its own PR, making things overall easier to review/merge

Indeed. I would try splitting into two PRs, one containing the fallback client itself, and one trying to integrate the fallback client into other code

op-service/client/dial.go
welkin22: I can't use this because its interface is too fragmented. For example, in the batcher, both `HeaderByNumber` and `SendTransaction` are required, but they come from two different interfaces. I cannot return multiple interfaces in `DialEthClientWithTimeoutAndFallback`. I need to return a unified EthClient.

tynes:
> > > For example, the places where *ethclient.Client is referenced are changed to ethereum.ChainStateReader, as only by replacing it with an interface can different implementations be introduced.
> > 
> > 
> > I am supportive of this kind of change and it is something that could easily be broken out into its own PR, making things overall easier to review/merge
> 
> Indeed. I would try splitting into two PRs, one containing the fallback client itself, and one trying to integrate the fallback client into other code

Would very much appreciate this approach :) thank you!

For comparing type `any`, shouldn't `reflect.DeepEqual` work? Also I was thinking of the `ethclient.Client` interface but with fallback support under the hood. https://github.com/ethereum/go-ethereum/blob/master/ethclient/ethclient.go

tynes:
> > I have a preference for the functionality to be as follows:
> > 
> > * upon construction, make an ethclient for each url
> > * when caller calls the client:
> >   
> >   * for each client, submit the request
> >   * compare the results
> >   * return error if mismatch
> >   * ignore backend errors/timeouts
> > 
> > This ensures that the backends are consistent altho it does use a linear number of requests. This feature is more useful for power users and power users can run their own nodes. What do you think @trianglesphere ?
> 
> @tynes Hi, I thought about the method you mentioned. It is indeed simple enough, but there are some problems that need to be solved:
> 
> 1. It is very difficult to compare the return value of rpc. After all, result is an any type. Is there any way to quickly compare the return values? Will the process of comparing return values slow down the overall request speed?
> 2. Multiple endpoints are requested at the same time. If the performance of one of them is poor (but less than the timeout period), it will affect the performance of the entire request.
> 3. If all endpoints fail, but the returned errors are different, which one should I use?
>    If all of the above issues can be resolved, I would be willing to re-implement a new version following this approach.
> 
> > Why does it need `firstRPC` and `currentRPC`? Why does it need `isInFallbackState`? Why does it have a `l1ChainId` and a `l1Block`? Having all of these makes the internals more complex than it needs to be.
> 
> firstRPC is the rpc that is used daily. Both `currentRPC` and `isInFallbackState` are used to manage the state when temporarily switching to an alternative rpc. The current design idea is that firstRPC will not fail all the time, but will only fail temporarily for a period of time. It usually works normally. We only need to switch to the alternative RPC during the failure. When firstRPC recovers, it will automatically switch back. a `l1ChainId` and a `l1Block` are used for checking the L1 endpoint. I noticed that `ValidateL1Config` has this logic, so when switching to a new RPC, this check should also be performed.

I would like to get some more feedback from my peers as well because this is a big design space, will follow up

welkin22:
> > > > For example, the places where *ethclient.Client is referenced are changed to ethereum.ChainStateReader, as only by replacing it with an interface can different implementations be introduced.
> > > 
> > > 
> > > I am supportive of this kind of change and it is something that could easily be broken out into its own PR, making things overall easier to review/merge
> > 
> > 
> > Indeed. I would try splitting into two PRs, one containing the fallback client itself, and one trying to integrate the fallback client into other code
> 
> Would very much appreciate this approach :) thank you!
> 
> For comparing type `any`, shouldn't `reflect.DeepEqual` work? Also I was thinking of the `ethclient.Client` interface but with fallback support under the hood. https://github.com/ethereum/go-ethereum/blob/master/ethclient/ethclient.go

@tynes Hi, I have deleted the code except fallbackClient and added some comments. Now we can focus on the discussion of fallbackClient implementation.
One problem is that I still can't write fallbackClient only in op-service. This is mainly because it is difficult to find a unified interface entry point. In op-node, if I want to replace the implementation, I can only start with the client.RPC interface. Everything else is a structure. If the implementation of the structure is replaced, all references to the structure must be modified. The Client in ethclient.go you mentioned is still a structure, and its implementation cannot be easily replaced. The same situation exists in op-batcher and op-proposer, so I chose to define the EthClient interface myself. But if I want to use this interface in op-node, I still need to modify a lot of code.

welkin22:
> > I understand the desire for this feature, but I am incredibly worried about consistency between different RPCs. Different RPC providers will not be in sync & these components are designed to handle re-orgs, but may not handle data inconsistencies other than re-orgs.
> > In addition, when a re-org is detected, a significant amount of work has to be done in the op-batcher/op-node so it is important than this client does not show something like a re-org occuring.
> 
> @trianglesphere Hi, I understand your concerns and I also considered this issue at the beginning of the design. Perhaps we can expand on this concern and discuss it.First of all, the l1 endpoint we fill in is most of the time a node we trust. The inconsistency in its data will only be the difference in the latest block height, and other inconsistencies rarely occur. Secondly, I have not found any examples in the code of op-node and op-batcher where data inconsistency may cause problems. It seems that there are necessary checking and retry mechanisms everywhere. Compared with the serious situation where the endpoint of l1 is paralyzed and cannot be accessed, causing the block to be unable to be produced, triggering the re-org logic and retry mechanism seems to be worthwhile. If you could further point out a specific data inconsistency problem case, I would be very grateful.

Sorry, I found that it is still necessary to avoid re-org . I will try to let the fallback client avoid re-org .

sebastianst:
Converted to draft until we have decided on how to proceed and tests are added.

PR #7320, opened 2023-09-20
Init Metamask Pass/Fail metrics

Spacesai1or:
This PR adds two counter metrics for the Metamask Test Service:

1. Number of successful transactions since last failure
2. Number of failed transaction since last success

A preconfigured Grafana dashboard was also added, and is available within the Grafana instance spun up via `docker-compose`


Spacesai1or:
* develop
  * **PR #7320** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7320" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7320?utm_source=stack-comment).

ufm-test-services/grafana/dashboards/metamask.json
```diff
@@ -0,0 +1,182 @@
...
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
```
nickbalestra, line 12: Should we have some share const where we pull different colors, so that is easy to configure a new  annotation?

ufm-test-services/grafana/dashboards/metamask.json
```diff
@@ -0,0 +1,182 @@
...
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 0,
+  "links": [],
+  "liveNow": false,
+  "panels": [
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "PBFA97CFB590B2093"
```
nickbalestra, line 27: Should we put those UID into a const file , perhaps as enums so that we can pull from there ? I see we already use it multiple place in this file and that could be a source of misconfuguration?

ufm-test-services/metamask/package.json
```diff
@@ -25,5 +25,9 @@
     "static-server": "^2.2.1",
     "typescript": "^5.1.6",
     "viem": "^1.10.8"
+  },
+  "dependencies": {
+    "node-fetch": "^3.3.2",
```
nickbalestra: what version of node are we running that we need node-fetch?

roninjin10: We could use `--experimental-fetch` in node 16. I believe @tynes has plans to upgrade us to node 18 in near future (if it hasn't already happened)

ufm-test-services/grafana/dashboards/metamask.json
roninjin10: This would be very nice if grafana supoprts this

ufm-test-services/grafana/dashboards/metamask.json
```diff
@@ -0,0 +1,182 @@
...
+    "from": "now-6h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "UFM: Metamask",
+  "uid": "f66f7076-c724-4f81-8ff9-58d6d99f2716",
+  "version": 1,
+  "weekStart": ""
+}
```
roninjin10:
```suggestion
}

```
Nit eol

ufm-test-services/metamask/tests/metamask.spec.ts
```diff
@@ -79,8 +96,8 @@ test('Send an EIP-1559 transaciton and verfiy success', async () => {
   await metamask.confirmTransaction()
   const txHash = await txHashPromise
 
-    // Waiting for Infura (Metamask given provider) to index our transaction
-    await sharedPage.waitForTimeout(10_000)
+  // Waiting for Infura (Metamask given provider) to index our transaction
+  await sharedPage.waitForTimeout(10_000)
```
roninjin10: Not from this pr but can we poll instead of waiting a hardcoded amount

ufm-test-services/metamask/tests/prometheusUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    return undefined
+  }
+}
+
+export const setMetamaskTxCounter = async (
+  isSuccess: boolean,
+  valueToSetTo: number
+) => {
+  if (process.env.PROMETHEUS_PUSHGATEWAY_URL === undefined)
+    throw new Error('Please provied PROMETHEUS_PUSHGATEWAY_URL in .env')
```
roninjin10: This is fine but I prefer using zod for this
```typescript
const prometheusGatewayUrl = z.string().url().parse(process.env.PROMETHEUS_PUSHGATEWAY_URL)
```
Or if you have multiple env variables
```
const env = z.object({
  PROMETHEUS_PUSHGATEWAY_URL: z.string().url()
  ...
}).parse(process.env)
```

ufm-test-services/metamask/tests/prometheusUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+  help: 'A counter signifying the number of failed transactions sent with Metamask since last successful transaction',
+})
+
+export const getMetricValueFromPrometheus = async (metricName: string) => {
+  if (process.env.PROMETHEUS_SERVER_URL === undefined)
+    throw new Error('Please provied PROMETHEUS_SERVER_URL in .env')
+
+  const response = await fetch(
+    `${process.env.PROMETHEUS_SERVER_URL}/api/v1/query?query=${metricName}`
+  )
```
roninjin10: not handling response.ok !== true and throwing a useful error

ufm-test-services/metamask/tests/prometheusUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+  const reponseJson = (await response.json()) as {
+    status: string
+    data: { resultType: string; result: any[] }
+  }
```
roninjin10:
```suggestion
  const reponseJson = z.object({
    status: z.string()
    data: z.object({ resultType: z.string(); result: z.array(z.any()) })
  }).parse(await response.json())
```
This is another place I would use zod

ufm-test-services/metamask/tests/prometheusUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    reponseJson.data &&
+    reponseJson.data.result &&
+    reponseJson.data.result.length > 0
+  ) {
+    return parseFloat(reponseJson.data.result[0].value[1])
+  } else {
+    console.warn(`No data found for metric ${metricName} in Prometheus`)
+    return undefined
+  }
+}
```
roninjin10: If you use zod this would already be validated. You can also use .transform to parse the float

roninjin10 (approved):
Lots of comments but it's a WIP so lgtm

ufm-test-services/grafana/dashboards/metamask.json
Spacesai1or: I setup the dashboard using Grafana and this file is an export of that dashboard - it shouldn't need to be modified by hand, and is probably better suited to be managed within Grafana's UI if changes need to be made

Specifically for this `datasource`, Grafana is being configured to use it by [this](https://github.com/ethereum-optimism/optimism/blob/wyatt/ufm/init-mm-metrics/ufm-test-services/grafana/provisioning/datasources/datasources.yml) file and [docker-compose.yml](https://github.com/ethereum-optimism/optimism/blob/933bedb46842584a969db4832bf95272dc1a2760/ufm-test-services/docker-compose.yml#L38), so the `uid` should always be in-sync since it's managed by these processes. I've tested completely tearing down the system and re-building the Docker images, and everything works fine with the `uid`s remaining static

ufm-test-services/grafana/dashboards/metamask.json
Spacesai1or: I don't think we need to do anything fancy with this file as it's generated by Grafana. This color is the default red that I selected to be used by the metrics through the Grafana dashboard editor UI. So, any metric that chooses to use red will default to use the same color

ufm-test-services/metamask/package.json
Spacesai1or: Previously the monorepo's `.nvmrc` was using `v16`, but it's now using `v18` - I'll update and remove `node-fetch`

nickbalestra (approved):

ufm-test-services/metamask/tests/prometheusUtils.ts
```diff
@@ -0,0 +1,98 @@
...
+  help: 'A counter signifying the number of failed transactions sent with Metamask since last successful transaction',
+})
+
+export const getMetamaskTxCounterValue = async (isSuccess: boolean) => {
+  const metricName = isSuccess ? txSuccessMetricName : txFailureMetricName
+  const prometheusMetricQuery = `${env.PROMETHEUS_SERVER_URL}/api/v1/query?query=${metricName}`
+
+  const response = await fetch(prometheusMetricQuery)
+  if (!response.ok)
+    throw new Error(`Failed to fetch metric from: ${prometheusMetricQuery}`)
```
roninjin10, line 32:
```suggestion
    throw new Error(`Failed to fetch metric from: ${prometheusMetricQuery}`)
```
console.error teh status code and metric

roninjin10: status and response.message

PR #7321, opened 2023-09-20
feat(op-challenger): Deep Positions

refcell:
**Description**

Upgrades position indices to use an arbitrary sized `big.Int` to support games with a depth greater than 63.

This PR is the start to adding support, other `Position` api changes should be stacked ontop, including cleaning up convenience helper functions, aliasing `big.Int` as a `PositionIndex` type to further abstract those convenience helper methods, etc, etc.

This PR is needed to support split games where the output root bisection game tree "stacks ontop" of classical game bisection.

**Tests**

Unit tests with large generalized indices > 64 bits.

**Metadata**

Progresses https://github.com/ethereum-optimism/client-pod/issues/38

### DIFF
--- a/op-challenger/game/fault/solver/game_solver_test.go
+++ b/op-challenger/game/fault/solver/game_solver_test.go
@@ -56,7 +56,6 @@ func TestCalculateNextActions(t *testing.T) {
 				builder.Seq().AttackCorrect()
 			},
 		},
-
 		{
 			name:                "RespondToAllClaimsAtDisagreeingLevel",
 			agreeWithOutputRoot: true,
@@ -70,7 +69,6 @@ func TestCalculateNextActions(t *testing.T) {
 				honestClaim.Defend(common.Hash{0xdd}).ExpectAttack()
 			},
 		},
-
 		{
 			name:                "StepAtMaxDepth",
 			agreeWithOutputRoot: true,
@@ -83,7 +81,6 @@ func TestCalculateNextActions(t *testing.T) {
 				lastHonestClaim.Attack(common.Hash{0xdd}).ExpectStepAttack()
 			},
 		},
-
 		{
 			name:                "PoisonedPreState",
 			agreeWithOutputRoot: true,
--- a/op-challenger/game/fault/solver/solver_test.go
+++ b/op-challenger/game/fault/solver/solver_test.go
@@ -2,6 +2,7 @@ package solver
 
 import (
 	"context"
+	"math/big"
 	"testing"
 
 	faulttest "github.com/ethereum-optimism/optimism/op-challenger/game/fault/test"
@@ -16,7 +17,8 @@ func TestAttemptStep(t *testing.T) {
 
 	// Last accessible leaf is the second last trace index
 	// The root node is used for the last trace index and can only be attacked.
-	lastLeafTraceIndex := uint64(1<<maxDepth - 2)
+	lastLeafTraceIndex := big.NewInt(1<<maxDepth - 2)
+	lastLeafTraceIndexPlusOne := big.NewInt(1<<maxDepth - 1)
 	ctx := context.Background()
 
 	tests := []struct {
@@ -32,9 +34,9 @@ func TestAttemptStep(t *testing.T) {
 		{
 			name:               "AttackFirstTraceIndex",
 			expectAttack:       true,
-			expectPreState:     claimBuilder.CorrectPreState(0),
-			expectProofData:    claimBuilder.CorrectProofData(0),
-			expectedOracleData: claimBuilder.CorrectOracleData(0),
+			expectPreState:     claimBuilder.CorrectPreState(common.Big0),
+			expectProofData:    claimBuilder.CorrectProofData(common.Big0),
+			expectedOracleData: claimBuilder.CorrectOracleData(common.Big0),
 			setupGame: func(builder *faulttest.GameBuilder) {
 				builder.Seq().
 					Attack(common.Hash{0xaa}).
@@ -45,9 +47,9 @@ func TestAttemptStep(t *testing.T) {
 		{
 			name:               "DefendFirstTraceIndex",
 			expectAttack:       false,
-			expectPreState:     claimBuilder.CorrectPreState(1),
-			expectProofData:    claimBuilder.CorrectProofData(1),
-			expectedOracleData: claimBuilder.CorrectOracleData(1),
+			expectPreState:     claimBuilder.CorrectPreState(big.NewInt(1)),
+			expectProofData:    claimBuilder.CorrectProofData(big.NewInt(1)),
+			expectedOracleData: claimBuilder.CorrectOracleData(big.NewInt(1)),
 			setupGame: func(builder *faulttest.GameBuilder) {
 				builder.Seq().
 					Attack(common.Hash{0xaa}).
@@ -58,9 +60,9 @@ func TestAttemptStep(t *testing.T) {
 		{
 			name:               "AttackMiddleTraceIndex",
 			expectAttack:       true,
-			expectPreState:     claimBuilder.CorrectPreState(4),
-			expectProofData:    claimBuilder.CorrectProofData(4),
-			expectedOracleData: claimBuilder.CorrectOracleData(4),
+			expectPreState:     claimBuilder.CorrectPreState(big.NewInt(4)),
+			expectProofData:    claimBuilder.CorrectProofData(big.NewInt(4)),
+			expectedOracleData: claimBuilder.CorrectOracleData(big.NewInt(4)),
 			setupGame: func(builder *faulttest.GameBuilder) {
 				builder.Seq().
 					AttackCorrect().
@@ -71,9 +73,9 @@ func TestAttemptStep(t *testing.T) {
 		{
 			name:               "DefendMiddleTraceIndex",
 			expectAttack:       false,
-			expectPreState:     claimBuilder.CorrectPreState(5),
-			expectProofData:    claimBuilder.CorrectProofData(5),
-			expectedOracleData: claimBuilder.CorrectOracleData(5),
+			expectPreState:     claimBuilder.CorrectPreState(big.NewInt(5)),
+			expectProofData:    claimBuilder.CorrectProofData(big.NewInt(5)),
+			expectedOracleData: claimBuilder.CorrectOracleData(big.NewInt(5)),
 			setupGame: func(builder *faulttest.GameBuilder) {
 				builder.Seq().
 					AttackCorrect().
@@ -97,9 +99,9 @@ func TestAttemptStep(t *testing.T) {
 		{
 			name:               "DefendLastTraceIndex",
 			expectAttack:       false,
-			expectPreState:     claimBuilder.CorrectPreState(lastLeafTraceIndex + 1),
-			expectProofData:    claimBuilder.CorrectProofData(lastLeafTraceIndex + 1),
-			expectedOracleData: claimBuilder.CorrectOracleData(lastLeafTraceIndex + 1),
+			expectPreState:     claimBuilder.CorrectPreState(lastLeafTraceIndexPlusOne),
+			expectProofData:    claimBuilder.CorrectProofData(lastLeafTraceIndexPlusOne),
+			expectedOracleData: claimBuilder.CorrectOracleData(lastLeafTraceIndexPlusOne),
 			setupGame: func(builder *faulttest.GameBuilder) {
 				builder.Seq().
 					AttackCorrect().
@@ -140,9 +142,9 @@ func TestAttemptStep(t *testing.T) {
 		{
 			name:               "CannotStepNearlyValidPath",
 			expectAttack:       true,
-			expectPreState:     claimBuilder.CorrectPreState(4),
-			expectProofData:    claimBuilder.CorrectProofData(4),
-			expectedOracleData: claimBuilder.CorrectOracleData(4),
+			expectPreState:     claimBuilder.CorrectPreState(big.NewInt(4)),
+			expectProofData:    claimBuilder.CorrectProofData(big.NewInt(4)),
+			expectedOracleData: claimBuilder.CorrectOracleData(big.NewInt(4)),
 			setupGame: func(builder *faulttest.GameBuilder) {
 				builder.Seq().
 					AttackCorrect().
--- a/op-challenger/game/fault/test/claim_builder.go
+++ b/op-challenger/game/fault/test/claim_builder.go
@@ -39,30 +39,30 @@ func (c *ClaimBuilder) CorrectClaimAtPosition(pos types.Position) common.Hash {
 }
 
 // CorrectPreState returns the pre-state (not hashed) required to execute the valid step at the specified trace index
-func (c *ClaimBuilder) CorrectPreState(idx uint64) []byte {
-	pos := types.NewPosition(c.maxDepth, int(idx))
+func (c *ClaimBuilder) CorrectPreState(idx *big.Int) []byte {
+	pos := types.NewPosition(c.maxDepth, idx)
 	preimage, _, _, err := c.correct.GetStepData(context.Background(), pos)
 	c.require.NoError(err)
 	return preimage
 }
 
 // CorrectProofData returns the proof-data required to execute the valid step at the specified trace index
-func (c *ClaimBuilder) CorrectProofData(idx uint64) []byte {
-	pos := types.NewPosition(c.maxDepth, int(idx))
+func (c *ClaimBuilder) CorrectProofData(idx *big.Int) []byte {
+	pos := types.NewPosition(c.maxDepth, idx)
 	_, proof, _, err := c.correct.GetStepData(context.Background(), pos)
 	c.require.NoError(err)
 	return proof
 }
 
-func (c *ClaimBuilder) CorrectOracleData(idx uint64) *types.PreimageOracleData {
-	pos := types.NewPosition(c.maxDepth, int(idx))
+func (c *ClaimBuilder) CorrectOracleData(idx *big.Int) *types.PreimageOracleData {
+	pos := types.NewPosition(c.maxDepth, idx)
 	_, _, data, err := c.correct.GetStepData(context.Background(), pos)
 	c.require.NoError(err)
 	return data
 }
 
 func (c *ClaimBuilder) incorrectClaim(pos types.Position) common.Hash {
-	return common.BigToHash(new(big.Int).SetUint64(pos.TraceIndex(c.maxDepth)))
+	return common.BigToHash(pos.TraceIndex(c.maxDepth))
 }
 
 func (c *ClaimBuilder) claim(pos types.Position, correct bool) common.Hash {
@@ -78,15 +78,15 @@ func (c *ClaimBuilder) CreateRootClaim(correct bool) types.Claim {
 	claim := types.Claim{
 		ClaimData: types.ClaimData{
 			Value:    value,
-			Position: types.NewPosition(0, 0),
+			Position: types.NewPosition(0, common.Big0),
 		},
 	}
 	return claim
 }
 
-func (c *ClaimBuilder) CreateLeafClaim(traceIndex uint64, correct bool) types.Claim {
-	parentPos := types.NewPosition(c.maxDepth-1, 0)
-	pos := types.NewPosition(c.maxDepth, int(traceIndex))
+func (c *ClaimBuilder) CreateLeafClaim(traceIndex *big.Int, correct bool) types.Claim {
+	parentPos := types.NewPosition(c.maxDepth-1, common.Big0)
+	pos := types.NewPosition(c.maxDepth, traceIndex)
 	return types.Claim{
 		ClaimData: types.ClaimData{
 			Value:    c.claim(pos, correct),
--- a/op-challenger/game/fault/test/game_builder.go
+++ b/op-challenger/game/fault/test/game_builder.go
@@ -1,6 +1,8 @@
 package test
 
 import (
+	"math/big"
+
 	"github.com/ethereum-optimism/optimism/op-challenger/game/fault/types"
 	"github.com/ethereum/go-ethereum/common"
 )
@@ -120,7 +122,7 @@ func (s *GameBuilderSeq) ExpectStepAttack() *GameBuilderSeq {
 }
 
 func (s *GameBuilderSeq) ExpectStepDefend() *GameBuilderSeq {
-	traceIdx := s.lastClaim.TraceIndex(s.builder.maxDepth) + 1
+	traceIdx := new(big.Int).Add(s.lastClaim.TraceIndex(s.builder.maxDepth), big.NewInt(1))
 	s.gameBuilder.ExpectedActions = append(s.gameBuilder.ExpectedActions, types.Action{
 		Type:       types.ActionTypeStep,
 		ParentIdx:  s.lastClaim.ContractIndex,
--- a/op-challenger/game/fault/trace/alphabet/provider.go
+++ b/op-challenger/game/fault/trace/alphabet/provider.go
@@ -35,7 +35,7 @@ func NewTraceProvider(state string, depth uint64) *AlphabetTraceProvider {
 
 func (ap *AlphabetTraceProvider) GetStepData(ctx context.Context, i types.Position) ([]byte, []byte, *types.PreimageOracleData, error) {
 	traceIndex := i.TraceIndex(int(ap.depth))
-	if traceIndex == 0 {
+	if traceIndex.Cmp(common.Big0) == 0 {
 		prestate, err := ap.AbsolutePreState(ctx)
 		if err != nil {
 			return nil, nil, nil, err
@@ -43,22 +43,23 @@ func (ap *AlphabetTraceProvider) GetStepData(ctx context.Context, i types.Positi
 		return prestate, []byte{}, nil, nil
 	}
 	// We want the pre-state which is the value prior to the one requested
-	traceIndex--
+	traceIndex = traceIndex.Sub(traceIndex, big.NewInt(1))
 	// The index cannot be larger than the maximum index as computed by the depth.
-	if traceIndex >= ap.maxLen {
+	if traceIndex.Cmp(big.NewInt(int64(ap.maxLen))) >= 0 {
 		return nil, nil, nil, ErrIndexTooLarge
 	}
 	// We extend the deepest hash to the maximum depth if the trace is not expansive.
-	if traceIndex >= uint64(len(ap.state)) {
-		return ap.GetStepData(ctx, types.NewPosition(int(ap.depth), len(ap.state)))
+	if traceIndex.Cmp(big.NewInt(int64(len(ap.state)))) >= 0 {
+		return ap.GetStepData(ctx, types.NewPosition(int(ap.depth), big.NewInt(int64(len(ap.state)))))
 	}
-	return BuildAlphabetPreimage(traceIndex, ap.state[traceIndex]), []byte{}, nil, nil
+	return BuildAlphabetPreimage(traceIndex, ap.state[traceIndex.Uint64()]), []byte{}, nil, nil
 }
 
 // Get returns the claim value at the given index in the trace.
 func (ap *AlphabetTraceProvider) Get(ctx context.Context, i types.Position) (common.Hash, error) {
 	// Step data returns the pre-state, so add 1 to get the state for index i
-	postPosition := types.NewPosition(int(ap.depth), int(i.TraceIndex(int(ap.depth)))+1)
+	ti := i.TraceIndex(int(ap.depth))
+	postPosition := types.NewPosition(int(ap.depth), new(big.Int).Add(ti, big.NewInt(1)))
 	claimBytes, _, _, err := ap.GetStepData(ctx, postPosition)
 	if err != nil {
 		return common.Hash{}, err
@@ -82,22 +83,14 @@ func (ap *AlphabetTraceProvider) AbsolutePreStateCommitment(ctx context.Context)
 }
 
 // BuildAlphabetPreimage constructs the claim bytes for the index and state item.
-func BuildAlphabetPreimage(i uint64, letter string) []byte {
-	return append(IndexToBytes(i), LetterToBytes(letter)...)
+func BuildAlphabetPreimage(i *big.Int, letter string) []byte {
+	return append(i.FillBytes(make([]byte, 32)), LetterToBytes(letter)...)
 }
 
 func alphabetStateHash(state []byte) common.Hash {
 	h := crypto.Keccak256Hash(state)
 	h[0] = mipsevm.VMStatusInvalid
 	return h
-}
-
-// IndexToBytes converts an index to a byte slice big endian
-func IndexToBytes(i uint64) []byte {
-	big := new(big.Int)
-	big.SetUint64(i)
-	out := make([]byte, 32)
-	return big.FillBytes(out)
 }
 
 // LetterToBytes converts a letter to a 32 byte array
--- a/op-challenger/game/fault/trace/cannon/provider_test.go
+++ b/op-challenger/game/fault/trace/cannon/provider_test.go
@@ -6,6 +6,7 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
+	"math/big"
 	"os"
 	"path/filepath"
 	"testing"
@@ -22,7 +23,7 @@ import (
 //go:embed test_data
 var testData embed.FS
 
-func PositionFromTraceIndex(provider *CannonTraceProvider, idx int) types.Position {
+func PositionFromTraceIndex(provider *CannonTraceProvider, idx *big.Int) types.Position {
 	return types.NewPosition(int(provider.gameDepth), idx)
 }
 
@@ -30,7 +31,7 @@ func TestGet(t *testing.T) {
 	dataDir, prestate := setupTestData(t)
 	t.Run("ExistingProof", func(t *testing.T) {
 		provider, generator := setupWithTestData(t, dataDir, prestate)
-		value, err := provider.Get(context.Background(), PositionFromTraceIndex(provider, 0))
+		value, err := provider.Get(context.Background(), PositionFromTraceIndex(provider, common.Big0))
 		require.NoError(t, err)
 		require.Equal(t, common.HexToHash("0x45fd9aa59768331c726e719e76aa343e73123af888804604785ae19506e65e87"), value)
 		require.Empty(t, generator.generated)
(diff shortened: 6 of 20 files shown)


refcell:
* develop
  * **PR #7321** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7321" target="_blank">Graphite</a>  👈
    * **PR #7354** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7354" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7321?utm_source=stack-comment).

op-challenger/game/fault/types/position.go
```diff
@@ -1,15 +1,21 @@
...
 // Position is a golang wrapper around the dispute game Position type.
 type Position struct {
 	depth        int
-	indexAtDepth int
+	indexAtDepth *big.Int
 }
 
-func NewPosition(depth, indexAtDepth int) Position {
-	return Position{depth, indexAtDepth}
+func NewPosition(depth int, indexAtDepth int) Position {
```
ajsutton: `indexAtDepth` probably should be a `big.Int` here for consistency.  We want to switch everything over to always assuming indexAtDepth is arbitrarily large - otherwise we risk missing a case that we think was safe but actually breaks stuff.

op-challenger/game/fault/types/game.go
```diff
@@ -167,7 +170,8 @@ func (g *gameState) getParent(claim Claim) *extendedClaim {
 
 func makeClaimEntry(claim Claim) claimEntry {
 	return claimEntry{
-		ClaimData:           claim.ClaimData,
+		GIndex:              claim.ToGIndex().Uint64(),
```
ajsutton: Doesn't this break the large index support? We're still using a uint64 for a position.  Why aren't we still using `ClaimData` here?

op-challenger/game/fault/types/position_test.go
```diff
@@ -85,7 +85,7 @@ func TestGINConversions(t *testing.T) {
 		pos := NewPosition(test.Depth, test.IndexAtDepth)
 		require.Equal(t, pos, from)
 		to := pos.ToGIndex()
-		require.Equal(t, test.GIndex, to)
+		require.Equal(t, test.GIndex, to.Uint64())
```
ajsutton: We need to add cases to each of these unit tests that use positions where the indexAtDepth is greater than the max uint64 value to confirm each individual function still returns the right result.

op-challenger/game/fault/types/position.go
```diff
@@ -36,14 +49,16 @@ func (p Position) TraceIndex(maxDepth int) uint64 {
 	// When we go right, we do a shift left and set the bottom bit to be 1.
 	// To do this in a single step, do all the shifts at once & or in all 1s for the bottom bits.
 	rd := maxDepth - p.depth
-	return uint64(p.indexAtDepth<<rd | ((1 << rd) - 1))
+	rhs := ((1 << rd) - 1)
+	ti := new(big.Int).Or(p.lshIndex(rd), big.NewInt(int64(rhs)))
+	return ti.Uint64()
```
ajsutton: This overflows for large depths.

ie this test fails:
```
func TestTraceIndexBigValue(t *testing.T) {
	traceIdx := new(big.Int).Mul(new(big.Int).SetUint64(math.MaxUint64), big.NewInt(4))
	pos := Position{depth: 100, indexAtDepth: traceIdx}
	actual := pos.TraceIndex(100)
	require.Equal(t, traceIdx, actual)
}
```

`TraceIndex` will have to return `big.Int` not `uint64`

op-challenger/game/fault/types/position.go
```diff
@@ -55,11 +70,19 @@ func boolToInt(b bool) int {
 	}
 }
 
+func (p Position) parentIndexAtDepth() *big.Int {
+	return big.NewInt(0).Div(p.IndexAtDepth(), big.NewInt(2))
+}
+
+func (p Position) DefendsParent(parentIndex *big.Int) bool {
+	return p.parentIndexAtDepth().Cmp(parentIndex) != 0
```
ajsutton: A position can't defend or attack its parent - claims attack or defend their parent by moving to a child or a different subtree. We're really asking if this position is to the right of the supplied position.

I'd also say we should pass the whole `Position` as argument rather than exposing the implementation details of it using index at depth.  So we'd get an api of:

```go
func (p Position) RightOf(pos Position) bool {
```

ajsutton (commented):
My suggestion would be to make this first PR get `Position` to a point where it completely works with large depths.  To avoid having to update every caller to also use `*big.Int` and flow that around the entire codebase in the same PR, I'd add a couple of helper methods for compatibility that cast things back to uint64.  Only ever use those methods to do the cast and then we'll know we're done when we can delete those methods.  E.g. we'd have:

```go
func (p Position) TraceIndex(depth int) *big.Int {
 // Nice, fully working code
}

func (p Position) EvilTraceIndexUInt64(depth int) uint64 {
  return p.TraceIndex(depth).UInt64()
}
```

And nothing else should ever call `TraceIndex(...).UInt64()` to do the cast itself.  Alternatively we could just have the one `TraceIndex` and a more generic conversion function like:
```go
func EvilUInt64(val *big.Int) {
  return val.UInt64()
}
```
Then we can use that for any cast we need and still easily find usages of it and work towards deleting it by the time we're done with this sequence of work.  Either way the idea is to make it very clear that it's not safe to just call `UInt64()` to extract a value from `big.Int` and let us track where we're doing that unsafe cast just because we want to make this change in smaller steps.

op-challenger/game/fault/types/position.go
trianglesphere: So this is pulling from the old `func (c *Claim) DefendsParent() bool {`. I think that implementing `(*Claim) DefendsParent` in terms of `(Position) RightOf` would work & make the most sense.

op-challenger/game/fault/types/game.go
refcell: Using the `ClaimData` object itself has a `*big.Int` pointer internally so comparisons between the objects become flaky since the internals of the `big.Int` can be nil if not instantiated properly

ajsutton: hmm, but we can't use a uint64 here as GIndex will need to exceed its max value. I'd be concerned if we're not instantiating objects properly - that's going to lead to all kinds of trouble regardless of whether you get a `nil` or a `0` since 0 is an invalid GIndex anyway.

op-challenger/game/fault/types/position.go
refcell: Introduced a new temporary `NewPositionFromUint64` method to minimize the diff here since there are quite a lot of places that construct new positions and importing and creating new big ints increases the diff.

In a stacked follow-on pr, will remove it.

op-challenger/game/fault/types/game.go
refcell: Ok I was having issues with the big ints in tests cases that compare the objects - it's that pesky big int internal NaN representation vs nil

refcell: So, dug into this a little bit more.

The main change here was that we now use a `*big.Int` pointer inside the `Position` struct for the `IndexAtDepth` value.

Then the `claimEntry` embeds a `ClaimData` internally which itself has a `Position` embedded. When using the `claimEntry` as a key in the `claims` mapping, the struct is _comparable_ because the big int pointer is comparable inside the position. It is not if the big int is not a pointer.

So the reason why using `ClaimData` fails with this big int pointer, is that checking duplicate claims fails since a a new `claimEntry` will be constructed for the duplicate claim and can hold a different big int pointer internally. Thus the `IsDuplicate` check fails.

I've added a game test to demonstrate this using a deep clone of the `ClaimData`.

Since we cannot use a `big.Int` as it is not comparable and does not work as a key in the map, I think the best way around this is to build a canonical claim identifier for the claim state. I have done a naive version of this in this pr by hashing the claim entry values. It's not the most efficient but it works.

Would appreciate some feedback on how you want to proceed with this. It's not clear to me what approach we should take to get this pr over the line.

refcell:
Apologies in advance that this PR is quite sizable @ajsutton - much of the added diff is due to altering the position tests to extend them to deep positions. Only added one in there so as to not increase the diff even further here.

op-challenger/game/fault/solver/game_solver_test.go
```diff
@@ -113,7 +110,7 @@ func TestCalculateNextActions(t *testing.T) {
 			game := builder.Game
 			for i, claim := range game.Claims() {
 				t.Logf("Claim %v: Pos: %v TraceIdx: %v ParentIdx: %v, Countered: %v, Value: %v",
-					i, claim.Position.ToGIndex(), claim.Position.TraceIndex(maxDepth), claim.ParentContractIndex, claim.Countered, claim.Value)
+					i, claim.Position.ToGIndex().Uint64(), claim.Position.TraceIndex(maxDepth), claim.ParentContractIndex, claim.Countered, claim.Value)
```
ajsutton: No need for `.Uint64()` here - `big.Int` works fine in a log statement.

op-challenger/game/fault/test/game_builder.go
```diff
@@ -112,22 +114,22 @@ func (s *GameBuilderSeq) ExpectStepAttack() *GameBuilderSeq {
 		Type:       types.ActionTypeStep,
 		ParentIdx:  s.lastClaim.ContractIndex,
 		IsAttack:   true,
-		PreState:   s.builder.CorrectPreState(traceIdx),
-		ProofData:  s.builder.CorrectProofData(traceIdx),
-		OracleData: s.builder.CorrectOracleData(traceIdx),
+		PreState:   s.builder.CorrectPreState(*traceIdx),
+		ProofData:  s.builder.CorrectProofData(*traceIdx),
+		OracleData: s.builder.CorrectOracleData(*traceIdx),
```
ajsutton: Not sure this makes sense - seems like the `ClaimBuilder` API should just use `*big.Int`. The `big.Int` docs specifically say:
> shallow copies of Ints are not supported and may lead to errors.

so we always need to use a pointer.

op-challenger/game/fault/types/position.go
```diff
@@ -1,56 +1,89 @@
...
 // Position is a golang wrapper around the dispute game Position type.
 type Position struct {
 	depth        int
-	indexAtDepth int
+	indexAtDepth *big.Int
 }
 
-func NewPosition(depth, indexAtDepth int) Position {
-	return Position{depth, indexAtDepth}
+func NewPosition(depth int, indexAtDepth big.Int) Position {
```
ajsutton: We should be use a pointer here to avoid shallow copies of `big.Int` which aren't supported.

op-challenger/game/fault/test/claim_builder.go
```diff
@@ -39,30 +39,30 @@ func (c *ClaimBuilder) CorrectClaimAtPosition(pos types.Position) common.Hash {
 }
 
 // CorrectPreState returns the pre-state (not hashed) required to execute the valid step at the specified trace index
-func (c *ClaimBuilder) CorrectPreState(idx uint64) []byte {
-	pos := types.NewPosition(c.maxDepth, int(idx))
+func (c *ClaimBuilder) CorrectPreState(idx big.Int) []byte {
```
ajsutton: Should be using `*big.Int` through this file.

op-challenger/game/fault/trace/alphabet/provider.go
```diff
@@ -35,30 +35,31 @@ func NewTraceProvider(state string, depth uint64) *AlphabetTraceProvider {
...
-	return BuildAlphabetPreimage(traceIndex, ap.state[traceIndex]), []byte{}, nil, nil
+	return BuildAlphabetPreimage(traceIndex, ap.state[traceIndex.Uint64()]), []byte{}, nil, nil
 }
 
 // Get returns the claim value at the given index in the trace.
 func (ap *AlphabetTraceProvider) Get(ctx context.Context, i types.Position) (common.Hash, error) {
 	// Step data returns the pre-state, so add 1 to get the state for index i
-	postPosition := types.NewPosition(int(ap.depth), int(i.TraceIndex(int(ap.depth)))+1)
+	ti := i.TraceIndex(int(ap.depth))
+	postPosition := types.NewPosition(int(ap.depth), *big.NewInt(0).Add(ti, big.NewInt(1)))
```
ajsutton:
```suggestion
	postPosition := types.NewPosition(int(ap.depth), new(big.Int).Add(ti, big.NewInt(1)))
```

op-challenger/game/fault/trace/cannon/provider_test.go
```diff
@@ -22,15 +23,15 @@ import (
 //go:embed test_data
 var testData embed.FS
 
-func PositionFromTraceIndex(provider *CannonTraceProvider, idx int) types.Position {
+func PositionFromTraceIndex(provider *CannonTraceProvider, idx big.Int) types.Position {
```
ajsutton: Needs to be `*big.Int`

op-challenger/game/fault/trace/outputs/provider.go
```diff
@@ -52,7 +52,7 @@ func NewTraceProviderFromInputs(logger log.Logger, rollupClient OutputRollupClie
 }
 
 func (o *OutputTraceProvider) Get(ctx context.Context, pos types.Position) (common.Hash, error) {
-	outputBlock := pos.TraceIndex(int(o.gameDepth)) + o.prestateBlock + 1
+	outputBlock := pos.TraceIndex(int(o.gameDepth)).Uint64() + o.prestateBlock + 1
```
ajsutton: Need to check that the trace index can fit in a uint64 and return an error if it doesn't rather than just blindly casting.

op-challenger/game/fault/types/game.go
```diff
@@ -41,6 +45,17 @@ type claimEntry struct {
...
+func NewClaimEntry(claim Claim) claimEntry {
+	return claimEntry{
+		ClaimData:           claim.ClaimData,
+		ParentContractIndex: claim.ParentContractIndex,
+	}
+}
+
+func (c claimEntry) Hash() common.Hash {
+	return HashFromClaimData(c.Position.ToGIndex(), c.Value, c.ParentContractIndex)
+}
```
ajsutton: I don't think the `claimEntry` struct is giving us any value here. Since we need to move to a derived identifier like a hash for claims, we should just remove the struct entirely.  We'd wind up with just:

```
func computeClaimID(claim Claim) common.Hash {
	return crypto.Keccak256Hash(
		claim.Position.ToGIndex().Bytes(),
		claim.Value.Bytes(),
		big.NewInt(int64(claim.ParentContractIndex)).Bytes(),
	)
}
```

and all the uses of `claimEntry` in the file become `common.Hash` (including for `children` in `extendedClaim`).

And for good measure I'd use a type alias like `type claimID common.Hash` so it's clear when we're using a claim ID and not just any hash.

op-challenger/game/fault/types/position_test.go
```diff
@@ -28,93 +30,99 @@ func TestMSBIndex(t *testing.T) {
...
 // TestGINConversions does To & From the generalized index on the treeNodesMaxDepth4 data
 func TestGINConversions(t *testing.T) {
-	for _, test := range treeNodesMaxDepth4 {
-		from := NewPositionFromGIndex(test.GIndex)
-		pos := NewPosition(test.Depth, test.IndexAtDepth)
-		require.Equal(t, pos, from)
+	for _, test := range treeNodes {
+		from := NewLargePositionFromGIndex(test.GIndex)
+		pos := NewPosition(test.Depth, *test.IndexAtDepth)
+		require.True(t, pos.Equals(from), "From GIndex %v vs pos %v", from.IndexAtDepth(), pos.IndexAtDepth())
```
ajsutton: I'd avoid adding the `Equals` method to production code just to support this test and just use:

```

		require.EqualValuesf(t, pos.Depth(), from.Depth(), "From GIndex %v vs pos %v", from.Depth(), pos.Depth())
		require.Zerof(t, pos.IndexAtDepth().Cmp(from.IndexAtDepth()), "From GIndex %v vs pos %v", from.IndexAtDepth(), pos.IndexAtDepth())
```

op-challenger/game/fault/types/position.go
```diff
@@ -1,56 +1,89 @@
...
+// todo(refcell): remove this to use the NewLargePositionFromGIndex.
 func NewPositionFromGIndex(x uint64) Position {
 	depth := MSBIndex(x)
 	indexAtDepth := ^(1 << depth) & x
-	return NewPosition(depth, int(indexAtDepth))
+	return NewPosition(depth, *big.NewInt(int64(indexAtDepth)))
+}
+
+func (p Position) Equals(other Position) bool {
+	return p.depth == other.depth && p.indexAtDepth.Cmp(other.indexAtDepth) == 0
```
ajsutton: As noted on the test, I wouldn't add this production code just to support the test. Particularly because `Equals` isn't automatically used for comparison by go so it potentially causes confusion.

op-challenger/game/fault/types/position.go
```diff
@@ -1,56 +1,89 @@
...
 }
 
 func (p Position) MoveRight() Position {
 	return Position{
 		depth:        p.depth,
-		indexAtDepth: int(p.indexAtDepth + 1),
+		indexAtDepth: big.NewInt(0).Add(p.indexAtDepth, big.NewInt(1)),
```
ajsutton:
```suggestion
		indexAtDepth: new(big.Int).Add(p.indexAtDepth, big.NewInt(1)),
```

op-challenger/game/fault/types/position.go
```diff
@@ -1,56 +1,89 @@
...
 
 // TraceIndex calculates the what the index of the claim value would be inside the trace.
 // It is equivalent to going right until the final depth has been reached.
-func (p Position) TraceIndex(maxDepth int) uint64 {
+func (p Position) TraceIndex(maxDepth int) *big.Int {
 	// When we go right, we do a shift left and set the bottom bit to be 1.
 	// To do this in a single step, do all the shifts at once & or in all 1s for the bottom bits.
 	rd := maxDepth - p.depth
-	return uint64(p.indexAtDepth<<rd | ((1 << rd) - 1))
+	rhs := ((1 << rd) - 1)
```
ajsutton: This still overflows for large depths:

```
func TestTraceIndexOfRootWithLargeDepth(t *testing.T) {
	traceIdx := new(big.Int).Lsh(big.NewInt(1), 100)
	pos := NewLargePositionFromGIndex(big.NewInt(1))
	actual := pos.TraceIndex(100)
	require.Equal(t, traceIdx, actual)
}
```

Not 100% my expected trace index is right, but I know the actual answer it's getting of `-1` is wrong.

op-challenger/game/fault/types/position.go
```diff
@@ -1,56 +1,89 @@
...
+	ti := new(big.Int).Or(p.lshIndex(rd), big.NewInt(int64(rhs)))
+	return ti
+}
+
+// UnsafeTraceIndex returns a uint64 representation of the trace index.
+// todo(refcell): This should be removed in a follow-on pr and any invocations
```
ajsutton, line 91: Can we make the todo reference the issue rather than you - then we know we need to fix the TODO before the issue is done.

op-challenger/game/fault/types/position.go
```diff
@@ -1,56 +1,89 @@
...
+func (p Position) UnsafeTraceIndex(maxDepth int) uint64 {
+	return p.TraceIndex(maxDepth).Uint64()
 }
 
 // move returns a new position at the left or right child.
 func (p Position) move(right bool) Position {
 	return Position{
 		depth:        p.depth + 1,
-		indexAtDepth: (p.indexAtDepth << 1) | boolToInt(right),
+		indexAtDepth: big.NewInt(0).Or(p.lshIndex(1), big.NewInt(int64(boolToInt(right)))),
```
ajsutton:
```suggestion
		indexAtDepth: new(big.Int).Or(p.lshIndex(1), big.NewInt(int64(boolToInt(right)))),
```

op-challenger/game/fault/types/position.go
```diff
@@ -81,11 +122,23 @@ func (p Position) Defend() Position {
 }
 
 func (p Position) Print(maxDepth int) {
-	fmt.Printf("GIN: %4b\tTrace Position is %4b\tTrace Depth is: %d\tTrace Index is: %d\n", p.ToGIndex(), p.indexAtDepth, p.depth, p.TraceIndex(maxDepth))
+	fmt.Printf("GIN: %4b\tTrace Position is %4b\tTrace Depth is: %d\tTrace Index is: %d\n", p.ToGIndex(), p.indexAtDepth.Uint64(), p.depth, p.TraceIndex(maxDepth))
```
ajsutton:
```suggestion
	fmt.Printf("GIN: %4b\tTrace Position is %4b\tTrace Depth is: %d\tTrace Index is: %d\n", p.ToGIndex(), p.indexAtDepth, p.depth, p.TraceIndex(maxDepth))
```

We should never be calling `.UInt64()` on the big ints or we're going to introduce bugs.

op-challenger/game/fault/types/position.go
```diff
@@ -81,11 +122,23 @@ func (p Position) Defend() Position {
...
+}
+
+func (p Position) ToGIndex() *big.Int {
+	return big.NewInt(0).Or(big.NewInt(0).Lsh(big.NewInt(1), uint(p.depth)), p.IndexAtDepth())
```
ajsutton:
```suggestion
	return new(big.Int).Or(new(big.Int).Lsh(big.NewInt(1), uint(p.depth)), p.IndexAtDepth())
```

The `new(big.Int)` syntax is a little more typically but also avoids having `0` in the equation which looks like it's important when its actually completely ignored and just being used as the final target for the result.

op-challenger/game/fault/types/position.go
```diff
@@ -81,11 +122,23 @@ func (p Position) Defend() Position {
...
 }
 
-func (p Position) ToGIndex() uint64 {
-	return uint64(1<<p.depth | p.indexAtDepth)
+// BigMSB returns the index of the most significant bit
+func BigMSB(x *big.Int) int {
```
ajsutton: Does this need to be exported?

ajsutton (commented):
Can you remove `fixes` from the PR description please? This leaves todos and other items to finish off before the story is done (which is fine, we just don't want to close the PR).

And I wouldn't hold back on adding tests to reduce diff size. We want to ensure each change we make is top quality which means adding all the required tests. If we're going to reduce size we do it by reducing the scope of the change, not by cutting testing.  This PR isn't actually that big because most of the changes are just types changing which is really easy to review - the fact that it's a lot of lines doesn't really matter.

op-challenger/game/fault/types/game.go
ajsutton: hmm, yeah it appears that we can't avoid moving to something like a Hash - I don't like it too much but it is what it is.  My suggestion then would be to simplify the heck out of the game implementation like https://github.com/ethereum-optimism/optimism/pull/7517 - I _think_ with that, we'll just need to `computeClaimID` it adds to handle `ToGIndex` returning a big.Int instead of having converting it to big.Int. This is inline with what I was suggesting in terms of calculating claim ID but I realised it was a good opportunity to simplify the game implementation a lot and would work as a separate PR.

op-challenger/game/fault/types/position.go
refcell: Done, created the issue https://github.com/ethereum-optimism/client-pod/issues/80

op-challenger/game/fault/types/game.go
refcell: Thank you so much for doing this - really meaningfully made the implementation cleaner

op-challenger/game/fault/types/game.go
refcell: Thank you for doing this in your pr!

op-challenger/game/fault/types/position.go
refcell: Fixed - the `rhs := ((1 << rd) - 1)` was overflowing so changed to use big integers.

op-challenger/game/fault/solver/solver_test.go
```diff
@@ -32,9 +34,9 @@ func TestAttemptStep(t *testing.T) {
 		{
 			name:               "AttackFirstTraceIndex",
 			expectAttack:       true,
-			expectPreState:     claimBuilder.CorrectPreState(0),
-			expectProofData:    claimBuilder.CorrectProofData(0),
-			expectedOracleData: claimBuilder.CorrectOracleData(0),
+			expectPreState:     claimBuilder.CorrectPreState(new(big.Int)),
+			expectProofData:    claimBuilder.CorrectProofData(new(big.Int)),
+			expectedOracleData: claimBuilder.CorrectOracleData(new(big.Int)),
```
ajsutton: These would more more typical as `big.NewInt(0)` or `common.Big0`.  😄  The `new(big.Int)` is typically just when it's being used as the receiver of some computation (ie `new(big.Int).Add(big.NewInt(0), big.NewInt(1))`.

op-challenger/game/fault/test/claim_builder.go
```diff
@@ -78,15 +78,15 @@ func (c *ClaimBuilder) CreateRootClaim(correct bool) types.Claim {
 	claim := types.Claim{
 		ClaimData: types.ClaimData{
 			Value:    value,
-			Position: types.NewPosition(0, 0),
+			Position: types.NewPosition(0, new(big.Int)),
```
ajsutton:
```suggestion
			Position: types.NewPosition(0, common.Big0),
```

op-challenger/game/fault/test/claim_builder.go
```diff
@@ -78,15 +78,15 @@ func (c *ClaimBuilder) CreateRootClaim(correct bool) types.Claim {
...
 		},
 	}
 	return claim
 }
 
-func (c *ClaimBuilder) CreateLeafClaim(traceIndex uint64, correct bool) types.Claim {
-	parentPos := types.NewPosition(c.maxDepth-1, 0)
-	pos := types.NewPosition(c.maxDepth, int(traceIndex))
+func (c *ClaimBuilder) CreateLeafClaim(traceIndex *big.Int, correct bool) types.Claim {
+	parentPos := types.NewPosition(c.maxDepth-1, new(big.Int))
```
ajsutton:
```suggestion
	parentPos := types.NewPosition(c.maxDepth-1, common.Big0)
```

op-challenger/game/fault/test/game_builder.go
```diff
@@ -120,7 +122,7 @@ func (s *GameBuilderSeq) ExpectStepAttack() *GameBuilderSeq {
 }
 
 func (s *GameBuilderSeq) ExpectStepDefend() *GameBuilderSeq {
-	traceIdx := s.lastClaim.TraceIndex(s.builder.maxDepth) + 1
+	traceIdx := big.NewInt(0).Add(s.lastClaim.TraceIndex(s.builder.maxDepth), big.NewInt(1))
```
ajsutton:
```suggestion
	traceIdx := new(big.Int).Add(s.lastClaim.TraceIndex(s.builder.maxDepth), big.NewInt(1))
```

op-challenger/game/fault/trace/alphabet/provider.go
```diff
@@ -35,30 +35,31 @@ func NewTraceProvider(state string, depth uint64) *AlphabetTraceProvider {
 
 func (ap *AlphabetTraceProvider) GetStepData(ctx context.Context, i types.Position) ([]byte, []byte, *types.PreimageOracleData, error) {
 	traceIndex := i.TraceIndex(int(ap.depth))
-	if traceIndex == 0 {
+	if traceIndex.Cmp(new(big.Int)) == 0 {
```
ajsutton:
```suggestion
	if traceIndex.Cmp(common.Big0) == 0 {
```

op-challenger/game/fault/trace/cannon/provider_test.go
```diff
@@ -22,15 +23,15 @@ import (
...
+func PositionFromTraceIndex(provider *CannonTraceProvider, idx *big.Int) types.Position {
 	return types.NewPosition(int(provider.gameDepth), idx)
 }
 
 func TestGet(t *testing.T) {
 	dataDir, prestate := setupTestData(t)
 	t.Run("ExistingProof", func(t *testing.T) {
 		provider, generator := setupWithTestData(t, dataDir, prestate)
-		value, err := provider.Get(context.Background(), PositionFromTraceIndex(provider, 0))
+		value, err := provider.Get(context.Background(), PositionFromTraceIndex(provider, new(big.Int)))
```
ajsutton:
```suggestion
		value, err := provider.Get(context.Background(), PositionFromTraceIndex(provider, common.Big0))
```

op-challenger/game/fault/trace/outputs/provider.go
```diff
@@ -52,7 +54,11 @@ func NewTraceProviderFromInputs(logger log.Logger, rollupClient OutputRollupClie
 }
 
 func (o *OutputTraceProvider) Get(ctx context.Context, pos types.Position) (common.Hash, error) {
-	outputBlock := pos.TraceIndex(int(o.gameDepth)) + o.prestateBlock + 1
+	traceIndex := pos.TraceIndex(int(o.gameDepth))
+	if traceIndex.Cmp(new(big.Int).SetUint64(math.MaxUint64)) > 0 {
```
ajsutton:
```suggestion
	if traceIndex.Cmp(common.Big0.SetUint64(math.MaxUint64)) > 0 {
```

op-challenger/game/fault/trace/outputs/provider_test.go
```diff
@@ -26,13 +27,21 @@ var (
 func TestGet(t *testing.T) {
 	t.Run("PrePrestateErrors", func(t *testing.T) {
 		provider, _ := setupWithTestData(t, 0, poststateBlock)
-		_, err := provider.Get(context.Background(), types.NewPosition(1, 0))
+		_, err := provider.Get(context.Background(), types.NewPosition(1, new(big.Int)))
```
ajsutton:
```suggestion
		_, err := provider.Get(context.Background(), types.NewPosition(1, common.Big0))
```

op-challenger/game/fault/trace/outputs/provider_test.go
```diff
@@ -82,7 +91,7 @@ func TestAbsolutePreStateCommitment(t *testing.T) {
 
 func TestGetStepData(t *testing.T) {
 	provider, _ := setupWithTestData(t, prestateBlock, poststateBlock)
-	_, _, _, err := provider.GetStepData(context.Background(), types.NewPosition(1, 0))
+	_, _, _, err := provider.GetStepData(context.Background(), types.NewPosition(1, new(big.Int)))
```
ajsutton:
```suggestion
	_, _, _, err := provider.GetStepData(context.Background(), types.NewPosition(1, common.Big0))
```

op-challenger/game/fault/types/game_test.go
```diff
@@ -15,14 +16,14 @@ func createTestClaims() (Claim, Claim, Claim, Claim) {
 	root := Claim{
 		ClaimData: ClaimData{
 			Value:    common.HexToHash("0x000000000000000000000000000000000000000000000000000000000000077a"),
-			Position: NewPosition(0, 0),
+			Position: NewPosition(0, new(big.Int)),
```
ajsutton:
```suggestion
			Position: NewPosition(0, common.Big0),
```

op-challenger/game/fault/types/game_test.go
```diff
@@ -15,14 +16,14 @@ func createTestClaims() (Claim, Claim, Claim, Claim) {
...
 		},
 		// Root claim has no parent
 	}
 	top := Claim{
 		ClaimData: ClaimData{
 			Value:    common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000364"),
-			Position: NewPosition(1, 0),
+			Position: NewPosition(1, new(big.Int)),
```
ajsutton:
```suggestion
			Position: NewPosition(1, common.Big0),
```

op-challenger/game/fault/types/position.go
```diff
@@ -1,56 +1,85 @@
...
 }
 
 func (p Position) Depth() int {
 	return p.depth
 }
 
-func (p Position) IndexAtDepth() int {
+func (p Position) IndexAtDepth() *big.Int {
+	if p.indexAtDepth == nil {
+		return new(big.Int)
```
ajsutton:
```suggestion
		return common.Big0
```

op-challenger/game/fault/types/position.go
```diff
@@ -1,56 +1,85 @@
...
+	}
 	return p.indexAtDepth
 }
 
 func (p Position) IsRootPosition() bool {
-	return p.depth == 0 && p.indexAtDepth == 0
+	return p.depth == 0 && new(big.Int).Cmp(p.indexAtDepth) == 0
```
ajsutton:
```suggestion
	return p.depth == 0 && common.Big0.Cmp(p.indexAtDepth) == 0
```

op-challenger/game/fault/types/position.go
```diff
@@ -84,8 +121,20 @@ func (p Position) Print(maxDepth int) {
...
+	return new(big.Int).Or(new(big.Int).Lsh(big.NewInt(1), uint(p.depth)), p.IndexAtDepth())
+}
+
+// bigMSB returns the index of the most significant bit
+func bigMSB(x *big.Int) int {
+	if x.Cmp(new(big.Int)) == 0 {
+		return 0
+	}
+	out := 0
+	for ; x.Cmp(new(big.Int)) != 0; out++ {
```
ajsutton:
```suggestion
	for ; x.Cmp(common.Big0) != 0; out++ {
```

op-challenger/game/fault/types/position.go
```diff
@@ -84,8 +121,20 @@ func (p Position) Print(maxDepth int) {
...
 
-func (p Position) ToGIndex() uint64 {
-	return uint64(1<<p.depth | p.indexAtDepth)
+func (p Position) ToGIndex() *big.Int {
+	return new(big.Int).Or(new(big.Int).Lsh(big.NewInt(1), uint(p.depth)), p.IndexAtDepth())
+}
+
+// bigMSB returns the index of the most significant bit
+func bigMSB(x *big.Int) int {
+	if x.Cmp(new(big.Int)) == 0 {
```
ajsutton:
```suggestion
	if x.Cmp(common.Big0) == 0 {
```

ajsutton (approved):
LGTM. The difference between `new(big.Int)` and `big.NewInt(0)` or `common.Big0` doesn't matter functionally but I find it much easier to understand the code when `new(big.Int)` is used consistently as the destination of a computation and not for when the 0 value is actually important.

PR #7337, opened 2023-09-21
op-node: un-hide beta flags, update superchain-registry for devnet protocol-versions monitoring

protolambda:
**Description**

- Un-hide the beta-flags, so the beta features are visible to more users
- Update the superchain-registry, to incude the op-labs devnet
- Fix runtime & L2 init order, so the L2 engine API client is created before it's used by the runtime config when signaling protocol version info
- Load the protocol-versions address from the superchain-registry into the rollup-config.

Flags to join the devnet:
```
                  --network=op-labs-devnet-0-goerli-dev-0 \
                  --beta.extra-networks \
                  --beta.rollup.halt=minor \
                  --beta.rollup.load-protocol-versions
```

Or, if using a manual `rollup.json`, add the `protocol_versions_address` manually to the config file.

**Tests**

Tested live on devnet:
```

# setting recommended to v3.1.0-1
cast send --rpc-url=$ETH_RPC_URL --private-key=$PRIVATE_KEY 0x79ADD5713B383DAa0a138d3C4780C7A1804a8090 'setRecommended(uint256)' 0x0000000000000000000000000000000000000003000000010000000000000001
https://goerli.etherscan.io/tx/0x3ca0a70d33b5618b5a3f543df06f3c449af8f40f3b017e0ed703c78146e1e045

# setting required to v3.2.0, to test out-of-date signal on live devnet
cast send --rpc-url=$ETH_RPC_URL --private-key=$PRIVATE_KEY 0x79ADD5713B383DAa0a138d3C4780C7A1804a8090 'setRequired(uint256)' 0x0000000000000000000000000000000000000003000000020000000000000000
https://goerli.etherscan.io/tx/0xd0935c2a112ba979fb12d24e4a87bdecddc4584ffb10da7d5a862048cd7a6c91

# setting required back to v3.1.0-1
cast send --rpc-url=$ETH_RPC_URL --private-key=$PRIVATE_KEY 0x79ADD5713B383DAa0a138d3C4780C7A1804a8090 'setRequired(uint256)' 0x0000000000000000000000000000000000000003000000010000000000000001
```


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.2.0
 	github.com/ethereum-optimism/go-ethereum-hdwallet v0.1.3
-	github.com/ethereum-optimism/superchain-registry/superchain v0.0.0-20230920213331-413695cf7906
+	github.com/ethereum-optimism/superchain-registry/superchain v0.0.0-20230921150749-b0edabdeac8c
```
trianglesphere: Would be nice to start versionning the superchain-regsitry. It's a pretty simple codebase, so should be easy to do so.

trianglesphere (dismissed):

protolambda:
Found a config issue where the devnet wasn't configured with the same defaults as the other newer networks, causing a bad rollup config. Will fix that and then this PR can be reviewed again. This only affects the devnet config through the beta-flag though.

tynes (approved):

go.mod
tynes: I agree but don't think its a blocker for this merge

tynes (approved):

op-node/rollup/superchain.go
```diff
@@ -99,5 +102,9 @@ func LoadOPStackRollupConfig(chainID uint64) (*Config, error) {
 	if superChain.Config.ProtocolVersionsAddr != nil { // Set optional protocol versions address
 		cfg.ProtocolVersionsAddress = common.Address(*superChain.Config.ProtocolVersionsAddr)
 	}
+	if chainID == labsDevnet {
```
trianglesphere, line 105: Why isn't this part of the baked in file? It's part of the rollup.json.

trianglesphere (changes requested):

op-node/rollup/superchain.go
protolambda: rollup.json is not baked into the registry. We didn't add one-off configuration to the registry, to keep chains more uniform.

trianglesphere: Chains should be kept uniform, however it is a valid per-chain config & I think we should be more explicit in storing the full chain config with each chain rather than having to override it here.

op-node/rollup/superchain.go
```diff
@@ -59,13 +60,15 @@ func LoadOPStackRollupConfig(chainID uint64) (*Config, error) {
...
+	// three goerli testnets test-ran Bedrock and later upgraded to Regolith.
 	// All other OP-Stack chains have Regolith enabled from the start.
 	switch chainID {
 	case baseGoerli:
 		regolithTime = 1683219600
 	case opGoerli:
 		regolithTime = 1679079600
+	case labsDevnet:
+		regolithTime = 1677984480
 	}
```
sebastianst, line 72: Similar to @trianglesphere's comment, is there a way to store the regolith timestamp in a per-chain config rather than putting it into the config loader function? What's the idea behind putting some chain-specific configuration into a seemingly generic loader function?

protolambda: The idea was that goerli and devnet will stop to exist, and so I thought it would be better to not bloat the registry with configuration options that nobody should change. Same thing for the other parameters like channel-timeout, which should be the same with every network.

I can add it as option to the registry, to reduce the complexity here in the op-node, but it does the opposite in the registry, adding more complexity there. It's a trade-off.

sebastianst: I see the argument that per-chain hf activation times should be out of scope for the superchain-registry. It's only a temporary situation and it can be deleted once Goerli shuts down. So I'm happy to keep as is.

op-node/rollup/superchain.go
sebastianst: See my other comment, there's an argument here that per-chain config of these values should be out of scope of the superchain-registry, since superchains shouldn't have differences in these configs. It's a temporary artifact from pre-superchain world. We can delete this when Goerli is shut down.

If we actually needed to officially support superchain chains in the future with some differences in the rollup config, then I would also suggest to add this to the superchain-registry. But in this case, I'd be fine with as it is. Still think that a code comment about this would be nice here.

trianglesphere (approved):

PR #7317, opened 2023-09-20
contracts-bedrock/periphery: create deployment scripts for configuring drippie with the faucet proxy contract

tremarkley:
Closes https://github.com/ethereum-optimism/ecopod/issues/534

This adds the `installFaucetDrippieConfigs` method to `DeployPeriphery`, which when called adds the drips to the Faucet `Drippie` contract that handle dripping funds to the Faucet `Proxy` and funding the Gelato account, which automatically checks if drips are execuatable.

PR #7203, opened 2023-09-12
Deployment script for contracts-periphery and faucet

tremarkley:
Part of https://github.com/ethereum-optimism/ecopod/issues/533
Part of https://github.com/ethereum-optimism/ecopod/issues/534 

This PR implements the following:
- A deployment system for deploying the contracts inside of the contracts-periphery directory. 
- A deployment script for deploying an upgradeable faucet contract

For UX reasons we would like the `FaucetProxy` contract to be deployed at the same address across all OP Chains. In order to do that the `ProxyAdmin`, `FaucetProxy` must provide a `salt` on contract creation.

PR #7342, opened 2023-09-21
contracts-bedrock/periphery: When deploying contracts with CREATE2, check if already deployed at address

tremarkley:
Part of https://github.com/ethereum-optimism/ecopod/issues/533
Part of https://github.com/ethereum-optimism/ecopod/issues/534

With this PR, `DeployPeriphery` will check if the contract has already been deployed prior to calling `CREATE2`.

### DIFF
--- a/packages/contracts-bedrock/scripts/DeployPeriphery.s.sol
+++ b/packages/contracts-bedrock/scripts/DeployPeriphery.s.sol
@@ -63,41 +63,68 @@ contract DeployPeriphery is Deployer {
 
     /// @notice Deploy the ProxyAdmin
     function deployProxyAdmin() public broadcast returns (address addr_) {
-        ProxyAdmin admin = new ProxyAdmin{ salt: keccak256(bytes("ProxyAdmin")) }({
-            _owner: msg.sender
-        });
-        require(admin.owner() == msg.sender);
+        bytes32 salt = keccak256(bytes("ProxyAdmin"));
+        bytes32 initCodeHash = keccak256(abi.encodePacked(type(ProxyAdmin).creationCode, abi.encode(msg.sender)));
+        address preComputedAddress = computeCreate2Address(salt, initCodeHash);
+        if (preComputedAddress.code.length > 0) {
+            console.log("ProxyAdmin already deployed at %s", preComputedAddress);
+            save("ProxyAdmin", preComputedAddress);
+            addr_ = preComputedAddress;
+        } else {
+            ProxyAdmin admin = new ProxyAdmin{ salt: salt }({
+              _owner: msg.sender
+            });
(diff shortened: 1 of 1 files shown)


tremarkley:
* develop
  * **PR #7203** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7203" target="_blank">Graphite</a> 
    * **PR #7342** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7342" target="_blank">Graphite</a>  👈
      * **PR #7317** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7317" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7342?utm_source=stack-comment).

packages/contracts-bedrock/scripts/DeployPeriphery.s.sol
```diff
@@ -63,14 +63,21 @@ contract DeployPeriphery is Deployer {
 
     /// @notice Deploy the ProxyAdmin
     function deployProxyAdmin() public broadcast returns (address addr_) {
-        ProxyAdmin admin = new ProxyAdmin{ salt: keccak256(bytes("ProxyAdmin")) }({
+        bytes32 initCodeHash = keccak256(abi.encodePacked(type(ProxyAdmin).creationCode, abi.encode(msg.sender)));
+        address preComputedAddress = computeCreate2Address(keccak256(bytes("ProxyAdmin")), initCodeHash);
+        if (preComputedAddress.code.length > 0) {
+          console.log("ProxyAdmin already deployed at %s", preComputedAddress);
+          save("ProxyAdmin", address(preComputedAddress));
```
tynes: you should be able to skip `save` here assuming the artifact is on disk

tynes:
We are considering moving away from the hardhat deploy style artifacts because we don't need them anymore now that we have the superchain-registry as the source of truth for all of the addresses, that would remove the whole `save` and `sync()` flow

packages/contracts-bedrock/scripts/DeployPeriphery.s.sol
tremarkley: Until the superchain-registry is ready, should we still call `save` here, so that the artifact gets saved to disk, and then in other places where the artifact is referenced it can still be located? (for example, [this function](https://github.com/ethereum-optimism/optimism/blob/adf55b3a60279e4750ab0682cb92d19921bbd92a/packages/contracts-bedrock/scripts/DeployPeriphery.s.sol#L105-L107) requires that the artifact be on disk)

tynes (approved):

PR #7348, opened 2023-09-22
bugfix(op-node): syncClient incorrectly removes peer issue

### LINKED ISSUE #7285: op-node: SyncClient may incorrectly remove peers.
**Describe the bug**
I noticed the following code
```
n.syncCl = NewSyncClient(log, rollupCfg, n.host.NewStream, gossipIn.OnUnsafeL2Payload, metrics, n.appScorer)
			n.host.Network().Notify(&network.NotifyBundle{
				ConnectedF: func(nw network.Network, conn network.Conn) {
					n.syncCl.AddPeer(conn.RemotePeer())
				},
				DisconnectedF: func(nw network.Network, conn network.Conn) {
					n.syncCl.RemovePeer(conn.RemotePeer())
				},
			})
```
When the staticPeer is configured with a DNS address (such as /dns/xxx.com/tcp/xxx/p2p/xxx), a peer can obtain multiple IP addresses through domain name resolution. However, in Libp2p, dialing concurrently can result in one peer corresponding to multiple connections, which is exacerbated by multiple IP addresses. If one of the connections is disconnected due to an unstable network or other reasons, the Notifiee.Disconnected event is triggered. At this time, the peer is removed from the syncClient, and the processing flow of the peerLoop exits. The log displays the message `stopped syncing loop of peer`. However, the peer still has other connections that can be used for p2p synchronization. Therefore, we should not remove this peer unless we determine that all of its connections are disconnected and there are no available connections, just like the check performed by the `monitorStaticPeers` function.

**To Reproduce**
1. Configuration --p2p.static=/dns/xxx.com/tcp/xxx/p2p/xxx
2. Run op-node
3. Sometimes it can be observed that the same peers establish multiple connections.There are multiple `connected to peer` records for the same peer in the log.
4. After a while, one of the connections was disconnected, causing a `stopped syncing loop of peer`

**Expected behavior**
we should not remove this peer unless we determine that all of its connections are disconnected and there are no available connections
Suggested changes:
```
DisconnectedF: func(nw network.Network, conn network.Conn) {
					// only when no connection is available, we can remove the peer
					if nw.Connectedness(conn.RemotePeer()) == network.NotConnected {
						n.syncCl.RemovePeer(conn.RemotePeer())
					}
				},
```

**Screenshots**
截屏2023-09-18 17 29 39

**System Specs:**
 - OS: linux
 - Package Version (or commit hash): release/bedrock-mainnet ~~but I believe the latest code also has this issue~~

welkin22:
**Description**
In the release/bedrock-mainnet branch, it is still using version 0.25.1 of go-libp2p. I found a problem in this version: a static peer sometimes has multiple connections, and when one of the connections is closed, the entire peer will be removed by syncClient.
we should not remove this peer unless we determine that all of its connections are disconnected and there are no available connections.
Although this problem no longer exists in the latest version of go-libp2p 0.31.0, I still think that syncClient should listen to the EvtPeerConnectednessChanged event to decide whether to remove the peer. This is better.

**Tests**
TestUseEvtPeerConnectednessChangedEvent
**Metadata**

- Fixes #7285


trianglesphere (commented):
If this issue is no longer present because it is fixed in libp2p, I think we rather keep it as is because callbacks are nicer to manage than a subscription & new go-routine.

welkin22:
> If this issue is no longer present because it is fixed in libp2p, I think we rather keep it as is because callbacks are nicer to manage than a subscription & new go-routine.

@trianglesphere Ok. How about the following modification? Add a conditional judgment protection.This change ensures that we do not delete peers by mistake, just like the logic in the AddPeer method to prevent repeated additions.
```
DisconnectedF: func(nw network.Network, conn network.Conn) {
					// only when no connection is available, we can remove the peer
					if nw.Connectedness(conn.RemotePeer()) == network.NotConnected {
						n.syncCl.RemovePeer(conn.RemotePeer())
					}
				},
```

trianglesphere:
@welkin22 that looks like a good check to add.

welkin22:
> @welkin22 that looks like a good check to add.

@trianglesphere I have modified the code, please review it again

op-node/p2p/sync_test.go
```diff
@@ -288,3 +288,52 @@ func TestMultiPeerSync(t *testing.T) {
...
+				syncCl.RemovePeer(conn.RemotePeer())
+			}
+		},
+	})
+	syncCl.Start()
+
+	err = hostA.Connect(context.Background(), peer.AddrInfo{ID: hostB.ID(), Addrs: hostB.Addrs()})
+	require.NoError(t, err, "failed to connect to peer B from peer A")
+	require.Equal(t, hostA.Network().Connectedness(hostB.ID()), network.Connected)
+	//wait for async add process done
```
trianglesphere, line 331: Is there any way to poll for these changes? Polling / waiting on a channel is much more reliable in CI rather than waiting for wall clock time

trianglesphere (commented):
One concern about flakey tests, actual changes look good.
Thanks for iterating on this.

op-node/p2p/sync_test.go
welkin22: I tried using channels instead of sleeping, please review again

trianglesphere (approved):
Thanks for the changes & going through all of the iterations

welkin22:
@trianglesphere Hi, can this PR be merged?

trianglesphere:
@welkin22 yes, sorry. It needed to be added to the merge queue. PRs used to automerge when ready

PR #7349, opened 2023-09-22
Experimental EIP-4844 support

protolambda:
**Description**

This PR depends on the L1 Dencun support changes of https://github.com/ethereum-optimism/optimism/pull/8131

This implements *experimental* EIP-4844 DA support.

Changes:
- introduce a Beacon-API endpoint option to the op-node, to fetch blobs from
- introduce deploy-config updates for:
  - L1 HF activation of Cancun (`l1CancunTimeOffset` relative to genesis)
  - L2 HF activation for Blobs processing (`l2BlobsUpgradeTimeOffset` relative to genesis)
- refactor the derivation slightly to have an optional blobs data source, switched over when the HF is active. The old calldata inbox still works, as fallback if the network experiences any problems post-4844.
- op-e2e utils extension: (cherry-picked into L1 dencun support PRs)
  - changes to build Cancun L1 blocks with the fake-proof-of-stake geth block builder
  - introduce a fake beacon node, which serves the minimal subset of the beacon-API to serve blobs, and holds on to the blobs after the geth node puts them in a block.
- op-node support for L1 Cancun block-header validation
- op-batcher changes to submit the data as blobs
- op-service tx manager changes to support blobs as tx-candidate option
- some new types to handle the blob and commitments in the op-node. (cherry-picked into L1 dencun support PRs)

TODO:
- [ ] more e2e testing
- [x] blob-gas fee estimation for tx-manager
- [x] utilize the blob-commitments/proofs of the sidecar, to verify the blob contents against the versioned hashes faster and more efficiently.
- [x] review upstream beacon-API changes. The op-e2e test relies on the beacon-API as it looked in an earlier 4844 prototype, this likely has changed.

**Tests**

Adds an op-e2e test that confirms an L2 transaction is included in a safe block, on a chain with 4844 blobs as DA!


tynes:
I am concerned the fee pricing is being overlooked here. We need to have the `blobbasefee` pushed to L2 through the system tx into the `L1Block` contract as well as updating the `L1CostFunc` in `op-geth` to new logic. There isn't consensus on what this logic should be yet but my proposal is to compute the fee using both calldata and blobspace and then take the min of the 2. This would allow `op-node` to still parse data from both calldata and blobspace after the upgrade and assumes the batcher is smart enough to use the cheaper resource at the time. This makes the spec more complex but would not break support for base chains that do not have blobspace. Also the implementation of the `GasPriceOracle` smart contract on L2 would need to be updated with the new `L1CostFunc` logic for gas estimation

protolambda:
Closing in favor of https://github.com/ethereum-optimism/optimism/pull/8434

PR #7352, opened 2023-09-22
L1 contract deploy: add support for customization of the IMPL_SALT

mdehoog:
**Description**
Currently any new L2 chain using this repo's Deploy script has to edit the hardcoded `IMPL_SALT` variable to be able to deploy to an L1 chain that has already been deployed to, otherwise the script will fail with:
```
    ├─ [0] → new OptimismPortal@0x8Cfa294bD0c6F63cD65d492bdB754eAcf684D871
    │   └─ ← 21596 bytes of code
    └─ ← "EvmError: Revert"
```
because the contract is already deployed.

This PR lets one customize the salt using the `IMPL_SALT` environment variable.

**Tests**
Tested locally using a forge simulation.

**Additional context**
We are trying to bootstrap our sepolia chain and running into this.

### DIFF
--- a/packages/contracts-bedrock/scripts/Deploy.s.sol
+++ b/packages/contracts-bedrock/scripts/Deploy.s.sol
@@ -47,11 +47,6 @@ import "src/libraries/DisputeTypes.sol";
 contract Deploy is Deployer {
     DeployConfig cfg;
 
-    /// @notice The create2 salt used for deployment of the contract implementations.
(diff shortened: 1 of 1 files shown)


tynes:
This is a good short term solution until we have the ability to just deploy the proxies and initialize them. There is no need to redeploy the implementations each time, we just need to deploy new proxies and point them to the current impls. Another solution is to compute the address and skip if the code exists

tynes (approved):

PR #7353, opened 2023-09-22
contracts-bedrock: legacy ISemver

tynes:
**Description**

Update the legacy contracts to use `ISemver` over `Semver`. Reduces the codesize and no longer needs execution to populate the semver. One step closer to killing the `Semver` contract.


refcell (approved):
Very nice!

PR #7358, opened 2023-09-24
contracts-bedrock: move to `ISemver`

tynes:
**Description**

Stop using `Semver` in favor of `ISemver` for
the EAS contracts. This helps to reduce the code
size and makes it so that the version is set directly in the bytecode instead of being set at runtime as an immutable.


packages/contracts-bedrock/src/EAS/SchemaRegistry.sol
```diff
@@ -19,9 +19,9 @@ contract SchemaRegistry is ISchemaRegistry, Semver {
     // Upgrade forward-compatibility storage gap
     uint256[MAX_GAP - 1] private __gap;
 
-    /// @dev Creates a new SchemaRegistry instance.
-    /// @custom:semver 1.2.0
-    constructor() Semver(1, 2, 0) { }
+    // @notice Semantic version.
```
clabby:
```suggestion
    /// @notice Semantic version.
```

packages/contracts-bedrock/src/EAS/EAS.sol
```diff
@@ -79,9 +79,12 @@ contract EAS is IEAS, Semver, EIP1271Verifier {
     // Upgrade forward-compatibility storage gap
     uint256[MAX_GAP - 3] private __gap;
 
+    // @notice Semantic version.
```
clabby:
```suggestion
    /// @notice Semantic version.
```

clabby (approved):
LGTM - just a few nits on the doc comments.

PR #7360, opened 2023-09-24
encapsulate data <-> blob conversion code, add unit tests

roberto-bayardo:
**Description**

Move the blob encoding/decoding code into op-service/eth/blob.go and add unit tests.

Update some obsolete comments.

**Tests**

Added unit tests for blob encoding/decoding.


protolambda (approved):
very nice

PR #7361, opened 2023-09-25
cannon: remove old test-util preimage-cheat function

protolambda:
**Description**

Remove the old `cheat` function from the pre-image oracle, to prepare it for secure usage onchain.

This util, in older versions of cannon, was used for testing purposes: to load the local game-specific preimages, without needing a localized context.

The `loadLocalData` replaced it, by hashing the key with the caller, which then attaches it to the dispute-game instance, through `addLocalData`, while preventing arbitrary data inputs.


packages/contracts-bedrock/.gas-snapshot
```diff
@@ -17,7 +17,7 @@ AttestationStationTest:test_attest_bulk_succeeds() (gas: 703749)
 AttestationStationTest:test_attest_individual_succeeds() (gas: 632087)
 AttestationStationTest:test_attest_single_succeeds() (gas: 651325)
 BlockOracle_Test:test_checkpointAndLoad_succeeds() (gas: 58341)
-BlockOracle_Test:test_load_noBlockHash_reverts() (gas: 12805)
+BlockOracle_Test:test_load_noBlockHash_reverts() (gas: 13659)
```
protolambda: tried updating the gas snapshot, but the gas changes make no sense to me. Is there a specific check I can do to fix this?

tynes: I am guessing you are using the wrong version of forge

protolambda: Can conffirm, my bad. Running `pnpm update:foundry` worked

Inphi (approved):
Thank you!

PR #7362, opened 2023-09-25
op-node: remove beta prefix from rollup.load-protocol-versions and rollup.halt flags

protolambda:
**Description**

This functionality is going out of beta, into regular release-candidate and release cycle.

PR #7363, opened 2023-09-25
op-node: update flags, remove BETA_ prefix

protolambda:
In #7362 we forgot to remove the `BETA_` prefix from the op-node flags.
Added a test for flag naming consistency so this can't happen again in the op-node.


sebastianst (approved):

protolambda:
Fixed go imports lint in the test. IDE put it in the wrong place

protolambda:
@Mergifyio refresh

tynes (approved):

PR #7364, opened 2023-09-25
Sequencer fencing endpoint

kahuang:
**Description**

Adds an optional fencing endpoint for block production. This provides a more fault-tolerant and consistent way to guarantee there is only one sequencer sequencing blocks at a time if you're running multiple sequencers with a consensus/election protocol.
**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/op-node/flags/flags.go
+++ b/op-node/flags/flags.go
@@ -139,6 +139,13 @@ var (
 		Required: false,
 		Value:    4,
 	}
+	SequencerFencingCheckEndpoint = &cli.StringFlag{
+		Name:     "sequencer.fencing-check-endpoint",
+		Usage:    "Optional endpoint that must return 200 when the sequencer should sequence blocks, otherwise return non-200.",
+		EnvVars:  prefixEnvVars("SEQUENCER_FENCING_CHECK_ENDPOINT"),
+		Required: false,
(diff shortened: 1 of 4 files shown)


op-node/flags/flags.go
```diff
@@ -139,6 +139,13 @@ var (
 		Required: false,
 		Value:    4,
 	}
+	SequencerFencingCheckEndpoint = &cli.StringFlag{
```
trianglesphere, line 142: Can we make this hidden & marked as still subject to change?

op-node/rollup/driver/state.go
```diff
@@ -266,6 +267,32 @@ func (s *Driver) eventLoop() {
 
 		select {
 		case <-sequencerCh:
+			if s.driverConfig.SequencerFencingCheckEndpoint != "" {
```
trianglesphere, line 270: Please pull this out into a discrete function.

trianglesphere (commented):
As mentioned, we can bring this in, but I do expect it to change in the future as we figure out better ways to perform this check.

kahuang:
I think what we'll do is just run this in our own fork and iterate, then come back when the interfaces are finalized

PR #7366, opened 2023-09-25
feat(op-challenger): Split Trace Provider

refcell:
**Description**

Introduces the Split Trace Provider that routes requests to its underlying trace providers
based on the game depth of the positions in the requests.

**Tests**

Unit tests around the split trace provider methods.

**Metadata**

Fixes https://github.com/ethereum-optimism/client-pod/issues/43

### DIFF
--- a/op-challenger/game/fault/trace/split/provider.go
+++ b/op-challenger/game/fault/trace/split/provider.go
@@ -1,0 +1,78 @@
+package split
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ethereum-optimism/optimism/op-challenger/game/fault/types"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+var (
+	GetStepDataErr = fmt.Errorf("GetStepData not supported")
+	NoProvidersErr = fmt.Errorf("no trace providers configured")
+)
+
+// todo(refcell): the Get method traceIndex is update to be a Position so we have the depth
+// var _ types.TraceProvider = (*SplitTraceProvider)(nil)
+
+// SplitTraceProvider is a [types.TraceProvider] implementation that
+// routes requests to the correct internal trace provider based on the
+// depth of the requested trace.
+type SplitTraceProvider struct {
+	logger     log.Logger
+	providers  []types.TraceProvider
+	depthTiers []uint64
+}
+
+func NewTraceProvider(logger log.Logger, providers []types.TraceProvider, depthTiers []uint64) *SplitTraceProvider {
+	return &SplitTraceProvider{
+		logger:     logger,
+		providers:  providers,
+		depthTiers: depthTiers,
+	}
+}
+
+func (s *SplitTraceProvider) providerForDepth(depth uint64) types.TraceProvider {
+	for i, tier := range s.depthTiers {
+		if depth <= tier {
+			return s.providers[i]
+		}
+	}
+	return s.providers[len(s.providers)-1]
+}
+
+// Get routes the Get request to the internal [types.TraceProvider] that
+// that serves the trace index at the depth.
+func (s *SplitTraceProvider) Get(ctx context.Context, pos types.Position) (common.Hash, error) {
+	if len(s.providers) == 0 {
+		return common.Hash{}, NoProvidersErr
+	}
+	return s.providerForDepth(uint64(pos.Depth())).Get(ctx, pos.TraceIndex(pos.Depth()))
+}
+
+// AbsolutePreStateCommitment returns the absolute prestate from the lowest internal [types.TraceProvider]
+func (s *SplitTraceProvider) AbsolutePreStateCommitment(ctx context.Context) (hash common.Hash, err error) {
+	if len(s.providers) == 0 {
+		return common.Hash{}, NoProvidersErr
+	}
+	return s.providers[len(s.providers)-1].AbsolutePreStateCommitment(ctx)
+}
+
+// AbsolutePreState routes the AbsolutePreState request to the lowest internal [types.TraceProvider].
+func (s *SplitTraceProvider) AbsolutePreState(ctx context.Context) (preimage []byte, err error) {
+	if len(s.providers) == 0 {
+		return nil, NoProvidersErr
+	}
+	return s.providers[len(s.providers)-1].AbsolutePreState(ctx)
+}
+
+// GetStepData routes the GetStepData request to the lowest internal [types.TraceProvider].
+func (s *SplitTraceProvider) GetStepData(ctx context.Context, i uint64) (prestate []byte, proofData []byte, preimageData *types.PreimageOracleData, err error) {
+	if len(s.providers) == 0 {
+		return nil, nil, nil, NoProvidersErr
+	}
+	return s.providers[len(s.providers)-1].GetStepData(ctx, i)
+}
--- a/op-challenger/game/fault/trace/split/provider.go
+++ b/op-challenger/game/fault/trace/split/provider.go
@@ -35,13 +35,17 @@ func NewTraceProvider(logger log.Logger, providers []types.TraceProvider, depthT
 	}
 }
 
-func (s *SplitTraceProvider) providerForDepth(depth uint64) types.TraceProvider {
+func (s *SplitTraceProvider) providerForDepth(depth uint64) (uint64, types.TraceProvider) {
+	reduced := uint64(0)
 	for i, tier := range s.depthTiers {
 		if depth <= tier {
-			return s.providers[i]
+			return reduced, s.providers[i]
+		}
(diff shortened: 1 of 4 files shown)


refcell:
* develop
  * **PR #7366** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7366" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7366?utm_source=stack-comment).

op-challenger/game/fault/trace/split/provider.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+
+	"github.com/ethereum-optimism/optimism/op-challenger/game/fault/types"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+var (
+	GetStepDataErr = fmt.Errorf("GetStepData not supported")
```
tynes: nit: `errors.New`

refcell:
This is currently blocked by #7185 which modifies the trace provider to accept positions as opposed to trace indices.

op-challenger/game/fault/trace/split/provider.go
ajsutton: Doesn't seem to have actually been changed. :)

op-challenger/game/fault/trace/split/provider.go
```diff
@@ -0,0 +1,83 @@
...
+	NoProvidersErr = fmt.Errorf("no trace providers configured")
```
ajsutton: errors.New here too.

op-challenger/game/fault/trace/split/provider.go
```diff
@@ -0,0 +1,83 @@
...
+
+var _ types.TraceProvider = (*SplitTraceProvider)(nil)
+
+// SplitTraceProvider is a [types.TraceProvider] implementation that
+// routes requests to the correct internal trace provider based on the
+// depth of the requested trace.
+type SplitTraceProvider struct {
+	logger     log.Logger
+	providers  []types.TraceProvider
+	depthTiers []uint64
```
ajsutton: We don't need to make this support a generic number of providers - just two. We should keep it as simple as possible.

op-challenger/game/fault/trace/split/provider.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return s.providers[len(s.providers)-1].AbsolutePreState(ctx)
+}
+
+// GetStepData routes the GetStepData request to the lowest internal [types.TraceProvider].
+func (s *SplitTraceProvider) GetStepData(ctx context.Context, i types.Position) (prestate []byte, proofData []byte, preimageData *types.PreimageOracleData, err error) {
+	if len(s.providers) == 0 {
+		return nil, nil, nil, NoProvidersErr
+	}
+	return s.providers[len(s.providers)-1].GetStepData(ctx, i)
```
ajsutton: This isn't translating the position correctly so the depth will be greater than the last provider expects (since it's using the depth of the combined game instead of the single bottom-most provider).

And while the request for step data should only come at the maximum game depth, for simplicity and consistency I'd just have this do exactly the same thing as `Get` does where it looks up the provider based on the depth and Localizes the position.

op-challenger/game/fault/types/position.go
```diff
@@ -25,6 +25,15 @@ func (p Position) MoveRight() Position {
...
 }
 
+// Localize returns a new position for a subtree.
+// reduced is the number of levels to reduce the depth by.
+func (p Position) Localize(reduced uint64) Position {
+	newPosDepth := uint64(p.depth) - reduced
+	nodesAtDepth := 1 << newPosDepth
+	newIndexAtDepth := p.indexAtDepth % nodesAtDepth
+	return NewPosition(int(newPosDepth), newIndexAtDepth)
+}
```
ajsutton: We should add a bunch of unit tests around this. It will be very hard to debug if it's wrong (and I'm old enough to have learnt not to trust my abilities with this kind of math... 😄).

I'm not really sold on the `Localize` name either but hard to think of something better.  The long version would be `RelativeToAncestorAtDepth` which I'd be seriously tempted to go with because it's quite explicit about what its doing.

op-challenger/game/fault/trace/split/provider_test.go
```diff
@@ -0,0 +1,148 @@
...
+	mockCommitment = common.HexToHash("0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb")
+)
+
+func TestGet(t *testing.T) {
+	t.Run("ErrorBubblesUp", func(t *testing.T) {
+		mockOutputProvider := mockTraceProvider{getError: mockGetError}
+		splitProvider := SplitTraceProvider{
+			logger:     testlog.Logger(t, log.LvlInfo),
+			providers:  []types.TraceProvider{&mockOutputProvider},
+			depthTiers: []uint64{40, 20},
```
ajsutton: I'm a little confused at having one trace provider but two depth tiers.  Though I suspect this becomes simpler if we just hard code that there's always two providers.

op-challenger/game/fault/trace/split/provider.go
refcell: Ok - that wasn't clear to me from [the issue description](https://github.com/ethereum-optimism/client-pod/issues/43) which detailed a generic split provider that routes based on depth

op-challenger/game/fault/types/position.go
refcell: I'm ok with `RelativeToAncestorAtDepth` as well. If methods have good inline documentation, it's propagated to tooltips and descriptors so it will be understandable. In fact, even having a method with a verbose name like `RelativeToAncestorAtDepth` can be very confusing without good inline docs IMHO. It's a habit I think our team needs to work on across the codebase.

op-challenger/game/fault/trace/split/provider.go
```diff
@@ -0,0 +1,72 @@
...
+
+var _ types.TraceProvider = (*SplitTraceProvider)(nil)
+
+// SplitTraceProvider is a [types.TraceProvider] implementation that
+// routes requests to the correct internal trace provider based on the
+// depth of the requested trace.
+type SplitTraceProvider struct {
+	logger         log.Logger
+	topProvider    types.TraceProvider
+	bottomProvider types.TraceProvider
```
ajsutton, line 20: This isn't going to work either - it needs to be a factory for the bottom provider as the trace will differ based on which output roots its going between. I'd forgotten that detail until I re-read the issue too actually. :)

> Should accept a  TraceProvider that provides the "top" of the game and a factory to create a TraceProvider given the the pre and post claims from the top level game. In the OUTPUT_ROOT_CANNON game type those pre and post claims will be the agreed output proposal and the first disputed output proposal.

We could probably do that as a separate PR though.

op-challenger/game/fault/trace/split/provider_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+func TestGet(t *testing.T) {
+	t.Run("ErrorBubblesUp", func(t *testing.T) {
+		mockOutputProvider := mockTraceProvider{getError: mockGetError}
+		splitProvider := newSplitTraceProvider(t, &mockOutputProvider, nil, 40)
+		_, err := splitProvider.Get(context.Background(), types.NewPosition(1, 0))
+		require.ErrorIs(t, err, mockGetError)
+	})
+
+	t.Run("ReturnsCorrectOutput", func(t *testing.T) {
```
ajsutton:
```suggestion
	t.Run("ReturnsCorrectOutputFromTopProvider", func(t *testing.T) {
```

op-challenger/game/fault/trace/split/provider_test.go
```diff
@@ -0,0 +1,148 @@
...
+		mockOutputProvider := mockTraceProvider{getOutput: mockOutput}
+		splitProvider := newSplitTraceProvider(t, &mockOutputProvider, nil, 40)
```
ajsutton:
```suggestion
		splitProvider := newSplitTraceProvider(t, &mockOutputProvider, &mockTraceProvider{}, 40)
```

Probably worth ensuring the setup is valid rather than having a `nil` bottom provider.

ajsutton (approved):
LGTM. We'll need to change the bottom provider to be a factory but that probably is better to do in a follow up PR.

op-challenger/game/fault/trace/split/provider.go
refcell: Since we're going to need to work towards this issue in other prs anyway, will put in a new pr so these changes can land

Issue #7075, opened 2023-08-31
Document CI environment and dependencies setup


jyellick:
The optimism monorepo utilizes CircleCI to execute its CI.  This configuration is documented in `.circleci/config.yaml`.  However, simply cloning this repo and associating it with a CircleCI account is unsufficient to actually make CI pass.  There are 3rd party services such as the GCP registry, and possibly others that are documented only insofar as they are referenced in the CI config file.

## Why is this a problem?

First and foremost, for anyone without write access to the Optimism repository, they cannot successfully run CI tests in their own fork.  This means that developers can't leverage CI to help them iterate in their own development while preparing code for a PR, or in their own general experimenation.

Secondly, for any individual or organization who does have a CircleCI account, their pull requests will end up executing in their own CI context resulting in failures on the PR.  Reviewers know that for instance the 'publish' steps of CI do not need to be green in order to merge, but it's still distracting and confusing.

Finally, as a community focused distribution with many op-stack derivatives, it's important for the health of the ecosystem that forks and other op-stacks can focus on contributing their own unique value, rather than duplicating effort to maintain parallel CI forks.

## How can this be addressed

Ideally, the CI configuration and its dependencies could be exposed through some sort of infrastructure as code repository.  Say, Terraform or something similar.  In this way, other individuals or organizations could simply review the code and execute it in a project context to stand up the full CI system.  Possibly this code already exists and is available, but it wasn't obvious to me where to find it.

Barring the ability to stand up the CI and its dependencies via infrastructure as code, even simple documentation about what services and accounts are required, what variables need to be set, and the general structure of the pipelines promotion, releases, etc. would be very helpful.

Note, @ajsutton [suggested that I open this issue](https://github.com/ethereum-optimism/optimism/pull/7049#issuecomment-1699839373).

BlocksOnAChain:
I will close this as a stale issue.

PR #7368, opened 2023-09-25
Reusable pipeline components

InoMurko:
**Description**
This is a response to issue:
https://github.com/ethereum-optimism/optimism/issues/7075
and comment:
https://github.com/ethereum-optimism/optimism/pull/7049#issuecomment-1699839373

It includes the necessery work to migrate away from deprecated Container Registry to Artifact Registry:
https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr

**Tests**

/

**Additional context**

Parameterizing the testing/image building/release pipeline that allows organizations outside of Optimisms to utilize the same pipeline flow in a forked or cloned repository. The simple explanation is that forks that work on the same codebase and pull upstream changes want to have as little conflicts as possible. 

To make this PR successful I would need the cirleCI env to be updated with 
```
GCP_ARTIFACT_REPOSITORY=oplabs-tools-artifacts
GCP_PROJECT_ID=<provide value>
```

Organizations outside of Optimism need a functioning GCP account with enabled APIs and Artifact Registry.
Link Cirleci and GCP via https://circleci.com/docs/openid-connect-tokens/ 

**Metadata**

- https://github.com/ethereum-optimism/optimism/issues/7075

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ commands:
 jobs:
   cannon-go-lint-and-test:
     docker:
-      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/images/ci-builder:latest
+      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/ci-builder:latest
     resource_class: medium
     steps:
       - checkout
@@ -92,7 +92,7 @@ jobs:
           path: /tmp/test-results
   cannon-build-test-vectors:
     docker:
-      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/images/ci-builder:latest
+      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/ci-builder:latest
     resource_class: medium
     steps:
       - checkout
@@ -105,7 +105,7 @@ jobs:
 
   pnpm-monorepo:
     docker:
-      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/images/ci-builder:latest
+      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/ci-builder:latest
     resource_class: xlarge
     steps:
       - checkout
@@ -320,7 +320,7 @@ jobs:
 
   contracts-bedrock-coverage:
     docker:
-      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/images/ci-builder:latest
+      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/ci-builder:latest
     resource_class: large
     steps:
       - checkout
@@ -345,7 +345,7 @@ jobs:
 
   contracts-bedrock-tests:
     docker:
-      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/images/ci-builder:latest
+      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/ci-builder:latest
     resource_class: xlarge
     steps:
       - checkout
@@ -365,7 +365,7 @@ jobs:
 
   contracts-bedrock-checks:
     docker:
-      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/images/ci-builder:latest
+      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/ci-builder:latest
     resource_class: xlarge
     steps:
       - checkout
@@ -452,7 +452,7 @@ jobs:
 
   contracts-bedrock-slither:
     docker:
-      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/images/ci-builder:latest
+      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/ci-builder:latest
     resource_class: large
     steps:
       - checkout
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ commands:
             echo $CIRCLE_OIDC_TOKEN > << parameters.oidc_token_file_path >>
             # Create a credential configuration for the generated OIDC ID Token
             gcloud iam workload-identity-pools create-cred-config \
-                "projects/${<< parameters.project_id >>}/locations/global/workloadIdentityPools/${<< parameters.workload_identity_pool_id >>}/providers/${<< parameters.workload_identity_pool_provider_id >>}"\
+                "projects/${<< parameters.project_number >>}/locations/global/workloadIdentityPools/${<< parameters.workload_identity_pool_id >>}/providers/${<< parameters.workload_identity_pool_provider_id >>}"\
```
InoMurko, line 37: reading the docs here: https://cloud.google.com/sdk/gcloud/reference/iam/workload-identity-pools/create-cred-config#AUDIENCE

this is actually project number, rather than project ID.

.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ commands:
 jobs:
   cannon-go-lint-and-test:
     docker:
-      - image: us-docker.pkg.dev/oplabs-tools-artifacts/images/ci-builder:latest
+      - image: us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/images/ci-builder:latest
```
InoMurko: Project IDs are sensitive information, but are not secrets. Disclosing a Project ID does not put your account at risk. I suggest storing your project ID in the cirleci variables.

.github/workflows/release-snapshot.yml
```diff
@@ -12,15 +12,23 @@ on:
 jobs:
   snapshot-snapshot:
     name: Publish snapshot release to npm
-    if: github.repository == 'ethereum-optimism/optimism'
```
InoMurko, line 15: GH actions if statements cannot be parametrized hence the hack below that achieves the same :)

.circleci/config.yml
InoMurko: image

Cirleci says project id, but it's actually project number - https://circleci.com/docs/openid-connect-tokens/ 
so I'm unsure of the purpose of their confusion.

.env
```diff
@@ -0,0 +1,3 @@
+GCP_ARTIFACT_REPOSITORY=oplabs-tools-artifacts
+GCP_PROJECT_ID=<>
+REPO=ethereumoptimism/optimism
```
sebastianst, line 3:
```suggestion
REPO=ethereum-optimism/optimism
```
?

trianglesphere:
This is breaking CI. It appears to be b/c it's not finding the appropriate CI builder image.

InoMurko:
> This is breaking CI. It appears to be b/c it's not finding the appropriate CI builder image.

Hey @trianglesphere yep! There could be two reasons:
CirleCI does not have environment variables set:
`GCP_ARTIFACT_REPOSITORY=oplabs-tools-artifacts and GCP_PROJECT_ID=<provide value>`

and then ci-builder would probably be found. If not, a new build would need to be spawned so that it's pushed into Artifact Registry (instead of Container Registry), but the docs says things are backwards compatible.

InoMurko:
@Inphi @trianglesphere are you interested in making this change perhaps? it helps forks adapt to their own pipeline.

Inphi:
Hey @InoMurko. Sorry for not getting to this earlier. I think this is a good problem to solve and the approach in this PR is generally fine. Appreciate you taking this on. Let me have the GCP_PROJECT_ID setup so we can unblock CI.

.circleci/config.yml
Inphi: This should be `us-docker.pkg.dev/${GCP_PROJECT_ID}/${GCP_ARTIFACT_REPOSITORY}/ci-builder:latest`. As `images` is the repository name we use.

I've setup both envs in circleci. You should be able to run CI once you've made the above changes.

InoMurko:
```
Starting container us-docker.pkg.dev///ci-builder:latest
  image cache not found on this host, downloading us-docker.pkg.dev///ci-builder:latest

```

can you double check the envs are there? thanks!

Inphi: Yeah, the envs are setup correctly for the project. I think the issue here is the envs can't be used in this context; as an `image` value.

Inphi: Perhaps defining them as a `parameter` would work. Then you can interpolate the values in the `image` value like:
`us-docker.pkg.dev/<<parameters.gcp_project_id>>/<<parameters.gcp_artifact_repository>>/ci-builder:latest`

InoMurko: They can be read. This is a screenshot from our cirleci run with the below config.yml.
image

https://github.com/bobanetwork/v3-anchorage/blob/develop/.circleci/config.yml#L70

I'm just thinking what else could be a problem. Thank you for sticking with me here.

Inphi: We have those envs setup for the optimism monorepo project, rather than in a circleci context. how are your envs setup?

InoMurko: I've set them here: https://app.circleci.com/settings/project/github/bobanetwork/v3-anchorage/environment-variables

Inphi: The environment variables are picked up when I trigger a circleci pipeline myself. See https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/29773/workflows/39eae03c-1b9e-4cdd-b71d-a59ae9fead8a/jobs/1406434.
Not sure if you have necessary permissions to do this, but could you try triggering a fresh new pipeline, rather than restart an existing one? My guess is there's some caching going on preventing the new envs from being picked up.

InoMurko: Oh cool, now they're being picked up!
I could open a new PR and see what happens there? 


Additionally, is this the right repo? Or should I drop `images` now?
```
      repo:
        description: Docker repo
        type: string
        default: "${GCP_ARTIFACT_REPOSITORY}/images"
```

Also,
the **.env** file would need to be populated in order to maintain equal _local development environment_ and _GH actions_ that use docker-compose for builds (like release-docker-canary.yml).
```
- name: Build and push
        uses: docker/build-push-action@v2
        with:
          context: .
          file: ./ops/docker/Dockerfile.packages
          target: fault-mon
          push: true
          build-args:
            - ${{ env.GCP_ARTIFACT_REPOSITORY }}
            - ${{ env.GCP_PROJECT_ID }}
          tags: ${{ env.ORG }}/fault-mon:${{ needs.canary-publish.outputs.canary-docker-tag }}
```

docker-compose.yml
```
  l2:
    build:
      context: .
      dockerfile: Dockerfile.l2
      args:
         - GCP_PROJECT_ID=${GCP_PROJECT_ID:-populate-manually}
         - GCP_ARTIFACT_REPOSITORY=${GCP_ARTIFACT_REPOSITORY:-oplabs-tools-artifacts}
```

Inphi: You can drop `images` so the repo is `${GCP_ARTIFACT_REPOSITORY}`.

> Also,
the .env file would need to be populated in order to maintain equal local development environment and GH actions that use docker-compose for builds (like release-docker-canary.yml).

Good callout. Not exactly sure how this is setup. I'll get back to you on whether we can update the gh envs.

InoMurko: Cool, will do. And let me know on other stuff. Thanks! 🙏🏻

PR #7372, opened 2023-09-25
chore: move `op-signer` into `op-service/signer`

### LINKED ISSUE #6731: op-service: crypto package / op-signer dependency cycle cleanup
The op-service / op-signer have an odd interdepdency.
We can fix this by moving the `opsigner.CLIConfig` into op-service.

Sabnock01:
**Description**

This PR removes the interdependency between `op-service` and `op-signer` by moving the `CLIConfig` struct and associated methods from `op-signer` into `op-service` and renaming it to `SignerCLIConfig`.

**Tests**

I've adjusted the tests in [`config_test.go`](https://github.com/ethereum-optimism/optimism/blob/develop/op-signer/client/config_test.go) to be compatible with this change. Let me know if we want anything more done on this front or if I missed something.

**Metadata**

- Fixes #6731


op-signer/client/config.go
```diff
@@ -32,37 +30,15 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
 
-type CLIConfig struct {
-	Endpoint  string
-	Address   string
-	TLSConfig optls.CLIConfig
-}
-
-func NewCLIConfig() CLIConfig {
-	return CLIConfig{
+func NewCLIConfig() optls.SignerCLIConfig {
```
sebastianst: You probably also want to move the ctors `NewCLIConfig` and `ReadCLIConfig` along with all tests that only test the `CLIConfig` to the new package.

op-service/crypto/signature.go
```diff
@@ -18,6 +18,7 @@ import (
 
 	hdwallet "github.com/ethereum-optimism/go-ethereum-hdwallet"
 	opsigner "github.com/ethereum-optimism/optimism/op-signer/client"
+	optls "github.com/ethereum-optimism/optimism/op-service/tls"
```
sebastianst: I think we haven't resolved the dependency "cycle" yet, because `SignerFactoryFromConfig` below calls `opsigner.NewSignerClientFromConfig` and the `op-signer/client` package still depends on the `tls` package.

sebastianst (commented):
There's still a "cycle" op-serice/crypto -> op-signer/client -> op-service/tls. Not sure if @protolambda's suggestion to just move the `CLIConfig` works.

op-signer/client/config.go
Sabnock01: Good call. Moved the rest of `config.go` and the associated tests over to `op-service`.

op-service/crypto/signature.go
Sabnock01: Nice catch. Will wait for @protolambda's feedback.

trianglesphere (commented):
This is definitely scope creep & a design question, but what about moving the entire op-signer package to `op-service/signer`? The op-signer is the op-signer client, not the server, so it's an easy fit into op-service which is a bunch of client/shared server/library components

Sabnock01:
> This is definitely scope creep & a design question, but what about moving the entire op-signer package to `op-service/signer`? The op-signer is the op-signer client, not the server, so it's an easy fit into op-service which is a bunch of client/shared server/library components

Willing to do this. Seems like a reasonable suggestion.

**Update**: Made `op-service/signer` virtually 1:1 with `op-signer/client` except renamed `config*.go` to `cli*.go` to match the `op-service` convention.

zhwrd (approved):

trianglesphere (approved):

PR #7381, opened 2023-09-26
[4844] remove duplicate initRuntimeConfig call

roberto-bayardo:
**Description**

initRuntimeConfig code ended up duplicated, probably from one of the rebases, which was causing the e2e test to fail


roberto-bayardo:
cc @protolambda

protolambda (approved):

PR #7387, opened 2023-09-26
contracts-bedrock: fix text coloring bug

tynes:
**Description**

Github seems to have issues with whatever library they are using to apply colors to the code for solidity code. The `'` triggers it to think that it is in a string when it is not actually in a string because the `'` is wrapped within two `"`. Just delete the `'` to fix this issue. The create2 salt is already not completely standard but it should become standard when we finally have tooling to easily add a chain to the superchain through the deploy script, this entrypoint would deploy the proxies and initialize them to the standard implementations.


maurelian (approved):
yay, thank you.

refcell (approved):

PR #7390, opened 2023-09-26
contracts-bedrock: fix compiler warnings

tynes:
**Description**

Fix the compiler warnings in the `DelayedVetoable` tests. There were unused variables returned from low level calls. There are alternative compiler warning if there is no assignment to the variables, so the solution is to use them either in an assertion or just reference them on their own line. Silly compiler.


maurelian (approved):
Thanks for fixing these

refcell (approved):

PR #7411, opened 2023-09-27
Revert: Claim Count Metrics

ajsutton:
**Description**

Reverts https://github.com/ethereum-optimism/optimism/pull/7264/ because the new game claim count metric uses an unbounded label value (the game address). This will create a new prometheus time series for every game the challenger touches. This is explicitly recommended against by prometheus: https://prometheus.io/docs/practices/naming/#labels

This reverts commit eb76a0db06214985940265e4a2fae6da91436fce, reversing changes made to b15e3ffea35eb877a13910dececa84e40ca4b9a3.

### DIFF
--- a/op-challenger/game/fault/agent.go
+++ b/op-challenger/game/fault/agent.go
@@ -30,7 +30,6 @@ type ClaimLoader interface {
 
 type Agent struct {
 	metrics                 metrics.Metricer
-	fdgAddr                 common.Address
 	solver                  *solver.GameSolver
 	loader                  ClaimLoader
 	responder               Responder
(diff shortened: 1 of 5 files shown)


ajsutton:
@refcell Not sure what the back story was for adding this metric. If there was a particularly useful thing we got out of it then it may be worth considering keeping but otherwise it's probably worth following the guidelines to avoid unbounded label values.

protolambda (approved):
Hope we can find another way of creating this type of metric, without the labels. Maybe we can try the status-chart type? (and has game IDs to the value instead, like we do with the blockhash ref-metric?).

tynes (approved):

PR #7412, opened 2023-09-27
Allow output deletion by Challenger, Security Council or Foundation

maurelian:
**Description**

The changes in this PR are primarily concerned with enabling any one of three entities to challenge (delete) an output. For any given OP Chain, those entities are Security Council, the Foundation, or the Challenger (as set by the Chain Governor).

This PR also allows the Chain Governor to update the Challenger role on the System Config.

A side effect of these changes was a 'stack too deep' error in solidity caused by adding a new argument to the `initialize` function. This was addressed by combining two values into a single `GasConfig` struct.


packages/contracts-bedrock/foundry.toml
```diff
@@ -23,6 +23,7 @@ extra_output = ['devdoc', 'userdoc', 'metadata', 'storageLayout']
 bytecode_hash = 'none'
 build_info = true
 build_info_path = 'artifacts/build-info'
+via_ir = true
```
tynes: This is going to cause build times to be way too slow

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
```diff
@@ -300,4 +308,12 @@ contract L2OutputOracle is Initializable, ISemver {
     function computeL2Timestamp(uint256 _l2BlockNumber) public view returns (uint256) {
         return startingTimestamp + ((_l2BlockNumber - startingBlockNumber) * L2_BLOCK_TIME);
     }
+
+    /// @notice Returns a boolean indicating whether the given address can delete an output.
+    /// @param _addr The address to check.
+    /// @return True if the address can delete an output, false otherwise.
+    function mayDelete(address _addr) public view returns (bool) {
+        return _addr == systemConfig.challenger() || _addr == systemConfig.superchainConfig().initiator()
```
tynes: This does extra calls that are not necessary

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -170,6 +177,7 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
         ResourceMetering.ResourceConfig memory _config,
         uint256 _startBlock,
         address _batchInbox,
+        address _challenger,
```
tynes: Stack too deep here? If so, we should consolidate the gas config into a struct that is read into memory

packages/contracts-bedrock/foundry.toml
maurelian: Ack. Just included it for debugging. Should've included a note as such.

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
maurelian: I think this is the best way to do it, here is my reasoning:

1. The addresses being retrieved should be stored in as few locations as possible: 
	- systemConfig in the L2OutputOracle
	- superchainConfig in the SystemConfig
	- initiator and vetoer in the SuperChainConfig
1. Based on the above assertion, the only optimization that I can see would be to add a getter to the SuperChainConfig which returns both the initiator and vetoer. IMO the additional code maintenance ($100/loc), is not justified, because:
1. This method will only be executed on-chain when an output is being deleted, so it is not a priority for optimization.

op-chain-ops/genesis/config.go
```diff
@@ -31,6 +31,16 @@ var (
 	ErrInvalidImmutablesConfig = errors.New("invalid immutables config")
 )
 
+// todo(maurelian): Is it possible to reuse the type def in the superchainconfig.go bindings?
+// SequencerKeys is an auto generated low-level Go binding around an user-defined struct.
+type SequencerKeys struct {
```
tynes: We can definitely reuse the typedef from the bindings, I am not super opinionated on whether we use it here or not

op-chain-ops/genesis/config.go
```diff
@@ -31,6 +31,16 @@ var (
...
+	BatcherHash       common.Hash
+	UnsafeBlockSigner common.Address
+}
+
+// SequencerKeysArray is an array of SequencerKeys.
+type SequencerKeysArray []SequencerKeys
```
tynes: nit on the naming here, I don't love calling things based on implementation details, ie the array suffix in this case. Perhaps we can use the bindings type or rename the above to `SequencerKey` (no s) or `SequencerKeyPair` (then we are coupled to 2 keys which i think is fine for now) or `SequencerIdentity` but perhaps it should match what is in solidity, i just want to not use the type `SequencerKeysArray` and rename it to something like `SequencerKeys` (with the plural) or `SequencerSet`. This begs the question, the solidity allows for a set where implementing as a slice allows for duplicates so its not technically a set

op-chain-ops/genesis/config.go
tynes: I think we should use our own type so that we can add a function to the type to compute the hash

op-chain-ops/genesis/config.go
tynes: I don't think it is worth the complexity to worry about implementing it as a set, just adds extra complexity to json parsing

op-chain-ops/genesis/config.go
```diff
@@ -204,6 +214,15 @@ type DeployConfig struct {
 	// RequiredProtocolVersion indicates the protocol version that
 	// nodes are recommended to adopt, to stay in sync with the network.
 	RecommendedProtocolVersion params.ProtocolVersion `json:"recommendedProtocolVersion"`
+	// SecurityCouncil represents the address of the Security Council role.
+	SecurityCouncil common.Address `json:"securityCouncil"`
```
tynes: For naming these things, what do you think about `"contractName" || "contractVariable"` as a standard, ie `l2OutputOracleChallenger` would follow this pattern. Out of scope of this PR but if we are breaking deploy config we should standardize on something sane and stick to it

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
tynes: What i mean is that you call `systemConfig.superchainConfig()` twice

maurelian: oh lol, fair point

tynes: If you feel strongly about it i'm fine with leaving it but there really isn't a reason to call `systemConfig.superchainConfig()` twice

```solidity
SuperchainConfig superchainConfig = systemConfig.superchainConfig();
return _addr == systemConfig.challenger() || _addr == superchainConfig().initiator() || _addr == superchainConfig().vetoer();
```

packages/contracts-bedrock/src/L1/OptimismPortal.sol
```diff
@@ -141,12 +141,19 @@ contract OptimismPortal is Initializable, ResourceMetering, ISemver {
...
+        return superchainConfig.guardian();
+    }
+
     /// @notice Getter for the Guardian
     /// @custom:legacy
     function guardian() external view returns (address) {
         return superchainConfig.guardian();
     }
 
+
```
tynes: nit: no need for this diff

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -188,13 +191,17 @@ contract SuperchainConfig is Initializable, ISemver {
...
+        paused_ = _getValue(PAUSED_SLOT) > block.timestamp;
     }
 
+    // todo(maurelian): we might need a repause() getter which is only callable by the
+    // security council. And is the only way to extend an active pause.
+    // This would enable us to distribute the one-time presigned pause tx, but restrict repausing.
     /// @notice Pauses withdrawals.
-    function pause() external {
+    function pause(uint256 duration) external {
         require(msg.sender == guardian(), "SuperchainConfig: only guardian can pause");
```
tynes, line 167: I feel like this should be eip712 based in the future but its fine for now

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -99,6 +101,9 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
     ///         This can only be set via upgrade.
     SuperchainConfig public superchainConfig;
 
+    /// @notice Challenger address
```
tynes: nit: better doc comment, what can the challenger do?

packages/contracts-bedrock/src/L1/SystemConfig.sol
tynes: A type that has the overhead, scalar, gaslimit together seems like a good type to combine into gas config

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -19,11 +19,13 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
...
     /// @custom:value UNSAFE_BLOCK_SIGNER  Represents an update to the signer key for unsafe
     ///                                    block distrubution.
+    /// @custom:value CHALLENGER           Represents an update to the challenger address.
     enum UpdateType {
         BATCHER,
         GAS_CONFIG,
         GAS_LIMIT,
-        UNSAFE_BLOCK_SIGNER
+        UNSAFE_BLOCK_SIGNER,
+        CHALLENGER
```
tynes: We don't need the challenger in this enum, it will break the `op-node` if an unknown type is emitted here

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -416,4 +425,19 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
...
+        _setChallenger(_challenger);
+    }
+
+    /// @notice Internal function for updating the challenger.
+    /// @param _challenger New challenger.
+    function _setChallenger(address _challenger) internal {
+        challenger = _challenger;
+
+        bytes memory data = abi.encode(_challenger);
+        emit ConfigUpdate(VERSION, UpdateType.CHALLENGER, abi.encode(data));
```
tynes: Remove this `emit ConfigUpdate` otherwise it'll break `op-node`

tynes: It was always said to be best practice to emit events so that the set of events can be used to reconstruct the state without needing to execute all txs. Then shadow events became a thing where you basically can insert events that don't consume gas or impact state into execution and index them which makes events kind of obsolete if you are running these shadow event indexers

tynes: I am torn on whether or not we want an event here, i dont think we need one

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
maurelian: ya, I'm on board.

op-chain-ops/genesis/config.go
maurelian: Addressed this in the parent PR.

op-chain-ops/genesis/config.go
maurelian: Addressed this in the parent PR.

packages/contracts-bedrock/src/L1/SystemConfig.sol
maurelian: OK, replaced with a dedicated event for challenger changes.

packages/contracts-bedrock/src/L1/SystemConfig.sol
maurelian: I think that including one is the safer option?
Have added it anyways.

packages/contracts-bedrock/scripts/DeployConfig.s.sol
```diff
@@ -15,6 +16,7 @@ contract DeployConfig is Script {
     string internal _json;
 
     address public finalSystemOwner;
+    // todo(maurelian): rename as this has been moved to the SuperchainConfig
```
tynes: its hard to tell which todos you want to resolve now and in the future

packages/contracts-bedrock/test/SystemConfig.t.sol
```diff
@@ -309,7 +313,7 @@ contract SystemConfig_Setters_TestFail is SystemConfig_Initializer {
 
 contract SystemConfig_Setters_Test is SystemConfig_Initializer {
     /// @dev Tests that `setSequencer` updates the batcher hash successfully.
-    function testFuzz_setSequencer_succeeds(Types.SequencerKeys calldata sequencer) external {
+    function testFuzz_setSequencer_succeeds(Types.SequencerKeyPair calldata sequencer) external {
```
tynes: I didn't realize you could fuzz with custom types

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -181,17 +196,18 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
 
         // These are set in ascending order of their UpdateTypes.
         _setBatcherHash(_batcherHash);
-        _setGasConfig({ _overhead: _overhead, _scalar: _scalar });
+        _setGasConfig({ _overhead: _gasConfig.overhead, _scalar: _gasConfig.scalar });
```
tynes: Thoughts on not unwrapping the type here? Did you not do that to preserve the abi?

maurelian: Good call out.

packages/contracts-bedrock/test/SystemConfig.t.sol
maurelian: I double checked, it generates unique inputs on each run.

packages/contracts-bedrock/scripts/DeployConfig.s.sol
maurelian: Yeah, that's fair, I've just removed them for now.

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
```diff
@@ -103,8 +105,10 @@ contract L2OutputOracle is Initializable, ISemver {
 
         startingTimestamp = _startingTimestamp;
         startingBlockNumber = _startingBlockNumber;
-        proposer = _proposer;
-        challenger = _challenger;
+        systemConfig = _systemConfig;
+
+        // Delete the proposer value which is no longer stored in this contract.
```
tynes, line 110: It feels like we don't have a standard on when to overwrite spacers and when not to

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -38,6 +40,18 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
         address optimismMintableERC20Factory;
     }
 
+    /// @notice Struct representing the overhead, scalar, and gas limit.
```
tynes: This comment is not correct, no gas limit

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -109,15 +129,22 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
...
     /// @notice Constructs the SystemConfig contract. Cannot set
     ///         the owner to `address(0)` due to the Ownable contract's
     ///         implementation, so set it to `address(0xdEaD)`
     constructor() {
         initialize({
             _owner: address(0xdEaD),
             _superchainConfig: address(0),
-            _overhead: 0,
-            _scalar: 0,
+            _gasConfig: SystemConfig.GasConfig({ overhead: 0, scalar: 0 }),
```
tynes: This has the `SystemConfig` prefix but the others do not

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -327,20 +356,18 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
...
-    function _setGasConfig(uint256 _overhead, uint256 _scalar) internal {
-        overhead = _overhead;
-        scalar = _scalar;
+    /// @param _gasConfig New gas config.
+    function _setGasConfig(GasConfig memory _gasConfig) internal {
+        overhead = _gasConfig.overhead;
+        scalar = _gasConfig.scalar;
 
-        bytes memory data = abi.encode(_overhead, _scalar);
+        bytes memory data = abi.encode(_gasConfig);
```
tynes: I have a slight preference for not encoding the struct and instead encoding the struct properties so it looks the same or would like a unit test showing equivalence

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -399,4 +426,45 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
 
         _resourceConfig = _config;
     }
+
+    /// @notice Checks if the given address is a proposal manager.
+    /// @param _manager The address to check.
+    /// @return A boolean indicating if the address is a proposal manager.
+    function isProposalManager(address _manager) public view returns (bool) {
```
tynes: This file should standardize on named returns

packages/contracts-bedrock/test/SystemConfig.t.sol
```diff
@@ -385,3 +385,28 @@ contract SystemConfig_CheckSequencer_TestFail is SystemConfig_Initializer {
         sysConf.checkSequencer();
     }
 }
+
+contract SystemConfig_isProposalManager_Test is SystemConfig_Initializer {
+    /// @dev Tests that `isProposalManager` returns the correct value.
+    function test_isProposalManager_succeeds() external {
+        // check if the challenger can delete an output
```
tynes: This technically doesn't check if the challenger can delete an output, it checks that `isProposalManager` returns true. It may or may not be the case now or in the future that this function must return true to delete an output

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
maurelian: Are there spacers that we haven't removed in the past?

I can remove this, but my inclination is that we should lean towards doing it when we have an opportunity.

packages/contracts-bedrock/test/SystemConfig.t.sol
maurelian: These comments are a bit superflous anyways, I will remove them.

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
tynes: I am fine with leaving it

tynes: Yes there are spacers in the xdm and bridge that we didn't remove before

maurelian: Right, IIRC not removing them was mostly a decision meant to limit scope creep. 
In this can we can easily fit it in with this PR. 

That said, there is also the option of removing the spacer entirely, and setting it to zero with `Storage.setUint(5,0)`. 
I prefer the current implementation with the explicit spacer as it prevents us from making a mistake prior to this getting out the door. 

Going to resolve this, feel free to reopen if you like.

packages/contracts-bedrock/src/L1/SystemConfig.sol
maurelian: I unwrapped the properties in the implementation. 
There is also a test of the initializer which used the encoded struct, showing that they are the same.

packages/contracts-bedrock/src/L1/SystemConfig.sol
maurelian: Good call out. Fixed.

op-chain-ops/genesis/config.go
maurelian: Very much on board with that and #7310 did implement that. 
When we update the deployments, we should also rename the challenger and proposer config vars.

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
```diff
@@ -79,19 +83,17 @@ contract L2OutputOracle is Initializable, ISemver {
...
+    /// @param _systemConfig        The address of the system config.
     function initialize(
         uint256 _startingBlockNumber,
         uint256 _startingTimestamp,
-        address _proposer,
-        address _challenger
+        SystemConfig _systemConfig
     )
         public
         reinitializer(2)
```
tynes, line 99: We will want to make this `reinitializer(3)` at some point in the future but its fine for now

packages/contracts-bedrock/test/L2OutputOracle.t.sol
```diff
@@ -197,6 +197,30 @@ contract L2OutputOracle_getter_Test is L2OutputOracle_Initializer {
...
+    function testFuzz_isProposalManagerCanSetProposerAndDelete_succeeds(address caller) external {
+        proposeAnotherOutput();
+
+        uint256 latestOutputIndex = oracle.latestOutputIndex();
+        address newProposer = makeAddr("New Proposer");
+
+        if (sysConf.isProposalManager(caller)) {
+            vm.prank(caller);
+            sysConf.setProposer(newProposer);
+            oracle.deleteL2Outputs(latestOutputIndex);
```
tynes, line 211: Do you need to prank as `newProposer` before calling `deleteL2Outputs` here? Isn't that supposed to be gated so it should revert?

packages/contracts-bedrock/test/L2OutputOracle.t.sol
```diff
@@ -197,6 +197,30 @@ contract L2OutputOracle_getter_Test is L2OutputOracle_Initializer {
...
 
+contract L2OutputOracle_proposalManagement_Test is L2OutputOracle_Initializer {
+    /// @dev Tests that `isProposalManager` is identically enforced both on deletion and proposer updating
+    function testFuzz_isProposalManagerCanSetProposerAndDelete_succeeds(address caller) external {
+        proposeAnotherOutput();
+
+        uint256 latestOutputIndex = oracle.latestOutputIndex();
+        address newProposer = makeAddr("New Proposer");
+
+        if (sysConf.isProposalManager(caller)) {
```
tynes, line 208: This feels like it should be split into 2 fuzz tests where 1 is always the top branch and 1 is always the bottom branch

packages/contracts-bedrock/test/L2OutputOracle.t.sol
```diff
@@ -393,6 +417,16 @@ contract L2OutputOracle_deleteOutputs_Test is L2OutputOracle_Initializer {
         vm.expectRevert("L2OutputOracle: cannot delete outputs that have already been finalized");
         oracle.deleteL2Outputs(latestOutputIndex);
     }
+
+    /// @dev Tests that `deleteL2Outputs` reverts for finalized outputs.
+    function test_deleteL2Outputs_unAuthed_reverts() external {
+        proposeAnotherOutput();
+        uint256 latestOutputIndex = oracle.latestOutputIndex();
+
```
tynes, line 425: This could be made stronger with as `assertFalse` on `isProposalManager`

packages/contracts-bedrock/test/L2OutputOracle.t.sol
```diff
@@ -416,21 +450,21 @@ contract L2OutputOracleUpgradeable_Test is L2OutputOracle_Initializer {
...
+        assertEq(oracle.l2BlockTime(), l2BlockTime);
 
         // The values that are set in the initialize function should be all
         // zero values in the implementation contract.
         assertEq(oracleImpl.startingBlockNumber(), 0);
         assertEq(oracleImpl.startingTimestamp(), 0);
-        assertEq(oracleImpl.PROPOSER(), address(0));
-        assertEq(oracleImpl.proposer(), address(0));
-        assertEq(oracleImpl.CHALLENGER(), address(0));
-        assertEq(oracleImpl.challenger(), address(0));
```
tynes, line 433: Are we breaking the ABI by removing these getters? Wondering if any of our tooling uses them

packages/contracts-bedrock/test/SystemConfig.t.sol
```diff
@@ -98,7 +98,7 @@ contract SystemConfig_Initialize_Test is SystemConfig_Initializer {
         vm.expectEmit(true, true, true, true, address(sysConf));
         emit ConfigUpdate(0, SystemConfig.UpdateType.BATCHER, abi.encode(batcherHash));
         vm.expectEmit(true, true, true, true, address(sysConf));
-        emit ConfigUpdate(0, SystemConfig.UpdateType.GAS_CONFIG, abi.encode(overhead, scalar));
+        emit ConfigUpdate(0, SystemConfig.UpdateType.GAS_CONFIG, abi.encode(gasConfig));
```
tynes, line 101: If this test passes, then the encoding is equivalent to referencing the properties, i thought it would be the same but its one of those things that could create a bug in some unexpected way

tynes (approved):

packages/contracts-bedrock/test/SystemConfig.t.sol
maurelian: Yeah, agree. I didn't really want to have a stand alone test sitting around that verifies it though. I can unwrap this event too lmk.

packages/contracts-bedrock/test/L2OutputOracle.t.sol
maurelian: Will do

packages/contracts-bedrock/test/L2OutputOracle.t.sol
maurelian: I think one is not a fuzz test though, and the fuzz test going to need to vm.assume and throw away these addresses.

packages/contracts-bedrock/src/L1/L2OutputOracle.sol
maurelian: Hmmm, is there any reason _not_ to make it `3` now?

packages/contracts-bedrock/test/L2OutputOracle.t.sol
maurelian: No, the Proposer is different from the Challenger. Pranking as the proposer should cause a revert.
This test is just meant to indicate that the abilities to
1. delete outputs, and 
2. update the Proposer address,
are shared by the same set of actors.

I'm starting to think that `isProposalManger()` isn't the best name, but I can't think of a better one. Nor can chatGPT.

But I'm adding this comment to that function to hopefully clarify:
```
///         The same entities who can delete an output should also be able to update the proposer; because in the
///         event that a faulty output is proposed, the malicious proposer will need to be removed prior to
///         deleting the output.
```

packages/contracts-bedrock/test/L2OutputOracle.t.sol
maurelian: I thought about it, and I really like this test as is. It clearly expresses the idea that there are two types of accounts, those that can both delete outputs AND change the proposer, and those that can do neither. There is necessarily a one to one mapping between them.

maurelian: will change the suffix on the test name from `_succeeds` to `_works` though to allow for both the happy and sad cases in one.

packages/contracts-bedrock/test/L2OutputOracle.t.sol
maurelian: I haven't removed the getters, they are still tested on the proxy on lines 415-418. But they fail on the implementation since it depends on a call to the `systemConfig` address which is zero on the impl.

PR #7415, opened 2023-09-27
allow release pipeline for hearbeat service

InoMurko:
**Description**

I could not find a way to properly release op-heartbeat through the standard route (gh actions tag that spawns a Cirleci build).
**Tests**

We're using this in Bobanetwork as of yesterday.

**Additional context**

/

**Metadata**

/

### DIFF
--- a/ops/tag-service/tag-service.py
+++ b/ops/tag-service/tag-service.py
@@ -17,7 +17,8 @@ MIN_VERSIONS = {
     'op-challenger': '0.0.4',
     'op-proposer': '0.10.14',
     'op-ufm': '0.1.0',
-    'proxyd': '3.16.0'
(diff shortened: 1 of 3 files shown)


ops/tag-service/tag-service.py
```diff
@@ -17,7 +17,8 @@
     'op-challenger': '0.0.4',
     'op-proposer': '0.10.14',
     'op-ufm': '0.1.0',
-    'proxyd': '3.16.0'
+    'proxyd': '3.16.0',
+    'op-heartbeat': '0.1.0'
```
InoMurko, line 21: not sure if this is right!

tynes:
I believe this looks right

zhwrd (approved):

PR #7416, opened 2023-09-27
feat(specs): Improve Subgame Rez explanation

clabby:
## Overview

Replaces the complicated digraph with an animation to iteratively walk through an example of the sub-game resolution process.


Inphi (commented):
Damn son. That's a cool ass animation! Loving it. Would also add a quick note explaining what the colors represent in the animation
Could you also keep the static image in there. As it's easiest to reference quickly.

Inphi (approved):

PR #7428, opened 2023-09-27
feat(ctb): Move storage getters and setters to own lib

maurelian:
**Description**

Moves the various low lever getters and setters which write directly to storage into a new lib `Slot.sol`

**Tests**

Each getter/setter pair is tested to ensure that they set/get the same value, and that the value corresponds with the test case input.


tynes:
I wouldn't be opposed to renaming `Slot` to `Storage` but def a good change

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
```diff
@@ -34,8 +35,8 @@ contract ProtocolVersions is OwnableUpgradeable, ISemver {
     event ConfigUpdate(uint256 indexed version, UpdateType indexed updateType, bytes data);
 
     /// @notice Semantic version.
-    /// @custom:semver 0.1.0
-    string public constant version = "0.1.0";
+    /// @custom:semver 0.1.1
+    string public constant version = "0.1.1";
```
tynes: Looks like we need to regen bindings

packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -99,8 +100,8 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
     uint256 public startBlock;
 
     /// @notice Semantic version.
-    /// @custom:semver 1.7.1
-    string public constant version = "1.7.1";
+    /// @custom:semver 1.7.2
+    string public constant version = "1.7.2";
```
tynes: Both semver changes should be minor and not patch because the changes impact the bytecode and not just the source code

tynes (changes requested):
Will approve when the semver is updated to minor bumps

tynes (approved):

PR #7433, opened 2023-09-27
update 4844 code to be compatible with latest beacon node api

roberto-bayardo:
**Description**

- Updates the code to support the latest [4844 beacon node api](https://ethereum.github.io/beacon-APIs/#/Beacon/getBlobSidecars); for example we can now fetch individual blobs by index rather than having to fetch all blobs in the block.

- Change blob verification to verify against the proof rather than recomputing the commitment.

**Tests**

Confirmed e2e 4844 test still passes.

### DIFF
--- a/op-e2e/e2eutils/fakebeacon/blobs.go
+++ b/op-e2e/e2eutils/fakebeacon/blobs.go
@@ -67,11 +67,11 @@ func (f *FakeBeacon) Start(addr string) error {
 			f.log.Error("config handler err", "err", err)
 		}
 	})
-	mux.HandleFunc("/eth/v1/beacon/blobs_sidecars/", func(w http.ResponseWriter, r *http.Request) {
-		blockID := strings.TrimPrefix(r.URL.Path, "/eth/v1/beacon/blobs_sidecars/")
+	mux.HandleFunc("/eth/v1/beacon/blob_sidecars/", func(w http.ResponseWriter, r *http.Request) {
+		blockID := strings.TrimPrefix(r.URL.Path, "/eth/v1/beacon/blob_sidecars/")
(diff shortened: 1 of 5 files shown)


roberto-bayardo:
cc @protolambda

op-e2e/e2eutils/fakebeacon/blobs.go
```diff
@@ -81,19 +81,47 @@ func (f *FakeBeacon) Start(addr string) error {
...
+		rawIndices := query["indices"]
+		indices := make([]int, 0, len(bundle.Blobs))
+		if len(rawIndices) == 0 {
+			// request is for all blobs
+			for i := range bundle.Blobs {
+				indices = append(indices, i)
+			}
+		} else {
+			for _, raw := range rawIndices {
+				ix, err := strconv.ParseUint(raw, 10, 64)
```
refcell, line 95: Since `ParseUint` will error if the raw index is sign-prefixed, the latter check that the index is negative should never be hit on line 110. Could we remove that check and just make the `indices` array of type `uint64`?

roberto-bayardo: That's certainly a reasonable approach. If a value's purpose is known to be an array index I prefer to convert to int at that point rather than deferring the conversion, but then as you note you have the awkward possibility of negative values that I always feel obliged to defensively check for (as done on L110).   I don't feel strongly about either approach TBH.

protolambda: The `ix < 0` should be fine, and it's only a testing codepath, not too critical

protolambda (approved):
Nice changes :+1:

PR #7434, opened 2023-09-27
op-challenger: Include game address in error message instead of listing all games

ajsutton:
**Description**

When there's an error scheduling a game update, the coordinator continues trying to schedule all games, but was then reporting the single game error with a log message that listed all games.  Change this to include the game address in the specific error message and not attach the full list of games.

**Metadata**

- Fixes https://github.com/ethereum-optimism/client-pod/issues/55


refcell (approved):
This is very nice

PR #7435, opened 2023-09-27
Add Base sepolia constants

mdehoog:
**Description**
Base sepolia has been genesis'd. The configuration can be verified in https://github.com/base-org/contract-deployments/edit/main/sepolia/2023-09-26-deploy/deployed/addresses.json and https://github.com/base-org/node/blob/main/sepolia/rollup.json.


packages/sdk/src/utils/chain-constants.ts
```diff
@@ -83,8 +83,9 @@ export const DEPOSIT_CONFIRMATION_BLOCKS: {
   [L2ChainID.OPTIMISM_HARDHAT_LOCAL]: 2 as const,
   [L2ChainID.OPTIMISM_HARDHAT_DEVNET]: 2 as const,
   [L2ChainID.OPTIMISM_BEDROCK_ALPHA_TESTNET]: 12 as const,
-  [L2ChainID.BASE_GOERLI]: 12 as const,
-  [L2ChainID.BASE_MAINNET]: 50 as const,
```
mdehoog, line 87: these are incorrect, we switched goerli to a 25 deposit conf time to avoid full node halts when goerli has deep reorgs

mainnet was decreased to 10 for user experience (faster deposits)

refcell (approved):

tynes:
Thank you for opening this PR! Ideally we can get to a single source of truth for these sorts of constants, we have been trying to consolidate in https://github.com/ethereum-optimism/superchain-registry but its not ready for typescript consumption yet

roninjin10 (approved):

op-node/rollup/superchain.go
```diff
@@ -19,6 +19,7 @@ const (
 	opGoerli    = 420
 	opSepolia   = 11155420
 	baseGoerli  = 84531
+	baseSepolia = 84532
```
protolambda: The other unexported constants here are for a few chain-specific exceptions, for configuration edge-cases that are not supported in the superchain-registry repository. I believe we can omit any unused constants here, and just register the chains in the registry.

protolambda (approved):
Changes look good to me, but we may want to update the superchain registry as well. The long-term ideal is for all monorepo components, and any other op-stack component, to pull the preset chain definitions from a single source of truth. See the `CONTRIBUTING.md` in the superchain-registry repo to get started.

op-node/rollup/superchain.go
mdehoog: happy to remove this and the other unused constants if preferred

trianglesphere:
```suggestion
```

refcell:
Can you rebase on develop? There were some e2e fixes that were might that should fix the WS flakes

PR #7437, opened 2023-09-28
feat: Multiple Mergify Queues

refcell:
**Description**

Updates the mergify ruleset to use 3 queues as opposed to 1.

The three queues (in decreasing priority) are: `urgent`, `default`, and `lowpriority`.

By default, pull requests are added to the `default` queue. Pull requests made by `dependabot` will be added to the `lowpriority` queue. When the https://github.com/ethereum-optimism/optimism/labels/C-urgent label is added to a pull request, it is added to the `urgent` queue.

In sum, the processing of queues is done in the order they are defined. So the `urgent` queue is processed before the
`default` queue, which in turn is processed before the `lowpriority` queue.

**On Queue Interruption**

By introducing multiple queues, interruption is supported whereby a pull request added to the `default` queue will _interrupt_
pull requests in the `lowpriority` queue, effectively "jumping" ahead and will be merged before `lowpriority` prs.

Better explained by the mergify docs:

> By default, when a pull request enters a queue that has a higher precedence over the one currently being processed, the processing of the current queue gets interrupted, and the higher precedence queue takes over. This ensures that your most critical pull requests are processed as soon as possible, regardless of what else is happening.

To learn more about merge queues, see the [mergify docs](https://docs.mergify.com/merge-queue/multi).

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,7 +1,7 @@
 queue_rules:
   - name: urgent
     queue_conditions:
-      - label = urgent
+      - label = C-urgent
   - name: default
     conditions: []
   - name: lowpriority
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,6 +1,12 @@
 queue_rules:
+  - name: urgent
+    queue_conditions:
+      - label = urgent
   - name: default
     conditions: []
+  - name: lowpriority
+    conditions:
+      - author=dependabot[bot]
 
 pull_request_rules:
   - name: Automatic merge on approval
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -3,9 +3,9 @@ queue_rules:
     queue_conditions:
       - label = C-urgent
   - name: default
-    conditions: []
+    queue_conditions: []
   - name: lowpriority
-    conditions:
+    queue_conditions:
       - author=dependabot[bot]
 
 pull_request_rules:


refcell:
* develop
  * **PR #7437** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7437" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7437?utm_source=stack-comment).

refcell:
This is a sizable devx change, so throwing the https://github.com/ethereum-optimism/optimism/labels/M-mergify-ignore label on here for now. Strong critiques + suggestions appreciated.

tynes:
I am a big fan of this change, thank you @refcell. I think we should try this out and see if it solves the problems for us

tynes (approved):

.github/mergify.yml
```diff
@@ -1,6 +1,12 @@
 queue_rules:
+  - name: urgent
+    queue_conditions:
```
sebastianst, line 3: what's the difference between the `queue_conditions` and `conditions` property?

sebastianst (approved):
Love it

.github/mergify.yml
refcell: Ah good catch, it looks like [`queue_conditions`](https://docs.mergify.com/configuration/file-format#:~:text=To%20validate%20your%20Mergify%20configuration,tab%20of%20your%20pull%20request.) is the canonical field here. My best guess is that `conditions` is a valid backwards-compatible field that mergify still supports hence why the bot did not freak out about the config being invalid. Updated to use `queue_conditions` now.

refcell: I double-checked that this config is valid in the editor fwiw :)
Screenshot 2023-09-28 at 8 11 02 PM

roninjin10 (approved):
Cool!

.github/mergify.yml
roninjin10: I think conditions is more general. you can do things with mergify such as make a comment under certain conditions or rebase a pr based on certain condtitions

PR #7438, opened 2023-09-28
op-e2e: Increase gas fee cap

ajsutton:
**Description**

Increase MaxFeeCap in TestBigL2Txs test to try to fix flakiness.


tynes (approved):
This didn't seem to fix the flakiness in practice but we could still include the change

ajsutton:
Going to close this - got a better understanding of the source of intermittency now and I don't think this is required. https://github.com/ethereum-optimism/optimism/pull/7442 is the better fix.

PR #7440, opened 2023-09-28
op-program: Add compatibility test to verify a section of goerli in CI

ajsutton:
**Description**

Adds the ability to capture pre-images from a run of `op-program` verifying a section of goerli along with the required arguments to re-run that same op-program invocation in offline mode. Uncompressed that's nearly 50,000 files that take up about 200Mb, so it's compressed into a tar.bz file that uses 34Mb. That data is now published as a release artifact in https://github.com/ethereum-optimism/chain-test-data/.  Creating a new tag in that repo will rerun the generation and upload the data to a new release automatically.

A new CI target is then added that expands that tarball and re-runs `op-program` to confirm it still produces the same result. Expanding and running op-program takes about 20s locally so is plenty fast enough to run in CI.

In `op-program` new test data can be captured with:
```
L1URL=$GOERLI_ALCHEMY L2URL=http://localhost:8544 make capture-goerli-verify
```

It takes a while, but only needs to be done when we need to regenerate the data for some reason. The resulting 34Mb file is then checked in.

Re-running op-program can be done with:
```
make run-goerli-verify
```

It also changes how the goerli verification selects inputs to find two adjacent proposed output roots that are present in the L1 finalized block. This is simpler and matches how the dispute game actually selects inputs.


.circleci/config.yml
```diff
@@ -1068,6 +1068,17 @@ jobs:
...
+  op-program-compat:
+    docker:
+      - image: cimg/go:1.21
+    steps:
+      - checkout
+      - run:
+          name: compat-goerli
+          command: |
+            make run-goerli-verify
+          working_directorY: op-program
```
sebastianst:
```suggestion
          working_directory: op-program
```

protolambda (changes requested):
To maintain a full copy of the git history, don't we need every tarball? 35 MB once might be fine, but repeating a few changes might quickly accumulate. Ideally we put it in a google cloud storage bucket or somewhere external, and use CI artifact caching to keep a copy closer to circleci.

op-program/Makefile
```diff
@@ -25,7 +25,7 @@ op-program-client-mips:
 	# result is mips32, big endian, R3000
 
 clean:
-	rm -rf bin
+	rm -rf bin temp/compat
```
tynes: This could be defined as a top level variable

tynes:
Another option is hosting the tarball in a cloud bucket. The concern would be cost of serving that data

ajsutton:
> To maintain a full copy of the git history, don't we need every tarball? 35 MB once might be fine, but repeating a few changes might quickly accumulate. Ideally we put it in a google cloud storage bucket or somewhere external, and use CI artifact caching to keep a copy closer to circleci.

It shouldn't need to change often, but yeah that is a concern.  We could potentially just use git lfs.  I never like including external dependencies in required CI jobs if we can avoid it - we do want the file to be versioned with git so tests keep working for old versions.  Doable but will require some more thought.

sebastianst:
> It shouldn't need to change often, but yeah that is a concern. We could potentially just use git lfs. I never like including external dependencies in required CI jobs if we can avoid it - we do want the file to be versioned with git so tests keep working for old versions. Doable but will require some more thought.

Also a fan of git lfs 👍🏻

ajsutton:
Ok, I've switched this to use git-lfs which is pretty nice in terms of just working with git but two downsides I can see:

1. Anyone who doesn't have git lfs installed will now get an error when working with the repo saying that git lfs isn't installed. It's a pretty helpful message but definitely a noticeable change.
2. I have seen a number of cases where the GitHub lfs quota gets hit and then things start breaking which would be quite disruptive.

One possible option is to not install the usual git lfs hooks and just have the make target run the appropriate git-lfs command to download the latest version of the file before running.  Committing a new version of the tarball would be more difficult though - the makefile currently just produces the tarball, so you'd have to run the right sequence of add, upload, commit type commands yourself.

Or we look at other options that aren't as integrated into git (e.g download from a server but reference the file by a hash and keep older versions around).

Or we put these in a different repo and the script just clones the appropriate version using a shallow checkout. Could potentially use submodules if we're prepared to switch target repos if the history starts getting too large...

Inphi:
Not a fan of git lfs for this. The tarball isn't terribly big to warrant it, and introducing git-lfs complicates git cloning for users.

trianglesphere:
Interesting: Github has really low quotas for LFS: https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage

I this seems like a good candidate for git LFS, but I haven't actually worked with git lfs, so I don't know the pitfalls.

protolambda:
> [Storage and bandwidth quotas](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas)
>
> Each file included in a release must be under 2 GiB. There is no limit on the total size of a release, nor bandwidth usage.

I think ideally we set up a separate public repository, which generates the data in a CI job, and then automatically publishes it as release with attached data. It should be in a separate repository, since I don't want to bloat the releases page of the monorepo.

We can then pull in the test data as `tar.gz` or something by release tag, and host it for free, without git LFS quota.

I think go-ethereum also does something like this when you run their `build/ci.go` and it pulls in testdata dynamically.

cc @ajsutton

ajsutton:
Yeah agreed on hosting this elsewhere - I just haven't had a chance to revisit this.  Interesting idea to have CI generate it.  I'm thinking:

* Create a new repo for this. Only thing in the repo would be the CI configuration.
* CI would be triggered to build and publish a release when a tag is created in the repo (thus we can just tag the repo whenever we want to regenerate the data).
* Actual generation scripts probably still lives in the monorepo. I'm tempted to just have CI always checkout the latest monorepo to keep things simple - I don't see any reason we need to be able to rebuild old artefacts and that avoids us needing to manually update the pinned version of the monorepo.
  * We need op-program to generate the pre-images so we'll always need to pull in the monorepo and so may as well have all the generation scripts there.
* Publishes the generated images as binaries attached to a new GitHub release.

In the monorepo CI we'd then have the test point to a the specific binaries.  So a "rebuild" of the pre-image data would mean tagging the external repo, wait for it to publish the release and then updating the monorepo to point to that new release.

Seems like a sensible approach to me.

ajsutton:
Made a start on this in https://github.com/ethereum-optimism/chain-test-data but need to work out how to get a GitHub access token that lets it create releases and upload artefacts.  Once that's working I'll tidy it up a bit, add a README with instructions etc and can then remove the .tar.bz from this PR (will do a rewrite of history to remove it completely).

ajsutton:
Ok, the automated data capture and release publishing is now working in https://github.com/ethereum-optimism/chain-test-data and this just downloads data from there.

.circleci/config.yml
```diff
@@ -1254,6 +1254,17 @@ jobs:
...
+  op-program-compat:
+    docker:
+      - image: cimg/go:1.21
+    steps:
+      - checkout
+      - run:
+          name: compat-goerli
+          command: |
+            make run-goerli-verify
+          working_directory: op-program
```
protolambda:
```suggestion
  op-program-compat:
    docker:
      - image: us-docker.pkg.dev/oplabs-tools-artifacts/images/ci-builder:latest
    steps:
      - checkout
      - restore_cache:
          name: Restore Go modules cache
          key: gomod-{{ checksum "go.sum" }}
      - restore_cache:
          key: golang-build-cache
      - run:
          name: compat-goerli
          command: |
            make run-goerli-verify
          working_directory: op-program
```

To reuse the Go build/mod caches.

protolambda (approved):
Very nice to have this running in CI. Will apply one small change to the CI job definition to use the CI-builder image and load the Go caches, but LGTM otherwise.

zhwrd (approved):

zhwrd (approved):

PR #7442, opened 2023-09-28
op-e2e: Wait for transactions to be ready to include in ActL2IncludeTx

ajsutton:
**Description**

Theory on the source of flakiness:
The test adds the transaction to the pool by calling `eth_sendRawTransaction` which winds up in `[api_backend` `SendTx`](https://github.com/ethereum-optimism/op-geth/blob/c06ca1374da9aafee1f2ff2ab44816b4cf78b11b/eth/api_backend.go#L289-L311). Importantly when it calls `txPool.Add` it specifies `sync: false`. The net result is that the tx is added to the pool’s `queue` initially (the unprocessable transactions) and then [calls `requestPromoteExecutables` to schedule an evaluation of which transactions can be promoted asynchronously](https://github.com/ethereum-optimism/op-geth/blob/8737298ff823f5a8397ac6533a8a227a4662b5d9/core/txpool/legacypool/legacypool.go#L1026-L1030).  Because `sync` is false, it does *not* wait for that promotion to complete.

If the test proceeds and reaches the point in `ActL2IncludeTx` where it retrieves transactions before the promotion completes, it will get the new transaction in the unprocessable list and thus fail.

This avoids that async process by calling `Add` on the tx pool directly and setting `sync=true` to ensure any promotion has completed before the method returns.

### DIFF
--- a/op-e2e/actions/l2_engine.go
+++ b/op-e2e/actions/l2_engine.go
@@ -171,13 +171,6 @@ func (e *L2Engine) ActL2RPCFail(t Testing) {
 	e.failL2RPC = errors.New("mock L2 RPC error")
 }
 
-// ActSendTx adds a transaction to the tx pool and ensures any required promotion to the pending tx queue is complete
-// before returning.
-func (e *L2Engine) ActSendTx(t Testing, tx *types.Transaction) {
-	err := errors.Join(e.eth.TxPool().Add([]*types.Transaction{tx}, true, true)...)
-	require.NoError(t, err, "Add tx to pool")
-}
-
 // ActL2IncludeTx includes the next transaction from the given address in the block that is being built
 func (e *L2Engine) ActL2IncludeTx(from common.Address) Action {
 	return func(t Testing) {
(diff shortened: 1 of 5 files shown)


protolambda:
See #7455 (PR to fix develop issue, may need to rebase once merged) and #7448 (extension of this) to reduce test flakes more

protolambda (approved):

refcell (approved):

op-e2e/actions/l2_engine.go
```diff
@@ -176,12 +186,22 @@ func (e *L2Engine) ActL2IncludeTx(from common.Address) Action {
...
+			i := e.engineApi.PendingIndices(from)
+			txs, _ := e.eth.TxPool().ContentFrom(from)
+			return uint64(len(txs)) > i, nil
+		})
+		require.NoError(t, err, "wait for tx to be in pending tx queue")
+
 		i := e.engineApi.PendingIndices(from)
 		txs, q := e.eth.TxPool().ContentFrom(from)
 		require.Greaterf(t, uint64(len(txs)), i,
 			"no pending txs from %s, and have %d unprocessable queued txs from this account", from, len(q))
```
ajsutton: We should combine the waiting and "getting" of values here so that when the waiter returns we use the values it just validated match.  It _shouldn't_ matter but this code can have the tx pool change again between when the waiter completes and the actual get which causes the test to fail again.

op-e2e/system_adminrpc_test.go
```diff
@@ -66,10 +68,11 @@ func TestStopStartSequencer(t *testing.T) {
 	require.NoError(t, err)
 	require.True(t, active, "sequencer should be active again")
 
-	blockBefore = latestBlock(t, l2Seq)
-	time.Sleep(time.Duration(cfg.DeployConfig.L2BlockTime+1) * time.Second)
-	blockAfter = latestBlock(t, l2Seq)
-	require.Greater(t, blockAfter, blockBefore, "Chain did not advance after starting sequencer")
+	require.NoError(
+		t,
+		wait.ForNextBlockWithTimeout(ctx, l2Seq, time.Duration(cfg.DeployConfig.L2BlockTime+1)*2*time.Second),
```
ajsutton: I think to reduce intermittency we probably need to increase the length of the timeout here

op-e2e/system_adminrpc_test.go
```diff
@@ -50,9 +52,9 @@ func TestStopStartSequencer(t *testing.T) {
 	require.NoError(t, err)
 	require.False(t, active, "sequencer should be inactive")
 
-	blockBefore = latestBlock(t, l2Seq)
+	blockBefore := latestBlock(t, l2Seq)
 	time.Sleep(time.Duration(cfg.DeployConfig.L2BlockTime+1) * time.Second)
-	blockAfter = latestBlock(t, l2Seq)
+	blockAfter := latestBlock(t, l2Seq)
```
ajsutton, line 57: Seems like we should use the waiter here too.

op-e2e/actions/l2_engine.go
```diff
@@ -168,6 +171,13 @@ func (e *L2Engine) ActL2RPCFail(t Testing) {
 	e.failL2RPC = errors.New("mock L2 RPC error")
 }
 
+// ActSendTx adds a transaction to the tx pool and ensures any required promotion to the pending tx queue is complete
+// before returning.
+func (e *L2Engine) ActSendTx(t Testing, tx *types.Transaction) {
+	err := errors.Join(e.eth.TxPool().Add([]*types.Transaction{tx}, true, true)...)
+	require.NoError(t, err, "Add tx to pool")
+}
```
ajsutton: Given we have the waiter in place I suspect we shouldn't do this now and should just stick to using the actual JSON RPC api.

op-e2e/system_adminrpc_test.go
ajsutton: Ah this is expecting to not advance so not suitable for the waiter.

op-e2e/system_adminrpc_test.go
ajsutton: Wound up simplifying things a lot and just modifying `ForNextBlock` to always apply a 60 second timeout.  That will be long enough to cover any sensible block time and means we have a timeout in place for all tests to avoid them hanging.  Test code is then kept really simple and declarative with the details being handled by the helper code.

PR #7448, opened 2023-09-28
fix: Wait for Tx Inclusion in Pool

refcell:
**Description**

The synchronous method of tx inclusion ensures tx processing is done in a timely manner, but multiple tests don't have direct access to the internal tx pool.

This pr updates the `ActL2IncludeTx` method to wait for the tx to be included in the pool with a timeout, thus preventing tests like `TestCrossLayerUser/NoRegolith` from flaking since they have to send the tx directly through a client.

### DIFF
--- a/op-e2e/actions/l2_engine.go
+++ b/op-e2e/actions/l2_engine.go
@@ -1,9 +1,12 @@
 package actions
 
 import (
+	"context"
 	"errors"
+	"time"
 
 	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils"
+	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils/wait"
 	"github.com/ethereum-optimism/optimism/op-program/client/l2/engineapi"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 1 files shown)


op-e2e/actions/l2_engine.go
```diff
@@ -183,12 +186,22 @@ func (e *L2Engine) ActL2IncludeTx(from common.Address) Action {
...
+		// Wait for the tx to be in the pending tx queue
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		err := wait.For(ctx, 2*time.Second, func() (bool, error) {
+			i := e.engineApi.PendingIndices(from)
+			txs, _ := e.eth.TxPool().ContentFrom(from)
+			return uint64(len(txs)) > i, nil
+		})
+		require.NoError(t, err, "wait for tx to be in pending tx queue")
+
```
protolambda, line 198: I'm not sure where we are inserting txs into the tx-pool asynchronously, but with a line like the following we can enforce synchronous execution in the tx-pool, and don't need the wait here.
```go
require.NoError(t, e.eth.TxPool().Add([]*types.Transaction{tx}, true, true)[0])
```

Maybe it's possible we change the `eth_sendRawTransaction` calls to the pool, and make these synchronous?
That could also really help op-e2e stability.

Geth implements the backend behind this RPC like:
```
func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction) error {
	return b.eth.txPool.Add([]*types.Transaction{signedTx}, true, false)[0]
```
I.e. it's async (`sync` bool argument is `false` here)

So I think this change in op-e2e becomes unnecessary, if we add a configuration option to op-geth to make tx-insertion synchronous, which we can then enable in tests.

protolambda (commented):
It's an ok workaround to wait for nonces, but I think we can make all of the testing more stable by adding a config option to op-geth instead.

op-e2e/actions/l2_engine.go
protolambda: Summary from slack thread: we can do something like this without introducing op-geth options by introducing a proxy around the API, to intercept the tx-sending, and wait for any events (pending nonce update etc.) as necessary, making all tests more stable. We can merge this as-is for op-e2e/actions to get more stable.

protolambda (approved):
LGTM. Introducing the wait call will avoid the race condition where the pool hasn't processed the previously submitted tx yet, but it's still possible that CI lag causes the wait to fail. When we introduce the proxy thing (see comment and slack threaD), to block right after tx-submission and wait for the txpool state change, we fix that.

refcell:
Ok let's see what's flaking now 😅

Inphi (approved):

PR #7449, opened 2023-09-28
fix: add protocol version to getting-started.json

Chomtana:
**Description**

getting-started.json is missing `requiredProtocolVersion` and `recommendedProtocolVersion`

.env is missing `IMPL_SALT` which is a random bytes32 hex string used for the create2 salt to enable deterministic deployment across chains.

This causes deploying contract errors on v1.1.6-rc2 if following the op stack document available at https://stack.optimism.io/docs/build/getting-started/#configure-your-network

**Tests**

I have a manual test on v1.1.6-rc2

**Before**

image

**After**

image

You can see that this error is fixed and contracts are deployed successfully.

**Additional context**

Fix for optimism v1.1.6-rc2

image

### DIFF
--- a/packages/contracts-bedrock/deploy-config/getting-started.json
+++ b/packages/contracts-bedrock/deploy-config/getting-started.json
@@ -53,5 +53,8 @@
   "eip1559Denominator": 50,
   "eip1559Elasticity": 10,
 
-  "systemConfigStartBlock": 0
+  "systemConfigStartBlock": 0,
+
+  "requiredProtocolVersion": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "recommendedProtocolVersion": "0x0000000000000000000000000000000000000000000000000000000000000000"
 }


Chomtana:
@tynes @sbvegan let's review this if it's necessary to have these field in order to deploy the contract

Chomtana:
Hi @tynes @smartcontracts @sbvegan , I have added proof of working deployment and a screenshot of the problem on v1.1.6-rc.2

It's great if you can help review and merge my pull request.

Thank you,

Inphi (commented):
The protocol version configs changes make sense, thanks.
IMPL_SALT is an optional config and defaults to "ethers phoenix". We can omit that from the example envrc to keep it minimal.

packages/contracts-bedrock/.envrc.example
```diff
@@ -1,9 +1,12 @@
 # RPC for the network to deploy to
 export ETH_RPC_URL=
 
-# Sets the deployer's key to match the first default hardhat account
+# Set the deployer's key to match the first default hardhat account
 export PRIVATE_KEY=ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
 
+# Set IMPL_SALT to a random bytes32 hex string
+export IMPL_SALT=0xdb6d...random.bytes32...5303
```
clabby: Would leave this commented out and note that this is an optional value with a default in the deploy script.

clabby (changes requested):
Nice, thanks for adding this! Just one nit

packages/contracts-bedrock/.envrc.example
refcell: For further context, see https://github.com/ethereum-optimism/optimism/blob/0b492be02342a7d32c09f114a3a976f961894511/packages/contracts-bedrock/scripts/Deploy.s.sol#L100

Chomtana: Hi @clabby @refcell ,

Deploying with "ethers phoenix" salt is resulting in an error in getting-started deployment because it's trying to redeploy into existing contract addresses. Same as using the same salt across projects.

Screenshot 2023-09-29 012041

I think it should default to a random string instead of "ethers phoenix"?

tynes: We can temporarily set it to a random value, we are in the process of migrating to a setup where all superchains use the same implementations, so the guide will need to be updated

Chomtana: Just seen that it's keccak256(bytes(...)) so IMPL_SALT can be any string not just bytes32

Chomtana: I think using a combination of finalSystemOwner, l1StartingBlockTag and l2ChainID as salt.

As same op stack chain is likely to have the same combination and if different chain one will be differ.

Using random salt will make the address change every time we run the deployment script which is not good and may cause more bugs.

Chomtana:
Hi @clabby @refcell @tynes @sbvegan , I have addressed IMPL_SALT issue by having it default to `keccak256(abi.encode(cfg.finalSystemOwner(), cfg.l1StartingBlockTag(), cfg.l2ChainID()))`

Idea behind three parameters 
- User may need to change admin (and others) account and redeploy
- If user want to deploy a new chain with same account they will need to change l1StartingBlockTag
- User may need to change l2ChainID and redeploy with the rest stay the same
- l1StartingBlockTag is unpredictable, so it's impossible to ddos attack.

However, I have encountered a new unknown bug #7484 that I have no idea how to solve it.

image

packages/contracts-bedrock/.envrc.example
Chomtana: Currently set IMPL_SALT default to `keccak256(abi.encode(cfg.finalSystemOwner(), cfg.l1StartingBlockTag(), cfg.l2ChainID()))` and have set IMPL_SALT requirement to optional.

tynes:
Can you update this PR to just include the changes to `getting-started.json` and then in combination with https://github.com/ethereum-optimism/optimism/pull/7492 it will fix the problems here. Thank you!

Chomtana:
Yes, will do it now

Chomtana:
@tynes Updated this PR to just include the changes to `getting-started.json`. Thank you!

tynes (approved):

tynes:
I will merge this if you can squash this PR into a single commit

rowlanja:
Hi can we merge this

Chomtana:
@tynes squashed this PR into a single commit

PR #7456, opened 2023-09-28
txmgr: Allow CLI Default values to be configured

trianglesphere:
**Description**

This lets the challenger change it's default CLI configs. The challenger has different needs from the batcher. Specifically, retries in transactions from the challenger are not as consequential or expensive as duplicating txns in the batcher. The challenger also has very good internal retries.

**Metadata**

- Fixes https://github.com/ethereum-optimism/client-pod/issues/61


op-service/txmgr/cli.go
```diff
@@ -48,17 +48,42 @@ var (
...
+	DefaultBatcherFlagValues = DefaultFlagValues{
+		NumConfirmations:          uint64(10),
+		SafeAbortNonceTooLowCount: uint64(3),
+		ResubmissionTimeout:       48 * time.Second,
+		NetworkTimeout:            10 * time.Second,
+		TxSendTimeout:             0 * time.Second,
+		TxNotInMempoolTimeout:     2 * time.Minute,
+		ReceiptQueryInterval:      12 * time.Second,
+	}
+	DefaultChallengerFlagValues = DefaultFlagValues{
```
trianglesphere, line 71: cc @ajsutton for the new values

ajsutton (approved):
LGTM but you'll need to fix the tests. I suspect the new defaults just need to be used in challengers config.NewConfig

trianglesphere:
Tests should be fixed now

ajsutton:
Just pulling this out of the merge queue briefly to get the flaky test fix in.  Then will pull that in here and the tests should pass...

PR #7469, opened 2023-09-29
op-e2e: Verify included tx were successful in TestERC20BridgeDeposits

ajsutton:
**Description**

Verify included tx were successful in TestERC20BridgeDeposits. Previously it only verified they were included which doesn't give as much info if the test fails.

Pads the gas estimate when sending the deposit transaction which is the most likely reason that the tx would fail.

Also applies a timeout in `wait.ForReceipt` so tests don't hang forever without tests needing to manually setup their own context with timeout.  The timeout is deliberately very long since it should only be reached in a failing case. If the test supplied a ctx with a shorter timeout that will still apply.

### DIFF
--- a/op-e2e/bridge_test.go
+++ b/op-e2e/bridge_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
 	"github.com/ethereum-optimism/optimism/op-bindings/predeploys"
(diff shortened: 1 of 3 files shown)


tynes (approved):

refcell:
Flag: testing the merge queues by adding the https://github.com/ethereum-optimism/optimism/labels/C-urgent label

PR #7471, opened 2023-09-29
specs: Honest challenger for new resolution

Inphi:
Update the honest challenger specs for the new resolution procedure.

Fixes https://github.com/ethereum-optimism/client-pod/issues/78

### DIFF
--- a/specs/honest-challenger-fdg.md
+++ b/specs/honest-challenger-fdg.md
@@ -67,8 +67,8 @@ This means an honest challenger only responds to claims made by the opposing tea
 The next step is to determine if the claim, now known to be for the opposing team,
 disputes another claim the honest challenger _agrees_ with.
 An honest challenger agrees with a claim iff every other claim along its path to the
-root claim commits to a valid `ClaimHash`. Thus, an honest challenger avoids supporting
-invalid claims on the same team.
+root claim commits to a valid `ClaimHash`. Put differently, an honest challenger will
+avoid countering a claim if it disagrees with the path of claims leading to that specific claim.
 
 The last step is to determine whether the claim has a valid commitment (i.e. `ClaimHash`).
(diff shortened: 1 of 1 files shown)


specs/honest-challenger-fdg.md
```diff
@@ -64,7 +64,13 @@ then it must dispute claims positioned at odd depths in the game tree.
 Otherwise, it disputes claims positioned at even depths in the game tree.
 This means an honest challenger only responds to claims made by the opposing team.
 
-The next step is to determine whether the claim has a valid commitment (i.e. `ClaimHash`).
+The next step is to determine if the claim, now known to be for the opposing team,
+disputes another claim the honest challenger _agrees_ with.
+An honest challenger agrees with a claim iff every other claim along its path to the
+root claim commits to a valid `ClaimHash`. Thus, an honest challenger avoids supporting
```
ajsutton: I'm unclear on the meaning of this. Our challenger only checks the claim's parent, not every claim along its path to the root claim.  But we wouldn't expect every claim along the path to the root claim to be valid, because about half of them should come from the person we're disagreeing with.

ajsutton: I take that back, I misread our code. :)

specs/honest-challenger-fdg.md
```diff
@@ -64,7 +64,13 @@ then it must dispute claims positioned at odd depths in the game tree.
...
+invalid claims on the same team.
```
ajsutton: We should make it clear that the honest challenger does not counter a claim at all if it disagrees with the path of claims leading to that claim. The current wording sounds like it would just disagree and therefore attack.

Inphi: Rewritten to make that clearer.

ajsutton (approved):
LGTM.

refcell (approved):

PR #7472, opened 2023-09-29
txmgr: don't wait the resubmission timeout for gas increases for underpriced errors

mdehoog:
**Description**
This PR splits the `publishAndWaitForTx` into two: `publishTx` and `waitForTx`. It then calls `publishTx` synchronously from `sendTx`, and ensures that whenever a `underpriced` error is received from the mempool, it immediately bumps the gas price in a loop.

Also pulls the `feeLimitMultiplier` constant into a configurable flag, with a default of `5` to match existing behavior.

**Tests**
Much of the test suite remains the same, however the fee bumping logic was changed to return an error if the feeMultiplier cap is hit, which required some test modifications.

**Additional context**
There is an edge case where the batcher can halt due to it continually sending underpriced transactions to the mempool. This is because the fee bump only happens on the resubmission interval, which for us is set to 30 seconds (and therefore only 2 bumps occur within the 2 minute sendTx timeout).

### DIFF
--- a/op-service/txmgr/cli.go
+++ b/op-service/txmgr/cli.go
@@ -26,6 +26,7 @@ const (
 	// TxMgr Flags (new + legacy + some shared flags)
 	NumConfirmationsFlagName          = "num-confirmations"
 	SafeAbortNonceTooLowCountFlagName = "safe-abort-nonce-too-low-count"
+	FeeLimitMultiplierFlagName        = "fee-limit-multiplier"
 	ResubmissionTimeoutFlagName       = "resubmission-timeout"
 	NetworkTimeoutFlagName            = "network-timeout"
 	TxSendTimeoutFlagName             = "txmgr.send-timeout"
@@ -51,6 +52,7 @@ var (
 type DefaultFlagValues struct {
 	NumConfirmations          uint64
 	SafeAbortNonceTooLowCount uint64
+	FeeLimitMultiplier        uint64
 	ResubmissionTimeout       time.Duration
 	NetworkTimeout            time.Duration
 	TxSendTimeout             time.Duration
@@ -62,6 +64,7 @@ var (
 	DefaultBatcherFlagValues = DefaultFlagValues{
 		NumConfirmations:          uint64(10),
 		SafeAbortNonceTooLowCount: uint64(3),
+		FeeLimitMultiplier:        uint64(5),
 		ResubmissionTimeout:       48 * time.Second,
 		NetworkTimeout:            10 * time.Second,
 		TxSendTimeout:             0 * time.Second,
@@ -71,6 +74,7 @@ var (
 	DefaultChallengerFlagValues = DefaultFlagValues{
 		NumConfirmations:          uint64(3),
 		SafeAbortNonceTooLowCount: uint64(3),
+		FeeLimitMultiplier:        uint64(5),
 		ResubmissionTimeout:       24 * time.Second,
 		NetworkTimeout:            10 * time.Second,
 		TxSendTimeout:             2 * time.Minute,
@@ -115,6 +119,12 @@ func CLIFlagsWithDefaults(envPrefix string, defaults DefaultFlagValues) []cli.Fl
 			Value:   defaults.SafeAbortNonceTooLowCount,
 			EnvVars: prefixEnvVars("SAFE_ABORT_NONCE_TOO_LOW_COUNT"),
 		},
+		&cli.Uint64Flag{
+			Name:    FeeLimitMultiplierFlagName,
+			Usage:   "The multiplier applied to fee suggestions to put a hard limit on fee increases",
+			Value:   defaults.FeeLimitMultiplier,
+			EnvVars: prefixEnvVars("FEE_LIMIT_MULTIPLIER"),
+		},
 		&cli.DurationFlag{
 			Name:    ResubmissionTimeoutFlagName,
 			Usage:   "Duration we will wait before resubmitting a transaction to L1",
@@ -158,6 +168,7 @@ type CLIConfig struct {
 	SignerCLIConfig           opsigner.CLIConfig
 	NumConfirmations          uint64
 	SafeAbortNonceTooLowCount uint64
+	FeeLimitMultiplier        uint64
 	ResubmissionTimeout       time.Duration
 	ReceiptQueryInterval      time.Duration
 	NetworkTimeout            time.Duration
@@ -170,6 +181,7 @@ func NewCLIConfig(l1RPCURL string, defaults DefaultFlagValues) CLIConfig {
 		L1RPCURL:                  l1RPCURL,
 		NumConfirmations:          defaults.NumConfirmations,
 		SafeAbortNonceTooLowCount: defaults.SafeAbortNonceTooLowCount,
+		FeeLimitMultiplier:        defaults.FeeLimitMultiplier,
 		ResubmissionTimeout:       defaults.ResubmissionTimeout,
 		NetworkTimeout:            defaults.NetworkTimeout,
 		TxSendTimeout:             defaults.TxSendTimeout,
@@ -188,6 +200,9 @@ func (m CLIConfig) Check() error {
 	}
 	if m.NetworkTimeout == 0 {
 		return errors.New("must provide NetworkTimeout")
+	}
+	if m.FeeLimitMultiplier == 0 {
+		return errors.New("must provide FeeLimitMultiplier")
 	}
 	if m.ResubmissionTimeout == 0 {
 		return errors.New("must provide ResubmissionTimeout")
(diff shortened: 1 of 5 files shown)


op-service/txmgr/txmgr.go
```diff
@@ -485,15 +504,13 @@ func (m *SimpleTxManager) increaseGasPrice(ctx context.Context, tx *types.Transa
...
-		m.l.Warn(fmt.Sprintf("bumped tip getting capped at %dx multiple of the suggested value", feeLimitMultiplier), "bumped", bumpedTip, "suggestion", tip)
-		bumpedTip.Set(maxTip)
+		return nil, fmt.Errorf("bumped tip 0x%s is over %dx multiple of the suggested value", bumpedTip.Text(16), m.cfg.FeeLimitMultiplier)
 	}
-	maxFee := calcGasFeeCap(new(big.Int).Mul(basefee, big.NewInt(feeLimitMultiplier)), maxTip)
+	maxFee := calcGasFeeCap(new(big.Int).Mul(basefee, big.NewInt(int64(m.cfg.FeeLimitMultiplier))), maxTip)
 	if bumpedFee.Cmp(maxFee) > 0 {
-		m.l.Warn("bumped fee getting capped at multiple of the implied suggested value", "bumped", bumpedFee, "suggestion", maxFee)
-		bumpedFee.Set(maxFee)
+		return nil, fmt.Errorf("bumped fee 0x%s is over %dx multiple of the suggested value", bumpedFee.Text(16), m.cfg.FeeLimitMultiplier)
```
mdehoog, line 520: switching the capped amounts to errors to avoid an infinite loop

op-service/txmgr/txmgr.go
```diff
@@ -403,6 +394,34 @@ func (m *SimpleTxManager) publishAndWaitForTx(ctx context.Context, tx *types.Tra
...
+			m.metr.TxPublished("tx_replacement_underpriced")
+		case errStringMatch(err, txpool.ErrUnderpriced):
+			l.Warn("transaction is underpriced", "err", err)
+			m.metr.TxPublished("tx_underpriced")
+		default:
+			return err
+		}
+
+		// Increase the gas price & retry
+		tx, err = m.increaseGasPrice(ctx, tx)
```
mdehoog: this is broken as is; the modified tx needs to be returned for the `waitMined` to work, will refactor

tynes:
Definitely a good idea

op-service/txmgr/txmgr.go
```diff
@@ -297,19 +294,17 @@ func (m *SimpleTxManager) sendTx(ctx context.Context, tx *types.Transaction) (*t
 
 	sendState := NewSendState(m.cfg.SafeAbortNonceTooLowCount, m.cfg.TxNotInMempoolTimeout)
 	receiptChan := make(chan *types.Receipt, 1)
-	sendTxAsync := func(tx *types.Transaction) {
-		defer wg.Done()
-		m.publishAndWaitForTx(ctx, tx, sendState, receiptChan)
+	sendTx := func(tx *types.Transaction, bumpFees bool) *types.Transaction {
```
roberto-bayardo: can you use a name other than sendTx to disambiguate it from the current function?

op-service/txmgr/txmgr.go
```diff
@@ -485,15 +515,13 @@ func (m *SimpleTxManager) increaseGasPrice(ctx context.Context, tx *types.Transa
 	bumpedTip, bumpedFee := updateFees(tx.GasTipCap(), tx.GasFeeCap(), tip, basefee, m.l)
 
 	// Make sure increase is at most 5x the suggested values
```
roberto-bayardo: fix the comment to remove the "5x" ?

op-service/txmgr/txmgr.go
```diff
@@ -350,57 +334,103 @@ func (m *SimpleTxManager) sendTx(ctx context.Context, tx *types.Transaction) (*t
 // publishAndWaitForTx publishes the transaction to the transaction pool and then waits for it with [waitMined].
 // It should be called in a new go-routine. It will send the receipt to receiptChan in a non-blocking way if a receipt is found
```
roberto-bayardo: comment needs updating (no longer should be called in goroutine?)

op-service/txmgr/txmgr.go
```diff
@@ -350,57 +334,103 @@ func (m *SimpleTxManager) sendTx(ctx context.Context, tx *types.Transaction) (*t
...
 			m.metr.TxPublished("unknown_error")
 		}
-		return
+
+		// on non-underpriced error return immediately; will retry on next resubmission timeout
+		return tx
 	}
-	m.metr.TxPublished("")
 
+	m.metr.TxPublished("")
```
roberto-bayardo: The contract of this function is pretty messy; I wonder if it's worth breaking it into two, one which just tries to publish the transaction to the mempool (returning the tx if so) or returns error, and then move the functionality for waiting for it to be mined elsewhere.

If one of those non-expected (non-fee bumping) errors occurred it seems the correct behavior may be to abort entirely rather than wait and try to resubmit exactly as we might do waiting for a successfully published transaction to confirm?

mdehoog: Agree, refactored to split publish from wait.

I want to avoid bringing in error handling modifications outside of the underpriced errors.

op-service/txmgr/txmgr.go
```diff
@@ -297,19 +294,26 @@ func (m *SimpleTxManager) sendTx(ctx context.Context, tx *types.Transaction) (*t
 
 	sendState := NewSendState(m.cfg.SafeAbortNonceTooLowCount, m.cfg.TxNotInMempoolTimeout)
 	receiptChan := make(chan *types.Receipt, 1)
-	sendTxAsync := func(tx *types.Transaction) {
-		defer wg.Done()
-		m.publishAndWaitForTx(ctx, tx, sendState, receiptChan)
+	publishAndWait := func(tx *types.Transaction, bumpFees bool) *types.Transaction {
+		tx, published := m.publishTx(ctx, tx, sendState, bumpFees)
```
roberto-bayardo, line 299: Should publishTx be called synchronously?  It shouldn't block for long, and as long as it's blocking we probably don't want to be retrying it.

roberto-bayardo: Hmm, I guess if it does error out though you are relying on the below loop timeout to handle the retry.  I wonder if retry for publish error should be a different timeout than retry-for-stuck-in-txpool?

op-service/txmgr/cli.go
```diff
@@ -115,6 +119,12 @@ func CLIFlagsWithDefaults(envPrefix string, defaults DefaultFlagValues) []cli.Fl
 			Value:   defaults.SafeAbortNonceTooLowCount,
 			EnvVars: prefixEnvVars("SAFE_ABORT_NONCE_TOO_LOW_COUNT"),
 		},
+		&cli.Uint64Flag{
+			Name:    FeeLimitMultiplierFlagName,
+			Usage:   "The multiplier applied to fee suggestions to put a hard limit on fee increases",
+			Value:   defaults.FeeLimitMultiplier,
+			EnvVars: prefixEnvVars("FEE_LIMIT_MULTIPLIER"),
```
trianglesphere:
```suggestion
			EnvVars: prefixEnvVars("TXMGR_FEE_LIMIT_MULTIPLIER"),
```
We're trying to use the txmgr prefix for new flags that don't need to maintain compatibility.

op-service/txmgr/txmgr.go
mdehoog: the reason it's called synchronously is because it modifies the `tx`... if it were async, we'd have to build messaging between the two go routines to get the updated tx back

op-service/txmgr/txmgr.go
```diff
@@ -297,19 +294,26 @@ func (m *SimpleTxManager) sendTx(ctx context.Context, tx *types.Transaction) (*t
...
+				m.waitForTx(ctx, tx, sendState, receiptChan)
+			}()
+		} else {
+			wg.Done()
+		}
+		return tx
 	}
 
 	// Immediately publish a transaction before starting the resumbission loop
 	wg.Add(1)
```
trianglesphere: Because publishAndWait starts as sync, we could now move `wg.Add` into that function

op-service/txmgr/txmgr.go
roberto-bayardo: right sorry, I was suggesting it should be called synchronously, somehow having missed it is already being called that way.

op-service/txmgr/txmgr.go
mdehoog: good idea 👍

roberto-bayardo (commented):
lgtm

trianglesphere (approved):
LGTM. Thanks for the improvement.

PR #7511, opened 2023-10-02
op-upgrade: superchain target

tynes:
**Description**

Allow for the superchain target to be passed in via flag. Previously an assumption existed that there would be a single superchain target per L1 network, now that there is a dev and prod superchain target on goerli, this flag needs to exist to differentiate them.

### DIFF
--- a/op-chain-ops/cmd/op-upgrade/main.go
+++ b/op-chain-ops/cmd/op-upgrade/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"slices"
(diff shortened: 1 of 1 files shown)


op-chain-ops/cmd/op-upgrade/main.go
```diff
@@ -116,23 +126,37 @@ func entrypoint(ctx *cli.Context) error {
...
 		log.Info(chainConfig.Name, "l1-chain-id", l1ChainID, "l2-chain-id", l2ChainID)
 
 		log.Info("Detecting on chain contracts")
 		// Tracking the individual addresses can be deprecated once the system is upgraded
 		// to the new contracts where the system config has a reference to each address.
-		addresses, ok := superchain.Addresses[l2ChainID.Uint64()]
+		addresses, ok := superchain.Addresses[chainConfig.ChainID]
 		if !ok {
-			return fmt.Errorf("no addresses for chain ID %d", l2ChainID.Uint64())
+			return fmt.Errorf("no addresses for chain ID %d", l2ChainID)
```
ajsutton: It's a bit confusing to use a different variable for the chain ID used to look up the addresses from what the log messages uses.  Both would probably be better as `l2ChainID` or just inline `l2ChainID` everywhere and always used `chainConfig.ChainID`.

ajsutton (approved):
LGTM - just the one small detail about being consistent about how we refer to the l2 chain ID.

op-chain-ops/cmd/op-upgrade/main.go
tynes: You are right that `l2ChainID` as a variable is probably not necessary

sebastianst (approved):

PR #7517, opened 2023-10-03
op-challenger: Simplify game implementation

ajsutton:
**Description**

Stores claims in an array with the same order as the contracts would. Computes a claim ID by hashing the values that must be unique.  This avoid the need for two custom structs and maintaining complex parent/child relationships while also preparing for the move to use `*big.Int` in `Position` which means the `claimEntry` struct won't be suitable for use as a map key.

**Tests**

Pure refactoring with no change to behaviour outside of the `game` struct as confirmed by existing tests.

### DIFF
--- a/op-challenger/game/fault/types/game.go
+++ b/op-challenger/game/fault/types/game.go
@@ -18,12 +18,6 @@ var (
 
 // Game is an interface that represents the state of a dispute game.
 type Game interface {
-	// Put adds a claim into the game state.
-	Put(claim Claim) error
-
-	// PutAll adds a list of claims into the game state.
-	PutAll(claims []Claim) error
-
 	// Claims returns all of the claims in the game.
 	Claims() []Claim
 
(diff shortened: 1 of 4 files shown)


op-challenger/game/fault/types/game.go
```diff
@@ -112,62 +106,40 @@ func (g *gameState) Put(claim Claim) error {
...
 	}
-	parent.children = append(parent.children, makeClaimEntry(claim))
-	claimWithExtension := &extendedClaim{
-		self:     claim,
-		children: make([]claimEntry, 0),
-	}
-	g.claims[makeClaimEntry(claim)] = claimWithExtension
-	g.contractIndicies[claim.ContractIndex] = claimWithExtension
+
+	g.claims = append(g.claims, claim)
```
Inphi: This would break claims ordering if Put wasn't called in the right order. This could easily be misused as the caller needs to make sure that they're Putting claims in order.

Perhaps we still need the old map for efficient updates to the game state.

ajsutton: I was thinking the next thing we should do is delete `Put` and `PutAll`. Production code never actually uses them.

ajsutton: Have pushed up an additional commit that removes `Put` and `PutAll`. Might cause a few merge conflicts with the deep position changes but should be ok.

ajsutton: We aren't performing the same validation on inputs from the contracts we were in terms of checking the parent index lines up etc, but we really need to trust our L1 node or we're going to lose all our bonds anyway so I think I'm ok with that.  I could add them back into the NewGameState or the loader component if we think they're needed though.

Inphi (approved):

PR #7329, opened 2023-09-21
op-node: Use RefMetrics and Event from op-service/metrics package

joohhnnn:
fix issue [6735](https://github.com/ethereum-optimism/optimism/issues/6735)

PR #7519, opened 2023-10-03
chore: op-service metrics cleanup

### LINKED ISSUE #6741: op-service metrics: cleanup metrics setup utils
- [ ] metrics-docs should be generalized, but currently op-proposer and op-batcher duplicate the cmd/doc metrics-doc-generator functionality.
- [ ] CLI flags should use cli.GenericFlag
- [x] Move `op-node/metrics/{caching.go,events.go}` into `op-service/metrics`

This was implemented partially in https://github.com/ethereum-optimism/optimism/pull/6242 but that prematurely introduced an `App` abstraction, which we should only introduce once the other op-service issues are cleaned up, and we can provide a more consistent app pattern that captures all of the op-service functionality.

Sabnock01:
**Description**

This PR introduces several changes part of the overall metrics cleanup:
* The metrics-docs are moved into `op-service/metrics/doc/` and deduplicated from `op-batcher` and `op-proposer`.
* The CLIFlags in `op-service/metrics/cli.go` have been converted from `cli.Flag` to `cli.GenericFlag`
* `op-node/metrics/caching.go` has been moved into `op-service/metrics`

**Additional context**

Pinging @sebastianst for visibility

**Metadata**

- Fixes #6741


protolambda (changes requested):
The intention was not to move the whole `op-node` metrics definition into op-service, but just the building blocks that are common with other packages. Can you move `metrics.go` back?

Part of this issue was already solved in https://github.com/ethereum-optimism/optimism/pull/7329
where the op-node deduplicated the metrics utils, other than the doc-generation.

So now it's only really the doc util that still remains to be deduplicated between services.
Note that the old `op-proposer` and `op-batcher` and `op-node` doc commands each imported a different metrics definition. We need to use some interface/generic to make the same metrics-docs generating code work with the different services.

op-service/metrics/cli.go
```diff
@@ -32,21 +35,44 @@ func CLIFlags(envPrefix string) []cli.Flag {
...
+	return string(fv)
+}
+
+func (fv *FlagValue) Clone() any {
+	cpy := *fv
+	return &cpy
+}
+
+var _ cliapp.CloneableGeneric = (*FlagValue)(nil)
+
```
protolambda: These can be `StringFlag` and `IntFlag`. The CI issue with flags earlier was a problem with `GenericFlag`, which has been resolved in `develop` for any existing generic flags.

If you like we can implement specialized host and port flag types with specialized value parsing, based on the `GenericFlag` type, but it's not required for the main cleanup of this PR: deduplicating the metrics docs functionality.

protolambda: You may need to rebase on `develop` to get the CI to pass though

Sabnock01: Very well, I'll revert them to `StringFlag` and `IntFlag` for now but would be open to doing that in a subsequent PR.

protolambda (approved):

PR #7521, opened 2023-10-03
feat(ctb): Add pre-pr:no-build script

maurelian:
Adds a new script `pre-pr:no-build` to package.json, which runs the same tasks as the `pre-pr` command, but without running `forge-clean` beforehand. 

This ensure that the compilation process is run at most once.


packages/contracts-bedrock/package.json
```diff
@@ -31,6 +31,7 @@
     "preinstall": "npx only-allow pnpm",
     "pre-pr": "pnpm clean && pnpm gas-snapshot && pnpm storage-snapshot && pnpm semver-lock && pnpm autogen:invariant-docs && pnpm lint && (cd ../../op-bindings && make)",
     "pre-pr:full": "pnpm test && pnpm slither && pnpm validate-deploy-configs && pnpm validate-spacers && pnpm pre-pr",
+    "pre-pr:no-build": "forge script scripts/SemverLock.s.sol && forge snapshot --no-match-test 'testDiff|testFuzz|invariant|generateArtifact' && ./scripts/storage-snapshot.sh",
```
tynes: Why is this not called in `pre-pr`? Feels like it should use the same impl

maurelian: OK, I made that happen.

packages/contracts-bedrock/package.json
```diff
@@ -11,25 +11,29 @@
   "scripts": {
     "bindings": "pnpm bindings:ts && pnpm bindings:go",
     "bindings:ts": "nx generate @eth-optimism/contracts-ts",
-    "bindings:go": "cd ../../op-bindings && make",
-    "build": "forge build",
+    "bindings:go:no-build": "cd ../../op-bindings && make binding-build",
+    "bindings:go": "pnpm clean && pnpm build && bindings:go:no-build",
```
maurelian, line 15: `bindings:go` forces a rebuild both before and after this change, but `bindings:go:no-build` provides a much faster option if you don't need to rebuild.

packages/contracts-bedrock/package.json
maurelian, line 1: The changes in this file are all intended to: 
1. maintain the result of the pre-existing command as is, ie. forcing a rebuild.
2. providing a faster (`:no-build`) alternative to the command. 

This allows for the no-build commands to be chained together.
Per the convention of the the `lint` command, the ordering of commands is made such that  the calling command follows the callee command.

packages/contracts-bedrock/package.json
maurelian, line 15: The result of calling `bindings:go` should be identical before and after this change, ie. it  forces a rebuild. 

See the relevant makefile section:

https://github.com/ethereum-optimism/optimism/blob/ded81369125cac37bcd76e53703627a46df5e08f/op-bindings/Makefile#L13-L27

packages/contracts-bedrock/package.json
```diff
@@ -11,25 +11,29 @@
...
     "prebuild": "./scripts/verify-foundry-install.sh",
+    "build": "forge build",
```
maurelian, line 17: This is the only command that breaks the composition of commands pattern (i.e `pnpm build` doesn't call `pnpm build:fuzz && pnpm build:differential`.  

But it's an important command referenced in a lot of places, so I don't think we should mess with it.

maurelian:
Hmmm, lemme see whats up with the gas snapshot

maurelian:
> Hmmm, lemme see whats up with the gas snapshot

there was a typo in the command name.

tynes (approved):

PR #7525, opened 2023-10-03
op-node: RPC Limit client does not respect context

jyellick:
This change addresses two bugs in the limit wrapper for client RPCs.

1. The existing `limit.go` implements its own semaphore which ignores the passed in context.  This means that the semaphore will block indefinitely, even when the context of the request has already expired.

2. The existing implementation does not guard against clients which invoke RPC methods after the channel underlying the custom semaphore has been closed.  This results in panics where the closed channel is written to during shutdown, and results in test flakiness.  This flakiness is most evident in the op-e2e-http-tests suite.

Along with these fixes comes a test which attempts to demonstrate the previous bad behavior.  Because these bugs are inherently tied to the interaction of multiple go routines, the test ends up being a bit complex, but is well commented and hopefully remains readable.

### DIFF
--- a/op-node/sources/limit.go
+++ b/op-node/sources/limit.go
@@ -2,17 +2,34 @@ package sources
 
 import (
 	"context"
+	"net"
 	"sync"
 
 	"github.com/ethereum-optimism/optimism/op-node/client"
 	"github.com/ethereum/go-ethereum"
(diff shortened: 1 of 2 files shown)


Inphi (approved):
A very welcome change to reduce flakes. tysm!

tynes (approved):

PR #7533, opened 2023-10-03
chore(Makefile): Fix Devnet Targets

refcell:
**Description**

Prior to this change, the `devnet-up` command had the `install-geth` and `cannon-prestate` graceful helpers, but the `devnet-allocs` and `devnet-test` didn't. Both of these targets would then fail when `geth` isn't installed.

This PR introduces a pre-devnet hook that allows for the graceful install of any required binaries to be installed or execution to occur before devnet targets are run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,13 +85,16 @@ nuke: clean devnet-clean
 	git clean -Xdf
 .PHONY: nuke
 
-devnet-up:
+pre-devnet:
 	@if ! [ -x "$(command -v geth)" ]; then \
 		make install-geth; \
 	fi
 	@if [ ! -e op-program/bin ]; then \
 		make cannon-prestate; \
 	fi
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -85,13 +85,17 @@ nuke: clean devnet-clean
...
+pre-devnet:
 	@if ! [ -x "$(command -v geth)" ]; then \
 		make install-geth; \
 	fi
 	@if [ ! -e op-program/bin ]; then \
 		make cannon-prestate; \
 	fi
+.PHONY: pre-devnet
+
+devnet-up:
```
tynes: The `make` way to do this would be

```make
devnet-up: pre-devnet
    ...
```

Inphi (approved):

tynes (approved):

Issue #7534, opened 2023-10-03
op-service: refactor utils with context-shutdown to `io.Closer` shutdown


protolambda:
Many of the current utils do not properly expose a shutdown mechanism: they assume things run to completion after getting the context cancellation.

Instead, especially for the several HTTP servers that are involved (RPC, PPROF, Metrics), we should call `Close()` and wait for completion.

*Work in progress*

PR #7538, opened 2023-10-04
op-node: Cleanup shutdown process and support idle after halt

protolambda:
**Description**

Changes:
- Introduce a `cliapp` package in `op-service` to put CLI building blocks in.
- Add a `Lifecycle` interface with Start/Stop/Stopped. It''s not `io.Closer` because `Stop` takes a context. We need a context to give the stopping flexibility: we want to stop, stop immediately, but also support idling after closing some resources before fully stopping (which too can be interrupted).
- The `OpNode` now implements this lifecycle interface. The main action simply creates an instance op-node, when given a CLI context. And the `cliapp` package handles the execution of the `Lifecycle` based on process signals.
- The `OpNode` can now `cancel` itself to stop, this avoids a synchronous call from one of its components to `Stop` which had to be worked around previously (since resources may be locked during execution, and can thus not close).
- Add idling on `Stop` when it was `halted`, after services are stopped, to avoid crash-loop when repeatedly opting to halt on startup. Metrics are up until after the idling, to signal the service is technically online and healthy.
- Clean up p2p shutdown which was giving errors on op-node shutdown: the event handler and subscription had to be cancelled before closing the underlying topic completely.
- Move pprof and metrics out of outer CLI action, into op-node. Opened #7534 to track refactoring of these service utils to be `io.Closer` rather than assuming proper cleanup with context-cancel.

**Tests**

- `TestRequiredProtocolVersionChangeAndHalt` still works
- Tested signal-based shutdown a few times on local Goerli node, to verify it's responsive and shutdown logging is as expected

**Metadata**

Fix https://github.com/ethereum-optimism/client-pod/issues/74

### DIFF
--- a/op-node/node/node.go
+++ b/op-node/node/node.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
+	"strconv"
 	"sync/atomic"
 	"time"
 
@@ -15,11 +17,14 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 
 	"github.com/ethereum-optimism/optimism/op-node/client"
+	"github.com/ethereum-optimism/optimism/op-node/heartbeat"
 	"github.com/ethereum-optimism/optimism/op-node/metrics"
 	"github.com/ethereum-optimism/optimism/op-node/p2p"
 	"github.com/ethereum-optimism/optimism/op-node/rollup/driver"
 	"github.com/ethereum-optimism/optimism/op-node/sources"
+	"github.com/ethereum-optimism/optimism/op-node/version"
 	"github.com/ethereum-optimism/optimism/op-service/eth"
+	oppprof "github.com/ethereum-optimism/optimism/op-service/pprof"
 	"github.com/ethereum-optimism/optimism/op-service/retry"
 )
 
@@ -49,15 +54,26 @@ type OpNode struct {
 	resourcesCtx   context.Context
 	resourcesClose context.CancelFunc
 
+	// resource context for anything that stays around for the post-processing phase: e.g. metrics.
+	postResourcesCtx   context.Context
+	postResourcesClose context.CancelFunc
+
 	// Indicates when it's safe to close data sources used by the runtimeConfig bg loader
 	runtimeConfigReloaderDone chan struct{}
 
 	closed atomic.Bool
+
+	// cancels execution prematurely, e.g. to halt. This may be nil.
+	cancel context.CancelCauseFunc
+	halted bool
 }
 
 // The OpNode handles incoming gossip
 var _ p2p.GossipIn = (*OpNode)(nil)
 
+// New creates a new OpNode instance.
+// The provided ctx argument is for the span of initialization only;
+// the node will immediately Stop(ctx) before finishing initialization if the context is canceled during initialization.
 func New(ctx context.Context, cfg *Config, log log.Logger, snapshotLog log.Logger, appVersion string, m *metrics.Metrics) (*OpNode, error) {
 	if err := cfg.Check(); err != nil {
 		return nil, err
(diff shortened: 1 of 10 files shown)


op-node/p2p/gossip.go
```diff
@@ -379,10 +380,22 @@ type GossipOut interface {
...
 
 type publisher struct {
-	log         log.Logger
-	cfg         *rollup.Config
+	log log.Logger
+	cfg *rollup.Config
+
+	// p2pCtx is used for downstream message-handling resources,
+	// these can be cancelled without blocking.
+	p2pCtx    context.Context
```
sebastianst: Looks like you never access this field, so it can probably removed. You only need access to `p2pCancel` in `Close`, if I see correctly in `JoinGossip`.

op-node/node/node.go
```diff
@@ -537,10 +610,25 @@ func (n *OpNode) Close() error {
 		n.closed.Store(true)
 	}
 
+	if n.halted {
```
trianglesphere, line 609: The code changes are fine, I'm a little curious about how this halt behavior interacts with different types of users. I'm a little worried that with config/setup errors we're suppressing it for too long for typical users here.

op-node/node/node.go
```diff
@@ -319,13 +350,53 @@ func (n *OpNode) initMetricsServer(ctx context.Context, cfg *Config) error {
...
+
+	go func(url string) {
+		if err := heartbeat.Beat(n.resourcesCtx, n.log, url, payload); err != nil {
+			log.Error("heartbeat goroutine crashed", "err", err)
+		}
+	}(cfg.Heartbeat.URL)
+	return nil
+}
+
+func (n *OpNode) initPProf(_ context.Context, cfg *Config) error {
```
sebastianst: This function always returns `nil`, can just remove the return value.

op-node/node/node.go
```diff
@@ -319,13 +350,53 @@ func (n *OpNode) initMetricsServer(ctx context.Context, cfg *Config) error {
...
 	go func() {
-		if err := n.metrics.Serve(ctx, cfg.Metrics.ListenAddr, cfg.Metrics.ListenPort); err != nil {
+		if err := n.metrics.Serve(n.postResourcesCtx, cfg.Metrics.ListenAddr, cfg.Metrics.ListenPort); err != nil {
 			log.Crit("error starting metrics server", "err", err)
 		}
 	}()
 	return nil
 }
 
+func (n *OpNode) initHeartbeat(_ context.Context, cfg *Config) error {
```
sebastianst: This function always returns nil, can just remove the return value.

op-node/node/node.go
```diff
@@ -229,27 +261,26 @@ func (n *OpNode) initRuntimeConfig(ctx context.Context, cfg *Config) error {
 				case errNodeHalt, nil:
 					n.log.Debug("reloaded runtime config", "l1_head", l1Head)
 					if err == errNodeHalt {
```
sebastianst: you use `errors.Is` above to check for `errNodeHalt`, maybe a good opportunity to fix it here too.

op-service/cliapp/lifecycle.go
```diff
@@ -0,0 +1,76 @@
...
+// With the close argument a lifecycle may choose to shut itself down.
+// A service may choose to idle, dump debug information or otherwise delay
+// a shutdown when the Stop context is not expired.
+type LifecycleAction func(ctx *cli.Context, close context.CancelCauseFunc) (Lifecycle, error)
+
+// LifecycleCmd turns a LifecycleAction into an CLI action,
+// by instrumenting it with CLI context and signal based termination.
+// The app may continue to run post-processing until fully shutting down.
+// The user can force an early shut-down during post-processing by sending a second interruption signal.
+func LifecycleCmd(fn LifecycleAction) cli.ActionFunc {
```
sebastianst, line 37: Wondering if this can be tested?

op-node/node/node.go
protolambda: The `go func` in there is a carry-over from the previous heartbeat setup, and I don't want to refactor the whole heartbeat service to `io.Closer` now, but we'll need the error-return back once the service doesn't drop the unhandled errors anymore. For now I'll simplify the code though, fair point.

op-node/node/node.go
protolambda: Same as with the heartbeat, see other comment.

op-node/node/node.go
protolambda: It's only idling this way if it's an intentional halt and not when it's an actual op-node error, so it won't delay any actual error from being surfaced to the user.

op-service/cliapp/lifecycle.go
protolambda: Added testing, and improved the context-cause reporting in the erorrs. Testing is a bit complicated, since it's dealing with a mock app lifecycle and process signals at any stage of the lifecycle, but it covers all the code now :+1:

sebastianst (approved):

PR #7543, opened 2023-10-04
fix(op-challenger): Bottom Provider Split Factory Pattern

refcell:
**Description**

Fixes the `SplitTraceProvider` to use a factory for the `bottomProvider`.

**Metadata**

- Fixes https://github.com/ethereum-optimism/client-pod/issues/83

### DIFF
--- a/op-challenger/game/fault/trace/split/provider.go
+++ b/op-challenger/game/fault/trace/split/provider.go
@@ -2,6 +2,8 @@ package split
 
 import (
 	"context"
+	"errors"
+	"math/big"
 
 	"github.com/ethereum-optimism/optimism/op-challenger/game/fault/types"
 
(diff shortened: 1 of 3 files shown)


refcell:
Opening this as RFR to ask for initial comments on the direction this takes wrt mapping lower trace providers using the top provider leaf node as a key.

op-challenger/game/fault/trace/split/provider.go
```diff
@@ -11,39 +13,89 @@ import (
...
+	leftIndex := new(big.Int).Sub(ancestorIndex, big.NewInt(1))
+	rightIndex := new(big.Int).Add(ancestorIndex, big.NewInt(1))
+	prePos := types.NewPosition(int(s.topDepth), leftIndex)
+	postPos := types.NewPosition(int(s.topDepth), rightIndex)
+	preClaim, err := s.upper.Get(ctx, prePos)
+	if err != nil {
+		return nil, err
 	}
-	return s.topDepth, s.bottomProvider
+	postClaim, err := s.upper.Get(ctx, postPos)
```
ajsutton, line 68: This is nice and simple but unfortunately doesn't work.  The pre and post claims must come from the claims made in the game, not from our trace provider, otherwise they may not match what is used on chain.  One or either of the pre and post-claim values may have been provided by our opponent and we're trying to disprove it.  The input to `op-program` should be the output root they provide and the cannon trace we claim to have must then have the invalid status.

We need to walk back up the claim ancestors until we find the ones that are from the trace index we need for pre and post claims.

op-challenger/game/fault/types/position.go
```diff
@@ -57,6 +57,14 @@ func (p Position) RelativeToAncestorAtDepth(ancestor uint64) (Position, error) {
 	return NewPosition(int(newPosDepth), newIndexAtDepth), nil
 }
 
+// IndexForParentDepth returns the index of a parent position at a given depth.
+func (p Position) IndexForParentDepth(depth int) *big.Int {
```
ajsutton, line 61: The parent is strictly one level higher than the current position.  So naming-wise I think this would just be `IndexAtDepth` or `IndexAtAncestorDepth`.

op-challenger/game/fault/trace/split/provider.go
```diff
@@ -11,39 +13,89 @@ import (
...
+		topDepth:     topDepth,
 	}
 }
 
-func (s *SplitTraceProvider) providerForDepth(depth uint64) (uint64, types.TraceProvider) {
-	if depth <= s.topDepth {
-		return 0, s.topProvider
+func (s *SplitTraceProvider) buildLowerProvider(ctx context.Context, pos types.Position) (types.TraceProvider, error) {
+	ancestorIndex := pos.IndexForParentDepth(int(s.topDepth))
+	provider, ok := s.lower[ancestorIndex]
```
ajsutton, line 54: This won't be unique enough because there may be two claims at the same upper-leaf-node position with different values and they will need to have different lower trace providers (see other comment about using the pre/post claim from the game, not from our trace provider).

We could use the contract index of the leaf claim involved. Otherwise we'd need to use the pre and post claim values as the key.

PR #7282, opened 2023-09-18
fix(op-bootnode): add missed flags and enable metrics server

ghost:
**Description**

Current op-bootnode has a issue: https://github.com/ethereum-optimism/optimism/issues/7026 
and this pr try to fix it

**Tests**
no additional tests for it

**Additional context**
none

**Metadata**

PR #7548, opened 2023-10-04
fix: pass in env var prefix for p2p flags

anikaraghu:
**Description**

In #7282  bootnode metrics were enabled, but the metrics defined in P2P flag have the `OP_NODE` prefix by default. Updating so we can pass in an env var prefix.
Recommend reading the diff ignoring whitespace: https://github.com/ethereum-optimism/optimism/pull/7548/files?w=1

**Tests**

N/A


trianglesphere (approved):
LGTM

PR #7549, opened 2023-10-04
contracts-bedrock: update reinitialize value

tynes:
**Description**

Updates the `reinitailze` value to 3 from 2. Now there is a constant used in all locations so that everything can be updated at once. This solution is not scalable long term because this means that the `reinitialize` value needs to be updated any time a new contract is deployed.


packages/contracts-bedrock/src/L1/ProtocolVersions.sol
```diff
@@ -35,8 +36,8 @@ contract ProtocolVersions is OwnableUpgradeable, ISemver {
     event ConfigUpdate(uint256 indexed version, UpdateType indexed updateType, bytes data);
 
     /// @notice Semantic version.
-    /// @custom:semver 0.2.0
-    string public constant version = "0.2.0";
+    /// @custom:semver 1.0.0
+    string public constant version = "1.0.0";
```
refcell, line 40: Major upgrade here? The others look to all be minor

refcell (approved):

packages/contracts-bedrock/src/L1/ProtocolVersions.sol
tynes: Yes, going to prod with this one

PR #7555, opened 2023-10-05
contracts-bedrock: SystemConfig reinit fix

tynes:
**Description**

Makes the `SystemConfig` able to be initialized again. It previously couldn't be because of a strict require statement on a state variable that was set in a previous time that the contract was initialized.

A full upgrade simulation using this implementation can be found here: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/e2d907e4-d5cb-4034-ae32-4b73d655cedf?trace=0.0.0.4.0.13.0


packages/contracts-bedrock/src/L1/SystemConfig.sol
```diff
@@ -249,18 +249,18 @@ contract SystemConfig is OwnableUpgradeable, ISemver {
...
     /// @param  _startBlock The start block override to set in storage.
     function _setStartBlock(uint256 _startBlock) internal {
-        require(startBlock == 0, "SystemConfig: cannot override an already set start block");
-        if (_startBlock != 0) {
-            // There is an override, it cannot already be set.
+        if (_startBlock != 0 && startBlock == 0) {
+            // There is an override and it is not already set, this is for legacy chains.
             startBlock = _startBlock;
-        } else {
+        } else if (startBlock == 0) {
```
refcell, line 261: I think the dup'd `startBlock` sload is ok here since the paths are mutually exclusive and should be handled by the compiler. But could always dedup the comparison with a `bool isStartBlockZero = startBlock == 0;` prior to the conditional.

refcell (approved):

PR #7557, opened 2023-10-05
Enable guardian to extend the pause

maurelian:
**Description**

Adds a partial spec document for the Superchain Config contract. The section on the pausability of the contract should be useful for reviewing this PR.

Beyond that, the primary changes to the pause functionality are: 

1. Adding an `identifier` field to the `pause` function.
2. Adding and `extendPause()` function which can be used to extend the pause before it expires.


packages/contracts-bedrock/src/libraries/Slot.sol
maurelian, line 1: This lib is a dupe of Storage.sol

specs/superchain_config.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Roles
+
+The superchain config contract defines the following roles:
+
+### System Owner
+
+The `systemOwner` account has the ability to to remove entries from the batcher allow list. The
+Optimism Superchain system will be configured such that the system owner is a
+[`DelayedVetoable`](./delayed-vetoable.md) contract.
```
maurelian: Link is currently broken, but I'll add it in a forthcoming PR.

packages/contracts-bedrock/.gas-snapshot
```diff
@@ -141,7 +141,7 @@ GasBenchMark_L1StandardBridge_Finalize:test_finalizeETHWithdrawal_benchmark() (g
 GasBenchMark_L2OutputOracle:test_proposeL2Output_benchmark() (gas: 100966)
 GasBenchMark_OptimismPortal:test_depositTransaction_benchmark() (gas: 75115)
 GasBenchMark_OptimismPortal:test_depositTransaction_benchmark_1() (gas: 75724)
-GasBenchMark_OptimismPortal:test_proveWithdrawalTransaction_benchmark() (gas: 155604)
```
maurelian, line 144: test comment

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -5,7 +5,7 @@ import { Initializable } from "@openzeppelin/contracts/proxy/utils/Initializable
 import { ISemver } from "src/universal/ISemver.sol";
 import { Types } from "src/libraries/Types.sol";
 import { Hashing } from "src/libraries/Hashing.sol";
-import { Slot } from "src/libraries/Slot.sol";
+import { Storage } from "src/libraries/Storage.sol";
```
maurelian, line 8: Because I'd merged in changes from `develop` to the base branch (`feat/security-council`), both the `Slot` and `Storage` libs were coexisting as duplicates. 
This PR removes `Slot`.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -129,51 +136,76 @@ contract SuperchainConfig is Initializable, ISemver {
...
-        paused_ = Slot.getUint(PAUSED_TIME_SLOT) > block.timestamp;
+        paused_ = Storage.getUint(PAUSED_TIME_SLOT) > block.timestamp;
+    }
+
+    /// @notice Getter for the paused address.
+    function pausedUntil() public view returns (uint256 paused_) {
+        paused_ = Storage.getUint(PAUSED_TIME_SLOT);
     }
 
     /// @notice Pauses withdrawals.
```
tynes: I am curious how you feel about reducing the ABI and having branching logic in `pause` based on whether or not it is already paused. The big difference between the branches is sloading the pause end time vs using the block timestamp. This will reduce some duplicate code and make the administration of pausing more simple. The difference here is a signed tx for extend pause can be invalidated by the pause duration running out but it will become valid again if the system is paused again, assuming no nonce invalidations

maurelian: Nice idea. The fact that auth and the storage written to is identical on both seems like a good indication that they can be combined. It simplifies both the interface and operations so sounds good.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -129,51 +136,70 @@ contract SuperchainConfig is Initializable, ISemver {
...
+            // Pause is not active, so set the pause end time to the current time plus the duration.
+            Storage.setUint(PAUSED_TIME_SLOT, uint256(block.timestamp) + duration);
+            emit Paused(duration, identifier);
+        } else {
+            // Pause is already active, so extend the current pause end time by the duration.
+            uint256 currentPausedUntil = Storage.getUint(PAUSED_TIME_SLOT);
+            uint256 newPausedUntil = currentPausedUntil + duration;
+
+            Storage.setUint(PAUSED_TIME_SLOT, newPausedUntil);
+            emit PauseExtended(duration, identifier);
```
tynes, line 192: Do we need 2 different events?

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -129,51 +136,70 @@ contract SuperchainConfig is Initializable, ISemver {
...
-        emit Paused();
+        require(duration <= Storage.getUint(MAX_PAUSE_SLOT), "SuperchainConfig: duration exceeds maxPause");
+
+        if (paused() == false) {
+            // Pause is not active, so set the pause end time to the current time plus the duration.
+            Storage.setUint(PAUSED_TIME_SLOT, uint256(block.timestamp) + duration);
+            emit Paused(duration, identifier);
+        } else {
+            // Pause is already active, so extend the current pause end time by the duration.
+            uint256 currentPausedUntil = Storage.getUint(PAUSED_TIME_SLOT);
```
tynes: naming nit: could remove intermediate variable here by doing read and addition in a single expression but not a big deal

tynes (approved):

tynes:
Looks like a race condition in `op-service-tests`

```
=== FAIL: op-service/txmgr TestSend/dual_threaded_blocking (3.00s)
    queue_test.go:236: 
                Error Trace:    /root/project/op-service/txmgr/queue_test.go:236
                Error:          Not equal: 
                                expected: []uint64{0x0, 0x1, 0x2, 0x3, 0x4}
                                actual  : []uint64{0x0, 0x1, 0x3, 0x4}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,5 +1,4 @@
                                -([]uint64) (len=5) {
                                +([]uint64) (len=4) {
                                  (uint64) 0,
                                  (uint64) 1,
                                - (uint64) 2,
                                  (uint64) 3,
                Test:           TestSend/dual_threaded_blocking
                Messages:       expected nonces do not match
```

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: In the past you've preferred to avoid nested expressions. I think it depends on the situation. 

Anyway, I looked at the code and I agree it's nicer to avoid these intermediate vars.
Changing that now.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: Yes, I think it's a different outcome and is worth differentiating the events to make that clear. 

ie. 
- `Paused(duration)` -> the system is paused until `now + duration`
- `PauseExtended(duration)` -> the system is paused until `getUint(PAUSED_TIME_SLOT) + duration`.

PR #7560, opened 2023-10-05
op-node: Synchronize node halted bit

Inphi:
Otherwise `OpNode.Stop` may not observe writes to `halted` occurring in the reloader goroutine


tynes (approved):
Good catch

protolambda (approved):

PR #7563, opened 2023-10-05
op-chain-ops: skip extra json serialization

tynes:
**Description**

Ignore JSON serialization of the skip calldata field as it is meant to be a helper for the implementation and should not be included in the serialized JSON.

### DIFF
--- a/op-chain-ops/safe/batch.go
+++ b/op-chain-ops/safe/batch.go
@@ -25,7 +25,7 @@ import (
 // the raw calldata when both the calldata and ABIs with arguments are
 // present.
 type Batch struct {
-	SkipCalldata bool
+	SkipCalldata bool               `json:"-"`
 	Version      string             `json:"version"`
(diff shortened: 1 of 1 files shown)


op-chain-ops/safe/batch.go
```diff
@@ -25,7 +25,7 @@ import (
 // the raw calldata when both the calldata and ABIs with arguments are
 // present.
 type Batch struct {
-	SkipCalldata bool
+	SkipCalldata bool               `json:"-"`
```
Inphi, line 28: nit: probably more idiomatic to make this a private field if it's not used outside this package.

Inphi (approved):

op-chain-ops/safe/batch.go
tynes: Then would need to make a `NewBatch(bool)` function to be able to set it

PR #7566, opened 2023-10-05
op-service: flag copy to protect against default-mutation of generic flag value

protolambda:
**Description**

The `cli.Generic` value of a `cli.GenericFlag` is being mutated, and makes a flag unsafe to reuse. This was causing CI test-failures in op-challenger, which re-runs the same command definition with same flag instances.

To fix this, we can protect the flags, by making a copy and replacing any generic flag definition with a flag definition copy that has a copied generic value.

This also implements a `Clone() any` func for the custom generic flag values we use around the codebase, in case we reuse those in the future.

The `ProtectFlags` function only has to be used when the app would otherwise re-run with the same flag instances, which only happens in the op-challenger.
I've added the same to op-batcher/proposer/node however, in case we want to test or re-run those the same way as the op-challenger.

**Tests**

Test that the protect function copies and protects the original values as expected.


ajsutton (approved):
LGTM. Thanks for digging into this.

refcell:
TBH it'd be cool to upstream this change directly to the urface cli module. Otherwise looks great, thanks for saving the day again @protolambda ❤️

refcell (approved):

PR #7581, opened 2023-10-06
contracts-bedrock: custom Initializable

tynes:
**Description**

Adds a new contract to replace the OpenZeppelin
`Initializable` contract. This contract makes the
`initializer()` modifier only callable by the
ERC1967 admin. This simplifies upgrades compared to using the OpenZeppelin `Initializable` contract because we want to be able to `reinitialize` the same account many times but don't want to worry about managing the `reinitialize` number. In practice, the admin is always calling the `initialize` function thru the `ProxyAdmin.upgradeAndCall` function.

When moving to this implementation of `Initializable`, it will require some storage layout updates to each of the contracts that used OpenZeppelin `Initializable`.

### DIFF
--- a/packages/contracts-bedrock/src/libraries/Initializable.sol
+++ b/packages/contracts-bedrock/src/libraries/Initializable.sol
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Constants } from "src/libraries/Constants.sol";
+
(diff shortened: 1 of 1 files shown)


maurelian (commented):
This looks great and very easy to reason about to me. I think we should ask OZ if they have considered and rejected this approach for any reason.

tynes:
Going to wait on this

tynes:
Going to close this for now

PR #7585, opened 2023-10-06
bugfix(indexer) index the mint field of the l1 deposit transactions and not the value field

hamdiallam:
We currently index the value field of the l1 deposit receipt which is different than
the amount of ETH bridged over from L1. What we want to index is the mint field which
is the amount bridged over (msg.value).


hamdiallam:
* develop
  * **PR #7585** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7585" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7585?utm_source=stack-comment).

roninjin10 (approved):

tynes:
Would be ideal to have a test case that fails previously and now passes

PR #7592, opened 2023-10-09
op-service: http-server util, Close explicitly to wait for resource cleanup before shutdown

### LINKED ISSUE #7534: op-service: refactor utils with context-shutdown to `io.Closer` shutdown

protolambda:
**Description**

Pay off tech-debt:
- Refactor "listen and serve" anti-pattern to explicit Server instance + `Close()` call
- We can now log the address after binding to a port, so the user will know what port to use when `0` was configured
- Explicitly `Close` to wait for listener close and server-shutdown before exiting.

In:
- HTTP util server
- PProf util server
- Metrics util server
- RPC-server of op-node

For:
- op-node
- op-proposer
- op-batcher
- op-heartbeat
- endpoint-monitor
- indexer
- op-bootnode
- op-wheel

In a follow-up PR we can then wrap the above services into a full lifecycle-managed CLI command, to make the calls to the `Close()` less tricky.

**Tests**

Test HTTPServer shutdown cases: basic, force-shutdown, graceful.

**Metadata**

Fixes https://github.com/ethereum-optimism/client-pod/issues/93

Fixes https://github.com/ethereum-optimism/optimism/issues/7534

### DIFF
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -4,8 +4,9 @@ import (
 	"context"
 	"fmt"
 	"math/big"
-	"net/http"
+	"net"
 	"runtime/debug"
+	"strconv"
 	"sync"
 
 	"github.com/ethereum/go-ethereum/log"
@@ -110,27 +111,27 @@ func (i *Indexer) startHttpServer(ctx context.Context) error {
 	r := chi.NewRouter()
 	r.Use(middleware.Heartbeat("/healthz"))
 
-	server := http.Server{Addr: fmt.Sprintf("%s:%d", i.httpConfig.Host, i.httpConfig.Port), Handler: r}
-	err := httputil.ListenAndServeContext(ctx, &server)
+	addr := net.JoinHostPort(i.httpConfig.Host, strconv.Itoa(i.httpConfig.Port))
+	srv, err := httputil.StartHTTPServer(addr, r)
 	if err != nil {
-		i.log.Error("http server stopped", "err", err)
-	} else {
-		i.log.Info("http server stopped")
+		return fmt.Errorf("http server failed to start: %w", err)
 	}
-
-	return err
+	i.log.Info("http server started", "addr", srv.Addr())
+	<-ctx.Done()
+	defer i.log.Info("http server stopped")
+	return srv.Close()
 }
 
 func (i *Indexer) startMetricsServer(ctx context.Context) error {
 	i.log.Info("starting metrics server...", "port", i.metricsConfig.Port)
-	err := metrics.ListenAndServe(ctx, i.metricsRegistry, i.metricsConfig.Host, i.metricsConfig.Port)
+	srv, err := metrics.StartServer(i.metricsRegistry, i.metricsConfig.Host, i.metricsConfig.Port)
 	if err != nil {
-		i.log.Error("metrics server stopped", "err", err)
-	} else {
-		i.log.Info("metrics server stopped")
+		return fmt.Errorf("metrics server failed to start: %w", err)
 	}
-
-	return err
+	i.log.Info("metrics server started", "addr", srv.Addr())
+	<-ctx.Done()
+	defer i.log.Info("metrics server stopped")
+	return srv.Close()
 }
 
 // Start starts the indexing service on L1 and L2 chains
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -106,7 +106,7 @@ func NewIndexer(
 }
 
 func (i *Indexer) startHttpServer(ctx context.Context) error {
-	i.log.Info("starting http server...", "port", i.httpConfig.Host)
+	i.log.Debug("starting http server...", "port", i.httpConfig.Host)
 
 	r := chi.NewRouter()
 	r.Use(middleware.Heartbeat("/healthz"))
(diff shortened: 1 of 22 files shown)


op-service/httputil/server.go
```diff
@@ -1,30 +1,75 @@
...
+}
+
+func (s *HTTPServer) Close() error {
+	// closes the underlying listener too
+	err := s.srv.Close()
+	if errors.Is(err, http.ErrServerClosed) {
+		return nil
+	}
+	return err
+}
```
sebastianst: The `ListenAndServeContext` helper that was there before used a `server.Shutdown(context.Background())` to shut down the http server (which was not ideal). This is now changed to a hard `Close()`. Was that deliberate?
I think it would be good to also expose the `Shutdown(ctx)` function on the `HTTPServer` and then use it instead of `Close` with some sane shutdown timeout where it makes sense.

op-service/httputil/server.go
```diff
@@ -1,30 +1,75 @@
...
+type HTTPServer struct {
+	listener net.Listener
+	srv      *http.Server
+	closed   atomic.Bool
+}
+
+// HTTPOption applies a change to an HTTP server
+type HTTPOption func(srv *HTTPServer) error
+
+func StartHTTPServer(addr string, handler http.Handler, opts ...HTTPOption) (*HTTPServer, error) {
```
sebastianst, line 23: Would be good to have some simple unit tests of the `HTTPServer`, e.g. that start it, close it (or shutdown it) and assert that `Closed()` is false before and true after close/shutdown.

op-challenger/game/service.go
```diff
@@ -79,24 +102,20 @@ func NewService(ctx context.Context, logger log.Logger, cfg *config.Config) (*Se
...
-		sched:   sched,
-	}, nil
+	return s, nil
 }
 
 // MonitorGame monitors the fault dispute game and attempts to progress it.
 func (s *Service) MonitorGame(ctx context.Context) error {
 	s.sched.Start(ctx)
 	defer s.sched.Close()
+	defer s.Close()
```
sebastianst: Since the `Service` `s` owns its `sched`, I suggest to move `s.sched.Close()` into `Service.Close()`.

op-heartbeat/service.go
```diff
@@ -56,32 +54,52 @@ func Main(version string) func(ctx *cli.Context) error {
...
+		result = errors.Join(result, hs.pprof.Close())
+	}
+	if hs.metrics != nil {
+		result = errors.Join(result, hs.metrics.Close())
+	}
+	if hs.http != nil {
+		result = errors.Join(result, hs.http.Close())
+	}
+	return result
+}
```
sebastianst: following up on the other comment, if the `HTTPServer` has a `Shutdown(context.Context)`, then we could call `Shutdown`s here and make this `Close` a `Shutdown` instead that could be called from the `Main` with some sensible timeout.

op-node/node/node.go
```diff
@@ -381,16 +383,18 @@ func (n *OpNode) initHeartbeat(cfg *Config) {
 	}(cfg.Heartbeat.URL)
 }
 
-func (n *OpNode) initPProf(cfg *Config) {
+func (n *OpNode) initPProf(cfg *Config) error {
```
sebastianst, line 386: now I see where the error return comes back ;)

op-node/node/node.go
```diff
@@ -617,9 +621,17 @@ func (n *OpNode) Stop(ctx context.Context) error {
...
+	if n.pprofSrv != nil {
+		if err := n.pprofSrv.Close(); err != nil {
+			result = multierror.Append(result, fmt.Errorf("failed to close pprof server: %w", err))
+		}
+	}
+	if n.metricsSrv != nil {
+		if err := n.metricsSrv.Close(); err != nil {
+			result = multierror.Append(result, fmt.Errorf("failed to close metrics server: %w", err))
+		}
+	}
```
sebastianst: could call here `Shutdown` with the `ctx` given to `Stop`

indexer/indexer.go
```diff
@@ -110,27 +111,27 @@ func (i *Indexer) startHttpServer(ctx context.Context) error {
...
-
-	return err
+	i.log.Info("http server started", "addr", srv.Addr())
+	<-ctx.Done()
+	defer i.log.Info("http server stopped")
+	return srv.Close()
 }
 
 func (i *Indexer) startMetricsServer(ctx context.Context) error {
 	i.log.Info("starting metrics server...", "port", i.metricsConfig.Port)
```
sebastianst: You changed this to a debug log in other places, I think it's a good idea since you now info-log when the server was started. Same above in `startHttpServer`.

op-service/httputil/server.go
protolambda: `Shutdown` will allow ongoing requests to complete gracefully, but when we are shutting down the node we generally don't intend to serve those, although it doesn't hurt. I'll add a `Stop(ctx)` method to allow the http-server user to perform the graceful shutdown, and then we can call `Close()` to force-close any non-idle connections on user-cancellation. The app lifecyle handles this close vs force-close difference, but the existing handling does not.

op-service/httputil/server.go
protolambda: Added tests for basic/force-shutdown/graceful cases now :+1:

hamdiallam (approved):

Inphi (approved):

trianglesphere (approved):

PR #7605, opened 2023-10-09
PoC: Abstract op-challenger from contract bindings

ajsutton:
**Description**

This is definitely not something that will merge as-is - it's exploring what it would take to add a layer of abstraction between the op-challenger and the contract bindings so that it's better positioned to work with multiple versions of the game contracts.

There are two big concepts bundled together here:

1. Introduces `GameSource` (`op-challenger/game/source.go`) which wraps the existing game factory loader and is used when loading games. Rather than only returning an address of games though, it has factory methods registered that can create the appropriate player instance for each specific game type.
  * The main reason for this is that the `GameSource` is then able to create the actual bindings for the game contracts rather than having that creation happen deep inside `op-challenger`.  So we've essentially brought all the binding creations into one place.
  * It's a bit of an unintentional but nice side effect that it also moves the challenger towards supporting multiple game types in a single instance based on the game type code. There's CLI and config work that would still be needed to do this properly.
2. Introduces the `contracts` package that provides a wrapper for the contracts bindings. Ultimately the `contracts` package should be the only place where the autogenerated bindings are used directly (currently the game factory doesn't have a wrapper - only the fault dispute game and oracle). When contracts change we can implement a different wrapper and have it still implement the same interface and so be used through the rest of the code.  We may need to adjust the interface exposed to be a bit more generic or work at a slightly higher level, but we can do that because the code isn't being autogenerated.


ajsutton:
Closing this. Merged the first part of supporting multiple game types in a better way.  Investigating better ways to abstract contracts in https://github.com/ethereum-optimism/optimism/pull/7712.

PR #7607, opened 2023-10-10
contracts-bedrock: mainnet deployment

tynes:
**Description**

Latest contracts release deployed to mainnet. These are only implementation addresses. The deploy config is updated with the `systemConfigStartBlock` which should be set to the block number that the contract was initialized in. This happened in the following tx: https://etherscan.io/tx/0x76bceccd7d44656f5a129a600a6120091570b897c1d45c18cd7134cfe67c2537 which was included in block 17422444.


Inphi:
IIUC, we should only care about events from the new `SystemConfig` deployment. Are events emitted by prior versions of the `SystemConfig` no longer needed by the op-node for the sync?

tynes:
> IIUC, we should only care about events from the new `SystemConfig` deployment. Are events emitted by prior versions of the `SystemConfig` no longer needed by the op-node for the sync?

I am confused by what you mean - the set of all events that come from the proxy account are required for syncing. It doesn't really make sense to commit in the artifacts for the implementations anymore but we still do it until we completely migrate to the `superchain-registry` repo

Inphi:
> > IIUC, we should only care about events from the new `SystemConfig` deployment. Are events emitted by prior versions of the `SystemConfig` no longer needed by the op-node for the sync?
> 
> I am confused by what you mean - the set of all events that come from the proxy account are required for syncing. It doesn't really make sense to commit in the artifacts for the implementations anymore but we still do it until we completely migrate to the `superchain-registry` repo

This patch sets the `systemConfigStartBlock` to the block the latest system config was deployed on. Does this not mean we only consider events starting from that block? I'm wondering why we no longer consider config changes prior to that?

tynes:
>  Does this not mean we only consider events starting from that block?

Yes exactly - it is by definition impossible to emit events from the contract before it was initialized. See [this](https://www.notion.so/oplabs/New-L1-Contracts-6b22e34beb304ead95304c3180f21991?pvs=4) doc for an explanation. Also its the block that the proxy was initialized at, not the block the implementation was deployed

Inphi:
I see. I wanted to make sure that this is the desired behavior. But I also missed this bit in the doc, which explains how the op-node can complete the sync:
> Ideally we can hardcode the initial system config values for older op stack chains and then not need them for newer op stack chains.

Inphi (approved):

PR #7609, opened 2023-10-10
optimize(op-node): make sequencer block produce trigger time more stable

ghost:
**Description**

This PR adds an optimization to op-node eventLoop, which could make sequencer block produce trigger time more stable.

In original implementation, when timestamp is ready for sequencer to trigger sequencerStep, sequencer may not be able to do it timely(exactly) due to other caseSteps in for-select-loop(e.g. stepReqCh, altSyncTicker.C, etc) may take over the execution chance, which could result to sequencerStep delay ~xxx ms than expected time.  In normal case, this is not a issue, but if we do high TPS load test, the delay of sequencerStep may reduce the time for op-geth to do execution work and may result to a empty block(without txns from txpool) built, which is a barrier to high TPS achievement.

This PR adds higher priority to sequencerStep to make block produce trigger time more stable, and will let op-geth full block building work more stable in extremely high TPS test. In my local test, 4K TPS of native transfer txns is passed (i tuned the block gas limit and some other configs as well)

**Tests**

I did performance test and stability test locally, no issue found.

**Additional context**

N/A

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/op-node/rollup/driver/state.go
+++ b/op-node/rollup/driver/state.go
@@ -222,6 +222,24 @@ func (s *Driver) eventLoop() {
 		sequencerTimer.Reset(delay)
 	}
 
+	sequencerStep := func() error {
+		payload, err := s.sequencer.RunNextSequencerAction(ctx)
+		if err != nil {
+			s.log.Error("Sequencer critical error", "err", err)
+			return err
+		}
+		if s.network != nil && payload != nil {
+			// Publishing of unsafe data via p2p is optional.
+			// Errors are not severe enough to change/halt sequencing but should be logged and metered.
+			if err := s.network.PublishL2Payload(ctx, payload); err != nil {
+				s.log.Warn("failed to publish newly created block", "id", payload.ID(), "err", err)
+				s.metrics.RecordPublishingError()
+			}
+		}
+		planSequencerAction() // schedule the next sequencer action to keep the sequencing looping
+		return nil
+	}
+
 	// Create a ticker to check if there is a gap in the engine queue. Whenever
 	// there is, we send requests to sync source to retrieve the missing payloads.
 	syncCheckInterval := time.Duration(s.config.BlockTime) * time.Second * 2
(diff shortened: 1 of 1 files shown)


op-node/rollup/driver/state.go
```diff
@@ -266,20 +284,18 @@ func (s *Driver) eventLoop() {
...
-			if s.network != nil && payload != nil {
-				// Publishing of unsafe data via p2p is optional.
-				// Errors are not severe enough to change/halt sequencing but should be logged and metered.
-				if err := s.network.PublishL2Payload(ctx, payload); err != nil {
-					s.log.Warn("failed to publish newly created block", "id", payload.ID(), "err", err)
-					s.metrics.RecordPublishingError()
-				}
+			continue
+		default:
+		}
```
protolambda, line 292: This loop shouldn't run so hot that it needs to prioritize the sequencer channel this way. With 12s L1 blocks and 2s L2 blocks it's not expected to have this type of priority issue with that many events. If it's running hot, it likely means it's blocked or lagging on a synchronous component, like the execution engine.

We are aware of some issues related to timing and latency there, that may propagate back into this loop. Fixing the core of the problem would be preferred over changing this loop, as the even with these changes the node would still be stressed, and sequencing would keep up (due to priority) but not be timed accurately.

See https://github.com/ethereum/go-ethereum/issues/28242 for geth block building, specifically the sub-issue about the engine-API having no control to finish block-building when it requests it to, causing late blocks.

ghost: **Yeah, the core improvement should be done in op-geth👍**.  As to this pr, it changes the schedule of sequencerStep and otherSteps(e.g. derivateStep...) a little bit:
image
* former schedule, there is chance derivateStep hit the block-produce-timestamp rather than sequencerStep
* after schedule, more likely to let sequencerStep hit the block-produce-timestamp, and arrange the rest steps(e.g. derivateStep) to the rest of time window

This may be a little optimization i think.

PR #7611, opened 2023-10-10
contracts-bedrock: `StorageSetter` contract

tynes:
**Description**

Add + test a `StorageSetter` contract that can be
used by upgrades and storage migrations. This contract exists to decouple migrations from implementations. The existing contracts use `Initialize` which is a leaky abstraction because it includes the fact that the implementation is behind a proxy in the implementation code itself. The logic in `initialize` must include logic that can ensure the storage is correct for any version upgrading to any version as well as going from empty storage for new deployments, adding a lot of complexity. This contract should be used as an intermediate step in upgrades, so that this contract can be used to migrate storage between different implementations.


packages/contracts-bedrock/src/universal/StorageSetter.sol
```diff
@@ -0,0 +1,46 @@
...
+pragma solidity 0.8.15;
+
+import { ISemver } from "src/universal/ISemver.sol";
+import { Storage } from "src/libraries/Storage.sol";
+
+/// @title StorageSetter
+/// @notice A simple contract that allows setting arbitrary storage slots.
+///         WARNING: this contract is not safe to be called by untrusted parties.
+///         It is only meant as an intermediate step during upgrades.
+contract StorageSetter is ISemver {
```
maurelian, line 11: I'm having trouble understanding the flow in which this contract will be used. It has a bunch of public setter functions, but what calls them?
Will it be inherited by a contract which is customized for the specific contract being upgraded?

tynes: Right now a single upgrade and call is hand coded, keeping it simple because that is all we really need for now, see the logic in https://github.com/ethereum-optimism/optimism/pull/7612

maurelian (approved):
This LGTM as a general utility, and the functionality is well tested. 

Any usage of this contract needs to be carefully reviewed as well.

packages/contracts-bedrock/src/universal/StorageSetter.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { ISemver } from "src/universal/ISemver.sol";
+import { Storage } from "src/libraries/Storage.sol";
+
+/// @title StorageSetter
+/// @notice A simple contract that allows setting arbitrary storage slots.
+///         WARNING: this contract is not safe to be called by untrusted parties.
+///         It is only meant as an intermediate step during upgrades.
```
refcell, line 10: Could enforce with an `onlyDeployer` modifier where the deployer is set during construction. But this is also just simple enough and clear enough to use like so.

packages/contracts-bedrock/src/universal/StorageSetter.sol
```diff
@@ -0,0 +1,46 @@
...
+contract StorageSetter is ISemver {
+    /// @notice Semantic version.
+    /// @custom:semver 1.0.0
+    string public constant version = "1.0.0";
```
refcell, line 14: Should this be kept to < 0 major so as to add another indicator that this shouldn't be used outside the upgrade workflow (non-prod environments)?

refcell (approved):

packages/contracts-bedrock/src/universal/StorageSetter.sol
tynes: I see what you mean, but this is being used in the prod upgrade flow

packages/contracts-bedrock/src/universal/StorageSetter.sol
tynes: We can't use `onlyDeployer` unless the multisig deploys it, if we did only owner it would have to have an immutable owner. It is safe to be used the way that it is used at the moment

PR #7615, opened 2023-10-10
op-node: Span Batch Type, Encoding, and Decoding

protolambda:
See #7288 

Added one commit to ensure the new span-batch decoding that's introduced does not get accepted before HF activation.

### DIFF
--- a/op-node/rollup/derive/channel.go
+++ b/op-node/rollup/derive/channel.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
+
 	"github.com/ethereum-optimism/optimism/op-service/eth"
 	"github.com/ethereum/go-ethereum/rlp"
 )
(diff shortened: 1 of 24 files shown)


op-node/rollup/derive/channel.go
```diff
@@ -157,6 +159,12 @@ func BatchReader(r io.Reader, l1InclusionBlock eth.L1BlockRef) (func() (BatchWit
 			L1InclusionBlock: l1InclusionBlock,
 		}
 		err := rlpReader.Decode(&ret.Batch)
-		return ret, err
+		if err != nil {
+			return ret, err
+		}
+		if ret.Batch.BatchType == SpanBatchType && !cfg.IsSpanBatch(ret.L1InclusionBlock.Time) {
```
trianglesphere, line 165: Have we specified how activation is done in the specs?

trianglesphere (approved):

op-node/rollup/derive/channel.go
protolambda: In the implementation there's a placeholder span-batch-specific activation timestamp in the configuration, but with specs and final implementation I believe we can merge it with the Canyon hardfork definition.

PR #7621, opened 2023-10-10
Span batch derivation [by Test In Prod, rebased on develop, see #7289]

protolambda:
**Description**

Rebased to resolve conflicts with `develop` branch and changes that were implemented as part of the prior PRs (#7615, #7288)

See #7289

### DIFF
--- a/op-node/cmd/batch_decoder/reassemble/reassemble.go
+++ b/op-node/cmd/batch_decoder/reassemble/reassemble.go
@@ -107,14 +107,18 @@ func processFrames(cfg *rollup.Config, id derive.ChannelID, frames []FrameWithMe
 	var batches []derive.SingularBatch
 	invalidBatches := false
 	if ch.IsReady() {
-		br, err := derive.BatchReader(cfg, ch.Reader(), eth.L1BlockRef{})
+		br, err := derive.BatchReader(ch.Reader())
 		if err == nil {
 			for batch, err := br(); err != io.EOF; batch, err = br() {
 				if err != nil {
 					fmt.Printf("Error reading batch for channel %v. Err: %v\n", id.String(), err)
 					invalidBatches = true
 				} else {
-					batches = append(batches, batch.Batch.SingularBatch)
+					if batch.BatchType != derive.SingularBatchType {
+						batches = append(batches, batch.SingularBatch)
+					} else {
+						fmt.Printf("batch-type %d is not supported", batch.BatchType)
+					}
 				}
 			}
 		} else {
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -78,7 +78,7 @@ func (bq *BatchQueue) popNextBatch(safeL2Head eth.L2BlockRef) *SingularBatch {
 	return nextBatch
 }
 
-func (bq *BatchQueue) advanceEpoch(nextBatch *SingularBatch) {
+func (bq *BatchQueue) advanceEpochMaybe(nextBatch *SingularBatch) {
 	if nextBatch.GetEpochNum() == rollup.Epoch(bq.l1Blocks[0].Number)+1 {
 		// Advance epoch if necessary
 		bq.l1Blocks = bq.l1Blocks[1:]
@@ -89,7 +89,7 @@ func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) 
 	if len(bq.nextSpan) > 0 {
 		// If there are cached singular batches, pop first one and return.
 		nextBatch := bq.popNextBatch(safeL2Head)
-		bq.advanceEpoch(nextBatch)
+		bq.advanceEpochMaybe(nextBatch)
 		return nextBatch, nil
 	}
 
@@ -169,7 +169,7 @@ func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) 
 		return nil, NewCriticalError(fmt.Errorf("unrecognized batch type: %d", batch.GetBatchType()))
 	}
 
-	bq.advanceEpoch(nextBatch)
+	bq.advanceEpochMaybe(nextBatch)
 	return nextBatch, nil
 }
 
--- a/op-node/rollup/derive/batch_queue.go
+++ b/op-node/rollup/derive/batch_queue.go
@@ -29,7 +29,12 @@ import (
 
 type NextBatchProvider interface {
 	Origin() eth.L1BlockRef
-	NextBatch(ctx context.Context) (*BatchData, error)
+	NextBatch(ctx context.Context) (Batch, error)
+}
+
+type SafeBlockFetcher interface {
+	L2BlockRefByNumber(context.Context, uint64) (eth.L2BlockRef, error)
+	PayloadByNumber(context.Context, uint64) (*eth.ExecutionPayload, error)
 }
 
 // BatchQueue contains a set of batches for every L1 block.
@@ -42,16 +47,22 @@ type BatchQueue struct {
 
 	l1Blocks []eth.L1BlockRef
 
-	// batches in order of when we've first seen them, grouped by L2 timestamp
-	batches map[uint64][]*BatchWithL1InclusionBlock
+	// batches in order of when we've first seen them
+	batches []*BatchWithL1InclusionBlock
+
+	// nextSpan is cached SingularBatches derived from SpanBatch
+	nextSpan []*SingularBatch
+
+	l2 SafeBlockFetcher
 }
 
 // NewBatchQueue creates a BatchQueue, which should be Reset(origin) before use.
-func NewBatchQueue(log log.Logger, cfg *rollup.Config, prev NextBatchProvider) *BatchQueue {
+func NewBatchQueue(log log.Logger, cfg *rollup.Config, prev NextBatchProvider, l2 SafeBlockFetcher) *BatchQueue {
 	return &BatchQueue{
 		log:    log,
 		config: cfg,
 		prev:   prev,
+		l2:     l2,
 	}
 }
 
@@ -59,7 +70,29 @@ func (bq *BatchQueue) Origin() eth.L1BlockRef {
 	return bq.prev.Origin()
 }
 
-func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) (*BatchData, error) {
+func (bq *BatchQueue) popNextBatch(safeL2Head eth.L2BlockRef) *SingularBatch {
+	nextBatch := bq.nextSpan[0]
+	bq.nextSpan = bq.nextSpan[1:]
+	// Must set ParentHash before return. we can use safeL2Head because the parentCheck is verified in CheckBatch().
+	nextBatch.ParentHash = safeL2Head.Hash
+	return nextBatch
+}
+
+func (bq *BatchQueue) advanceEpoch(nextBatch *SingularBatch) {
+	if nextBatch.GetEpochNum() == rollup.Epoch(bq.l1Blocks[0].Number)+1 {
+		// Advance epoch if necessary
+		bq.l1Blocks = bq.l1Blocks[1:]
+	}
+}
+
+func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) (*SingularBatch, error) {
+	if len(bq.nextSpan) > 0 {
+		// If there are cached singular batches, pop first one and return.
+		nextBatch := bq.popNextBatch(safeL2Head)
+		bq.advanceEpoch(nextBatch)
+		return nextBatch, nil
+	}
+
 	// Note: We use the origin that we will have to determine if it's behind. This is important
 	// because it's the future origin that gets saved into the l1Blocks array.
 	// We always update the origin of this stage if it is not the same so after the update code
@@ -89,7 +122,7 @@ func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) 
 	} else if err != nil {
 		return nil, err
 	} else if !originBehind {
-		bq.AddBatch(batch, safeL2Head)
+		bq.AddBatch(ctx, batch, safeL2Head)
 	}
 
 	// Skip adding data unless we are up to date with the origin, but do fully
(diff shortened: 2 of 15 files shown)


clabby:
Looks like we lost [pcw109550](https://github.com/pcw109550)'s co-author here in the rebase.

Otherwise, monstrous PR - thank you @ImTei & @pcw109550 🚀

protolambda:
@clabby pcw co-authored a commit that was already merged in develop here: https://github.com/ethereum-optimism/optimism/commit/d6e1259b0d6b2aba946c788ec6b60b72fdec3516  It shows up in the original version of this PR since that one isn't based on latest `develop` and has slightly conflicting history. I'm happy to amend any other commits here to add co-author data where needed though!

op-node/rollup/derive/batch_queue.go
```diff
@@ -42,24 +47,55 @@ type BatchQueue struct {
...
+		l2:     l2,
 	}
 }
 
 func (bq *BatchQueue) Origin() eth.L1BlockRef {
 	return bq.prev.Origin()
 }
 
-func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) (*BatchData, error) {
+func (bq *BatchQueue) popNextBatch(safeL2Head eth.L2BlockRef) *SingularBatch {
```
trianglesphere, line 75: Needs documentation that `len(nextSpan) > 0`

op-node/rollup/derive/batch_queue.go
```diff
@@ -42,24 +47,55 @@ type BatchQueue struct {
...
+	nextBatch := bq.nextSpan[0]
+	bq.nextSpan = bq.nextSpan[1:]
+	// Must set ParentHash before return. we can use safeL2Head because the parentCheck is verified in CheckBatch().
+	nextBatch.ParentHash = safeL2Head.Hash
+	return nextBatch
+}
+
+func (bq *BatchQueue) advanceEpochMaybe(nextBatch *SingularBatch) {
```
trianglesphere: nit: `maybeAdvanceEpoch` or `try...` or `attempt...` reads better in english.

op-node/rollup/derive/channel.go
```diff
@@ -146,25 +144,21 @@ func (ch *Channel) Reader() io.Reader {
 
 // BatchReader provides a function that iteratively consumes batches from the reader.
 // The L1Inclusion block is also provided at creation time.
-func BatchReader(cfg *rollup.Config, r io.Reader, l1InclusionBlock eth.L1BlockRef) (func() (BatchWithL1InclusionBlock, error), error) {
```
trianglesphere, line 149: Why do we no longer track the L1 Inclusion block?

op-node/rollup/derive/batch_queue.go
```diff
@@ -111,43 +147,70 @@ func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef)
...
+		if err != nil {
+			return nil, NewCriticalError(err)
+		}
+		bq.nextSpan = singularBatches
+		nextBatch = bq.popNextBatch(safeL2Head)
+	default:
+		return nil, NewCriticalError(fmt.Errorf("unrecognized batch type: %d", batch.GetBatchType()))
+	}
+
+	bq.advanceEpochMaybe(nextBatch)
```
trianglesphere: Why are we attempting to advance the epoch here?

op-node/rollup/derive/batch_queue.go
```diff
@@ -42,24 +47,55 @@ type BatchQueue struct {
 
 	l1Blocks []eth.L1BlockRef
 
-	// batches in order of when we've first seen them, grouped by L2 timestamp
```
trianglesphere, line 45: Is the map impossible to use? I see why the array is simpler, but the benefit of the map is that we could do constant time lookups per timestamp rather than having to iterate the entire array.

op-node/rollup/derive/batch_queue.go
tynes: Should it panic if the above doesn't hold?

op-node/rollup/derive/batches.go
```diff
@@ -28,14 +31,37 @@ const (
...
+			return BatchDrop
+		}
+		return checkSingularBatch(cfg, log, l1Blocks, l2SafeHead, singularBatch, batch.L1InclusionBlock)
+	case SpanBatchType:
+		spanBatch, ok := batch.Batch.(*SpanBatch)
+		if !ok {
+			log.Error("failed type assertion to SpanBatch")
+			return BatchDrop
+		}
+		if !cfg.IsSpanBatch(batch.Batch.GetTimestamp()) {
```
tynes, line 50: Is there a plan to unify between a specific span batch activation vs a canyon activation? Seems like we want to dedupe additional hardforks if possible but then it becomes a bit more difficult to test

op-node/rollup/derive/channel.go
ImTei: In the original code, `BatchReader()` returns `BatchWithL1InclusionBlock` but the L1 inclusion block is not used in `ChannelInReader`

`ChannelInReader` [just picks `BatchData`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/channel_in_reader.go#L83-L92) from the `BatchWithL1InclusionBlock`, and [the L1 inclusion block is set again in `BatchQueue`.](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/batch_queue.go#L134-L137)

So, this change is to remove unnecessary code.

op-node/rollup/derive/batch_queue.go
ImTei: Attempting to advance the epoch is moved to here [from `deriveNextBatch`](https://github.com/ethereum-optimism/optimism/pull/7621/files/0b719a9678a71204eaa82c9642042d98697b00c4#diff-70e161ad6625124a29f49468d591c738ddec2e962fcc198d3a9db85b242c1096L212-L215), because the epoch can be advanced at the middle of a single span batch. We should attempt to advance the epoch for every singular batch.

op-node/rollup/derive/batch_queue.go
ImTei: Changing to a single batch queue is required to allow overlapping span batches(span batches that partially overlap the current safe chain). In the original code, overlapping batches may not be in the queue at nextTimestamp. So we have to scan every batches from the past, in order of the inclusion on L1.

Additionally, if I understand correctly, there might be a memory leak in the current code since we don't clean up remaining batches from the queue after accepting a batch. This code change also resolves this leak.

ImTei: FYI, The reason why we have to allow overlapping span batches is to make the derivation is working from any starting block. If we don't allow overlapping batches, the node cannot derive the chain when the safe head is reset to the middle of the Span Batch.

op-node/rollup/derive/batches.go
ImTei: Yes we have a plan. Span batch has its own specific hard fork config now for easy development and isolated tests, but it will be merged with Canyon in the final implementation.

op-node/rollup/derive/batch_queue.go
ImTei: It would be nice to have both comment and panic 👍

ImTei:
@protolambda can you apply the simple changes suggested in the review comments? (adding [panic](https://github.com/ethereum-optimism/optimism/pull/7621#discussion_r1353498478) & [renaming method](https://github.com/ethereum-optimism/optimism/pull/7621#discussion_r1353498918)) sorry to bother you but I don't have permission on this branch 🥲

op-node/rollup/derive/batch_queue.go
protolambda: This made me realize we were allowing empty span batches, which results in more fun edge-cases like this. I changed the raw-span-batch decoding, and the raw to full span-batch derive function, to error when it's empty. And added the panic check + doc comments :+1:

protolambda:
Implemented review suggestions, added error case for empty span batch + decoding test case, and rebased on `develop`. Ready for final review/merge. Edit: and fixed the go lint.

op-node/cmd/batch_decoder/reassemble/reassemble.go
```diff
@@ -107,14 +107,18 @@ func processFrames(cfg *rollup.Config, id derive.ChannelID, frames []FrameWithMe
...
 				if err != nil {
 					fmt.Printf("Error reading batch for channel %v. Err: %v\n", id.String(), err)
 					invalidBatches = true
 				} else {
-					batches = append(batches, batch.Batch.SingularBatch)
+					if batch.BatchType != derive.SingularBatchType {
+						batches = append(batches, batch.SingularBatch)
+					} else {
+						fmt.Printf("batch-type %d is not supported", batch.BatchType)
+					}
```
ajsutton, line 121: Can we make sure we have an issue somewhere to update batch decoder with span batch support? Definitely shouldn't be done in this PR but it is a really useful tool that would be good to update.

op-node/rollup/derive/channel_in_reader.go
```diff
@@ -89,7 +89,23 @@ func (cr *ChannelInReader) NextBatch(ctx context.Context) (*BatchData, error) {
...
+	case SingularBatchType:
+		return &batchData.SingularBatch, nil
+	case SpanBatchType:
+		if origin := cr.Origin(); !cr.cfg.IsSpanBatch(origin.Time) {
+			return nil, NewTemporaryError(fmt.Errorf("cannot accept span batch in L1 block %s at time %d", origin, origin.Time))
+		}
+		// If the batch type is Span batch, derive block inputs from RawSpanBatch.
+		spanBatch, err := batchData.RawSpanBatch.derive(cr.cfg.BlockTime, cr.cfg.Genesis.L2Time, cr.cfg.L2ChainID)
+		if err != nil {
+			return nil, err
```
ajsutton, line 102: Is it an issue that this isn't a temporary error?

op-node/rollup/derive/batch_queue.go
```diff
@@ -42,24 +47,60 @@ type BatchQueue struct {
...
 }
 
 func (bq *BatchQueue) Origin() eth.L1BlockRef {
 	return bq.prev.Origin()
 }
 
-func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef) (*BatchData, error) {
+// popNextBatch pops the next batch from the current queued up span-batch nextSpan.
+// The queue must be non-empty, or the function will panic.
+func (bq *BatchQueue) popNextBatch(safeL2Head eth.L2BlockRef) *SingularBatch {
```
ajsutton, line 75: nit: Should this be called `popNextSpanBatch` given it works only with `nextSpan`?

op-node/rollup/derive/batch_queue.go
```diff
@@ -111,43 +152,71 @@ func (bq *BatchQueue) NextBatch(ctx context.Context, safeL2Head eth.L2BlockRef)
...
+	case SpanBatchType:
+		spanBatch, ok := batch.(*SpanBatch)
+		if !ok {
+			return nil, NewCriticalError(errors.New("failed type assertion to SpanBatch"))
+		}
+		// If next batch is SpanBatch, convert it to SingularBatches.
+		singularBatches, err := spanBatch.GetSingularBatches(bq.l1Blocks, safeL2Head)
+		if err != nil {
+			return nil, NewCriticalError(err)
+		}
```
ajsutton, line 173: To confirm my understanding - this will throw a critical error if we are processing a span batch that spans more L1 blocks that we have in `bq.l1Blocks`. This works because part of reading new channel data from L1 blocks means updating the origin and we then add the L1 block to `l1Blocks`.  All L1 blocks referenced by a span batch must be before the last part of the span batch is included on chain and thus we must have seen all l1 blocks. This is definitely enforced by returning `BatchUndecided` for a span batch if the last l1 block it needs isn't in `l1Blocks` as it needs to check the span batch l1 origin.

Then we only remove blocks from l1Blocks in the startup corner case where the `base` L1Block is added but we've already processed it so we basically just skip it (the `originBehind` case).  Or in the normal case, we only remove l1Blocks when we've processed all batches from that l1 block.  There's two cases of that, in `deriveNextBatch` when we've produced every batch that can possibly be produced from an L1 block (including any forced empty blocks) or in `maybeAdvanceEpoch` called when we return the next single batch (either standalone or from a span batch).

The one concern I have is if we process part of a span batch and have to roll back (eg we find a block in the span batch that includes non-deposit tx that aren't eligible for inclusion), then we'd need to reset the safe head and move on to processing a different span batch, but we will have discarded some of the l1 origin blocks that span batch needs. If we do a full pipeline reset as part of the rollback it would be ok, but then I feel like we'd try to process the invalid span batch again and get stuck in a reset loop.

op-node/rollup/derive/batch_queue.go
```diff
@@ -170,19 +239,15 @@ func (bq *BatchQueue) deriveNextBatch(ctx context.Context, outOfData bool, l2Saf
...
-			return nil, NewCriticalError(fmt.Errorf("found batch with timestamp %d marked as future batch, but expected timestamp %d", batch.Batch.Timestamp, nextTimestamp))
+			remaining = append(remaining, batch)
+			continue
 		case BatchDrop:
-			bq.log.Warn("dropping batch",
-				"batch_timestamp", batch.Batch.Timestamp,
-				"parent_hash", batch.Batch.ParentHash,
-				"batch_epoch", batch.Batch.Epoch(),
-				"txs", len(batch.Batch.Transactions),
+			batch.Batch.LogContext(bq.log).Warn("dropping batch",
```
ajsutton:
```suggestion
			batch.Batch.LogContext(bq.log).Warn("Dropping batch",
```

Not a big deal, just my OCD....

op-node/rollup/derive/batches.go
```diff
@@ -28,14 +31,37 @@ const (
...
+			log.Warn("received SpanBatch before SpanBatch hard fork")
+			return BatchDrop
+		}
+		return checkSpanBatch(ctx, cfg, log, l1Blocks, l2SafeHead, spanBatch, batch.L1InclusionBlock, l2Fetcher)
+	default:
+		log.Warn("unrecognized batch type: %d", batch.Batch.GetBatchType())
```
ajsutton:
```suggestion
		log.Warn("Unrecognized batch type: %d", batch.Batch.GetBatchType())
```

Though I do wonder if this should be an error log given we recognised the type to read the batch (ie `BatchData.decodeType` had support for this batch type) but then haven't updated this code to be able to validate it.  That feels like a programming error.

op-node/rollup/derive/batches.go
```diff
@@ -141,3 +167,204 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
+			log.Warn("batch has misaligned timestamp")
+			return BatchDrop
+		}
+		if (l2SafeHead.Time-batch.GetTimestamp())%cfg.BlockTime != 0 {
+			log.Warn("batch has misaligned timestamp")
+			return BatchDrop
+		}
+		parentNum = l2SafeHead.Number - (l2SafeHead.Time-batch.GetTimestamp())/cfg.BlockTime - 1
+		var err error
+		parentBlock, err = l2Fetcher.L2BlockRefByNumber(ctx, parentNum)
```
ajsutton, line 211: We don't have any caching for block by number and can't add it because reorgs could change the result. It should be ok, but we should check how many extra engine API calls this causes in practical usage (ie on a devnet).

op-node/rollup/derive/batches.go
```diff
@@ -141,3 +167,204 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
+	// finding parent block of the span batch.
+	// if the span batch does not overlap the current safe chain, parentBLock should be l2SafeHead.
+	parentNum := l2SafeHead.Number
+	parentBlock := l2SafeHead
+	if batch.GetTimestamp() < nextTimestamp {
+		if batch.GetTimestamp() > l2SafeHead.Time {
+			// batch timestamp cannot be between safe head and next timestamp
+			log.Warn("batch has misaligned timestamp")
+			return BatchDrop
+		}
```
ajsutton, line 204: None of the unit tests hit this case (ie everything passes when I comment it out)

op-node/rollup/derive/batches.go
```diff
@@ -141,3 +167,204 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
+	if startEpochNum+cfg.SeqWindowSize < l1InclusionBlock.Number {
+		log.Warn("batch was included too late, sequence window expired")
+		return BatchDrop
+	}
+
+	// Check the L1 origin of the batch
+	if startEpochNum > parentBlock.L1Origin.Number+1 {
+		log.Warn("batch is for future epoch too far ahead, while it has the next timestamp, so it must be invalid", "current_epoch", epoch.ID())
+		return BatchDrop
+	}
```
ajsutton, line 235: None of the unit tests hit this case (ie everything passes when I comment it out)

op-node/rollup/derive/batches.go
```diff
@@ -141,3 +167,204 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
+	if startEpochNum < parentBlock.L1Origin.Number {
+		log.Warn("dropped batch, epoch is too old", "minimum", parentBlock.ID())
+		return BatchDrop
+	}
+
+	originIdx := 0
+	originAdvanced := false
+	if startEpochNum == parentBlock.L1Origin.Number+1 {
+		originAdvanced = true
+	}
```
ajsutton, line 263: None of the unit tests hit this case (ie everything passes when I comment it out).

op-node/rollup/derive/batches.go
```diff
@@ -141,3 +167,204 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
+
+	for i := 0; i < batch.GetBlockCount(); i++ {
+		if batch.GetBlockTimestamp(i) <= l2SafeHead.Time {
+			continue
+		}
```
ajsutton, line 268: None of the unit tests hit this case (ie everything passes when I comment it out)

op-node/rollup/derive/batches.go
```diff
@@ -141,3 +167,204 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
+				originIdx = j
+				break
+			}
+
+		}
+		if i > 0 {
+			originAdvanced = false
+			if batch.GetBlockEpochNum(i) > batch.GetBlockEpochNum(i-1) {
+				originAdvanced = true
+			}
```
ajsutton, line 282: None of the unit tests hit this case (ie everything passes when I comment it out)

op-node/rollup/derive/batches.go
```diff
@@ -141,3 +167,204 @@ func CheckBatch(cfg *rollup.Config, log log.Logger, l1Blocks []eth.L1BlockRef, l
...
+			if len(safeBlockTxs)-depositCount != len(batchTxs) {
+				log.Warn("overlapped block's tx count does not match", "safeBlockTxs", len(safeBlockTxs), "batchTxs", len(batchTxs))
+				return BatchDrop
+			}
+			for j := 0; j < len(batchTxs); j++ {
+				if !bytes.Equal(safeBlockTxs[j+depositCount], batchTxs[j]) {
+					log.Warn("overlapped block's transaction does not match")
+					return BatchDrop
+				}
+			}
```
ajsutton, line 356: None of the unit tests hit this case (ie everything passes when I comment it out)

ajsutton (commented):
From what I can see this looks safe to merge while span batches are off and is generally good anyway.

The test coverage is very impressive but I've commented on a few places where the immediately relevant unit tests passed even with code commented out - it's possible those cases either are tested elsewhere or are better tested at a different level so feel free to push back, it just seemed the most effective way to evaluate the tests.

The one case that does worry me is that we're applying blocks from span batches and updating the safe head, origin and things like the batch queue `l1Blocks` based on those batches being applied, but we may find that one of the blocks in the span batch is actually invalid and I don't see how we'll roll back the span batch properly.  I think the main thing is we should write e2e tests around that kind of case and make sure that the derivation as a whole handles it since there could easily be downstream effects. This case can't be hit with span batches disabled so we don't need to solve it before merging.

The other thing I think we should consider is that its potentially misleading for users to have the op-geth safe head update when we're part way through processing a span batch, particularly if the safe head could then be rolled back if the span batch is found to be invalid.  I guess that could happen with an L1 reorg anyway so maybe not a showstopper.  That said, it seems like it wouldn't be too hard to attach a flag to the `SingularBatch` we feed down the pipeline from span batches with a flag to say to import but not update the safe head. There's definitely some downstream impacts of that so may not be simple, but it may help with handling span batches that turn out to be invalid or even just restarts in the middle of applying a span batch. I'm not across all the conversations in this area though so feel free to ignore. :)

op-node/rollup/derive/batch_queue.go
ImTei: Thanks for pointing this out! as you said, we should avoid full pipeline reset for span batch rollback. then we must consider batch queue's l1 blocks also. I was missing that part. 👍

op-node/rollup/derive/channel_in_reader.go
ImTei: All errors from `RawSpsanBatch.derive()` mean that the span batch is invalid and we should drop the batch. There's no case that we need to retry.

op-node/rollup/derive/batch_queue.go
ImTei: My intended meaning is "pop the next singular batch from the span batch." So I think `popNextBatch` makes sense. what do you think?

op-node/rollup/derive/batches.go
ImTei: Yes we should reduce usage of engine RPC as much as possible. Fortunately, additional RPC reqs occur only for overlapping batches. And current op-batcher won't make overlapping batches if working normally. So in most cases, op-node handles overlapping batch when resetting the pipeline, expecting not very often.

ImTei:
@ajsutton Thank you for the review and checking test coverage 👍 
As you have seen, there is a discussion about span batch atomicity and rollback. I'm working on it now. 
And of course we need more test cases, both unit and e2e tests. I think we can make them as follow-up PRs. your coverage check will be very helpful to me!

op-node/rollup/derive/batch_queue.go
ajsutton: hmm, I see the confusion - for me it screams that it's the method that is used to get the next batch regardless of source and has no indication that it's span batch specific.  Maybe `popNextBatchFromSpan` but yeah it is hard to name. It has the comment so I can live with it either way.

op-node/cmd/batch_decoder/reassemble/reassemble.go
protolambda: https://github.com/ethereum-optimism/optimism/issues/7851

protolambda:
Opened an issue to track the span-batch validation test-coverage items @ajsutton raised so we don't lose track of it. https://github.com/ethereum-optimism/optimism/issues/7853

ajsutton (approved):

op-node/rollup/derive/batch_queue.go
joohhnnn: Maybe `popNextCachedBatch` ?

PR #7622, opened 2023-10-11
op-upgrade: cleanup implementation

tynes:
**Description**

Use a higher level API that is more appropriate for building calldata for calls. The lower level API did not append the 4byte selectors and that had to be done manually. Using the low level API was error prone so it should be avoided. The high level API does exactly what we want it to do - ABI encode the data and then append it to the correct 4byte selector.

To review this PR, ensure that the arguments to the old `PackValues` functions are the same as the new `Pack` function, excluding the first argument which is the name of the function to ABI encode for.

### DIFF
--- a/op-chain-ops/upgrades/l1.go
+++ b/op-chain-ops/upgrades/l1.go
@@ -14,6 +14,10 @@ import (
 	"github.com/ethereum-optimism/superchain-registry/superchain"
 )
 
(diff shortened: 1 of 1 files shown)


ajsutton (approved):
LGTM.

tynes:
> LGTM.

Thank you for the review!

Issue #6728, opened 2023-08-11
L2 Shapella (Shanghai + Capella) hardfork support


protolambda:
This issue tracks the necessary changes to support the Shapella L1 upgrade in the OP-Stack. Starting with specs, op-node and op-geth changes.

These changes should be stacked together as PRs before merging, since end-to-end testing largely depends on various op-geth & engine API shapella features all being activated.

# Introduced L1 EIPs

*Source: https://blog.ethereum.org/2023/03/28/shapella-mainnet-announcement*

- [EIP-3651: Warm COINBASE](https://eips.ethereum.org/EIPS/eip-3651)
    - Affects EVM execution
- [EIP-3855: PUSH0 instruction](https://eips.ethereum.org/EIPS/eip-3855)
    - Affects EVM execution
- [EIP-3860: Limit and meter initcode](https://eips.ethereum.org/EIPS/eip-3860)
    - Affects EVM execution
    - Need to signal this to users, if they depend on pre-shanghai behavior (large initcode during `CREATE` / `CREATE2`)
- [EIP-4895: Beacon chain push withdrawals as operations](https://eips.ethereum.org/EIPS/eip-4895)
    - Affects block-format
    - Affects Engine API
    - Affects state-transition outside EVM
- [EIP-6049: Deprecate SELFDESTRUCT](https://eips.ethereum.org/EIPS/eip-6049)
    - Soft-change, need to signal this to users

# Upgrading

The shanghai hardfork affects both op-geth and op-node: the fork affects the Engine API.

We need to add a `shanghaiTime` rollup-config attribute, and can use the existing geth shanghai upgrade functionality.

# op-node changes

## Rollup Config `shanghaiTime` attribute

The rollup-config needs to be made aware of the shanghai hardfork time to enable some of the below changes to apply based on the hardfork timestamp: the Engine API usage is dependent on the hardfork timestamp.

This needs to be documented in the network-upgrades spec: https://github.com/ethereum-optimism/optimism/blob/develop/specs/network-upgrades.md

## SSZ ExecutionPayload encoding

The SSZ encoding is used in P2P gossip: we need to support the withdrawals list attribute here (even if empty, since empty is different than non-existent as in pre-shanghai blocks).

This change needs to be documented in the P2P spec: https://github.com/ethereum-optimism/optimism/blob/develop/specs/rollup-node-p2p.md

And requires a gossip-topic upgrade (the topic identifier is already versioned).

## L2 Eth-RPC client bindings

[...]

PR #7623, opened 2023-10-11
Migrate from V1 to V2 APIs

danyalprout:
**Description**
Migrate from the V1 to the V2 APIs for Shanghai support. The V2 APIs in are backwards compat:

* [GetPayload](https://github.com/ethereum-optimism/op-geth/blob/optimism/eth/catalyst/api.go#L437-L448)
* [NewPayload](https://github.com/ethereum-optimism/op-geth/blob/optimism/eth/catalyst/api.go#L465-L485)
* [ForkChoiceUpdated](https://github.com/ethereum-optimism/op-geth/blob/optimism/eth/catalyst/api.go#L174-L194)
    * [verifyPayloadAttributes](https://github.com/ethereum-optimism/op-geth/blob/optimism/eth/catalyst/api.go#L206-L218)

**Tests**

Updated current op-e2e tests to support the V2 methods.

**Metadata**
This PR is the first part of the work this [issue](https://github.com/ethereum-optimism/optimism/issues/6728). I'll finish up the P2P changes tomorrow and push a PR for those 🙏


specs/derivation.md
```diff
@@ -679,14 +679,14 @@ To interact with the engine, the [execution engine API][exec-engine] is used, wi
 
 [exec-engine]: exec-engine.md
```
tynes, line 681: Perhaps a note in here that a backwards compatible change was made would be nice to have

op-program/client/l2/engineapi/l2_engine_api.go
```diff
@@ -177,6 +177,31 @@ func (ea *L2EngineAPI) endBlock() (*types.Block, error) {
...
 
 func (ea *L2EngineAPI) GetPayloadV1(ctx context.Context, payloadId eth.PayloadID) (*eth.ExecutionPayload, error) {
+	return ea.getPayload(ctx, payloadId)
+}
+
+func (ea *L2EngineAPI) GetPayloadV2(ctx context.Context, payloadId eth.PayloadID) (*eth.ExecutionPayloadEnvelope, error) {
+	payload, err := ea.getPayload(ctx, payloadId)
+	if err != nil {
+		return nil, err
+	} else {
```
tynes: This doesn't necessarily need the `else` block but does the same thing, I think it reads better without it

op-program/client/l2/engineapi/l2_engine_api.go
```diff
@@ -177,6 +177,31 @@ func (ea *L2EngineAPI) endBlock() (*types.Block, error) {
...
+		return &eth.ExecutionPayloadEnvelope{ExecutionPayload: payload}, nil
+	}
+}
+
+func (ea *L2EngineAPI) ForkchoiceUpdatedV1(ctx context.Context, state *eth.ForkchoiceState, attr *eth.PayloadAttributes) (*eth.ForkchoiceUpdatedResult, error) {
+	return ea.forkchoiceUpdated(ctx, state, attr)
+}
+
```
tynes, line 205: Is there any difference between fork choice updated v1 and v2 at all?

specs/derivation.md
danyalprout: Great suggestion -- just added thanks!

op-program/client/l2/engineapi/l2_engine_api.go
danyalprout: I believe it's mostly validations on whether withdrawals are present or not -- depending on the current fork (which I've just ported over).

op-program/client/l2/engineapi/l2_engine_api.go
```diff
@@ -177,6 +178,82 @@ func (ea *L2EngineAPI) endBlock() (*types.Block, error) {
...
+func (ea *L2EngineAPI) config() *params.ChainConfig {
+	return ea.backend.Config()
+}
+
+func (ea *L2EngineAPI) ForkchoiceUpdatedV1(ctx context.Context, state *eth.ForkchoiceState, attr *eth.PayloadAttributes) (*eth.ForkchoiceUpdatedResult, error) {
+	if attr != nil {
+		if attr.Withdrawals != nil {
+			return STATUS_INVALID, engine.InvalidParams.With(errors.New("withdrawals not supported in V1"))
+		}
+		if ea.config().IsShanghai(ea.config().LondonBlock, uint64(attr.Timestamp)) {
```
trianglesphere, line 198: Is there any easy way to get the current block? As far as I can tell, the mix between the shanghai block & the london block will mean that this is always true or always false.

op-program/client/l2/engineapi/l2_engine_api.go
```diff
@@ -177,6 +178,82 @@ func (ea *L2EngineAPI) endBlock() (*types.Block, error) {
...
+	}
+
+	return ea.forkchoiceUpdated(ctx, state, attr)
+}
+
+func (ea *L2EngineAPI) verifyPayloadAttributes(attr *eth.PayloadAttributes) error {
+	c := ea.config()
+
+	// Verify withdrawals attribute for Shanghai.
+	if err := checkAttribute(c.IsShanghai, attr.Withdrawals != nil, c.LondonBlock, uint64(attr.Timestamp)); err != nil {
```
trianglesphere, line 220: ditto london block / activation

op-program/client/l2/engineapi/l2_engine_api.go
danyalprout: I believe it should work as expected, the underlying `IsShanghai` [method](https://github.com/ethereum-optimism/op-geth/blob/optimism/params/config.go#L582-L585) uses the second arg, in this case `attr.Timestamp` to detect if it's Shanghai or not. The first arg (the london block) is used just to see if we've forked past London.

For context, this code is ported from [op-geth](https://github.com/ethereum-optimism/op-geth/blob/optimism/eth/catalyst/api.go#L175-L182), my intention was to try and keep it as similar as possible. However happy to try and resolve the closest block to `attr.Timestamp` if you think it's worth it.

op-service/eth/types.go
```diff
@@ -228,6 +233,8 @@ type PayloadAttributes struct {
 	PrevRandao Bytes32 `json:"prevRandao"`
 	// suggested value for the coinbase field of the new payload
 	SuggestedFeeRecipient common.Address `json:"suggestedFeeRecipient"`
+	// Withdrawals to include into the block -- should be nil or empty depending on Shanghai enablement
+	Withdrawals *[]Withdrawal `json:"withdrawals,omitempty"`
```
tynes, line 237: Should this be `[]*Withdrawal`?

op-program/client/l2/engineapi/l2_engine_api.go
trianglesphere: My bad, I didn't realize that geth requires timestamp forks to be post london.

op-service/eth/types.go
trianglesphere: have we tested this pre & post shanghai? geth uses `[]*Withdrawal`, but does require it to be `nil` pre-shanghai and not `nil` post-shanghai

danyalprout: > have we tested this pre & post shanghai?

Yeah, I've tested this pre/post. Have some tests similar to the regolith ones that test withdrawals etc are returned post Shanghai. My intention is to submit those as part of the p2p changes PR I'm currently working on.

> geth uses []*Withdrawal, but does require it to be nil pre-shanghai and not nil post-shanghai

That's spot on. This is `*[]Withdrawal` so the omitempty tag works by default pre/post Shanghai. The intention is:

**Pre-Shanghai:**
Withdrawals = nil
`{}`

**Post-Shanghai:**
Withdrawals = []
`{"withdrawals": []}`

I believe doing this without the pointer would require custom serialization based on the canyon timestamp. Happy to do that in EngineClient if ya'll prefer.

trianglesphere: Sweet. Only question is if we want to do `*[]*Withdrawal`? I feel like `*[]Withdrawal` does work

trianglesphere (approved):

tynes (approved):

op-service/eth/types.go
tynes: @danyalprout What do you think of the above comment? If you don't want to change it, then you can resolve this comment and this PR will auto merge

op-service/eth/types.go
```diff
@@ -293,3 +300,11 @@ type SystemConfig struct {
 	GasLimit uint64 `json:"gasLimit"`
 	// More fields can be added for future SystemConfig versions.
 }
+
+// Withdrawal represents a validator withdrawal from the consensus layer.
+type Withdrawal struct {
```
roberto-bayardo, line 305: why do you need to paste this struct here rather than just using types.Withdrawal to use the one already defined by op-geth?

tynes: We did have a bug because of this previously when the withdrawals were added we didn't. know how to deserialize

op-service/eth/types.go
danyalprout: I don't have a strong opinion either way. My thinking is to leave it as `*[]Withdrawal` for now, as that's what I've tested. If I find any issues when doing the P2P layer, I'll change it as part of that work.

op-service/eth/types.go
danyalprout: My thinking was mostly to stick with the current convention of duplicating. e.g. [ExecutionPayload - optimism](https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/types.go#L128) and [ExecutableData - op-geth](https://github.com/ethereum-optimism/op-geth/blob/optimism/beacon/engine/types.go#L60).

I think decoupling the client/server types keeps it a bit safer, i.e. if we have some reflection somewhere and an additional field gets added and breaks some encoding.

PR #7627, opened 2023-10-11
feat(contracts-bedrock): add `DataAvailabilityChallenge` contract

alvrs:
TODO
- [ ] integrate into deploy script

cc @tynes


packages/contracts-bedrock/src/L1/DataAvailabilityChallenge.sol
```diff
@@ -0,0 +1,250 @@
...
+    function setBondSize(uint256 _bondSize) public onlyOwner {
+        bondSize = _bondSize;
+    }
+
+    /// @notice Initializes the contract.
+    /// @param _owner The owner of the contract.
+    /// @param _challengeWindow The block interval during which a commitment can be challenged.
+    /// @param _resolveWindow The block interval during which a challenge can be resolved.
+    /// @param _bondSize The amount required to post a challenge.
+    function initialize(address _owner, uint256 _challengeWindow, uint256 _resolveWindow, uint256 _bondSize) public initializer {
```
tynes: nit: can you move `initialize` to be directly under the constructor?

packages/contracts-bedrock/src/L1/DataAvailabilityChallenge.sol
```diff
@@ -0,0 +1,250 @@
...
+        __Ownable_init();
+        setChallengeWindow(_challengeWindow);
+        setResolveWindow(_resolveWindow);
+        setBondSize(_bondSize);
+        _transferOwnership(_owner);
+    }
+
+    /// @notice Post a bond as prerequisite for challenging a commitment.
+    function deposit() external payable {
```
tynes: Can you add a receive function that calls `deposit`? So you can just send ether directly to the contract without needing an ABI

packages/contracts-bedrock/src/L1/DataAvailabilityChallenge.sol
```diff
@@ -0,0 +1,250 @@
...
+        balances[_msgSender()] += msg.value;
```
tynes: Is `_msgSender` coming from `OwnableUpgradable`? We usually just use `msg.sender`, i know that fn is supposed to abstract away the sender for smart contract wallets or something, I don't think we necessarily need it unless you believe there is a good reason to have it

tynes:
The deploy script will need to have 3 functions added
- `deployDataAvailabilityChallengeProxy`
- `deployDataAvailabilityChallenge`
- `initializeDataAvailabilityChallenge`

We will need to add the config values to the `deploy-config`, there is a smart contract that reads the config and also there are JSON files in the `deploy-config` directory per network. This may also trigger CI to break on the deploy config checks where the values will need to be added to `op-chain-ops/genesis#DeployConfig`

packages/contracts-bedrock/src/L1/DataAvailabilityChallenge.sol
alvrs: yep it's from `OwnableUpgradable` - no strong opinion though, happy to use `msg.sender` for consistency if that's what you usually use

tynes: Can we use `msg.sender`?

alvrs: yep already reverted the `msgSender()` change!

op-chain-ops/genesis/config.go
```diff
@@ -204,9 +204,15 @@ type DeployConfig struct {
 	// RequiredProtocolVersion indicates the protocol version that
 	// nodes are required to adopt, to stay in sync with the network.
 	RequiredProtocolVersion params.ProtocolVersion `json:"requiredProtocolVersion"`
-	// RequiredProtocolVersion indicates the protocol version that
+	// RecommendedProtocolVersion indicates the protocol version that
```
alvrs, line 207: unrelated to this PR but noticed this comment is was wrong

op-bindings/bindings/mips_more.go
alvrs, line 1: unclear why this and `preimageoracle_more.go` changed, should i remove these changes?

tynes: Sometimes different contracts cause the `more` files to change in seemingly unrelated ways, its ok if this happens

tynes:
Looks like the remaining thing here is the gas snapshot with `pnpm gas-snapshot`

tynes:
> Looks like the remaining thing here is the gas snapshot with `pnpm gas-snapshot`

It could be that different versions of forge create different gas snapshots, the version we use is in `.foundryrc` in the root of the repo

packages/contracts-bedrock/src/L1/DataAvailabilityChallenge.sol
```diff
@@ -0,0 +1,266 @@
...
+        challengeWindow = _challengeWindow;
+    }
+
+    /// @notice Sets the resolve window.
+    /// @param _resolveWindow The block interval during which a challenge can be resolved.
+    function setResolveWindow(uint256 _resolveWindow) public onlyOwner {
+        resolveWindow = _resolveWindow;
+    }
+
+    /// @notice Sets the bond size.
```
tynes, line 138: We should probably emit events in these setters but fine with doing that in a follow up PR

packages/contracts-bedrock/src/L1/DataAvailabilityChallenge.sol
```diff
@@ -0,0 +1,266 @@
+// SPDX-License-Identifier: MIT
```
clabby, line 1: Alot of this looks to duplicate the logic from the dispute protocol (challenge status, the rough interface, etc.) - is there any specific reason that this isn't an `IDisputeGame` that can be integrated into the `DisputeGameFactory`?

packages/contracts-bedrock/src/L1/DataAvailabilityChallenge.sol
alvrs: done in https://github.com/ethereum-optimism/optimism/pull/7627/commits/247a49b8f3dd6067431ecccb21cea31cd61fbaaa

packages/contracts-bedrock/src/L1/DataAvailabilityChallenge.sol
alvrs: Didn't know the `IDisputeGame` interface existed! Agree there is some overlap. But it seems like the `DisputeGameFactory` creates a new `DisputeGame` for every dispute, which is a different approach than this challenge contract. Since challenging a hash and resolving a challenge by providing the pre-image of the hash are both simple atomic operations, they can be managed in a single contract and it would feel like an overkill to create a new `DataAvailabilityDispute` contract for each challenge.

PR #7628, opened 2023-10-11
contracts-bedrock: remove contracts-ts generator

tynes:
**Description**

This should probably be a temporary change but the tooling around `contracts-ts` is broken. This removes the auto generation of the `contracts-ts` package during the bindings command in `contracts-bedrock`.

cc @roninjin10


roninjin10:
Fix is here https://github.com/ethereum-optimism/optimism/pull/7631

PR #7629, opened 2023-10-11
feat(ctb): Deeper local key contexts

clabby:
## Overview

Introduces further contextualization of local keys in `MIPS.sol` and the `PreimageOracle` contract for use in output bisection.

### Rationale

When output bisection is added, we'll need the ability for a single `FaultDisputeGame` contract to post multiple sets of local keys to the `PreimageOracle`. The solution here is to add another element of contextualization. With this, the `FaultDisputeGame` can compute which L2 block number is being disputed and provide it as the `localContext` to the `step` function internally. The caller of the `FaultDisputeGame`'s step function will not have to provide the local context, however they will have to provide it for the `addLocalData` function.


clabby:
* develop
  * **PR #7629** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7629" target="_blank">Graphite</a>  👈
    * **PR #7630** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7630" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7629?utm_source=stack-comment).

packages/contracts-bedrock/src/cannon/MIPS.sol
```diff
@@ -621,7 +622,11 @@ contract MIPS {
 
     /// @notice Executes a single step of the vm.
     ///         Will revert if any required input state is missing.
-    function step(bytes calldata stateData, bytes calldata proof) public returns (bytes32) {
+    /// @param _stateData The encoded state witness data.
+    /// @param _proof The encoded proof data for leaves within the MIPS VM's memory.
+    /// @param _localContext The local key context for the preimage oracle. Optional, can be set as a constant
+    ///                     if the caller only requires one set of local keys.
```
refcell: very small nit
```suggestion
    /// @param _localContext The local key context for the preimage oracle. Optional, can be set as a constant
    ///                      if the caller only requires one set of local keys.
```

packages/contracts-bedrock/src/cannon/interfaces/IPreimageOracle.sol
```diff
@@ -13,6 +13,8 @@ interface IPreimageOracle {
 
     /// @notice Loads of local data part into the preimage oracle.
     /// @param _ident The identifier of the local data.
+    /// @param _localContext The local key context for the preimage oracle. Optional, can be set as a constant
+    ///                     if the caller only requires one set of local keys.
```
refcell: ibid very small nit
```suggestion
    /// @param _localContext The local key context for the preimage oracle. Optionally, can be set as a constant
    ///                      if the caller only requires one set of local keys.
```

packages/contracts-bedrock/src/dispute/interfaces/IBigStepper.sol
```diff
@@ -29,8 +29,16 @@ interface IBigStepper {
     ///         hash.
     /// @param _stateData The preimage of the prestate hash.
     /// @param _proof A proof for the inclusion of the prestate's memory in the merkle tree.
+    /// @param _localContext The local key context for the preimage oracle. Optional, can be set as a constant
+    ///                     if the caller only requires one set of local keys.
```
refcell: ibid

packages/contracts-bedrock/src/dispute/interfaces/IFaultDisputeGame.sol
```diff
@@ -70,8 +70,10 @@ interface IFaultDisputeGame is IDisputeGame {
 
     /// @notice Posts the requested local data to the VM's `PreimageOralce`.
     /// @param _ident The local identifier of the data to post.
+    /// @param _l2BlockNumber The L2 block number being disputed. This serves as the local context for the
+    ///                      `PreimageOracle` key.
```
refcell: ibid

refcell (commented):
As I understand it, this looks good!

cannon/mipsevm/evm.go
```diff
@@ -24,9 +24,9 @@ import (
 )
 
 var (
-	StepBytes4                      = crypto.Keccak256([]byte("step(bytes,bytes)"))[:4]
+	StepBytes4                      = crypto.Keccak256([]byte("step(bytes,bytes,uint256)"))[:4]
 	LoadKeccak256PreimagePartBytes4 = crypto.Keccak256([]byte("loadKeccak256PreimagePart(uint256,bytes)"))[:4]
-	LoadLocalDataBytes4             = crypto.Keccak256([]byte("loadLocalData(uint256,bytes32,uint256,uint256)"))[:4]
+	LoadLocalDataBytes4             = crypto.Keccak256([]byte("loadLocalData(uint256,uint256,bytes32,uint256,uint256)"))[:4]
```
tynes: What do you think about pulling this from the binding instead of hardcoding it?

`bindings.Mips.GetAbi().Methods["StepBytes4"].ID()`

cannon/mipsevm/evm_test.go
```diff
@@ -76,13 +76,13 @@ func (m *MIPSEVM) Step(t *testing.T, stepWitness *StepWitness) []byte {
 
 	if stepWitness.HasPreimage() {
 		t.Logf("reading preimage key %x at offset %d", stepWitness.PreimageKey, stepWitness.PreimageOffset)
-		poInput, err := stepWitness.EncodePreimageOracleInput()
+		poInput, err := stepWitness.EncodePreimageOracleInput(0)
```
tynes, line 79: Should local context be its own type? What does 0 mean? ie make it human readable

clabby: I've removed these in the PR above - #7630. They're not consumed by anything else other than the tests, so it's just wasted space on disk and wasted cycles during the cannon run.

Will add a custom type for the tests.

refcell (approved):
lgtm

cannon/mipsevm/evm.go
```diff
@@ -16,17 +16,16 @@ import (
...
 )
 
 var (
-	StepBytes4                      = crypto.Keccak256([]byte("step(bytes,bytes)"))[:4]
-	LoadKeccak256PreimagePartBytes4 = crypto.Keccak256([]byte("loadKeccak256PreimagePart(uint256,bytes)"))[:4]
-	LoadLocalDataBytes4             = crypto.Keccak256([]byte("loadLocalData(uint256,bytes32,uint256,uint256)"))[:4]
+	StepBytes4                      []byte
+	LoadKeccak256PreimagePartBytes4 []byte
+	LoadLocalDataBytes4             []byte
 )
```
tynes, line 29: These are part of the public API so therefore they should be populated in an `init` function instead of a function that is called internal to another function

op-challenger/game/fault/trace/cannon/updater_test.go
```diff
@@ -104,12 +104,13 @@ func TestCannonUpdater_BuildLocalOracleData(t *testing.T) {
 	txData, err := updater.BuildLocalOracleData(oracleData)
 	require.NoError(t, err)
 
-	var addLocalDataBytes4 = crypto.Keccak256([]byte("addLocalData(uint256,uint256)"))[:4]
+	var addLocalDataBytes4 = crypto.Keccak256([]byte("addLocalData(uint256,uint256,uint256)"))[:4]
```
tynes: Same thing with `.GetAbi()` here

tynes:
Looks good to me altho i don't have the right context to give an approval

cannon/mipsevm/witness.go
```diff
@@ -63,6 +66,7 @@ func (wit *StepWitness) EncodePreimageOracleInput() ([]byte, error) {
 		var input []byte
 		input = append(input, LoadLocalDataBytes4...)
 		input = append(input, wit.PreimageKey[:]...)
+		input = append(input, uint32ToBytes32(uint32(localContext))...) // local context in bytes
```
ajsutton: Why are we using a `uint64` that we then case to `uint32`? Seems like we should declare it as `uint32` for accuracy.  Converting from `big.Int` will be slightly more difficult but that's where we should be bounds checking the conversion anyway.

packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol
```diff
@@ -174,7 +174,10 @@ contract FaultDisputeGame is IFaultDisputeGame, Clone, ISemver {
         // SAFETY:    While the `attack` path does not need an extra check for the post
         //            state's depth in relation to the parent, we don't need another
         //            branch because (n - n) % 2 == 0.
-        bool validStep = VM.step(_stateData, _proof) == Claim.unwrap(postState.claim);
+        // TODO(clabby): Once output bisection is implemented, the local context will no longer
```
ajsutton: Can we reference an issue number for the TODO rather than just you - then we know it should have been done if that issue is closed.

ajsutton (approved):
LGTM. I'm not strongly opinionated on the uint32 vs uint64 type change - we're already doing a bunch of unsafe casts because these numbers should always be fairly small but get pushed up to big types due to the RPC APIs that always return `big.Int`.

cannon/mipsevm/evm.go
```diff
@@ -16,19 +16,33 @@ import (
...
+		panic(fmt.Errorf("failed to load MIPS ABI: %w", err))
+	}
+	StepBytes4 = mipsAbi.Methods["step"].ID[:4]
+
+	preimageAbi, err := bindings.PreimageOracleMetaData.GetAbi()
+	if err != nil {
+		panic(fmt.Errorf("failed to load pre-image oracle ABI: %w", err))
+	}
+	LoadKeccak256PreimagePartBytes4 = preimageAbi.Methods["loadKeccak256PreimagePart"].ID[:4]
+	LoadLocalDataBytes4 = preimageAbi.Methods["loadLocalData"].ID[:4]
```
Inphi, line 43: nit: It's better to manually compute the function selector because it's more robust against changes to the ABI. Relying on the bindings means when we update the ABI, we may accidentally provide invalid inputs - worst case the call doesn't revert! But prior to this change the function call would simply revert with the incorrect selector.

Then again, since we're pulling in the bindings, we can use `abi.Pack` when constructing inputs for safety. This can be done later on another PR.

packages/contracts-bedrock/src/cannon/PreimageKeyLib.sol
```diff
@@ -7,29 +7,36 @@ library PreimageKeyLib {
...
             // the [1, 32) bytes in this value.
             key_ := or(shl(248, 1), and(_ident, not(shl(248, 0xFF))))
         }
-        key_ = localize(key_);
+        key_ = localize(key_, _localContext);
     }
 
     /// @notice Localizes a given local data key for the caller's context.
     /// @dev The localization operation is defined as:
     ///      localize(k) = H(k .. sender) & ~(0xFF << 248) | (0x01 << 248)
```
Inphi: nit: Comments need to be updated.

Inphi (approved):
Looks good. Only thing note-worthy is the local context width that Adrian pointed out. It's fine for now as uint32, but we should port to `big.Int` (or uint.256) at some point so it's consistent with the width of block numbers.

cannon/mipsevm/evm.go
clabby: For sure. Tracking in https://github.com/ethereum-optimism/client-pod/issues/102

PR #7632, opened 2023-10-11
Add Delayed Vetoable to deploy scripts

maurelian:
**Changes to `SuperchainConfig`:**

The `systemOwner` variable is now read directly from the `PROXY_OWNER_ADDRESS` slot in storage, rather than having a separate storage slot dedicated to it. Although this is breaking the separation of concerns between proxy and implementation, this change is important to avoid redundant config, and potential configuration errors. 

**Changes to `DelayedVetoable`**:

- Refactored `DelayedVetoable` to read from `SuperchainConfig`.
- Replaced immutable reads with internal getters.
- Added separate internal and external getters to properly support the `readOrHandle` modifier

Note: the goal of making `DelayedVetoable` an easily inheritable and reusable contract was abandoned, as it would have significantly impacted the maintainability and readability of the contract to keep both the pre-existing immutable vars, while overriding the internal methods to read from the `SuperchainConfig`.

**Changes to deployment scripts:**

- Deploys a Safe for the Initiator (security council) and Vetoer (foundation)
- Deploys a new `DelayedVetoable` in as admin of the `SuperchainConfig` 
- Deploys a new `DelayedVetoable` as owner of the `ProxyAdmin`
- Renamed config vars:
	- `finalSystemOwner` → `systemConfigOwner`
	- `portalGuardian` → `superchainConfigGuardian`


packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -31,11 +30,6 @@ contract SuperchainConfig is Initializable, ISemver {
         REMOVE_SEQUENCER
     }
 
-    /// @notice The address of the systemOwner who may trigger an upgrade or change to critical config values.
-    ///         This will be a DelayedVetoable contract.
-    ///         It can only be modified by an upgrade.
-    bytes32 public constant SYSTEM_OWNER_SLOT = bytes32(uint256(keccak256("superchainConfig.systemowner")) - 1);
```
tynes, line 37: Hmm I don't love leaking the proxy admin abstraction into the implementation here

maurelian: I 100% get where you're coming from with that, it felt a bit dirty to me too. I'm kind of on the fence, but in the interest of finding the best solution, here's my steel man argument for this approach:

The requirement is that both [upgrades](https://www.notion.so/oplabs/External-Security-Council-Product-Requirements-FAQ-10425231ed5a47138c388b8705c313f1?pvs=4#aebe5405146f48dd893ec4b344cb5401), and the [removal of a sequencer](https://www.notion.so/oplabs/External-Security-Council-Product-Requirements-FAQ-10425231ed5a47138c388b8705c313f1?pvs=4#98a95e1bb0b243d980d3b4edb4fbe9f2) must be  enacted by Optimism Governance, and subject to a delay.

The implementation can satisfy that requirement directly by reading the same address from storage, the alternative is to have:

- 2 separate values in the config which we need to assert are identical
- 2 separate storage values that need to be kept in sync

So yes, in one sense the proxy abstraction is leaking into the implementation. But in another sense, the intended behavior is being implemented directly, rather than enforced with additional checks.

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -416,6 +425,83 @@ contract Deploy is Deployer {
         console.log("SuperchainConfig deployed at %s", address(superchainConfig));
     }
 
+    /// @notice Deploy the DelayedVetoable contract for the SuperchainConfig
+    ///         Note: Because the getters on DelayedVetoable must be read by pranking as the zero
+    ///         address, this function does not use the broadcast modifier, and instead starts and
+    ///         stops broadcasting, then starts and stops pranking, within the function body.
+    function deployDelayedVetoableForSuperchain() public {
```
tynes: There is a `broadcast` modifier - does that not work because of the transparent nature based on `address(0)`?

maurelian: Correct, you apparently can't prank during a broadcast.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: I'm going to make sure this is documented in the PR description itself. Resolving.

packages/contracts-bedrock/deploy-config/getting-started.json
```diff
@@ -1,6 +1,6 @@
 {
-  "finalSystemOwner": "ADMIN",
-  "portalGuardian": "ADMIN",
+  "systemConfigOwner": "ADMIN",
+  "superchainConfigGuardian": "ADMIN",
```
tynes, line 4: We are going to need to add something to the op stack tutorial that says to check out an old commit or update it when this gets merged

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -71,45 +73,47 @@ contract Deploy is Deployer {
     /// @notice Deploy all of the L1 contracts
     function run() public {
         console.log("Deploying a fresh OP Stack including SuperchainConfig");
-        address systemOwnerSafe = deploySafe();
-        address supConfProxy = setupSuperchainConfig();
-        setupOpChain({ supConfProxy: supConfProxy, systemOwnerSafe: systemOwnerSafe });
+
+        address supConfInitiator;
```
tynes, line 77: This should be like:

```solidity
address supConfIgInitiator = cfg.superchainCOnfigInitiator();
if (supConfigInitiator == address(0) {
  supconfigInitator = deploySafe()
}
```

Otherwise supconfInitiator could be `address(0)` when passed to `setupOpChain`

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -191,7 +199,20 @@ contract Deploy is Deployer {
...
+        initializeL1StandardBridge();
+        initializeL1ERC721Bridge();
+        initializeOptimismMintableERC20Factory();
+        initializeL1CrossDomainMessenger();
+        initializeL2OutputOracle();
+        initializeOptimismPortal();
+        initializeProtocolVersions();
+    }
     // @notice Gets the address of the SafeProxyFactory and Safe singleton for use in deploying a new GnosisSafe.
+
```
tynes, line 222: extra newline?

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -209,7 +230,7 @@ contract Deploy is Deployer {
     }
 
     /// @notice Deploy the Safe
-    function deploySafe() public broadcast returns (address addr_) {
+    function deploySafe(string memory safeName) public broadcast returns (address addr_) {
```
tynes: nit: prefixed arg

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -416,6 +438,89 @@ contract Deploy is Deployer {
...
+        vm.startBroadcast();
+        SuperchainConfig superchainConfigProxy = SuperchainConfig(mustGetAddress("SuperchainConfigProxy"));
+        DelayedVetoable delayedVetoable = new DelayedVetoable({
+                superchainConfig_: superchainConfigProxy,
+                target_: address(superchainConfigProxy)
+            });
+        vm.stopBroadcast();
+
+        vm.startPrank(address(0));
+        require(delayedVetoable.superchainConfig() == address(superchainConfigProxy));
```
tynes: Perhaps we could modularize the 2 deployments for delayed vetoable with an internal function to dedupe code

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -661,10 +774,11 @@ contract Deploy is Deployer {
                     abi.encodeCall(
                         SuperchainConfig.initialize,
                         (
-                            cfg.finalSystemOwner(), // systemOwner
-                            cfg.superchainConfigInitiator(), // initiator
+                            // todo: fix this
```
tynes: Wondering about this todo

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -78,39 +73,61 @@ contract DelayedVetoable is ISemver {
     string public constant version = "1.0.0";
```
tynes: nit: semver bump

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -42,17 +43,11 @@ contract DelayedVetoable is ISemver {
...
-    /// @notice The address that can veto a call.
-    address internal immutable VETOER;
-
-    /// @notice The address that can initiate a call.
-    address internal immutable INITIATOR;
-
-    /// @notice The delay which will be set after the initial system deployment is completed.
-    uint256 internal immutable OPERATING_DELAY;
+    /// @notice The superchain config contract.
+    SuperchainConfig internal _superchainConfig;
```
tynes: any reason this has to be in storage?

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -91,7 +85,6 @@ contract SuperchainConfig is Initializable, ISemver {
     /// @notice Constructs the SuperchainConfig contract.
```
tynes, line 85: nit: semver bump

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -622,14 +725,22 @@ contract Deploy is Deployer {
...
+    ///         DelayedVetoable contract .
+    function _upgradeAndCallViaSafe(
+        Safe _safe,
+        address _proxy,
+        address _implementation,
+        bytes memory _innerCallData
+    )
+        internal
+    {
+        address delayedVetoableForOpChain = mustGetAddress("DelayedVetoableForOpChain");
```
maurelian: I think this is broken. 
It shouldn't hard code the delayed vetoable contract to use. This'll require a minor refactor.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
```diff
@@ -237,13 +227,6 @@ contract SuperchainConfig is Initializable, ISemver {
         emit ConfigUpdate(UpdateType.REMOVE_SEQUENCER, abi.encode(_sequencer));
     }
 
-    /// @notice Sets the system owner address.
-    /// @param _systemOwner The new system owner address.
-    function _setSystemOwner(address _systemOwner) internal {
-        Storage.setAddress(SYSTEM_OWNER_SLOT, _systemOwner);
-        emit ConfigUpdate(UpdateType.SYSTEM_OWNER, abi.encode(_systemOwner));
```
tynes, line 244: Interesting, now we lose this event, i think that is fine

op-chain-ops/genesis/config.go
```diff
@@ -217,13 +215,19 @@ type DeployConfig struct {
...
 	// SuperchainConfigDelay represents the time after which an upgrade can occur ifit
 	SuperchainConfigDelay uint64 `json:"superchainConfigDelay"`
 	// SuperchainConfigMaxPause represents the time after which an upgrade can occur ifit
 	SuperchainConfigMaxPause uint64 `json:"superchainConfigMaxPause"`
 	// SequencerKeys represents the set of allowed Sequencer key pairs (batcher hash and unsafe
 	// block signer).
 	SuperchainConfigSequencerKeyPairs []SequencerKeyPair `json:"superchainConfigSequencerKeyPairs"`
+	// Legacy config values.
+	// These are no longer used, but are necessary for parsing legacy deploy-configs
+	PortalGuardian   common.Address `json:"portalGuardian"`
```
tynes, line 229: We will want to remove these longer term

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: My bad. 
The thing that needed to be fixed was the fact that if the initiator address is set in the config, then it won't be in the state when this script is run on a devnet. 

The fix is the code in `setupSuperchain` that deploys a new Safe if one is not provided.

I'll remove this todo.

op-chain-ops/genesis/config.go
maurelian: Agreed... I did not feel good about this, but I didn't really see another way available here.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: As is the only place it was emitted was in the initializer.

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: No, good catch. It can be made immutable. I'll fix that.

Same goes for the `TARGET` var above. It looks like I put it into storage while moving the other immutables (`VETOER`, `INITIATATOR`, etc.) to the SuperchainConfig.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: `DelayedVetoable` needs a bump, yes, but this contract isn't on develop yet, so I think it should be OK at `1.0.0`, right?

op-chain-ops/genesis/config.go
tynes: We should just delete these imo and bite the bullet

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -986,13 +1114,29 @@ contract Deploy is Deployer {
...
+            _target: mustGetAddress("DelayedVetoableForSuperchain"),
+            _data: hex""
+        });
+
+        _callViaSafe({
+            _safe: Safe(mustGetAddress("SuperchainConfigInitiatorSafe")),
+            _target: mustGetAddress("DelayedVetoableForOpChain"),
+            _data: hex""
+        });
+        console.log("Delay activated on DelayedVetoable contracts");
```
tynes: What do you think about 2 loglines and include the addresses in the log?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -42,17 +43,11 @@ contract DelayedVetoable is ISemver {
     /// @param data The data forwarded to the target.
```
tynes, line 43: Just a thought: this could be made into a clones-with-immutable-args factory + proxy pattern longer term, it would be cheaper to deploy minimal proxies that delegatecall a single impl, we shouldn't do this now but just something to think about

packages/contracts-bedrock/test/CommonTest.t.sol
```diff
@@ -97,11 +98,12 @@ contract SuperchainConfig_Initializer is CommonTest {
...
     event ConfigUpdate(SuperchainConfig.UpdateType indexed updateType, bytes data);
 
-    address systemOwner = makeAddr("SystemOwner");
+    address systemOwner = multisig;
     address initiator = makeAddr("initiator");
     address vetoer = makeAddr("vetoer");
     address guardian = makeAddr("guardian");
-    uint256 delay = 100;
+    uint256 operatingDelay = 14 days;
+    // uint256 delay = operatingDelay;
```
tynes: nit: dead code

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: I have a strong feeling there is room for improvement wrt the operations of this contract.

op-chain-ops/genesis/config.go
maurelian: We agree to live with this for the moment.

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Fixed this.

packages/contracts-bedrock/src/L1/SuperchainConfig.sol
maurelian: bumped it anyways

packages/contracts-bedrock/deploy-config/getting-started.json
maurelian: Surfaced this in #7656.

op-chain-ops/genesis/config.go
```diff
@@ -217,13 +215,19 @@ type DeployConfig struct {
...
+	FinalSystemOwner common.Address `json:"finalSystemOwner"`
```
tynes, line 230: Thinking about this again, different values are used in the go code vs the solidity code, I am worried that if we leave both it will result in bugs. I would like to have these deduped before going to `develop`

packages/contracts-bedrock/test/CommonTest.t.sol
```diff
@@ -125,14 +126,60 @@ contract SuperchainConfig_Initializer is CommonTest {
...
+    event Forwarded(bytes32 indexed callHash, bytes data);
+    event Vetoed(bytes32 indexed callHash, bytes data);
+
+    address target = makeAddr("target");
+    DelayedVetoable delayedVetoable;
+
+    function setUp() public override {
+        super.setUp();
+        delayedVetoable = new DelayedVetoable({
+            superchainConfig_: supConf,
```
tynes: Why are these args postfixed instead of prefixed?

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -133,18 +150,18 @@ contract DelayedVetoable is ISemver {
     function _handleCall() internal {
         // The initiator and vetoer activate the delay by passing in null data.
```
tynes, line 154: nit: can cache calls onto the stack instead of making the same calls multiple times

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
```diff
@@ -74,43 +69,65 @@ contract DelayedVetoable is ISemver {
...
-    /// @param target_ Address of the target.
-    /// @param operatingDelay_ Time to delay when the system is operational.
-    constructor(address vetoer_, address initiator_, address target_, uint256 operatingDelay_) {
-        // Note that the _delay value is not set here. Having an initial delay of 0 is helpful
-        // during the deployment of a new system.
-        VETOER = vetoer_;
-        INITIATOR = initiator_;
+    /// @param superchainConfig_ Address of the superchain config contract.
+    /// @param target_ Address of the target contract.
+    constructor(SuperchainConfig superchainConfig_, address target_) {
```
tynes: Why are these postfixed instead of prefixed?

op-chain-ops/upgrades/l1.go
```diff
@@ -420,7 +420,7 @@ func SystemConfig(batch *safe.Batch, implementations superchain.ImplementationLi
 	}
 
 	calldata, err := initialize.Inputs.PackValues([]any{
-		finalSystemOwner,
```
tynes, line 423: The base that this is implemented on has bugs in this file, should be a simple rebase when its ready

packages/contracts-bedrock/test/CommonTest.t.sol
maurelian: Probably autosuggested tbh.
Will fix, but it's crazy that we don't have linting for this.

tynes (approved):

packages/contracts-bedrock/src/L1/DelayedVetoable.sol
maurelian: Good call.
This pattern was safer and more iterative during refactoring, but caching finishes the job.

packages/contracts-bedrock/test/CommonTest.t.sol
maurelian: OK, just for posterity, the reason for the postfix was that:
1. the internal getter function `_target` already had the prefix, so there was a naming collision.
2. I wanted to be consistent with the arg names on this function.

But I've resolved that differently by renaming it to `_targetContract`.

op-chain-ops/genesis/config.go
maurelian: I'm not sure I understand, but I've tracked this issue in the PR to `develop`. 

https://github.com/ethereum-optimism/optimism/pull/7656/files#r1358504501

op-chain-ops/upgrades/l1.go
maurelian: I'd like to do a rebase ASAP, appreciate the heads up.

PR #7642, opened 2023-10-12
feat(ci): TODO GitHub issue checker

clabby:
## Overview

Adds a script to `ops` that greps through the codebase for TODO comments that reference GitHub issues. If any of the issues
referenced are closed, an error is thrown, and the script exits with a non-zero exit code and an error asking the user
to remove the stale TODO comment.

The script also has several options:
* `--verbose`: Prints out warnings and a summary for issues that were either not found as well as TODO comments with invalid formats.
* `--strict`: Fails if any TODO comments are found that fit an invalid format.

### Accepted `TODO` formats
* `TODO(<issue_number>): <description>` (Default org & repo: "ethereum-optimism/optimism")
* `TODO(repo#<issue_number>): <description>` (Default org "ethereum-optimism")
* `TODO(org/repo#<issue_number>): <description>`

### Examples
Screenshot 2023-10-12 at 3 58 20 AM
Screenshot 2023-10-12 at 3 58 38 AM

### TODO
- [x] Get a PAT secret for the `ethereum-optimism` organization in order for CI to be able to reference issues in `client-pod`.

### DIFF
--- a/ops/scripts/todo-checker.sh
+++ b/ops/scripts/todo-checker.sh
@@ -1,8 +1,15 @@
 #!/bin/bash
 
+set -uo pipefail
+
+# Flags
+FAIL_INVALID_FMT=false
+VERBOSE=false
+
 # Github API access token (Optional - necessary for private repositories.)
-TOKEN=""
-if [[ $TOKEN != "" ]]; then
+GH_API_TOKEN=""
+AUTH=""
+if [[ $GH_API_TOKEN != "" ]]; then
     AUTH="Authorization: token $TOKEN"
 fi
 
@@ -25,11 +32,10 @@ CYAN='\033[0;36m'
 PURPLE='\033[0;35m'
 NC='\033[0m' # No Color
 
-# Toggle strict mode; Will fail if any TODOs are found that don't match the expected
-# formats:
-# * TODO(<issue_number>): <description> (Default org & repo: "ethereum-optimism/client-pod")
-# * TODO(repo#<issue_number>): <description> (Default org "ethereum-optimism")
-# * TODO(org/repo#<issue_number>): <description>
+# Parse flags
+#
+# `--strict`: Toggle strict mode; Will fail if any TODOs are found that don't match the expected
+# `--verbose`: Toggle verbose mode; Will print out details about each TODO
 for arg in "$@"; do
   case $arg in
     --strict)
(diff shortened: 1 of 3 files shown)


clabby:
* develop
  * **PR #7642** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7642" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/optimism/7642?utm_source=stack-comment).

ajsutton (commented):
Nice.  :)

ajsutton:
Somehow I lost the comment I was actually trying to add...

I'd suggest we run this only on a schedule and not with every PR.  Otherwise closing an issue could prevent unrelated PRs from failing to merge which would be very disruptive.  We can either add a slack notification on failures or possibly better, just reopen the issue. :)

clabby:
> Somehow I lost the comment I was actually trying to add...
> 
> I'd suggest we run this only on a schedule and not with every PR. Otherwise closing an issue could prevent unrelated PRs from failing to merge which would be very disruptive. We can either add a slack notification on failures or possibly better, just reopen the issue. :)

Added this - great point 👍

ops/scripts/todo-checker.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+
```
ajsutton, line 2: Typically I'd suggest using `set -euo pipefail` for bash scripts. By default bash will just carry on even after some command fails which can lead to pretty unexpected results and be hard to debug.

I think given we aren't using `--fail` with `curl` that it should work out ok even for issues that can't be found. Not sure if `rg` will cause issues though so maybe this is a script where its not worth it...

or maybe could just go with `set -uo pipefail` so you get failures for undeclared variables (catches a lot of typos) and sensible error codes when using pipes but it won't abort the script on a non-zero exit code.

ajsutton (approved):
LGTM. Can try to tighten up the bash mode for a bit more safety but not worth stressing about too much given this is just informational.

clabby:
Adding https://github.com/ethereum-optimism/optimism/labels/M-do-not-merge. Need a GH PAT secret in CircleCI to read the private issues in the `client-pod` repo before we move everything into the public.

sebastianst:
@clabby I added a PAT that should have the right permissions to CCI's env var `CI_TODO_CHECKER_PAT`. Please give it a try.

PR #7662, opened 2023-10-13
op-challenger: Pass through full game data when creating player instead of just address

ajsutton:
**Description**

Passes the full `FaultDisputeGame` (now `GameData`) struct further through the pipeline so that the game type is available when creating the actual player.

This is the first step towards selecting the game to create based on game type and supporting multiple game types in one instance, and ultimately to being less tightly coupled to the contract bindings because we'll be able to create the bindings in one place rather than scattered through the codebase.

The game loader also got moved into a `factory` package - helps avoid dependency cycles and is just a bit tidier.

**Tests**

No change to functionality yet, just updated the tests to use the new types.

**Metadata**

- Part of https://github.com/ethereum-optimism/client-pod/issues/100

### DIFF
--- a/op-challenger/game/types/types.go
+++ b/op-challenger/game/types/types.go
@@ -2,6 +2,8 @@ package types
 
 import (
 	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
 )
 
 type GameStatus uint8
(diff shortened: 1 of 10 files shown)


op-challenger/game/types/types.go
```diff
@@ -33,3 +35,9 @@ func GameStatusFromUint8(i uint8) (GameStatus, error) {
 	}
 	return GameStatus(i), nil
 }
+
+type GameData struct {
```
Inphi: nit: GameMetadata could be a better name.

Inphi (approved):

op-challenger/game/types/types.go
ajsutton: Yeah good call, I'll change that.

PR #7663, opened 2023-10-13
op-challenger: Introduce game registry

ajsutton:
**Description**

Introduces a game type registry that is used to create the appropriate game player based on the game type. Currently the CLI/config still only allows the user to specify one supported trace type at a time, but this sets up the code to support multiple game types once the cli and config are updated.

This also means that the challenger will now log an error if it finds games with a type it doesn't support rather than attempting to play the game anyway.

**Tests**

Added unit tests.

**Additional context**

Builds on #7662

**Metadata**

- Part of https://github.com/ethereum-optimism/client-pod/issues/100

### DIFF
--- a/op-challenger/game/fault/register.go
+++ b/op-challenger/game/fault/register.go
@@ -37,7 +37,7 @@ func RegisterGameTypes(
 ) {
 	switch cfg.TraceType {
 	case config.TraceTypeCannon:
-		registerFaultGameType(ctx, registry, cannonGameType, logger, m, cfg, txMgr, client, func(addr common.Address, gameDepth uint64, dir string) (faultTypes.TraceProvider, faultTypes.OracleUpdater, error) {
+		resourceCreator := func(addr common.Address, gameDepth uint64, dir string) (faultTypes.TraceProvider, faultTypes.OracleUpdater, error) {
 			provider, err := cannon.NewTraceProvider(ctx, logger, m, cfg, client, dir, addr, gameDepth)
 			if err != nil {
 				return nil, nil, fmt.Errorf("create cannon trace provider: %w", err)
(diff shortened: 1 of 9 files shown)


op-challenger/game/fault/register.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/ethereum-optimism/optimism/op-service/txmgr"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+var (
+	cannonGameType   = uint8(0)
+	alphabetGameType = uint8(255)
+)
```
refcell, line 23: Should these game types be fault-specific (since this is in the `fault` package)?

Thinking down the road if there are non-fault game types, having one place for these type definitions may be more canonical.

op-challenger/game/fault/register.go
```diff
@@ -0,0 +1,74 @@
...
+	m metrics.Metricer,
+	cfg *config.Config,
+	txMgr txmgr.TxManager,
+	client bind.ContractCaller,
+	creator resourceCreator,
+) {
+	registry.RegisterGameType(gameType, func(game types.GameMetadata, dir string) (scheduler.GamePlayer, error) {
+		return NewGamePlayer(ctx, logger, m, cfg, dir, game.Proxy, txMgr, client, creator)
+	})
+}
```
refcell: Splitting the registry creation into it's own function here is difficult for me to intuit what's going on and passes a lot of arguments around. Would prefer to keep it simple by inlining and use explicit declarations like so

```suggestion
func RegisterGameTypes(
	registry Registry,
	ctx context.Context,
	logger log.Logger,
	m metrics.Metricer,
	cfg *config.Config,
	txMgr txmgr.TxManager,
	client bind.ContractCaller,
) {
	switch cfg.TraceType {
	case config.TraceTypeCannon:
		resourceCreator := func(addr common.Address, gameDepth uint64, dir string) (faultTypes.TraceProvider, faultTypes.OracleUpdater, error) {
			provider, err := cannon.NewTraceProvider(ctx, logger, m, cfg, client, dir, addr, gameDepth)
			if err != nil {
				return nil, nil, fmt.Errorf("create cannon trace provider: %w", err)
			}
			updater, err := cannon.NewOracleUpdater(ctx, logger, txMgr, addr, client)
			if err != nil {
				return nil, nil, fmt.Errorf("failed to create the cannon updater: %w", err)
			}
			return provider, updater, nil
		}
		playerCreator := func(game types.GameMetadata, dir string) (scheduler.GamePlayer, error) {
			return NewGamePlayer(ctx, logger, m, cfg, dir, game.Proxy, txMgr, client, resourceCreator)
		}
		registry.RegisterGameType(cannonGameType, playerCreator)
	case config.TraceTypeAlphabet:
		resourceCreator := func(addr common.Address, gameDepth uint64, dir string) (faultTypes.TraceProvider, faultTypes.OracleUpdater, error) {
			provider := alphabet.NewTraceProvider(cfg.AlphabetTrace, gameDepth)
			updater := alphabet.NewOracleUpdater(logger)
			return provider, updater, nil
		}
		playerCreator := func(game types.GameMetadata, dir string) (scheduler.GamePlayer, error) {
			return NewGamePlayer(ctx, logger, m, cfg, dir, game.Proxy, txMgr, client, resourceCreator)
		}
		registry.RegisterGameType(alphabetGameType, playerCreator)
	}
}
```

refcell (commented):
Minor nits, otherwise looks like a nice refactor

op-challenger/game/fault/register.go
ajsutton: My current thinking is that each type of game knows how to register its variants.  I could see a centralised list of game type codes but I'm not sure where that would live currently, so it made sense to just keep it with the registration code for now.  The canonical list of these game types is actually in the contract deployment code since that's what defines the "source of truth".

refcell (approved):

PR #7032, opened 2023-08-28
urfave v2 shutdown patterns

ethenotethan:
**Description**
TBD - Still WIP

A clear and concise description of the features you're adding in this pull request.

**Tests**

Please describe any tests you've added. If you've added no tests, or left important behavior untested, please explain why not.

**Additional context**
TBD - Still WIP

**Metadata**

- Fixes #6742

PR #7664, opened 2023-10-13
fix(cli): shutdown cleanly on interrupts

### LINKED ISSUE #6742: op-service: opio change to use cleaner urfave v2 shutdown pattern
- [ ] `opio` waits for interrupt signal, but doesn’t cleanly close urfave CLI, we should use `app.RunContext` (urfave v2) and close the context on interrupt.
- [ ] Update the main entrypoints of all services to use the cleaner shutdown pattern.
- [ ] Force-exit on e.g. receiving 10 consecutive interrupts.

hyunchel:
**Description**

This commit updates main entry points to feed-in a context from the top level, and adds an easy cancel upon interrupt signals. The change allows graceful shutdowns on future features as long as the feature utilizes the context.

**Tests**

No test added. More than glad to add one with some guidance.

**Metadata**

- Fixes #6742

### DIFF
--- a/endpoint-monitor/cmd/main.go
+++ b/endpoint-monitor/cmd/main.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 
 	oplog "github.com/ethereum-optimism/optimism/op-service/log"
+	"github.com/ethereum-optimism/optimism/op-service/opio"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/urfave/cli/v2"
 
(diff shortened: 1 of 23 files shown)


protolambda:
Thank you for the contribution; the issue you are fixing was already started in https://github.com/ethereum-optimism/optimism/pull/7032 but that went stale due to inactivity and testing issues. In the meantime, we have made other related improvements in the op-service package and the resource-closing of the different CLI apps, and I would like to include this in the next iteration of improvements of service shutdowns.

The async block on interrupt + context-based unblock does not effectively change the shutdown behavior alone: services are still deferring their resource-closing, waiting for a context to finish, returning, and dropping resource-closing errors, while not handling force-shutdown cases well.

I thus think we should be refactoring to utilize the new `op-service/cliapp.LifecycleCmd`, and refactor the main modules to have proper start/stop/force-close handling.

And the op-program client deliberately uses a different startup flow, without context handling or go-routines or any advanced CLI entrypoint, since the program needs to be synchronous to be fault-proven. It is not a CLI app.

PR #7669, opened 2023-10-13
refactor(ctb): Add transferControl for more DRY code

maurelian:
Some clean up and deduplication of the deploy scripts.

This PR builds on #7632, but I wanted to split it out to simplify review of changes.

### DIFF
--- a/packages/contracts-bedrock/scripts/Deploy.s.sol
+++ b/packages/contracts-bedrock/scripts/Deploy.s.sol
@@ -95,13 +95,22 @@ contract Deploy is Deployer {
         deploySuperchainConfig();
         initializeSuperchainConfig();
         deployDelayedVetoable(supConfProxy_, "DelayedVetoableForSuperchain");
-        transferSuperchainConfigProxyOwnership();
+
+        transferControl({
+            _controlled: "SuperchainConfigProxy",
+            _newController: "DelayedVetoableForSuperchain",
+            _transferSelector: Proxy.changeAdmin.selector
+        });
 
         address protocolVersionsProxy = deployProtocolVersionsProxy();
         deployProtocolVersions();
         initializeProtocolVersions();
         deployDelayedVetoable(protocolVersionsProxy, "DelayedVetoableForProtocolVersions");
-        transferProtocolVersionsProxyOwnership();
+        transferControl({
+            _controlled: "ProtocolVersionsProxy",
+            _newController: "DelayedVetoableForProtocolVersions",
+            _transferSelector: Proxy.changeAdmin.selector
+        });
     }
 
     /// @notice Deploy a new OP Chain, with an existing SuperchainConfig provided
@@ -178,7 +187,11 @@ contract Deploy is Deployer {
         deployAddressManager();
         address proxyAdmin = deployProxyAdmin();
         deployDelayedVetoable(proxyAdmin, "DelayedVetoableForOpChain");
-        transferProxyAdminOwnership();
+        transferControl({
+            _controlled: "ProxyAdmin",
+            _newController: "DelayedVetoableForOpChain",
+            _transferSelector: Ownable.transferOwnership.selector
+        });
 
         deployOptimismPortalProxy();
         deployL2OutputOracleProxy();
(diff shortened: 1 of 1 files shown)


packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -472,44 +494,25 @@ contract Deploy is Deployer {
...
+    )
+        public
+        broadcast
+    {
         _callViaSafe({
             _safe: Safe(mustGetAddress("SuperchainConfigInitiatorSafe")),
-            _target: protocolVersionsProxy,
-            _data: abi.encodeCall(Proxy.changeAdmin, (delayedVetoableForProtocolVersions))
+            _target: mustGetAddress(_controlled),
+            _data: abi.encodeWithSelector(_transferSelector, mustGetAddress(_newController))
```
tynes, line 497: This has an assumption that the function will always accept a single address arg

packages/contracts-bedrock/scripts/Deploy.s.sol
```diff
@@ -188,8 +200,11 @@ contract Deploy is Deployer {
         deployOptimismMintableERC20FactoryProxy();
         deployL1ERC721BridgeProxy();
         deployDisputeGameFactoryProxy();
-
-        transferAddressManagerOwnership(); // to the ProxyAdmin
+        transferControl({
+            _controlled: "AddressManager",
+            _newController: "ProxyAdmin",
+            _transferSelector: Ownable.transferOwnership.selector
+        });
```
tynes: Can we bring back the higher level functions like `transferAddressManagerOwnership` and then just wrap `transferControl`?

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Yes, but I think it's fine for now. AFAIK in the whole codebase we don't have any ownership transfer like functions which take two args. 

Supporting an arbitrary number of arguments is possible using some variant on `abi.encodeX()` or just passing in the calldata, but that feels to me like it walks back the value of abstracting this operation into a single reusable method.

maurelian: Going to resolve, please feel free to reopen if this is important to you.

packages/contracts-bedrock/scripts/Deploy.s.sol
tynes: Looks like this one comment was resolved without addressing

packages/contracts-bedrock/semver-lock.json
```diff
@@ -1,14 +1,14 @@
 {
   "src/EAS/EAS.sol": "0xea04b7a9ec170ce9cbd466ede87650f9e9ffe3d725d7951cef5777a98a840173",
   "src/EAS/SchemaRegistry.sol": "0x5ee1a0c3b2bf1eb5edb53fb0967cf13856be546f0f16fe7acdc3e4f286db6831",
-  "src/L1/DelayedVetoable.sol": "0x276c6276292095e6aa37a70008cf4e0d1cbcc020dbc9107459bbc72ab5ed744f",
+  "src/L1/DelayedVetoable.sol": "0xc59a6b5e7fe976000d6e9a702d97f048b512938a2ff6dc17f32678ab50da73f3",
```
tynes: Does this need a rebase? Why would the semver lock be changing here?

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: oh sorry, I fixed it locally, but that change got lost in the shuffle. Going to wait for #7670 to merge, then I'll rebase, rebuild and push.

packages/contracts-bedrock/semver-lock.json
maurelian: Rebased

packages/contracts-bedrock/scripts/Deploy.s.sol
maurelian: Pushed the fix

PR #7679, opened 2023-10-13
contracts-bedrock: initialize tests

tynes:
**Description**

Make the ERC721Bridge test setup be part of the same process as the
StandardBridge setup so that its possible to have the entire L1 system
set up in a single unit test block. This makes writing tests that
touch all of the system L1 contracts much more easy. These changes
were done with some find and replaces.

Test coverage of the L1 `Initializable` contracts to cover the
case of contracts being reinitialized. No contract should be able
to have `initialize()` called twice.


maurelian (approved):
Good stuff

packages/contracts-bedrock/test/Initializable.t.sol
```diff
@@ -0,0 +1,61 @@
...
+import { SystemConfig } from "src/L1/SystemConfig.sol";
+import { ResourceMetering } from "src/L1/ResourceMetering.sol";
+import { OptimismPortal } from "src/L1/OptimismPortal.sol";
+
+/// @title Initializer_Test
+/// @dev Ensures that the `initialize()` function on contracts cannot be called more than
+///      once. This contract inherits from `ERC721Bridge_Initializer` because it is the
+///      deepest contract in the inheritance chain for setting up the system contracts.
+contract Initializer_Test is ERC721Bridge_Initializer {
+    function test_cannotReinitializeL1_succeeds() public {
```
Inphi, line 17: I know this PR focuses on L1 contracts, but is it also worth asserting that the `L2Messenger` cannot be reinitialized?

tynes: Thinking about how to handle this because really what we want to test is in practice these cannot be reinitialized. To do this, we would want to test against the L1 contracts being set up by the deploy script and test against the L2 contracts being set up by the genesis gen script

Inphi: i see. That sounds quite involved to add to this PR. We should look into doing this soon.

Inphi (approved):

packages/contracts-bedrock/test/Initializable.t.sol
```diff
@@ -0,0 +1,61 @@
...
+import { CrossDomainMessenger } from "src/universal/CrossDomainMessenger.sol";
+import { L2OutputOracle } from "src/L1/L2OutputOracle.sol";
+import { SystemConfig } from "src/L1/SystemConfig.sol";
+import { ResourceMetering } from "src/L1/ResourceMetering.sol";
+import { OptimismPortal } from "src/L1/OptimismPortal.sol";
+
+/// @title Initializer_Test
+/// @dev Ensures that the `initialize()` function on contracts cannot be called more than
+///      once. This contract inherits from `ERC721Bridge_Initializer` because it is the
+///      deepest contract in the inheritance chain for setting up the system contracts.
```
mds1, line 15: Maybe overkill, but is it worth considering how we can ensure we aren't missing any contracts in this test? It seems we'd currently just verify manually which might be sufficient. A more robust way might be something like:

```solidity
// Define initialization calls for each contract.
struct InitializeableContract {
  address target;
  bytes initializationCalldata;
}

InitializeableContract[] initializeableContracts;
initializeableContracts.push(InitializeableContract({
    target: address(L1Messenger), 
    initializationCalldata: abi.encodeCall(L1Messenger.initialize, (OptimismPortal(payable(address(0)))))
}));
initializeableContracts.push(etc...);

// Read artifacts for all contracts in src/L1.
// --- snip ---

// If it has a method called initialize, increment counter.
uint256 numInitializeableContracts;
// --- snip ---

// Verify no contracts are missing.
assertEq(initializeableContracts.length, numInitializeableContracts, "Missing contracts");

// Now execute the initialization calls that we expect to revert.
// --- snip ---
```

Alternatively could even dynamically generate the `InitializeableContract[] contracts` array from the initialization sigs read from the artifacts

tynes: This approach is far superior. It can add additional contracts that are `Initializable` automatically over time with no need for a diff. The problem is that the way that the tests are initialized, the code is not modular enough to make it really easy for all possible tests to be initialized correctly in the same test suite without rewriting a bunch of initialization code. This also begs the question of, why even test the non production contract setup code? Its possible that we are initializing the contracts correctly in the tests but not in the deploy script

Everything is pointing towards redoing the unit tests based on the deploy script so that we have 1 unified contract setup step and then all test coverage can be against the way that tests are configured in practice.

mds1: This is helpful context, thanks. Given that, I'd support leaving this PR as-is for now, and having an issue to track the modification of this `test_cannotReinitializeL1_succeeds` test to be artifact-based once the change to testing based on deploy scripts is made

tynes: https://github.com/ethereum-optimism/optimism/issues/7691

mds1 (approved):

Issue #7671, opened 2023-10-13
Extend op-service LifecycleCmd usage to all long-running Go services


protolambda:
Op-node uses `cliapp.LifecycleCmd` to handle startup/shutdown/force-shutdown, while ensuring resources are closed and any shutdown errors are propagated properly.

Previous context-based shutdown patterns encouraged bad code that shuts down like `defer thing.Close()` without propagating the error, or even returning before go-routines of the service fully shutdown, since context closure triggers shutdown, but does not await the results.

Services to update:
- [x] op-batcher
- [x] op-proposer
- [x] endpoint-monitor
- [x] indexer
- [x] op-wheel (`engine auto` sub-command is long-running, other sub-commands are not, may require some more work)
- [x] op-bootnode
- [x] op-heartbeat
- [x] op-challenger

mslipper:
@protolambda are you still working on this?

protolambda:
@mslipper an external contributor implemented the op-proposer changes last week. op-batcher and op-node are already done. Op-challenger remains I think.

protolambda:
The indexer is also done. And I think op-bootnode, op-heartbeat and op-wheel are nice to have; they work relatively well as-is since they don't have a lot of resources to start/close, but I may be able to implement the changes there soon now that I'm back from devconnect.

PR #7682, opened 2023-10-14
op-batcher: service lifecycle cleanup

protolambda:
**Description**

Changes:
- create `BatcherService`: one consistent place that hosts all the typed resources
- refactor `BatchSubmitter` driver to only depend on minimal interfaces
- rename start/stop parts of driver, to clarify it's about the batch-submitting work, and not the service as a whole
- cleanup the initialization of the batcher service to be in one place with consistent error handling
- initialize the driver as one of the resources of the service: this way the service can cleanly start/stop it as necessary, and use it as backend for the start/stop-submitting admin RPC endpoints
- cleanup double config thing: there is only a CLI config, and the driver really depends on a setup of initialized dependencies, rather than a config that is not a config.
- refactor all resources to be properly `io.Close`-ed on service shutdown, with error joining
  - including RPC clients
  - including HTTP server services
  - including batch-submission work
  - every resource, even the balance reporting background loop
    - introduced a new op-service util to make it easier to create looping background functions that can be closed nicely.
- The `BatcherService` implements the CLI lifecycle interface, used by the batcher CLI entrypoint

**Tests**

Minor op-e2e updates to start/stop the `BatchSubmitter`, and update any tests that really start/stop the batch-submission, making the service/driver separation more clear.

All service start/close code is covered by the existing tests, but should be a lot cleaner now that resources are more properly managed.

**Additional context**

Also found the issue with the already-closed error that was causing CI flakes, will stack a PR on top of this to update the batcher logic. *This PR avoids making any logic changes to the batcher driver.*

**Metadata**

First part of https://github.com/ethereum-optimism/optimism/issues/7671

~TODO: I need to open a few issues to track minor TODOs (will use TODO linking from @clabby !)~


op-service/clock/loop.go
```diff
@@ -0,0 +1,62 @@
...
+	defer lf.wg.Done()
+	defer lf.ticker.Stop() // clean up the timer
+	for {
+		select {
+		case <-lf.ctx.Done():
+			return
+		case <-lf.ticker.Ch():
+			ctx, cancel := context.WithCancel(lf.ctx)
+			lf.fn(ctx)
+			cancel()
```
sebastianst: What's the purpose of deriving a local context here? Why not just
```suggestion
			lf.fn(lf.ctx)
```
?

And if there is a purpose, I'd make sure that a panic in `lf.fn` doesn't drop a call to `cancel` like
```suggestion
			ctx, cancel := context.WithCancel(lf.ctx)
			func() {
				defer cancel()
				lf.fn(ctx)
			}()
```

op-service/clock/loop_test.go
```diff
@@ -0,0 +1,27 @@
...
+	loopFn := NewLoopFn(cl, func(ctx context.Context) {
+		calls <- struct{}{}
+	}, func() error {
+		close(calls)
+		return testErr
+	}, time.Second*10)
+	cl.AdvanceTime(time.Second * 15)
+	<-calls
+	cl.AdvanceTime(time.Second * 10)
+	<-calls
```
sebastianst, line 25: I suggest adding
```suggestion
	<-calls
	select {
	case <-calls:
		t.Fatal("more calls than expected")
	default:
	}
```
to assert that exactly 2 calls have been made at this point, and not more.

op-e2e/system_fpp_test.go
```diff
@@ -93,7 +93,7 @@ func testVerifyL2OutputRootEmptyBlock(t *testing.T, detached bool) {
 	l2OutputRoot := agreedL2Output.OutputRoot
 
 	t.Log("=====Stopping batch submitter=====")
-	err = sys.BatchSubmitter.Stop(ctx)
+	err = sys.BatchSubmitter.Driver().StopBatchSubmitting(context.Background())
```
sebastianst: There's a `ctx` available here, so I'd use it. Although it is just the background context right now, if it is ever changed, we don't miss using it here.

op-batcher/batcher/service.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	bs.rpcServer = server
+	return nil
+}
+
+// Check ensures that the [BatcherService] is valid
+func (bs *BatcherService) Check() error {
+
+	return nil
+}
```
sebastianst: Is it planned to add something here soon? Otherwise I'd delete.

sebastianst (changes requested):
Nice cleanup 🧼

op-batcher/batcher/service.go
protolambda: Ah, it used to have the rollup/channel config checks, but I moved those to check at initialization time, so the configs are checked before they are used. Will drop the emtpy function

op-service/clock/loop.go
protolambda: The local context here ensures that things that are opened in the function call don't stick around for longer than the function call, to avoid a looping function from creating lots of unclosed resources

op-batcher/batcher/service.go
```diff
@@ -0,0 +1,351 @@
...
+		return fmt.Errorf("failed to start pprof server: %w", err)
+	}
+	bs.initDriver()
+	if err := bs.initRPCServer(cfg); err != nil {
+		return fmt.Errorf("failed to start RPC server: %w", err)
+	}
+	// Validate the setup
+	if err := bs.Check(); err != nil {
+		return fmt.Errorf("failed post-initialization check: %w", err)
+	}
```
protolambda:
```suggestion
```

op-batcher/batcher/service.go
protolambda:
```suggestion
```

op-batcher/batcher/service.go
```diff
@@ -0,0 +1,351 @@
...
+	bs.rpcServer = server
+	return nil
+}
+
+// Check ensures that the [BatcherService] is valid
+func (bs *BatcherService) Check() error {
+
+	return nil
+}
+
```
protolambda:
```suggestion
```

op-e2e/system_fpp_test.go
protolambda:
```suggestion
	err = sys.BatchSubmitter.Driver().StopBatchSubmitting(ctx)
```

sebastianst (approved):

op-batcher/batcher/config.go
```diff
@@ -107,15 +72,15 @@ func (c CLIConfig) Check() error {
...
+	if err := c.RPC.Check(); err != nil {
 		return err
 	}
 	return nil
 }
 
 // NewConfig parses the Config from the provided flags or environment variables.
-func NewConfig(ctx *cli.Context) CLIConfig {
-	return CLIConfig{
+func NewConfig(ctx *cli.Context) *CLIConfig {
```
trianglesphere, line 82: Why is this a pointer now?

op-service/dial/dial.go
```diff
@@ -21,8 +21,8 @@ const defaultRetryTime = 2 * time.Second
 // DialEthClientWithTimeout attempts to dial the L1 provider using the provided
 // URL. If the dial doesn't complete within defaultDialTimeout seconds, this
 // method will return an error.
-func DialEthClientWithTimeout(timeout time.Duration, log log.Logger, url string) (*ethclient.Client, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+func DialEthClientWithTimeout(ctx context.Context, timeout time.Duration, log log.Logger, url string) (*ethclient.Client, error) {
```
trianglesphere, line 24: Once you've added the context, there's not much point of retaining the timeout as callers can set the timeout

op-batcher/batcher/config.go
protolambda: To avoid unnecessary copies of the config struct, with all the op-service sub-configs it's gotten pretty large, and it might not be clone-able with just a shallow-copy in the future.

op-service/dial/dial.go
protolambda: Good point, I'll refactor it as part of the remaining op-proposer work, when the remaining callers of this method handle their contexts better here

PR #7683, opened 2023-10-14
op-batcher: adjust error handling on pending-channels after close

protolambda:
**Description**

This PR depends on https://github.com/ethereum-optimism/optimism/pull/7682

This fixes a CI flake by adjusting the error reporting of pending-channels.

CI would previously fail when there is remaining pending channel-data that does not fit all in one channel, due to the compressor-full status.

-----

Here's the rabbit-hole I went down to find this issue:

`ChannelOut.Close()` is being called twice on a closed state, which should never happen.

CI Log snippet:
```
    driver.go:155:              INFO [10-12|17:46:15.226] Stopping Batch Submitter                 role=batcher
    driver.go:313:              ERROR[10-12|17:46:15.226] error closing the channel manager        role=batcher      err="creating frames with channel builder: closing channel out: already closed"
```

Batcher :
- `ChannelOut.Close()` is called as part of `channelBuilder.closeAndOutputAllFrames()`, `closing channel out` is logged.
- `channelBuilder.closeAndOutputAllFrames()` is called as part of `outputFrames()`, `creating frames with channel builder` is logged.
- `outputFrames()` is called both as part of `channelManager.Close()` and `channelManager.TxData()`
- `channelManager.Close()` is only called like `BatchSubmitter.state.Close()`, mutually exclusive driver loop events:
  - Option 1: closed and then cleared (reopened) on reorg
  - Option 2: closed on shutdown. `error closing the channel manager` is logged.
- `channelManager.TxData()` is only called `publishTxToL1()`
- `publishTxToL1` is called in a go-routine by `publishStateToL1`, but caller always blocks till it is done publishing.
- `publishStateToL1` is called *after* closing the state and logging closing error on shutdown, in the driver loop.
- `publishStateToL1` is also called in more driver-loop places.

Given that `publishStateToL1` waits till it is done, there seems to be no concurrency issue, and it's called many times over in regular operation.
But `closeAndOutputAllFrames` followed by `Close` seems to be the issue: only then we can hit the log from CI fail.

On `channelManager.Close()`, the `channelBuilder.Close` happens, but this only really marks it as "full", with `ErrTerminated`
However, the `ChannelOut` of the `channelBuilder` should have been removed if it were completely unutilized, and have nothing to submit.

The test-cases, when not dropping the error, then indicate the exact issue: the compressor is "full", and when an existing pending channel is "fully" submitted, it actually does not get fully submitted, and will appear already-closed.

----

**Changes**

- ~Do not make the channel-manager `Close` func try to output frames: this happens as part of the later `TxData` calls anyway.~ Edit(seb): Only `Close` a channel and `outputFrames` in `channelManager.Close` if the channel wasn't already full. This was causing the double-closed error, when things didn't fit in a single channel.
- Return an error that channel data remains to be submitted.
- Add clear logging on channel-manager close on what is being dropped as never-submitted and what is still pending after close
- Adjust the error log to a warn-log if pending data remains to be submitted: the driver is capable of draining that by submitting it all to L1 before final shutdown.
- Include channel full-cause in error on non-err output of channel frames, so we can better understand issues like the previous already-closed error.
- Comments to explain what's happening, because it was not clear.
- Minor cleanup of error naming/duplication.

**Tests**

`Close` errors in the channel-manager tests are no longer dropped.

`ChannelManagerClosePendingChannel` tested the exact case of remaining pending-channels work after channel-manager close, but ignored the error.

Added new `NonCompressor` that can be used in testing. When writing to it, it first flushes data from any previous write. This makes it behave more predicable which is better suited for tests.

Using this one, added a new test `ChannelManager_Close_PartiallyPendingChannel` that shows that the `outputFrames` call is necessary in `channelManager.Close` in case there are unflushed blocks in the compressor.

### DIFF
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -371,12 +371,22 @@ func (s *channelManager) Close() error {
 		return nil
 	}
 
-	// Force-close the remaining open channel early (if not already closed):
-	// it will be marked as "full" due to service termination.
-	s.currentChannel.Close()
+	// If the channel is already full, we don't need to close it or output frames.
+	// This would already have happened in TxData.
+	if !s.currentChannel.IsFull() {
+		// Force-close the remaining open channel early (if not already closed):
+		// it will be marked as "full" due to service termination.
+		s.currentChannel.Close()
 
-	// We do not s.outputFrames():
-	// this will only error if we cannot submit it in one go due to channel limits.
-	// Instead, make it clear to the caller that there is remaining pending work.
-	return ErrPendingAfterClose
+		// Final outputFrames call in case there was unflushed data in the compressor.
+		if err := s.outputFrames(); err != nil {
+			return fmt.Errorf("outputting frames during close: %w", err)
+		}
+	}
+
+	if s.currentChannel.HasFrame() {
+		// Make it clear to the caller that there is remaining pending work.
+		return ErrPendingAfterClose
+	}
+	return nil
 }
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -241,7 +241,7 @@ func (s *channelManager) processBlocks() error {
 		} else if err != nil {
 			return fmt.Errorf("adding block[%d] to channel builder: %w", i, err)
 		}
-		s.log.Debug("Added block to channel", "channel", s.currentChannel.ID(), "block", block)
+		s.log.Debug("Added block to channel", "id", s.currentChannel.ID(), "block", block)
 
 		blocksAdded += 1
 		latestL2ref = l2BlockRefFromBlockAndL1Info(block, l1info)
@@ -359,10 +359,10 @@ func (s *channelManager) Close() error {
 	// Any pending state can be proactively cleared if there are no submitted transactions
 	for _, ch := range s.channelQueue {
 		if ch.NoneSubmitted() {
-			s.log.Info("Channel has no past or pending submission and might as well not exist, channel will be dropped", "channel", ch.ID(), "")
+			s.log.Info("Channel has no past or pending submission - dropping", "id", ch.ID(), "")
 			s.removePendingChannel(ch)
 		} else {
-			s.log.Info("Channel is in-flight and will need to be submitted after close", "channel", ch.ID(), "confirmed", len(ch.confirmedTransactions), "pending", len(ch.pendingTransactions))
+			s.log.Info("Channel is in-flight and will need to be submitted after close", "id", ch.ID(), "confirmed", len(ch.confirmedTransactions), "pending", len(ch.pendingTransactions))
 		}
 	}
 	s.log.Info("Reviewed all pending channels on close", "remaining", len(s.channelQueue))
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -365,7 +365,7 @@ func (s *channelManager) Close() error {
 			s.log.Info("Channel is in-flight and will need to be submitted after close", "channel", ch.ID(), "confirmed", len(ch.confirmedTransactions), "pending", len(ch.pendingTransactions))
 		}
 	}
-	s.log.Info("reviewed all pending channels on close", "remaining", len(s.channelQueue))
+	s.log.Info("Reviewed all pending channels on close", "remaining", len(s.channelQueue))
 
 	if s.currentChannel == nil {
 		return nil
--- a/op-batcher/batcher/channel_manager.go
+++ b/op-batcher/batcher/channel_manager.go
@@ -337,9 +337,15 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
 	}
 }
 
-// Close closes the current pending channel, if one exists, outputs any remaining frames,
-// and prevents the creation of any new channels.
-// Any outputted frames still need to be published.
+var ErrPendingAfterClose = errors.New("pending channels remain after closing channel-manager")
+
+// Close clears any pending channels that are not in-flight already, to leave a clean derivation state.
+// Close then marks the remaining current open channel, if any, as "full" so it can be submitted as well.
+// Close does NOT immediately output frames for the current remaining channel:
+// as this might error, due to limitations on a single channel.
+// Instead, this is part of the pending-channel submission work: after closing,
+// the caller SHOULD drain pending channels by generating TxData repeatedly until there is none left (io.EOF).
+// A ErrPendingAfterClose error will be returned if there are any remaining pending channels to submit.
 func (s *channelManager) Close() error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -348,19 +354,29 @@ func (s *channelManager) Close() error {
 	}
 
 	s.closed = true
+	s.log.Info("Channel manager is closing")
 
 	// Any pending state can be proactively cleared if there are no submitted transactions
 	for _, ch := range s.channelQueue {
 		if ch.NoneSubmitted() {
+			s.log.Info("Channel has no past or pending submission and might as well not exist, channel will be dropped", "channel", ch.ID(), "")
 			s.removePendingChannel(ch)
+		} else {
+			s.log.Info("Channel is in-flight and will need to be submitted after close", "channel", ch.ID(), "confirmed", len(ch.confirmedTransactions), "pending", len(ch.pendingTransactions))
 		}
 	}
+	s.log.Info("reviewed all pending channels on close", "remaining", len(s.channelQueue))
 
 	if s.currentChannel == nil {
 		return nil
 	}
 
+	// Force-close the remaining open channel early (if not already closed):
+	// it will be marked as "full" due to service termination.
 	s.currentChannel.Close()
 
-	return s.outputFrames()
+	// We do not s.outputFrames():
+	// this will only error if we cannot submit it in one go due to channel limits.
+	// Instead, make it clear to the caller that there is remaining pending work.
+	return ErrPendingAfterClose
 }
--- a/op-batcher/batcher/driver.go
+++ b/op-batcher/batcher/driver.go
@@ -265,9 +265,9 @@ func (l *BatchSubmitter) loop() {
 				err := l.state.Close()
 				if err != nil {
 					if errors.Is(err, ErrPendingAfterClose) {
-						l.Log.Warn("closed channel manager to handle L2 reorg, but pending channel(s) remain to be submitted")
+						l.Log.Warn("Closed channel manager to handle L2 reorg, but pending channel(s) remain to be submitted")
 					} else {
-						l.Log.Error("error closing the channel manager to handle a L2 reorg", "err", err)
+						l.Log.Error("Error closing the channel manager to handle a L2 reorg", "err", err)
 					}
 				}
 				l.publishStateToL1(queue, receiptsCh, true)
@@ -283,13 +283,13 @@ func (l *BatchSubmitter) loop() {
 			err := l.state.Close()
 			if err != nil {
 				if errors.Is(err, ErrPendingAfterClose) {
-					l.Log.Warn("closed channel manager, but pending channel(s) remain to be submitted")
+					l.Log.Warn("Closed channel manager, but pending channel(s) remain to be submitted")
 				} else {
-					l.Log.Error("error closing the channel manager", "err", err)
+					l.Log.Error("Error closing the channel manager", "err", err)
 				}
 			}
 			l.publishStateToL1(queue, receiptsCh, true)
-			l.Log.Info("finished publishing all remaining channel data")
+			l.Log.Info("Finished publishing all remaining channel data")
 			return
 		}
 	}
--- a/op-batcher/batcher/driver.go
+++ b/op-batcher/batcher/driver.go
@@ -278,8 +278,8 @@ func (l *BatchSubmitter) loop() {
 		case r := <-receiptsCh:
 			l.handleReceipt(r)
 		case <-l.shutdownCtx.Done():
-			// This removes any pending channels, so these do not have to be drained .
-			// But it also tries to write ev
+			// This removes any never-submitted pending channels, so these do not have to be drained with transactions.
+			// Any remaining unfinished channel is terminated, so its data gets submitted.
 			err := l.state.Close()
 			if err != nil {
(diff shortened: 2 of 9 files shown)


protolambda:
fixed semgrep todo issues in base PR, semgrep comment is stale

op-batcher/batcher/driver.go
```diff
@@ -266,11 +270,18 @@ func (l *BatchSubmitter) loop() {
 		case r := <-receiptsCh:
 			l.handleReceipt(r)
 		case <-l.shutdownCtx.Done():
+			// This removes any pending channels, so these do not have to be drained .
+			// But it also tries to write ev
```
sebastianst: typos

op-batcher/batcher/driver.go
```diff
@@ -256,7 +256,11 @@ func (l *BatchSubmitter) loop() {
 			if err := l.loadBlocksIntoState(l.shutdownCtx); errors.Is(err, ErrReorg) {
 				err := l.state.Close()
 				if err != nil {
-					l.Log.Error("error closing the channel manager to handle a L2 reorg", "err", err)
+					if errors.Is(err, ErrPendingAfterClose) {
+						l.Log.Warn("closed channel manager to handle L2 reorg, but pending channel(s) remain to be submitted")
```
sebastianst: I think we can downgrade this message to `Info` because this is an expected situation. Same below.

op-batcher/batcher/channel_manager.go
```diff
@@ -343,19 +349,29 @@ func (s *channelManager) Close() error {
...
 
+	// Force-close the remaining open channel early (if not already closed):
+	// it will be marked as "full" due to service termination.
 	s.currentChannel.Close()
 
-	return s.outputFrames()
+	// We do not s.outputFrames():
+	// this will only error if we cannot submit it in one go due to channel limits.
+	// Instead, make it clear to the caller that there is remaining pending work.
+	return ErrPendingAfterClose
```
sebastianst: Why can we safely skip the `outputFrames()` call here? Where are the frames output instead? Note that the `TxData` call [exits early for closed channel managers](https://github.com/ethereum-optimism/optimism/blob/8a3c8964700f7890d021c76841c7853b1c4d5a89/op-batcher/batcher/channel_manager.go#L158) and doesn't call it.

I'm also wondering if we're missing a test (not the channel manager unit test) for this scenario, where the batcher is closed with a pending channel and correct submission is asserted.

sebastianst (dismissed):

op-batcher/batcher/driver.go
protolambda: It's an expected situation, but still risky w.r.t. unclean shutdowns, so I think it's worth a warn-log.

op-batcher/batcher/channel_manager.go
protolambda: My impression was that the `outputFrames` call [here](https://github.com/ethereum-optimism/optimism/blob/8a3c8964700f7890d021c76841c7853b1c4d5a89/op-batcher/batcher/channel_manager.go#L183C14-L183C26) is sufficient. Is it possible that without the `outputFrames` here, it potentially submits less frame data in the first tx-data, but then still submits everything in the following txs? Note that `TestChannelManagerClosePendingChannel` passes.

op-batcher/batcher/driver.go
protolambda:
```suggestion
			// This removes any never-submitted pending channels, so these do not have to be drained with transactions.
			// Any remaining unfinished channel is terminated, so its data gets submitted.
```

op-batcher/batcher/channel_manager.go
sebastianst: This `outputFrames` call should never be reached if called on a closed channel manager because of the early exit I linked in my first comment.

I'm worried that the test is no good, and that with `outputFrames` skipped, we might end up with a truncated channel. When the underlying `channel` is `Close`d, it has to be guaranteed that `OutputFrames` is called on it once, or the compressor is not flushed, afaik.

It's probably a good idea to expand the channel manager tests, at least this one, to assert that all the tx data output by the `TxData` calls, if reassembled, contains all the blocks that were previously put into the channel manager. I'm otherwise also not sure what's going on here, but I don't trust the test as it is.

op-batcher/batcher/channel_manager.go
```diff
@@ -343,19 +349,29 @@ func (s *channelManager) Close() error {
...
 	// Any pending state can be proactively cleared if there are no submitted transactions
 	for _, ch := range s.channelQueue {
 		if ch.NoneSubmitted() {
+			s.log.Info("Channel has no past or pending submission and might as well not exist, channel will be dropped", "channel", ch.ID(), "")
 			s.removePendingChannel(ch)
+		} else {
+			s.log.Info("Channel is in-flight and will need to be submitted after close", "channel", ch.ID(), "confirmed", len(ch.confirmedTransactions), "pending", len(ch.pendingTransactions))
 		}
 	}
+	s.log.Info("reviewed all pending channels on close", "remaining", len(s.channelQueue))
```
ajsutton:
```suggestion
	s.log.Info("Reviewed all pending channels on close", "remaining", len(s.channelQueue))
```

op-batcher/batcher/driver.go
ajsutton:
```suggestion
						l.Log.Warn("Closed channel manager to handle L2 reorg, but pending channel(s) remain to be submitted")
```

op-batcher/batcher/driver.go
```diff
@@ -256,7 +256,11 @@ func (l *BatchSubmitter) loop() {
...
+					} else {
+						l.Log.Error("error closing the channel manager to handle a L2 reorg", "err", err)
```
ajsutton:
```suggestion
						l.Log.Error("Error closing the channel manager to handle a L2 reorg", "err", err)
```

op-batcher/batcher/driver.go
```diff
@@ -266,11 +270,18 @@ func (l *BatchSubmitter) loop() {
 		case r := <-receiptsCh:
 			l.handleReceipt(r)
 		case <-l.shutdownCtx.Done():
+			// This removes any never-submitted pending channels, so these do not have to be drained with transactions.
+			// Any remaining unfinished channel is terminated, so its data gets submitted.
 			err := l.state.Close()
 			if err != nil {
-				l.Log.Error("error closing the channel manager", "err", err)
+				if errors.Is(err, ErrPendingAfterClose) {
+					l.Log.Warn("closed channel manager, but pending channel(s) remain to be submitted")
```
ajsutton:
```suggestion
					l.Log.Warn("Closed channel manager, but pending channel(s) remain to be submitted")
```

op-batcher/batcher/driver.go
```diff
@@ -266,11 +270,18 @@ func (l *BatchSubmitter) loop() {
...
+				} else {
+					l.Log.Error("error closing the channel manager", "err", err)
+				}
 			}
 			l.publishStateToL1(queue, receiptsCh, true)
+			l.Log.Info("finished publishing all remaining channel data")
```
ajsutton:
```suggestion
			l.Log.Info("Finished publishing all remaining channel data")
```

op-batcher/batcher/driver.go
```diff
@@ -266,11 +270,18 @@ func (l *BatchSubmitter) loop() {
...
 		case <-l.shutdownCtx.Done():
+			// This removes any never-submitted pending channels, so these do not have to be drained with transactions.
+			// Any remaining unfinished channel is terminated, so its data gets submitted.
 			err := l.state.Close()
 			if err != nil {
-				l.Log.Error("error closing the channel manager", "err", err)
+				if errors.Is(err, ErrPendingAfterClose) {
+					l.Log.Warn("closed channel manager, but pending channel(s) remain to be submitted")
+				} else {
+					l.Log.Error("error closing the channel manager", "err", err)
```
ajsutton:
```suggestion
					l.Log.Error("Error closing the channel manager", "err", err)
```

op-batcher/batcher/channel_manager.go
ajsutton: I'm not familiar with the batcher code so take this with a grain of salt, but I have the sense that there's a critical fork in `channelBuilder.Close` where it checks if the the channel is already marked full.  I suspect we want tests that cover both when it is and isn't full. `TestChannelManagerClosePendingChannel` is hitting the case where it's already full because `compressor is full`. I _think_ the net result of that is that all the data has already been output to frames and so it doesn't matter if we have this call to `outputFrames` or not.

With the channel already full, the call to `channelBuilder.OutputFrames` that results from the first call to `m.TxData` in the test will have already run the IsFull case where it calls `closeAndOutputAllFrames`.

I can't work out how to adjust the test to hit the other case where the full err actually gets set to `ErrTerminated` but I suspect we need to get there for this extra call to be important. Then `channelBuilder.OutputFrames` will only call `outputReadyFrames` initially and there will be some extra data that hasn't yet been added to the output frame list which would leave the channel unclosed.

It's telling that I can modify `channelBuilder.Close` to:
```
// Close immediately marks the channel as full with an ErrTerminated
// if the channel is not already full.
func (c *channelBuilder) Close() {
	println("Closing channel")
	if !c.IsFull() {
		panic("Setting ErrTerminated")
		c.setFullErr(ErrTerminated)
	}
}
```

and all the `channel_manager_test` cases still pass, so none of them is actually hitting the case where `channelBuilder.Close` is actually doing anything.

sebastianst: Finally got around to this. Managed to write a test that does hit this code path and that shows that the call to `outputFrames` is actually needed.

op-batcher/compressor/non_compressor_test.go
```diff
@@ -0,0 +1,39 @@
...
+	data := make([]byte, dlen)
+	rng := rand.New(rand.NewSource(42))
+	rng.Read(data)
+
+	n, err := c.Write(data)
+	require.NoError(err)
+	require.Equal(n, dlen)
+	l0 := c.Len()
+	require.Less(l0, dlen)
+	require.Equal(7, l0)
```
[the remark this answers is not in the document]
sebastianst: this is before flushing, so the 7 bytes is just the zlib preamble overhead or so

op-batcher/compressor/non_compressor.go
```diff
@@ -0,0 +1,82 @@
...
+		return 0, err
+	}
+	n1, err := t.compress.Write(p[hlen:])
+	if err != nil {
+		return 0, err
+	}
+	if uint64(t.buf.Len()) > t.config.TargetFrameSize*uint64(t.config.TargetNumFrames) {
+		t.fullErr = derive.CompressorFullErr
+	}
+	return n0 + n1, nil
```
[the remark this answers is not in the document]
sebastianst: thanks rabbit, that was a test I never meant to commit

op-batcher/batcher/channel_manager.go
```diff
@@ -348,19 +354,39 @@ func (s *channelManager) Close() error {
 	}
 
 	s.closed = true
+	s.log.Info("Channel manager is closing")
 
 	// Any pending state can be proactively cleared if there are no submitted transactions
 	for _, ch := range s.channelQueue {
 		if ch.NoneSubmitted() {
+			s.log.Info("Channel has no past or pending submission and might as well not exist, channel will be dropped", "channel", ch.ID(), "")
```
ajsutton: nit: This log message makes it sounds like an error or at least unusual condition. If it's expected then we can probably reduce the potential for confusion by making it say something like `Channel has no past or pending submission - dropping`.

op-batcher/batcher/driver.go
```diff
@@ -274,11 +278,18 @@ func (l *BatchSubmitter) loop() {
 		case r := <-receiptsCh:
 			l.handleReceipt(r)
 		case <-l.shutdownCtx.Done():
+			// This removes any never-submitted pending channels, so these do not have to be drained with transactions.
+			// Any remaining unfinished channel is terminated, so its data gets submitted.
 			err := l.state.Close()
 			if err != nil {
-				l.Log.Error("error closing the channel manager", "err", err)
+				if errors.Is(err, ErrPendingAfterClose) {
+					l.Log.Warn("Closed channel manager, but pending channel(s) remain to be submitted")
```
ajsutton: Should change this log message to mention that the pending channels will be submitted? It sounds like this is an invalid state and we've left the channel unclosed.

ajsutton (approved):
LGTM. Just a couple of suggestions to avoid log messages sending me down a rabbit hole in the future. :)

op-batcher/batcher/driver.go
sebastianst: Changed to "Closed channel manager on shutdown with pending channel(s) remaining - submitting", similarly for L2 reorg warning.

op-batcher/batcher/driver.go
```diff
@@ -274,11 +278,18 @@ func (l *BatchSubmitter) loop() {
...
+			// This removes any never-submitted pending channels, so these do not have to be drained with transactions.
+			// Any remaining unfinished channel is terminated, so its data gets submitted.
 			err := l.state.Close()
 			if err != nil {
-				l.Log.Error("error closing the channel manager", "err", err)
+				if errors.Is(err, ErrPendingAfterClose) {
+					l.Log.Warn("Closed channel manager on shutdown with pending channel(s) remaining - submitting")
+				} else {
+					l.Log.Error("Error closing the channel manager on shutdown", "err", err)
+				}
```
axelKingsley, line 289: This duplicate error handling seems to indicate that closing the channel manager with pending work isn't really an error at all, right?

Can `Close` return how much remaining work it has along with the error? That way its signal can be handled without these `errors.Is` calls

op-batcher/batcher/channel_manager.go
```diff
@@ -337,9 +337,15 @@ func l2BlockRefFromBlockAndL1Info(block *types.Block, l1info derive.L1BlockInfo)
 	}
 }
 
-// Close closes the current pending channel, if one exists, outputs any remaining frames,
-// and prevents the creation of any new channels.
-// Any outputted frames still need to be published.
+var ErrPendingAfterClose = errors.New("pending channels remain after closing channel-manager")
+
+// Close clears any pending channels that are not in-flight already, to leave a clean derivation state.
+// Close then marks the remaining current open channel, if any, as "full" so it can be submitted as well.
```
axelKingsley, line 343: It seems like we're marking non-full channels as full specifically so they'll be handled in a certain way. Is that an appropriate use of the marking? I see elsewhere that we emit errors about a "error while closing full channel", and worry that will be confusing if the channel isn't *actually* full.

PR #7690, opened 2023-10-16
op-service: fix exponential backoff strategy in case of high number of attempts

protolambda:
**Description**

Simple overflow bugs in the backoff computation of the exponential stategy resulted in the minimum duration being returned, instead of the maximum duration.  Also, jitter should always be applied, to avoid issues after reaching the max.

**Tests**

Add test-cases for high number of attempts.

**Additional context**

Detected the issue when hitting a high amount of re-attempts during OP-Goerli issue, and subsequent high log volume.

### DIFF
--- a/op-service/retry/strategies.go
+++ b/op-service/retry/strategies.go
@@ -35,16 +35,13 @@ func (e *ExponentialStrategy) Duration(attempt int) time.Duration {
 	if attempt < 0 {
 		return e.Min + jitter
 	}
-	durFloat := float64(e.Min) * float64(time.Nanosecond)
+	durFloat := float64(e.Min)
 	durFloat += math.Pow(2, float64(attempt)) * float64(time.Second)
 	dur := time.Duration(durFloat)
 	if durFloat > float64(e.Max) {
 		dur = e.Max
 	}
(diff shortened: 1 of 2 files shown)


op-service/retry/strategies.go
```diff
@@ -23,15 +23,24 @@ type ExponentialStrategy struct {
...
+		return e.Min + jitter
+	}
+	durFloat := float64(e.Min) * float64(time.Nanosecond)
+	durFloat += math.Pow(2, float64(attempt)) * float64(time.Second)
+	dur := time.Duration(durFloat)
+	if durFloat > float64(e.Max) {
+		dur = e.Max
+	}
 	dur += jitter
 	if dur > e.Max {
```
trianglesphere: Also saturating here. Feels like b/c of this cap, it's not following the behavior indicated in the comments.

trianglesphere (approved):
Changes look good. Only one question about the second `dur > e.Max`

op-service/retry/strategies.go
protolambda: fixed, and also removed an unnecessary `* float64(time.Nanosecond)`

PR #7692, opened 2023-10-16
contracts-bedrock: L2 contracts reinit

tynes:
**Description**

Ensures that the L2 contracts cannot be reinitialized by making their storage value for `_initialze` match the value that is hardcoded in the contract. Since the call to `initialize()` never actually runs in the L2 genesis generation, we only have contract over setting state. Trying to call `reinitialize` again with the same value will cause it to revert.

Also remove the unnecessary argument to the `L2ERC721Bridge`. This was noticed by trail of bits during the audit. It adds extra assurance that the `initialize` functions take no arguments on L2, meaning that even if they could be reinitialized, nothing horrible could happen. Note that this is very application specific and would result in nothing horrible happening based on how the optimism contracts are designed. We should stil not allow for contracts to be reinitialized.

In the far future, we will move towards removing the concept of `initialize` completely by breaking storage layout migrations into their own upgrade step.


packages/contracts-bedrock/test/CommonTest.t.sol
maurelian, line 1: Comment only: I guess we don't explicitly test the L2 ERC721 Bridge or L2 xDM. That's not unreasonable given the implementations are nearly identical, but it is something we should consider in future test refactoring.

op-chain-ops/genesis/config.go
```diff
@@ -26,6 +26,11 @@ import (
 	"github.com/ethereum-optimism/optimism/op-service/eth"
 )
 
+// initialzedValue represents the `Initializable` contract value. It should be kept in
+// sync with the constant in `Constants.sol`.
+// https://github.com/ethereum-optimism/optimism/blob/b6104508cb11ac85252484e212c199acf512a874/packages/contracts-bedrock/src/libraries/Constants.sol#L49
+const initialzedValue = 3
```
mds1: Typo in variable name
```suggestion
// initializedValue represents the `Initializable` contract value. It should be kept in
// sync with the constant in `Constants.sol`.
// https://github.com/ethereum-optimism/optimism/blob/b6104508cb11ac85252484e212c199acf512a874/packages/contracts-bedrock/src/libraries/Constants.sol#L49
const initializedValue = 3
```

op-chain-ops/genesis/config.go
```diff
@@ -26,6 +26,11 @@ import (
 	"github.com/ethereum-optimism/optimism/op-service/eth"
 )
 
+// initialzedValue represents the `Initializable` contract value. It should be kept in
+// sync with the constant in `Constants.sol`.
+// https://github.com/ethereum-optimism/optimism/blob/b6104508cb11ac85252484e212c199acf512a874/packages/contracts-bedrock/src/libraries/Constants.sol#L49
```
mds1: Should we use a regular link to `Constants.sol` instead of a permalink? Linking to this commit may make it look like `3` is always the most up-to-date value

mds1 (commented):
LGTM, just small comments

op-chain-ops/genesis/config.go
tynes: That is a great catch

op-chain-ops/genesis/config.go
mds1, line 1: Thoughts on adding a simple shell script in CI that parses the text in this file + `Constants.sol` to extract and compare the initialized values, and exit with an error if they don't match? This way we ensure they stay in sync

tynes: This is definitely a good idea in the meantime before we can move away from the initialize pattern

tynes:
This now needs a rebase

tynes:
Creating a ticket to track the CI assertion work, merging this now

PR #7698, opened 2023-10-16
chore(ctb): Only record gas costs of benchmark tests

maurelian:
**Description**

The gas snapshot file has reach a size such that it is difficult to use for the purpose of tracking gas costs overtime. 
Moreover, we should only be interested in the costs of calling the implementation contracts. As The snapshot must be updated after any changes to a foundry test.

To the extent that we wish to measure and track gas cost changes in our implementation code, we should write short and simple benchmark tests which wrap the functions we wish to measure with as little gas overhead as possible.

We already have this in `BenchmarkTest.t.sol`, so this change simply makes that approach the 'one right way' to record gas costs.

**Credit**

@mds1 suggested this approach in conversation.


packages/contracts-bedrock/package.json
```diff
@@ -21,8 +21,7 @@
     "test": "pnpm build:differential && pnpm build:fuzz && forge test",
     "coverage": "pnpm build:differential && pnpm build:fuzz && forge coverage",
     "coverage:lcov": "pnpm build:differential && pnpm build:fuzz && forge coverage --report lcov",
-    "gas-snapshot:no-build": "forge snapshot --no-match-test 'testDiff|testFuzz|invariant|generateArtifact'",
-    "gas-snapshot": "pnpm build:differential && pnpm build:fuzz && pnpm gas-snapshot:no-build",
+    "gas-snapshot": "forge snapshot --match-contract GasBenchMark",
```
maurelian: Although all the tests that this picks up are currently stored in `Benchmark.t.sol`, this will catch any contract with the string `GasBenchMark` in it, enabling benchmarks to be written in any test file.

packages/contracts-bedrock/package.json
```diff
@@ -21,7 +21,7 @@
     "test": "pnpm build:differential && pnpm build:fuzz && forge test",
     "coverage": "pnpm build:differential && pnpm build:fuzz && forge coverage",
     "coverage:lcov": "pnpm build:differential && pnpm build:fuzz && forge coverage --report lcov",
-    "gas-snapshot:no-build": "forge snapshot --no-match-test 'testDiff|testFuzz|invariant|generateArtifact'",
+    "gas-snapshot:no-build": "forge snapshot --match-contract GasBenchMark",
```
maurelian, line 24: This will catch any contract with the string `GasBenchMark` in it, enabling benchmarks to be written in any test file.

clabby:
I'd prefer to keep the way that we do this now - is this causing anyone explicit trouble?

It's nice to be able to track gas consumption differences among unit tests in SC PRs. While they aren't entirely accurate because they aren't isolated measurements, the current gas snapshot gives a rough idea on the positive or negative correlation between the changes and gas consumption across a much wider range of codepaths than our `GasBenchmark` tests. With this, we'd have to make alternative gas benchmarks for each unit test we'd like to measure.

maurelian:
Converting to draft pending further discussion

PR #7707, opened 2023-10-16
Canyon: Blocks V2 P2P & Testing

danyalprout:
**Description**
P2P changes for this [issue](https://github.com/ethereum-optimism/optimism/issues/6728).

* Add blocksv2 p2p layer 
     * New blocksV2 topic for post Shanghai/Canyon blocks
     * SSZ encoding changes to support v2 blocks
* Canyon testing
    * Flag to enable / `OP_E2E_USE_CANYON=true` to enable the tests to use Canyon
    * Op Geth testing (withdrawals + push0 opcode)
* Set withdrawals for new payloads to `[]` 

**TODO**
* Add an e2e transition test Regolith to Canyon
* Add a build-server check where `OP_E2E_USE_CANYON=true`

### DIFF
--- a/op-node/p2p/sync.go
+++ b/op-node/p2p/sync.go
@@ -621,8 +621,10 @@ func (s *SyncClient) doRequest(ctx context.Context, id peer.ID, expectedBlockNum
 		return fmt.Errorf("failed to read response: %w", err)
 	}
 
+	expectedBlockTime := s.cfg.TimestampForBlock(expectedBlockNum)
+
 	blockVersion := eth.BlockV1
-	if s.cfg.IsCanyon(expectedBlockNum) {
+	if s.cfg.IsCanyon(expectedBlockTime) {
 		blockVersion = eth.BlockV2
 	}
 	var res eth.ExecutionPayload
--- a/op-node/p2p/sync.go
+++ b/op-node/p2p/sync.go
@@ -571,7 +571,7 @@ func (r requestResultErr) ResultCode() byte {
 	return byte(r)
 }
 
-func (s *SyncClient) doRequest(ctx context.Context, id peer.ID, n uint64) error {
+func (s *SyncClient) doRequest(ctx context.Context, id peer.ID, expectedBlockNum uint64) error {
 	// open stream to peer
 	reqCtx, reqCancel := context.WithTimeout(ctx, streamTimeout)
 	str, err := s.newStreamFn(reqCtx, id, s.payloadByNumber)
@@ -582,8 +582,8 @@ func (s *SyncClient) doRequest(ctx context.Context, id peer.ID, n uint64) error 
 	defer str.Close()
 	// set write timeout (if available)
 	_ = str.SetWriteDeadline(time.Now().Add(clientWriteRequestTimeout))
-	if err := binary.Write(str, binary.LittleEndian, n); err != nil {
-		return fmt.Errorf("failed to write request (%d): %w", n, err)
+	if err := binary.Write(str, binary.LittleEndian, expectedBlockNum); err != nil {
+		return fmt.Errorf("failed to write request (%d): %w", expectedBlockNum, err)
 	}
 	if err := str.CloseWrite(); err != nil {
 		return fmt.Errorf("failed to close writer side while making request: %w", err)
@@ -620,14 +620,20 @@ func (s *SyncClient) doRequest(ctx context.Context, id peer.ID, n uint64) error 
 	if err != nil {
 		return fmt.Errorf("failed to read response: %w", err)
 	}
+
+	blockVersion := eth.BlockV1
+	if s.cfg.IsCanyon(expectedBlockNum) {
+		blockVersion = eth.BlockV2
+	}
 	var res eth.ExecutionPayload
-	if err := res.UnmarshalSSZ(uint32(len(data)), bytes.NewReader(data)); err != nil {
+	if err := res.UnmarshalSSZ(blockVersion, uint32(len(data)), bytes.NewReader(data)); err != nil {
 		return fmt.Errorf("failed to decode response: %w", err)
 	}
+
 	if err := str.CloseRead(); err != nil {
 		return fmt.Errorf("failed to close reading side")
 	}
-	if err := verifyBlock(&res, n); err != nil {
+	if err := verifyBlock(&res, expectedBlockNum); err != nil {
 		return fmt.Errorf("received execution payload is invalid: %w", err)
 	}
 	select {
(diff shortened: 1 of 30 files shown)


op-service/eth/ssz_test.go
```diff
@@ -122,28 +204,131 @@ func TestOPB01(t *testing.T) {
...
 	}
-	var buf bytes.Buffer
-	_, err := payload.MarshalSSZ(&buf)
-	require.NoError(t, err)
-	buf.Reset()
 
-	payload = &ExecutionPayload{
-		ExtraData: make([]byte, math.MaxUint32-executionPayloadFixedPart+1),
+	for _, test := range tests {
+		// First, test the maximum len - which in this case is the max uint32
```
tynes: do you need `test := test`?

specs/rollup-node-p2p.md
```diff
@@ -246,10 +246,13 @@ The extended validator emits one of the following validation signals:
...
 ## Gossip Topics
+There are two topics for distributing blocks to other nodes faster than proxying through L1 would. These are:
 
-### `blocks`
+### `blocksv1`
+Pre-Canyon/Shanghai blocks are broadcast on `/optimism/<chainId>/0/blocks`.
 
-The primary topic of the L2, to distribute blocks to other nodes faster than proxying through L1 would.
+### `blocksv2`
+Post-Canyon/Shanghai blocks are broadcast on `/optimism/<chainId>/0/blocks`.
```
tynes, line 260: This is the same as `blocksv1`?

trianglesphere: The block content is different between v1 & v2 (withdrawals field). P2P uses SSZ and I think we need to separate because otherwise we would start sending blocks that couldn't be parsed (at least that's why there is the V1/V2 split in the engine API).

I'm not sure if it's easy to have a SSZ parser which tells the difference between a ExecutionPayload with withdrawals versus one without withdrawals.

danyalprout: Sorry -- that's a typo in the docs (just fixed). We're setting up two topics [here](https://github.com/ethereum-optimism/optimism/pull/7707/files#diff-03658205f9bcc2e4c45b33bcf3ce3e0e1a194cd6b6399303933ec50abe9bc469R500-R513).

op-service/eth/ssz_test.go
danyalprout: I don't believe so, as we're not capturing it in anything that can run async.

op-node/p2p/sync.go
```diff
@@ -620,14 +620,20 @@ func (s *SyncClient) doRequest(ctx context.Context, id peer.ID, n uint64) error
 	if err != nil {
 		return fmt.Errorf("failed to read response: %w", err)
 	}
+
+	blockVersion := eth.BlockV1
+	if s.cfg.IsCanyon(expectedBlockNum) {
```
trianglesphere: The activation is done by timestamp, not block number

trianglesphere (commented):
We also need to add a check to https://github.com/ethereum-optimism/optimism/blob/56c7254726619826cc5647250922a2c56ecd0d39/op-node/rollup/derive/engine_consolidate.go#L17 to ensure that we verify that the withdrawals from gossip are the same as we created locally (i.e. empty).

We could even completely disable withdrawals and require them to be empty (but not-nil) for the comment b/c we don't have a need for them. It kinda looks like deposits, but our deposits can do much more than withdrawals.

op-service/eth/ssz.go
```diff
@@ -9,16 +9,33 @@ import (
...
 // This is more efficient than RLP, and matches the L1 consensus-layer encoding of ExecutionPayload.
 
 // All fields (4s are offsets to dynamic data)
-const executionPayloadFixedPart = 32 + 20 + 32 + 32 + 256 + 32 + 8 + 8 + 8 + 8 + 4 + 32 + 32 + 4
+const blockV1FixedPart = 32 + 20 + 32 + 32 + 256 + 32 + 8 + 8 + 8 + 8 + 4 + 32 + 32 + 4
+
+// V1 + Withdrawals offset
+const blockV2FixedPart = blockV1FixedPart + 4
+
+const withdrawalSize = 8 + 8 + 32 + 8
```
trianglesphere: Could it be `8 + 8 + 20 + 8`?
The withdrawal spec appears to point to the 3rd field being an address. https://github.com/ethereum/consensus-specs/blob/ef434e87165e9a4c82a99f54ffd4974ae113f732/specs/capella/beacon-chain.md#withdrawal

cc @ajsutton

trianglesphere (commented):
I've reviewed everything. There's a couple things to look into, but broadly looks good.

op-service/eth/ssz.go
ajsutton: Yeah it winds up being `UInt64, Uint64, Vector[byte, 20], UInt64`.  Teku's schema calculates it as having 44 bytes so `8 + 8 + 20 + 8`.

op-program/client/l2/engineapi/l2_engine_api.go
```diff
@@ -350,6 +350,25 @@ func (ea *L2EngineAPI) forkchoiceUpdated(ctx context.Context, state *eth.Forkcho
 	return valid(nil), nil
 }
 
+func toGethWithdrawals(payload *eth.ExecutionPayload) []*types.Withdrawal {
+	if payload.Withdrawals == nil {
+		return nil
+	}
+
+	result := []*types.Withdrawal{}
```
sebastianst:
```suggestion
	result := make([]*types.Withdrawal, 0, len(payload))
```
leads to more efficient memory alloc

op-service/eth/ssz.go
danyalprout: Excellent catch -- thank you 🙏 Just fixed this.

op-program/client/l2/engineapi/l2_engine_api.go
danyalprout: 👍 Added this in the latest commit.

op-node/p2p/sync.go
danyalprout: Good catch, thanks. I've just added a function to go from blockNumber to blockTimestamp on the config. `TimestampForBlock`. 

The specs lead me to believe that this should be deterministic. Would appreciate a sanity check on this, as I maybe e misunderstanding.

trianglesphere: It is deterministic. Code looks correct. I skimmed it & there doesn't seem to be an easy way to introduce this otherwise.

trianglesphere (approved):
TYVM for implementing this

PR #7712, opened 2023-10-17
op-challenger: Move away from generated bindings and batch load claims.

ajsutton:
**Description**

Introduces `contracts.FaultDisputeGameContract` that uses the contract ABI, but not the generated go code to read contract data. Since we're using the ABI directly we can also leverage batch RPC calls when retrieving claims.

The `MultiCaller` struct encapsulates the details of working with the ABI and batching calls to keep the wrapper code we need to write pretty simple.

The tests for this are pretty messy at the moment, but show that we can reasonably test that the calls we're making are actually compatible with the contract ABI.  If you pass a `string` to something that expects a `*big.Int` the test fails, similarly if the method name is wrong or if the wrong number of arguments are provided etc. So we still get a strong assertion that the calls are compatible with the actual contracts. The down side is we have to write tests for each call rather than just depending on the generation being correct.

There are some pretty big advantages to this approach though:

1. The same approach works whether you use batched calls or not. The generated go bindings only support single calls and you have to switch to using the ABI directly (like this code) to use batching.
2. We have a proper adapter layer (or in domain driven development terms an anti-corruption layer) that we control between our program code and the contracts.
   * The go APIs we expose from that layer can be much nicer and more natural. For example we have `GetMaxDepth(ctx)` rather than `MAXDEPTH(&bind.CallOpts{Context: ctx})`
   * If the contract API changes we may be able to adapt to it with only changes to our adapter code instead of having that change ripple through large parts of the code base (especially for things like contract method renames)
3. We are in a much better position to support multiple versions of the contracts with the same go API, just by having different versions of the adapter.  And even the different adapters may only need to vary for a few method calls and can still reuse a lot of code.
4. It's easy for us to support specifying any block we want as reference (ie specifying the block by hash which can't be done with the generated go bindings).

The major downside obviously is having to write code rather than just autogenerating it. I think the production code is quite straight forward to write so I'm not too worried about that - we'd need to do some work on the testing side to make it easier to write the appropriate tests, but I think that's probably doable.

We also need to look at how to integrate this with sending tx via the txmgr, but those already use the ABI directly so I'm hoping it will be a nice fit.

https://github.com/ethereum-optimism/client-pod/issues/154


op-challenger/game/fault/contracts/multicall.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+type ContractCall struct {
+	Abi    *abi.ABI
+	Addr   common.Address
+	Method string
+	Args   []interface{}
+}
+
+func NewContractCall(abi *abi.ABI, addr common.Address, method string, args ...interface{}) *ContractCall {
```
trianglesphere: Go bindings wrap this into a bound contract & then you can create a specific call by only passing the methods & args. I kinda like it, but also have both or this isn't bad.

trianglesphere (commented):
I like this approach. Interacting with the ABI doesn't seem bad & we'd get a lot more flexibility and control.

op-challenger/game/fault/contracts/multicall.go
ajsutton: I'd say the aim of `NewContractCall` is to be the low level building block that lets you manually build the wrapper that provides a nice API (which is the `FaultDisputeGameContract` in this example). It's mostly nice to not have to repeat the rather long winded and very error-prone interactions with the ABI to pack args and `ContractCall` handles that for you (in a way that then works with making multiple calls in a single batch request)

trianglesphere: Talking in code, I'm suggesting the following as syntatic sugar/convenience

```
type BoundContract struct {
    Abi  *abi.ABI
    Addr common.Address
}

func (b *BoundContract) NewContractCall(method string, args ...interface{}) *ContractCall {
    return &ContractCall{
		Abi:    b.abi,
		Addr:   b.addr,
		Method: method,
		Args:   args,
	}
}


```
It's not a huge deal, but I think it makes it slightly nicer in some cases.

ajsutton: oh sorry, yeah I completely misunderstood what you were saying. :)

Yes that's a great idea. I suspect we'd actually always use the `BoundContract` type approach since you'd create `BoundContract` when creating the wrapper (eg `FaultDisputeGameContract`) and would then naturally use it everywhere. Both options would pretty naturally be supported though I think.

ajsutton:
I think it's worth merging this now - tests have been added and it moves the read part of the fault dispute game contract over to the new style.  The `ContractCall` is in a good place to start using it with the txmgr to send transactions, though there's still some room for experimentation of the best way to fully integrate that.  Next step though is probably to convert the read calls of the dispute game factory over to this new style.

op-service/sources/batching/call.go
```diff
@@ -0,0 +1,115 @@
...
+	return &CallResult{out: out}, nil
+}
+
+func toCallArg(msg ethereum.CallMsg) interface{} {
+	arg := map[string]interface{}{
+		"from": msg.From,
+		"to":   msg.To,
+	}
+	if len(msg.Data) > 0 {
+		arg["data"] = hexutil.Bytes(msg.Data)
```
trianglesphere: Is it worth already switching this to `input`?

trianglesphere (approved):

op-service/sources/batching/call.go
ajsutton: Yes, good spot. That changed in geth while this PR was in draft. :)

PR #7715, opened 2023-10-17
ops: shared cached go docker build

protolambda:
**Description**

- Merge the Go Dockerfile services into one; this helps de-duplicate the Go build work, since there are so many shared Go packages. It adds very little time to the total step, with the benefit of greatly reducing computation and downloading work, and a single docker go-mod/build cache.
- The individual go services still have their own dockerfile and individual build, but simply copy the necessary binary from the above.
- Utilize the Go mod cache and Go build cache in docker builds
- Update CI to use the above docker build
- Share the docker build CI step with the devnet CI step, so the devnet does not have to re-build all docker containers, and we devnet-test exactly what was built in docker.
- The cannon pre-state is now a job that persists its results to the workspace, so it can be reused by other jobs
- The devnet-allocs is now a job that persists its results to the workspace, so it can be reused by other jobs
- The Go linting has been merged into one job: with the go caches, the go linter cache, and so many common packages to analyze it makes little sense to add different linting jobs when one suffices.
- The geth install now checks if the existing install is the right version, and doesn't re-install if it's already there and matches.

The overall CI time is still slow-ish due to the docker-caches not being fruitful CI yet; it works across jobs, but not across workflows, and the worst-case path is thus still in the 20+min range. But compute credits will be lower, so maybe we can start using more powerful/faster/expensive CI instances.

In a follow-up we can also try to do manual docker-layer caching, to save some time across workflows by putting overlay2 in the CI cache, and thus speeding up the builds a bunch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -168,13 +168,42 @@ jobs:
         description: Docker repo
         type: string
         default: "oplabs-tools-artifacts/images"
+      build_args:
+        description: Docker build args
+        type: string
+        default: ""
+      load_base_image:
+        description: Load docker image as base
+        type: string
+        default: ""
+      save_image_tag:
+        description: Save docker image as
+        type: string
+        default: ""
     machine:
       image: ubuntu-2204:2022.07.1
       resource_class: medium
+      docker_layer_caching: true
     steps:
       - checkout
+      - restore_cache:
+          name: Restore docker build cache
+          key: docker-build-cache
+      - attach_workspace:
+          at: /tmp/docker_images
       - run:
           command: mkdir -p /tmp/docker_images
+      - when:
+          condition: "<<parameters.load_base_image>>"
+          steps:
+            - run:
+                name: Load OP-Stack Go base image
+                command: |
+                  docker load < "/tmp/docker_images/<<parameters.load_base_image>>.tar"
+      - run:
+          name: build args
+          command: |
+            echo "build args: <<parameters.build_args>>"
       - run:
           name: Build
           command: |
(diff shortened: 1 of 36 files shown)


.circleci/config.yml
```diff
@@ -771,6 +798,10 @@ jobs:
     resource_class: xlarge
     steps:
       - checkout
+      - restore_cache:
+          key: golang-build-cache
+          paths:
+            - "/root/.cache/go-build"
```
tynes: Should this be `$HOME` instead of `root` to be more portable?

ops-bedrock/docker-compose.yml
```diff
@@ -42,11 +52,15 @@ services:
 
   op-node:
     depends_on:
+      - op_stack_go_builder
       - l1
       - l2
     build:
       context: ../
       dockerfile: ./op-node/Dockerfile
+      args:
```
tynes, line 61: Are you sure this will work for local dev? ie you make the change locally and then want to run it in the devnet?

tynes (approved):

op-batcher/Dockerfile
```diff
@@ -1,32 +1,9 @@
...
-COPY ./op-service /app/op-service
-
-COPY ./.git /app/.git
-
-WORKDIR /app/op-batcher
-
-ARG TARGETOS TARGETARCH
-
-RUN make op-batcher VERSION="$VERSION" GOOS=$TARGETOS GOARCH=$TARGETARCH
+ARG OP_STACK_GO_BUILDER=us-docker.pkg.dev/oplabs-tools-artifacts/images/op_stack_go:latest
```
trianglesphere, line 1: This pattern means that `docker build -f op-batcher/Dockerfile .` is completely broken at the moment & not likely what the developer wants (because it's pulling the latest tag, not building the docker image).

Doing a shared build & multi-step build seems really nice, but this behavior is pretty weird.

op-batcher/Dockerfile
```diff
@@ -1,32 +1,9 @@
...
+FROM $OP_STACK_GO_BUILDER as builder
```
roninjin10, line 2: This feels overly complicated compared to just doing a multistage build in a single dockerfile

protolambda: We may merge it into a single dockerfile and do the target-selection thing to output a single stage yes

roninjin10: Oh nice I see this is more of a step in that direction. Nice improvement overall!

PR #7729, opened 2023-10-18
set of changes for it to get `make devnet-test` to work

h4ck3rk3y:
These are hardcoded but we can clean it up and pass as environment variables


tynes:
What is the purpose of this PR?

PR #7734, opened 2023-10-18
contracts-bedrock: update gas snapshot

tynes:
**Description**

The gas snapshot has changed because of the new foundry
version. This commit updates the gas snapshot so that
CI will pass. Every open PR will need to rebase on top
of this to have the smart contract gas snapshot check pass.


maurelian (approved):
Surprisingly big jump in some tests, I wonder what they did.

clabby (approved):

PR #7750, opened 2023-10-19
op-node/rollup: remove redundant nil check in engine_queue

joohhnnn:
`tryNextSafeAttributes` is only used in func `func (eq *EngineQueue) Step`, and already check in  `func (eq *EngineQueue) Step`


ajsutton (commented):
Is there a particular reason to remove this check? It seems like a good defensive check so we don't get surprises if `tryNextSafeAttributes` gets called from a different place in the future.

joohhnnn:
No particular reason here, just walk around in the code base and found it check twice, feel free to close this PR :)

ajsutton:
Cool, thanks for submitting, always good to review if we can tidy things up but in this case I think the defensive value is worth it.

PR #7754, opened 2023-10-19
feat(proxyd): avoid caching debug_getRawReceipts responses with 0 receipts

felipe-op:
**Description**

Recently we experienced an unusual situation in which a backend responded `debug_getRawReceipts` with 0 receipts and the client was expecting a `> 0` receipt count.
We believe the backend that served the request was providing bad data.

This change avoids caching bad responses like this. 

It also adds a config to serve an `x-served-by` header, so we can introspect where the bad data came from.

**Tests**

Tested locally + new test cases


Inphi (approved):
Nice!

PR #7755, opened 2023-10-19
op-node: on fetching error, accurately report the L1 block that it fails at

protolambda:
**Description**

The error was logging the parent block reference as if it's the current block that failed, while the parent was already successfully fetched.

In the case of the earlier goerli network issue, it was reporting an issue with `0xdc6cd3bd7c41f7e2cf0d7986edf89718cb99ada79fe9fed87c436e79d2a52875:9876474` expecting 67 transactions, but `9876474` is actually the parent block of the one that failed and has 92 as expected. It is block `9876475` that has the 67 transactions that the inner error says it expected, while getting 0 from the L1 RPC.
So the op-node was behaving correctly, but reporting the error badly.


trianglesphere (approved):

Inphi (approved):
aish! ty

PR #7759, opened 2023-10-19
op-service,op-node: track proxyd debug info

protolambda:
**Description**

Introduces an "HTTP Interceptor"; a bit of an anti-pattern, but it allows us to modify/inspect HTTP requests by the underlying HTTP client of the geth RPC.

With the websocket RPC it might capture the initial WS upgrade, if connecting from HTTP endpoint with upgrade. Untested.

This allows us to track Proxyd HTTP headers, and implement more future RPC-debugging utils (many services use RPC but hae little insight in the underlying server data).

This intercepting of the requests/responses should not be used for control-flow, but can provide very valuable debug information.

If we make interceptor support the default, perhaps we can implement HTTP metrics and more, to attach to all Json RPC clients.

This implements a new `--l1.proxyd-debug` flag that enables the intercepting & error logging.

**Tests**
TODO: this is quite experimental still. We should add some op-e2e tests maybe. I did add unit-tests for the HTTP interceptor.

**Additional context**

Feature request by @felipe-op

### DIFF
--- a/op-service/httputil/proxyd_inspector.go
+++ b/op-service/httputil/proxyd_inspector.go
@@ -1,0 +1,28 @@
+package httputil
+
+import (
+	"net/http"
+
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// ProxydInspector implements a HTTPInterceptor to monitor cache headers in HTTP responses.
+type ProxydInspector struct {
+	ServedBy          string
+	ProxydCacheStatus string
+}
+
+var _ HTTPInterceptor = (*ProxydInspector)(nil)
+
+func (chi *ProxydInspector) Intercept(req *http.Request, inner http.RoundTripper) (resp *http.Response, err error) {
+	resp, err = inner.RoundTrip(req)
+	if resp != nil {
+		chi.ServedBy = resp.Header.Get("X-Served-By")
+		chi.ProxydCacheStatus = resp.Header.Get("X-Proxyd-Cache-Status")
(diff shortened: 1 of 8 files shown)


clabby:
apologies ^ misclicked on mobile.

op-service/httputil/proxyd_inspector.go
```diff
@@ -0,0 +1,28 @@
...
+	ServedBy          string
+	ProxydCacheStatus string
+}
+
+var _ HTTPInterceptor = (*ProxydInspector)(nil)
+
+func (chi *ProxydInspector) Intercept(req *http.Request, inner http.RoundTripper) (resp *http.Response, err error) {
+	resp, err = inner.RoundTrip(req)
+	if resp != nil {
+		chi.ServedBy = resp.Header.Get("X-Served-By")
```
felipe-op, line 20: lower case: `x-served-by`
nit: const

op-service/httputil/proxyd_inspector.go
```diff
@@ -0,0 +1,28 @@
...
+		chi.ProxydCacheStatus = resp.Header.Get("X-Proxyd-Cache-Status")
```
felipe-op, line 21: nit: const

felipe-op (changes requested):

felipe-op (changes requested):
we should log the headers after validateReceipts return false

PR #7760, opened 2023-10-20
op-chain-ops: add additional assertions the `check-l2`

tynes:
**Description**

The `check-l2` script works against new networks but not legacy networks. We need a solution that can work against both new networks and legacy networks. This adds test coverage for new networks as `check-l2` runs in CI against the local devnet.

Add test coverage for reinitialize on L2 contracts. Also ensures that the `_initialized` value is set as expected.

### DIFF
--- a/op-chain-ops/cmd/check-l2/main.go
+++ b/op-chain-ops/cmd/check-l2/main.go
@@ -13,6 +13,7 @@ import (
 	"github.com/mattn/go-isatty"
 	"github.com/urfave/cli/v2"
 
+	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 
 	"github.com/ethereum-optimism/optimism/op-bindings/bindings"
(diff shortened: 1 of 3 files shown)


op-chain-ops/cmd/check-l2/main.go
```diff
@@ -429,6 +440,25 @@ func checkL2ERC721Bridge(addr common.Address, client *ethclient.Client) error {
...
+	}
+	calldata, err := abi.Pack("initialize")
+	if err != nil {
+		return err
+	}
+	msg := ethereum.CallMsg{
+		To:   &addr,
+		Data: calldata,
+	}
+	if _, err := client.CallContract(context.Background(), msg, nil); err.Error() != errAlreadyInitialized.Error() {
```
Inphi: A bit brittle relying on error strings here as we upgrade go-ethereum. If we're sure that this error string format is something go-ethereum will continue to maintain in the foreseeable future then disregard this comment.
But I'd be more comfortable with this if we had a Negative Test that asserts the error is returned when an invalid initialized value is used.

op-chain-ops/cmd/check-l2/main.go
```diff
@@ -566,6 +596,25 @@ func checkOptimismMintableERC20Factory(addr common.Address, client *ethclient.Cl
...
+		return err
+	}
```
Inphi: Let's create a function that runs this check, taking an ABI as input, to dedup code.

tynes:
Going to circle back around on this

op-chain-ops/cmd/check-l2/main.go
tynes: @Inphi We cannot do this because `initialize` takes different numbers of args but we could do something similar

op-chain-ops/cmd/check-l2/main.go
tynes: We will soon have solidity test coverage of this that will reduce this from being a critical test

tynes: I updated the logic to be more lenient to changes to geth by using string matching instead of an exact comparison

op-chain-ops/cmd/check-l2/main.go
Inphi: Could use variadics for a slice of `interface{}`: `func initialize(inputs ...interface{})`. It's not any worse type-safety wise as `abi.Pack` already takes an interface slice. But this is OK too.

Inphi (approved):

PR #7761, opened 2023-10-20
Support optional whitespace in XFF Header

danyalprout:
**Description**
In XFF headers the whitespace around the commas is [optional](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax).

This PR updates `stripXFF` to support the optional whitespace case along with the `, ` case.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1170,6 +1170,6 @@ func RecordBatchRPCForward(ctx context.Context, backendName string, reqs []*RPCR
 }
 
 func stripXFF(xff string) string {
-	ipList := strings.Split(xff, ", ")
+	ipList := strings.Split(xff, ",")
 	return strings.TrimSpace(ipList[0])
 }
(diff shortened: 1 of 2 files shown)


tynes (approved):

felipe-op:
can we test all outputs? I believe we need to trim the strings from extra spaces

ie not clear if
```
{"1.2.3, 4.5.6, 7.8.9", "1.2.3"},
```
will result in `" 4.5.6"`

felipe-op:
> can we test all outputs? I believe we need to trim the strings from extra spaces
> 
> ie not clear if
> 
> ```
> {"1.2.3, 4.5.6, 7.8.9", "1.2.3"},
> ```
> 
> will result in `" 4.5.6"`

nvm, just saw we actually TrimSpaces and there is a 3rd test case with extraneous spaces

felipe-op (approved):

